PR #25, opened 2015-02-01
Makefile cleanup

juliusv:
This borrows a lot of the latest changes to the node_exporter Makefile, as well as some tricks from the Prometheus one.

It also removes some unused JS/CSS files in a separate commit.

@beorn7


Makefile.INCLUDE
```diff
@@ -13,11 +11,39 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-.SUFFIXES:
+VERSION  := 0.1.0
```
beorn7: For consistency, do we want perhaps a VERSION file in every repo?

beorn7:
:+1: (The VERSION thing is more a long-term vision thing. But feel free to adopt it if you like it.)

juliusv:
Agree with the VERSION comment in general, but let's do a general cleanup sweep through the repos for that at some point.

PR #27, opened 2015-02-08
Make flag names consistent across projects.

juliusv:
@beorn7

### DIFF
--- a/web/web.go
+++ b/web/web.go
@@ -29,8 +29,8 @@ import (
 
 // Commandline flags.
 var (
-	listenAddress  = flag.String("listenAddress", ":9090", "Address to listen on for web interface.")
-	useLocalAssets = flag.Bool("useLocalAssets", false, "Read assets/templates from file instead of binary.")
+	listenAddress  = flag.String("web.listen-address", ":9093", "Address to listen on for the web interface and API.")
+	useLocalAssets = flag.Bool("web.use-local-assets", false, "Serve assets and templates from local files instead of from the binary.")
 )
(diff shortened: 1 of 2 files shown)


web/web.go
```diff
@@ -29,8 +29,8 @@ import (
 
 // Commandline flags.
 var (
-	listenAddress  = flag.String("listenAddress", ":9090", "Address to listen on for web interface.")
-	useLocalAssets = flag.Bool("useLocalAssets", false, "Read assets/templates from file instead of binary.")
+	listenAddress  = flag.String("web.listen-address", ":9090", "Address to listen on for the web interface and API.")
```
brian-brazil: If we're going to be changing this flag name I think we should also ensure all the default ports are unique across projects while we're at it.

juliusv: Good point... I'll think of a list of ports and update the PRs.

beorn7:
:+1:

web/web.go
```diff
@@ -29,8 +29,8 @@ import (
 
 // Commandline flags.
 var (
-	listenAddress  = flag.String("listenAddress", ":9090", "Address to listen on for web interface.")
-	useLocalAssets = flag.Bool("useLocalAssets", false, "Read assets/templates from file instead of binary.")
+	listenAddress  = flag.String("web.listen-address", ":9093", "Address to listen on for the web interface and API.")
```
juliusv: Changed this from 9092 -> 9093 to avoid collision with Apache Kafka.

PR #29, opened 2015-02-09
Expand alertmanager README.md

juliusv:
@brian-brazil @discordianfish


README.md
```diff
@@ -1,6 +1,63 @@
...
+
+    make test
+
+## Caveats and roadmap
+
+Alertmanager is still in an experimental state. Some of the known caveats which
+are going to be addressed in the future:
+
+* Alertmanager is run as a single instance and does not provide high
+  availability yet. We plan on using a consensus protocol between multiple
```
brian-brazil: With alerts you want AP, saying "consensus" implies we're going for C.

brian-brazil:
:+1:

README.md
juliusv: Ah. Didn't other solutions in this space use Paxos? What approach/formulation would you think makes most sense here?

brian-brazil: Paxos isn't appropriate for this sort of usage, as it can't handle partition. A single completely isolated alertmanager needs to be able to send alerts.

I'd have them all talk to each other, and choose a consistent leader (e.g. lowest IP). We'd then need some way to synchronise notification and silence state across them, likely using CRDTs so we get sane eventual consistency.

juliusv: That makes sense. Gossip/share the current alert states to every other node, but sync on notification/silence states. I just changed it to the vague "We plan on clustering multiple replicated Alertmanager instances" for now.

discordianfish:
LGTM

PR #32, opened 2015-02-16
Add From/To (non-envelope) headers to email notifications.

juliusv:


brian-brazil:
:+1: 

When we allow custom templates this will get more interesting.

PR #33, opened 2015-02-16
add pushover as notification method

antonlindstrom:
This adds https://pushover.net/ as a way of notification.

I'm not sure if this is entirely useful for everybody but I saw myself using it so I thought I'd contribute.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,9 @@ following aspects:
 * inhibiting alerts based on alert dependencies
 * aggregating alerts by labelset
 * handling notification repeats
-* sending alert notifications via external services (currently email or [PagerDuty](http://www.pagerduty.com/))
+* sending alert notifications via external services (currently email,
+[PagerDuty](http://www.pagerduty.com/) or
+[Pushover](https://www.pushover.net/))
 
 See [config/fixtures/sample.conf.input](config/fixtures/sample.conf.input) for
 an example config. The full configuration schema including a documentation for
--- a/config/config.go
+++ b/config/config.go
@@ -56,6 +56,14 @@ func (c Config) Validate() error {
 				return fmt.Errorf("Missing email address in email notification config: %s", proto.MarshalTextString(ec))
 			}
 		}
+		for _, ec := range nc.PushoverConfig {
+			if ec.Token == nil {
+				return fmt.Errorf("Missing token in Pushover notification config: %s", proto.MarshalTextString(ec))
+			}
+			if ec.UserKey == nil {
+				return fmt.Errorf("Missing user key in Pushover notification config: %s", proto.MarshalTextString(ec))
+			}
+		}
 
 		if _, ok := ncNames[nc.GetName()]; ok {
 			return fmt.Errorf("Notification config name not unique: %s", nc.GetName())
--- a/manager/notifier.go
+++ b/manager/notifier.go
@@ -26,6 +26,7 @@ import (
 	"text/template"
 
 	"github.com/golang/glog"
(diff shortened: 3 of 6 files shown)


config/config.go
```diff
@@ -56,6 +56,14 @@ func (c Config) Validate() error {
 				return fmt.Errorf("Missing email address in email notification config: %s", proto.MarshalTextString(ec))
 			}
 		}
+		for _, ec := range nc.PushoverConfig {
+			if ec.Token == nil {
+				return fmt.Errorf("Missing token in Pushover notification config: %s", proto.MarshalTextString(ec))
+			}
+			if ec.UserKey == nil {
+				return fmt.Errorf("Missing userkey in Pushover notification config: %s", proto.MarshalTextString(ec))
```
juliusv: I'd say "userkey" -> "user key" in a message like this.

manager/notifier.go
```diff
@@ -190,6 +191,15 @@ func (n *notifier) sendEmailNotification(email string, a *Alert) error {
 	return writeEmailBody(wc, a)
 }
 
+func (n *notifier) sendPushoverNotification(token, userKey string, a *Alert) error {
+	po, err := pushover.NewPushover(token, userKey)
+	if err != nil {
+		return err
+	}
+
+	return po.Message("Alertmanager " + a.Description)
```
juliusv: Hmm, looking at https://pushover.net/api#messages, pushover seems to support titles for messages, which would be great to put the alert Summary in. But I guess the problem is that the client library so far doesn't support setting a title?

juliusv:
Thanks, looks great otherwise! :+1: 

(yeah, we should add tests both for this and PagerDuty soon...)

juliusv:
And yeah, I think in general, we want to support quite a few notification mechanisms, so this is very useful. If it gets too much, we'll split the code up and make it modular/pluggable...

manager/notifier.go
juliusv: Actually, the client library does support setting titles:

http://godoc.org/github.com/thorduri/pushover#Pushover.Push

So how about:

```go
_, _, err := po.Push(&po.Message{
	Title: a.Summary,
	Message: a.Description,
})
return err
```

(there are potentially other interesting fields in http://godoc.org/github.com/thorduri/pushover#Message)

antonlindstrom:
@juliusv Thanks for the awesome feedback! I've updated the code based on your comments.

Thanks!

README.md
```diff
@@ -9,7 +9,9 @@ following aspects:
 * inhibiting alerts based on alert dependencies
 * aggregating alerts by labelset
 * handling notification repeats
-* sending alert notifications via external services (currently email or [PagerDuty](http://www.pagerduty.com/))
+* sending alert notifications via external services (currently email,
+[PagerDuty](http://www.pagerduty.com/) or
+[Pushover](https://www.pushover.net/))
```
juliusv: Sorry, just noticed this now: There's one closing parentheses too many.

antonlindstrom: I'm not sure if I understand correctly, the last of the parentheses are ending the one between `..services (currently email..` on row 12. Is that the one you mean?

Here's an image on how the markup renders:

screen shot 2015-02-17 at 12 59 04

Thanks!

juliusv: Oh! Sorry. Due to how it got displayed visually by GitHub in the side-by-side diff variant, I totally missed that. All good.

juliusv:
Thanks a bunch!

PR #35, opened 2015-02-23
Fix makefile parallelism

juliusv:


grobie:
Makefile.COMMON?

juliusv:
@grobie Makefile.COMMON needs some additions to work with the extra requirements here. So for now, only this fix.

juliusv:
But Makefile.COMMON has the same problem... (will fix either today or tomorrow - probably tomorrow).

juliusv:
Will merge if no further comments for now?

PR #36, opened 2015-02-23
Makefile: Use $(CURDIR) instead of $(PWD), which is more dependable

atombender:
When building a Debian package with `debuild`, which uses fakeroot, Make doesn't have `$(PWD)` for some reason. Apparently this is a shell-dependent thing. For that reason, the `touch` command gets the wrong path in `$@`:

```
make[1]: Entering directory `/tmp/d20150222-20503-whiv4g/prometheus-alertmanager_0.0git942cd35'
mkdir -p .deps
curl -o .deps/go1.4.1.linux-amd64.tar.gz -L https://golang.org/dl/go1.4.1.linux-amd64.tar.gz
tar -C .deps -xzf .deps/go1.4.1.linux-amd64.tar.gz
touch /.deps/go/bin/go
touch: cannot touch ‘/.deps/go/bin/go’: No such file or directory
make[1]: *** [/.deps/go/bin/go] Error 1
```

(The makefile in `node_exporter` is correct, by the way.)

### DIFF
--- a/Makefile.INCLUDE
+++ b/Makefile.INCLUDE
@@ -31,11 +31,13 @@ else
 RELEASE_SUFFIX ?=
 endif
 
+MAKEFILE_DIR ?= $(realpath $(dir $(lastword $(MAKEFILE_LIST))))
+
 GO_VERSION ?= 1.4.1
 GOURL      ?= https://golang.org/dl
 GOPKG      ?= go$(GO_VERSION).$(GOOS)-$(GOARCH)$(RELEASE_SUFFIX).tar.gz
-GOROOT     := $(PWD)/.deps/go
-GOPATH     := $(PWD)/.deps/gopath
+GOROOT     := $(MAKEFILE_DIR)/.deps/go
+GOPATH     := $(MAKEFILE_DIR)/.deps/gopath
 GOCC       := $(GOROOT)/bin/go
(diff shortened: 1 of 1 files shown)


atombender:
Hold on, this actually breaks the makefile under `web`. Let me fix that.

atombender:
Had to use `realpath` instead. Works now.

web/Makefile
```diff
@@ -18,6 +18,7 @@ all: blob/files.go
 SUFFIXES:
 
 blob/files.go: $(shell find templates/ static/ -type f)
+	echo $(MAKEFILE_DIR)
```
grobie: Debug leftover?

atombender:
@grobie: Sorry, it's late. Removed it.

grobie:
:+1:

The goal is to move all the common makefile beahvior into a Makefile.COMMON eventually which will be authoritatively maintened here: https://github.com/prometheus/utils/blob/master/Makefile.COMMON We use $(CURDIR). Your solution seems to be more correct when executing a target outside of the root dir though.

atombender:
Don't know about correct, but at least I'm not able to build with `debuild` without it. I can only speculate that `$PWD` isn't set under `fakeroot`, although there may be a way to force-set it. At any rate, this change worked for me.

atombender:
Same for [prometheus](https://github.com/prometheus/prometheus/pull/550).

juliusv:
Thanks!

PR #37, opened 2015-03-12
Add -smtpAuth for mail servers that require authentication

logan:
I'm trying out alertmanager in Amazon EC2, and the easiest delivery of notifications I could think of is through Amazon's email sending service. It requires auth, though, so I threw this change together.

### DIFF
--- a/manager/notifier.go
+++ b/manager/notifier.go
@@ -194,6 +194,8 @@ func (n *notifier) sendEmailNotification(to string, a *Alert) error {
 	if hasAuth && username != "" {
 		for _, mech := range strings.Split(mechs, " ") {
 			switch mech {
+			default:
+				continue
 			case "CRAM-MD5":
 				secret := os.Getenv("SMTP_AUTH_SECRET")
 				if secret == "" {
@@ -202,7 +204,6 @@ func (n *notifier) sendEmailNotification(to string, a *Alert) error {
 				if err := c.Auth(smtp.CRAMMD5Auth(username, secret)); err != nil {
 					return fmt.Errorf("cram-md5 auth failed: %s", err)
 				}
-				break
 			case "PLAIN":
 				password := os.Getenv("SMTP_AUTH_PASSWORD")
 				if password == "" {
@@ -219,8 +220,8 @@ func (n *notifier) sendEmailNotification(to string, a *Alert) error {
 				if err := c.Auth(smtp.PlainAuth(identity, username, password, host)); err != nil {
 					return fmt.Errorf("plain auth failed: %s", err)
 				}
-				break
 			}
+			break
 		}
 	}
 
--- a/manager/notifier.go
+++ b/manager/notifier.go
@@ -15,13 +15,16 @@ package manager
 
 import (
 	"bytes"
+	"crypto/tls"
 	"encoding/json"
 	"flag"
 	"fmt"
 	"io"
 	"io/ioutil"
+	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


brian-brazil:
That'd put a password on the command line, you should read the password from a file.

I use a local postfix in null client mode http://www.postfix.org/STANDARD_CONFIGURATION_README.html#null_client

discordianfish:
I would prefer and ENV variable. It's easier to configure and at similar secure.

logan:
How does SMTP_AUTH sound?

logan:
Or maybe I should go for a more complete implementation.

If SMTP_AUTH_USERNAME is given, then check for the AUTH and TLS extensions and a compatible auth mechanism supported by the server.

If the server supports PLAIN auth and SMTP_AUTH_PASSWORD is given, then use PLAIN auth. An optional SMTP_AUTH_IDENTITY will also be passed to the server.

If the server supports CRAM-MD5 auth and SMTP_AUTH_SECRET is given, then use CRAM-MD5 auth.

If neither of these conditions is met, then plain old unencrypted, unauthenticated mail sending takes place.

juliusv:
That sounds good to me (although I'm not an email auth expert, so I'll trust you there or wait for other comments). Also, :+1: to env vars for credentials.

logan:
Ok, this change is ready for another round of review. I've updated the title of the pull request to reflect this new approach.

I don't have a good way to test CRAM-MD5 or the SMTP_AUTH_IDENTITY variable, but I can confirm that PLAIN auth works for me with SMTP_AUTH_USERNAME and SMTP_AUTH_PASSWORD set.

manager/notifier.go
```diff
@@ -184,6 +188,42 @@ func (n *notifier) sendEmailNotification(to string, a *Alert) error {
...
+			switch mech {
+			case "CRAM-MD5":
+				secret := os.Getenv("SMTP_AUTH_SECRET")
+				if secret == "" {
+					continue
+				}
+				if err := c.Auth(smtp.CRAMMD5Auth(username, secret)); err != nil {
+					return fmt.Errorf("cram-md5 auth failed: %s", err)
+				}
+				break
```
juliusv: Wouldn't this `break` only leave the `switch`? From the Go spec: 'A "break" statement terminates execution of the innermost "for", "switch", or "select" statement within the same function.'

I guess your intention is to break out of the `for` loop (not that it would matter in most cases), which could be done via a label?

manager/notifier.go
```diff
@@ -184,6 +188,42 @@ func (n *notifier) sendEmailNotification(to string, a *Alert) error {
...
+			case "PLAIN":
+				password := os.Getenv("SMTP_AUTH_PASSWORD")
+				if password == "" {
+					continue
+				}
+				identity := os.Getenv("SMTP_AUTH_IDENTITY")
+
+				// PLAIN auth requires TLS to be started first.
+				host, _, _ := net.SplitHostPort(*smtpSmartHost)
```
juliusv: To be 100% safe, could you also check the error return value here?

juliusv:
:+1: besides comments.

logan:
Good catch on the switch break issue, I keep forgetting that Go shares that with C. Instead of a label I've added a default continue case, and moved the break to after the switch. If you'd prefer the labeled breaks I can switch to that easily enough.

discordianfish:
LGTM :+1:

juliusv:
:+1: Thanks!

PR #38, opened 2015-03-13
Separate out SMTP auth logic and test it.

logan:
This code feels a lot cleaner in its own function, which now has 100% test coverage.


juliusv:
:+1: Nice, thanks a lot!

PR #41, opened 2015-03-22
docker: run alertmanager as nobody

### LINKED ISSUE #39: Dockerfile: don’t run alertmanager as root
I don’t see a reason for alertmanager to run as root, not even in a docker container. Could we change it so that it runs as nobody/nogroup instead?

stapelberg:
fixes #39


juliusv:
@stapelberg Thanks!

CC @discordianfish

discordianfish:
LGTM

PR #42, opened 2015-03-23
Only serve alerts page from /alerts or /.

juliusv:
This fixes https://github.com/prometheus/prometheus/issues/607


juliusv:
@beorn7

beorn7:
:+1:

PR #40, opened 2015-03-20
Adding functionality to Prometheus to have hipchat alerts

rekhajoshm:
Adding functionality to Prometheus to have hipchat alerts.Thanks.

PR #45, opened 2015-04-10
adding simple support for hipchat

a86c6f7964:


config/config.proto
```diff
@@ -34,6 +34,17 @@ message PushoverConfig {
...
 }
 
+message HipChatConfig {
+  // Hipchat auth token, https://www.hipchat.com/docs/api/auth
+  optional string auth_token = 1;
+  // Hipchat room id, https://www.hipchat.com/rooms/ids
+  optional int32 room_id = 2;
+  // color of message
+  optional string color = 3 [default = "red"];
+  // should this message notify or not
```
brian-brazil: What does this mean?

config/config.proto
```diff
@@ -34,6 +34,17 @@ message PushoverConfig {
   optional string user_key = 2;
 }
 
+message HipChatConfig {
+  // Hipchat auth token, https://www.hipchat.com/docs/api/auth
+  optional string auth_token = 1;
+  // Hipchat room id, https://www.hipchat.com/rooms/ids
+  optional int32 room_id = 2;
+  // color of message
+  optional string color = 3 [default = "red"];
```
brian-brazil: Can we go with something a bit softer by default? Red is a bit jarring.

config/config.proto
a86c6f7964: in hipchat you can send a message to a room, and it will not make a notification for others

brian-brazil: Found the docs, let's default this to no. Waking up everyone when there's an alert is bad practice, you want to leave that to the oncall/interrupts person.

config/config.proto
a86c6f7964: we can. but i guess i was assuming that these are alerts. want to change it to be be 'purple' and 'do not notify by default'?

matthiasr: +1 to that. It's useful to have alerts also got to channels, but this should not be the main method of paging someone.

manager/notifier.go
```diff
@@ -164,6 +165,38 @@ func (n *notifier) sendPagerDutyNotification(serviceKey string, a *Alert) error
 	return nil
 }
 
+func (n *notifier) sendHipChatNotification(authToken string, roomId int32, color string, notify bool, a *Alert) error {
+	// https://www.hipchat.com/docs/apiv2/method/send_room_notification
+	incidentKey := a.Fingerprint()
+	buf, err := json.Marshal(map[string]interface{}{
+		"color":  color,
+		"message":  a.Description,
```
brian-brazil: The description tends to be pretty light, I think we'll want a bit more than that. I'd suggest a lighter version of what goes in the email, and a link (configurable via flag) back to the alertmanager

manager/notifier.go
```diff
@@ -164,6 +165,38 @@ func (n *notifier) sendPagerDutyNotification(serviceKey string, a *Alert) error
...
+		"notify": notify,
+	})
+	if err != nil {
+		return err
+	}
+
+	resp, err := http.Post(
```
brian-brazil: Is there a timeout here?

brian-brazil:
Thanks for this, can you add a unittest?

a86c6f7964:
I am not sure that I know what to test? I guess I could test the message if we want to make it more complicated than just the description.

brian-brazil:
I'd suggest testing that the expected http endpoint is hit.

a86c6f7964:
I could write that test, but that test is kind of silly IMHO. If I copy and paste the URL from the code that I wrote and it is wrong it will be wrong in both the test and the code. Then the test will pass but it will still not work. Really we would need to test that the interaction between hipchat and alertmanager worked correctly with an integration test, but there is not really a way that I know to just spin up a fake hipchat server.

manager/notifier.go
a86c6f7964: I am thinking that the answer is no. It looks like golang does not provide a simple way to do timeouts using the http.Post api. Unless I am mistaken? Would it be good to make a ticket to address this in both this implementation and the pagerduty implementation?

brian-brazil: You can add a timeout on the client, see http://golang.org/pkg/net/http/#Client

As hipchat will generally get less-critical alerts than pagerduty I think it's best to ensure it doesn't gum things up given it's all serial at the moment.

a86c6f7964:
okay I think that I took most of your comments except for the unit tests into consideration. I feel like I could break out the message into a unit test if we really wanted to, but the only thing that I would know to test is the message. Which all of the golang maps should just default to "" anyhow so I am not sure if it would be worth it.

juliusv:
@a86c6f7964 I wouldn't demand it here, but the thing that'd make sense to test in this situation would be whether the correct JSON and headers are sent. That can be done by setting the hipchat URL to a custom one (which would point to an instance of http://golang.org/pkg/net/http/httptest/#Server). However, since we currently don't do that for the other mechanisms, I think I wouldn't require that for adding this for now.

config/config.go
```diff
@@ -64,6 +64,14 @@ func (c Config) Validate() error {
 				return fmt.Errorf("Missing user key in Pushover notification config: %s", proto.MarshalTextString(ec))
 			}
 		}
+		for _, hcc := range nc.HipchatConfig {
+			if hcc.AuthToken == nil {
+				return fmt.Errorf("Missing token in HipChat config: %s", proto.MarshalTextString(hcc))
+                        }
```
juliusv: Mix of tabs and spaces for indents here - please run `go fmt ./...`.

juliusv:
Could you add the same `README.md` addition as in https://github.com/prometheus/alertmanager/pull/40/files?

config/config.proto
```diff
@@ -34,6 +34,19 @@ message PushoverConfig {
   optional string user_key = 2;
 }
 
+message HipChatConfig {
```
juliusv: Please add a comment above this: `// Configuration for notification via HipChat.`, as with the other messages.

config/config.proto
```diff
@@ -34,6 +34,19 @@ message PushoverConfig {
...
+  // https://www.hipchat.com/docs/apiv2/method/send_room_notification
+
+  // Hipchat auth token, https://www.hipchat.com/docs/api/auth
+  optional string auth_token = 1;
+  // Hipchat room id, https://www.hipchat.com/rooms/ids
+  optional int32 room_id = 2;
+  // color of message
```
juliusv: Sentence-style doc-comments please (start capitalized, end with period).

config/config.proto
```diff
@@ -44,6 +57,8 @@ message NotificationConfig {
   repeated EmailConfig email_config = 3;
   // Zero or more pushover notification configurations.
   repeated PushoverConfig pushover_config = 4;
+  // Zero or more hipchat notification configuration.
```
juliusv: s/configuration/configurations/

juliusv:
:+1: otherwise, thanks!

a86c6f7964:
do you want me to squash any of this? or would you rather just leave all of the commits.

juliusv:
Yep, squash it please (after last comment) :) In general, commit messages should in the end be self-sufficient. For example, "adding 5 second timeout" is ambiguous without the context of a PR, so "Add 5 second timeout for Hipchat notifications" is better. So something like "Add simple support for HipChat notifications" would be great for the squashed overall commit.

a86c6f7964:
okay squashed

juliusv:
Thanks!

PR #47, opened 2015-04-14
Make notification flag names consistent with Prometheus style.

juliusv:


juliusv:
FYI @a86c6f7964 @discordianfish @brian-brazil

brian-brazil:
:+1:

discordianfish:
Meh, this might break lot of existing setups.. Agree that it should be consistent but I'm neutral on this I guess.

juliusv:
@discordianfish Well, Prometheus itserlf is `0.x` and alertmanager is clearly marked experimental and not production-ready. We'll still have minor breakages like this for a while, but at least they should be noticeable immediately upon startup.

PR #50, opened 2015-04-21
Visually distinguish alertname and summary

quinox:
Put the alert title in bold to make it visually more pleasing
Hipchat example


juliusv:
:+1: Thanks!

PR #51, opened 2015-04-24
Add support to notify for resolved alerts

discordianfish:
Instead of relying on prometheus to send notifications for resolved alerts, this change adds support to send resolve notification once an alert expired.

### DIFF
--- a/config/config.proto
+++ b/config/config.proto
@@ -24,14 +24,18 @@ message PagerDutyConfig {
 message EmailConfig {
   // Email address to notify.
   optional string email = 1;
+  // Notify when resolved.
+  optional bool send_resolved = 2 [default = false];
 }
 
 // Configuration for notification via pushover.net.
 message PushoverConfig {
-  // Pushover token
+  // Pushover token.
   optional string token = 1;
-  // Pushover user_key
+  // Pushover user_key.
   optional string user_key = 2;
+  // Notify when resolved.
+  optional bool send_resolved = 3 [default = false];
 }
 
 // Configuration for notification via HipChat.
@@ -42,10 +46,14 @@ message HipChatConfig {
   optional string auth_token = 1;
   // HipChat room id, (https://www.hipchat.com/rooms/ids).
   optional int32 room_id = 2;
-  // Color of message.
+  // Color of message when triggered.
   optional string color = 3 [default = "purple"];
+  // Color of message when resolved.
+  optional string color_resolved = 5 [default = "green"];
   // Should this message notify or not.
   optional bool notify = 4 [default = false];
+  // Notify when resolved.
+  optional bool send_resolved = 6 [default = false];
 }
 
 // Notification configuration definition.
--- a/manager/notifier.go
+++ b/manager/notifier.go
@@ -37,12 +37,17 @@ import (
 	pb "github.com/prometheus/alertmanager/config/generated"
 )
 
-const contentTypeJson = "application/json"
+const (
+	contentTypeJson = "application/json"
+
+	notificationOpTrigger notificationOp = iota
+	notificationOpResolve
+)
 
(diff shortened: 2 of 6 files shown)


manager/notifier.go
```diff
@@ -37,12 +37,17 @@ import (
 	pb "github.com/prometheus/alertmanager/config/generated"
 )
 
-const contentTypeJson = "application/json"
+const (
+	contentTypeJson = "application/json"
+
+	opTrigger = iota
```
juliusv: maybe explicitly type these as `Op`?

manager/notifier.go
```diff
@@ -62,11 +67,13 @@ var (
 	hipchatUrl             = flag.String("notification.hipchat.url", "https://api.hipchat.com/v2", "HipChat API V2 URL.")
 )
 
+type Op int
```
juliusv: `Op` seems a bit ambiguous in a package that does a lot of things. Maybe `notificationOp`?

manager/notifier.go
```diff
@@ -168,13 +186,22 @@ func (n *notifier) sendPagerDutyNotification(serviceKey string, a *Alert) error
...
 	// https://www.hipchat.com/docs/apiv2/method/send_room_notification
 	incidentKey := a.Fingerprint()
+	color := ""
+	status := ""
+	switch op {
+	case opTrigger:
+		color = config.GetColor()
+	case opResolve:
+		color = config.GetColorResolved()
+		status = " resolved"
```
juliusv: I'd move the space in this string to the format string where it's used below.

manager/notifier.go
```diff
@@ -168,13 +186,22 @@ func (n *notifier) sendPagerDutyNotification(serviceKey string, a *Alert) error
 	return nil
 }
 
-func (n *notifier) sendHipChatNotification(authToken string, roomId int32, color string, notify bool, a *Alert) error {
+func (n *notifier) sendHipChatNotification(op Op, config *pb.HipChatConfig, a *Alert) error {
 	// https://www.hipchat.com/docs/apiv2/method/send_room_notification
 	incidentKey := a.Fingerprint()
+	color := ""
+	status := ""
+	switch op {
```
juliusv: The PD notification implementation has a `default` error case. Should it be in the other ones too?

config/config.proto
```diff
@@ -24,6 +24,8 @@ message PagerDutyConfig {
 message EmailConfig {
   // Email address to notify.
   optional string email = 1;
+  // Notify when resolved
```
juliusv: add period

config/config.proto
```diff
@@ -32,6 +34,8 @@ message PushoverConfig {
   optional string token = 1;
   // Pushover user_key
   optional string user_key = 2;
+  // Notify when resolved
```
juliusv: period here and in the copy below

config/config.proto
```diff
@@ -24,6 +24,8 @@ message PagerDutyConfig {
 message EmailConfig {
   // Email address to notify.
   optional string email = 1;
+  // Notify when resolved
+  optional bool send_resolved = 2 [default = false];
```
juliusv: I'd vote for setting these to `true` by default :)

juliusv:
:+1: AFAICS besides comments - keep in mind I haven't dealt with this code base in quite a long time.

manager/notifier.go
discordianfish: Removed the default case from PD.

manager/notifier.go
```diff
@@ -168,13 +184,22 @@ func (n *notifier) sendPagerDutyNotification(serviceKey string, a *Alert) error
...
+	case notificationOpTrigger:
+		color = config.GetColor()
+	case notificationOpResolve:
+		color = config.GetColorResolved()
+		status = " resolved"
```
juliusv: "firing" / "resolved" perhaps?

PR #52, opened 2015-04-27
Path prefix to support reverse proxies

quinox:
Same as https://github.com/prometheus/prometheus/pull/614 and https://github.com/prometheus/promdash/pull/369

Using this code my AlertManager receives alerts from my Prometheus (after changing the API URL in Prometheus) and nothing in the interface is obviously broken.

One thing to note: there was a peculiar check inside `web/web.go`:

```python
		// The "/" pattern matches everything, so we need to check
		// that we're at the root here.
		if req.URL.Path != "/" {
			http.NotFound(rw, req)
			return
		}
```

Prometheus doesn't have something like this, and after moving the root handler to the bottom of that functions thing seem to work okay without it.

### DIFF
--- a/main.go
+++ b/main.go
@@ -16,6 +16,7 @@ package main
 import (
 	"flag"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/golang/glog"
@@ -30,10 +31,18 @@ var (
 	configFile       = flag.String("config.file", "alertmanager.conf", "Alert Manager configuration file name.")
 	silencesFile     = flag.String("silences.file", "silences.json", "Silence storage file name.")
 	minRefreshPeriod = flag.Duration("alerts.min-refresh-period", 5*time.Minute, "Minimum required alert refresh period before an alert is purged.")
+	pathPrefix       = flag.String("web.path-prefix", "/", "Prefix for all web paths.")
 )
 
 func main() {
 	flag.Parse()
+
+	if !strings.HasPrefix(*pathPrefix, "/") {
+		*pathPrefix = "/" + *pathPrefix
+	}
+	if !strings.HasSuffix(*pathPrefix, "/") {
+		*pathPrefix = *pathPrefix + "/"
+	}
 
 	versionInfoTmpl.Execute(os.Stdout, BuildInfo)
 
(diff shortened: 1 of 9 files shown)


juliusv:
Hey, awesome. The "peculiar" check is needed though. It fixes https://github.com/prometheus/prometheus/issues/607. Otherwise, e.g. http://localhost:9093/foo/bar does not return a `404`, but shows the alerts page instead.

quinox:
Thanks for the help. I've read up on https://github.com/prometheus/prometheus/issues/607 and [ServeMux](http://golang.org/pkg/net/http/#ServeMux) and changed the code accordingly: 
```bash
$ ./alertmanager --web.path-prefix=/promalert/ &

$ curl -I http://localhost:9093/                                                                                                                                                                           
HTTP/1.1 302 Found
Location: /promalert/
Date: Tue, 28 Apr 2015 19:13:33 GMT
Content-Type: text/plain; charset=utf-8

$ curl -I http://localhost:9093/promalert/
HTTP/1.1 200 OK
Date: Tue, 28 Apr 2015 19:13:44 GMT
Content-Type: text/html; charset=utf-8

$ curl -I http://localhost:9093/promalert/foo/bar
HTTP/1.1 404 Not Found
Content-Type: text/plain; charset=utf-8
Date: Tue, 28 Apr 2015 19:13:50 GMT
Content-Length: 19

$ curl -I http://localhost:9093/foo/bar          
HTTP/1.1 404 Not Found
Content-Type: text/plain; charset=utf-8
Date: Tue, 28 Apr 2015 19:13:54 GMT
Content-Length: 19
```

web/web.go
```diff
@@ -40,51 +40,58 @@ type WebService struct {
...
 
+
 	http.HandleFunc("/", prometheus.InstrumentHandlerFunc("index", func(rw http.ResponseWriter, req *http.Request) {
 		// The "/" pattern matches everything, so we need to check
 		// that we're at the root here.
-		if req.URL.Path != "/" {
+		if req.URL.Path == pathPrefix {
+			w.AlertsHandler.ServeHTTP(rw, req)
+		} else if req.URL.Path == "/" {
+			// We're running under a prefix but the user requested "/"
```
juliusv: nit: add period at end of comment :)

main.go
```diff
@@ -30,16 +31,24 @@ var (
...
+	if !strings.HasSuffix(*pathPrefix, "/") {
+		*pathPrefix = *pathPrefix + "/"
+	}
+
 	versionInfoTmpl.Execute(os.Stdout, BuildInfo)
 
 	conf := config.MustLoadFromFile(*configFile)
 
-	silencer := manager.NewSilencer()
+	silencer := manager.NewSilencer(*pathPrefix)
```
juliusv: I don't think the Silencer needs the path prefix at all, right? Only the web components need it...

juliusv: Confirmed that there is no user of this field at all, and AM still builds when simply removing it from this constructor and the struct type.

juliusv: Please apply: https://gist.github.com/juliusv/e4f5b3e640db3fb0758b

quinox:
Running `make test` succeeds, Travis fails on `package code.google.com/p/goprotobuf/proto: unable to detect version control system for code.google.com/ path`.

juliusv:
@quinox Ah, the protobuf package moved to GitHub. I updated the path and pushed to master. Could you rebase on master? Thanks!

quinox:
Rebased. 

As a side note, I tried to add a testcase for https://github.com/prometheus/prometheus/issues/607 but my knowledge of golang isn't sufficient to work around the `ListenAndServe` part

discordianfish:
LGTM

juliusv:
:+1: Thanks! Sorry for the late merge...

PR #53, opened 2015-04-27
we do not need the deps around after the executable is created

a86c6f7964:


juliusv:
:+1: 

/cc @discordianfish 

Actually, the only thing that's needed in the end is the static binary, but Docker seems to have this culture where all the cruft ends up in the container image... hope we can fix that for all our components in a general way soon. This is a good start.

PR #54, opened 2015-04-27
fix notifier, return means no other notifications will work. 

a86c6f7964:
instead continue in case other notifications should alert


juliusv:
@a86c6f7964 Oh my. Thanks for catching this! :+1: 

/cc @discordianfish

PR #55, opened 2015-05-05
Use Makefile.COMMON

discordianfish:
This now uses the shared Makefile.COMMON but since the alertmanager requires to run make -C web before building, it doesn't work. We'll have the same problem with prometheus itself.

So how should we solve that? We could make it possible to execute arbitrary build steps by changing the Makefile.COMMON, we could commit the files.go or we use go-generate.

Related: https://github.com/prometheus/prometheus/issues/548


discordianfish:
So for now I added the files.go, I guess we should discuss in https://github.com/prometheus/prometheus/issues/548 what to do on the long run.

web/Makefile
```diff
@@ -11,16 +11,14 @@
...
 
-include ../Makefile.INCLUDE
-
 all: blob/files.go
 
 SUFFIXES:
 
 blob/files.go: $(shell find templates/ static/ -type f)
-	./blob/embed-static.sh static templates | $(GOFMT) > $@
+	./blob/embed-static.sh static templates | gofmt > $@
```
juliusv: If we kept the `make web` step in the parent Makefile (to be executed manually), we could leave this as is, no?

Makefile
```diff
@@ -11,47 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-include Makefile.INCLUDE
+VERSION  := 0.1.0
```
juliusv: With alertmanager as an exception, I would go for 0.0.1 to point out to people just how experimental it is...

brian-brazil:
:+1:

juliusv:
I'd include the Makefile.COMMON at the top, but otherwise :+1:

PR #57, opened 2015-05-10
Add simple support for Slack notifications

sdurrheimer:
Like the HipChat one, here a simple implementation for Slack notifications.

alertmanager_slack_notifcations

@juliusv

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -72,6 +72,11 @@ func (c Config) Validate() error {
 				return fmt.Errorf("Missing room in HipChat config: %s", proto.MarshalTextString(hcc))
 			}
 		}
+		for _, sc := range nc.SlackConfig {
+			if sc.WebhookUrl == nil {
+				return fmt.Errorf("Missing webhook URL in Slack config: %s", proto.MarshalTextString(sc))
+			}
+		}
 
 		if _, ok := ncNames[nc.GetName()]; ok {
 			return fmt.Errorf("Notification config name not unique: %s", nc.GetName())
--- a/manager/notifier.go
+++ b/manager/notifier.go
@@ -38,7 +38,7 @@ import (
 )
 
 const (
-	contentTypeJson = "application/json"
+	contentTypeJSON = "application/json"
 
 	notificationOpTrigger notificationOp = iota
 	notificationOpResolve
@@ -61,10 +61,10 @@ Payload labels:
 
 var (
 	notificationBufferSize = flag.Int("notification.buffer-size", 1000, "Size of buffer for pending notifications.")
-	pagerdutyApiUrl        = flag.String("notification.pagerduty.url", "https://events.pagerduty.com/generic/2010-04-15/create_event.json", "PagerDuty API URL.")
+	pagerdutyAPIURL        = flag.String("notification.pagerduty.url", "https://events.pagerduty.com/generic/2010-04-15/create_event.json", "PagerDuty API URL.")
 	smtpSmartHost          = flag.String("notification.smtp.smarthost", "", "Address of the smarthost to send all email notifications to.")
 	smtpSender             = flag.String("notification.smtp.sender", "alertmanager@example.org", "Sender email address to use in email notifications.")
-	hipchatUrl             = flag.String("notification.hipchat.url", "https://api.hipchat.com/v2", "HipChat API V2 URL.")
+	hipchatURL             = flag.String("notification.hipchat.url", "https://api.hipchat.com/v2", "HipChat API V2 URL.")
 )
 
 type notificationOp int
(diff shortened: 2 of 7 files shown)


manager/notifier.go
```diff
@@ -231,6 +231,70 @@ func (n *notifier) sendHipChatNotification(op notificationOp, config *pb.HipChat
...
+		status = "firing"
+	case notificationOpResolve:
+		color = config.GetColorResolved()
+		status = "resolved"
+	}
+
+	buf, err := json.Marshal(map[string]interface{}{
+		"channel": config.GetChannel(),
+		"attachments": []interface{}{
+			map[string]interface{}{
```
fabxc: How about a properly typed struct here? Even though more verbose, I always prefer those over maps and empty interfaces. Especially in bad cases of nesting like this one.

I know the other implementations are not leading by best example - but we can start improving here.

manager/notifier.go
```diff
@@ -392,6 +456,14 @@ func (n *notifier) handleNotification(a *Alert, op notificationOp, config *pb.No
 			glog.Error("Error sending HipChat notification: ", err)
 		}
 	}
+	for _, scConfig := range config.SlackConfig {
+		if op == notificationOpResolve && !scConfig.GetSendResolved() {
+			continue
+		}
+		if err := n.sendSlackNotification(op, scConfig, a); err != nil {
+			glog.Error("Error sending Slack notification: ", err)
```
fabxc: `Errorln` would be more idiomatic here.
Maybe you could change that in the calls above as well.

manager/notifier.go
```diff
@@ -231,6 +231,70 @@ func (n *notifier) sendHipChatNotification(op notificationOp, config *pb.HipChat
...
+	if err != nil {
+		return err
+	}
+
+	timeout := time.Duration(5 * time.Second)
+	client := http.Client{
+		Timeout: timeout,
+	}
+	resp, err := client.Post(
+		fmt.Sprintf("%s", config.GetWebhookUrl()),
```
fabxc: No need for the call to `Sprintf` here.

fabxc:
Thanks! Looks good otherwise.

config/config.go
```diff
@@ -72,6 +72,11 @@ func (c Config) Validate() error {
 				return fmt.Errorf("Missing room in HipChat config: %s", proto.MarshalTextString(hcc))
 			}
 		}
+		for _, sc := range nc.SlackConfig {
+			if sc.WebhookUrl == nil {
+				return fmt.Errorf("Missing webhook url in Slack config: %s", proto.MarshalTextString(sc))
```
juliusv: nit: s/url/URL/

manager/notifier.go
```diff
@@ -231,6 +231,100 @@ func (n *notifier) sendHipChatNotification(op notificationOp, config *pb.HipChat
...
+	Pretext   string                 `json:"pretext,omitempty"`
+	Title     string                 `json:"title,omitempty"`
+	TitleLink string                 `json:"title_link,omitempty"`
+	Text      string                 `json:"text"`
+	Color     string                 `json:"color,omitempty"`
+	MrkdwnIn  []string               `json:"mrkdwn_in,omitempty"`
+	Fields    []slackAttachmentField `json:"fields,omitempty"`
+}
+
+// Slack fields to display in a table inside the message attachment
```
fabxc: Strictly speaking, godoc style would be `slackAttachment is ...`. As those are not exported I don't feel strong about this.
They should end with a `.`, though.

fabxc:
Looks good for merge in general :) — after nits and squashing :+1:

config/config.go
sdurrheimer: In the same style, should I rename `contentTypeJson`, `pagerdutyApiUrl` and `hipchatUrl` ?

juliusv: Oh yes, I hadn't even looked that far yet. While this was about the word "url" in the error message (also error-messages should all start out lower-cased, but the others here are all wrong too, so maybe keep yours starting upper-cased for now), the general rule is also to avoid "initialisms" in variable names: https://github.com/golang/go/wiki/CodeReviewComments#initialisms

So `contentTypeJson` -> `contentTypeJSON`.

The exception are protobuf-generated names like the `WebhookUrl`. You can't use `WebhookURL` there because the Go name is autogenerated from the protocol buffer schema definition.

manager/notifier.go
```diff
@@ -231,6 +231,100 @@ func (n *notifier) sendHipChatNotification(op notificationOp, config *pb.HipChat
 	return nil
 }
 
+// Slack body request
```
juliusv: Please always use sentence-style doc comments (end with a period): https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

juliusv:
:+1: Looking forward to using this at SoundCloud at some point :)

sdurrheimer:
Waiting for the final go-ahead to squash in one commit.

fabxc:
Squash it! (Plus the s/url/URL/ nit.)
Thanks for fixing the other parts as well.

fabxc:
Great!

PR #60, opened 2015-05-18
Flowdock notifications

t0mk:
Hi, I wrote a Flowdock notifier.

I did it a bit different way than the HipChat and Slack notifiers, I didn't want to duplicate code in e.g. serializing to JSON, and posting to HTTP URL.

There is quite a lot of duplicated code in manager/notifier.go. I was thinking I could rewite it in similar manner in another pull-request. Would you be interested in that?

I didn't write any tests, it wasn't straightforward to mock the config and the alert. Maybe it would be easier after the refactoring I suggest.

### DIFF
--- a/config/config.proto
+++ b/config/config.proto
@@ -71,11 +71,11 @@ message SlackConfig {
 }
 
 message FlowdockConfig {
-  // Flowdock flow API token
+  // Flowdock flow API token.
   optional string api_token = 1;
-  // Flowdock from_address
+  // Flowdock from_address.
   optional string from_address = 2;
-  // Flowdock flow tags
+  // Flowdock flow tags.
   repeated string tag = 3;
   // Notify when resolved.
   optional bool send_resolved = 4 [default = false];
--- a/config/config.proto
+++ b/config/config.proto
@@ -70,6 +70,17 @@ message SlackConfig {
   optional bool send_resolved = 5 [default = false];
 }
 
+message FlowdockConfig {
+  // Flowdock flow API token
+  optional string api_token = 2;
+  // Flowdock from_address
+  optional string from_address = 3;
+  // Flowdock flow tags
+  repeated string tags = 4;
+  // Color of message when triggered.
+  optional bool send_resolved = 7 [default = false];
+}
+
 // Notification configuration definition.
 message NotificationConfig {
   // Name of this NotificationConfig. Referenced from AggregationRule.
@@ -84,6 +95,8 @@ message NotificationConfig {
   repeated HipChatConfig hipchat_config = 5;
   // Zero or more slack notification configurations.
   repeated SlackConfig slack_config = 6;
+  // Zero or more Flowdock notification configurations.
+  repeated FlowdockConfig flowdock_config = 7;
 }
 
 // A regex-based label filter used in aggregations.
--- a/config/config.proto
+++ b/config/config.proto
@@ -72,13 +72,13 @@ message SlackConfig {
 
 message FlowdockConfig {
   // Flowdock flow API token
-  optional string api_token = 2;
+  optional string api_token = 1;
   // Flowdock from_address
-  optional string from_address = 3;
+  optional string from_address = 2;
   // Flowdock flow tags
-  repeated string tags = 4;
-  // Color of message when triggered.
-  optional bool send_resolved = 7 [default = false];
+  repeated string tag = 3;
+  // Notify when resolved.
+  optional bool send_resolved = 4 [default = false];
 }
 
 // Notification configuration definition.
--- a/manager/notifier.go
+++ b/manager/notifier.go
@@ -326,10 +326,6 @@ func (n *notifier) sendSlackNotification(op notificationOp, config *pb.SlackConf
 	return nil
 }
 
-// https://www.flowdock.com/api/team-inbox
-// compulsory fields in Flowdock JSON: source, from_address, subject, content
-// Content-type: application/json
-// POST to https://api.flowdock.com/v1/messages/team_inbox/:flow_api_token
 type flowdockMessage struct {
 	Source      string   `json:"source"`
 	FromAddress string   `json:"from_address"`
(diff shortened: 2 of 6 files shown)


config/config.proto
```diff
@@ -70,6 +70,17 @@ message SlackConfig {
   optional bool send_resolved = 5 [default = false];
 }
 
+message FlowdockConfig {
+  // Flowdock flow API token
+  optional string api_token = 2;
```
juliusv: Heh. Please start numbers at 1 and then increment by 1 for each field :)

config/config.proto
```diff
@@ -70,6 +70,17 @@ message SlackConfig {
...
+  // Flowdock from_address
+  optional string from_address = 3;
+  // Flowdock flow tags
+  repeated string tags = 4;
```
juliusv: You'll probably want to call this just `tag`, because when using a repeated field, you just list it multiple times with as the same field name:

```
tag: "foo"
tag: "bar"
...
```

config/config.proto
```diff
@@ -70,6 +70,17 @@ message SlackConfig {
...
+  // Color of message when triggered.
```
juliusv: The comment doesn't match the field below here...

manager/notifier.go
```diff
@@ -325,6 +326,59 @@ func (n *notifier) sendSlackNotification(op notificationOp, config *pb.SlackConf
 	return nil
 }
 
+// https://www.flowdock.com/api/team-inbox
+// compulsory fields in Flowdock JSON: source, from_address, subject, content
+// Content-type: application/json
+// POST to https://api.flowdock.com/v1/messages/team_inbox/:flow_api_token
+type FlowdockMessage struct {
```
juliusv: AFAICS this type doesn't need to be exported / capitalized (but the fields need to be, for JSON unmarshaling).

manager/notifier.go
```diff
@@ -325,6 +326,59 @@ func (n *notifier) sendSlackNotification(op notificationOp, config *pb.SlackConf
...
+		Subject:     html.EscapeString(a.Summary),
+		Format:      "html",
+		Content:     fmt.Sprintf("*%s %s*: %s (<%s|view>)", html.EscapeString(a.Labels["alertname"]), status, html.EscapeString(a.Summary), a.Payload["GeneratorURL"]),
+		Link:        a.Payload["GeneratorURL"],
+		Tags:        config.GetTags(),
+	}
+	return msg
+}
+
+func postJSONtoURL(jsonMessage []byte, url string) *http.Response {
```
juliusv: I'd just call it `postJSON` (avoids the lower-case `to` awkardness etc.)

manager/notifier.go
```diff
@@ -325,6 +326,59 @@ func (n *notifier) sendSlackNotification(op notificationOp, config *pb.SlackConf
...
+}
+
+func jsonize(msg interface{}) []byte {
+	buf, err := json.Marshal(msg)
+	if err != nil {
+		glog.Errorln("Error compiling JSON:", err)
+	}
+	return buf
+}
+func getFlowdockNotificationMessage(op notificationOp, config *pb.FlowdockConfig, a *Alert) *FlowdockMessage {
```
juliusv: `get...` -> `new...` seems to be the naming convention in Go for functions which just build objects for you.

manager/notifier.go
```diff
@@ -325,6 +326,59 @@ func (n *notifier) sendSlackNotification(op notificationOp, config *pb.SlackConf
...
+	Source      string   `json:"source"`
+	FromAddress string   `json:"from_address"`
+	Subject     string   `json:"subject"`
+	Content     string   `json:"content"`
+	Format      string   `json:"format"`
+	Link        string   `json:"link"`
+	Tags        []string `json:"tags,omitempty"`
+}
+
+func jsonize(msg interface{}) []byte {
```
juliusv: `jsonize` sounds a bit... colloquial. How about `marshalJSON`?

manager/notifier.go
```diff
@@ -325,6 +326,59 @@ func (n *notifier) sendSlackNotification(op notificationOp, config *pb.SlackConf
...
+	buf, err := json.Marshal(msg)
+	if err != nil {
+		glog.Errorln("Error compiling JSON:", err)
```
juliusv: compiling -> marshalling or serializing would be the best term :)

juliusv:
Looks pretty cool otherwise.

It would be great indeed to refactor things more in the future and make tests easier.

t0mk:
I fixed issues from your comments. Plus,
- I added status (firing|resolved} as a tag for a Flowdock message.
- I mentioned Flowdock in readme
- I improved HTTP response logging in notifier.go

Once you merge this, will you rebuild the prom/alertmanager Docker image?

manager/notifier.go
```diff
@@ -494,6 +563,15 @@ func (n *notifier) handleNotification(a *Alert, op notificationOp, config *pb.No
 			glog.Errorln("Error sending Slack notification:", err)
 		}
 	}
+	for _, fdConfig := range config.FlowdockConfig {
+		if op == notificationOpResolve && !fdConfig.GetSendResolved() {
+			continue
+		}
+		flowdockMessage := newFlowdockNotificationMessage(op, fdConfig, a)
+		url := *flowdockURL + "/" + fdConfig.GetApiToken()
+		httpResponse := postJSON(marshallJSON(flowdockMessage), url)
```
fabxc: Why would we want to make the request with invalid data?
I don't think the the `marshallJSON` helper is beneficial in general. It just hides the error from the control flow.

juliusv: Oh yeah, I didn't even notice that. I agree... also, the name would be `marshalJSON` (with one `l`) - see also `MarshalJSON` in the stdlib.

config/config.proto
```diff
@@ -70,6 +70,17 @@ message SlackConfig {
   optional bool send_resolved = 5 [default = false];
 }
 
+message FlowdockConfig {
+  // Flowdock flow API token
```
juliusv: Sorry if I didn't notice this before, but style nit: period after each doc comment please.

t0mk:
I fixed issues from previous comments. Also, I made the flowdock code look a bit more like the existing. I tried to put the code which I think can be reused in separate functions.

juliusv:
@t0mk Nice, thanks!

PR #62, opened 2015-05-22
Added support for @mentions in HipChat notifications

SQiShER:
Hi! So I was planning to pester my fellow co-workers with HipChat notifications, whenever Prometheus fires alerts. In order to do so, I needed a way to add `@all` or `@somename` to the message body. (I know there is a notify flag, but that didn't seem to work as well and doesn't allow to target specific people or groups like `all` or `here`.) 

The alertmanager sends HipChat notifications in HTML format. This opens the door for nicely formatted messages, but also has the consequence, that @mentions don't work. (For some strange reason, that only Atlassian knows...)

Here is an excerpt from the `message_format` section of the HipChat documentation:

> * html - Message is rendered as HTML and receives no special treatment. Must be valid HTML and entities must be escaped (e.g.: '&amp;' instead of '&'). May contain basic tags: a, b, i, strong, em, br, img, pre, code, lists, tables.
> * text - Message is treated just like a message sent by a user. Can include @mentions, emoticons, pastes, and auto-detected URLs (Twitter, YouTube, images, etc).

In order to make it possible to use mentions, I added two new properties to the configuration: `message_format` (which can be `text` or `html`) and `prefix`, which is a free-form text that's simply prepended to the message.

In order to keep it backward-compatible `html` remains the default and both new properties are optional. However, if present, the `prefix` will also be applied to the HTML message.

I'd love to see this change make its way into master. If you have any concerns or know a better way to achieve the same result, please let me know! Also, since I'm new to Go and Protobuf please let me know, if there is anything I should change about the code.

Best regards,
Daniel

### DIFF
--- a/config/config.proto
+++ b/config/config.proto
@@ -54,6 +54,14 @@ message HipChatConfig {
   optional bool notify = 4 [default = false];
   // Notify when resolved.
   optional bool send_resolved = 6 [default = false];
+  // Prefix to be put in front of the message (useful for @mentions, etc.).
+  optional string prefix = 7 [default = ""];
+  // Format the message as "html" or "text".
+  enum MessageFormat {
+    HTML = 0;
+    TEXT = 1;
+  }
+  optional MessageFormat message_format = 8 [default = HTML];
 }
 
 // Configuration for notification via Slack.
--- a/manager/notifier.go
+++ b/manager/notifier.go
@@ -190,6 +190,8 @@ func (n *notifier) sendHipChatNotification(op notificationOp, config *pb.HipChat
 	incidentKey := a.Fingerprint()
 	color := ""
 	status := ""
+	message := ""
+	messageFormat := ""
 	switch op {
 	case notificationOpTrigger:
 		color = config.GetColor()
@@ -198,11 +200,18 @@ func (n *notifier) sendHipChatNotification(op notificationOp, config *pb.HipChat
 		color = config.GetColorResolved()
 		status = "resolved"
 	}
+	if config.GetMessageFormat() == pb.HipChatConfig_TEXT {
+		message = fmt.Sprintf("%s%s %s: %s", config.GetPrefix(), a.Labels["alertname"], status, a.Summary)
+		messageFormat = "text"
+	} else {
+		message = fmt.Sprintf("%s<b>%s %s</b>: %s (<a href='%s'>view</a>)", config.GetPrefix(), html.EscapeString(a.Labels["alertname"]), status, html.EscapeString(a.Summary), a.Payload["GeneratorURL"])
+		messageFormat = "html"
+	}
 	buf, err := json.Marshal(map[string]interface{}{
 		"color":          color,
-		"message":        fmt.Sprintf("<b>%s %s</b>: %s (<a href='%s'>view</a>)", html.EscapeString(a.Labels["alertname"]), status, html.EscapeString(a.Summary), a.Payload["GeneratorURL"]),
+		"message":        message,
 		"notify":         config.GetNotify(),
-		"message_format": "html",
+		"message_format": messageFormat,
 	})
 	if err != nil {
 		return err
--- a/config/generated/config.pb.go
+++ b/config/generated/config.pb.go
@@ -30,6 +30,40 @@ import math "math"
 var _ = proto.Marshal
 var _ = math.Inf
 
+// Format the message as "html" or "text".
+type HipChatConfig_MessageFormat int32
+
+const (
+	HipChatConfig_HTML HipChatConfig_MessageFormat = 0
+	HipChatConfig_TEXT HipChatConfig_MessageFormat = 1
+)
+
+var HipChatConfig_MessageFormat_name = map[int32]string{
+	0: "HTML",
+	1: "TEXT",
+}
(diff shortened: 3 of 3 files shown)


config/config.proto
```diff
@@ -54,6 +54,10 @@ message HipChatConfig {
   optional bool notify = 4 [default = false];
   // Notify when resolved.
   optional bool send_resolved = 6 [default = false];
+  // Prefix to be put in front of the message (useful for @mentions, etc.)
```
juliusv: style nit: please end doc comments with a period.

config/config.proto
```diff
@@ -54,6 +54,10 @@ message HipChatConfig {
...
+  optional string prefix = 7 [default = ""];
+  // Format the message as "html" or "text"
+  optional string message_format = 8 [default = "html"];
```
juliusv: This would need a check in `Config.Validate()` (https://github.com/prometheus/alertmanager/blob/master/config/config.go#L43) to make sure it's set to a valid value.

But I think it may be better to use an enum (and then AFAIK it wouldn't need extra validation either):

```
enum MessageFormat {
  HTML = 0;
  TEXT = 1;
}
optional MessageFormat message_format = 8 [default = HTML];
```

manager/notifier.go
```diff
@@ -190,6 +190,8 @@ func (n *notifier) sendHipChatNotification(op notificationOp, config *pb.HipChat
 	incidentKey := a.Fingerprint()
 	color := ""
 	status := ""
+	message := ""
+	message_format := ""
```
juliusv: Please stick to `camelCase` for Go identifiers.

juliusv:
:+1: Otherwise. Thanks for the great commit/PR description!

fabxc:
Great PR, thanks! :+1: besides what @juliusv pointed out.

brian-brazil:
I think what we should be looking at is general templating for all of the notification methods, so that you can configure what the various outputs look like to your heart's content.

This PR solves only one use case in a large sea of use cases, and adding special cases for each individual deployment's use cases isn't something that scales.

juliusv:
@brian-brazil :+1: in general for templating of message texts, but for the particular feature in this PR to work, the actual JSON of the HipChat API request would have to be generated by a template - that seems a bit intense.

SQiShER:
Thank you all for your feedback!

@juliusv I made the proposed changes and squashed the commits. I followed your suggestion and used an enum. I like that, because this way the `config.proto` file serves as good documentation for the allowed options without any hidden caveats. The only thing I find confusing about it is that the configuration now looks like this:

```yaml
hipchat_config {
    auth_token: "authtoken"
    room_id: 4815162342
    prefix: "@all "
    message_format: TEXT
}
```

It took me a while to figure out, that the value of `message_format` must not be quoted. Although it looks kinda cool and one could simply blame it on my inexperience with Protobuf, I can imagine I won't be the last one who's confused by that. Since there is no explicit validation, any mistake will simply cause the notifier to fallback to HTML.

What do you think? Should we rather go the other route with custom validation and string values?

fabxc:
@SQiShER, we switched Prometheus itself to YAML with the new version. We probably want to go the same route with the alertmanager to be consistent.

While the self-documenting feature of the `config.proto` file is nice for devs, so are Go structs we parse YAML into . It has limited benefit for the end-user – also for the general unfamiliarity with ASCII protobufs.

Regardless of future changes, In this case enumerations are the right approach.

brian-brazil:
> the actual JSON of the HipChat API request would have to be generated by a template 

You'd just template the message. The other bits of the json could be static/configured or templated themselves.

SQiShER:
Nevermind that part of my last comment:

> Since there is no explicit validation, any mistake will simply cause the notifier to fallback to HTML.

I tried again and validation works fine and as expected: only `TEXT` and `HTML` will be accepted. Any other value causes an error. I think I got confused there for a moment.

@brian-brazil Do you have something like this in mind?

```yaml
hipchat_config {
    message_template: "<b>{{alert.alertname}} {{status}}</b>: {{alert.summary}} (<a href='{{alert.payload.GeneratorURL}}'>view</a>)"
    message_format: HTML
}
```

Or for plain text:

```yaml
hipchat_config {
    message_template: "@all {{alert.alertname}} {{status}}: {{alert.summary}}"
    message_format: TEXT
}
```

I like that and it would definitely solve my problem in a more flexible way. It would even allow to use the alert description instead of the summary; or to include certain labels. One could even go further and allow different templates for `firing` and `resolved` status. On the other hand the whole templating mechanism probably adds quite some complexity to the notifier implementations. But it could be worth it.

SQiShER:
@fabxc Alright, that's good to know!

brian-brazil:
@SQiShER Something like that, there's a few things to be thought of beyond that such as dealing with multiple alerts in one template and if we need an abstraction to let templates be shared across notification sections.

SQiShER:
Should I change this PR to the template approach or is that something you'd see on the roadmap? /cc @brian-brazil @juliusv

juliusv:
IMO the current approach is fine for now and the templating is a long-term thing. We'll probably change the config to YAML soon, so there's going to be a breaking change anyways.

juliusv:
:+1: Merging for now.

PR #64, opened 2015-05-22
add status to pushover title

a86c6f7964:
assume that the summary is the message, and that the alertname is a very short concise name for the alert


juliusv:
Thanks!

:+1: from my side - just /cc-ing @antonlindstrom, who initially implemented Pushover support to see if this works for him too.

antonlindstrom:
Works for me too! :+1: Thanks @a86c6f7964!

fabxc:
Thanks to both of you! :+1:

PR #68, opened 2015-05-23
Migrate logging to use `prometheus/log`.

juliusv:


juliusv:
@fabxc

fabxc:
:+1:

juliusv:
Also removed another reference in the README.md. Dockerfile is fine already.

stapelberg:
This pull request broke my Dockerfile because -logtostderr is no longer defined.

At which point will alertmanager and prometheus itself be stable enough to no longer break the UI?

beorn7:
No real guarantee before v1.0.0. But even now we try to not have breaking changes for all releases 0.x.y where x doesn't change. If you build from head, you will always run into breaking changes.
Alertmanager is still so raw that releases were only introduced recently. Still, if you pin to 0.0.1, you should not run into surprises, and breaking changes between releases are always documented.

juliusv:
See also [this FAQ](http://prometheus.io/docs/introduction/faq/#how-stable-are-prometheus-features,-storage-formats,-and-apis?) about UI/API stability.

The alertmanager is arguably the most experimental component, as pointed out in the README.md. It'll likely get a complete overhaul at some point, given that its current architecture is more of a proof-of-concept than a production-ready service.

PR #69, opened 2015-05-23
Update fsnotify to its new source location

benley:
Probably ought to update to v1.x of fsnotify at some point as well.

[Review on Reviewable](https://reviewable.io/reviews/prometheus/alertmanager/69)

### DIFF
--- a/config/watcher.go
+++ b/config/watcher.go
@@ -14,8 +14,8 @@
 package config
 
 import (
-	"github.com/howeyc/fsnotify"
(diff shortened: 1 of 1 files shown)


fabxc:
:+1: 

Not PR-related:
Eventually we should switch the alertmanager to YAML as well for consistency. For Prometheus itself we decided for `SIGHUP` only for config reloading - that's also something we'll want to be consistent with.

PR #70, opened 2015-05-26
Add a generic webhook notifier.

brian-brazil:
This allows for use cases such as kicking off shell scripts,
logging notifications, or anything else the alertmanager doesn't
directly support.

@juliusv 

Is there a better way to test this? This seems kinda clunky.

### DIFF
--- a/config/config.proto
+++ b/config/config.proto
@@ -58,7 +58,7 @@ message HipChatConfig {
 
 // Configuration for notification via Slack.
 message SlackConfig {
-  // Slack webhook url, (https://api.slack.com/incoming-webhooks).
+  // Slack webhook URL, (https://api.slack.com/incoming-webhooks).
   optional string webhook_url = 1;
   // Slack channel override, (like #other-channel or @username).
   optional string channel = 2;
@@ -70,6 +70,7 @@ message SlackConfig {
   optional bool send_resolved = 5 [default = false];
 }
 
+// Configuration for notification via Flowdock.
 message FlowdockConfig {
   // Flowdock flow API token.
   optional string api_token = 1;
@@ -79,6 +80,14 @@ message FlowdockConfig {
   repeated string tag = 3;
   // Notify when resolved.
   optional bool send_resolved = 4 [default = false];
+}
+
+// Configuration for notification via generic webhook.
+message WebhookConfig {
+  // URL to send POST request to.
+  optional string url = 1;
+  // Notify when resolved.
+  optional bool send_resolved = 2 [default = false];
 }
 
 // Notification configuration definition.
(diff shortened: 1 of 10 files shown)


config/config.proto
```diff
@@ -81,6 +81,13 @@ message FlowdockConfig {
   optional bool send_resolved = 4 [default = false];
 }
 
+message WebhookConfig {
+  // Url to send POST request to.
```
juliusv: Url -> URL in the comment

config/config.proto
```diff
@@ -81,6 +81,13 @@ message FlowdockConfig {
   optional bool send_resolved = 4 [default = false];
 }
 
+message WebhookConfig {
```
juliusv: Docstring above this (and the FlowdockConfig above, seems it was forgotten there too).

juliusv:
I'd generally like to keep too many unrelated changes out of the same PR or at least commit. Especially the big files.go update is not related to this, right?

:+1: otherwise.

As for testing, the current way is ok, but a more generic way would definitely be nicer in the future. Some way of just defining the test alerts only once, and then reusing them for all the different notification methods (using table-based tests referencing notification methods and fixture files containing the resulting JSON or other expected outputs). Something like that...

fabxc:
What `protoc` version/flags were you using? Somehow all the comments were stripped from the generated code.

Also the embedding of static files in `blob/files.go` changed but none the files themselves. Why is it not gitingored in the alertmanager anyway?

juliusv:
Ah, good point @fabxc. @brian-brazil please make sure you're using the latest protoc version and that comments are rendered out.

config/config.proto
brian-brazil: Done, and above.

config/config.proto
brian-brazil: Done.

brian-brazil:
I'm using the latest protoc that comes with Ubuntu 12.04. If we need a specific version, then the build should be hermetic and pull it in itself.

juliusv:
Ubuntu 12.04 is ancient :) I think I used https://github.com/google/protobuf/releases/tag/v2.6.1.

Pulling and installing the right version automatically might be nice, but from a recent discussion it felt to me like multiple people wanted to get away from trying to build the perfect auto-installer for all build dependencies. Further, we will likely switch away from protobufs for the AM too at some point too, so time investments in this area might not really be worth so much right now.

fabxc:
A consistent config format across our components is essential. Switching to YAML will be one of the first things to do once we shift our focus towards the Alertmanager.

I don't even mind whether the generated code has doc strings, tbh.

brian-brazil:
I've managed to get a recent version by compiling from source. 

I don't think it's reasonable for us to require that level of effort of contributors using a modern, supported, and widely used OS - we can't assume that everyone is on the bleeding edge or has the technical skills and time to deal with that sort of issue.

fabxc:
I generally agree with that.

Unfortunately, the PR can no longer be merged.

fabxc:
For the future, I hope that we can limit our additional tools required for contributing to go-generate for embedding static assets and a build tool (I have high hopes for [gb](http://getgb.io)).

brian-brazil:
That should be good now.

I reset the files.go, but I think it needs to be regenerated.

fabxc:
:+1:

PR #71, opened 2015-05-30
Change case of alert json to initial lower letter.

brian-brazil:
Matching change to what's on the prometheus server side.

See https://github.com/prometheus/prometheus/pull/751 and #70


discordianfish:
LGTM

juliusv:
:+1:

PR #72, opened 2015-06-03
Fix silences.json permission error with docker image

sdurrheimer:
This fix the `Error saving silences to file: open silences.json: permission denied` error when using `prom/alertmanager` docker image.

@juliusv @discordianfish


discordianfish:
LGTM

PR #75, opened 2015-06-10
Fix email template body writer being called with parameters in wrong order

wrouesnel:
The email body writer function had it's parameters in the wrong order, leading to the insertion of the email address of the sender rather then the actual status field.

I can confirm from testing that this pull request fixes the problem.


brian-brazil:
:+1: Thanks

PR #78, opened 2015-06-11
Add a GET api to /api/alerts which pulls JSON formatted AlertAggregates.

wrouesnel:
This is a tiny change which extends the /api/alerts function to write JSON formatted AlertAggregates when polled with a GET request.

This allows integrating the alertmanager with existing dashboard/notification walls which either want to poll the alerts output or get an initial scan of the current alert state.

### DIFF
--- a/web/api/alert.go
+++ b/web/api/alert.go
@@ -22,6 +22,12 @@ import (
 	"github.com/prometheus/alertmanager/manager"
 )
 
+// Return all currently firing alerts as JSON.
+func (s AlertManagerService) getAlerts(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+	aggs := s.Manager.GetAll(nil)
+	respondJSON(w, aggs)
+}
+
 func (s AlertManagerService) addAlerts(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
 	alerts := manager.Alerts{}
 	if err := parseJSON(w, r, &alerts); err != nil {
--- a/web/api/api.go
+++ b/web/api/api.go
@@ -32,6 +32,7 @@ type AlertManagerService struct {
 
 func (s AlertManagerService) Handler() http.Handler {
 	r := httprouter.New()
+	r.GET(s.PathPrefix+"api/alerts", s.getAlerts)
 	r.POST(s.PathPrefix+"api/alerts", s.addAlerts)
 	r.GET(s.PathPrefix+"api/silences", s.silenceSummary)
 	r.POST(s.PathPrefix+"api/silences", s.addSilence)


juliusv:
This should probably go in the new v1 API endpoint at `/api/v1/...` (https://github.com/prometheus/prometheus/tree/master/web/api/v1) if we decide to support this.

Just checking, are you aware that you can already get information about pending and firing alerts as time series (`ALERTS` metric name)? That won't show you completely inactive alert definitions though.

wrouesnel:
Yes, but since you might have several prometheii feeding the one alert manager if makes more sense to get the data from there. In a practical sense this is just the machine-parseable version of the alerts page.

juliusv:
Ah sorry - I totally did not see that this PR was for alertmanager and not for Prometheus (since that also has an alerts page).

web/api/alert.go
```diff
@@ -39,4 +46,4 @@ func (s AlertManagerService) addAlerts(w http.ResponseWriter, r *http.Request, p
 	}
 
 	s.Manager.Receive(alerts)
-}
+}
```
juliusv: Please configure your editor to end files with newlines (http://stackoverflow.com/questions/729692/why-should-files-end-with-a-newline).

web/api/alert.go
```diff
@@ -16,12 +16,19 @@ package api
...
 	"fmt"
 	"net/http"
+	"encoding/json"
 
 	"github.com/julienschmidt/httprouter"
 
 	"github.com/prometheus/alertmanager/manager"
 )
 
+// Return all currently firing alerts as JSON
```
juliusv: Style nit: period after comment.

web/api/alert.go
```diff
@@ -16,12 +16,19 @@ package api
...
+func (s AlertManagerService) getAlerts(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+	aggs := s.Manager.GetAll(nil)
+	json.NewEncoder(w).Encode(aggs)
```
juliusv: Could you do:

```
respondJSON(w, aggs)
```

This handles error cases (failure to marshal JSON) correctly and is consistent with the API methods in `silence.go`.

juliusv:
:+1: in general, but under the caveat that Alertmanager's API can still change at any time for the moment, so this might be broken in the future.

wrouesnel:
That commit should take care of the issues. Agree with the caveat - but the basic method will be needed however the api changes.

juliusv:
@wrouesnel Thanks! Could you squash, and I'll merge.

wrouesnel:
Squashed and ready.

brian-brazil:
If we're now exposing AlertAggregate, we should be ensuring it's fields are camelCase rather than CamelCase to be consistent with everything else we expose JSON wise.

web/api/alert.go
```diff
@@ -40,3 +46,4 @@ func (s AlertManagerService) addAlerts(w http.ResponseWriter, r *http.Request, p
 
 	s.Manager.Receive(alerts)
 }
+
```
juliusv: My earlier newline wording might have been confusing :) Unix text files should end with a single newline character (`0x0a`). This one ends with two now. Thus, there is now an empty line at the end (please remove it).

juliusv:
@brian-brazil The alertmanager is still exposing all JSON properties with a capital first letter (e.g. `/api/silences` and `/api/silences/1`). The only place where it's already lower-case is alert data sent *to* Alertmanager from Prometheus.

I propose we fix all of that right after getting this PR in.

wrouesnel:
That should be right now - gedit silently adds newlines, Eclipse doesn't.

juliusv:
Thanks!

PR #80, opened 2015-06-18
HTML template for notifications with FlowDock

t0mk:
I created an HTML template, similar to the existing email body template. I use it for content in the Flowdock message. It's rather generic, and I think it can be used also in other HTML-capable notification receivers like Hipchat.

In Flowdock it looks like:

### DIFF
--- a/manager/notifier.go
+++ b/manager/notifier.go
@@ -20,6 +20,7 @@ import (
 	"flag"
 	"fmt"
 	"html"
+	htmltemplate "html/template"
 	"io"
 	"io/ioutil"
 	"net"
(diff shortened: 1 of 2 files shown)


manager/notifier.go
```diff
@@ -55,10 +56,26 @@ Grouping labels:
...
 {{range $label, $value := .Alert.Payload}}
   {{$label}} = "{{$value}}"{{end}}`))
 
+var contentTmpl = htmltemplate.Must(htmltemplate.New("content").Parse(
+	`<p><b><i>{{.Alert.Description}}</i></b></p>
+
+<div><i>Grouping labels</i></div>
+<ul>
+{{range $label, $value := .Alert.Labels}}
+<li><b>{{$label}}:</b> {{$value}}</li>
```
juliusv: Indent the `li` within the `ul`?

manager/notifier.go
```diff
@@ -370,17 +390,22 @@ func newFlowdockMessage(op notificationOp, config *pb.FlowdockConfig, a *Alert)
 	case notificationOpResolve:
 		status = "resolved"
 	}
+	contentBuf := new(bytes.Buffer)
```
juliusv: Style nit: we only use `new` or `make` when required. Instead use `contentBuf := &bytes.Buffer{}` here (or `var contentBuf bytes.Buffer` and then `&contentBuf` as the argument).

juliusv:
:+1: otherwise. We will want to migrate this to a generalized and user-configurable templating system soon.

juliusv:
@t0mk Sorry, this one slipped through the cracks due to the lack of another ping after the last update. Could you squash the commits (and rebase on current master if necessary).

t0mk:
Yeah, I'll do it on Monday.

t0mk:
uh, let me fix the test

t0mk:
I squashed the commits, fixed the test, and wrote a test for flowdock notification.

juliusv:
@t0mk Thanks! Sorry again - was on vacation :) Now finally merging.

PR #84, opened 2015-06-25
Add runbook and alertmanager URLs to PD+email notifications.

juliusv:
I don't have a way to test all the other notification mechanisms, which
is something we should fix in general. For now, only PagerDuty and email
have the new runbook and alertmanager URL information.

Not very happy with the overall cleanliness of this, and the codebase
overall, of course, but since we need this urgently tomorrow, I hope
this is fine for now.

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,6 +15,8 @@ package main
 
 import (
 	"flag"
+	"fmt"
+	"net"
 	"os"
 	"strings"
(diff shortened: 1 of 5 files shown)


juliusv:
@fabxc

fabxc:
:+1:

Issue #28, opened 2015-02-09
Label name should be a string rather than regex in config.proto


brian-brazil:
Currently the label name is specified as a regex in the config for aggregation. I can't think of a sane label scheme where you'd want to aggregate based on label name rather than value, so I propose we offer `label` instead for simplicity.

juliusv:
:+1: (probably won't get to this soon, but feel free to do it and announce it)

brian-brazil:
This is obsolete with the new alertmanager.

PR #85, opened 2015-06-29
switch to YAML config

fabxc:
@juliusv


Makefile
```diff
@@ -11,7 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
```
juliusv: extra empty line at end of Makefile now

config/config.go
```diff
@@ -15,138 +15,474 @@ package config
 
 import (
 	"fmt"
+	"io/ioutil"
+	"regexp"
+	"strings"
 	"time"
 
-	"github.com/golang/protobuf/proto"
+	"gopkg.in/yaml.v2"
```
juliusv: according to our awesome import grouping rules, this'd actually go in the same group as the `common` one.

config/config.go
```diff
@@ -15,138 +15,474 @@ package config
...
+	if err != nil {
+		return nil, err
+	}
+	return Load(string(content))
+}
 
-// Config encapsulates the configuration of an Alert Manager instance. It wraps
-// the raw configuration protocol buffer to be able to add custom methods to
-// it.
+// Config is the top-level configuration for  Alertmanager's config files.
```
juliusv: nit: double space before "Alertmanager"

config/config.go
```diff
@@ -15,138 +15,474 @@ package config
...
+// Notification configuration definition.
+type NotificationConfig struct {
+	// Name of this NotificationConfig. Referenced from AggregationRule.
+	Name string `yaml:"name"`
+
+	// Notify when resolved.
+	SendResolved bool `yaml:"send_resolved"`
+
+	// Zero or more PagerDuty notification configurations.
+	PagerdutyConfigs []*PagerdutyConfig `yaml:"pagerduty_configs"`
```
juliusv: A pity that repeated elements come at a higher cost in YAML, but probably worth keeping all these repeated.

config/config.go
```diff
@@ -15,138 +15,474 @@ package config
...
+	if len(r.NotificationConfigs) == 0 {
+		return fmt.Errorf("aggregation rule needs at least one notification config")
+	}
+	return checkOverflow(r.XXX, "aggregation rule")
+}
 
-	return nil
+// A regex-based label filter used in aggregations.
+type Filter struct {
+	Name  string  `yaml:"name"`
```
brian-brazil: With this we can close #28

config/config.go
```diff
@@ -15,138 +15,474 @@ package config
...
+	Regex *Regexp `yaml:"regex"`
```
juliusv: This was represented in this way because protobufs didn't have maps. In YAML, it would be better to represent this as key-value pairs. Ok to do in a later change though...

juliusv:
:+1: tests are obviously missing and completely crucial, but they don't make sense at the moment, since we're going to revamp the configuration structure completely as the next step. But then!

fabxc:
Yep.

PR #90, opened 2015-07-09
Set Pushover notification priority to "emergency".

juliusv:


fabxc:
:+1:

Luzifer:
Cool. Thanks! :+1: Thought this would be more discussion…

juliusv:
@Luzifer Well, now that you made me look at it again, I wonder whether "resolved" notifications should not be emergencies :)

Luzifer:
Good point… Maybe they shouldn't be… :D

PR #97, opened 2015-07-21
Alertmanager: Add OpsGenie Webhook notification.

miekg:
This adds notifications to OpsGenie. It is using the webhook api. The
notification_config should look like

  opsgenie_config {
    api_key: "bliep"
    teams: "team-one"
    send_resolved: true
  }

It further sets the alert fingerprint in Alias and adds all the
alert labels into the Details field.

See https://www.opsgenie.com/docs/web-api/alert-api#createAlertRequest
for the details.

It can also close the alert by using the Alias field again. See
https://www.opsgenie.com/docs/web-api/alert-api#closeAlertRequest

### DIFF
--- a/config/config.proto
+++ b/config/config.proto
@@ -90,6 +90,18 @@ message FlowdockConfig {
   optional bool send_resolved = 4 [default = false];
 }
 
+// Configuration for notification via OpsGenie.
+message OpsGenieConfig {
+  // OpsGenie API token.
+  optional string api_key = 1;
+  // List of team names which will be responsible for the alert.
+  repeated string teams = 2;
+  // List of labels that have their values converted to tags.
+  repeated string labels_to_tags = 3;
+  // Notify when resolved.
+  optional bool send_resolved = 4 [default = false];
+}
+
 // Configuration for notification via generic webhook.
 message WebhookConfig {
   // URL to send POST request to.
@@ -116,6 +128,8 @@ message NotificationConfig {
   repeated FlowdockConfig flowdock_config = 7;
   // Zero or more generic web hook notification configurations.
   repeated WebhookConfig webhook_config = 8;
+  // Zero or more OpsGenieConfig notification configurations.
+  repeated OpsGenieConfig opsgenie_config = 9;
 }
 
 // A regex-based label filter used in aggregations.
--- a/manager/notifier.go
+++ b/manager/notifier.go
@@ -27,6 +27,7 @@ import (
 	"net/http"
 	"net/smtp"
 	"os"
+	"strconv"
 	"strings"
 	"sync"
 	"text/template"
@@ -83,6 +84,7 @@ var contentTmpl = htmltemplate.Must(htmltemplate.New("content").Parse(
 var (
 	notificationBufferSize = flag.Int("notification.buffer-size", 1000, "Size of buffer for pending notifications.")
 	pagerdutyAPIURL        = flag.String("notification.pagerduty.url", "https://events.pagerduty.com/generic/2010-04-15/create_event.json", "PagerDuty API URL.")
+	opsgenieAPIURL         = flag.String("notification.opsgenie.url", "https://api.opsgenie.com/v1/json/alert", "OpsGenie API URL.")
 	smtpSmartHost          = flag.String("notification.smtp.smarthost", "", "Address of the smarthost to send all email notifications to.")
 	smtpSender             = flag.String("notification.smtp.sender", "alertmanager@example.org", "Sender email address to use in email notifications.")
 	hipchatURL             = flag.String("notification.hipchat.url", "https://api.hipchat.com/v2", "HipChat API V2 URL.")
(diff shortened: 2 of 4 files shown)


config/config.proto
```diff
@@ -116,6 +126,8 @@ message NotificationConfig {
   repeated FlowdockConfig flowdock_config = 7;
   // Zero or more generic web hook notification configurations.
   repeated WebhookConfig webhook_config = 8;
+  // Zero or more OpsGenieConfig notification configuration.
```
juliusv: s/configuration/configurations/

manager/notifier.go
```diff
@@ -544,6 +545,77 @@ func (n *notifier) sendEmailNotification(to string, op notificationOp, a *Alert)
 	return writeEmailBody(wc, *smtpSender, to, status, a, n.alertmanagerURL)
 }
 
+// opGenieMessageCreate is the request for sending a opsGenie notification. We are not specifying all the
```
juliusv: s/a/an/

manager/notifier.go
```diff
@@ -544,6 +545,77 @@ func (n *notifier) sendEmailNotification(to string, op notificationOp, a *Alert)
...
+// fields. The Details field is populated with the labels from the alert.
+type opsGenieMessageCreate struct {
+	Message     string            `json:"message"`
+	Description string            `json:"description"`
+	Alias       string            `json:"alias"`
+	Teams       []string          `json:"teams"`
+	Details     map[string]string `json:"details"`
+}
+
+// opsGenieMessageClose closes an open alert in OpsGenie
```
juliusv: Period at the end of comment.

miekg: [ Quoting <notifications@github.com> in "Re: [alertmanager] Alertmanager: Ad..." ]
>> @@ -544,6 +545,77 @@ func (n *notifier) sendEmailNotification(to string, op notificationOp, a *Alert)
>>  	return writeEmailBody(wc, *smtpSender, to, status, a, n.alertmanagerURL)
>>  }
>>
>> +// opGenieMessageCreate is the request for sending a opsGenie notification. We are not specifying all the
>> +// fields. The Details field is populated with the labels from the alert.
>> +type opsGenieMessageCreate struct {
>> +	Message     string            `json:"message"`
>> +	Description string            `json:"description"`
>> +	Alias       string            `json:"alias"`
>> +	Teams       []string          `json:"teams"`
>> +	Details     map[string]string `json:"details"`
>> +}
>> +
>> +// opsGenieMessageClose closes an open alert in OpsGenie
>
>Period at the end of comment.

Ack.

Note I still need to fully test this at our end. Hopefully that will happen
today.

manager/notifier.go
```diff
@@ -544,6 +545,77 @@ func (n *notifier) sendEmailNotification(to string, op notificationOp, a *Alert)
...
+		err         error
+	)
+	switch op {
+	case notificationOpTrigger:
+		msg := &opsGenieMessageCreate{
+			Message:     a.Summary,
+			Description: a.Description,
+			Alias:       strconv.FormatUint(uint64(incidentKey), 10),
+			Teams:       config.Teams,
+			Details:     a.Labels,
```
juliusv: This is not including the `a.Payload` labels anywhere, which can be very useful. If the `Details` field is the only one that may contain a nested JSON map, how about adding two sub-properties in `Details` which each contain a map for the grouping and payload labels?

juliusv: Or maybe use the `tags` field in the OpsGenie API for the grouping labels (`a.Labels`)? https://www.opsgenie.com/docs/web-api/alert-api#createAlertRequest

juliusv: Ah dang, those are really just single-word tags, not key-value ones. Guess you'd need the former approach then.

config/config.proto
```diff
@@ -90,6 +90,16 @@ message FlowdockConfig {
...
 }
 
+// Configuration for notification via OpsGenie.
+message OpsGenieConfig {
+  // https://api.opsgenie.com/v1/json/alert
+  optional string webhook_url = 1;
+  // See (https://www.opsgenie.com/docs/web-api/alert-api).
+  optional string api_key = 2;
+  repeated string teams = 3;
+  optional bool send_resolved = 4 [default = false];
```
juliusv: Please add a doc comment to every protobuf field here.

juliusv: (Ok, I guess the link covers them all, but the doc comments here will actually be transformed into Go doc comments on each field in the generated protobuf structs, and then they will seem missing).

manager/notifier.go
```diff
@@ -544,6 +545,77 @@ func (n *notifier) sendEmailNotification(to string, op notificationOp, a *Alert)
...
 }
 
+// opGenieMessageCreate is the request for sending a opsGenie notification. We are not specifying all the
+// fields. The Details field is populated with the labels from the alert.
+type opsGenieMessageCreate struct {
+	Message     string            `json:"message"`
+	Description string            `json:"description"`
+	Alias       string            `json:"alias"`
+	Teams       []string          `json:"teams"`
+	Details     map[string]string `json:"details"`
```
juliusv: You might also want to include the `source` field and set that to `n.alertmanagerURL` like we do for PagerDuty notifications?

manager/notifier.go
```diff
@@ -544,6 +545,77 @@ func (n *notifier) sendEmailNotification(to string, op notificationOp, a *Alert)
...
+// opsGenieMessageClose closes an open alert in OpsGenie
+type opsGenieMessageClose struct {
+	Alias string `json:alias"`
+}
+
+func (n *notifier) sendOpsGenieNotification(op notificationOp, config *pb.OpsGenieConfig, a *Alert) error {
+	// https://api.slack.com/incoming-webhooks
+	var (
+		incidentKey = a.Fingerprint()
+		buf         = []byte{}
```
juliusv: Just do `buf []byte`.

config/config.proto
```diff
@@ -90,6 +90,16 @@ message FlowdockConfig {
   optional bool send_resolved = 4 [default = false];
 }
 
+// Configuration for notification via OpsGenie.
+message OpsGenieConfig {
+  // https://api.opsgenie.com/v1/json/alert
+  optional string webhook_url = 1;
```
juliusv: This URL should probably not be configurable here, but as a global flag (similar to how we do it with most of the others in the file, except for Slack - Slack seems to need auth/identification params in the URL, but those could actually be appended dynamically too and the prefix statically configured).

miekg:
Squashed the commits, added another feature and tested it. PTAL

manager/notifier.go
```diff
@@ -570,6 +572,94 @@ func (n *notifier) sendEmailNotification(to string, op notificationOp, a *Alert)
...
+	Source      string                 `json:"source"`
+	Tags        []string               `json:"tags"`
+	Teams       []string               `json:"teams"`
+	Details     map[string]interface{} `json:"details"`
+}
+
+// opsGenieMessageClose closes an open alert in OpsGenie.
+type opsGenieMessageClose struct {
+	ApiKey string `json:"apiKey"`
+	Alias  string `json:alias"`
```
juliusv: Does the `alias` need quotes too?

juliusv: (Oh, it's just the first quote that is missing - should be added)

juliusv:
:+1: otherwise, thanks!

manager/notifier.go
miekg: [ Quoting <notifications@github.com> in "Re: [alertmanager] Alertmanager: Ad..." ]
>> +// opsGenieMessageClose closes an open alert in OpsGenie.
>> +type opsGenieMessageClose struct {
>> +	ApiKey string `json:"apiKey"`
>> +	Alias  string `json:alias"`
>
>(Oh, it's just the first quote that is missing - should be added)

Wow, sharp :) Yes, missing the quote. Fixed. And pushed again.

/Miek

--
Miek Gieben

juliusv:
:+1: :tada: Thanks, merging!

PR #102, opened 2015-07-24
Add web.external-url flag

### LINKED ISSUE #95: Support web.external-url flag per prometheus
Rather than just `web.hostname`, otherwise alertmanager can't be accessed behind a reverse proxy.

pdf:
Allows setting the full external URL for accessing Alertmanager, takes
precedence over web.hostname.

Fixes #95

I left the old flags there so we don't break any existing users, but they should
probably be dropped at some point.


juliusv:
Thanks! :+1:

Issue #100, opened 2015-07-22
Pushover notifications are broken


farcaller:
Enabling pushover results on the following error:

```
time="2015-07-22T17:10:40Z" level=error msg="Error sending Pushover notification: pushover: request 531f66f26c76006b84864f02a434b858: retry is invalid" file=notifier.go line=640
```

tokens we verified and correct.

juliusv:
@a86c6f7964 @antonlindstrom @Luzifer - you guys have been using Pushover in the past. Any clue about this problem?

antonlindstrom:
I've used this successfully for quite some time and never seen an error like this. My best bet is that the clock is out of sync, that may cause some strange problems.

farcaller:
this is not a clock issue as confirmed by ntp.

a86c6f7964:
I also have never seen this issue. Is it possible that the go library updated and alertmanager has not locked down the version?

juliusv:
That is possible. We don't vendor dependencies in alertmanager yet, but will do so at some point (we really should). If anyone is up for testing this against all current dependencies at that'd be awesome.

Luzifer:
Can confirm the issue with my own builds. Currently having an open alert but no alert in pushover web console.

Luzifer:
Does not look like an update in pushover lib… There is one change since 2014:
https://github.com/thorduri/pushover/compare/a8420a1935479cc266bda685cee558e86dad4b9f...master

This will not break anything… Maybe this is the problem: I'll take a look into the pushover API documentation.

Luzifer:
> To send an emergency-priority notification, the `priority` parameter must be set to `2` and the `retry` and `expire` parameters must be supplied.
> (Source: https://pushover.net/api#priority)

So we need to send those two parameters too when sending emergency notifications. Seeing this the breaking change was 15 days ago with b7919b338bdd6187a2f8a67307bdf62739cafe6d.

farcaller:
Didn't notice there was a fix in #103, sorry. Filed #104 as an alternative though, it should be configurable IMO.

Luzifer:
Closed #103 as #104 is a better solution!

PR #103, opened 2015-07-24
Fix: Pushover notifications not working

Luzifer:
Pushover notifications stopped working because of missing parameters which lead to an API error. Setting these two parameters should fix this issue.


Luzifer:
This should fix #100

Luzifer:
Closing as #104 is more configurable and therefor has a better usability.

PR #104, opened 2015-07-24
Fixed pushover api call with retry and expire flags

### LINKED ISSUE #100: Pushover notifications are broken

farcaller:
Fixes #100

### DIFF
--- a/manager/notifier.go
+++ b/manager/notifier.go
@@ -87,6 +87,8 @@ var (
 	smtpSender             = flag.String("notification.smtp.sender", "alertmanager@example.org", "Sender email address to use in email notifications.")
 	hipchatURL             = flag.String("notification.hipchat.url", "https://api.hipchat.com/v2", "HipChat API V2 URL.")
 	flowdockURL            = flag.String("notification.flowdock.url", "https://api.flowdock.com/v1/messages/team_inbox", "Flowdock API V1 URL.")
+	pushoverRetryTimeout   = flag.Int("notification.pushover.retry-interval", 60, "Interval in seconds at which Pushover should retry pushing a message to receiving users.")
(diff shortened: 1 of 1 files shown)


Luzifer:
Dunno if sane defaults but LGTM :+1:

manager/notifier.go
```diff
@@ -87,6 +87,8 @@ var (
 	smtpSender             = flag.String("notification.smtp.sender", "alertmanager@example.org", "Sender email address to use in email notifications.")
 	hipchatURL             = flag.String("notification.hipchat.url", "https://api.hipchat.com/v2", "HipChat API V2 URL.")
 	flowdockURL            = flag.String("notification.flowdock.url", "https://api.flowdock.com/v1/messages/team_inbox", "Flowdock API V1 URL.")
+	pushoverRetryTimeout   = flag.Int("notification.pushover.retry", 60, "Timeout in seconds between pushover retry attempts.")
```
juliusv: (nit: Please capitalize the proper noun "Pushover" in the help text (same below).)

Reading https://pushover.net/api, I think it's important to clarify that this is not a retry interval that happens in the Alertmanager, but internally in Pushover to retry pushing messages to users once Pushover has received a message. How about: "Interval in seconds at which Pushover should retry pushing a message to receiving users."

And let's call the flag `notification.pushover.retry-interval` to be more clear about its meaning (otherwise it could mean retry count or similar).

About the second flag, how about `notification.pushover.retry-expiry-interval`, with a doc string like "Timeout after which unacknowledged Pushover messages will not be retried further."

juliusv:
Thanks for spotting and fixing, and :+1: besides comments!

farcaller:
Fixed the flag names and description text as advised.

juliusv:
Woot, thanks! :+1:

PR #110, opened 2015-08-17
Allow specifying pushover message priority with a label

NuclearDog:
Basic patch to allow specifying the Pushover message priority by adding a `pushover_priority` label to alerts.

The Pushover application provides [different functionality depending on the priority of the alert](https://pushover.net/api#priority) so allowing the specification of a priority makes the alerting a lot more functional - for instance, I can receive a message with no vibrate/alert tone for a disk that is getting full, but a message which alerts me every minute until I acknowledge it for a down server even while I'm sleeping.

When creating a prometheus alerting rule, you can specify the priority as:

    ALERT DiskSpaceLow
        IF node_filesystem_free / node_filesystem_size < 0.2
        WITH {
                pushover_priority="low"
        }
        SUMMARY "Disk has less than 20% free!"
        DESCRIPTION "Disk has less than 20% free!"

The available priorities are named according to the Pushover documentation listed above:

* `lowest` / -2
* `low` / -1
* `normal` / 0
* `high` / 1
* `emergency` / 2

If no priority is specified, it will default to emergency/2, which was the value previously hardcoded, so this patch should present no issues with backward compatibility.

(This is the first bit of go I've ever written as well as my first github pull request, so apologies if I'm not doing this correctly. I needed this functionality myself and am just hoping it can be useful to someone else as well!)

Thanks!

### DIFF
--- a/manager/notifier.go
+++ b/manager/notifier.go
@@ -708,11 +708,29 @@ func (n *notifier) sendPushoverNotification(token string, op notificationOp, use
 	}
 	alertname := html.EscapeString(a.Labels["alertname"])
 
+	var pushoverPriority int = 2;
+
+	if priority, ok := a.Labels["pushover_priority"]; ok {
+		switch priority {
+			case "lowest":
+				pushoverPriority = -2;
+			case "low":
+				pushoverPriority = -1;
+			case "normal":
+				pushoverPriority = 0;
+			case "high":
+				pushoverPriority = 1;
+			default:
+			case "emergency":
+				pushoverPriority = 2;
+		}
+	}
+
 	// Send pushover message
 	_, _, err = po.Push(&pushover.Message{
 		Title:    fmt.Sprintf("%s: %s", alertname, status),
 		Message:  a.Summary,
-		Priority: pushover.Emergency,
+		Priority: pushoverPriority,
 		Retry:    *pushoverRetryTimeout,
 		Expire:   *pushoverExpireTimeout,
 	})


NuclearDog:
The reason I hardcoded magic values rather than use the constants provided by the library, is because the library alertmanager is using for Pushover didn't have a constant for the `lowest` priority. Looking into it a bit further though, it looks like it also validates the priority value against its out-of-date list... So this patch will work, however using the 'lowest' priority will cause the pushover message to fail to send.

Let me know if you'd rather I close this and make a new PR to remove that / use the library's constants.

brian-brazil:
Alert rules shouldn't be aware of how exactly the alertmanager is configured, as they're different concerns and it makes managing and maintaining alerts much more difficult. A setting like this belongs entirely on the alertmanager side.

juliusv:
Thanks for the PR, but I have to agree with @brian-brazil about not having this kind of coupling. The better way would be to create label-specific notification routing in your configuration file using separate aggregation rules for each severity level. That's still a bit verbose at the moment, but will become nicer after the Alertmanager reworking.

NuclearDog:
Makes sense - if I were to add a priority setting to the pushover_config config block, do you think you guys would accept that PR? I can get by with my solution, but wouldn't mind cleaning it up and resubmitting if it would make it useful/available (and if it turns out to not be too outside my skillset).

juliusv:
@NuclearDog Being able to configure it like that would make sense IMO. But then you need two settings: `trigger_priority` and `resolve_priority`.

NuclearDog:
After the rewrite this isn't really relevant - closing.

PR #112, opened 2015-08-17
Renamed TopicARN to TopicArn

jamiegwatkin:
Renamed TopicARN to TopicArn to match https://github.com/aws/aws-sdk-go/blob/54390aa3615c49d34e593de83ddc72e5ee35f97d/service/sns/examples_test.go#L518.


fabxc:
Thanks a lot! That's the first time not vendoring AM is biting us.
Also, why on earth would they change from idiomatic to unidiomatic Go...

brian-brazil:
To be consistent with all the other languages they support.

fabxc:
Not a good choice, IMO. One should implement client libraries idiomatic to the language they are written in – but not like I could change it.

PR #113, opened 2015-09-02
Fix version info string.

mrwacky42:
Broken by the following commit
e38bb7c Use Makefile.COMMON and add generated files.go

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,18 @@ TARGET   := alertmanager
 
 include Makefile.COMMON
 
+REV        := $(shell git rev-parse --short HEAD 2> /dev/null  || echo 'unknown')
+BRANCH     := $(shell git rev-parse --abbrev-ref HEAD 2> /dev/null  || echo 'unknown')
+HOSTNAME   := $(shell hostname -f)
+BUILD_DATE := $(shell date +%Y%m%d-%H:%M:%S)
+GOFLAGS    := -ldflags \
+	"-X=main.buildVersion $(VERSION)\
+	-X=main.buildRevision $(REV)\
+	-X=main.buildBranch $(BRANCH)\
+	-X=main.buildUser $(USER)@$(HOSTNAME)\
+	-X=main.buildDate $(BUILD_DATE)\
+	-X=main.goVersion $(GO_VERSION)"
+
 web: web/blob/files.go
 
 web/blob/files.go: $(shell find web/templates/ web/static/ -type f)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -16,6 +16,17 @@ TARGET   := alertmanager
 
 include Makefile.COMMON
 
+REV        := $(shell git rev-parse --short HEAD)
```
juliusv: Let's not depend on it being a git repo (for example, this will fail if someone downloads the automatic source zip file that GitHub generates for a release). In Prometheus, this was solved like this:

https://github.com/prometheus/prometheus/blob/master/Makefile.INCLUDE#L66-L67

juliusv:
:+1: otherwise. Long-term we can think whether this should always be set in Makefile.COMMON (and perhaps also used in all other Go binaries).

Makefile
```diff
@@ -16,6 +16,17 @@ TARGET   := alertmanager
 
 include Makefile.COMMON
 
+REV        := $(shell git rev-parse --short HEAD 2> /dev/null  || echo 'unknown')
+BRANCH     := $(shell git rev-parse --abbrev-ref HEAD 2> /dev/null  || echo 'unknown')
+HOSTNAME   := $(shell hostname -f)
```
mrwacky42: NB: ```hostname -f``` on Solaris will try to set the hostname to ```-f```

juliusv: Humm, I guess that's fine, since Solaris is not a supported Go platform? Do you know how it behaves on FreeBSD?

mrwacky42: FreeBSD man pages say it's the same as leenucks.
https://www.freebsd.org/cgi/man.cgi?query=hostname&sektion=1

mrwacky42:
This also never uses the VERSION directly.
How about:
```
GOFLAGS    := -ldflags \
	"-X=main.buildVersion $(VERSION)\
	-X=main.buildBranch $(BRANCH)@$(REV)\
[SNIP]
```

```
./alertmanager
alertmanager, version 0.0.3 (HEAD@982c968)
[SNIP]
```

juliusv:
@mrwacky42 Sounds good. Even better might be to do it exactly like in Prometheus (https://github.com/prometheus/prometheus/blob/master/Makefile.INCLUDE#L66-L76), with a separate version variable, and then do the formatting in build.go?

mrwacky42:
Rebased, now with separate ```Revision``` variable passed in.

juliusv:
@mrwacky42 Thanks! Merging...

PR #114, opened 2015-09-03
Use Go 1.5

brunoqc:


fabxc:
Currently Alertmanager does not experience any problems with 1.4.2. There are issues around 1.5 on the other hand. As this possibly introduces problems, we are currently not super motivated to upgrade.

1.5.1 is more realistic.

PR #115, opened 2015-09-05
Suggest to use email address as silence creator

grobie:
Using email addresses provides a unique identifier with a defined
format. It will reduce the amount of different and ambigious identifiers
used (first, last, full, initials, different case used, etc.) while
still allowing others to use a different format.

I could also imagine making an email address mandatory.

@juliusv @matthiasr


fabxc:
:+1: approved

juliusv:
:+1:

PR #116, opened 2015-09-07
Rearrange alert columns and allow expanding more alert details.

juliusv:
With this change, the alert overview and details look like this:

alerts_ui_change

This properly formats the alerting rule, but moves it into the expandable detailed view, along with other fields.


juliusv:
@grobie 

Regarding putting the summary before the labels, I don't like separating the labels from the alert name, as both together form the identity of the alert, so I prefer keeping them together if possible.

Regarding linkability of individual alert entries, I like the idea, but would keep it for a separate PR.

fabxc:
Wunderschoen.

web/templates/alerts.html
```diff
@@ -9,19 +9,20 @@
     <table class="table table-striped table-bordered table-hover">
       <thead>
         <tr>
+          <th class="alert_expand_col"></th>
           <th>Alert</th>
           <th>Labels</th>
+          <th>Summary</th>
```
brian-brazil: Summary is user-provided and could end up very long, I'd show it only when exapnded.

juliusv: I had that discussion with @grobie. He even wanted it to come first after the alertname because for him, the summary is the thing he is most interested in. Yes, it can become long, but then the table row will just become larger:

long_summary

So it can make the UI somewhat less beautiful, but wouldn't really break it.

brian-brazil: I'd expect that most of the key information in the summary is already covered by the alert name and labels.

For when someone is quickly scanning the table you want the representation to be as compact as possible.

juliusv: That sounds ok to me. But you might want to discuss with @grobie.

web/templates/alerts.html
```diff
@@ -61,6 +61,28 @@
...
+                <tr>
+                  <th>Last received</th>
+                  <td>{{timeSince .LastRefreshed}} ago</td>
+                </tr>
+                <tr>
+                  <th>Description</th>
+                  <td>{{.Alert.Description}}</td>
+                </tr>
+                <tr>
+                  <th>Runbook</th>
```
brian-brazil: This should only be displayed if it's non-empty

juliusv: Ok, updated.

juliusv:
Going to merge this for now. People with different opinions who haven't chimed in yet can send followup PRs.

PR #117, opened 2015-09-08
Sort alerts consistently in web UI.

juliusv:
Yeah, I could have gone through even more effort to sort alphabetically by exact labelset, but don't want
to put too much effort into the old web UI now. Main thing is that now entries will at least be sorted by
their alertname, as well as stay in the same position in general between page loads.

### DIFF
--- a/web/alerts.go
+++ b/web/alerts.go
@@ -15,6 +15,7 @@ package web
 
 import (
 	"net/http"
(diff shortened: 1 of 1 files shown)


juliusv:
@fabxc

brian-brazil:
:+1:

PR #118, opened 2015-09-08
Linkify alert rows.

juliusv:
This does:

* add a bookmark icon to each row that links to the row's ID (the ancient version of Bootstrap we're using doesn't have anchor, link, or pushpin icons yet, so I used the bookmark icon instead)
* upon page load, if the URL contains a hash that points to an alert row, the user is scrolled to that row and its details are expanded


juliusv:
linkify

juliusv:
Gonna fix the pixel alignment on the bookmark icon.

juliusv:
Done. Looks like this now:

linkify

juliusv:
@fabxc

fabxc:
:+1:

web/blob/files.go
```diff
@@ -1911,14 +1914,34 @@ var files = map[string]map[string][]byte{
 			0x3d, 0xfd, 0x94, 0xf8, 0x1f, 0xb1, 0xab, 0xd8, 0x1f, 0x3b, 0x19, 0x00, 0x00,
 		},
 
+		"css/.default.css.swp": {
```
grobie: Wait, why is there a vim swp here? @juliusv

web/blob/files.go
```diff
@@ -1952,83 +1975,135 @@ var files = map[string]map[string][]byte{
...
+			0xdb, 0x13, 0x67, 0x90, 0x43, 0x07, 0x3d, 0xa1, 0x78, 0x9c, 0x0f, 0xff, 0x4f, 0xce, 0x0c, 0x8a, 0x4a, 0xe5, 0x39, 0x8e, 0xc9, 0x7f, 0xaf, 0xa8, 0x6a, 0xfd, 0xa3, 0x50, 0x37, 0xa8,
+			0xf4, 0xc5, 0x5c, 0xd4, 0x56, 0xfe, 0xea, 0x3d, 0x5d, 0xe3, 0xfc, 0xc3, 0x19, 0x36, 0x3f, 0x63, 0x7b, 0x84, 0x23, 0x41, 0xc2, 0xc0, 0x90, 0x46, 0x7a, 0xe5, 0x77, 0x86, 0x5d, 0x2b,
+			0xf4, 0xef, 0x06, 0xaf, 0x12, 0x1d, 0x0e, 0xd7, 0x4f, 0xb4, 0xf8, 0xc4, 0x16, 0xb5, 0xcf, 0x98, 0x61, 0x0d, 0xcc, 0xaa, 0x11, 0x59, 0x44, 0xdd, 0xf5, 0xc9, 0x67, 0x4d, 0x29, 0x2b,
+			0x53, 0x67, 0x0c, 0xb9, 0xe5, 0x9c, 0xd4, 0x7e, 0xfc, 0xe2, 0x8f, 0x4c, 0x5b, 0xfe, 0x33, 0x60, 0xf1, 0x1b, 0x18, 0xdc, 0xc7, 0xf4, 0xa9, 0xf5, 0x4f, 0xe6, 0xc6, 0x90, 0xc2, 0xe9,
+			0x08, 0xa8, 0x27, 0x39, 0x7d, 0x99, 0xb7, 0xcf, 0x1d, 0x67, 0x4b, 0x86, 0x46, 0xc9, 0xcb, 0xae, 0x99, 0x2a, 0xc9, 0x85, 0xed, 0x63, 0x10, 0xfc, 0x0b, 0x4a, 0x2d, 0x72, 0xa6, 0xf7,
+			0x67, 0x46, 0x8f, 0x77, 0x7e, 0x3c, 0xf6, 0x9e, 0xf4, 0xdd, 0xdd, 0x59, 0x74, 0x3e, 0xb8, 0x5b, 0xb6, 0xd7, 0xf7, 0xbf, 0x01, 0xcf, 0xc5, 0x80, 0x1d, 0xe4, 0x10, 0x00, 0x00,
+		},
+
+		".alerts.html.swp": {
```
grobie: Same @juliusv

grobie:
The bookmark adds a new column which causes the detail row not to stretch all columns anymore.

I also don't really like the bookmark icon. It just clutters the UI even more. What I would expect is that if I open a detail view of an alert then the URL automatically updates to include the fingerprint. If I close it, the anchor gets removed as well again. So there wouldn't be the need for any new - and ambigious - icon.

web/blob/files.go
juliusv: Because I had a vim running at that time. I already pushed a fix for that and the other web asset issues yesterday:

https://github.com/prometheus/alertmanager/commit/e00f5b183ed432688de9c02961b36df5c0c70797

juliusv:
Thanks for pointing out the column issue, didn't notice that, will fix.

For the other point, PRs welcome!

PR #120, opened 2015-09-17
Update sample config to include opsgenie example

jbarratt:
I was trying to figure out how to set up opsgenie and ended up digging out the syntax from an older PR. This seems like the canonical place for it to be documented.


brian-brazil:
Thanks!

PR #123, opened 2015-09-21
Basic state persistence for the alertmanager silences

wrouesnel:
Extends the silences.json model to have the memoryAlertManager persist its state to an on-disk JSON file. This allows restarts to avoid re-dispatching sent notifications. The persistence file can be disabled by setting it to a blank string "".

### DIFF
--- a/main.go
+++ b/main.go
@@ -32,6 +32,7 @@ import (
 var (
 	configFile       = flag.String("config.file", "alertmanager.conf", "Alert Manager configuration file name.")
 	silencesFile     = flag.String("silences.file", "silences.json", "Silence storage file name.")
+	memoryManagerFile = flag.String("manager.memory.file", "aggregates.json", "Persistence file for memory alert manager")
 	minRefreshPeriod = flag.Duration("alerts.min-refresh-period", 5*time.Minute, "Minimum required alert refresh period before an alert is purged.")
 	listenAddress    = flag.String("web.listen-address", ":9093", "Address to listen on for the web interface and API.")
 	pathPrefix       = flag.String("web.path-prefix", "/", "Prefix for all web paths.")
@@ -105,8 +106,10 @@ func main() {
 		Silencer:           silencer,
 		Notifier:           notifier,
 		MinRefreshInterval: *minRefreshPeriod,
+		PersistenceFile:	*memoryManagerFile,
 	}
 	alertManager := manager.NewMemoryAlertManager(options)
+	
 	alertManager.SetAggregationRules(conf.AggregationRules())
 	go alertManager.Run()
 
--- a/manager/manager.go
+++ b/manager/manager.go
@@ -18,6 +18,9 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"io/ioutil"
+	"encoding/json"
+	"strconv"
 
 	"github.com/prometheus/log"
 )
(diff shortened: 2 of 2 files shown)


juliusv:
Hey, thanks for sending this! Since we're planning on rewriting the Alertmanager completely (which will include features like this), I'm thinking whether to make this ready for merging. On the other hand, it doesn't have to be perfect then if it gets replaced soon anyways.

Besides minor issues (and lack of tests, which is a general AM problem), there are at least some things that make me unsure about the full correctness here: a restart should at least be treated like a configuration reload to make sure aggregates are (re)associated with the correct (potentially new) rules, so `alertManager.SetAggregationRules(conf.AggregationRules())` should be called *after* loading the persisted file. In general, just serializing and deserializing whole datastructures including pointered objects that weren't initially designed for that makes me a bit nervous - I couldn't immediately see a problem with that though. Are you reasonably confident that's all ok and no references will be broken after reload?

manager/manager.go
```diff
@@ -172,7 +188,85 @@ func NewMemoryAlertManager(o *MemoryAlertManagerOptions) AlertManager {
...
 		notifier:           o.Notifier,
+		persistenceFile:	o.PersistenceFile,
+	}
+}
+
+// Loads a JSON representation of the memoryManager from a file.
+func (s *memoryAlertManager) loadFromFile() error {
+	s.mu.Lock()
+	
+	aggregateJson, err := ioutil.ReadFile(s.persistenceFile)
```
juliusv: `aggregateJson` -> `aggregateJSON`

wrouesnel:
In my (limited) testing it all seemed fine but I'll run it through it's paces a bit more extensively in the next few days. Good point about the rule changes - I'll test that to see what the damage is.

I pulled this together to solve an immediate short term problem (namely, when I do a config reload I inevitably have to take down the alertmanager anyway, so saving my alert emits stops me spamming out a whole bunch of new notifications).

juliusv:
Also noticed that `alertManager.Run()`, which does the loading, is called asynchronously in main.go, so it's possible for the web interface to start up and accept alerts already and those would be removed again when the alertstates are loaded afterwards...

main.go
```diff
@@ -105,8 +106,10 @@ func main() {
 		Silencer:           silencer,
 		Notifier:           notifier,
 		MinRefreshInterval: *minRefreshPeriod,
+		PersistenceFile:	*memoryManagerFile,
```
juliusv: Run `go fmt` over changed files please.

manager/manager.go
```diff
@@ -133,6 +145,8 @@ type memoryAlertManager struct {
 	silencer *Silencer
 	// Notifier for dispatching notifications.
 	notifier Notifier
+	// JSON file which memory alert manager will persist itself to
```
juliusv: Periods at the end of any doc comments.

manager/manager.go
```diff
@@ -172,7 +188,85 @@ func NewMemoryAlertManager(o *MemoryAlertManagerOptions) AlertManager {
 		inhibitor:          o.Inhibitor,
 		silencer:           o.Silencer,
 		notifier:           o.Notifier,
+		persistenceFile:	o.PersistenceFile,
+	}
+}
+
+// Loads a JSON representation of the memoryManager from a file.
+func (s *memoryAlertManager) loadFromFile() error {
+	s.mu.Lock()
```
juliusv: You probably wouldn't even need any locking here if you called this synchronously before `Run` and starting up anything else from main.

wrouesnel:
Since it looks like the new alertmanager is ramping up for roll out, and I'm moving over to it myself, I'm going to close this off for the time being since it won't apply to it.

PR #127, opened 2015-10-05
Add inhibiting test, fix some uncovered issues

juliusv:
The inhibition acceptance test is still failing because the `Subscribe` method needs fixing
(it currently does not return all alerts).


juliusv:
@fabxc

fabxc:
:+1:

Issue #79, opened 2015-06-15
web/static/js/alerts.js should use -web.path-prefix


bluecmd:
Currently the the API is assumed to be located at /api/ which if -web.path-prefix is set it is most likely not.

Looking around how the URLs are constructed, this might be a case of just changing to 'api/..' instead of '/api/..'. I'd be happy to do a pull request either way if I get some guidance on how you want to solve this.

a86c6f7964:
This just him me as well. I would think that you should just bubble up the web.path-prefix using relative urls is not always the most safe.

juliusv:
For Prometheus, we inject the path prefix into the HTML as a global JS variable and then base any self-links on that (https://github.com/prometheus/prometheus/blob/master/web/blob/templates/_base.html#L13). 

So I'd do the same in https://github.com/prometheus/alertmanager/blob/master/web/templates/_base.html and then use that variable in https://github.com/prometheus/alertmanager/blob/master/web/static/js/alerts.js for the API paths.

Actually, let me try to do that right now...

juliusv:
See PR at https://github.com/prometheus/alertmanager/pull/132. I wouldn't want to put much more work into the old Alertmanager at this point, as the new one is making good progress finally. But this should hopefully fix it.

PR #132, opened 2015-10-08
Fix path prefix usage in UI.

juliusv:
Just a quick hotfix before the new Alertmanager comes out.

The path prefix handling is kind of inconsistent in AM (we have both a
-web.path-prefix flag and a -web.external-url flag and with different
trailing-backslash variant than in Prometheus). This should all be
cleaned up in the new one.

Fixes https://github.com/prometheus/alertmanager/issues/79


beorn7:
:+1:

Issue #133, opened 2015-10-13
hipchat room_id as string or int32


triggity:
hipchat room_id's can be the integer number of the room or its string name. https://www.hipchat.com/docs/apiv2/method/send_message

currently, the configuration only allows for int32 room id's.
https://github.com/prometheus/alertmanager/blob/216d9dd46792f7a97ace168e1b1ba92c2509310e/config/config.proto#L48

any way we can change this?

juliusv:
Since the HipChat API defines that as a string, I think it's fine to file a PR that changes that config field to string type and uses that instead of int32.

0x46616c6b:
With the new implementation it is an string. See the [new documentation](https://github.com/prometheus/docs/blob/amdoc/content/docs/alerting/configuration.md#hipchat-receiver-hipchat_config). Also new is the usage of the [Room Notification API](https://www.hipchat.com/docs/apiv2/method/send_room_notification) from Hipchat.

triggity:
Thanks for the update!

PR #134, opened 2015-10-14
 hipchat room_id field to string

triggity:
https://github.com/prometheus/alertmanager/issues/133
hipchat api accepts room names or room id's in room_id field. this changes that.

not sure if you wanted me to run ```make config``` or if that should be done in a seperate commit


fabxc:
Thanks. We are currently rewriting the alertmanager in the `dev` branch. So we will most likely not release another version of the current one.

It might more sense to look into integrating this in the new branch.

Unless you want to run an alertmanager build from master. If so, we can reconsider.

triggity:
it's fine for now. looking forward to the proposed changes in alertmanager

PR #143, opened 2015-10-29
Ensure data directory exists

brian-brazil:
This saves the user having to create it by hand the first time around.

@fabxc


fabxc:
Can you reopen against `dev`. Switched development there – will delete `fabxc/clean`.

PR #146, opened 2015-11-09
Adding documentation to make clear that port is required

trumant:
This caught me out, and I would have appreciated being made aware of it when running `alertmanager -h` instead of finding out when I saw `level=error msg="Error sending email notification: dial tcp: missing port in address` in my alertmanager.log

@juliusv - Hoping I save someone else the trouble.


juliusv:
:+1: Thanks!

/cc @fabxc in case it's relevant for the new version.

Issue #126, opened 2015-09-25
Alert notification templates


brian-brazil:
All fields that we send out should be configurable via Go templates.

Tricky bit is how we handle numeric fields.

Issue #74, opened 2015-06-10
Load templates from file


wrouesnel:
The current email config is in `notifier.go` as a blob of text.

This can pretty obviously be loaded up from a file specified as a configuration option, since email is one of those things you'd want to pretty heavily customize.

I'm just dropping this here so I can reference it later when I get a chance to have a poke at it.

brian-brazil:
More generally, all notifications should be configurable via a templating language. I'm thinking inline in the config, as many will be brief.

brian-brazil:
The new alertmanager supports this.

PR #147, opened 2015-11-16
First-pass at improving template system.

### LINKED ISSUE #135: Emails are not encoded properly
If alert summary and/or description contains non-ascii symbols, it breaks emails.

* There is no content type headers in email and most of mail clients don't use UTF-8 by default - the message body is broken in most of mail clients.
* If non-ascii symbols will end up in email subject, it's even worse. All headers (including subject) need to be encoded properly and if they are not, various sanity checks in mail transport (including virus and spam scanners) could throw these mails away. For example Amavis complains here:

``BAD HEADER Non-encoded 8-bit data (char C3 hex) in message header 'Subject'``

brian-brazil:
- Cut back to bare minimum to make the rest simpler
- Consistency in config naming
- Have one data strucutre that's the same for all templates
- Pass in common labels to templates
- Support templates almost everywhere
- Support multiple SMTP recipients
- Support non-ASCII SMTP headers
- Handle colour logic via templates
- Make $subjects have consistent output, go maps aren't sorted.
- Make tests pass when v6 is disabled

@fabxc 

Fixes #135 #126 #74 

Still needs improvement, but this should be a good base.


config/notifies.go
```diff
@@ -15,62 +15,50 @@ package config
...
 	"fmt"
+	"strings"
 )
 
 var (
 	// DefaultHipchatConfig defines default values for Hipchat configurations.
 	DefaultHipchatConfig = HipchatConfig{
-		ColorFiring:   "purple",
-		ColorResolved: "green",
+		Color:         `{{ if eq .Status "firing" }}purple{{ else }}green{{ end }}`,
```
fabxc: This is the mix of template files and inlined templating I'm not yet sure about.
Then again, this is one of a bunch of configs taken from the old implementation that are not backed by an implementation.

Maybe it makes sense to remove these as well to start unbiased by the old structure when adding new integrations.

config/notifies.go
```diff
@@ -15,62 +15,50 @@ package config
...
-
-		Templates: SlackTemplates{
-			Title:     "slack.default.title",
-			TitleLink: "slack.default.title_link",
-			Pretext:   "slack.default.pretext",
-			Text:      "slack.default.text",
-			Fallback:  "slack.default.fallback",
-		},
+		Color:     `{{ if eq .Status "firing" }}warning{{ else }}good{{ end }}`,
+		Title:     `{{template "slack.default.title" . }}`,
```
fabxc: I see where you are going.
I actually like this approach – it's a clean way to mix between inlining and files.

notify/impl.go
```diff
@@ -62,6 +66,79 @@ func Build(confs []*config.Receiver, tmpl *template.Template) map[string]Fanout
...
 const contentTypeJSON = "application/json"
 
+// End-users should not be exposed to Go's type system,
+// as this will confuse them and prevent simple things like
+// simple equality checks to fail. Map everything to float64/string.
+type templateData struct {
+	Status                      string
+	Alerts                      []templateAlert
+	AlertCommonLabels           map[string]string
+	AlertCommonLabelnamesSorted []string
```
fabxc: The `Sorted` suffix just seems noisy. That can be part of the documentation. As the user has no influence on iteration order anyway the info is not helping much anyway.

notify/impl.go
```diff
@@ -62,6 +66,79 @@ func Build(confs []*config.Receiver, tmpl *template.Template) map[string]Fanout
 
 const contentTypeJSON = "application/json"
 
+// End-users should not be exposed to Go's type system,
```
fabxc: I agree that the `model.LabelName`/`model.LabelValue` types do not need to be exposed. Don't agree with the statement in general though, e.g. in Prometheus templating, I don't think dates and such should be plain strings.

Lots of infrastructure software is written in Go. Most of them (Docker, Kubernetes) allow accessing data by defining templates. Anyone making use of such features can be expected to be able to deal with the well-defined template system.

notify/impl.go
```diff
@@ -62,6 +66,79 @@ func Build(confs []*config.Receiver, tmpl *template.Template) map[string]Fanout
...
+// as this will confuse them and prevent simple things like
+// simple equality checks to fail. Map everything to float64/string.
+type templateData struct {
+	Status                      string
+	Alerts                      []templateAlert
```
fabxc: I would export both types for visibility in the godoc.
This should remain `model.Alerts` in my opinion – of course then labels are no longer strings and such.

I wonder how this would even come up if everything consistently is `model.LabelName` though. Comparisons would work with each other and string literals.
I cannot immediately see rich use-cases that would cause much trouble here.

notify/impl.go
```diff
@@ -183,11 +262,32 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) error {
...
+		return fmt.Errorf("error executing from template: %s", err)
 	}
-	if err := c.Rcpt(n.conf.Email); err != nil {
-		return err
+	addrs, err := mail.ParseAddressList(from)
+	if err != nil {
+		return fmt.Errorf("error parsing from addresses: %s", err)
+	}
+	if len(addrs) != 1 {
+		return fmt.Errorf("error, must be exactly one from address")
```
fabxc: No need to prefix this with error if it's not required to make it a readable statement.

notify/impl.go
```diff
@@ -62,6 +66,79 @@ func Build(confs []*config.Receiver, tmpl *template.Template) map[string]Fanout
...
+	GroupLabels                 map[string]string
+	GroupLabelnamesSorted       []string
+}
+
+type templateAlert struct {
+	Labels      map[string]string
+	Annotations map[string]string
+}
+
+func generateTemplateData(ctx context.Context, as ...*types.Alert) *templateData {
```
fabxc: Summing up the above, I'm not sure about this whole mapping here. I would like to see some use cases causing trouble exposing the original types/data structures with their potentially useful methods.

config/notifies.go
brian-brazil: I'm okay with kicking this one down to the default.tmpl. For things this brief I was thinking it may make sense to do it inline.

There's a few other equivalents that can be replaced by an if in templates, mostly color related.

notify/impl.go
```diff
@@ -268,34 +363,25 @@ func (n *PagerDuty) Notify(ctx context.Context, as ...*types.Alert) error {
...
 		return s
 	}
+	details := make(map[string]string, len(n.conf.Details))
+	for k, v := range n.conf.Details {
+		details[k] = tmpl(v)
+	}
 
 	msg := &pagerDutyMessage{
-		ServiceKey:  n.conf.ServiceKey,
+		ServiceKey:  tmpl(n.conf.ServiceKey),
```
fabxc: What's the use case here? Just having the service key in a different file?

template/template.go
```diff
@@ -49,140 +41,33 @@ func FromGlobs(paths ...string) (*Template, error) {
...
+	err = tmpl.Execute(&buf, data)
 	return buf.String(), err
 }
 
-func (t *Template) ExecuteText(w io.Writer, name string, data interface{}) error {
-	return t.text.ExecuteTemplate(w, name, data)
-}
-
-func (t *Template) ExecuteHTMLString(name string, data interface{}) (string, error) {
+func (t *Template) ExecuteHTMLString(html string, data interface{}) (string, error) {
```
fabxc: Compiling the templates on each execution is an anti-pattern in my opinion. Even though we have it in Prometheus too. Templates are compilable for a reason.

Not feeling very strongly, but I also see it as an advantage to avoid races and as a result broken notifications.
I don't see much benefit to be constantly able to change the templates – it happens to rarely that sending a `SIGHUP` is not too much overhead.

test/acceptance.go
```diff
@@ -86,7 +86,7 @@ func NewAcceptanceTest(t *testing.T, opts *AcceptanceOpts) *AcceptanceTest {
 func freeAddress() string {
 	// Let the OS allocate a free address, close it and hope
 	// it is still free when starting Alertmanager.
-	l, err := net.Listen("tcp", ":0")
+	l, err := net.Listen("tcp4", ":0")
```
fabxc: What was the issue with the version-free one?

fabxc:
Needs rebase.

notify/impl.go
brian-brazil: > I don't think dates and such should be plain strings.

Timestamps are currently float64 in Prometheus templates, the alertmanager doesn't expose any dates yet. I can see us exposing some though for alert starting time, so for consistency we'd be looking at float64 and a humanize function there.

> Anyone making use of such features can be expected to be able to deal with the well-defined template system.

I don't think this is a reasonable expectation. Go's templating is designed for use by the same people writing the code, who will obviously already be familiar with Go's type system and have control of the surrounding code to handle anything that comes up.

What we're doing here is different as we're exposing Go's template system to users who in general will have no knowledge of Go, and may not even know what a type system is. I think losing a handful of minor utility functions is a small price to make things more easily understandable by our whole userbase.

config/notifies.go
fabxc: Actually fine with that. This was just the first one I saw. But it's consistent what is happening in other places below :+1:

notify/impl.go
brian-brazil: Removed.

> I guess you want it to allow in-order iteration over the maps

Yip.

notify/impl.go
brian-brazil: > I would export both types for visibility in the godoc.

Done.

> I cannot immediately see rich use-cases that would cause much trouble here.

I can imagine users wanting to compare the labelname with the labelvalue, and also across with annotations.

The more common things that'll break will be string functions, they can only take in one type.

notify/impl.go
brian-brazil: Done

notify/impl.go
brian-brazil: I can imagine routing being plumbed through from service discovery. I had a system many years back that used this technique very successfully for email, and can see the same being done for pagerduty etc.

template/template.go
brian-brazil: You'll need to send a SIGHUP anyway as they're coming from the config. As mentioned on IRC this is just a first pass, ultimately they should be compiled at the config load stage both for efficiency and to catch syntax errors earlier.

test/acceptance.go
brian-brazil: Breaks if you've disabled v6 on your Linux machine, such as I have on my entire home network.

brian-brazil:
> Needs rebase.

I've already done so, I suspect you accidentally `--forced` over #144 so the datadir change is coming up again.

notify/impl.go
fabxc: I see :+1:

template/template.go
fabxc: :+1:

test/acceptance.go
fabxc: I always understood this version-less format to chose any working one – okay then.

fabxc: Actually, the normal stdlib `ListenAndServe()` function, which is essentially used everywhere also uses [this](https://golang.org/src/net/http/server.go#L1868).

Accordingly none of our tools serving HTTP content should work for you then – confused.

brian-brazil: As far as I can tell it's when the test tries to connect to `[::1]` that things break.

fabxc: I guess with a normal port it defaults to ipv4 and with :0 to ipv6 then or something – doesn't matter.

notify/impl.go
```diff
@@ -62,6 +66,85 @@ func Build(confs []*config.Receiver, tmpl *template.Template) map[string]Fanout
...
 
+// TemplateData is the data passed to notification templates.
+// End-users should not be exposed to Go's type system,
+// as this will confuse them and prevent simple things like
+// simple equality checks to fail. Map everything to float64/string.
+type TemplateData struct {
+	Status            string
+	Alerts            []TemplateAlert
+	AlertCommonLabels map[string]string
+
```
fabxc: You don't see any use for common annotations? Both suffer from the problem of changing as more alerts are sorted into the group. But if for example `summary` is common across all of them, that can be useful for some templates.

brian-brazil: I don't think the use case is as strong as for labels, I can see potential use but only in a similar way to what I'm doing to subject which won't tend to have space to fit a summary. I figure a wait and see approach is best, we can always add it in later.

fabxc:
Okay, let's merge so we have the base in place :ship:

PR #149, opened 2015-11-17
slack: add icon_emoji parameter

qzio:
I was not able to run the tests using a docker container based on `golang:latest`.

I managed to get `make config` to run, but it rewrote large parts of `config/generated/config.pb.go`.
However, I only commited the parts relevant to this PR.

Is there a canonical docker image where all the deps are installed, or installable which I can use to do `make config` and `make test`?

### DIFF
--- a/config/config.proto
+++ b/config/config.proto
@@ -76,6 +76,10 @@ message SlackConfig {
   optional string color_resolved = 4 [default = "good"];
   // Notify when resolved.
   optional bool send_resolved = 5 [default = false];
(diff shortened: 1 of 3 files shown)


fabxc:
Thanks. We are currently rewriting the alertmanager in alertmanager/dev. So we will most likely not release another version of the current one.

It might more sense to look into integrating this in the new branch.

brian-brazil:
This was added in the new AM in #253

PR #152, opened 2015-11-19
Remove data races

fabxc:
@beorn7

### DIFF
--- a/dispatch.go
+++ b/dispatch.go
@@ -29,6 +29,7 @@ type Dispatcher struct {
 	marker types.Marker
 
 	aggrGroups map[*Route]map[model.Fingerprint]*aggrGroup
+	mtx        sync.RWMutex
 
 	done   chan struct{}
 	ctx    context.Context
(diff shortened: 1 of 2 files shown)


dispatch.go
```diff
@@ -273,6 +281,17 @@ func (ag *aggrGroup) String() string {
 	return fmt.Sprintf("%v", ag.fingerprint())
 }
 
+func (ag *aggrGroup) alertSlice() []*types.Alert {
+	ag.mtx.RLock()
+	defer ag.mtx.RUnlock()
+
+	var alerts []*types.Alert
```
beorn7: Not that it matters a lot, but you could
```
alerts := make([]*types.Alert, 0, len(ag.alerts))
```
to avoid reallocations.

beorn7:
:+1: (Feel free to disregard my comment if you prefer a few characters fewer over a few allocations fewer.)

dispatch.go
beorn7: Actually, the most idiomatic way is probably the following (in lieu of lines 288–292):

`return append(make([]*types.Alert, 0, len(ag.alerts)), ag.alerts...)`

beorn7:
Please disregard my updated comment. :-)

dispatch.go
fabxc: `ag.alerts` is a map, not a slice. So this does not work.

fabxc:
Merging for now, see comment.

PR #153, opened 2015-11-19
Improve default subject template

fabxc:
This adds the number of alerts in the grouping. Helpful information for few characters.
Currently this is the total number. We might want to provide an easy way to separate resolved and firing alerts in the templating.

Fixes a mis-placed `)`.

Adds label name for everything but "alertname". By default that seems more sensible.

Before:
`[firing] ManyOpenConnections dd bazooka ()))critical)`
After:
` [firing:3] ManyOpenConnections cluster=dd service=bazooka (severity=critical)`

Longer but still fits into a single PagerDuty notification. If being truncated, mostly the additional common labels would be affected, which is okay.

@brian-brazil 

Added so that "alertname" is always at the first position of `data.GroupLabelnames`.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -120,11 +120,19 @@ func generateTemplateData(ctx context.Context, as ...*types.Alert) *TemplateData
 		data.Alerts = append(data.Alerts, alert)
 	}
 
+	sortStart := 0
 	for k, v := range groupLabels {
 		data.GroupLabels[string(k)] = string(v)
-		data.GroupLabelnames = append(data.GroupLabelnames, string(k))
+
+		// Always have the alertname label at the first position.
+		if k == model.AlertNameLabel {
+			data.GroupLabelnames = append([]string{string(k)}, data.GroupLabelnames...)
+			sortStart = 1
+		} else {
+			data.GroupLabelnames = append(data.GroupLabelnames, string(k))
+		}
 	}
-	sort.Strings(data.GroupLabelnames)
+	sort.Strings(data.GroupLabelnames[sortStart:])
 
 	if len(alerts) >= 1 {
 		common := alerts[0].Labels.Clone()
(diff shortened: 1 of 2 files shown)


brian-brazil:
That all seems okay, bar adding the labelnames. There could easily be 10 labels, and some page transport mechanisms (such as certain email->SMS gateways in the US) you need to keep strictly below 140 characters if you want reliable delivery.

fabxc:
I would rather ditch the common labels than label names to be honest.
For us at least, it's mostly `severity="critical"`, wich I do not care about.

brian-brazil:
The purpose of grouping labels is to prevent many alerts of one class from each creating their own notification. You're using it mainly for cluster, service and alertname but that's not the only case that's going to be common. 

I'd expect many users not to have alertname as a grouping label, just cluster+service (if you tend to have pager storms when something goes bad, this is one way to mitigate this). Similarly you may have a 'shard' label and while in the cases where several shards are bad we can't do much with that, if only one shard is being dodgy having that in the subject would be useful to determine which dashboards you need to look at.

I think that for our default subject that needs to be generic enough to handle as many use cases in as useful a fashion as practical that we should have the common labels included.

fabxc:
All true and fair points. The part on keeping the common labels makes total sense.
I would still argue that dropping the label names will be confusing for many people. Not everyone who is on-call defined the alerts, maybe doesn't even know the setup of the affected system.
Especially cryptic identifiers like clusters, shards, etc. are hard to grasp without it.

If I understood correctly, your opinion is that a page is only for calling you to a laptop anyway, where you see those label names then. I've talked to other people with extensive on-call experience though who see this differently.

After all these are just defaults. For those I would go for slightly more verbosity until the opposite has proven to be saner.

brian-brazil:
> Not everyone who is on-call defined the alerts, maybe doesn't even know the setup of the affected system.
> Especially cryptic identifiers like clusters, shards, etc. are hard to grasp without it.

I'd disagree on that point, the form of things like clusters and machines tend to be very recognisable without providing any context - both within an organisation and without. For example if I say `eu-west-1`, `LHR01` or `ip-10-43-17-98` you'll likely know what those are (and if you're heavy on the operations side also exactly where it is).  Similarly with job labels, though those are more org-specific.

If you don't know these basic aspects of how your production environment is laid out, then you shouldn't be oncall. This is going to be part of the standard new-hire and oncall training within an organisation.

Things more application specific like shards and customers are trickier. For those not too familiar with the system (say a large shared oncall rotation) these are unlikely to be grouping labels, so they'll be in the parens and unlikely to cause confusion with the most "standard" grouping labels. For those familiar, they'll know what they mean without the labelname.

In essence, I see the labelnames as syntactic noise in a place where space is very limited. Things like android notifications and subject lines in your inbox gmail are going to truncate. It's my belief that fitting in more labels for those with some experience is going to be of more benefit than hand-holding someone who's brand new to oncall who likely needs the whole email anyway.

> If I understood correctly, your opinion is that a page is only for calling you to a laptop anyway, where you see those label names then. I've talked to other people with extensive on-call experience though who see this differently.

If you're getting pages where your first reaction isn't to run to your laptop, then they shouldn't be pages as they're not urgent.

fabxc:
Fair enough. Generally, we shouldn't assume though that an organization has an on-call setup we consider good. At the same time it shouldn't be a blocker for us.

I removed the label names from the default. In the end everyone can overwrite as it suits them.

jimmidyson:
I know this is closed but I'm just trying out the new AM for the first time & the subject line is way too noisy IMO. Also missing spaces it seems. e.g. I see a subject line of:

    [firing:1] ServiceDown (quickstart-springboot-webmvcspring-bootquickstartsblackbox.default.svc:80kubernetes-servicesspringboot-web

As these are discovered from kubernetes, they are arbitrary k/v pairs so the idea that you'd be able to identify what they mean from their value as @brian-brazil mentioned above doesn't hold true.

brian-brazil:
This is more in indication that there's too much noise in your target labels. There is useful information there, we should fix the spaces.

I note that this will come in at just under the size of an SMS.

jimmidyson:
I've fixed up the space already.

Agreed that labels are too noisy in this case but the point is that arbitrary labels are possible & are hard to understand without context (name).

Issue #131, opened 2015-10-07
Nil pointer derefrence


a86c6f7964:
I have been having an alert going off for a while. It is just a warning. It also is flapping back and forth. I think that this is possibly causing this Nil pointer. I am not sure what is needed to make this issue more clear, so tell me if I can help.

I notice that there is a lot of logs about 'Recomputing notifications'
If I run `increase(prometheus_notifications_latency_milliseconds_count[10m])` I get 40. This to mean implies that prometheus is sending the alert every 15 seconds. There are 3 alerts going off. So this does not really make sense to me that I would see the 'Recomputing' every second.

I am running version 216d9dd:0.0.4 of prom/alertmanager from the docker registry, and I am running 64349aa:0.15.1 of prometheus.

```
Oct 07 15:58:51 ip-172-31-23-215.us-west-2.compute.internal docker[18598]: time="2015-10-07T15:58:51Z" level=info msg="Recomputing notification outputs (active alerts have changed)" file=manager.go line=398
Oct 07 15:58:52 ip-172-31-23-215.us-west-2.compute.internal docker[18598]: time="2015-10-07T15:58:52Z" level=info msg="Recomputing notification outputs (active alerts have changed)" file=manager.go line=398
Oct 07 15:58:53 ip-172-31-23-215.us-west-2.compute.internal docker[18598]: time="2015-10-07T15:58:53Z" level=info msg="Recomputing notification outputs (active alerts have changed)" file=manager.go line=398
Oct 07 15:58:54 ip-172-31-23-215.us-west-2.compute.internal docker[18598]: time="2015-10-07T15:58:54Z" level=info msg="Recomputing notification outputs (active alerts have changed)" file=manager.go line=398
Oct 07 15:58:55 ip-172-31-23-215.us-west-2.compute.internal docker[18598]: time="2015-10-07T15:58:55Z" level=info msg="Recomputing notification outputs (active alerts have changed)" file=manager.go line=398
Oct 07 15:58:56 ip-172-31-23-215.us-west-2.compute.internal docker[18598]: time="2015-10-07T15:58:56Z" level=info msg="Recomputing notification outputs (active alerts have changed)" file=manager.go line=398
Oct 07 15:58:57 ip-172-31-23-215.us-west-2.compute.internal docker[18598]: time="2015-10-07T15:58:57Z" level=info msg="Recomputing notification outputs (active alerts have changed)" file=manager.go line=398
[...]

Issue #122, opened 2015-09-18
Null pointer dereference


qed-:
I received the following panic, not doing anything in particular, there were two active alerts at the time, one silenced. Let me know if there's any further information I can provide.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x4e532d]

goroutine 15 [running]:
github.com/prometheus/alertmanager/manager.(*memoryAlertManager).removeExpiredAggregates(0xc208054280)
        /home/mavdev/dave/prometheus/alertmanager/alertmanager/.build/gopath/src/github.com/prometheus/alertmanager/manager/manager.go:282 +0x23d
github.com/prometheus/alertmanager/manager.(*memoryAlertManager).runIteration(0xc208054280)
        /home/mavdev/dave/prometheus/alertmanager/alertmanager/.build/gopath/src/github.com/prometheus/alertmanager/manager/manager.go:395 +0x33
github.com/prometheus/alertmanager/manager.(*memoryAlertManager).Run(0xc208054280)
        /home/mavdev/dave/prometheus/alertmanager/alertmanager/.build/gopath/src/github.com/prometheus/alertmanager/manager/manager.go:408 +0x9a
created by main.main
        /home/mavdev/dave/prometheus/alertmanager/alertmanager/main.go:111 +0xd50

goroutine 1 [chan receive, 15 minutes]:
github.com/prometheus/alertmanager/manager.(*notifier).Dispatch(0xc2080fc420)
        /home/mavdev/dave/prometheus/alertmanager/alertmanager/.build/gopath/src/github.com/prometheus/alertmanager/manager/notifier.go:813 +0x54
main.main()
        /home/mavdev/dave/prometheus/alertmanager/alertmanager/main.go:157 +0x15c0

goroutine 14 [chan receive]:
main.func·001()
        /home/mavdev/dave/prometheus/alertmanager/alertmanager/main.go:85 +0x84
created by main.main
        /home/mavdev/dave/prometheus/alertmanager/alertmanager/main.go:90 +0x659

goroutine 16 [IO wait, 3 minutes]:
net.(*pollDesc).Wait(0xc208108fb0, 0x72, 0x0, 0x0)
        /usr/lib/golang/src/net/fd_poll_runtime.go:84 +0x47
net.(*pollDesc).WaitRead(0xc208108fb0, 0x0, 0x0)
[...]

Issue #82, opened 2015-06-19
Null pointer dereference when reloading config via HUP


wrouesnel:
Just a contribution of an interesting crash case. Noting here for my own and others reference.
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x48d82d]
goroutine 14 [running]:
github.com/prometheus/alertmanager/manager.(*memoryAlertManager).removeExpiredAggregates(0xc20803d720)
/home/wrouesnel/src/scripts/go/src/github.com/prometheus/alertmanager/.build/gopath/src/github.com/prometheus/alertmanager/manager/manager.go:282 +0x23d
github.com/prometheus/alertmanager/manager.(*memoryAlertManager).runIteration(0xc20803d720)
/home/wrouesnel/src/scripts/go/src/github.com/prometheus/alertmanager/.build/gopath/src/github.com/prometheus/alertmanager/manager/manager.go:395 +0x33
github.com/prometheus/alertmanager/manager.(*memoryAlertManager).Run(0xc20803d720)
/home/wrouesnel/src/scripts/go/src/github.com/prometheus/alertmanager/.build/gopath/src/github.com/prometheus/alertmanager/manager/manager.go:408 +0x9a
created by main.main
/home/wrouesnel/src/scripts/go/src/github.com/prometheus/alertmanager/main.go:82 +0xb42
goroutine 1 [chan receive, 4 minutes]:
github.com/prometheus/alertmanager/manager.(*notifier).Dispatch(0xc2080fc8a0)
/home/wrouesnel/src/scripts/go/src/github.com/prometheus/alertmanager/.build/gopath/src/github.com/prometheus/alertmanager/manager/notifier.go:639 +0x54
main.main()
/home/wrouesnel/src/scripts/go/src/github.com/prometheus/alertmanager/main.go:128 +0x1380
goroutine 39 [IO wait, 1 minutes]:
net.(*pollDesc).Wait(0xc2080a3100, 0x72, 0x0, 0x0)
/home/wrouesnel/opt/go/src/net/fd_poll_runtime.go:84 +0x47
net.(*pollDesc).WaitRead(0xc2080a3100, 0x0, 0x0)
/home/wrouesnel/opt/go/src/net/fd_poll_runtime.go:89 +0x43
[...]

PR #154, opened 2015-11-20
manager: Guard agg.Rule against dereferencing when nil.

octo:
This bug has been around since at least July, and multiple people have reported it in #82, #122 and #131. I hope that pulling this despite the rewrite is easier than closing yet another duplicate.


brian-brazil:
Thanks!

PR #155, opened 2015-11-23
Remove 'assets' from default target

fabxc:
There are issues with changing timestamps for unmodified files
in go-bindata+git.

@beorn7


beorn7:
:+1:

PR #156, opened 2015-11-23
PR with changes after code review

beorn7:
Now to be reverse-reveiewed.

@fabxc Happy to discuss details in person.


types/match.go
```diff
@@ -80,20 +82,21 @@ func NewMatcher(name model.LabelName, value string) *Matcher {
...
 // expression which is used for matching.
-func NewRegexMatcher(name model.LabelName, value string) (*Matcher, error) {
-	re, err := regexp.Compile(value)
-	if err != nil {
-		return nil, err
+func NewRegexMatcher(name model.LabelName, re *regexp.Regexp) *Matcher {
+	value := strings.TrimSuffix(strings.TrimPrefix(re.String(), "^(?:"), ")$")
+	if len(re.String())-len(value) != 6 {
+		// Any non-anchored regexp is a bug.
+		panic(fmt.Errorf("regexp %q not properly anchored", re.String()))
```
fabxc: No need for the call to `String()` I think.

template/template.go
```diff
@@ -16,19 +16,22 @@ package template
 import (
 	"bytes"
 
-	html_tmpl "html/template"
-	text_tmpl "text/template"
```
fabxc: For package renaming underscores are actually a common pattern.

main.go
```diff
@@ -24,7 +24,7 @@ import (
 	"path/filepath"
 	"strings"
 	"syscall"
-	template_text "text/template"
+	texttemplate "text/template"
```
fabxc: As below, underscore is not an anti-pattern for that use case.
I generally prefer the common ground to be the prefix rather than the suffix.

fabxc:
Looks all good in general :+1: 
Comments are more of a style thing, don't care too much.

types/match.go
beorn7: True. Fixed.

template/template.go
beorn7: I couldn't find a single example in the stdlib, outside of weird, close-to-C code.
Might be, of course, that you don't really need named imports in the stdlib.

Of course, there is the exception of something_test, but that's special and ever more of a reason to avoid underscores in normal package (re-)names.

So I'd like to avoid underscores until proven otherwise.

main.go
beorn7: Don't feel strongly here.

So template first, but I'll make it consistent with the other case, and keep it abbreviated.

PR #158, opened 2015-11-24
Add OpsGenie notifier

jimmidyson:
Seems to be working OK, although resolving alerts seems a bit hit & miss - not due to this code I don't think as the OpsGenie API calls are pretty simple, but more the firing/resolve flow.

/cc @fabxc

### DIFF
--- a/config/notifies.go
+++ b/config/notifies.go
@@ -47,6 +47,12 @@ var (
 		Text:      `{{template "slack.default.text" . }}`,
 		Fallback:  `{{template "slack.default.fallback" . }}`,
 	}
+
+	// DefaultOpsGenieConfig defines default values for OpsGenie configurations.
+	DefaultOpsGenieConfig = OpsGenieConfig{
+		Description: `{{template "opsgenie.default.description" .}}`,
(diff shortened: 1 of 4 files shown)


config/notifies.go
```diff
@@ -280,3 +286,27 @@ func (c *WebhookConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface.
+func (c *OpsGenieConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	*c = DefaultOpsGenieConfig
+	type plain OpsGenieConfig
+	if err := unmarshal((*plain)(c)); err != nil {
+		return err
+	}
+	if c.APIKey == "" {
+		return fmt.Errorf("missing service key in OpsGenie config")
```
fabxc: s/service/API/

Might have been forgotten in other places too after renaming to APIKey.

fabxc:
:+1: cool thanks!

jimmidyson:
Updated.

PR #159, opened 2015-11-25
Sync alert filtering with URL.

fabxc:
This commit keeps the alert filters synchronized with URL search parameters
so they can be shared.

@beorn7


beorn7:
:+1: modulo "I can't read Angular".

PR #160, opened 2015-11-25
Iterate over templating

fabxc:
I was about to do some templating and gave it another iteration. Data structures moved into `template/`.
Added some functions I consider necessary. `pairs` being a way for ordered iteration that I think is understandable to everyone and simplifies our data structure and also the usage for simple ordered iteration (no use of `index` function and reference to parent scope).

I want a first round of feedback before continuing with templates themselves.

@brian-brazil @beorn7

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -25,7 +25,6 @@ import (
 	"net/mail"
 	"net/smtp"
 	"os"
-	"sort"
 	"strings"
 	"time"
 
@@ -68,93 +67,6 @@ func Build(confs []*config.Receiver, tmpl *template.Template) map[string]Fanout 
 }
 
 const contentTypeJSON = "application/json"
-
-// TemplateData is the data passed to notification templates.
-// End-users should not be exposed to Go's type system,
-// as this will confuse them and prevent simple things like
-// simple equality checks to fail. Map everything to float64/string.
-type TemplateData struct {
-	Status            string
-	Alerts            []TemplateAlert
-	AlertCommonLabels map[string]string
-
-	// AlertCommonLabelnames is sorted.
-	AlertCommonLabelnames []string
-	GroupLabels           map[string]string
-
-	// GroupLabelnames is sorted.
-	GroupLabelnames []string
-}
-
-// TemplateAlert holds one alert for notification templates.
-type TemplateAlert struct {
-	Labels      map[string]string
-	Annotations map[string]string
-}
-
-func generateTemplateData(ctx context.Context, as ...*types.Alert) *TemplateData {
-	alerts := types.Alerts(as...)
-
-	groupLabels, ok := GroupLabels(ctx)
-	if !ok {
-		log.Error("missing group labels")
-	}
-
-	data := &TemplateData{
-		Status:                string(alerts.Status()),
-		Alerts:                make([]TemplateAlert, 0, len(alerts)),
-		AlertCommonLabels:     map[string]string{},
-		AlertCommonLabelnames: []string{},
-		GroupLabels:           map[string]string{},
-		GroupLabelnames:       make([]string, 0, len(groupLabels)),
-	}
-
-	for _, a := range alerts {
-		alert := TemplateAlert{
-			Labels:      make(map[string]string, len(a.Labels)),
-			Annotations: make(map[string]string, len(a.Annotations)),
-		}
-		for k, v := range a.Labels {
-			alert.Labels[string(k)] = string(v)
-		}
-		for k, v := range a.Annotations {
-			alert.Annotations[string(k)] = string(v)
-		}
-		data.Alerts = append(data.Alerts, alert)
-	}
-
-	sortStart := 0
-	for k, v := range groupLabels {
-		data.GroupLabels[string(k)] = string(v)
-
-		// Always have the alertname label at the first position.
-		if k == model.AlertNameLabel {
-			data.GroupLabelnames = append([]string{string(k)}, data.GroupLabelnames...)
-			sortStart = 1
-		} else {
-			data.GroupLabelnames = append(data.GroupLabelnames, string(k))
-		}
-	}
-	sort.Strings(data.GroupLabelnames[sortStart:])
-
-	if len(alerts) >= 1 {
-		common := alerts[0].Labels.Clone()
-		for _, a := range alerts[1:] {
-			for ln, lv := range common {
-				if a.Labels[ln] != lv {
-					delete(common, ln)
-				}
-			}
-		}
-		for k, v := range common {
-			data.AlertCommonLabels[string(k)] = string(v)
-			data.AlertCommonLabelnames = append(data.AlertCommonLabelnames, string(k))
-		}
-	}
-	sort.Strings(data.AlertCommonLabelnames)
-
-	return data
-}
 
 // Webhook implements a Notifier for generic webhooks.
 type Webhook struct {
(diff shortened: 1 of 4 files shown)


notify/impl.go
```diff
@@ -562,7 +477,7 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) error {
 		if err != nil {
 			return
 		}
-		s, err = n.tmpl.ExecuteTextString(name, data)
+		s, err = n.tmpl.ExecuteTextString(name, template.NewData(groupLabels(ctx), as...))
```
brian-brazil: This is recalculating the template data each time, it'd be better to do it just once

template/template.go
```diff
@@ -76,3 +84,137 @@ func (t *Template) ExecuteHTMLString(html string, data interface{}) (string, err
 	err = tmpl.Execute(&buf, data)
 	return buf.String(), err
 }
+
+type FuncMap map[string]interface{}
+
+var DefaultFuncs = FuncMap{
+	"upper": strings.ToUpper,
```
brian-brazil: These names are problematic as they don't follow the Go names for these functions. What we should probably do is rename `title` to `toTitle` over in Prometheus.

template/template.go
```diff
@@ -76,3 +84,137 @@ func (t *Template) ExecuteHTMLString(html string, data interface{}) (string, err
...
+			}
+		}
+		sort.Strings(keys[sortStart:])
+
+		for _, k := range keys {
+			pairs = append(pairs, Pair{k, m[k]})
+		}
+		return pairs
+	},
+	"firing": func(alerts []Alert) []Alert {
```
brian-brazil: I'm not sure these make sense, a given alert template is only going to be called with either all firing or all resolved alerts. Passing in a single parameter to the templates with this information would seem better to me.

template/template.go
fabxc: I understand making them equivalent to the Go functions. But why then making it inconsistent again by turning `title` into `toTitle`?

template/template.go
```diff
@@ -76,3 +84,137 @@ func (t *Template) ExecuteHTMLString(html string, data interface{}) (string, err
...
+	"title": strings.Title,
+	// pairs allows for in-order iteration of key/value pairs.
+	"pairs": func(m map[string]string) []Pair {
+		var (
+			pairs     = make([]Pair, 0, len(m))
+			keys      = make([]string, 0, len(m))
+			sortStart = 0
+		)
+		for k := range m {
+			if k == string(model.AlertNameLabel) {
```
brian-brazil: This isn't quite safe to apply to annotations, but is probably the way to go

notify/impl.go
fabxc: Ah yes, missed that this was within a function.

template/template.go
fabxc: Currently we are handing in all alerts of a group, firing or resolved. The total status of a notification is firing unless all alerts it contains are resolved.

template/template.go
fabxc: Yes, this would be a very rare edge case with little impact. Worth it for the gain of simplicity.

template/default.tmpl
```diff
@@ -1,19 +1,17 @@
-{{ define "__subject" }}{{$dot := .}}[{{ .Status }}:{{ .Alerts | len }}] {{ range $k := .GroupLabelnames }}{{ index $dot.GroupLabels $k }} {{ end }}{{if gt (len .AlertCommonLabels) (len .GroupLabels) }}({{ range $k := .AlertCommonLabelnames }}{{ if eq "" (index $dot.GroupLabels $k) }}{{ index $dot.AlertCommonLabels $k }}{{ end }} {{ end }}){{ end }}{{ end }}
-
-{{ define "slack.default.fallback" }}
-{{ template "__subject" . }}
-{{ end }}
+{{ define "__alertmanager" }}AlertManager{{ end }}
+{{ define "__alertmanagerURL" }}{{ .ExternalURL }}{{ end }}
+{{ define "__subject" }}{{$dot := .}}[{{ .Status | upper }}:{{ .Alerts | firing | len }}] {{ range $i, $p := .GroupLabels | pairs }}{{ $p.Value }} {{ end }}{{if gt (len .CommonLabels) (len .GroupLabels) }}({{ range $i, $p := .CommonLabels | pairs }}{{ if eq "" (index $dot.GroupLabels $p.Name) }}{{ $p.Value }} {{ end }}{{ end }}){{ end }}{{ end }}
```
brian-brazil: > `{{ .Alerts | firing | len }}`

This is misleading, if it's a resolved notification it'll incorrectly report 0.

> `{{ range $i, $p := .GroupLabels | pairs }}`

Hmm, might `sortedPairs` be clearer?

template/template.go
brian-brazil: Go also has a `toTitle` function, so it'd be consistent in that way.

fabxc: Ah, didn't even realize. Makes sense then.

template/template.go
brian-brazil: I think that's going to be very confusing as a naieve template is going to blindly print out everything (like we're doing right now). I'm also generally not a fan of resolved notifications as I see them as mostly spam. This way of doing things doesn't feel quite right to me, from either a operations-theory or ease-of-use standpoint.

Was there a particular reason you went for doing it this way?

template/default.tmpl
fabxc: The first depends on the other comment.
The second, I would consider this part of the documentation. If I don't care about order I have no reason to use that function in the first place.

template/template.go
fabxc: For PagerDuty and friends, resolving alerts as they stop firing is completely reasonable (only happens if the whole group is resolved). For more info-level stuff like email, I agree that it's noise but not everyone will feel the same.

Generally, this way ensures that my last notification for a group always gives me the last updates. Consider a dashboard behind a webhook, which might want that information in a timely manner. An alternative option would be to always keep resolved and firing separate and send two notifications if there is a mix within a group. That's generally more noisy though and can also be misleading.

I'm open to alternatives. Any suggestions?

brian-brazil: Thinking a bit and considering how the system is designed, I think the sanest thing is to add an options to all the notification methods where this is not something ~everyone would want (i.e. not Pagerduty) to filter out the resolved alerts. The default would be to filter out the resolved alerts.

We'd still need the default templates to handle all this in case someone flipped the option, but it'll put most users on the right path.

template/default.tmpl
brian-brazil: > The second, I would consider this part of the documentation. If I don't care about order I have no reason to use that function in the first place.

It depends on your expectations for how a map behaves. If you're used to say an older version of Python you might expect them to come out ordered in the first place. If I were to read `.GroupLabels | pairs` as a non-Go user I'd probably think that whatever structure that is isn't iterable in Go, and that `pairs` was working around that.

template/template.go
fabxc: We have that in one way: https://github.com/prometheus/alertmanager/blob/dev/notify/notify.go#L255-L257
Seems like this would make more sense to be configured per receiver configuration then?

brian-brazil: Yeah, that's what the old-AM has.

template/default.tmpl
fabxc: Thinking again tomorrow, `sortedPairs` seems more intuitive regardless of docs.

template/template.go
fabxc: Updated addressing most comments. This will be subject to a later change.

template/template.go
```diff
@@ -76,3 +84,137 @@ func (t *Template) ExecuteHTMLString(html string, data interface{}) (string, err
 	err = tmpl.Execute(&buf, data)
 	return buf.String(), err
 }
+
+type FuncMap map[string]interface{}
+
+var DefaultFuncs = FuncMap{
+	"toUpper": strings.ToUpper,
+	"toLower": strings.ToLower,
+	"toTitle": strings.ToTitle,
```
brian-brazil: https://github.com/prometheus/prometheus/issues/1241 filed to get this updated in Prometheus

brian-brazil:
:+1:

PR #161, opened 2015-11-26
Improve PagerDuty templating

fabxc:
WIP on some nits.

@brian-brazil thoughts on that as a default level of information?

This stuff a string list of alerts into a single field. Having full JSON objects might be an option too. But more complicated and not generally supported by clients (e.g. OpsGenie).

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -340,7 +340,7 @@ func (c *Receiver) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return checkOverflow(c.XXX, "receiver config")
 }
 
-// Regexp encapsulates a regexp.Regexp and makes it YAML marshallable.
+// Regexp encapsulates a regexp.Regexp and makes it YAML marshalable.
 type Regexp struct {
 	*regexp.Regexp
 }
--- a/config/config.go
+++ b/config/config.go
@@ -20,6 +20,7 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"time"
 
 	"github.com/prometheus/common/model"
 	"gopkg.in/yaml.v2"
@@ -181,6 +182,8 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 // DefaultGlobalConfig provides global default values.
 var DefaultGlobalConfig = GlobalConfig{
+	ResolveTimeout: model.Duration(5 * time.Minute),
+
 	PagerdutyURL:    "https://events.pagerduty.com/generic/2010-04-15/create_event.json",
 	OpsGenieAPIHost: "https://api.opsgenie.com/",
 }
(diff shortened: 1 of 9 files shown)


config/config.go
```diff
@@ -193,6 +193,7 @@ type GlobalConfig struct {
 	SlackURL        string `yaml:"slack_url"`
 	PagerdutyURL    string `yaml:"pagerduty_url"`
 	OpsGenieAPIHost string `yaml:"opsgenie_api_host"`
+	ExternalURL     string `yaml:"external_url"`
```
brian-brazil: We have this as a flag on Prometheus, we should keep this consistent

config/notifies.go
```diff
@@ -120,7 +126,7 @@ func (c *EmailConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	}
 	c.Headers = normalizedHeaders
 	if _, ok := c.Headers["Subject"]; !ok {
-		c.Headers["Subject"] = DefaultEmailSubject
```
brian-brazil: This change will mean that if the user specifies a subject with a case other than `Subject`, it'll break in the for loop above. It's messy, but the reason I did things this way.

jimmidyson:
OpsGenie does support a JSON map for details (up to 8k chars long) - see `details` param documented at https://www.opsgenie.com/docs/web-api/alert-api#createAlertRequest

template/default.tmpl
```diff
@@ -1,22 +1,38 @@
...
+{{ define "slack.default.titlelink" }}{{ template "__alertmanagerURL" }}{{ end }}
 {{ define "slack.default.text" }}{{ template "__subject" . }}{{ end }}
 
+
 {{ define "pagerduty.default.description" }}{{ template "__subject" . }}{{ end }}
 {{ define "pagerduty.default.client" }}{{ template "__alertmanager" . }}{{ end }}
-{{ define "pagerduty.default.clientURL" }}{{ template "__alertmanagerURL" . }}{{ end }}
+{{ define "pagerduty.default.clientURL" }}{{ template "__alertmanagerURL" }}{{ end }}
+
+{{ define "pagerduty.default.instances" }}{{ range . }}
```
brian-brazil: We should probably have this as a generic __text_alert_list similar to __subject.

config/notifies.go
```diff
@@ -92,6 +97,7 @@ type EmailConfig struct {
 	From      string            `yaml:"from"`
 	Smarthost string            `yaml:"smarthost,omitempty"`
 	Headers   map[string]string `yaml:"headers"`
+	Subject   string            `yaml:"subject"`
```
brian-brazil: Subject is a header, I don't see much point in treating it specially.

fabxc:
@jimmidyson good to know about the limit, thanks.

config/config.go
fabxc: Everything that's trivial to change at runtime should not be a flag if there is a config file anyway.
I would rather change it in Prometheus.

config/notifies.go
fabxc: Yea, I'll revert that.

config/config.go
brian-brazil: I don't think that this is a setting that's going to change at runtime, as it's more an infrastructure-level thing.

fabxc: Me neither, still find it preferable. But not feeling strongly right now if it unblocks the rest.

brian-brazil: Let's not change it until we figure out what the division is.

Having things that are easy to change in the config is doing things the wrong way around. It's things that need to be easy to change or that are too complex to represent in a flag that need to go in the config (and thus implicitly be easy to change).

fabxc: That's another angle. I can see arguments for both sides.
Went back to the flag for now – should be ready then.

fabxc:
Added configurable resolve timeout.

brian-brazil:
:+1:

PR #162, opened 2015-11-28
Use title rather than toTitle.

brian-brazil:
It turns out that 'title case' is a thing in unicode,
in the same way upper/lower case is.

What our users will want is the first character of a word
being titlecased, which `title` will do.

@fabxc 

I don't think we should expose `toTitle`, as I'd expect most users are actually looking for `title` and this would just cause confusion.

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -93,7 +93,7 @@ type FuncMap map[string]interface{}
 var DefaultFuncs = FuncMap{
 	"toUpper": strings.ToUpper,
 	"toLower": strings.ToLower,
-	"toTitle": strings.ToTitle,
+	"title":   strings.Title,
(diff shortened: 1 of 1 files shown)


template/template.go
```diff
@@ -93,7 +93,7 @@ type FuncMap map[string]interface{}
 var DefaultFuncs = FuncMap{
 	"toUpper": strings.ToUpper,
 	"toLower": strings.ToLower,
-	"toTitle": strings.ToTitle,
+	"title": strings.title,
```
fabxc: `strings.Title`, `go fmt`

brian-brazil: I'm clearly not fully awake yet, fixed.

fabxc:
:+1:

PR #166, opened 2015-11-30
Transfer template functions into types+methods

fabxc:
Thinking a bit, having type specific functions might not be the easiest to understand for users. This is another approach using dedicated types (based on strings only).

I expect the case of excluding certain labels etc. to be frequent enough to add a `Without` method and `strings.Join`. For the default subject this simplifies the previous conditional with scoping issues into 

`{{ with .CommonLabels.Without .GroupLabels.Names }}{{ join .Values " " }}{{ end }}`

Also reduces the need for another conditional to remove the trailing space.

@brian-brazil

### DIFF
--- a/template/default.tmpl
+++ b/template/default.tmpl
@@ -1,14 +1,13 @@
 {{ define "__alertmanager" }}AlertManager{{ end }}
 {{ define "__alertmanagerURL" }}{{ .ExternalURL }}/#/alerts?receiver={{ .Receiver }}{{ end }}
 
-
-{{ define "__subject" }}{{$dot := .}}[{{ .Status | toUpper }}{{ if eq .Status "firing" }}:{{ .Alerts | firing | len }}{{ end }}] {{ range .GroupLabels | sortedPairs }}{{ .Value }} {{ end }}{{ if gt (len .CommonLabels) (len .GroupLabels) }}({{ range .CommonLabels | sortedPairs }}{{ if eq "" (index $dot.GroupLabels .Name) }}{{ .Value }} {{ end }}{{ end }}){{ end }}{{ end }}
+{{ define "__subject" }}[{{ .Status | toUpper }}{{ if eq .Status "firing" }}:{{ .Alerts.Firing | len }}{{ end }}] {{ .GroupLabels.SortedPairs.Values | join " " }} {{ if gt (len .CommonLabels) (len .GroupLabels) }}({{ with .CommonLabels.Remove .GroupLabels.Names }}{{ .Values | join " " }}{{ end }}){{ end }}{{ end }}
 {{ define "__description" }}{{ end }}
 
 {{ define "__text_alert_list" }}{{ range . }}Labels: 
-{{ range .Labels | sortedPairs }} - {{ .Name }} = {{ .Value }}
+{{ range .Labels.SortedPairs }} - {{ .Name }} = {{ .Value }}
 {{ end }}Annotations: 
-{{ range .Annotations | sortedPairs }} - {{ .Name }} = {{ .Value }}
+{{ range .Annotations.SortedPairs }} - {{ .Name }} = {{ .Value }}
 {{ end}}
 {{ end }}{{ end }}
 
--- a/template/template.go
+++ b/template/template.go
@@ -94,51 +94,89 @@ var DefaultFuncs = FuncMap{
 	"toUpper": strings.ToUpper,
 	"toLower": strings.ToLower,
 	"title":   strings.Title,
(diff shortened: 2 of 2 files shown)


template/template.go
```diff
@@ -94,46 +94,7 @@ var DefaultFuncs = FuncMap{
...
-	"resolved": func(alerts []Alert) []Alert {
-		res := []Alert{}
-		for _, a := range alerts {
-			if a.Status == string(model.AlertResolved) {
-				res = append(res, a)
-			}
-		}
-		return res
-	},
+	"join":    strings.Join,
```
brian-brazil: You should switch the order of the arguments so it's easier to use in a pipeline.

fabxc: I was considering that but figured you wouldn't like it for differing from stdlib.
Apparently not the case.

brian-brazil: I've flipped a few over in Prometheus for just this reason, it'd make your example:
```
{{ .CommonLabels.Without .GroupLabels.Names | index "Values" |  join " " }}
```

template/template.go
```diff
@@ -141,6 +102,79 @@ type Pair struct {
...
+	}
+	sort.Strings(keys[sortStart:])
+
+	for _, k := range keys {
+		pairs = append(pairs, Pair{k, kv[k]})
+	}
+	return pairs
+}
+
+// Without returns a copy of the key/value set without the given keys.
```
brian-brazil: I'm not sure about this function. It isn't related to the `with` statement that's in the templates, and to someone not familiar with Go's templates this looks like a field access rather than a method call.

I think we might be better off keeping this as a separate function.

template/template.go
fabxc: I find the use of `index` very unintuitive and not worth the few saved characters. Especially in this case where we want to access a method.

I changed it already, the `{{ with }}` remains and inside it's ``{{ .Values | join " " }}`.

brian-brazil:
I like the general idea, the Without function should probably remain separate though.

template/template.go
fabxc: Relating to the `with` statement goes a bit far I think. We should make things reasonably convenient. But if one wants to do templating there's no way around learning the basics to a reasonable degree.

If it is a separate function though, it will actually be `without` rather than `myvar.Without` which makes it more similiar to `with`.

brian-brazil: What I'm saying is that calling this `without` will incorrectly make users think there's some relation syntactically or semantically to `with`.

Calling it remove or setDifference or something would avoid this.

fabxc: Yea, that's the easiest solution. Done.

template/default.tmpl
```diff
@@ -1,14 +1,13 @@
 {{ define "__alertmanager" }}AlertManager{{ end }}
 {{ define "__alertmanagerURL" }}{{ .ExternalURL }}/#/alerts?receiver={{ .Receiver }}{{ end }}
 
-
-{{ define "__subject" }}{{$dot := .}}[{{ .Status | toUpper }}{{ if eq .Status "firing" }}:{{ .Alerts | firing | len }}{{ end }}] {{ range .GroupLabels | sortedPairs }}{{ .Value }} {{ end }}{{ if gt (len .CommonLabels) (len .GroupLabels) }}({{ range .CommonLabels | sortedPairs }}{{ if eq "" (index $dot.GroupLabels .Name) }}{{ .Value }} {{ end }}{{ end }}){{ end }}{{ end }}
+{{ define "__subject" }}[{{ .Status | toUpper }}{{ if eq .Status "firing" }}:{{ .Alerts.Firing | len }}{{ end }}] {{ range .GroupLabels.SortedPairs }}{{ .Value }} {{ end }}{{ if gt (len .CommonLabels) (len .GroupLabels) }}({{ with .CommonLabels.Remove .GroupLabels.Names }}{{ .Values | join " " }}{{ end }}){{ end }}{{ end }}
```
brian-brazil: This can be simplified further, the if is no longer required as the with covers it being empty already.

`{{ range .GroupLabels.SortedPairs }}{{ .Value }}` becomes `{{ .GroupLabels.SortedPairs.Values | join " " }}`

brian-brazil:
:+1: 

I'm not quite happy with the function being a method, but we can try it and see how it goes. The templates up to now have avoided any OOP.

fabxc:
Since they depend on certain non-primitive input types, I think it's better to attach them directly to those types rather than having different functions with generic names that take different, possibly non-obvious, types.

Especially `Names` and `Values` are now easy to use on `Pairs` and `KV` with the same result without any runtime magic. From the user end I find it more intuitive.

PR #168, opened 2015-12-01
Statically compile default templates

fabxc:
This simplifies deployment for all the people using default templates.
It also prevents people from messing with default templates rather than creating proper overrides.

@brian-brazil


brian-brazil:
:+1:

PR #170, opened 2015-12-02
Remove default.tmpl from Docker image

jimmidyson:
Now that default templates are compiled in.


jimmidyson:
cc @fabxc

fabxc:
We could have removed the creation of the template dir and the config file entry.

PR #174, opened 2015-12-03
Hide secret configuration data

fabxc:
@brian-brazil

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -26,6 +26,16 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+var patAuthLine = regexp.MustCompile(`((?:api_token|api_key|service_key|api_url):\s+)(".+"|'.+'|[^\s]+)`)
+
+// Secret is a string that must not be revealed on marshaling.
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -26,6 +26,16 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
+var patAuthLine = regexp.MustCompile(`((?:api_token|api_key|service_key|api_url):\s+)(".+"|'.+'|[^\s]+)`)
```
brian-brazil: I don't see an `api_url` in the config

config/config.go
```diff
@@ -26,6 +26,16 @@ import (
...
+
+// Secret is a string that must not be revelead on marshaling.
```
brian-brazil: revealed

fabxc: Fixed.

brian-brazil:
:+1: 

Having something as generic as `api_url` hidden may come back to bite us as URLs are not generally secure (proxies log them for example).

PR #179, opened 2015-12-06
[RFC] Add postgres storage backend and allow selecting providers.

wrouesnel:
This commit does an initial refactor of the code to support the ql,mem and a new postgres backend as backing storage for the alertmanager.

The postgres backend is introduced to provide a persistent storage layer that avoids QL, which may have concurrency bugs at the present time. The SQL notification layer is split up, as SQL differs too much between implementations to really make it practical to have a "merged" mode.

The storage backend selection code is preliminary, and probably needs some architecting to make it a more modular for the frontend.

It's worth noting that with QL I usually see a crash in a few hours at the moment, but the postgres backend seems solid (though obviously not entirely practical or desireable for a lightweight application - I just happen to have a postgres database available on my notification server).


provider/postgres/postgres.go
```diff
@@ -0,0 +1,661 @@
...
+	tx, err := db.Begin()
+	if err != nil {
+		return nil, err
+	}
+	if _, err := tx.Exec(createNotifyInfoTable); err != nil {
+		tx.Rollback()
+		return nil, err
+	}
+
+	// XXX(bug-ish): The selection of pending alerts uses a NOT IN clause
```
fabxc: All this noise below was to work around QL behavior I couldn't really explain myself. Probably not necessary for Postgres though, right?

provider/ql/ql.go
```diff
@@ -19,24 +19,25 @@ import (
 	"fmt"
 	"sync"
 
-	"github.com/cznic/ql"
+	qldriver "github.com/cznic/ql"
```
fabxc: `ql` doesn't collide with the package being named `ql` itself.

fabxc:
Thanks. It would be interesting to know why QL is causing trouble for you but has been working for everyone else (a handful of people only though) so far.
The initial choice fell on QL because we wanted to be stand-alone without using CGO. Doesn't seem worth the trouble and I'm personally okay with CGO. The stand-alone part is very important to me though, which is why I would say we should use SQLite.

Multiple backends are generally a great idea. I would rather wait with that until everything else has stabilized. 

The memory backed provider was initially used during implementation but eventually fell behind as I switched to persistence. So it doesn't fulfill the contract right now.

For now going for a single SQLite backend seems most reasonable to me. Your Postgres work could be used at a later point to provide that as a backend too.

wrouesnel:
Postgres is a practical choice - my alert box has a PG database already on it, and lib/pg is pure go so I was fairly confident it would work. This code has been running flawlessly for over a day now, so everything else is looking good.

I've filed an issue open with cznic about this and an update's already been pushed for 1 of the data races it uncovered - but it looks like there's still another https://github.com/cznic/ql/issues/113

fabxc:
I understand that the Postgres solution is very practical for you as you've already set it up. We cannot generally require a running Postgres database to run Alertmanager though. Since QL apparently has issues, we need a more stable embedded choice.

I did an initial implementation using SQLite. It still has issues of delayed inserts though that I can hopefully get rid off.

Providing a multitude of possible backends at this point is difficult, since we might add features that require us to alter tables, adjust interfaces, etc.
Then having to adjust 3+ backends accordingly will generate a lot of work and expose a lot of surface area for regressions. I absolutely want to consider it, but not before we have a well-working default with some milage on it.

wrouesnel:
Agree 100%. I submitted the PR mostly to prove that switching from QL was possible. SQLite is definitely a better option.

vtolstov:
Im need ql, why not providw build tags for different storages?
08 дек. 2015 г. 5:37 пользователь "Will Rouesnel" <notifications@github.com>
написал:

fabxc:
@vtolstov We can provide multiple storages within the same binary without a problem (c.f. the PR code which does just that). The reasoning is that at this early stage maintaining multiple backends doesn't seem to be the best idea.

wrouesnel:
Closing this PR since sqlite3 support got merged, which accomplishes the same thing.

PR #181, opened 2015-12-07
Update webhook for new alertmanager.

brian-brazil:
As we've a new data format the version number needs a bump.
Also make alerts with no annotations easier to handle for
users on the receiving end.

@fabxc


fabxc:
:+1:

PR #183, opened 2015-12-08
Switch to SQLite persistence

fabxc:
@beorn7 @brian-brazil


fabxc:
@wrouesnel you might want to try out this branch as well.

fabxc:
Because UnmarshalYAML is bottom-up. We apply the global SMTP sender if the local one is not set after that receiver config has been parsed. Thus the writing to the header map doesn't kick in here. 
This is the simplest solution to fix it without touching much else.

brian-brazil:
:+1:

PR #184, opened 2015-12-08
Add better mail template

fabxc:
@brian-brazil


template/default.tmpl
```diff
@@ -32,190 +32,89 @@
...
-</td>
-</tr>
-  {{ end }}
-
+<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
+<html xmlns="http://www.w3.org/1999/xhtml" xmlns="http://www.w3.org/1999/xhtml" style="font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif; box-sizing: border-box; font-size: 14px; margin: 0;">
+<head>
+<meta name="viewport" content="width=device-width" />
+<meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+<title>Actionable emails e.g. reset password</title>
```
brian-brazil: I don't think this is what you meant

template/email.html
```diff
@@ -0,0 +1,392 @@
...
+}
+</style>
+</head>
+
+<body itemscope itemtype="http://schema.org/EmailMessage">
+
+<table class="body-wrap">
+  <tr>
+    <td></td>
+    <td class="container" width="600">
```
brian-brazil: A fixed width may not look great on very wide screens

template/email.html
```diff
@@ -0,0 +1,392 @@
...
+                    <strong>[{{ .Alerts.Firing | len }}] Firing</strong>
+                  </td>
+                </tr>
+                {{ end }}
+                {{ range .Alerts.Firing }}
+                <tr>
+                  <td class="content-block">
+                    <strong>Labels</strong><br />
+                    {{ range .Labels.SortedPairs }}{{ .Name }} = {{ .Value }}<br />{{ end }}
+                    <strong>Annotations</strong><br />
```
brian-brazil: There well be no annotations by default, we should probably exclude them in that case to avoid clutter.

template/email.html
```diff
@@ -0,0 +1,392 @@
...
+                    {{ range .Annotations.SortedPairs }}{{ .Name }} = {{ .Value }}<br />{{ end }}
+                  </td>
+                </tr>
+                {{ end }}
+                {{ if gt (len .Alerts.Resolved) 0 }}
```
brian-brazil: There should probably be a notable vertical gap between resolved and firing, if there's a very long list it'll be easier to notice when scrolling that way.

template/email.html
fabxc: 600 is standard width for responsive mail templates.
Not that I've any idea what I'm doing but that's why I picked a pre-made thing from a company that does nothing but email.

fabxc:
Fixes were made.

brian-brazil:
:+1:

PR #196, opened 2015-12-17
Move send_resolved parameter to notifier configuration

fabxc:
@brian-brazil

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -19,22 +19,29 @@ import (
 )
 
 var (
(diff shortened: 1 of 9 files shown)


config/notifiers.go
```diff
@@ -19,22 +19,29 @@ import (
 )
 
 var (
+	// DefaultWebhookConfig defines default vaues for Webhook configurations.
```
brian-brazil: values

brian-brazil:
:+1:

PR #208, opened 2016-01-05
Readd Hipchat Integration

0x46616c6b:


doc/examples/simple.yml
```diff
@@ -114,4 +114,11 @@ receivers:
 
 - name: 'team-DB-pager'
   pagerduty_configs:
-  - service_key: <team-DB-key>
\ No newline at end of file
+  - service_key: <team-DB-key>
+- name: 'team-X-hipchat'
+  hipchat_configs:
+    -
```
brian-brazil: The style here is inconsistent with the rest of the file

config/notifiers.go
```diff
@@ -67,6 +67,18 @@ var (
...
+	// DefaultHipchatConfig defines default values for Hipchat configurations.
+	DefaultHipchatConfig = HipchatConfig{
+		NotifierConfig: NotifierConfig{
+			VSendResolved: false,
+		},
+		Color:         `{{ if eq .Status "firing" }}red{{ else }}green{{ end }}`,
+		From:          `{{ template "hipchat.default.from" . }}`,
+		Notify:        `{{ template "hipchat.default.notify" . }}`,
+		Message:       `{{ template "hipchat.default.message" . }}`,
+		MessageFormat: `{{ template "hipchat.default.message_format" . }}`,
```
brian-brazil: You're not using this as a template anymore, so just put in `text`

template/default.tmpl
```diff
@@ -20,6 +20,12 @@
 {{ define "slack.default.text" }}{{ end }}
 
 
+{{ define "hipchat.default.from" }}{{ template "__alertmanager" . }}{{ end }}
+{{ define "hipchat.default.notify" }}false{{ end }}
+{{ define "hipchat.default.message" }}{{ template "__subject" . }}{{ end }}
+{{ define "hipchat.default.message_format" }}text{{ end }}
```
brian-brazil: This line can be removed

brian-brazil:
This generally looks okay to me, just some minor comments.

config/notifiers.go
```diff
@@ -67,6 +67,18 @@ var (
 		Fallback:  `{{ template "slack.default.fallback" . }}`,
 	}
 
+	// DefaultHipchatConfig defines default values for Hipchat configurations.
+	DefaultHipchatConfig = HipchatConfig{
+		NotifierConfig: NotifierConfig{
+			VSendResolved: false,
```
fabxc: Why false by default. Slack is quite similar and has it as true.
I don't think there's one right default, but consistency is important.

brian-brazil: Slack being true is an oversight, I've already sent a PR to correct it.

Resolved notifications are almost pure spam. Unfortunately some people come believe that they're essential in the same way that email alert spam is believed to be essential. To discourage this, the default should be to not send them so that our users are less likely to get caught in a downwards spiral of being overloaded by notifications for which they take no action.

It should only be enabled by default for systems where they do something meaningful with it, such as Pagerduty and friends. Otherwise it should be strongly discouraged.

fabxc: Don't really care either way.

config/notifiers.go
```diff
@@ -188,6 +200,41 @@ func (c *SlackConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	*c = DefaultHipchatConfig
+	type plain HipchatConfig
+	if err := unmarshal((*plain)(c)); err != nil {
+		return err
+	}
+	if c.RoomID == "" {
+		return fmt.Errorf("missing room id in Hipchat config")
+	}
+
+	if c.AuthToken == "" {
```
fabxc: For Slack the webhook URL already contains the authentication information and can be set in the `global` section. For PagerDuty for example that makes little sense as you have different service keys that influence the destination of notifications.

For Slack and Hipchat that is determined via channel/room though, so I imagine one would rarely have more than one token/key/etc.

Should we make this settable via the `global` section?

0x46616c6b: I absolutely agree to make this configurable in the global section. I was a little bit confused while it is inconsistent.

config/notifiers.go
```diff
@@ -188,6 +200,41 @@ func (c *SlackConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
 
+// HipchatConfig configures notifications via Hipchat.
+type HipchatConfig struct {
+	NotifierConfig `yaml:",inline"`
+
+	APIURL        string `yaml:"api_url"`
+	AuthToken     Secret `yaml:"auth_token"`
+	RoomID        string `yaml:"room_id"`
+	From          string `yaml:"from"`
+	Notify        string `yaml:"notify"`
```
fabxc: Is this a boolean?

brian-brazil:
:+1:

0x46616c6b:
I changed the Notify parameter to bool

fabxc:
Awesome, thanks! :+1:

0x46616c6b:
Yeah, i will open a PR in the docs asap.

PR #209, opened 2016-01-05
Don't send resolved to Slack by default

brian-brazil:
Slack is a general chat system, it has no notion
of resolved messages. Default it to false to avoid
spamming people as we do with all other such systems.

@fabxc

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -56,7 +56,7 @@ var (
 	// DefaultSlackConfig defines default values for Slack configurations.
 	DefaultSlackConfig = SlackConfig{
 		NotifierConfig: NotifierConfig{
-			VSendResolved: true,
+			VSendResolved: false,
 		},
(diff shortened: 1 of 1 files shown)


grobie:
I wonder if that matches the expectation of most users. The Pagerduty slack integration sends resolved messages by default to name the most prominent example. I don't mind to much as it can be changed, but the current default makes more sense to me.

brian-brazil:
See comment at https://github.com/prometheus/alertmanager/pull/208#discussion_r48877203

I've come across many monitoring systems that spam hipchat and slack with pointless messages, that doesn't mean that we should do it too (Pagerduty being only one of them). We should be providing users with good defaults that encourage sane practices.

grobie:
I don't agree that every resolved message is entirely pointless, but the noise:signal ratio is probably high enough to warrant your suggested default.

:+1:

PR #210, opened 2016-01-05
Fixing UI paths to be relative to index.html

sjoerdmulder:
Using relative paths from the `index.html` file solves the issue when running the alertmanager behind a reverse proxy (for authentication) on a different path (like `/alertmanager/`)

@fabxc I have only modified the static resources, not sure if something in go needs to be regenerated?

### DIFF
--- a/ui/app/index.html
+++ b/ui/app/index.html
@@ -1,32 +1,32 @@
 <!DOCTYPE html>
 <html lang="en" ng-app="am">
 <head>
-	<script src="/lib/jquery.min.js"></script>
-	<script src="/lib/moment.min.js"></script>
+	<script src="lib/jquery.min.js"></script>
+	<script src="lib/moment.min.js"></script>
 
-	<script src="/lib/angular.min.js"></script>
-	<script src="/lib/angular-sanitize.min.js"></script>
-	<script src="/lib/angular-route.min.js"></script>
-	<script src="/lib/angular-resource.min.js"></script>
-	<script src="/lib/angular-moment.min.js"></script>
+	<script src="lib/angular.min.js"></script>
+	<script src="lib/angular-sanitize.min.js"></script>
+	<script src="lib/angular-route.min.js"></script>
+	<script src="lib/angular-resource.min.js"></script>
+	<script src="lib/angular-moment.min.js"></script>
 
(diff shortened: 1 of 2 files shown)


grobie:
This would add a hard dependency on never introducing any structures with sub paths, or changing this (weird) angular routing schema. In all other projects projects we generate such paths with a function AFAIK. Could we do the same here?

sjoerdmulder:
This doesn't create a hard dependency, structures with subpaths are fine, everything is relative from the `index.html` file so if you change a path you need to update it (but you need todo that anyway, no difference here). 

The angular routing scheme (which is completly unrelated from static resources or api resources) is unaffected by this change. 

Yes you could also do this by a function that has some configuration from the server (prometheus you mean?) but it isn't required if you just use relative paths (relative from the location if the `index.html` file) Since the `index.html` file is they only file that is the entrypoint.

grobie:
The filesystem location of index.html is irrelevant, the browser will
construct the asset URL relative to the current HTML path. If index.html is
served under `/foo/bar`, the assets will be tried to be loaded from
`/foo/lib/jquery.min` which will fail.

The current Angular routing scheme uses the hashbang approach. We discussed
changing that to the HTML5 routing scheme already. That would generate
paths like `/silences/342dsfq` instead of `/#/silences/342dsfq`. This will
quickly cause problems then.

sjoerdmulder:
I wasn't aware that you want to use the non-hashbang approach. This will
not work then indeed.

I basically tried to get the current 0.1.0-beta release working behind my
reverse proxy and this was the result. Pull request can be closed if you
want.
Keep up the good work with the Prometheus tools!
Op 5 jan. 2016 23:51 schreef "Tobias Schmidt" <notifications@github.com>:

grobie:
Your use case is absolutely relevant. The probelm that alertmanager doesn't
load assets when run behind a reverse proxy is yet another regression of
the rewrite. The old alertmanager handled that fine.

In general, have a look at the -web.external-url flag which is supported by
most prometheus components and also accepted by the rewrite. It's not
enough to simply use relative locations as alertmanager also needs to know
its own location when generating notifications.

grobie:
Please follow #212 for updates.

PR #213, opened 2016-01-06
remove checking auth token in config/notifiers.go

0x46616c6b:
I found a bug, if you configure a global `hipchat_auth_token`, then it doesnt start and print the message `missing auth token in Hipchat config`. So in the global configuration is the handling implemented which overrides the token.


fabxc:
Thanks :+1:

PR #216, opened 2016-01-07
Remove template section from example config

fabxc:
Since our default templates are now compiled into the binary, this is only
relevant to advanced users and will cause trouble when running a derivative
of the example config within the repository directory.

@RichiH @brian-brazil


brian-brazil:
:+1:

RichiH:
+1

RichiH:
To qualify that +1: This change works as expected.

PR #218, opened 2016-01-08
Simplify and fix notification grouping.

fabxc:
This commit changes the notification grouping behavior
to simply send all alerts of a group as soon as a single
one of them needs updating.

This fixes a critical bug which caused erroneous resolved
notifications to be sent.

@brian-brazil


brian-brazil:
:+1:

beorn7:
:+1:

PR #219, opened 2016-01-09
Add basic instrumentation

fabxc:
@brian-brazil

### DIFF
--- a/api.go
+++ b/api.go
@@ -32,6 +32,25 @@ import (
 	"github.com/prometheus/alertmanager/version"
 )
 
+var (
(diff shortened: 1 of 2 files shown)


api.go
```diff
@@ -31,6 +32,25 @@ import (
 	"github.com/prometheus/alertmanager/version"
 )
 
+var (
+	numReceivedAlerts = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "alertmanager",
+		Name:      "received_alerts_total",
```
brian-brazil: I'd do alerts_received rather than received alerts, so the link between the two metrics is more obvious.

brian-brazil:
:+1:

PR #221, opened 2016-01-09
Add instrumentation about notifications

fabxc:
@brian-brazil

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -28,6 +28,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/log"
 	"github.com/prometheus/common/model"
 	"golang.org/x/net/context"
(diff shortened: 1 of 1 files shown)


notify/impl.go
```diff
@@ -38,6 +39,25 @@ import (
...
+	numNotifications = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "alertmanager",
+		Name:      "notifications_total",
+		Help:      "The total number of attempted notifications.",
+	}, []string{"receiver", "integration"})
+
+	numFailedNotifications = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "alertmanager",
+		Name:      "notifications_failed_total",
+		Help:      "The total number failed notifications.",
```
brian-brazil: of

notify/impl.go
```diff
@@ -68,14 +93,27 @@ func Build(confs []*config.Receiver, tmpl *template.Template) map[string]Fanout
 			if len(res) == 0 {
 				return nil
 			}
-			return n.Notify(ctx, res...)
+
+			var (
+				integr = n.name()
+				// The receiver is of the form <name>/<integration>/<num>.
+				// We only care about the first component.
+				recv = strings.Split(receiver(ctx), "/")[0]
```
brian-brazil: Are names limited from having slashes?

notify/impl.go
```diff
@@ -38,6 +39,25 @@ import (
 	"github.com/prometheus/alertmanager/types"
 )
 
+var (
+	numNotifications = prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: "alertmanager",
+		Name:      "notifications_total",
+		Help:      "The total number of attempted notifications.",
+	}, []string{"receiver", "integration"})
```
brian-brazil: I can imagine the number of receivers getting really high (thousands) in large companies, I wonder if we should just split out by integration and leave the rest to logs.

notify/impl.go
fabxc: Currently not. It's sort of dirty in general restoring the info from a string that should be opaque – I might have to think about some tweaks elsewhere to do this better.

notify/impl.go
fabxc: Made it so.

brian-brazil:
:+1:

PR #223, opened 2016-01-09
New release process using docker, circleci and a centralized building tool

sdurrheimer:
Here is my current work on replacing the heavy `golang` based image by a `From scratch` or `busybox` one.

The binary is build by a building image [`sdurrheimer/prom-golang-builder`](https://github.com/sdurrheimer/prom-golang-builder) which will then launch the final docker image build.

This system can't use the automated build system from the Docker Hub. It can't ~~build a image which will in turn build a docker image~~ build golang binaries and then build the final docker image.
This is why CircleCI is needed. It will cross-build the project, build the project docker image and push it to the Docker Hub according to different cases:
- The `master` and `dev` branches
- Pre-release tags, for example: 1.0.0-beta
- Release tags, for example: 1.0.0. Those tags are also used for the `latest` docker tag.

The PR files are currently referring to my fork for the development purpose. I will rename those references before the merge.

Link to my alertmanager docker image: https://hub.docker.com/r/sdurrheimer/alertmanager/
Link to CircleCI builds: https://circleci.com/gh/sdurrheimer/alertmanager

In bonus, by using CircleCI with the Github repo, PRs will also be builded but not pushed to the docker hub.

There will be some Docker Hub and CircleCi configurations  to do. I can help with that if you want me to.

@fabxc @discordianfish @juliusv


fabxc:
This is fantastic!

Just to clarify how this whole thing works:
The `prom-golang-builder` image will work with the only requirement of a `make build` target to exist.
The repository is mounted into the build container in which then the scratch-image is built by talking to the Docker of the host system?

Using CI for that makes perfect sense.

sdurrheimer:
> Just to clarify how this whole thing works:
> The prom-golang-builder image will work with the only requirement of a make build target to exist.
Yes, a `make build` target and a `Dockerfile` into the root of the sources files. The project source files are mounted with a volume to `/app`. The `docker.sock` is also required in order to build the final docker image.

> The repository is mounted into the build container in which then the scratch-image is built by talking > to the Docker of the host system?
Yes, the `/var/run/docker.sock` in mounted into the build container in order to be able to talk to the host docker engine. This a better approach compared to Docker-in-Docker.

Moreover, I can improve the `prom-golang-builder` to be able to build the project without necessary building a docker image at the end. This will let people build the project without having a working golang environment.
And I guess we can also add the possibility to cross-compile the project for different platforms.

fabxc:
That all sounds great so far.

I discussed the approach with @matthiasr and he brought up that we should use `busybox` rather than `scratch` to allow for debugging, which makes a lot of sense to me. Thoughts?

sdurrheimer:
@fabxc @matthiasr Do you guys have some debugging scenario as examples ? Examples that can't be done with a sidekick container or some debugging tool like Sysdig ?
There is some things I don't like about `busybox` like non-normal GNU versions of UNIX tools for example.

Makefile
```diff
@@ -37,8 +41,17 @@ build:
 	@echo ">> building binaries"
 	@./scripts/build.sh
 
-docker:
-	@docker build -t alertmanager:$(shell git rev-parse --short HEAD) .
+docker: dockerpull
+	@echo ">> running builder docker image"
+	@docker run -t -v "$(shell pwd):/app" -v /var/run/docker.sock:/run/docker.sock:ro $(DOCKER_BUILDER_IMAGE) -p "github.com/prometheus/alertmanager" -t "$(DOCKER_IMAGE_NAME):$(DOCKER_IMAGE_TAG)"
```
discordianfish: Maybe do a `docker run --rm` to not leave old build containers around.

discordianfish:
@sdurrheimer:
That's pretty awesome, finally someone gets to do this :)
One thing though, I would really prefer having the prom-golang-builder repo in the prometheus org. I think it might seem confusing / less trustworthy to users especially since it won't be a automated build anymore and people need to trust the content of the images.

I think it should be okay to add you to the prometheus org and give you permissions to prom-golang-builder repo. @fabxc @beorn7 @juliusv - What do you think?

matthiasr:
I strongly object to calling GNU versions of tools "normal".

The main use case for a shell is to be able to `docker exec` and more importantly, `kubectl exec` into the container and poke around. In cluster deployments like Kubernetes, you generally don't get and should not need a shell on the node. A sidekick container can't be spun up on demand, and it won't have quite the same view of the world.

Makefile
sdurrheimer: I have tried, but it doesn't work on CircleCi. I had to remove it.

matthiasr:
I can't follow the control flow of this. Why does the builder need access to Docker at all? Why not just have a builder image that has the blessed Go version in it, and explicitly call out to that in the build? I.e. just use docker 1. as a way to pull in a defined build environment and 2. to package and ship the result but keep the two completely separate?

It feels to me like the current approach carries a lot of baggage from trying to `docker build` in one go.

fabxc:
@discordianfish I believe this was mostly for development. I will take care of the permission stuff.

@sdurrheimer To me the most important part is separation of a build and a run container. Name resolution is handled differently and there are some indications that the glibc resolver might be favorable.
So forcing a static binary under any circumstances might not be what we want after all. And it will not be possible for all tools we want to release in the future anyway (e.g. exporters written in Python).

I also just learned that the same build flags result in static and non-static builds depending on where one cross-compiles, which means our releases for Linux had different name resolution depending on who built them in the past.
So having a build container that cross compiles so we have reproducible release builds is really important.

sdurrheimer:
> One thing though, I would really prefer having the prom-golang-builder repo in the prometheus org. I think it might seem confusing / less trustworthy to users especially since it won't be a automated build anymore and people need to trust the content of the images.

@discordianfish Agreed, same as my previous docker images, I'm totally open to move it to the prometheus organization. But it might be cool to give me some permissions in order to be able to maintain it.

> I strongly object to calling GNU versions of tools "normal".

@matthiasr Sorry about using some bad shortcut words. What I wanted to say that some tools in busybox doesn't really act/work as most people might think. Like `wget` which doesn't work with HTTPS. In the past, there were also some tools missing some common options, I don't know if it is fixed now or not.

> The main use case for a shell is to be able to docker exec and more importantly, kubectl exec into the container and poke around. In cluster deployments like Kubernetes, you generally don't get and should not need a shell on the node. A sidekick container can't be spun up on demand, and it won't have quite the same view of the world.

I agree but I also like the fact to not to be able to executing other stuffs with from scratch. This might prevent privilege escalation with the root user. Yes of course we can use the `USER` directive in the Dockerfile, but I'm not sure that all prometheus tools can be run with a non-privilege user.

> I can't follow the control flow of this. Why does the builder need access to Docker at all? Why not just have a builder image that has the blessed Go version in it, and explicitly call out to that in the build? I.e. just use docker 1. as a way to pull in a defined build environment and 2. to package and ship the result but keep the two completely separate?
> It feels to me like the current approach carries a lot of baggage from trying to docker build in one go.

My objective was to replace the current heavy weight golang based image and try the `From scratch` approache that we already talked about on IRC with @juliusv and @fabxc etc.
So I was working on the Docker side of this. I pretty much like the way of building the golang project and building the docker image of it in one step. It makes it easier to reproduce this building mechanics for others projects (for the prometheus organization or for the prometheus community).
If this is problematic, we can work on breaking it into two steps. Maybe with a 'Dockerfile.build' for the building process and 'Dockerfile' as the final image.

> To me the most important part is separation of a build and a run container. Name resolution is handled differently and there are some indications that the glibc resolver might be favorable.
So forcing a static binary under any circumstances might not be what we want after all.

@fabxc Hum, you think the `From scratch` container won't work the same based on different distros ?

> And it will not be possible for all tools we want to release in the future anyway (e.g. exporters written in Python).

This is a specific way to build golang containers. For sure, we can't do the same with python or java exporters for example.

> I also just learned that the same build flags result in static and non-static builds depending on where one cross-compiles, which means our releases for Linux had different name resolution depending on who built them in the past.
So having a build container that cross compiles so we have reproducible release builds is really important.

You mean, you want to be able to use  the `prom-golang-builder`, not only for the docker image, but also for all the binaries included in the Github release ?

fabxc:
@sdurrheimer you should have sufficient permissions by now.

> You mean, you want to be able to use the prom-golang-builder, not only for the docker image, but also for all the binaries included in the Github release ?

Yes, using the same to cross-compile our releases would be neat.

matthiasr:
> This might prevent privilege escalation with the root user. 

Not really. If they crack open Prometheus they already _have_ code execution, there not being a shell won't save you. I would prefer debuggability over the illusion of security.

matthiasr:
Not sure if that's possible with CircleCI, but IMO the ideal process would be to use an image that has Go to run the tests and build binaries for all platforms, then put the binary for linux-amd64 in a Docker image, attach the binaries to the Github tag, and push to Docker hub… that way, the binary in Docker is also the same as on the release page.

discordianfish:
Best would be if Docker supports more shared namespaces (https://github.com/docker/docker/issues/10163). Until then.. I guess yeah especially with kubernetes and similar scenarios where you don't have host access I see how you want something like busybox for now.

juliusv:
Naive question from the sidelines: at Google, the cluster manager always mounted a simple base image (with shell, etc.) into every container, while the container image itself only contained the service binary. Couldn't people who want a shell do the same with Docker (using `-v` in read-only mode) or does the handling of that get too hard when using schedulers like Kubernetes on top of Docker?

No opinion of whether that's better, but it's an alternative to consider. I do like the idea of container images that only contain the service binary and anything it really needs to run.

matthiasr:
That is pretty much what "FROM busybox" does - that the base image is
called a "layer" instead of an image in its own right is a difference in
terminology.

I don't think either Docker or Kubernetes allow doing this after the fact.

sdurrheimer:
My question about busybox:
- I've seen some mention about glibc. Busybox use uClibc, does it fit well ? I don't know.
One tag seems to have glibc : `busybox:ubuntu-14.04` but it is a old tag.
Maybe busybox:ubuntu is a updated equivalent ? I don't know.
- There is no package manager, is it ok for you ? Will people complain because they can't do much except be able to use a shell and some available binaries ?

matthiasr:
Whether we need glibc depends more on the build- than the runtime container I guess. I'm not sure what it takes to really make a static go binary these days, that would be ideal. Otherwise you'll need to pull in whatever the binary needs, because the `busybox` image comes with the same libraries as `scratch`.

Yes, for debugging purposes I think busybox is enough. True that it doesn't have a package manager but that is more relevant for pulling in libraries at image build time.

sdurrheimer:
The need of glibc for the build-time is already the case in the `prom-golang-builder` and we are also already building static go binaries.
My question was more about @fabxc comment:

> Name resolution is handled differently and there are some indications that the glibc resolver might be favorable.
So forcing a static binary under any circumstances might not be what we want after all.

So if the binary might not  be necessarely static, we might then need a proper busybox with glibc, am I misunderstanding something here ?

matthiasr:
That was based on a verbal comment I made from being a bit miffed at the current (musl base + glibc prometheus) situation, which makes it hard to debug potential DNS issues because you can't reproduce what Prometheus sees when the base tools don't support search domains. This is not really an issue with the `busybox` image. While the resolvers used aren't _exactly_ the same it's still close enough so disregard that.

jimmidyson:
Seeing https://github.com/docker-library/busybox/tree/master/glibc seems like there's an up to date glibc busybox (1.24.1) but not published on Docker hub - at least can't find tag referencing it.

matthiasr:
let's not worry about the details of the base system too much. We don't want to _depend_ on it anyway, it's just there for operators. This means we can easily switch it out should the standard busybox prove unsuitable.

sdurrheimer:
I'm currently working on being able to cross-compile with the `prom-golang-builder` image (@see https://github.com/sdurrheimer/prom-golang-builder/tree/dev). The image will be huge with all these dependencies.

sdurrheimer:
@fabxc Cross-building Alertmanager for windows is currently failing due to the following error:
```
/go/src/github.com/prometheus/alertmanager/vendor/github.com/prometheus/common/log/syslog_formatter.go:15: undefined: syslog.Writer
```
After testing, it seems that it can be fixed by updating the `github.com/prometheus/common/log` vendor.

sdurrheimer:
Here some updates:
* I have split the`prom-golang-builder` image into 4 images:
  * The `main` image tag for platforms like `linux/amd64 linux/386`, `darwin/amd64 darwin/386`, `windows/amd64 windows/386` etc.
  * The `arm` image tag
  * The `powerpc` image tag
  * The `mips` image tag. This one is not used for now given that `go 1.50` doesn't support cross-building for `mips` architectures.
* I have remove the `docker-in-docker via docker socket` feature from the builder image. The project docker image is build after the crossbuild process.
* I have created a `prom-busybox` image that adds ca-certificates and zoneinfos.
* The CircleCI build is now cross-building the project, executing tests and then building the project docker image.

fabxc:
Thanks for the update. This is great!
Queued this for a proper look and to think about how we can automate our build/release process with that.

sdurrheimer:
I also need to check if we could also upload artifacts to the github release. Maybe by using [aktau/github-release](https://github.com/aktau/github-release)

sdurrheimer:
Here we go guys,

Here are the different scenarios:

##### 1/ Builds when branches are updated:
- `master` only by default. The `dev` branch is also builded for alertmanager but since it's now an outdated branch, should we keep it ? Those branches are configurable by modifying the `circle.yml` file.
- Steps:
  -  `make crossbuild`: cross-compiling the project using the `scripts/crossbuild.sh` script.
    - This script use 3 differents builder images. [`prom-golang-builder:main`](https://github.com/sdurrheimer/prom-golang-builder/tree/master/main), [`prom-golang-builder:arm`](https://github.com/sdurrheimer/prom-golang-builder/tree/master/arm) and the [`prom-golang-builder:powerpc`](https://github.com/sdurrheimer/prom-golang-builder/tree/master/powerpc)
  - `make docker`: building the project docker image with the branch name as docker tag
    - The project docker image use the [`prom-busybox`](https://github.com/sdurrheimer/prom-busybox) as base image. I have extended the official image to be able to add additional files like `ca-certificates.crt` for HTTPS, zoneinfos or any required files.
    - There are currently 2 tags:
      - `latest` for the classic `uclibc` version of `busybox`.
      - `glibc` for `busybox` with `glibc`. This one is used for projects that can't be statically builded like `node_exporter` for example.
  - Saving cross-builded binaries as CircleCI artifacts
  - `make dockertest`: Running tests in the [`prom-golang-builder:main`](https://github.com/sdurrheimer/prom-golang-builder/tree/master/main) image.
  - Pushing the branch's tagged image to the Docker Hub.

##### 2/ Pull Requests:
- Pull requests are builded like branches unlike that they are not pushed to the Docker Hub.

##### 3/ Github pre-releases:
- The tag version must respect the following RegEx : `/[0-9]+(\.[0-9]+)*-.*/`. For example: `1.2.0-rc1`
- Steps:
  -  `make crossbuild`: cross-compiling the project using the `scripts/crossbuild.sh` script.
    - This use 3 different builder images. [`prom-golang-builder:main`](https://github.com/sdurrheimer/prom-golang-builder/tree/master/main), [`prom-golang-builder:arm`](https://github.com/sdurrheimer/prom-golang-builder/tree/master/arm) and the [`prom-golang-builder:powerpc`](https://github.com/sdurrheimer/prom-golang-builder/tree/master/powerpc)
  - `make docker`: building the project docker image with the Github tag version as docker tag
    - The project docker image use the [`prom-busybox`](https://github.com/sdurrheimer/prom-busybox) as base image. I have extended the official image to be able to add additional files like `ca-certificates.crt` for HTTPS, zoneinfos or any required files.
    - There are currently 2 tags:
      - `latest` for the classic `uclibc` version of `busybox`.
      - `glibc` for `busybox` with `glibc`. This one is used for projects that can't be statically builded like `node_exporter` for example.
  - Saving cross-builded binaries as CircleCI artifacts
  - `make dockertest`: Running tests in the [`prom-golang-builder:main`](https://github.com/sdurrheimer/prom-golang-builder/tree/master/main) image.
  - `make release`: Creating tarballs for each cross-builded platforms with binaries, any config files, the `LICENSE` file and the `NOTICE` file. Then the tarballs are automatically uploaded to the Github pre-release.
  - Saving tarballs as CircleCI artifacts
  - Pushing the version's tagged image to the Docker Hub.

##### 3/ Github releases:
 - The tag version must respect the following RegEx : `/[0-9]+(\.[0-9]+)*/`. For example: `1.2.0`
 - Steps:
   -  `make crossbuild`: cross-compiling the project using the `scripts/crossbuild.sh` script.
     - This use 3 different builder images. [`prom-golang-builder:main`](https://github.com/sdurrheimer/prom-golang-builder/tree/master/main), [`prom-golang-builder:arm`](https://github.com/sdurrheimer/prom-golang-builder/tree/master/arm) and the [`prom-golang-builder:powerpc`](https://github.com/sdurrheimer/prom-golang-builder/tree/master/powerpc)
   - `make docker`: building the project docker image with the Github tag version as docker tag
     - The project docker image use the [`prom-busybox`](https://github.com/sdurrheimer/prom-busybox) as base image. I have extended the official image to be able to add additional files like `ca-certificates.crt` for HTTPS, zoneinfos or any required files.
     - There are currently 2 tags:
       - `latest` for the classic `uclibc` version of `busybox`.
       - `glibc` for `busybox` with `glibc`. This one is used for projects that can't be statically builded like `node_exporter` for example.
   - Saving cross-builded binaries as CircleCI artifacts
   - `make dockertest`: Running tests in the [`prom-golang-builder:main`](https://github.com/sdurrheimer/prom-golang-builder/tree/master/main) image.
   - `make release`: Creating tarballs for each cross-builded platforms with binaries, any config files, the `LICENSE` file and the `NOTICE` file. Then the tarballs are automatically uploaded to the Github release.
   - Saving tarballs as CircleCI artifacts
   - Docker tag the version's tagged image as latest
   - Pushing the latest and version's tagged image to the Docker Hub.

@fabxc @discordianfish @juliusv @matthiasr @grobie

matthiasr:
This looks really good, I am impressed :+1:

fabxc:
This is fantastic work! Thanks a ton.
Looking into it now.

Makefile
```diff
@@ -14,6 +14,14 @@
 GO   := GO15VENDOREXPERIMENT=1 go
 pkgs  = $(shell $(GO) list ./... | grep -v /vendor/)
 
+
+REPO_PATH 	                := github.com/prometheus/alertmanager
+VERSION 	                ?= $(shell cat version/VERSION)
+DOCKER_BUILDER_IMAGE 		:= sdurrheimer/prom-golang-builder:main
+DOCKER_BUILDER_PLATFORMS	?= $(shell go env GOOS)/$(shell go env GOARCH)
```
sdurrheimer: Unused variable. I will remove it soon.

circle.yml
```diff
@@ -0,0 +1,55 @@
...
+
+test:
+  pre:
+    - ln -s alertmanager-linux-amd64 alertmanager
+  override:
+    - make dockertest
+
+deployment:
+  hub_branch:
+    branch: [master, dev]
```
sdurrheimer: `dev` branch seems to be outdated now. I guess we can remove it from here.

fabxc: Yes, that one is dead.

sdurrheimer:
Further updates:

- [x] [`prometheus/busybox`](https://github.com/prometheus/busybox) created and linked to a automated build on Docker Hub [`prom/busybox`](https://hub.docker.com/r/prom/busybox/).
- [x] [`prometheus/golang-builder`](https://github.com/prometheus/golang-builder) created, build by [CircleCI](https://circleci.com/gh/prometheus/golang-builder) and pushed to Docker Hub [`prom/golang-builder`](https://hub.docker.com/r/prom/golang-builder/)
- [x] PR updated to point to the new images.

Missing :

- [ ] Rename `sdurrheimer` mentions to `prometheus` when the PR is validated.
- [ ] Create a github access token for `github-release` uploads.
- [ ] Setting up `prometheus/alertmanager` CircleCI build.

sdurrheimer:
Ready to move on.

discordianfish:
@sdurrheimer is the circleci error expected? Beside that, let's move forward. What would be the next/remaining steps? Maybe some rollout plan would be a good idea.

sdurrheimer:
Hum, CircleCI build PRs with a branch name like `pull/223`. Docker tags with a `/` are not allowed.
I think I have to change the default `SUFFIX` for non-master branches builds, maybe the commit ID instead of the branche name.

sdurrheimer:
CircleCI build fixed by replacing `/` in branch name by `-`.

sdurrheimer:
@discordianfish for `alertmanager` no remaining step for the moment. The new release process is already merged for `prometheus`. Another PR is also ready to be merged for the `node_exporter`.

For other exporters, I'm gonna make PRs progressively. Priority suggestion are welcome.

fabxc:
👍 💯

PR #229, opened 2016-01-26
Update common/* package vendoring

fabxc:
This will enable ms precision for configuration. Unlikely to be relevant for the Alertmanager but configuration shouldn't be different from Prometheus.


brian-brazil:
:+1:

PR #230, opened 2016-01-26
Add routing validation

fabxc:
The root route in the configuration must not exclude any alerts,
i.e. not have any matchers.
All receivers referenced in the routing tree are validated to exist.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -207,7 +207,36 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		}
 		names[rcv.Name] = struct{}{}
 	}
+
+	// The root route must not have any matchers as it is the fallback node
+	// for all alerts.
+    if c.Route == nil {
+        return fmt.Errorf("No routes provided")
+    }
+	if len(c.Route.Match) > 0 || len(c.Route.MatchRE) > 0 {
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -207,9 +207,35 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		}
 		names[rcv.Name] = struct{}{}
 	}
+
+	// The root route must not have any matchers as it is the fallback node
+	// for all alerts.
+	if len(c.Route.Match) > 0 || len(c.Route.MatchRE) > 0 {
```
brian-brazil: c.Route could be null, we should check for that first or change it not to be a pointer

fabxc:
Fixed

config/config.go
```diff
@@ -207,9 +207,38 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		}
 		names[rcv.Name] = struct{}{}
 	}
+
+	// The root route must not have any matchers as it is the fallback node
+	// for all alerts.
+    if c.Route == nil {
```
brian-brazil: You need a go fmt

brian-brazil:
:+1:

PR #237, opened 2016-02-02
fix build command

cherti:
`make` alone doesn't work to build alertmanager, throws a pile of errors instead here. `make build` works perfectly fine, though.

@fabxc

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ $ mkdir -p $GOPATH/src/github.com/prometheus
 $ cd $GOPATH/src/github.com/prometheus
 $ git clone https://github.com/prometheus/alertmanager.git
 $ cd alertmanager
-$ make
+$ make build
 $ ./alertmanager -config.file=<your_file>
 ```
 


fabxc:
`make build` is indeed sufficient for the end-user.

However,

> pile of errors 

specifics would really help here since changing the README is certainly not a fix for that.
Probably just some integration tests that are flaking due to a slow machine, but we should make sure.

cherti:
git cloned to /home/me/alertmanager, `cd alertmanager`, then `make`

```
can't load package: package _/home/me/alertmanager: cannot find package "_/home/me/alertmanager" in any of:
	/usr/lib/go/src/_/home/me/alertmanager (from $GOROOT)
	/home/me/.go/src/_/home/me/alertmanager (from $GOPATH)
[… more erros of the same kind …]
Makefile:29: die Regel für Ziel „format“ scheiterte
make: *** [format] Fehler 1
```

Basically make/go doesn't find a lot of packages with `make` only while `make build` doesn't have this issue.

Any further info I can help with, @fabxc?

Packages in question:
alertmanager
alertmanager/config
alertmanager/notify
am/provider
am/provider/sqlite
am/template
alertmanager/template/internal/deftmpl
alertmanager/test
alertmanager/test/acceptance
am/types
am/ui
am/version

juliusv:
It cannot find that package because `make` (but also `make build`) expects the `alertmanager` checkout directory to be in the right place in your `$GOPATH`, that is, in  `$GOPATH/src/github.com/prometheus/alertmanager`.

I wonder why `make build` works for you then.

cherti:
just tested again, `make build` reproducably works while only `make` produces mentioned errors.

fabxc:
That's probably be because the build script provides the package path directly rather than relative from the current directory as the `format` target does.
Is it possible that you have another copy of Alertmanager properly checked out in your `$GOPATH`? That's probably the one that's build rather than the one in the directory you are calling `make build` from.

Having a repository properly located in your `GOPATH` is a fundamental requirement for the `go` tool to work without issues. Our README instructions are very clear about that as well.
The behavior here is all correct and there's nothing to be done from our side.

fabxc:
Merged because `make build` is indeed sufficient for the end-user.

cherti:
> Is it possible that you have another copy of Alertmanager properly checked out in your $GOPATH? Then that's the one that's build rather than the one in the directory you are calling make build from.

Yeps, that's very precisely the case. `make build` fails with basically "cannot find package alertmanager" after I deleted that sourcedir.

Thx for clearing that up, @fabxc!

PR #238, opened 2016-02-02
OpsGenie: include missing source option

lukaf:
OpsGenie receiver has a source option:
```
# A backlink to the sender of the notification.
[ source: <tmpl_string> | default = '{{ template "opsgenie.default.source" . }}' ]
```
Which sets a field in OpsGenie alarm definition. `OpsGenieConfig` struct is correctly populated but the field was missing when creating a new message.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -614,6 +614,7 @@ type opsGenieCreateMessage struct {
 	*opsGenieMessage `json:,inline`
 	Message          string            `json:"message"`
 	Details          map[string]string `json:"details"`
+	Source           string            `json:"source"`
 }
 
 type opsGenieCloseMessage struct {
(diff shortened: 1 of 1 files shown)


notify/impl.go
```diff
@@ -658,6 +659,7 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) error {
 			opsGenieMessage: &apiMsg,
 			Message:         tmpl(n.conf.Description),
 			Details:         details,
+			Source:          tmpl(n.conf.Source),
```
brian-brazil: We should probably be defaulting this to ExternalURL

lukaf: @brian-brazil it already is: https://github.com/prometheus/alertmanager/blob/master/config/notifiers.go#L88

Which is expanded to:
https://github.com/prometheus/alertmanager/blob/master/template/default.tmpl#L34
https://github.com/prometheus/alertmanager/blob/master/template/default.tmpl#L2

brian-brazil:
Ah missed that, thanks!

PR #239, opened 2016-02-03
Fix disappearing alerts

### LINKED ISSUE #231: Continuously firing alerts disappearing from alertmanager UI
Both with the current master and with 0.1.0-beta1, continuously firing alerts first show up in the AM interface and then disappear after a couple of minutes when I use the following Prometheus and Alertmanager configurations:

https://gist.github.com/juliusv/14787b894356ac189023

Notably, when I restart Prometheus, the alerts reappear for a while and then disappear again. So maybe it has something to do with their starting timestamp?

fabxc:
Fixes #231 
This only happened with the changed Integration in the upcoming Prometheus version. So with Prometheus clients 0.16.2 and earlier it didn't cause issues.

@juliusv can you verify?

### DIFF
--- a/api.go
+++ b/api.go
@@ -235,7 +235,7 @@ func (api *API) insertAlerts(w http.ResponseWriter, r *http.Request, alerts ...*
 		// is marked resolved if it is not updated.
 		if alert.EndsAt.IsZero() {
 			alert.Timeout = true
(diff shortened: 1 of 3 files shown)


juliusv:
I'm not able to recreate the problem anymore with this change at least :) Thank you!

juliusv:
:+1:

Issue #207, opened 2015-12-30
Slack messages can't send urls as markup


dmclain:
When I create a configuration that tries to do formatting of the `text` field like so:

    slack_configs:
        - text: "{{ range .Alerts }}*{{ .Annotations.description }}*<{{ .Annotations.generatorURL }}|View in Prometheus>\n{{ end }}"
          ...

The json that is sent to sentry escapes the `<` and `>` characters.

    "text":"*my alert's description*\u0026lt;http://prometheus:9090/graph#%5B%7B%22expr%22%3A%22node_filesystem_avail%20%2F%20node_filesystem_size%20%3C%200.5%20and%20node_filesystem_avail%20%2F%20-delta%28node_filesystem_avail%5B1d%5D%29%20%3E%200%20%3C%207%22%2C%22tab%22%3A0%7D%5D|View in Prometheus\u003e\n"

I believe this is because the json encoder object in the standard library makes the decision for the programmer to always escape those characters. (see https://golang.org/pkg/encoding/json/#Marshal and https://github.com/golang/go/issues/8592) 

For the slack notifier to support sending links as anything other than the full url (which would be useful for sending generatorURL) either a different method of converting to json or a second pass that unescapes those values will be necessary.

fabxc:
It's really a shame that the `encoding/json` enforces this with no option to turn it off.
Seems like a post-processing step on the raw JSON that undos it is the only way – not really pretty.

brian-brazil:
Have you filed a bug with Slack about this? It's likely that all other Go slack clients are affected, so they will probably want to fix this bug on their end.

tisinno:
I think the problem is that the text field is not allowing markup. After adding the text field to this line:
		MrkdwnIn:  []string{"fallback", "pretext", "text"},
in notify/impl.go I can send markup and urls

brian-brazil:
This still sounds like a bug in Slack if it can't handle escaped characters in json strings.

tisinno:
the doc is a bit ambiguous to me but I think escaping < > & is actually required for slack api, see https://api.slack.com/docs/formatting. Here it seems to have used some unicode escaping as well, but apparently that doesn't matter, after allowing markup for the "text" field using the following in a template shows up fine in slack client: <http://github.com|TEST URL>

fabxc:
So #240 will essentially fix this without post-processing the JSON?

tisinno:
[...]

PR #244, opened 2016-02-08
Slack: Use the text template processor for the text field

dmclain:
In #207, I speculated that the golang json encoder was overescaping the values sent to slack, but slack's API handles the escaped json fine. The real culprit is the html template call that is being used for text which is turning `<` into `&lt;`.

Using alertmanager with this patch, I can use the following configuration:
```
receivers:
- name: 'debug-slack'
  slack_configs:
  - api_url: "https://hooks.slack.com/services/crypto/crypto"
    username: "{{ .GroupLabels.instance }}"
    channel: "#prometheus-debug"
    text: "{{ range .Alerts }}*{{ .Annotations.description }}* <{{ .Annotations.generatorURL }}|View in Prometheus>\n{{ end }}"
```

and then running:
```
curl -XPOST http://localhost:9093/api/v1/alerts -d '[{"labels": {
        "alertname": "DMcLain Testing",
        "severity": "warning",
        "instance": "qa-dmclain-flaming-bourbon"
    },
    "annotations": {
        "summary": "High latency is high!",
        "description": "qa-dmclain-flaming-bourbon has 0.02316945256093281 days of disk left on rootfs mounted at /",
        "generatorURL": "http://infra-prometheus-happy-pbr:9090/graph#%5B%7B%22expr%22%3A%22node_filesystem_avail%20%2F%20node_filesystem_size%20%3C%200.5%20and%20node_filesystem_avail%20%2F%20-delta%28node_filesystem_avail%5B1d%5D%29%20%3E%200%20%3C%207%22%2C%22tab%22%3A0%7D%5D"
    }
}]'
```

I got a message like the following in Slack:
slack

With the "View In Prometheus" linking to the generatorURL as hoped/expected.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -481,14 +481,13 @@ func (n *Slack) Notify(ctx context.Context, as ...*types.Alert) error {
 	var (
 		data     = n.tmpl.Data(receiver(ctx), groupLabels(ctx), as...)
 		tmplText = tmplText(n.tmpl, data, &err)
-		tmplHTML = tmplHTML(n.tmpl, data, &err)
 	)
 
 	attachment := &slackAttachment{
 		Title:     tmplText(n.conf.Title),
 		TitleLink: tmplText(n.conf.TitleLink),
 		Pretext:   tmplText(n.conf.Pretext),
-		Text:      tmplHTML(n.conf.Text),
+		Text:      tmplText(n.conf.Text),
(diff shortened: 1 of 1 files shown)


brian-brazil:
Thanks!

grobie:
We should add tests for such cases. It'll just break again if we don't codify the expectations.

PR #249, opened 2016-02-10
Upgrade webhook protocol

fabxc:
We basically want the same info for webhook integrations as for templating.
This will also make maintenance easier since we don't have two separate formats we have to version.

@brian-brazil

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -200,29 +200,30 @@ func (kv KV) Values() []string {
 	return kv.SortedPairs().Values()
 }
 
-// Data is the data passed to notification templates.
-// End-users should not be exposed to Go's type system,
-// as this will confuse them and prevent simple things like
-// simple equality checks to fail. Map everything to float64/string.
+// Data is the data passed to notification templates and webhook pushes.
+//
+// End-users should not be exposed to Go's type system, as this will confuse them and prevent
+// simple things like simple equality checks to fail. Map everything to float64/string.
 type Data struct {
-	Receiver string
-	Status   string
-	Alerts   Alerts
+	Receiver string `json:"receiver"`
+	Status   string `json:"status"`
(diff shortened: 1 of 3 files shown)


beorn7:
Note: tests are failing.

template/template.go
```diff
@@ -205,24 +205,25 @@ func (kv KV) Values() []string {
...
 type Alert struct {
-	Status       string
-	Labels       KV
-	Annotations  KV
-	WasSilenced  bool
-	WasInhibited bool
+	Status       string `json:"status"`
+	Labels       KV     `json:"labels"`
+	Annotations  KV     `json:"annotations"`
+	WasSilenced  bool   `json:"-`
```
brian-brazil: Missing a " ?

template/template.go
```diff
@@ -205,24 +205,25 @@ func (kv KV) Values() []string {
 // as this will confuse them and prevent simple things like
 // simple equality checks to fail. Map everything to float64/string.
 type Data struct {
```
brian-brazil: We should add a note here that this is also going to the webhook. Sending things over the network is more expensive than rendering them in memory.

fabxc: The whole point of AM is that there are only very few notifications. So I'm not worried about any network load.
Adding comment anyway.

fabxc:
Updated

brian-brazil:
:+1:

template/template.go
brian-brazil: "Few" is a relative term. It's not hard to imagine a situation where you'd have thousands of alerts per second worth of webhook firing.

fabxc: Honestly, we are worrying about 1000x scale way to often for Alertmanager at this point.
And if we reach that point, 200 more bytes of JSON will certainly not be the bottleneck.

PR #252, opened 2016-02-15
Fix vetting errors

fabxc:
@brian-brazil

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -607,14 +607,15 @@ type opsGenieMessage struct {
 }
 
 type opsGenieCreateMessage struct {
-	*opsGenieMessage `json:,inline`
-	Message          string            `json:"message"`
-	Details          map[string]string `json:"details"`
-	Source           string            `json:"source"`
+	*opsGenieMessage `json:",inline"`
+
+	Message string            `json:"message"`
+	Details map[string]string `json:"details"`
+	Source  string            `json:"source"`
 }
 
 type opsGenieCloseMessage struct {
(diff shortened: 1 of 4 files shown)


notify/impl.go
```diff
@@ -549,7 +549,7 @@ func (n *Hipchat) Notify(ctx context.Context, as ...*types.Alert) error {
 		data     = n.tmpl.Data(receiver(ctx), groupLabels(ctx), as...)
 		tmplText = tmplText(n.tmpl, data, &err)
 		tmplHTML = tmplHTML(n.tmpl, data, &err)
-		url      = fmt.Sprintf("%sv2/room/%s/notification?auth_token=%s", n.conf.APIURL, n.conf.RoomID, n.conf.AuthToken)
+		url      = fmt.Sprintf("%sv2/room/%d/notification?auth_token=%s", n.conf.APIURL, n.conf.RoomID, n.conf.AuthToken)
```
brian-brazil: RoomID is a string, %s sounds right to me

fabxc: Mh, that's odd. My editor show's it as type int in auto-completion and `go vet` did this:

```>> vetting code
notify/impl.go:552: arg n.conf.RoomID for printf verb %s of wrong type: int
```

brian-brazil: It was an int in the previous alertmanager, do you have a cache or the old version lying around or something?

fabxc: Nope, checked everything. Weird.
Changed it back.

PR #253, opened 2016-02-16
Adding support for slack icon_emoji in slack receiver config.

alexissavin:
As it may be useful to quickly identify notifications in a channel stream, we needed support for "icon_emoji" in the slack receiver configuration.

Configuration look like this :
```
slack_configs:
  - send_resolved: true
    username: 'Prometheus'
    channel: '#monitoring'
    icon_emoji: ':prometheus:'
    api_url: 'https://hooks.slack.com/services/<token>'
```

@fabxc, is this small addition ok for you ? Thanks in advance for your feedback.


brian-brazil:
Thanks!

fabxc:
Merged too early. Template changes need a manual run of `make assets`.
(It's not run by the default target since it rewrites timestamps aggressively every time)

PR #256, opened 2016-02-19
Make source link in emails more prominent

grobie:
The huge improvement of Prometheus alerting is that it provides me with
a history of the alert expression over time. One of my most common
actions after receiving an alert is to gather more information in the
prometheus graph explorer.

@fabxc 

I don't know how to test this? What do you usually do?

### DIFF
--- a/template/email.html
+++ b/template/email.html
@@ -325,7 +325,7 @@ a {
           <tr>
             <td class="alert alert-warning">
               {{ .Alerts | len }} alert{{ if gt (len .Alerts) 1 }}s{{ end }} for {{ range .GroupLabels.SortedPairs }}
-                {{ .Name }}={{ .Value }} 
+                {{ .Name }}={{ .Value }}
               {{ end }}
             </td>
           </tr>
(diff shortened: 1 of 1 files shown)


fabxc:
I increment the counter on my backlog for integration tests by 1 and keep moving :)

Unfortunately this is based on the assumption that the alerts were grouped by alertname, are all from the same Prometheus server, and all originate from the same alerting expression.

This is not generally true. While it will common for most new/small setups, all these aspects will change over time in more advanced environments.

Thus, `GeneratorURL` is an attribute of each alert and not of the entire notification and this template would not evaluate.

fabxc:
What is certainly helpful is common information in the beginning of a message.
We are doing that in a few places in our internal templates – the `CommonAnnotations` and `CommonAnnotations` attributes are handy for that. 

This should really be used in email templates as well.

In the same vein one could also display the `GeneratorURL` at the top if it is equal for all alerts in the notification. It might just be confusing if people expect it to be there in every notification. A short message "alerts have different sources" would help there.

grobie:
You're right of course, I wonder how more often I'll run into this. I'll close for now.

> I increment the counter on my backlog for acceptance tests by 1 and keep moving :)

I loled. Then I shed some tears.

PR #257, opened 2016-02-23
Changed incident_key

andriisoldatenko:
I need to change `incident_key` in pagerduty post request to `alertname`? 
Could you please help me.


fabxc:
What do you want do?
I assume you meant to change the incident key to be the literal string "alertname". That would simply cause all pages to be the same PagerDuty incident.

andriisoldatenko:
@fabxc It'll be great if `incident_key` would be not random integer number. For example description or alertname?

fabxc:
Can you elaborate why you want that?
Descriptions, alertname, etc. are in respective fields. For example the PagerDuty description and the data body which contains information about the alerts in the notification.

`incident_key` is a field that's meant for machine consumption, i.e. PagerDuty itself.
The `incident_key` is what gives us identity for a notification, so that we can update incidents as the alerts for it change. It's not a random number.

bratchenko:
Hi Fabian,

the reason for this is that incident key is the only distinguishing field of pager duty incident that is returned through [Incidents API](https://developer.pagerduty.com/documentation/rest/incidents).

We want to consume list of incidents elsewhere and understand what kind of incident it is. Therefore we want to set "incident key" to a value that our automation would be able to interpret.

This change does not need to go to master. As I understand, Andrii created this PR to discuss the right approach of doing it in our fork (or maybe add ability to specify incident key as a feature to prometheus).

brian-brazil:
This seems like a failing of the Pagerduty API that should be taken up with them. One option on our side is to use a webhook to capture notifications and do whatever processing you'd like with that. 

I don't think this is something we should make templatable, as our current behaviour meets the contract required by Pagerduty and there's a pretty good chance that if this was exposed to users that they'd accidentally end up breaking their paging setup (for example this change would cause every alert to cause it's own page, ignoring the user's requested grouping).

juliusv:
@brian-brazil The current PD incident key is a combination of the hash of the alert grouping labels (and a not yet used routing config hash, it seems). Since the incident key can be any string in PD, what about the option of just spelling out the grouping labels like a time series, i.e. `{alertname="foo",job="api-server"}` instead of hashing them together?

brian-brazil:
We could do that, but it'd have to be clear that we may change the format at any time based on technical needs.

juliusv:
Yeah, just an idea.

fabxc:
Not really an option. Generating unique incident keys is not trivial and especially problematic across config changes. The route fingerprint should be used of course, that's an issue/bug.
Right now same groupings from completely different routes will result in the same incident.

There is no meaningful short way to express an incident in a human-readable format. Especially not if the `routeFP` is used correctly.

That said, I don't find this to be a satisfying solution nonetheless. It should a) probably not use fnv to be less prone to collision (collision tables are not really an option) b) for full correctness a hash would not only have to include what did match along the way through the routing tree, but also what did not match based on the `continue` flags of previous routes.

fabxc:
Regarding this PR: the PD API docs pretty clearly state that the `incident_key` is a meant to be used for deduplication. There are plenty of other fields that can contain data for human consumption or automation.

> the reason for this is that incident key is the only distinguishing field of pager duty incident that is returned through Incidents API.

Any changes to what it is now will take the "distinguishing" aspect away entirely.
If you want to maintain an internal fork, you could suffix the number we set with something – but you would have to be extremely careful that this suffix wouldn't change over time while the incident_key remains the same.

How many incidents do you have? Assuming it's not more than hundreds/second, you can retrieve a list and then get the details for each incident. There you can put basically any data in any format you want via the templating we provide.

andriisoldatenko:
@fabxc Thanks. I'll close the PR. We'll use description

Issue #107, opened 2015-08-01
Pushover: Resolved alerts are sent as emergency alerts


Luzifer:
This is a follow-up from #90:

> Well, now that you made me look at it again, I wonder whether "resolved" notifications should not be emergencies :) (@juliusv)

Confirmed, this totally should be the case! Shocked me a little bit to receive three new emergency notifications after the first bunch… xD

![2015-08-01-22 45 39](https://cloud.githubusercontent.com/assets/20142/9023876/6c3a52e6-38b0-11e5-8403-7e1f6d9e68e3.png)

juliusv:
@Luzifer Wanna send a fix? :)

PR #258, opened 2016-02-26
Re-implement pushover notifications

stapelberg:
This feature was dropped during the rewrite, but I use and like
Pushover.

The resulting notification looks like this:
screenshot_20160226-093434

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -16,6 +16,7 @@ package config
 import (
 	"fmt"
 	"strings"
+	"time"
 )
 
 var (
@@ -88,6 +89,19 @@ var (
 		Description: `{{ template "opsgenie.default.description" . }}`,
 		Source:      `{{ template "opsgenie.default.source" . }}`,
 		// TODO: Add a details field with all the alerts.
+	}
+
+	// DefaultPushoverConfig defines default values for Pushover configurations.
+	DefaultPushoverConfig = PushoverConfig{
+		NotifierConfig: NotifierConfig{
+			VSendResolved: true,
+		},
+		Title:    `{{ template "pushover.default.title" . }}`,
+		Message:  `{{ template "pushover.default.message" . }}`,
+		URL:      `{{ template "pushover.default.url" . }}`,
+		Priority: `{{ if eq .Status "firing" }}2{{ else }}0{{ end }}`, // emergency (firing) or normal
+		Retry:    duration(1 * time.Minute),
+		Expire:   duration(1 * time.Hour),
 	}
 )
 
@@ -282,4 +296,50 @@ func (c *OpsGenieConfig) UnmarshalYAML(unmarshal func(interface{}) error) error 
 		return fmt.Errorf("missing API key in OpsGenie config")
 	}
 	return checkOverflow(c.XXX, "opsgenie config")
+}
+
+type duration time.Duration
+
+func (d *duration) UnmarshalText(text []byte) error {
+	parsed, err := time.ParseDuration(string(text))
+	if err == nil {
+		*d = duration(parsed)
+	}
+	return err
+}
+
+func (d duration) MarshalText() ([]byte, error) {
+	return []byte(time.Duration(d).String()), nil
+}
+
+type PushoverConfig struct {
+	NotifierConfig `yaml:",inline"`
+
+	UserKey  Secret   `yaml:"user_key"`
+	Token    Secret   `yaml:"token"`
+	Title    string   `yaml:"title"`
+	Message  string   `yaml:"message"`
+	URL      string   `yaml:"url"`
+	Priority string   `yaml:"priority"`
+	Retry    duration `yaml:"retry"`
+	Expire   duration `yaml:"expire"`
+
+	// Catches all undefined fields and must be empty after parsing.
+	XXX map[string]interface{} `yaml:",inline"`
+}
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface.
+func (c *PushoverConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	*c = DefaultPushoverConfig
+	type plain PushoverConfig
+	if err := unmarshal((*plain)(c)); err != nil {
+		return err
+	}
+	if c.UserKey == "" {
+		return fmt.Errorf("missing user key in Pushover config")
+	}
+	if c.Token == "" {
+		return fmt.Errorf("missing token in Pushover config")
+	}
+	return checkOverflow(c.XXX, "pushover config")
 }
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -19,11 +19,13 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"mime"
 	"net"
 	"net/http"
 	"net/mail"
 	"net/smtp"
+	"net/url"
 	"os"
 	"strings"
 	"time"
@@ -132,6 +134,10 @@ func Build(confs []*config.Receiver, tmpl *template.Template) map[string]Fanout 
 		}
 		for i, c := range nc.HipchatConfigs {
 			n := NewHipchat(c, tmpl)
+			add(i, n, filter(n, c))
+		}
+		for i, c := range nc.PushoverConfigs {
+			n := NewPushover(c, tmpl)
 			add(i, n, filter(n, c))
 		}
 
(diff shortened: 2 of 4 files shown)


notify/impl.go
```diff
@@ -682,6 +688,72 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) error {
...
+	}
+	data := n.tmpl.Data(receiver(ctx), groupLabels(ctx), as...)
+
+	log.With("incident", key).Debugln("notifying Pushover")
+
+	var err error
+	tmpl := tmplText(n.tmpl, data, &err)
+
+	parameters := url.Values{}
+	parameters.Add("token", n.conf.Token)
```
brian-brazil: Token and user should probably be templated

notify/impl.go
```diff
@@ -682,6 +688,72 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) error {
...
+	title := tmpl(n.conf.Title)
+	message := tmpl(n.conf.Message)
+	parameters.Add("title", title)
+	if len(title)+len(message) > 512 {
+		message = message[:512]
+		log.With("incident", key).Debugf("Truncated message to %q due to Pushover message limit", message)
+	}
+	parameters.Add("message", message)
+	parameters.Add("url", tmpl(n.conf.URL))
+	parameters.Add("priority", "2")  // emergency
```
brian-brazil: These should all be user configurable

notify/impl.go
```diff
@@ -682,6 +688,72 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) error {
...
+	var err error
+	tmpl := tmplText(n.tmpl, data, &err)
+
+	parameters := url.Values{}
+	parameters.Add("token", n.conf.Token)
+	parameters.Add("user", n.conf.UserKey)
+	title := tmpl(n.conf.Title)
+	message := tmpl(n.conf.Message)
+	parameters.Add("title", title)
+	if len(title)+len(message) > 512 {
```
brian-brazil: This is quite little space, so that listing all alerts doesn't make sense once you've more than 1 or 2 - and for the defaults we need to consider the case where there could be hundreds. Could you cut this down to say only the labels in the message?

notify/impl.go
brian-brazil: and be aware of #107

notify/impl.go
fabxc: I recommend using the `CommonLabels` and `CommonAnnotations` here plus a sane fallback should neither exist.

brian-brazil: Common labels are already up in the title by default.

fabxc: Ah right, was just skimming through on mobile.
We should probably have a default detail description putting `CommonAnnotations` (and possibly common `GeneratorURL`) at the top. It has come up quite often now.

brian-brazil: In this case the GeneratorURL is already covered in another field. It's unclear to me what we should do in cases where we've some space, but not enough to cover more than a handful of alerts. The common annotations and then start listing the alerts is best I guess. We may want to add some better logic for truncation than a simple cutoff, as there's quite a few fields across the various notifications methods that have limits.

notify/impl.go
stapelberg: Done.

notify/impl.go
stapelberg: Done.

notify/impl.go
stapelberg: I’m not sure what the result of this discussion is. Do I need to change anything right now, or can we just file an issue to revisit this later with a common solution? If you want me to make changes right now, can you outline what they should be please?

stapelberg:
Addressed your comments. PTAL.

fabxc:
We did the whole thing of accepting integrations without tests in the past for AM and service discovery in Prometheus. While it speeds up features and all, this of course did not turn out to great.

Full integration tests are just not realistic as I realize.
But I'd like to start requiring acceptance tests, i.e. at least a simple mocking of the API against which we implement.

Would you mind providing these? For AM integrations that's relatively simple.

config/notifiers.go
```diff
@@ -89,6 +89,20 @@ var (
...
+	// DefaultPushoverConfig defines default values for Pushover configurations.
+	DefaultPushoverConfig = PushoverConfig{
+		NotifierConfig: NotifierConfig{
+			VSendResolved: true,
+		},
+		Title:            `{{ template "pushover.default.title" . }}`,
+		Message:          `{{ template "pushover.default.message" . }}`,
+		URL:              `{{ template "pushover.default.url" . }}`,
+		Priority:         "2",  // emergency
+		PriorityResolved: "0",  // normal
```
brian-brazil: We've used templating to handle this elsewhere, rather than a separate option.

notify/impl.go
brian-brazil: Please change to listing the common annotations first, and then the listing of alerts.

config/notifiers.go
stapelberg: I’ve looked at the other notifiers and can’t find a use of templating that uses a conditional on whether the alerts are resolved or not.

Could you please point me to an example of what you’re thinking of?

brian-brazil: https://github.com/prometheus/alertmanager/blob/master/config/notifiers.go#L61

stapelberg: Done.

notify/impl.go
stapelberg: Done.

stapelberg:
> But I'd like to start requiring acceptance tests, i.e. at least a simple mocking of the API against which we implement.
> 
> Would you mind providing these? For AM integrations that's relatively simple.

I’m willing to add them if you add an example first. I don’t have time for doing significant alertmanager development work.

config/notifiers.go
```diff
@@ -283,3 +296,35 @@ func (c *OpsGenieConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
...
+type PushoverConfig struct {
+	NotifierConfig `yaml:",inline"`
+
+	UserKey  Secret `yaml:"user_key"`
+	Token    Secret `yaml:"token"`
+	Title    string `yaml:"title"`
+	Message  string `yaml:"message"`
+	URL      string `yaml:"url"`
+	Priority string `yaml:"priority"`
+	Retry    int    `yaml:"retry"`
```
brian-brazil: retry_seconds and expire_seconds would be clearer.

template/default.tmpl
```diff
@@ -162,3 +162,10 @@ SOFTWARE.
 </html>
 
 {{ end }}
+
+{{ define "pushover.default.title" }}{{ template "__subject" . }}{{ end }}
+{{ define "pushover.default.message" }}
```
brian-brazil: This will put a newline at the start of the message, is that what you want?

stapelberg: Done.

config/notifiers.go
stapelberg: Done.

stapelberg:
PTAL.

config/notifiers.go
```diff
@@ -283,3 +296,35 @@ func (c *OpsGenieConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
...
+type PushoverConfig struct {
+	NotifierConfig `yaml:",inline"`
+
+	UserKey  Secret `yaml:"user_key"`
+	Token    Secret `yaml:"token"`
+	Title    string `yaml:"title"`
+	Message  string `yaml:"message"`
+	URL      string `yaml:"url"`
+	Priority string `yaml:"priority"`
+	Retry    int    `yaml:"retry_seconds"`
```
grobie: I know you just changed that as asked by @brian-brazil, but why are we not using golangs `time.Duration` here? `retry: 1m`, `expire: 1h`, maye with an added `interval` suffix or so.

stapelberg: Done.

brian-brazil:
:+1:

stapelberg:
Not sure why the tests fail on travis, they do pass locally.

What’s missing until this PR can be merged? Who can make that call?

fabxc:
Travis seems to be really slow and runs into timing issues even though the tests are quite generous already.

I'd still prefer having acceptance tests but don't have time right now either.
Please don't be surprised in case something accidentally breaks in the future though that would have been caught by them.

:+1:

Issue #193, opened 2015-12-11
Error sending email notification: starttls failed: x509: certificate signed by unknown authority


gregbkr:
Hello,

How can I get pass the TLS error for sending email notification to a remote server?

I configured the var: 
SMTP_AUTH_USERNAME
SMTP_AUTH_PASSWORD
and command: "-config.file=/alertmanager.conf -notification.smtp.smarthost=mail.xxx.com:587  -notification.smtp.sender=support-testing@xxx.com"

I have this error:
    
    time="2015-12-11T16:06:48Z" level=error msg="Error sending email notification: starttls failed: x509: certificate signed by unknown authority" file=notifier.go line=758

I tried to use the prometheus config: tls_config :

Disable validation of the server certificate. [ insecure_skip_verify: ] http://prometheus.io/docs/operating/configuration/

File prometheus.yml:

    scrape_configs:  

      # Container via Cadvisor
      - job_name: 'container'

        target_groups:
          - targets: ['192.168.33.10:8888'] # cadvisor
            labels:
              group: 'dev'

        tls_config:
          insecure_skip_verify: true

I am not sure really where to put the tls_config parameters, anywhere is failing except here...

Any help welcome ;-)
Thank you! Good evening!
Greg

fabxc:
You are setting the TLS options for scraping the cAdvisvor metrics endpoint. The error is related to validating certifcates when sending email notifications for which you cannot configure skipping verification.
You have to setup your system in a way that your certificates can be verified.

gregbkr:
Thanks a lot Fabxc! So I can't use that option. 
Do you guys target a remote email server and validate your certificate, 
OR do you use a post fix null client with no certificate to relay the email?
Thank you for you help!

opskumu:
@grebkr maybe you need modify the source code `manager/notifier.go`:

```
@@ -572,7 +586,7 @@ func (n *notifier) sendEmailNotification(to string, op notificationOp, a *Alert)
                return fmt.Errorf("invalid address: %s", err)
        }
 
-       tlsConfig := &tls.Config{ServerName: host}
+       tlsConfig := &tls.Config{ServerName: host, InsecureSkipVerify: true}
        if err := c.StartTLS(tlsConfig); err != nil {
                return fmt.Errorf("starttls failed: %s", err)
        }
```
then, go build it.

opskumu:
@grebkr maybe you need modify the source code `manager/notifier.go`:

```
[...]

PR #163, opened 2015-11-28
smtp: STARTTLS before querying auth mechanisms

stapelberg:
Some servers (notably smtp.gmail.com) require you to use STARTTLS before
they will even list PLAIN authentication. This commit makes using gmail
for smtp notifications work.

Aside from STARTTLS being required for authentication, it’s of course
also a good idea to just use STARTTLS whenever possible, regardless of
the authentication requirements :).

PR #266, opened 2016-03-04
Re-do #163: smtp: STARTTLS before querying auth mechanisms

stapelberg:
This was not ported over with the rewrite.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -227,7 +227,7 @@ func NewEmail(c *config.EmailConfig, t *template.Template) *Email {
 func (*Email) name() string { return "email" }
 
 // auth resolves a string of authentication mechanisms.
-func (n *Email) auth(mechs string) (smtp.Auth, *tls.Config, error) {
+func (n *Email) auth(mechs string) (smtp.Auth, error) {
 	username := os.Getenv("SMTP_AUTH_USERNAME")
 
 	for _, mech := range strings.Split(mechs, " ") {
@@ -237,7 +237,7 @@ func (n *Email) auth(mechs string) (smtp.Auth, *tls.Config, error) {
 			if secret == "" {
 				continue
 			}
-			return smtp.CRAMMD5Auth(username, secret), nil, nil
+			return smtp.CRAMMD5Auth(username, secret), nil
 
 		case "PLAIN":
 			password := os.Getenv("SMTP_AUTH_PASSWORD")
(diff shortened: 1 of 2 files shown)


grobie:
How will this work if TLS is not available?

fabxc:
To clarify, this change did not carry over from the old AM because people reported issues with it. See #193.

stapelberg:
> How will this work if TLS is not available?

I’ve modified the pull request to check whether STARTTLS is announced by the server.

> To clarify, this change did not carry over from the old AM because people reported issues with it.

The next time people report issues with features that I contribute, I’d be happier if you could notify me and give me a chance to fix the issues instead of effectively reverting my contributions.

I think what #193 is asking for is an option for specifying InsecureSkipVerify in the TLS config, which is orthogonal to what I’m implementing here. Of course, I do realize that the change I’m making has the side-effect of exposing broken configurations that used to work before.

How do you suggest we proceed? Should #193 be fixed before this can go in, or can we fix #193 independently?

fabxc:
By the time of your contribution the rewrite was already in process and I only carried over very few changes – and none that might have been source of issues.
So it was no intentional revert of your changes but it drowned in a lot of other work – sorry about that.

I'll take a look at the details tomorrow.

notify/impl.go
```diff
@@ -270,16 +266,24 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) error {
 	}
 	defer c.Quit()
 
+	// We need to know the hostname for both auth and TLS.
+	host, _, err := net.SplitHostPort(n.conf.Smarthost)
+	if err != nil {
+		return fmt.Errorf("invalid address: %s", err)
+	}
+
+	if ok, _ := c.Extension("STARTTLS"); ok {
```
grobie: Doesn't this open a downgrade attack vector? An attacker could just strip the STARTTLS field from the extension list and trick alertmanager into sending messages unencrypted.

What do you think about using a flag instead, so that alertmanager would send emails either always with TLS (and failing if it can't) or always without?

stapelberg: Yeah, having a flag which defaults to always enforcing TLS sounds good to me.

fabxc:
We can handle the other issue independently. It should just be done before the next release.

notify/impl.go
fabxc: This sounds like something we should add within this PR then.

stapelberg:
Added the flag. PTAL.

grobie:
:+1:

juliusv:
:+1: Thanks, Herr @stapelberg :)

PR #269, opened 2016-03-08
Adjust slack notifier default config to match documentation

pdbogen:
Docs:
```
Slack receiver <slack_config>

Slack notifications are sent via Slack webhooks.

# Whether or not to notify about resolved alerts.
[ send_resolved: <boolean> | default = true ]
```

Implementation (https://github.com/prometheus/alertmanager/blob/1748a0e304a928a4238375bb142ae4627bdddade/config/notifiers.go#L60):
```
	VSendResolved: false,
```

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -57,7 +57,7 @@ var (
 	// DefaultSlackConfig defines default values for Slack configurations.
 	DefaultSlackConfig = SlackConfig{
 		NotifierConfig: NotifierConfig{
-			VSendResolved: false,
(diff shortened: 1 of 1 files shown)


fabxc:
If I recall our agreement correctly, the docs and not the code have to be adjusted.

pdbogen:
I don't think we had any discussion about that, but if that's the preference it doesn't bother me.

fabxc:
Yes, this was referring to a discussion I had with @brian-brazil

PR #273, opened 2016-03-09
Remove races from acceptance tests

fabxc:
Due to the generous timing these are probably never hit, but
should still be fixed.


brian-brazil:
:+1:

sdurrheimer:
Looks like Travis builds are still failing.

PR #277, opened 2016-03-11
Notifications: Improve OpsGenie error handling, and minor fixes

amorken:
Hello @fabxc,

I'm trying to upgrade our alertmanager installation and had to make a minor fix to make it handle the case where an alert is closed more than once, which can happen if i.e. an operator closes the alert manually. The behavior is not documented in OpsGenie's API docs, so the code is based on exploratory testing.

Perhaps in poor style, but very conveniently for myself, I also drive-by-fixed a few minor things while touching the file. I hope you don't mind that I included them here, but I'll be happy to break them out in a separate PR if you wish.

Thanks,
-AndersM

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -198,7 +198,7 @@ func (w *Webhook) Notify(ctx context.Context, alerts ...*types.Alert) error {
 	resp.Body.Close()
 
 	if resp.StatusCode/100 != 2 {
-		return fmt.Errorf("unexpected status code %v", resp.StatusCode)
+		return fmt.Errorf("unexpected status code %v from %s", resp.StatusCode, w.URL)
 	}
 
 	return nil
@@ -602,7 +602,7 @@ type OpsGenie struct {
 	tmpl *template.Template
 }
 
-// NewOpsGenieDuty returns a new OpsGenie notifier.
+// NewOpsGenie returns a new OpsGenie notifier.
 func NewOpsGenie(c *config.OpsGenieConfig, t *template.Template) *OpsGenie {
 	return &OpsGenie{conf: c, tmpl: t}
 }
@@ -626,6 +626,11 @@ type opsGenieCreateMessage struct {
 
 type opsGenieCloseMessage struct {
 	*opsGenieMessage `json:",inline"`
+}
+
+type opsGenieErrorResponse struct {
+	Code  int    `json:"code"`
+	Error string `json:"error"`
 }
 
 // Notify implements the Notifier interface.
@@ -684,11 +689,25 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) error {
 	if err != nil {
 		return err
 	}
-	resp.Body.Close()
+	defer resp.Body.Close()
 
-	if resp.StatusCode/100 != 2 {
+	if resp.StatusCode == 400 && alerts.Status() == model.AlertResolved {
 		body, _ := ioutil.ReadAll(resp.Body)
-		log.With("incident", key).Debugf("unexpected OpsGenie response %s: %s", resp.Status, body)
+
+		var responseMessage opsGenieErrorResponse
+		if err := json.Unmarshal(body, &responseMessage); err != nil {
+			return fmt.Errorf("could not parse error response %q", body)
+		}
+		const alreadyClosedError = 5
+		if responseMessage.Code == alreadyClosedError {
+			return nil
+		}
+		return fmt.Errorf("error when closing alert: code %d, error %q",
+			responseMessage.Code, responseMessage.Error)
+	} else if resp.StatusCode/100 != 2 {
+		body, _ := ioutil.ReadAll(resp.Body)
+		log.With("incident", key).Debugf("unexpected OpsGenie response from %s (POSTed %s), %s: %s",
+			apiURL, msg, resp.Status, body)
 		return fmt.Errorf("unexpected status code %v", resp.StatusCode)
 	}
 	return nil


notify/impl.go
```diff
@@ -680,12 +685,27 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) error {
...
-		log.With("incident", key).Debugf("unexpected OpsGenie response %s: %s", resp.Status, body)
-		return fmt.Errorf("unexpected status code %v", resp.StatusCode)
+
+		var responseMessage opsGenieErrorResponse
+		if err := json.Unmarshal(body, &responseMessage); err != nil {
+			return fmt.Errorf("OpsGenie: Could not parse error response \"%s\"", body)
+		}
+		// Response code 5 appears to indicate that an alert was already closed when we tried to close it.
+		// Consider that a success.
+		if responseMessage.Code == 5 {
```
grobie: Can we define a constant for that magic number? Probably removes the need for a comment as well.

grobie:
Looks great! :+1: I'll reach out to OpsGenie and ask them for details here.

notify/impl.go
amorken: Now that you mention it, yes, of course. =)

notify/impl.go
```diff
@@ -198,7 +198,7 @@ func (w *Webhook) Notify(ctx context.Context, alerts ...*types.Alert) error {
 	resp.Body.Close()
 
 	if resp.StatusCode/100 != 2 {
-		return fmt.Errorf("unexpected status code %v", resp.StatusCode)
+		return fmt.Errorf("Webhook: unexpected status code %v from %s", resp.StatusCode, w.URL)
```
fabxc: The integration names should be provided via structured logging `log.With("integration", "webhook").X(err)` on the caller side.

notify/impl.go
```diff
@@ -680,12 +685,27 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) error {
...
 
-	if resp.StatusCode/100 != 2 {
+	if resp.StatusCode == 400 && alerts.Status() == model.AlertResolved {
 		body, _ := ioutil.ReadAll(resp.Body)
-		log.With("incident", key).Debugf("unexpected OpsGenie response %s: %s", resp.Status, body)
-		return fmt.Errorf("unexpected status code %v", resp.StatusCode)
+
+		var responseMessage opsGenieErrorResponse
+		if err := json.Unmarshal(body, &responseMessage); err != nil {
+			return fmt.Errorf("OpsGenie: Could not parse error response \"%s\"", body)
```
fabxc: `%q` is a good replacement for `\"%s\"`

notify/impl.go
```diff
@@ -680,12 +685,27 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) error {
...
+		}
+		// Response code 5 appears to indicate that an alert was already closed when we tried to close it.
+		// Consider that a success.
+		if responseMessage.Code == 5 {
+			return nil
+		}
+		return fmt.Errorf("OpsGenie: Error when closing alert: Code %d, error \"%s\"",
```
fabxc: We generally keep error messages lower cased.

fabxc:
Thanks a lot!
We should merge this if it solves an issue but it would be good to validate with OpsGenie whether this is actually how it works.

zwopiR:
Hi @amorken, @grobie,

I chatted with OpsGenie just a moment ago. They confirmed Code: 5 means "already closed". The documentation is not yet public, but they are planning to release a doc about this.

brian-brazil:
That's good to know. It looks like you've still two pending comments from fabxc.

zwopiR:
Hi @bbrazil, 

It's not my PR, but I'd like to have it in master too. Is there anything I can do?

brian-brazil:
This change is quite close to being ready, so there's not much to be done.

If Anders doesn't respond within a few days and this is important to you, you could take their change, handle the pending comments and make a new PR.

notify/impl.go
amorken: Thanks, this Go language has some neat things, I should learn it some day. =)

notify/impl.go
amorken: Ack, fixed, thanks. =)

notify/impl.go
amorken: OK, I tried figuring out how this would work when the logging is made from the RetryNotifier - which has a reference to a Notifier instance, but not an Integration. Now, as I'm a complete Go noob I didn't find a trivial and non-ugly way to do this based on what I had in https://github.com/prometheus/alertmanager/blob/master/notify/notify.go#L193 - I tried a few things like modifying the Notifier interface to also include a name and such, but ... well, I'll just say I wasn't satisfied with what I came up with, so I never got around to responding to this.

I can remove these integration names from the error strings entirely if you like? They were very handy for pinning down the problem that led me to submitting this PR, though. =)

amorken:
Hi everyone, sorry about my extended radio silence.

I was stumped by the comment about logging the integration names on the calling sides, and before I was able to resolve that, good old Real Life caught up with me - we had a baby a couple of weeks ago, and it turns out keeping a baby boy as well as his two older siblings (and their mother... =) fed clean and happy turns out to leave little bandwidth for catching up with GitHub PR comments. =)

This means that I don't think I can promise to get any large changes done in a timely manner before next week, but I'll be more than happy to back out the integration names in error strings stuff if that is the only remaining objection. (Or, if someone can share some insight about how I could handle the Notifier/Integration naming conundrum that I can turn into a not-too-complex patch, I'll be happy to take a look at that too.

Thanks,
AndersM

brian-brazil:
Congratulations!

If you could just squash your commits, between @zwopiR's offer of help and the rest of us we should be able to take it from here.

amorken:
Thanks, I've squashed things into one commit, omitting the integration names in error strings.

-AndersM

fabxc:
Thanks a lot :+1:

PR #278, opened 2016-03-11
OpsGenie notifier: Add support for tags, teams, and using labels as tags.

sebzz:
@fabxc

This PR exposes some extra settings available in the OpsGenie alert creation API. Teams and LabelsToTags used to be available on a previous version but have since been removed.

OpsGenieConfig now has the following extra settings:
- Teams: Templated comma-separated string of teams  that this alert should be assigned to.
- Tags: Templated comma-separated string of tags that should be set on alert.

(Create Alert Api)[https://www.opsgenie.com/docs/web-api/alert-api#createAlertRequest]

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -280,6 +280,8 @@ type OpsGenieConfig struct {
 	Description string            `yaml:"description"`
 	Source      string            `yaml:"source"`
 	Details     map[string]string `yaml:"details"`
+	Teams       string            `yaml:"teams"`
+	Tags        string            `yaml:"tags"`
 
 	// Catches all undefined fields and must be empty after parsing.
 	XXX map[string]interface{} `yaml:",inline"`
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -620,6 +620,8 @@ type opsGenieCreateMessage struct {
 	Message string            `json:"message"`
 	Details map[string]string `json:"details"`
 	Source  string            `json:"source"`
+	Teams   string            `json:"teams,omitempty"`
+	Tags    string            `json:"tags,omitempty"`
 }
 
 type opsGenieCloseMessage struct {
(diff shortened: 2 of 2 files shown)


config/notifiers.go
```diff
@@ -275,11 +275,14 @@ func (c *WebhookConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
-	Source      string            `yaml:"source"`
-	Details     map[string]string `yaml:"details"`
+	APIKey       Secret            `yaml:"api_key"`
+	APIHost      string            `yaml:"api_host"`
+	Description  string            `yaml:"description"`
+	Source       string            `yaml:"source"`
+	Details      map[string]string `yaml:"details"`
+	Teams        []string          `yaml:"teams"`
+	Tags         []string          `yaml:"tags"`
+	LabelsToTags []string          `yaml:"labels_to_tags"`
```
brian-brazil: This seems messy to me, as it handles only one use case and doesn't offer any templating.

Could we do something like a delimited string?

sebzz: Hey @brian-brazil. Thanks for taking a look. What exactly do you mean by a delimited string? Is this regarding `LabelsToTags` only ?

sebzz: Also this was previously merged in #97

brian-brazil: Yes, only `LabelsToTags`.

If say semicolon is forbidden in Opsgenie tags you could have a single string like `a;b;c;d` that's then split out. That'd permit the full power of templating.

sebzz:
Thanks @brian-brazil . I've updated the PR. Actually the OpsGenie Create Alert Api is documented as taking tags as a comma delimited tag list so there was no need to split the string. It just so happened that their API worked with an array of tags as well.
This version just applies the templating function onto the passed in tags string.

brian-brazil:
That makes it simple alright. What shall we do about teams?

sebzz:
@brian-brazil - tested with a comma separated string for Teams.

fabxc:
Seems :+1: then to me.

brian-brazil:
Can you squash your commits please?

sebzz:
@brian-brazil @fabxc - squashed commit.

brian-brazil:
Thanks!

Issue #285, opened 2016-03-30
question about glibc


joeblew99:
Is this a dependency ? Its in the readme, but i just want to check.

juliusv:
It's a build-time dependency, but not a run-time one.

```
$ ldd alertmanager
	not a dynamic executable
```

juliusv:
Closing this as it seems answered. Please reopen if you have further questions!

juliusv:
@joeblew99 I made a PR to clarify that: https://github.com/prometheus/alertmanager/pull/286

joeblew99:
Ok thanks all. Loving Prometheus. Took 10 minutrs to deploy and config. Great stuff !!!

PR #286, opened 2016-03-30
Clarify build- vs. run-time dependencies in README.md

juliusv:
See https://github.com/prometheus/alertmanager/issues/285


beorn7:
:+1:

Issue #287, opened 2016-03-30
OpsGenie notifier fails if teams and tags is not provided


zwopiR:
the OpsGenie notifier fails with 
```
WARN[0034] Notify attempt 4 failed: templating error: template: : "" is an incomplete or empty template; defined templates are: "slack.default.titlelink", "hipchat.tmpl", "__subject", "slack.default.pretext", "slack.default.title", "pagerduty.default.description", "__description", "", "slack.default.username", "opsgenie.default.source", "__text_alert_list", "email.default.html", "opsgenie.description", "opsgenie.default.description", "pushover.default.title", "slack.default.fallback", "hipchat.default.message", "__alertmanagerURL", "pagerduty.default.client", "pagerduty.default.instances", "pushover.default.url", "slack.default.iconemoji", "slack.default.text", "pagerduty.default.clientURL", "opsgenie.tmpl", "hipchat.default.from", "hipchat.from", "__alertmanager", "hipchat.message", "pushover.default.message", "email.default.subject"  source=notify.go:193
^CINFO[0034] Received SIGTERM, exiting gracefully...       source=main.go:206
```
if the config item `teams` and/or `tags` is not provided.
In notify/impl.go (https://github.com/prometheus/alertmanager/blob/master/notify/impl.go#L670) `ExecuteTextString` is called on the values Teams/Tags which are initialized to nil, if the config params are not specified resulting the above template error.

I would provide a PR, but don't know what solution to pick:

- make team and tags mandatory: I thinks this is not suitable for all OpsGenie users, since teams and tags can be configured too within OpsGenie (in the API integration)
- provide default values: not suitable too, since alertmanager can't know the team names
- don't send the teams and tags key/value pairs in the call to OpsGenie API if they are not provided: probably the best idea, but I don't know how

Any ideas?

brian-brazil:
@sebzz

zwopiR:
ah - it duplicates https://github.com/prometheus/alertmanager/pull/281

PR #288, opened 2016-03-30
fix for Issue 287 (https://github.com/prometheus/alertmanager/issues/…

zwopiR:
a proposal for https://github.com/prometheus/alertmanager/issues/287 to be discussed

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -82,6 +82,9 @@ func FromGlobs(paths ...string) (*Template, error) {
 
 // ExecuteTextString needs a meaningful doc comment (TODO(fabxc)).
 func (t *Template) ExecuteTextString(text string, data interface{}) (string, error) {
+	if text == "" {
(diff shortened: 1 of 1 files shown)


brian-brazil:
We should be fixing this for all notifiers over in ExecuteTextString, rather than just for opsgenie.

zwopiR:
@brian-brazil here is a proposal that fixes #287 as well as similar cases in other notifiers. Can you help me with the failing test? I can't reproduce it here on my machine

fabxc:
Thanks :+1: 

Restarted the tests. The acceptance tests are unfortunately very flaky on the low resources CI provides.

zwopiR:
Thanks fabxc

brian-brazil:
This is a general resilience and usability change, so we'll take this one anyway. Thanks!

Issue #212, opened 2016-01-05
Allow alertmanager to be run behind a reverse proxy


grobie:
The `-web.external-url` flag is no longer respected when loading assets. It's no  longer possible to run alertmanager behind a reverse proxy.

a86c6f7964:
FYI I just did https://github.com/novilabs/alertmanager/commit/50937b09100abcc1d28f933baa38e2b22aa0ef01 to be able to use a prefix path of ```/alerts``` and run it behind my reverse proxy. It seems to be working well. I do not know a way to pass it through as configuration to the html and js files though. If anybody wants to know what files would need to be touched.

hryamzik:
```-web.external-url``` does seem to be broken (again?). @grobie, any comments on this?

grobie:
It never got fixed so far unfortunately. Pull requests very welcome, I think all core contributors are very overworked right now.

juliusv:
Just looking at it, I got it almost working, except for one thing: the `index.html` is a static HTML page (not a Go template), so we can't inject the path prefix into it to fix up the JS and CSS asset links. Should we serve the web pages (at least `index.html`) through the Go template engine, like we do in Prometheus?

fabxc:
Without a closer look – that doesn't seem like something that should depend on passing through a template engine.
If it comes with complexity I cannot think of right now, sure – it doesn't have a relevant overhead.

juliusv:
@fabxc True. The alternative is to make all the paths between the assets relative. Trying that out now, it seems to work. It would mean that one has to be careful to not introduce any absolute paths anywhere, ever. But requiring a path prefix everywhere in the HTML+JS would have the same problem. Should I go for relative paths then?

fabxc:
Sure, it's not that many files and won't change frequently.

mrwacky42:
+1 to relative paths.

juliusv:
Cool. Fix is out in https://github.com/prometheus/alertmanager/pull/289/files.

sdwerwed:
I made it work like this (`--web.external-url` not `-web.external-url`)
`/usr/local/bin/alertmanager --config.file /etc/alertmanager/alertmanager.yml --web.external-url=http://example.com:9093/alertmanager` then add to nginx `location /alertmanager {proxy_pass http:localhost:9093/;}`

PR #289, opened 2016-03-30
Re-add full -web.external-url functionality

juliusv:
Fixes https://github.com/prometheus/alertmanager/issues/212


juliusv:
@fabxc @mrwacky42

main.go
```diff
@@ -183,8 +186,8 @@ func main() {
 
 	router := route.New()
 
-	RegisterWeb(router)
-	api.Register(router.WithPrefix("/api"))
+	RegisterWeb(router.WithPrefix(amURL.Path))
+	api.Register(router.WithPrefix(amURL.Path + "/api"))
```
fabxc: path.Join

juliusv: Hmm, we're usually just +-ing together pre-sanitized paths in Prometheus, including in your own https://github.com/prometheus/common/blob/master/route/route.go. Not that it matters much, but maybe we should keep it consistent?

ui/app/js/app.js
```diff
@@ -406,21 +406,21 @@ angular.module('am').config(
...
       controller: 'SilencesCtrl',
       reloadOnSearch: false
     }).
     when('/status', {
-      templateUrl: '/app/partials/status.html',
+      templateUrl: 'app/partials/status.html',
       controller: 'StatusCtrl'
     }).
     otherwise({
       redirectTo: '/alerts'
```
mrwacky42: This redirect to `/alerts` won't work correctly when proxying to a different base URL.
(In our case `/alertmanager`...)

juliusv: As far as I can tell, this all works fine for me. Have you tested it? E.g. I run:

```
./alertmanager -web.external-url=http://localhost:9093/foo/bar
```

Going to `http://localhost:9093/foo/bar` correctly redirects me to `http://localhost:9093/foo/bar/#/alerts`.

Note that this is an angular route name that appears in the hash part of the URL, and is thus not part of the normal path.

Or am I misunderstanding something?

main.go
grobie: path.Join does not work well for URLs in general, I usually decide against using it.

https://play.golang.org/p/xGEOg3Gw7p

juliusv: path.Join is fine in general, but can only be applied to paths, not full URLs. Otherwise it'll get rid of duplicated slashes and do all kinds of other things you might not want.

ui/app/index.html
```diff
@@ -1,32 +1,32 @@
 <!DOCTYPE html>
 <html lang="en" ng-app="am">
 <head>
-	<script src="/lib/jquery.min.js"></script>
```
grobie: Not a fan of relative locations. This makes it a hard requirement that all entrypoints are mounted under the root (like `/silences/elapsed` would not be possible to add).

Alertmanager unfortunatley uses the awful angular path syntax though, so the requirement is already fulfilled.

main.go
mrwacky42: Aren't side effects fun?

juliusv: Note that this is not an issue here though, since we're dealing with paths only. Both `path.Join` or `+` would work fine, just that I think `+` is what we usually use in Prometheus in this situation.

ui/app/index.html
juliusv: I'm not a fan of relative paths either, but the alternative would be to introduce Go templating into the asset serving in order to be able to inject the path prefix. See https://github.com/prometheus/alertmanager/issues/212.

main.go
fabxc: A full URL is not a path. Of course that doesn't work and it shouldn't. But this is a path (`url.URL.Path`) :)

Do as you wish. But it would be correct to apply here and prevent keep things working, when sanitization somewhere up the chain changes.

ui/app/index.html
fabxc: We can make it all nice and shiny when we redo the UI ;)

grobie: :+1:

grobie:
:+1:

main.go
juliusv: Na gut, `path.Join` for @fabxc :)

mrwacky42:
Doesn't this merit a 0.1.2 release? Our configuration won't work if we can't proxy. 
And if we can't upgrade Alertmanager, we can't upgrade Prometheus.
See my sad panda face: 😦

flecno:
hey... when do you will release this fix?

juliusv:
@fabxc would probably know when a next release is planned for the Alertmanager.

fabxc:
`-web.external-url` is hacky implementation-wise in many ways and prometheus/prometheus#1191 that suggests we need something different to generally work anyway.

I'd like some ideas or opinions on that.

flecno:
Thanks for your fast reply... You are right.. it's a little bit hacky... Why you are using angular for this frontend? Mhm. But I need this fix to get alertmanger working behind a reverser proxy. I should build my own release for now...
After all: **THANK YOU** FOR YOUR BEAUTIFUL prometheus!

fabxc:
It's hacky independent of the frontend. Look at the Prometheus code for comparison...
There's a good chance we will release a version with the storage changes to resolve problems people are encountering with SQLite persistence. In that case the flag would be included again soon.

Still would prefer a counter-proposal.

fabxc:
FYI, @flecno @mrwacky42 there was a `0.2` release that contains these changes.

fabxc:
FYI, @flecno @mrwacky42 there was a `0.2` release that contains these changes.

PR #294, opened 2016-04-07
Change fallback fonts to something readable

ben51:
Change-Id: I5f5ce8c7113b2188083ef0ec9ef2bcc4d74f5500


fabxc:
Mh, what's not readable about it?
A screenshot would help.

ben51:
Because the first three fonts are not installed on my machine, it falls back to "Courier" which is pretty ugly.
screen

PR #295, opened 2016-04-07
Limit 'from' length to 64 chars to ensure conformity with the Hipchat API

utobi:
@fabxc : 'from' can be longer than 64 chars which results in a bad request to the Hipchat API and a dropped notification. This change is supposed to cap the length of 'from' to ensure that the alertmanager never sends something that is too long and does not comply with the API.

Still very new to go so if there is a way to solve this better I'm happy to look into it. :)

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -573,8 +573,13 @@ func (n *Hipchat) Notify(ctx context.Context, as ...*types.Alert) error {
 		msg = tmplText(n.conf.Message)
 	}
 
+	from := tmplText(n.conf.From)
+	if len(from) > 64 {
+		from = from[:64]
+	}
+
 	req := &hipchatReq{
-		From:          tmplText(n.conf.From),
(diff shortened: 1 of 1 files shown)


notify/impl.go
```diff
@@ -573,8 +573,13 @@ func (n *Hipchat) Notify(ctx context.Context, as ...*types.Alert) error {
 		msg = tmplText(n.conf.Message)
 	}
 
+	from := tmplText(n.conf.From)
+	if len(from) > 64 {
```
brian-brazil: We're going to have a lot of these, I'd suggest breaking this out to a function.

fabxc:
Good catch.
As @brian-brazil said a function would be nice. It could then also truncate with a `...` suffix so it doesn't look like some sort of bug to the user.

fabxc:
Superseded by #449

PR #296, opened 2016-04-08
Use runtime.Version() instead of injected version

juliusv:
@fabxc @sdurrheimer

### DIFF
--- a/version/info.go
+++ b/version/info.go
@@ -13,7 +13,11 @@
 
 package version
 
-import "github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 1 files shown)


sdurrheimer:
:+1:

sdurrheimer:
@juliusv whoops missing removal here : https://github.com/prometheus/alertmanager/blob/master/scripts/build.sh#L37

juliusv:
@sdurrheimer Ah right, thanks! Pushed fix to master.

Issue #275, opened 2016-03-10
alertmanager: how to send mail notification with smtp server?


dannik1976:
Good morning friends, I need your help:
I have configured all successfuly, alertmanager takes notifications from prometheus but i would send e-mail alert notification. I have installed sendmail to test outgoing mail with my user and psw smtp and this test was successful.
Can you tell me what I have to write in alertmanager.conf (Actually my alertmanager.conf is empty) in email_configs section please?
if my smtp server is smtp.xxx.it 
if my user is xxx@xxx.it 
if my psw is xxx

I did not find any example about that.
Thanks in advance.

Daniele

beorn7:
You have to configure alertmanager in `alertmanager.conf`.
See http://prometheus.io/docs/alerting/configuration/#email-receiver-<email_config>
ro configure the mail server and
http://prometheus.io/docs/alerting/configuration/#route-route to
configure the alert routing.

dannik1976:
# The email address to send notifications to.
to: mymail@gmail.com
# The sender address.
from: I have to insert my smtp server? smtp.xxx.it?
# The SMTP host through which emails are sent.
smarthost: I must necessary configure a smarthost?

help please.
Thanks

beorn7:
For "from", you use anything that makes sense as a sender address for alerts.
And yes, you must configure a smarthost.

Usually, both would be set in the global section as default values for all notifications.

Example:
```
global:
  smtp_smarthost: localhost:25
  smtp_from: alertmanager@example.com
```

sjoerdmulder:
You have too use the 587 port and not 465; i found this out the hard-way and it is not documented at all but the underlying Go API says it in the docs: https://golang.org/pkg/net/smtp/#SendMail 

SendMail **connects** to the server at addr, switches to TLS **if possible**. 

_So SSL from the start doesn't work_
```
global:
  # The smarthost and SMTP sender used for mail notifications.
  smtp_smarthost: 'smtp.xxx.it:587'
  smtp_from: 'Alertmanager <alertmanager@example.org>'
```
Next you need to set 2 environment variables containing your user / passsword
```
SMTP_AUTH_USERNAME=xxx@xxx.it
SMTP_AUTH_PASSWORD=xxx
```
We use Sendgrid with this configuration

grobie:
@sjoerdmulder Please have a look at https://github.com/prometheus/alertmanager/pull/266. The behavior around TLS has recently changed.

fabxc:
[...]

PR #300, opened 2016-04-10
Clarify SMTP authentication in the docs

mpchadwick:
Wasn't initially clear to me what was needed to authenticate with e.g. Mandrill, Mailgun, etc..

Seems some others likely had [similar issues](https://github.com/prometheus/alertmanager/issues/275)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ This is an example configuration that should cover most relevant aspects of the 
 ```yaml
 global:
   # The smarthost and SMTP sender used for mail notifications.
+  # If the host requires authentication, set SMTP_AUTH_USERNAME and SMTP_AUTH_PASSWORD
+  # environment variables.
   smtp_smarthost: 'localhost:25'
   smtp_from: 'alertmanager@example.org'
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -49,6 +49,8 @@ This is an example configuration that should cover most relevant aspects of the
 ```yaml
 global:
   # The smarthost and SMTP sender used for mail notifications.
+  # If the host requires authentication, set SMTP_AUTH_USERNAME and SMTP_AUTH_PASSWORD
+  # environment variables
```
brian-brazil: Comments should be full sentences, so add a full stop

brian-brazil:
Can you squash your commits toot?

mpchadwick:
Updated. I was actually thinking that it might make sense to remove the example from `README.md` and just link to `docs/example/simple.yml`. What do you think? Having two examples makes it more likely that things will get out of sync / something will be missed in one or the other.

brian-brazil:
This is probably our most common user question about the alertmanager, so it's best to have it in the readme.

Thanks!

pydolan:
FYI: the prometheus documentation could use this change as well: https://prometheus.io/docs/alerting/configuration/ (this link currently has no mention of the environment variables)

PR #301, opened 2016-04-11
Provide feedback in console about port

mpchadwick:
Other than by consulting the source code or looking at the suggested nginx proxy configuration [the default port isn't really documented anywhere](https://github.com/prometheus/alertmanager/search?utf8=%E2%9C%93&q=9093). As a newbie, when I started the service, I wasn't exactly sure what port it was running on. This could be helpful for some people.

### DIFF
--- a/main.go
+++ b/main.go
@@ -186,6 +186,7 @@ func main() {
 	RegisterWeb(router)
 	api.Register(router.WithPrefix("/api"))
 
+	log.Infof("Listening on %s", *listenAddress)
 	go listen(router)
 
 	var (


main.go
```diff
@@ -173,6 +173,7 @@ func main() {
 		disp = NewDispatcher(alerts, NewRoute(conf.Route, nil), build(conf.Receivers), marker)
 
 		go disp.Run()
+		log.Infof("Alertmanager is now listening on port %s", strings.TrimPrefix(*listenAddress, ":"))
```
sdurrheimer: Hello,

I think it would be more appropriate to put this log above `go listen(router)`.
Also, why not just display the full `listenAddress` like it is done in [`prometheus/prometheus`](https://github.com/prometheus/prometheus/blob/master/web/web.go#L239) ?
```
log.Infof("Listening on %s", *listenAddress)
```

mpchadwick:
Hi - I've updated the PR based on your feedback.

grobie:
:+1: Ideally we'd also log version and the other build information as we do in Prometheus for example. Can be done in a follow-up change.

sdurrheimer:
@grobie other informations are already logged:
```
$ ./alertmanager -config.file doc/examples/simple.yml
alertmanager, version 0.1.1 (branch: mpchadwick-listening-port, revision: 458f0cc)
  build user:       steve@steve-Latitude-e7440
  build date:       20160412-05:49:29
  go version:       go1.5.3
INFO[0000] Loading configuration file                    file=doc/examples/simple.yml source=main.go:150
INFO[0000] Listening on :9093                            source=main.go:193
```

@mpchadwick Thanks for the contribution, LGTM :+1:

grobie:
Ha, even better!

PR #302, opened 2016-04-11
Add routing-tree visualization

stuartnelson3:
Adds a visualization of the routing tree defined in `config.yml`, as well as the ability to trace entered labelSets.

The tree is rendered as the final component on the Status page.

Additionally, changed the html indentation from tabs to spaces. You're welcome.

All styling is up to debate. Since alert manager needs a redesign I didn't labor over integrating this.

Also, the javascript code is barely cleaner than a spike. There are probably unintentional globals floating around still.


ui/lib/routing-tree.js
```diff
@@ -0,0 +1,260 @@
...
+    .style("background-color", "white")
+    .style("border", "1px solid #ddd")
+    .style("font", "9px monospace")
+    .style("padding", "4px 2px")
+    .style("z-index", "10")
+    .style("visibility", "hidden");
+
+// Global for now so we can play with it from the console
+function parseSearch(searchString) {
+  var labels = searchString.replace(/{|}|\"|\s/g, "").split(",");
```
fabxc: Is GH syntax highlighting not getting the `\"` in the regex?

fabxc:
Great! Others haven't seen any screenshots yet. Could you add those?

ui/lib/routing-tree.js
stuartnelson3: Yeah, that seems to be the case. It happens on my own repo with this code as well, even though the code compiles and runs correctly. I guess it's something about the syntax highlighter not picking up that it's within a regexp literal.

stuartnelson3:
Here's how it looks with the example config.yml. Since there are so few routes and the algorithm balances the branches equally in a circle, it looks pretty sparse. I'll send you a private one with real configuration so you can see how it would normally look.

image

SuperQ:
:+1: Nice.

beorn7:
Just FTR: Looks breathtakingly beautiful with complex routing trees... we just cannot make Soundcloud's tree public here... ;)

stuartnelson3:
I'll wait for a :+1: from @fabxc before I squash n merge

fabxc:
Cool, I don't think I've much qualification to comment on JS etc.
Should we just get it in and see whether we've to iterate on things later?

stuartnelson3:
That sounds reasonable to me. The code is all contained within a single js file, so it should be easy to modify.

fabxc:
👍  Nice work

Issue #306, opened 2016-04-13
Move smtp auth parameters into config


brian-brazil:
For historical reasons, the smtp auth config is in environment variables. We should move it into the config file like all the rest of the notifier auth parameters.

mpchadwick:
This was fixed in #308

PR #308, opened 2016-04-15
Move SMTP auth to the config file

mpchadwick:
Implements enhancement proposed in #306.

Note: I have left support for environment variables as a fallback since we probably don't want people's alerts to break if they pull down master and don't move their credentials from env variables to the config file.

Documentation additionally updated to reflect the fact that the config file is the preferred place to store creds.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,8 +49,6 @@ This is an example configuration that should cover most relevant aspects of the 
 ```yaml
 global:
   # The smarthost and SMTP sender used for mail notifications.
-  # If the host requires authentication, set SMTP_AUTH_USERNAME and SMTP_AUTH_PASSWORD
-  # environment variables.
   smtp_smarthost: 'localhost:25'
   smtp_from: 'alertmanager@example.org'
 
--- a/config/config.go
+++ b/config/config.go
@@ -26,7 +26,7 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-var patAuthLine = regexp.MustCompile(`((?:api_key|service_key|api_url|token|user_key):\s+)(".+"|'.+'|[^\s]+)`)
+var patAuthLine = regexp.MustCompile(`((?:api_key|service_key|api_url|token|user_key|password|secret):\s+)(".+"|'.+'|[^\s]+)`)
 
 // Secret is a string that must not be revealed on marshaling.
 type Secret string
@@ -160,6 +160,18 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 				}
 				ec.From = c.Global.SMTPFrom
 			}
+			if ec.AuthUsername == "" {
+				ec.AuthUsername = c.Global.SMTPAuthUsername
+			}
+			if ec.AuthPassword == "" {
+				ec.AuthPassword = c.Global.SMTPAuthPassword
+			}
+			if ec.AuthSecret == "" {
+				ec.AuthSecret = c.Global.SMTPAuthSecret
+			}
+			if ec.AuthIdentity == "" {
+				ec.AuthIdentity = c.Global.SMTPAuthIdentity
+			}
 		}
 		for _, sc := range rcv.SlackConfigs {
 			if sc.APIURL == "" {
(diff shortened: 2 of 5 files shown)


config/config.go
```diff
@@ -225,6 +243,20 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return checkOverflow(c.XXX, "config")
 }
 
+func smtpAuthParam(key string, gc *GlobalConfig) (string, Secret) {
+	// Try to pull it from the global config
+	globalKey := "SMTPAuth" + key
+	reflected := reflect.Indirect(reflect.ValueOf(gc))
```
brian-brazil: This is getting a bit magic, I think it'd be okay to just switch to config file for this.

README.md
```diff
@@ -49,10 +49,10 @@ This is an example configuration that should cover most relevant aspects of the
 ```yaml
 global:
   # The smarthost and SMTP sender used for mail notifications.
-  # If the host requires authentication, set SMTP_AUTH_USERNAME and SMTP_AUTH_PASSWORD
-  # environment variables.
   smtp_smarthost: 'localhost:25'
+  smtp_auth_username: 'alertmanager'
```
brian-brazil: With this in the config now, we can remove this from here and just have it in the docs and simple.yml.

config/config.go
```diff
@@ -257,6 +289,10 @@ type GlobalConfig struct {
 
 	SMTPFrom         string `yaml:"smtp_from"`
 	SMTPSmarthost    string `yaml:"smtp_smarthost"`
+	SMTPAuthUsername string `yaml:"smtp_auth_username"`
+	SMTPAuthPassword Secret `yaml:"smtp_auth_password"`
```
brian-brazil: These need to be added to patAuthLine

mpchadwick:
@brian-brazil updated per your feedback. I can squash as needed if you're OK with this (all into 1 commit or into 2 commits, one for code changes and the other for documentation).

config/config.go
```diff
@@ -26,7 +26,7 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-var patAuthLine = regexp.MustCompile(`((?:api_key|service_key|api_url|token|user_key):\s+)(".+"|'.+'|[^\s]+)`)
+var patAuthLine = regexp.MustCompile(`((?:api_key|service_key|api_url|token|user_key|auth_password|auth_secret):\s+)(".+"|'.+'|[^\s]+)`)
```
fabxc: Just making that `password` and `secret` will be enough. We are just so specific for `key` and `url` because it could be something normal.

brian-brazil:
That looks good, can you squash please?

config/notifiers.go
```diff
@@ -120,12 +120,16 @@ type EmailConfig struct {
...
-	To         string            `yaml:"to"`
-	From       string            `yaml:"from"`
-	Smarthost  string            `yaml:"smarthost,omitempty"`
-	Headers    map[string]string `yaml:"headers"`
-	HTML       string            `yaml:"html"`
-	RequireTLS bool              `yaml:"require_tls"`
+	To           string            `yaml:"to"`
+	From         string            `yaml:"from"`
+	Smarthost    string            `yaml:"smarthost,omitempty"`
+	AuthUsername string            `yaml:"auth_username"`
```
fabxc: These all have the same prefix. Would it be worthwhile to move them into their own YAML block?

fabxc:
@mpchadwick Thanks a lot! Added two comments.

mpchadwick:
Rebased to a single commit.

> Just making that `password` and `secret` will be enough. We are just so specific for key and url because it could be something normal.

I updated the RegEx to be more loose. I agree that probably we'd always want to obscure anything matching these terms.

> These all have the same prefix. Would it be worthwhile to move them into their own YAML block?

I guess it would be something like this?

```yml
global:
  # The smarthost and SMTP sender used for mail notifications.
  smtp_smarthost: 'localhost:25'
  smtp_auth:
  - username: 'alertmanager'
    password: 'password'  
```

I didn't address it because your comment was a question, but if you'd prefer that syntax I can.

fabxc:
Yes, that's what I had in mind. I would say grouping things together in a block is a good idea if you have a common prefix. We did it similar to TLS configs in Prometheus.

Of course any objections can be discussed.

brian-brazil:
For TLS there were 20-30 potential fields as things expand, there's only 4 here. I'd see this along the same lines as us not having an `auth` field in Prometheus for basic auth vs. token vs tls.

fabxc:
But we do: https://github.com/prometheus/prometheus/blob/master/config/config.go#L384

brian-brazil:
We have a `basic_auth`, a `bearer_token` and a `bearer_token_file` all of which are auth related. These are not however nested under an `auth`.

fabxc:
Ah, I see what you mean.
So it would be symmetric to put username and password in a separate block. However, don't see strong need in general. We can leave it as it is.

brian-brazil:
Yeah. There are a few other possible auth options, but if they come up (sound to be most likely in a full-Windows environment) we can refactor this a bit.

fabxc:
Okay. Thanks @mpchadwick 👍

pedro-caseonit:
Dear @kaihendry could you be more specific or just simply paste an example of your *alertmanager.yml* because I'm having the same issue trying to send notification through SES and I don't understand what goes on!
```
time="2016-07-11T12:09:57Z" level=warning msg="Notify attempt 1 failed: *smtp.plainAuth failed: 535 Authentication Credentials Invalid" source="notify.go:193"
```
Thank in advance.

PR #309, opened 2016-04-15
CSS changes for readability

jjneely:
Being somewhat visually challenged, I find these changes improve
readability drastically.  This removes some opacity settings, removes
smaller fonts, and address a couple colors.  I now find the native and
more subtle bootstrap mouse overs to be very helpful in locating where
the mouse is with sacrificing being able to read the alerts with a quick
glance.

@fabxc, I'm not a UI designer by any means.

### DIFF
--- a/ui/app/css/main.css
+++ b/ui/app/css/main.css
@@ -61,13 +61,6 @@ a.gen-link button {
 	margin-right: 12px;
 }
 
-.alert-item .expand, .silence-item .expand {
-	opacity: .4;
-}
-.alert-item:hover .expand, .silence-item:hover .expand {
-	opacity: 1;
-}
-
 .silence-item .labels, .alert-item .labels {
 }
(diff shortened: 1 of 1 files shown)


fabxc:
Thanks, any help is greatly appreciated.
It would be helpful if you could attach screenshots showing the difference for everyone.

jjneely:
Current styling:

am-before

With my changes:

am-after

Minus the horrible obfuscation, of course.

fabxc:
I agree that its more readable but looks pretty raw in some ways. I probably shouldn't make judgements on design – maybe someone else can provide an opinion.

@juliusv @grobie

grobie:
I don't really have an opinion on this change, none of the versions makes me really happy :p I'm not visually impaired when it comes to computer screens, but if it helps others I'm fine with it.

fabxc:
New UI will hopefully fix readability more generally :) Closing here

Issue #314, opened 2016-04-20
Add safeHtml template function


brian-brazil:
We should copy https://github.com/prometheus/prometheus/blob/master/template/template.go#L148 over to the alertmanager.

Reference: http://stackoverflow.com/questions/36735479/prometheus-using-html-content-in-alerts-annotations-and-using-it-in-email-templa

PR #315, opened 2016-04-20
Add safeHtml template function.

juliusv:
Fixes https://github.com/prometheus/alertmanager/issues/314


juliusv:
@brian-brazil Why is this called `safeHtml`, not `safeHTML`, in Prometheus actually? Because it's not a Go function and we apply a different style? Or accident?

brian-brazil:
I forget the reason, I was probably keeping everything camelCase.

brian-brazil:
:+1:

juliusv:
Ok.

Issue #313, opened 2016-04-20
Hide silenced alerts


maximilianvw:
A new filter on the main page that only shows alerts that are not silenced would be great!

juliusv:
:+1:

fabxc:
Fixed via #319

PR #319, opened 2016-04-22
Add a filter to hide silenced alerts

mpchadwick:
An implementation for the feature request in #313.

Just uses CSS `display: none` to hide the silenced alerts. Worth noting that the alert group node will not be removed if all alerts are silenced (the UI feels like an accordion in this case).

**Before Filtering**

image

**After Filtering**

image


matthiasr:
I like this. What do you think about moving the check box to be beside the
receivers filter? That way the alerts don't move further down, and we have
a lot of space there for more filters.

fabxc:
Nice, thanks! 👍

ui/app/partials/alerts.html
```diff
@@ -2,14 +2,20 @@
...
+		</section>
+
+		<section>
+			<label class="checkbox"><input type="checkbox" ng-model="hideSilenced"> Hide silenced</label>
+		</section>
 	</fieldset>
 </form>
 
-<div id="alert-groups">
+<div id="alert-groups" ng-class="{'hide-silenced': hideSilenced}">
```
juliusv: Can this just be `ng-hide="hideSilenced"`? Then you don't need any CSS... same for the other one in alert.html.

mpchadwick: @juliusv Hmm...I think that would hide the entire `alert-groups` node. In order to avoid the DOM / CSS approach we'd probably need to do something like this on the `alert-item` (untested)...

```js
ng-hide="alert.silenced && hideSilenced"
```

`hideSilenced` is currently a property on `AlertsCtrl`s `$scope`. So we'd need to make it available in `AlertCtrl` either via `$rootScope` or a service. Thoughts?

mpchadwick:
@matthiasr I initially thought about floating and using that real estate, but then when I added it initially at the bottom I thought it looked fine. If we start having tons of filters we would probably want the two columns (and maybe even show more / show less). I thought it might look weird / lonely to have a second column with a lone checkbox. I can add a screenshot later though.

ui/app/partials/alerts.html
juliusv: Ah, the nested nature of this wasn't apparent to me. All makes sense then, I guess keep it like it is.

mpchadwick:
@matthiasr updated to use 2 columns for the filters. I tested it out and *do* think it looks better after all

image

@fabxc anything holding back the merge here?

matthiasr:
LGTM

fabxc:
Thanks!

PR #326, opened 2016-04-27
Add resolved alerts to pushover template

teemow:
The pushover notification of resolved alerts can still end up in an empty
message (only a newline). The result is an `unexpected status code 400` with a `message cannot be blank` response from pushover. I've fixed the check for an empty message with
trimming the whitespace. But I also thought that adding the resolved
alerts to the message would be helpful.

ping @fabxc


fabxc:
Thanks! 👍

PR #337, opened 2016-05-01
Bolt intermediate storage

fabxc:
As discussed, we need a solution that satisfies the current feature set in an acceptable way.
This gives us time to do an potential replacement without time pressure.

@brian-brazil @juliusv

### DIFF
--- a/.promu.yml
+++ b/.promu.yml
@@ -1,4 +1,4 @@
-go: 1.5.4
+go: 1.6.2
 repository:
     path: github.com/prometheus/alertmanager
 build:
--- a/provider/boltmem/boltmem.go
+++ b/provider/boltmem/boltmem.go
@@ -1,0 +1,174 @@
+package boltmem
+
+import (
+	"encoding/binary"
+	"path/filepath"
+	"sync"
+
+	"github.com/boltdb/bolt"
+	"github.com/prometheus/alertmanager/provider"
+	"github.com/prometheus/alertmanager/types"
+	"github.com/prometheus/common/model"
+)
+
+// Alerts gives access to a set of alerts. All methods are goroutine-safe.
+type Alerts struct {
+	mtx    sync.RWMutex
+	alerts map[model.Fingerprint]*types.Alert
+
+	listeners map[int]chan *types.Alert
+}
+
+func NewAlerts() (*Alerts, error) {
+	return nil, nil
+}
+
+// Subscribe returns an iterator over active alerts that have not been
+// resolved and successfully notified about.
+// They are not guaranteed to be in chronological order.
+func (a *Alerts) Subscribe() provider.AlertIterator {
+	return nil
+}
+
+// GetPending returns an iterator over all alerts that have
+// pending notifications.
+func (a *Alerts) GetPending() provider.AlertIterator {
+	return nil
+}
+
+// Get returns the alert for a given fingerprint.
+func (a *Alerts) Get(model.Fingerprint) (*types.Alert, error) {
+	return nil, nil
+}
+
+// Put adds the given alert to the set.
+func (a *Alerts) Put(...*types.Alert) error {
+	return nil
+}
+
+// Silences gives access to silences. All methods are goroutine-safe.
+type Silences struct {
+	db *bolt.DB
+}
+
+func NewSilences(path string) (*Silences, error) {
+	db, err := bolt.Open(filepath.Join(path, "silences.db"), 0666, nil)
+	if err != nil {
+		return nil, err
+	}
+	return &Silences{db: db}, nil
+}
+
+// The Silences provider must implement the Muter interface
+// for all its silences. The data provider may have access to an
+// optimized view of the data to perform this evaluation.
+func (s *Silences) Mutes(lset model.LabelSet) bool {
+	return false
+}
+
+// All returns all existing silences.
+func (s *Silences) All() ([]*types.Silence, error) {
+	return nil, nil
+}
+
+// Set a new silence.
+func (s *Silences) Set(*types.Silence) (uint64, error) {
+	return 0, nil
+}
+
+// Del removes a silence.
+func (s *Silences) Del(uint64) error {
+	return nil
+}
+
+// Get a silence associated with a fingerprint.
+func (s *Silences) Get(uint64) (*types.Silence, error) {
+	return nil, nil
+}
+
+// Notifies provides information about pending and successful
+// notifications. All methods are goroutine-safe.
+type Notifies struct {
+	db *bolt.DB
+}
+
+func NewNotifies(path string) (*Notifies, error) {
+	db, err := bolt.Open(filepath.Join(path, "notifies.db"), 0666, nil)
+	if err != nil {
+		return nil, err
+	}
+	err = db.Update(func(tx *bolt.Tx) error {
+		_, err := tx.CreateBucketIfNotExists(bktNotifies)
+		return err
+	})
+	return &Notifies{db: db}, err
+}
+
+var (
+	bktNotifies = []byte("notifies")
+)
+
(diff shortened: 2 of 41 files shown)


fabxc:
Looks like CI environment decodes/encodes time values differently and simple recursive comparison fails...

provider/boltmem/boltmem.go
```diff
@@ -0,0 +1,426 @@
...
+			ab := b.Get(fp)
+
+			// Merge the alert with the existing one.
+			if ab != nil {
+				var old types.Alert
+				if err := json.Unmarshal(ab, &old); err != nil {
+					return fmt.Errorf("decoding alert failed: %s", err)
+				}
+				// Merge alerts if there is an overlap in activity range.
+				if (alert.EndsAt.Sub(old.StartsAt) > 0 && alert.EndsAt.Sub(old.EndsAt) < 0) ||
```
warebot: Should these comparisons also be inclusive of "0" for ranges that have exact overlaps?

fabxc: As we are talking about nanoseconds in this case it doesn't really matter.
But I can add that later. Not a blocker for testing.

warebot: :thumbsup:

provider/boltmem/boltmem.go
```diff
@@ -0,0 +1,426 @@
...
+
+	"github.com/boltdb/bolt"
+	"github.com/prometheus/alertmanager/provider"
+	"github.com/prometheus/alertmanager/types"
+	"github.com/prometheus/common/log"
+	"github.com/prometheus/common/model"
+)
+
+var (
+	bktNotifies = []byte("notifies")
```
juliusv: Is there a chance we can name this "notifications", since "notifies" confuses me every time I read it? "notifies" is just a conjugation of the verb "to notify", but never a valid noun.

Just saying this now since this will be persisted and will thus be harder to change later on.

provider/boltmem/boltmem.go
```diff
@@ -0,0 +1,426 @@
...
+// They are not guaranteed to be in chronological order.
+func (a *Alerts) Subscribe() provider.AlertIterator {
+	var (
+		ch   = make(chan *types.Alert, 200)
+		done = make(chan struct{})
+	)
+	alerts, err := a.getPending()
+
+	i := a.next
+	a.next++
```
juliusv: Should this also be lock-protected? Or will the method never be called concurrently?

provider/boltmem/boltmem.go
```diff
@@ -0,0 +1,426 @@
...
+
+		for _, a := range alerts {
+			select {
+			case ch <- a:
+			case <-done:
+				return
+			}
+		}
+
+		<-done
```
juliusv: Here you have an extra `<-done` to wait with the cleanup until `done` is closed, not just if `a` is closed. But in the method below you don't. Is there a reason that makes it necessary to have this here?

juliusv: Ah, ok, now I get it. Because `ch` is still used to continuously send alerts even after the initial sweep is sent.

provider/boltmem/boltmem.go
juliusv: To me this'd be more grokkable if expressed using `Before` and `After`:

```go
if (alerts.EndsAt.After(old.StartsAt) && alert.EndsAt.Before(old.EndsAt)) || ...
```

provider/boltmem/boltmem.go
```diff
@@ -0,0 +1,426 @@
...
+				// Merge alerts if there is an overlap in activity range.
+				if (alert.EndsAt.Sub(old.StartsAt) > 0 && alert.EndsAt.Sub(old.EndsAt) < 0) ||
+					(alert.StartsAt.Sub(old.StartsAt) > 0 && alert.StartsAt.Sub(old.EndsAt) < 0) {
+					alert = old.Merge(alert)
+				}
+			}
+
+			ab, err := json.Marshal(alert)
+			if err != nil {
+				return fmt.Errorf("encoding alert failed :%s", err)
```
juliusv: The space should go after the colon.

provider/boltmem/boltmem.go
juliusv: (at least the "All methods are goroutine-safe." comment wouldn't hold true here)

provider/boltmem/boltmem.go
```diff
@@ -0,0 +1,426 @@
...
+	}
+
+	for _, sil := range sils {
+		if sil.Mutes(lset) {
+			s.mk.SetSilenced(lset.Fingerprint(), sil.ID)
+			return true
+		}
+	}
+
+	s.mk.SetSilenced(lset.Fingerprint())
```
juliusv: Out of scope for this PR, but I find this interface confusing. It does two different things based on whether there is a variadic arg (while the method name doesn't make that clear), but then it also only cares about the first variadic arg and ignores any further ones.

I think two separate methods would be easier to understand.

provider/boltmem/boltmem.go
```diff
@@ -0,0 +1,426 @@
...
+func (n *Notifies) Get(recv string, fps ...model.Fingerprint) ([]*types.NotifyInfo, error) {
+	var res []*types.NotifyInfo
+
+	err := n.db.View(func(tx *bolt.Tx) error {
+		b := tx.Bucket(bktNotifies)
+
+		for _, fp := range fps {
+			k := make([]byte, 8+len([]byte(recv)))
+			binary.BigEndian.PutUint64(k, uint64(fp))
+			copy(k[16:], []byte(recv))
```
juliusv: Why `k[16:]` and not `k[8:]`? Same for the method below.

provider/boltmem/boltmem.go
```diff
@@ -0,0 +1,426 @@
...
+	return sil, err
+}
+
+// Notifies provides information about pending and successful
+// notifications. All methods are goroutine-safe.
+type Notifies struct {
+	db *bolt.DB
+}
+
+func NewNotifies(path string) (*Notifies, error) {
```
juliusv: Missing doc strings here and below.

juliusv:
Did the go-fmting of the bindata.go and types.go make it into this PR by accident?

fabxc:
Yep.

provider/boltmem/boltmem.go
fabxc: Mistake.

provider/boltmem/boltmem.go
fabxc: Totally agree, will change that in a subsequent one.

fabxc:
Done.

.promu.yml
```diff
@@ -1,4 +1,4 @@
-go: 1.5.4
+go: 1.6.2
```
sdurrheimer: You will need to rebase your work given that this change and the one in `.travis.yml` are already in master.

provider/boltmem/boltmem.go
```diff
@@ -0,0 +1,432 @@
+package boltmem
```
sdurrheimer: Just in case, don't miss to add license's headers here.

juliusv:
@fabxc Cool - the accidental gofmt changes are still included though?

provider/boltmem/boltmem_test.go
```diff
@@ -0,0 +1,548 @@
+package boltmem
```
sdurrheimer: Same here, license's header.

fabxc:
Are we :+1: aside from these?

.promu.yml
fabxc: That's the master commit. It will merge fine.

juliusv:
Yep :+1:

mattbostock:
@fabxc: Can you please elaborate on the pull request description; does SQLite not satisfy the current feature set? Why was Bolt chosen?

PR #338, opened 2016-05-02
*: bump default Go version to 1.6.2

fabxc:


.travis.yml
```diff
@@ -2,7 +2,7 @@ sudo: false
 language: go
 
 go:
-- 1.5
+- 1.6.2
```
sdurrheimer: Wouldn't be better to continue to test against go `1.5` and `1.6` ? People might continue to build it by hand on `go 1.5`.

```
go:
- 1.5.4
- 1.6.2
```

fabxc: Ah, if that's possible. We can do that for a while.

fabxc: Done

sdurrheimer:
Is it normal the following files are always reformated by the make command:
- `template/internal/deftmpl/bindata.go`
- `ui/bindata.go`

fabxc:
Yea... timestamps always change.

PR #340, opened 2016-05-04
Exotel sms

sarathsp06:
Added **SMS** Notification (Using [exotel](http://exotel.in/)) support

How to enable sms alert ?
  Add the `sms_configs` in receivers

How sms is sent?
 SMS is sent using [exotel](http://exotel.in/) a cloud telephony solution

Structure of `sms_configs`
-----------------------------------
  ` from` : From number
   `to`     : Virtual Number To be used
  `account_sid` : exotel account name / sid 
  `account_token` : exotel account token
  `text`  : sms template 

Global config 
------------------
   `sms_account_sid` and `sms_account_token`


RomainBelorgey:
Hi,

Great pull request, but is it possible to make the url generic to send the SMS (inside the yaml config) ?
It will make the config a real SMSConfig (and not ExotelSMSConfig).

I'm also currently using Kannel to send SMS (http://www.kannel.org) which use a GET request and not POST. 
Example : http://smsbox.host.name:13013/cgi-bin/sendsms?username=foo&password=bar&to=0123456&text=Hello+world
And I would like to send alerts with Kannel.

I don't know if it's also possible to make the http request generic to respond to this or should I make a different config part (KannelConfig) ?

Thanks and I can help if it's needed to implement that.

brian-brazil:
Thanks for the PR.

I don't it's sane to support every possible SMS provider, and a SMS provider that works in one particular country isn't much use for the majority of our users who aren't in that country. Any SMS gateways we add support in the core to should be relatively popular internationally and intended to be reliable across all continents.

I'd recommend using the webhook notifier for this.

sarathsp06:
@brian-brazil . I do understand that , i was using exotel internally and it does not have global reach and popularity. Would [infobip](http://www.infobip.com) work ? .I found infobip to be highly reliable among most SMS Gateways

brian-brazil:
We'll probably need to discuss how we're going to approach this among the developers. I also had a private contact looking to add nexmo. There's likely quite a few players in this market.

marcelcorso:
I agree that there are many players, yes. I work for one (@messagebird) definitely interested in providing sms support for alertmanager and I'm still to do the pull request for Nexmo (we want to use something different than our own dog food for alerting).

The problem is that all providers implementations are slightly different but not small enough to allow a unified, generic code to conquer them all. 
There is a protocol though, SMPP, but I'm not sure if many providers have it as an option and it is not as simple as a POST request.

fabxc:
Without a widely used protocol we won't be able to have a generic SMS integration.
Since this seems to be interesting to quite a few people, how about we have a collection of programs that integrate with the Alertmanager webhook? This way people can still make use of each others work.

sarathsp06:
@fabxc That would be good, a framework for alertmanager webhook integration and applications made on top of that. I have added call and SMS integrations for Prometheus alerts for my use would be happy to  contribute on the framework and applications

StianOvrevage:
I did some research recently and it seems that Clickatell can be a good one-size-fits-all SMS provider ( https://www.clickatell.com ). They cover basically the whole world, there are not fixed monthly fees and the per-SMS pricing is competitive. I'm about to hack together an integration myself but the best would be to have a native alertmanager plugin for it.

fabxc:
The thing is, no matter how you, and maybe the rest of us, find this provider superior and clearly the one everyone should go with... someone will disagree. Providers change, they change prices or quality, competitors come up. As an open source project there no good way we can favor one commercial provider over others.
That's why these things need open protocols.

StianOvrevage:
I'm not going to start a fight, but you do support PagerDuty and Slack, which are commercial providers ;)

fabxc:
Okay, from that perspective my response was pretty BS :) – ignore and forgive please.
Probably I'm just lacking overview over the market. The integrations we have right now came very natural I suppose due to the wide adoption.

If sufficient demand is visible, we can probably rethink. But it has to go beyond one person/company finding it great.

StianOvrevage:
Forgiven!

I don't have a strong preference for any provider, and I suspect there are others like me. If there was one provider that was not overly expensive and worked in my country I would gladly just use that for Prometheus. SMS is a commodity so I would gladly use one provider for Prometheus and another one if I would ever need another SMS sending feature for some other system that did not support the same as Prometheus (but that will probably never happen).

Is there a way to poll the community of whether they care about SMS alerts, and second, which providers people use?

marcelcorso:
Hey! @samwierema took up on the idea of separating the sms notification stuff to another service and started it last Friday. 

https://github.com/marcelcorso/sachet

The idea is to allow users to configure different behaviour, sms provider per alertmanager receiver. 

On the "providers" section of the configuration file the user can add their favorite service access keys. 

The "provider" key on the receiver list selects one of the configured providers.

Implementing a provider should be easy:
1) Do a struct that implements "Provider" 
2) Instantiate it on the function "providerByName" 

Sam tested it yesterday and it send the first sms. 
I created a few issues of things that sure need to be done. Sure there is way more. 

what you guys think?

StianOvrevage:
We discussed this on the #prometheus IRC channel a few days ago and we think maybe the best and most flexible way is to extend the current webhook alert with JSON templates. That way it's possible to integrate towards any remote REST API endpoint just by writing a simple template for that provider. That would make it very simple to integrate not only SMS providers but any other remote REST services as well.

marcelcorso:
Sounds good.
For Nexmo json templates would do.
But in the case of sending an sms with MessageBird authentication goes on
the header and currently there is no option to add headers to webhooks.

StianOvrevage:
Yep, I know. The same for Clickatell. So adding Header-support to webhooks would be on the list as well. I'll see if we open another issue and keep track of things there and gather all the different SMS related issues there.

PR #347, opened 2016-05-10
[WIP] check availability of templates when loading the config

### LINKED ISSUE #311: using an unknown template makes AM flood your $alerttarget
##### Problem:
When specifying an unknown template at least to the "html"-field of a receiver makes AM flood the receiver's inbox with mails where the according field (in case of " html:" the mail body) is empty.

##### Steps to reproduce:
use e.g. `html: '{{ email.shinanigans.html }}'` in a receiver-block and provoke an alert for that receiver.

@fabxc

cherti:
### Work in progress, do not merge yet!

This is proof-of-concept-code to be discussed with upstream.
Fixes #311 when finished and merged.
@fabxc 

## Problem diagnosis
When sending out alerts, Prometheus tries sending the alert and if that fails with an error, it tries again with exponential backoff. Unfortunately, an error that can cause this is, besides others, the use of a nonexistent template, which (at least for mail) results in AM flooding the receivers inbox by repeatedly trying (and succeeding) sending mails. This has primarily been tested for mail, but might occur as well for other notification types.

## Basic idea for fixing
There are two options to fix this: 
1) fall back onto the default template when a template is not found
2) Check upon loading of the config that all templates in use are actually found

in the current structure, `1)` would require internal silencing of errors due to the mechanics of the used notification module when falling back onto the default, which could be done, but is highly unelegant.
So we went for `2)` and verify the integrity of the entire config set by checking it upon loading and make config-loading throw an error for inconsistent configurations.

## State of this PR
This PR implements a general Proof-of-concept of fixing the issue mentioned above to be discussed with upstream, to find the optimal solution for this before carrying out work that might end up being unnecessary and wasting time. For this reason, currently only the `html`-field of the mail-configs is checked for invalid templates. For other notification methods the code will be fairly similar.

## Further plans to be discussed besides the obvious

* Currently the significant part of the fix lives directly in `main.go`, it might be a reasonable choice to move it into the `config`-package and just call that package from within `main.go`.

Feedback is highly appreciated so that we can finish fixing this issue. :)

Shoutout to @joker234 with whom I debugged and fixed this!

### DIFF
--- a/main.go
+++ b/main.go
@@ -169,6 +169,36 @@ func main() {
 		}
 		tmpl.ExternalURL = amURL
 
+
+
+		for _, recv := range conf.Receivers {
+			for _, ec := range recv.EmailConfigs {
+				if ec.HTML != config.DefaultEmailConfig.HTML {
+					if err := tmpl.CheckTemplateExistence("html", ec.HTML); err != nil {
+						return err
+					}
+				}
+			}
+//			for pc := range recv.PagerdutyConfigs {
+//
+//			}
+//			for hc := range recv.HipchatConfigs {
+//
+//			}
+//			for sc := range recv.SlackConfigs {
+//
+//			}
+//			for wc := range recv.WebhookConfigs {
+//
+//			}
+//			for ogc := range recv.OpsGenieConfigs {
+//
+//			}
+//			for pc := range recv.PushoverConfigs {
+//
+//			}
+		}
+
 		disp.Stop()
 
 		inhibitor = NewInhibitor(alerts, conf.InhibitRules, marker)
--- a/template/template.go
+++ b/template/template.go
@@ -15,6 +15,7 @@ package template
 
 import (
 	"bytes"
+	"fmt"
 	"net/url"
 	"path/filepath"
 	"sort"
@@ -36,6 +37,28 @@ type Template struct {
 	html *tmplhtml.Template
 
 	ExternalURL *url.URL
+}
+
+// checkTemplateExistence checks if a given template name exists for a given
+// type in a prometheus-template-struct
+func (tmpl *Template) CheckTemplateExistence(tmplType string, name string) error {
+	switch tmplType {
+	case "html":
+		if t := tmpl.html.Lookup(name); t == nil {
+			return fmt.Errorf("Template '%s' not found", name)
+		} else {
+			return nil
+		}
+
+	case "text":
+		if t := tmpl.html.Lookup(name); t == nil {
+			return fmt.Errorf("Template '%s' not found", name)
+		} else {
+			return nil
+		}
+	default:
+		return fmt.Errorf("There are no Templates of type '%s'", tmplType)
+	}
 }
 
 // FromGlobs calls ParseGlob on all path globs provided and returns the


brian-brazil:
Thanks for this. What I was thinking is that we'd add a new `Template` type in the config that'd compile the template at load time (similar to how we compile regexes). This will automatically work for all new fields, the only ones we need to watch are the ones which can be either text or HTML.

We'll still need 1), as there's various runtime ways a template fails.

cherti:
So you say upon loading the config your idea will expand the templates immediately, so AM won't actually expand the template upon sending the alert?

brian-brazil:
You can only expand the template when sending it, however we can compile the template at any point before that. That'll catch basic syntax errors early.

cherti:
but wouldn't account for using template {{ foo.bar }} whereas the only template I ever define is {{ foo.baz }} (althoug foo.baz is syntactically correct), right? that would still flood the receiver in the current situation, from what I understand. We would still have to account for that problem, right?

brian-brazil:
That sort of error can only be caught at expansion time as far as I'm aware.

cherti:
ok, so we still should do this for all fields right at the start imho, to voice that upon start.

brian-brazil:
Yes, doing the compiling early will catch the more obvious problems.

cherti:
What about the general approach we took here? can be considered solid for that purpose?

brian-brazil:
I don't think the approach here will scale. I'd suggest making a new string type for templates and compiling them at the UnmarshallYaml stage.

cherti:
I don't think that compiling is the issue here. The problem is that the templates are parsed after the main config. So when we compile the config, we currently cannot know which templates will exist. So we have to verify that all necessary templates can be found afterwards.
This is kinda unavoidable, because the templates are specified within the config, so we necessarily need to parse that first to actually know where to look for templates. And as the config is parsed by yaml.v2, I don't think that we can interfere there reasonably to parse the templates before parsing the receiver-section of the main config-file.

This leaves us with matching templates and receivers after the templates are loaded (which happens after the receiver's section is loaded). Whether they are compiled or not doesn't really matter, because what we need to know is if we can expand all templates specified somewhere within the receiver-section to avoid the mentioned problem.

brian-brazil:
> So when we compile the config, we currently cannot know which templates will exist. So we have to verify that all necessary templates can be found afterwards.

That's merely one way in which templates can not work at runtime. I don't think we should be trying to do verification beyond what Go makes easy for us, as that'll be a lot of effort.

cherti:
I don't think that the alternative of silencing at runtime is much more attractive. And due to the structure to github.com/cenkalti/backoff the only thing we can do is log the error, but not notify anyone about it. Not mentioning that this way the error is detected only if there is an alert routing to that receiver, and when there is an alert, the template is broken. I don't think it is reasonable to detect configuration errors in a monitoring system when the error actually occurs while we have the possibility to detect that very error immediately upon reloading the configuration.
(This also assumes implicitly that the default template is always an appropriate. Which might be a fair assumption, but I find that still a little unsatisfactory to rely on that if it is avoidable. That would be the only alternative to checking prior, though, without rewriting notification internals, quite a bit, as far as I can see.)

The code for checking for configuration errors is a little verbose, that's true, but it's pretty straightforward. Just loop over the receiver-section and check all fields where a template can occur in contrast to internally silencing an error and falling back onto the default template (because that's the only thing we can do once the alertmanager is running).

Once we verified that we have templates for all receivers do have a matching template, we can do the rest of verification within the Template-Parsing-Code, even when unmarshalling or directly after, e.g. compiling the templates immediately.

brian-brazil:
> The code for checking for configuration errors is a little verbose, that's true, but it's pretty straightforward. Just loop over the receiver-section and check all fields where a template can occur

It's not that straightforward. You'd need to walk to parse tree to find all template references, and then cross-check them. I don't believe the Go template library offers the APIs to do this.

cherti:
actually it's a simple call to `tmpl.{text,html}.Lookup(templatename)`. This either returns our template or `nil`.

brian-brazil:
How are you getting the name of the template?

cherti:
just accessing the templatable field. The name is provided there.

brian-brazil:
That value is the content of a template, not a template name.

cherti:
ok, I'm not familiar with the terminology here. But it works fine, handing the field's content over to `Lookup` detects if the template is nonexisting and throws the patched alertmanager throws an error "cannot find template".

brian-brazil:
I'm pretty sure this won't work. Have you tested this with a valid template end-to-end?

cherti:
I think so, but I can verify that. Not today, though, if you don't mind. :)

cherti:
Yes, you've been right, Lookup is not sufficient indeed. I'm thinking about how to verify a templates' existence currently. I think that most of the obvious errors could indeed be caught with your compiling-idea, but simple typos in the config we cannot (the qeustion of existence).

However, I still think that this is desirable to report those upon loading the config, because it's something that can easily happen (and with a bigger config, one might not see it immediately). I'm thinking about if it would be a possibility to test-expand the templates once and see if that goes through. What do you think about that idea?

brian-brazil:
> I'm thinking about if it would be a possibility to test-expand the templates once and see if that goes through. What do you think about that idea?

I've pondered that and it's an interesting usability vs. correctness tradeoff. It's possible that the user is making assumptions about the nature of the alerts they'll be getting (e.g. always having a "severity" label) which generally work in their setup, but that we obviously don't know when doing a test expansion. Should be be forcing users to be more correct in their templates for cases they'll never hit which will be annoying, or leave them to find out at runtime for cases they do hit?

cherti:
> leave them to find out at runtime for cases they do hit?

When an alert hits my {inbox,pushover,$channel}, I'd personally not be confronted with the realization that my alertmanager is misconfigured in *that* situation. So I'd go for forcing correctness upon loading the config.

I'd be more relaxed with other kinds of software, but especially with montioring you usually have contact with that once something went wrong. So I'd personally prefer to have dealt with those issues before an alert, if we can, because people rely on their monitoring to behave as they expect. Especially because monitoring is one of the components that shall notify you if that is not the case somewhere else.

brian-brazil:
The challenge is that the template language is quite powerful, so we can't catch every potential failure at compile time. I think catching the more blatant problems at compile time such as syntax errors is enough, and then 1) for the rest.

cherti:
in that case [this](https://github.com/prometheus/alertmanager/blob/master/notify/notify.go#L192) is the place (or rather function) to fix this. I haven't yet traced down *which* notify that actually is, but that notify must either fall back onto the default templates internally or we have to check if the error is of the kind that we can fall back and then restart the notification process with a default template. The former is probably the more desirable way, although I haven't looked into that part of the code.

cherti:
Not a feasible solution for the referenced bug, therefore closing this PR and updating the issue with the information found.

PR #358, opened 2016-05-27
provider/boltmem: make alerts purely in-memory.

fabxc:
Initial testing has shown BoltDB in plain usage to be a bottleneck
at a few thousand alerts or more (especially JSON decoding.
This commit actually makes them purely memory.

@beorn7

### DIFF
--- a/provider/boltmem/boltmem.go
+++ b/provider/boltmem/boltmem.go
@@ -16,9 +16,9 @@ package boltmem
 import (
 	"encoding/binary"
 	"encoding/json"
-	"fmt"
 	"path/filepath"
 	"sync"
+	"time"
 
 	"github.com/boltdb/bolt"
 	"github.com/prometheus/alertmanager/provider"
@@ -30,38 +30,59 @@ import (
 var (
 	bktNotificationInfo = []byte("notification_info")
 	bktSilences         = []byte("silences")
-	bktAlerts           = []byte("alerts")
+	// bktAlerts           = []byte("alerts")
 )
 
 // Alerts gives access to a set of alerts. All methods are goroutine-safe.
 type Alerts struct {
-	db *bolt.DB
+	mtx    sync.RWMutex
+	alerts map[model.Fingerprint]*types.Alert
+	stopGC chan struct{}
 
-	mtx       sync.RWMutex
 	listeners map[int]chan *types.Alert
 	next      int
 }
 
 // NewAlerts returns a new alert provider.
 func NewAlerts(path string) (*Alerts, error) {
-	db, err := bolt.Open(filepath.Join(path, "alerts.db"), 0666, nil)
-	if err != nil {
-		return nil, err
-	}
-	err = db.Update(func(tx *bolt.Tx) error {
-		_, err := tx.CreateBucketIfNotExists(bktAlerts)
-		return err
-	})
-	return &Alerts{
-		db:        db,
+	a := &Alerts{
+		alerts:    map[model.Fingerprint]*types.Alert{},
+		stopGC:    make(chan struct{}),
 		listeners: map[int]chan *types.Alert{},
 		next:      0,
-	}, err
+	}
+	go a.runGC()
+
+	return a, nil
+}
+
+func (a *Alerts) runGC() {
+	for {
+		select {
+		case <-a.stopGC:
+			return
+		case <-time.After(30 * time.Minute):
+		}
+
+		a.mtx.Lock()
+
+		for fp, alert := range a.alerts {
+			// As we don't persist alerts, we no longer consider them after
+			// they are resolved. Alerts waiting for resolved notifications are
+			// held in memory in aggregation groups redundantly.
+			if alert.EndsAt.Before(time.Now()) {
+				delete(a.alerts, fp)
+			}
+		}
+
+		a.mtx.Unlock()
+	}
 }
 
 // Close the alert provider.
 func (a *Alerts) Close() error {
-	return a.db.Close()
+	close(a.stopGC)
+	return nil
 }
 
 // Subscribe returns an iterator over active alerts that have not been
@@ -128,42 +149,28 @@ func (a *Alerts) GetPending() provider.AlertIterator {
 }
 
 func (a *Alerts) getPending() ([]*types.Alert, error) {
-	var alerts []*types.Alert
+	a.mtx.RLock()
+	defer a.mtx.RUnlock()
 
-	err := a.db.View(func(tx *bolt.Tx) error {
-		b := tx.Bucket(bktAlerts)
-		c := b.Cursor()
+	res := make([]*types.Alert, 0, len(a.alerts))
 
-		for k, v := c.First(); k != nil; k, v = c.Next() {
-			var a types.Alert
-			if err := json.Unmarshal(v, &a); err != nil {
-				return err
-			}
-			alerts = append(alerts, &a)
-		}
+	for _, alert := range a.alerts {
+		res = append(res, alert)
+	}
 
-		return nil
-	})
-	return alerts, err
+	return res, nil
 }
 
 // Get returns the alert for a given fingerprint.
 func (a *Alerts) Get(fp model.Fingerprint) (*types.Alert, error) {
-	var alert types.Alert
-	err := a.db.View(func(tx *bolt.Tx) error {
-		b := tx.Bucket(bktAlerts)
+	a.mtx.RLock()
+	defer a.mtx.RUnlock()
 
-		fpb := make([]byte, 8)
-		binary.BigEndian.PutUint64(fpb, uint64(fp))
-
-		ab := b.Get(fpb)
-		if ab == nil {
-			return provider.ErrNotFound
-		}
-
-		return json.Unmarshal(ab, &alert)
-	})
-	return &alert, err
+	alert, ok := a.alerts[fp]
+	if !ok {
+		return nil, provider.ErrNotFound
+	}
+	return alert, nil
 }
 
 // Put adds the given alert to the set.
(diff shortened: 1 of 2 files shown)


provider/boltmem/boltmem.go
```diff
@@ -30,38 +30,59 @@ import (
 var (
 	bktNotificationInfo = []byte("notification_info")
 	bktSilences         = []byte("silences")
-	bktAlerts           = []byte("alerts")
+	// bktAlerts           = []byte("alerts")
```
juliusv: Remove commented-out line?

juliusv:
:+1: otherwise, but tests are still failing?

fabxc:
Yea, this was for Beorn's visbility. This reduces the feature set so some tests fail.

beorn7:
OK, I got a vague understanding (but not enough for a detailed review). 

Strangely, only the Go1.6 test fail, not Go1.5. Related to vendoring (that might need updates)?

What's the plan now? Merge this and then test it in production? I could re-create that mirrored setup here at SC to test it with a real production load.

mattbostock:
Definitely agree to optimising this, we've seen very high CPU usage in AlertManager v0.1.1 (using SQLite) due to JSON unmarshalling in `getPending()`:

screen shot 2016-06-08 at 13 16 06

That said, it'd be a shame to lose alerts between application restarts. Instead, could we use [gob](https://golang.org/pkg/encoding/gob/) to serialise the data and store the gobs in BoltDB?

fabxc:
The issue is not serialization per se. The O(n^2) complexity of inhibition is, which is a naive implementation we haven't gotten around to improve. No matter how fast your serialization, this will be an issue. It even is for memory, but it becomes apparent only a fair bit later.

So in-memory would buy us a bit more time once more but not solve the fundamental problem.

fabxc:
Right now there's no way to access historic alerts by the way. And active alerts will reappear within 0-60 seconds. So at the current state, in-memory vs persisted is not really different for the user.

mattbostock:
@fabxc So there are other issues, but the JSON serialisation is a significant one. It seems like the complexity of inhibition is a separate issue to the one being addressed by this pull request.

What would make active alerts reappear? My only concern with not persisting alerts is this scenario:

- Prometheus sends an alert to AlertManager
- AlertManager sends notification
- AlertManager restarted, alerts wiped because they're in-memory
- AlertManager no longer has a record of the alert that Prometheus sent, so doesn't know to re-alert based on the `group_interval`.

Admittedly, resending alerts after `group_interval` is perhaps not that important but it would be good to understand what happens in that scenario so that the tradeoffs are clear.

beorn7:
Prometheus re-sends alerts to Alertmanager all the time. It's more the other way round: If Alertmanager restarts, it will re-send notifications because it has no notion of having it sent already.

fabxc:
It does and still would as that info is persisted.

mattbostock:
I see, thanks both.

mattbostock:
What needs to happen to get this merged? I'm having to regularly truncate the alerts table to prevent AlertManager from using excessive CPU time (request time for 'list_alerts' handler was taking as long as 160s in the 90th percentile in version 0.1.1).

fabxc:
@mattbostock going back to the original problem, I just filed #379 – would you mind giving that branch a go? If that makes things smoother this is the change we'd actually want to release ASAP.

mattbostock:
@fabxc We were seeing the CPU issues before any inhibition rules were configured, so I don't think #379 will help in our use case. I would expect that this PR will, however.

fabxc:
Your flame graph above indicates that it is all coming from inhibition.
The Inhibitor is acutally scanning all alerts even if there are no
inhibition rules configured, i.e. an early return is missing – so the
problem doesn't go away with 0 inhibition rules as one would think.

https://github.com/prometheus/alertmanager/blob/master/inhibit.go#L49-L79

I'd appreciate this to be tested. If it doesn't help I promise we'll move
forward with this one quickly.

mattbostock:
@fabxc Ah, thank you, that makes more sense to me now. I'll try out #379 ASAP and get back to you.

mattbostock:
Early indications are that #379 is a very significant improvement - we're seeing request times for the 'list_alerts' handler in the region of 200-250ms versus > 60s previously. Thanks @fabxc!

fabxc:
Happy to hear that! That change is part of the latest Alertmanager 0.2.0
release.

fabxc:
Closing for now. Will revisit the general idea for the HA mode.

mattbostock:
Since upgrading to version 0.2.0, we're seeing response time for the 'list_alerts' handler slowly creep up as new alerts fire (from ~100ms to ~500ms in the 90th percentile).

Here's a new flamegraph over a 30 second period, which shows that a substantial amount of CPU time is spent unmarshaling JSON:

screen shot 2016-06-20 at 18 59 54

Could you reconsider merging this PR, or would you accept a PR that uses another serialisation format other than JSON, e.g. gob?

fabxc:
Hey again. So that luckily still seems heaps better than before the inhibition fix, right?
So now we are running into the next case of silences (instead of alerts as before) being read from persistence on every notification attempt.

Here is a benchmark of different serialization formats: https://github.com/alecthomas/go_serialization_benchmarks

As we see gob is indeed about 4x better than JSON. However, no matter how good the serialization format, it's only a constant factor and the scaling will not change. So if it makes your handler slow down to 1s/req now, with 10x the silences it will be 10s/req. 
Picking a better serialization format will reduce these to 250ms and 2.5s respectively – in total we don't solve the problem.

For now, let me add an in-memory cache in front of our persisted silences. This will by far be the best. For future scaling we should look into some smarter mechanisms.

For the record, I totally agree that JSON is not the ideal persistence format and will consider alternatives for the persistence of the HA Alertmanager.

Also found a bug – accidentally decoding the JSON silences twice certainly doesn't help your situation right now ;)

mattbostock:
Indeed, much better than before.

The cache makes sense - so you'd like to continue to persist alerts in the database? What's the reason for persisting them rather than just keeping them in-memory given the reasons that @beorn7 cites above?

fabxc:
I'm writing an HA Alertmanager currently, which will have semantical differences in the storage. It will indeed be purely in-memory by default as it will fetch back state from the other peers if it's restarted.
It will have some periodic snapshotting for persistence in case you still want to run it as a single instance.

The 0.2.0 release was cut mainly for the issues you were encountering. I want people to have something useable without me having to rush the HA stuff.
For that purpose I didn't want to make too many deep changes. The in-memory only part specifically has implications for the user – minor but these things always come with a risk.

PR #397 should alleviate your problem and I'll cut 0.2.1 with it tomorrow.

mattbostock:
@fabxc Thanks! I'll put #397 in Staging and see how it compares.

mattbostock:
@fabxc, when you say [above](https://github.com/prometheus/alertmanager/pull/358#issuecomment-225362892) that notification information is persisted, I presume that's because it's persisted separately to alerts in the `notification_info` bucket?

Unfortunately #397 doesn't seem to be helping much in my case in our Staging environment. Here's another flamegraph taken from that instance (version 0.1.1-114-g8dc2f97):

screen shot 2016-06-27 at 14 04 37

If I'm reading the graph correctly, I believe the changes in this pull request would help alleviate the issue? I'm currently seeing request times of around 1.2s in the 90th percentile for the 'list_alerts' handler.

I'll test this PR in Staging and see how it compares.

fabxc:
Okay, yet another flame graph :) Thanks for keeping us updated.

Just to summarize:
* first iteration we fixed the inhibition, which does the same walk through all existing alerts for each alert we notify about. That was really bad and happens way more often than opening the alert listing page. This caused load times of 60s+ because of Alertmanager was very busy in the background
* second flamegraph than showed a similar load from sending notifications due to walking all silences for sent alerts.

Both of these are gone in the last flame graph.
I hoped reducing the background load was sufficient but as we see now, the load the UI alone is also pretty slow.

Taking alerts into memory or adding a cache as I did with silences would indeed help.

The thing I'm happy with is that 1.2s is still bearable compared to before and more importantly that the notification path, which is more critical, is no longer affected.

I'm currently working on the highly available Alertmanager, which will have a different storage due to some fundamental differences. Part of that will be that everything will be in memory by default.
So my question to you is, how badly is this affecting your workflow? I can fix the issue here and do another release or spend the time to get the HA Alertmanager out of the door (alpha/beta) faster. If it's not too troubling, I'd like to opt for that of course.

mattbostock:
I see you agree that flamegraphs are awesome ;-)

Thanks for confirming.

1.2s is a definite improvement over before, and as you say, we could just increase our timeouts. 1.2s is still a relatively long time for a HTTP request to be fulfilled for a Production service and while the HA feature sounds great, I imagine it's probably going to be a while before we can run it Production?

We can live with this as is, though if you'd be happy merging this PR without necessarily cutting a new release, that'd help us avoid having to patch.

mattbostock:
Follow-up for completeness: after several days running in a Staging environment with Production load, AlertManager 0.2.1 with the changes in this PR has a 5ms 90th percentile request time for the 'list_alerts' handler, versus around 1.9 seconds with AlertManager 0.2.0.

Previous testing of the silence cache (#397) had a negligible effect so I think the speedup is attributable to the changes in this PR.

fabxc:
ACK, a cache would arguably be better so we don't lose any state at all on shutdown. But that would require a different set of changes now, eating up time.

I discussed with @brian-brazil and the effects of no alert persistence are pretty minor. So we can release a 0.3.0 with this change.

mattbostock:
@fabxc Great, thanks. We're been running this patch in Staging sending (almost) the same notifications as our Production instance (not actually paging anyone) and the functionality is empirically the same.

PR #359, opened 2016-05-30
Add support for NATS as a notifier.

diogogmt:
Initial PR to add support for new NATS notifier.

NATS is a simple and fast message queue: http://nats.io/

@brian-brazil I have a couple of questions regarding adding  a new notifier:

* The WebHook notifier defines a WebhookMessage struct with a version and a group key, does the NATS notifier needs to define something similar?
* Should the VSendResolved flag be true or false for the DefaultNatsConfig?

**Please ignore the printf's since I'm still testing the code, I'm opening the PR to get some initial feedback and make sure I'm on the right track.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -406,6 +406,7 @@ type Receiver struct {
 	WebhookConfigs   []*WebhookConfig   `yaml:"webhook_configs,omitempty"`
 	OpsGenieConfigs  []*OpsGenieConfig  `yaml:"opsgenie_configs,omitempty"`
 	PushoverConfigs  []*PushoverConfig  `yaml:"pushover_configs,omitempty"`
+	NatsConfigs   	 []*NatsConfig      `yaml:"nats_configs,omitempty"`
 
 	// Catches all undefined fields and must be empty after parsing.
 	XXX map[string]interface{} `yaml:",inline"`
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -104,6 +104,13 @@ var (
 		Retry:    duration(1 * time.Minute),
 		Expire:   duration(1 * time.Hour),
 	}
+
+	// DefaultNatsConfig defines default values for NATS configurations.
+	DefaultNatsConfig = NatsConfig{
+		NotifierConfig: NotifierConfig{
+			VSendResolved: false,
+		},
+	}
 )
 
 // NotifierConfig contains base options common across all notifier configurations.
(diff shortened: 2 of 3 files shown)


brian-brazil:
Hmm, I think for something like this the webhook receiver is more appropriate. There's many many queuing systems out there and I don't think it's sane to start on the path of supporting them all. We've already decided against SMS notifiers on the same grounds, and there's a stronger argument there as they're more alerty. It's also likely there's some pre-processing you'd want to do before putting notifications on the queue.

diogogmt:
@brian-brazil 

The use case to add support for NATS  I would argue is a bit different than SMS, for example; With SMS in most cases an operator wants to be notified if an alert is triggered, the key is that the message is being delivered to a person, the same I would argue is the use case for  PagerDuty and other notification services where the goal is to get a person to acknowledge and act on a given alert.

Now the use case for NATS is much more similar to the one for the WebHook where a third party application is integrating with prometheus and processing its alerts through its own notification engine.

To give you a bit of background on the use case that I'm coming from:
We have an application composed of several microservices that use NATS to power the messaging layer instead HTTP for inter service communication.
Our application provides a platform to manage containers in production and uses prometheus to power the policy engine that allows users to manage SLA auto scale policies.

We need to be notified by the prometheus alert manager when an alert is triggered and instead of building a service to map the HTTP request sent by the alert manager web hook we figured that since the alert manager has a plugin architecture to support different notifiers we would add NATS support as a native notifier directly on the alert manager itself removing the need for adding an extra layer in our application.

The NATS protocol doesn't require that the message be pre-processed. The same JSON object that is sent in the WebHook can be published to NATS and it is up to the topic subscriber to make sense of the message and take an action accordingly.

IMHO adding support to NATS can really simplify the integration effort for other apps the same way I'm sure having added support for PagerDuty, OpsGenie and PushOver did.

brian-brazil:
The thing is that NATS is just one queuing system, there's also Kafka, RabbitMQ, ActiveMQ, MQTT, AMQP and many others. We've chosen to have the webhook receiver handle use cases which are too varied and custom to support directly, rather than having the alertmanager end up with 10s to 100s of integrations of this type to maintain.

> Our application provides a platform to manage containers in production and uses prometheus to power the policy engine that allows users to manage SLA auto scale policies.

Be very careful in how you do this, control systems are hard. I would strongly recommend not having the alertmanager as part of this control path.

juliusv:
Thanks for this PR, but I will have to unfortunately agree with @brian-brazil that it doesn't make sense to add (and support, document, maintain, ...) direct interfaces to arbitrary queueing systems in the Alertmanager. For this kind of generic use case, please use the webhook notifier instead.

Closing this.

PR #365, opened 2016-06-03
[WIP] Alertmanager HA mode

fabxc:
This adds an HA mode for Alertmanager based on the [weaveworks/mesh](https://github.com/weaveworks/mesh) library, which implements a gossip protocol. This means we can run any number of Alertmanagers and remain available for writes under partitions. It's eventually consistent for notification and silence state.

It requires Prometheus servers to dispatch alerts to all Alertmanagers, which is implemented in prometheus/prometheus#1699. Consistent history of alerts is explicitly not a goal.

This is still lacking tests but contains a `Procfile` for manual testing using contents of `examples/ha`.
PRd here for visibility for anyone interested.

TODOs:
- [x] Unit testing
- [ ] Functional testing
- [x] Retention for all data (could GC be conveniently embedded in the gossip merge procedure?)

@peterbourgon there are some questions regarding mesh I want to resolve before considering this ready but you might want to take a look already.

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/alertmanager/365)


Procfile
```diff
@@ -0,0 +1,5 @@
+a1: ./alertmanager -log.level=debug -storage.path=$TMPDIR/a1 -web.listen-address=:9093 -mesh.hardware-address=00:00:00:00:00:01 -mesh.nickname=a -mesh.listen-address=:8001 -config.file=examples/ha/alertmanager.yaml
+a2: ./alertmanager -log.level=debug -storage.path=$TMPDIR/a2 -web.listen-address=:9094 -mesh.hardware-address=00:00:00:00:00:02 -mesh.nickname=b -mesh.listen-address=:8002 -mesh.peer=127.0.0.1:8001 -config.file=examples/ha/alertmanager.yaml
+a3: ./alertmanager -log.level=debug -storage.path=$TMPDIR/a3 -web.listen-address=:9095 -mesh.hardware-address=00:00:00:00:00:03 -mesh.nickname=c -mesh.listen-address=:8003 -mesh.peer=127.0.0.1:8001 -config.file=examples/ha/alertmanager.yaml
+wh: ./examples/ha/webhook.py
+
```
beorn7: This file probably shouldn't be checked in. But it is convenient for me to create a test setup... ;)

beorn7: But perhaps `examples/ha/alertmanager.yaml` should be checked in?

beorn7:
Might just be a UI glitch, but all alerts are always displayed as "silenced" in my setup.

beorn7:
And expired silences shouldn't have a "delete" button if that just expires them again... :)

beorn7:
Caveat: My understanding of AM internals has suffered a lot ever since I did my last review (which was when we merged @fabxc 's rewrite into master), and even then, I was more concerned with other things…

Having said that, this seems to me to be a simple and pragmatic solution that's "good enough" to work with. However, to be of any real help and comment on semantic intricacies, I need a better understanding. Would love to have a Q&A session with whiteboarding of some "what-ifs" in person with @fabxc and @peterbourgon, similar to what we did when we first discussed HA issues.

Procfile
fabxc: Yes, it's useful – test setups is what people checking out the repo rather than downloading the distribution will want to do. Which is why it should be checked in. The config file as well.

fabxc:
It needs UI changes to match some semantic differences. But it's a robust way to do what we want, not an intermediate solution while searching for something better.

Procfile
beorn7: Perhaps document then how to use the Procfile. For me, it was only ever a Heroku thing, but there are apparently other tools consuming it, too. I don't think it can be assumed as generally known, though.

fabxc:
Yes, that needs better documenting. In this case the Procfile is meant to be run with goreman.
The silencing is probably a bug – either backend or UI.

examples/ha/webhook.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python
+# Reflects the requests from HTTP methods GET, POST, PUT, and DELETE
+# Written by Nathan Hamiel (2010)
```
brian-brazil: Is there a license on this?

examples/ha/webhook.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python
```
brian-brazil: This doesn't seem ha specific, so can go up a directory.

main.go
```diff
@@ -211,21 +230,78 @@ func main() {
...
-func extURL(s string) (*url.URL, error) {
-	if s == "" {
+type peerDescSlice []mesh.PeerDescription
+
+func (s peerDescSlice) Len() int           { return len(s) }
+func (s peerDescSlice) Less(i, j int) bool { return s[i].UID < s[j].UID }
+func (s peerDescSlice) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
+
+// meshWait returns a function that inspects the current peer state and returns
+// a duration of one base timeout for each  peer with a higher ID than ourselves.
```
brian-brazil: extra space after each

main.go
```diff
@@ -211,21 +230,78 @@ func main() {
...
+		sort.Sort(peers)
+
+		k := 0
+		for _, desc := range peers {
+			if desc.Self {
+				break
+			}
+			k++
+		}
+		log.Warnf("timeout multiplier: %d", k)
```
brian-brazil: This is Info I'd say

main.go
```diff
@@ -211,21 +230,78 @@ func main() {
...
+		log.Fatalf("mesh address: %s: %v", addr, err)
+	}
+	port, err := strconv.Atoi(portStr)
+	if err != nil {
+		log.Fatalf("mesh address: %s: %v", addr, err)
+	}
+
+	name, err := mesh.PeerNameFromString(hwaddr)
+	if err != nil {
+		log.Fatalf("%s: %v", hwaddr, err)
```
brian-brazil: It could be clearer what's wrong here, I'd not expect everyone to recognise a MAC address

main.go
```diff
@@ -239,8 +315,49 @@ func extURL(s string) (*url.URL, error) {
...
+	slice := make([]string, 0, len(ss))
+	for k := range ss {
+		slice = append(slice, k)
+	}
+	sort.Strings(slice)
+	return slice
+}
+
+func mustHardwareAddr() string {
+	ifaces, err := net.Interfaces()
```
brian-brazil: Are there any protections against two AMs having the same MAC address?

MAC addresses only need to be unique on a segment, which cloud computing providers may be taking advantage of.

provider/boltmem/boltmem_test.go
```diff
@@ -123,7 +123,7 @@ func TestNotifiesSet(t *testing.T) {
 func TestSilencesSet(t *testing.T) {
 	var (
 		t0 = time.Now()
-		t1 = t0.Add(10 * time.Minute)
+		// t1 = t0.Add(10 * time.Minute)
```
brian-brazil: debug code?

provider/mesh/peer.go
```diff
@@ -0,0 +1,241 @@
...
+
+	"github.com/prometheus/alertmanager/provider"
+	"github.com/prometheus/alertmanager/types"
+	"github.com/prometheus/common/log"
+	"github.com/prometheus/common/model"
+	"github.com/satori/go.uuid"
+	"github.com/weaveworks/mesh"
+)
+
+type NotificationInfos struct {
```
brian-brazil: `NotificationsInfo` would be better, as this is information about multiple notifications rather than many pieces of information about one notification.

provider/mesh/peer.go
```diff
@@ -0,0 +1,241 @@
...
+
+func (s *Silences) Run(retention time.Duration) {
+	s.st.run(retention)
+}
+
+func (s *Silences) Stop() {
+	s.st.stop()
+}
+
+func (s *Silences) Mutes(lset model.LabelSet) bool {
```
brian-brazil: Based on the Muter interface description, I wouldn't expect this function to be mutating.

examples/ha/webhook.py
fabxc: Just slapped in the first thing I could find Googling – will replace this.

main.go
fabxc: Not even that – Debug at most. Testing leftover.

main.go
fabxc: Talked to @peterbourgon about this. He was reasonably confident that it's fine – but in doubt the user can overwrite it with something custom and unique.

provider/mesh/peer.go
fabxc: By that it would rather be `NotificationsInfos` – I'd love to have generally better naming for the whole area but no ideas so far.

main.go
brian-brazil: I think it's worth adding a sanity check for at least, I don't trust all cloud providers and network abstractions to have globally unique MAC addresses.

main.go
brian-brazil: I think we should log it at startup, it's an important thing to know.

provider/mesh/state.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+	if !b.StartsAt.Equal(a.StartsAt) {
+		if a.StartsAt.Before(now) {
+			return fmt.Errorf("start time of active silence must not be modified")
+		}
+		if b.StartsAt.Before(now) {
+			return fmt.Errorf("start time cannot be moved into the past")
+		}
+	}
+	if a.EndsAt.Before(now) {
```
brian-brazil: Shouldn't there be a test here that the end time has changed?

provider/mesh/state.go
```diff
@@ -0,0 +1,327 @@
...
+
+	return st.mergeComplete(o.m)
+}
+
+func (st *silenceState) mergeComplete(set map[uuid.UUID]*types.Silence) *silenceState {
+	st.mtx.Lock()
+	defer st.mtx.Unlock()
+
+	for k, v := range set {
+		if prev, ok := st.m[k]; !ok || prev.UpdatedAt.Before(v.UpdatedAt) {
```
brian-brazil: We should ensure it's clearly documented that silences are last writer wins.

brian-brazil:
That generally seems okay.

main.go
fabxc: As its a mesh peers are constantly coming and going so this might change at runtime.

brian-brazil: We should probably make it a metric then.

provider/mesh/state.go
```diff
@@ -0,0 +1,327 @@
...
+			// From b being valid it follows that EndsAt will also be
+			// in the future.
+			return fmt.Errorf("new silence may not start in the past")
+		}
+		return nil
+	}
+	if a.ID != b.ID {
+		return fmt.Errorf("IDs do not match")
+	}
+	if !b.StartsAt.Equal(a.StartsAt) {
```
beorn7: This runs into a precision problem (start time parsed from the UI will almost be slightly different from the start time in the silence itsilf).

fabxc:
* [ ] silence delete should actually delete unstarted silences

beorn7:
# Notes from the brainstorming session with @fabxc and myself.

The issues here are net necessarily meant to be addressed in this PR. They can ass well be addressed in the future. (Record them as issues or as TODOs in the code in that case.)

## Edit semantics
- The only fields that are editable are start time, end time, creator, and comment.
- Creator and comment can be edited freely at any time (it's really just an annotation-like field).
- Start time can not be modified if already in the past, and it can never be set to anything in the past.
- Same is true for end time.
- Edits of any other fields will end the current silence and create a now one starting now (which allows tracking history of silences). (Not implemented yet.)
- If editing the start or end date of an expired silence, a new one will be created in the same way. (Not implemented yet.)
- The various options above should be transparent from the UI perspective. It's just an "edit", and decisions are made behind the scenes.
- Note that concurrent edits (or netsplits) will create multiple new silences. Only in case of direct edits (dates, creator, comment), the last write wins.
- UI quirk: An expired silence should not have a delete button. (Perhaps call it "expire" anyway.)

## Persistence
- Mesh has no persistence. Thus, if AM is run with a single instance, there is no persistence for silences and notifications. (And same should all instances be restarted at the same time.)
- Some persistence needs to be added for those. Could be just snapshotting of the state in memory.
- GC saved state upon startup, after that it's safe to merge the state with the one in Mesh.

## Config skew
- Config (via file and via command line flags) is not synchronized.
- Config skew should be avoided, but the idea is that config skew would not have disastrous effects.

provider/mesh/state.go
```diff
@@ -0,0 +1,327 @@
...
+		set: make(map[string]notificationEntry, len(s.set)),
+	}
+	for k, v := range s.set {
+		res.set[k] = v
+	}
+	return res
+}
+
+// Encode the notification state into a single byte slices.
+func (st *notificationState) Encode() [][]byte {
```
beorn7: Inconsistent naming of receiver (s vs st).

beorn7:
Fixing golint warnings would be good (mainly doc comments)

`go vet` is already perfect. :)

provider/mesh/peer.go
fabxc: You mean the silence marker?
It has always been like this and while I agree, this should be addressed in its own PR.

provider/mesh/state.go
fabxc: Good catch, fixed & test added.

beorn7:
Let me know when this is ready for the final review.

fabxc:
You can take another look. I added docs and a few TODOs. Fixed an issue.

For everything subsequent issues + new PRs would be better.

beorn7:
OK, I'll go through it ASAP and file issues where I see fit.

beorn7:
Still on my review queue… sorry for dealy.

beorn7:
Merge question: Do you want to develop the HA version in a different branch for now? In that case, this PR shouldn't go into master.

fabxc:
Let's keep merging new stuff into this branch here until some fundamental additions are in. Eventually we can merge it into master.

Procfile
beorn7: Thanks for the explanation given in person. Could you add a short paragraph to README.md about the possibility of running a test setup with goreman?

provider/boltmem/boltmem.go
```diff
@@ -268,90 +270,72 @@ func (s *Silences) All() ([]*types.Silence, error) {
...
 	err := s.db.Update(func(tx *bolt.Tx) error {
-		b := tx.Bucket(bktSilences)
-
-		k := make([]byte, 8)
-		binary.BigEndian.PutUint64(k, uid)
-
-		return b.Delete(k)
+		// TODO(fabxc): this does not yet comply with the semantics of
+		// deleting just setting EndsAt to now if it is in the future.
+		// As long as we are don't have strong semantics we might still
```
beorn7: Delete "are".

beorn7:
For my understanding: The `Silences` and `NotificationInfo` from the boltmem provider are not used anymore. Have you left them in because you might potentially use them to solve persistence (#387)? Or are their other reasons?

fabxc:
Probably going to use snapshotting. Probably will delete them as they are also semantically different now so they are not even useful for educational purpose anymore.

provider/mesh/peer.go
```diff
@@ -0,0 +1,273 @@
...
+
+// NotificationInfos provides gossiped information about which
+// receivers have been notified about which alerts.
+type NotificationInfos struct {
+	st     *notificationState
+	send   mesh.Gossip
+	logger log.Logger
+}
+
+func NewNotificationInfos(logger log.Logger) *NotificationInfos {
```
beorn7: doc comment to satisfy golint.

provider/mesh/peer.go
```diff
@@ -0,0 +1,273 @@
...
+	"github.com/prometheus/alertmanager/provider"
+	"github.com/prometheus/alertmanager/types"
+	"github.com/prometheus/common/log"
+	"github.com/prometheus/common/model"
+	"github.com/satori/go.uuid"
+	"github.com/weaveworks/mesh"
+)
+
+// NotificationInfos provides gossiped information about which
+// receivers have been notified about which alerts.
```
beorn7: Add a hint that it implements the Gossiper interface.

provider/mesh/peer.go
fabxc: Meh, think we discussed these things in the past. Golint has way less
confidence than govet and commenting by saying "NewSilencer returns a new
silencer" seems slightly useless.

provider/mesh/peer.go
```diff
@@ -0,0 +1,273 @@
...
+				Timestamp: e.Timestamp,
+			})
+		} else {
+			res = append(res, nil)
+		}
+	}
+	return res, nil
+}
+
+// Silences provides gossiped silences.
```
beorn7: ... and implements the mesh.Gossiper interface.

provider/mesh/peer.go
```diff
@@ -0,0 +1,273 @@
...
+type Silences struct {
+	st     *silenceState
+	mk     types.Marker
+	send   mesh.Gossip
+	logger log.Logger
+}
+
+func NewSilences(mk types.Marker, logger log.Logger) *Silences {
```
beorn7: Doc comment to please golint.

provider/mesh/peer.go
beorn7: On 21 June 2016 at 17:56, Fabian Reinartz <notifications@github.com> wrote:
>
> Meh, think we discussed these things in the past. Golint has way less
> confidence than govet and commenting by saying "NewSilencer returns a new silencer"
> seems slightly useless.

We might have discussed these things, but apparently, we came to
different conclusions:

- Yes, doc comments sometimes appear useless.
- Still forcing oneself to write them down has the advantage of
thinking twice if there is a way to create a useful doc comment after
all.
- Furthermore, it silences a golint warning. Golint warnings might be
false positives, but easily avoided warnings should still be avoided,
so that we don't create useless noise and are able to spot meaningful
golint warnings.

All provider sub packages are, btw, completely free of golint
warnings. Those two are the only ones.






-- 
Björn Rabenstein, Engineer
http://soundcloud.com/brabenstein

SoundCloud Ltd. | Rheinsberger Str. 76/77, 10115 Berlin, Germany
Managing Director: Alexander Ljung | Incorporated in England & Wales
with Company No. 6343600 | Local Branch Office | AG Charlottenburg  |
HRB 110657B

provider/mesh/state.go
```diff
@@ -0,0 +1,350 @@
...
+)
+
+func utcNow() time.Time { return time.Now().UTC() }
+
+type notificationEntry struct {
+	Resolved  bool
+	Timestamp time.Time
+}
+
+type notificationState struct {
```
beorn7: Doc comment: notificationState implements mesh.GossipData.
(Those things are really helpful when reading the code. Go-oracle can tell me the same thing, but it takes long enough that I wouldn't hit the hotkey for each struct I see.)

beorn7:
OK, I think I have seen and (kind of) understood everything.
How to handle persistence and exact semantics after a longer partition might become interesting topics during our meeting with @peterbourgon.

I'll file more GH issues for the TODOs discussed so far. Feel free to amend/correct/close them as you see fit.

provider/mesh/peer.go
fabxc: Yea, I'll add respective comments. I still find it utterly useless and a tool that forces me to do useless things doesn't seem like a particularly good tool. After all the tool should help me, not the other way around.

beorn7: On 22 June 2016 at 09:39, Fabian Reinartz <notifications@github.com> wrote:

> Yea, I'll add respective comments. I still find it utterly useless and a
> tool that forces me to do useless things doesn't seem like a particularly
> good tool. After all the tool should help me, not the other way around.


Thanks for adding the doc comments.

The tool correctly implements the style guide of the Go community, which we
adopted for Prometheus. The rule is that every exported identifier requires
a doc comment. Obviously, we could decide to change the style guide, but
based on my personal experience, I predict the debate will be moot.  (I had
that debate quite often for different languages, taking either side at
times, but the outcome never changed.)

-- 
Björn Rabenstein, Engineer
http://soundcloud.com/brabenstein

SoundCloud Ltd. | Rheinsberger Str. 76/77, 10115 Berlin, Germany
Managing Director: Alexander Ljung | Incorporated in England & Wales with
Company No. 6343600 | Local Branch Office | AG Charlottenburg  | HRB 110657B

fabxc: Plenty of examples in the standard library where docs are omitted where they don't add any value.

A few right here:
https://golang.org/src/encoding/json/encode.go#L211

beorn7: This is really not the right place to open a debate about changing the
style guide. If you really want to go forward with it, I suggest to start
an email thread on prometheus-team@.

On 22 June 2016 at 10:38, Fabian Reinartz <notifications@github.com> wrote:

> In provider/mesh/peer.go
> <https://github.com/prometheus/alertmanager/pull/365#discussion_r68014440>
> :
>
> > +	"github.com/prometheus/alertmanager/types"
> > +	"github.com/prometheus/common/log"
> > +	"github.com/prometheus/common/model"
> > +	"github.com/satori/go.uuid"
> > +	"github.com/weaveworks/mesh"
> > +)
> > +
> > +// NotificationInfos provides gossiped information about which
> > +// receivers have been notified about which alerts.
> > +type NotificationInfos struct {
> > +	st     *notificationState
> > +	send   mesh.Gossip
> > +	logger log.Logger
> > +}
> > +
> > +func NewNotificationInfos(logger log.Logger) *NotificationInfos {
>
> Plenty of examples in the standard library where docs are omitted where
> they don't add any value.
>
> A few right here:
> https://golang.org/src/encoding/json/encode.go#L211
>
> —
> You are receiving this because you commented.
> Reply to this email directly, view it on GitHub
> <https://github.com/prometheus/alertmanager/pull/365/files/656738451e34e0fcd9d8d2218133d2656718ee83#r68014440>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AFWZnnu7PPp8sdc9MeST9xERgBC2z5wqks5qOPSJgaJpZM4ItRTv>
> .
>



-- 
Björn Rabenstein, Engineer
http://soundcloud.com/brabenstein

SoundCloud Ltd. | Rheinsberger Str. 76/77, 10115 Berlin, Germany
Managing Director: Alexander Ljung | Incorporated in England & Wales with
Company No. 6343600 | Local Branch Office | AG Charlottenburg  | HRB 110657B

fabxc: I'm not opening a debate - I said I'll add the comments.
I'm just trying to convey that, other than govet, golint is not right by default but merely provides suggestions. The noisiness is even tweak-able via configuration flag.

beorn7: On 22 June 2016 at 12:00, Fabian Reinartz <notifications@github.com> wrote:

> I'm just trying to convey that, other than govet, golint is not right by
> default but merely provides suggestions. The noisiness is even tweak-able
> via configuration flag.
>
That's not disputed. But in this case, nothing is wrong with golint. It
merely implements what's clearly written in the style guide:
https://github.com/golang/go/wiki/CodeReviewComments#doc-comments .
If you have no intention to change the style guide, then there is indeed
nothing to debate.

-- 
Björn Rabenstein, Engineer
http://soundcloud.com/brabenstein

SoundCloud Ltd. | Rheinsberger Str. 76/77, 10115 Berlin, Germany
Managing Director: Alexander Ljung | Incorporated in England & Wales with
Company No. 6343600 | Local Branch Office | AG Charlottenburg  | HRB 110657B

fabxc: It says "should non-trivial unexported type or function declarations" – "should" and "non-trivial". Triviality was exactly my argument.

beorn7: On 22 June 2016 at 13:15, Fabian Reinartz <notifications@github.com> wrote:

> It says "should non-trivial unexported type or function declarations" –
> "should" and "non-trivial". Triviality was exactly my argument.


But you forgot the "unexported". Non-triviality only applies to unexported
type or function declarations. The function my review comment was about is
exported. golint never complains about doc comments for unexported type or
function declaration, exactly because it is not able to judge about
triviality.


-- 
Björn Rabenstein, Engineer
http://soundcloud.com/brabenstein

SoundCloud Ltd. | Rheinsberger Str. 76/77, 10115 Berlin, Germany
Managing Director: Alexander Ljung | Incorporated in England & Wales with
Company No. 6343600 | Local Branch Office | AG Charlottenburg  | HRB 110657B

fabxc:
Merging into master and continuing work there.

PR #367, opened 2016-06-04
Use localhost for tests

mattbostock:
Previously, the tests would listen on all available interfaces.

Instead, have the tests use localhost only; using all available
interfaces is unnecessary.

On Mac OS X with the builtin firewall enabled, it triggers annoying
prompts to allow the tests to listen on all interfaces.


fabxc:
Thanks 👍

PR #368, opened 2016-06-04
Allow route receiver to be inherited

mattbostock:
This is useful when you're using a label to determine the receiver but
want to override other options such as `group_by`. Currently you'd have
to duplicate the matchers for the receivers to be able to do this.

`checkReceiver()` now returns no error if a receiver is empty and we add
a check to ensure that the root route has a receiver defined. I've added
a test for this.

This brings the `receiver` option into line with the other options
(`group_by`, `group_wait`, etc) in the sense that routes can now inherit
the receiver from the parent node.

From https://prometheus.io/docs/alerting/configuration/:

> A route block defines a node in a routing tree and its children. Its
> optional configuration parameters are inherited from its parent node
> if not set.

Also, add tests for:
- inheriting a receiver from the default route
- overriding the receiver when the parent route has not set one
- inheriting a receiver from a parent route that's not the default

To be sure of edge case behaviour, we test inheriting the receiver
from a child route. Hopefully no one would actually implement this edge
case in reality as it's overly complex but we test anyway to be sure
that it doesn't trigger unexpected behaviour.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -225,6 +225,9 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if c.Route == nil {
 		return fmt.Errorf("No routes provided")
 	}
+	if len(c.Route.Receiver) == 0 {
+		return fmt.Errorf("Root route must specify a default receiver")
+	}
 	if len(c.Route.Match) > 0 || len(c.Route.MatchRE) > 0 {
 		return fmt.Errorf("Root route must not have any matchers")
 	}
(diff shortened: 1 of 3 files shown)


mattbostock:
I'm wary of introducing changes that makes routing overly complex but  I think the changes here are justified for the amount of flexibility that it provides when grouping notifications. I'd be grateful for any feedback or comments.

fabxc:
Thanks! Sorry about the delay but this looks good to me 👍 
Objections @brian-brazil ?

brian-brazil:
Makes sense from a consistency standpoint.

Issue #353, opened 2016-05-17
Add /-/reload endpoint


ZhenyangZhao:
I found reload code in the main.go - line:149 

But i don't know how to use it make alertmanger reload config file .In prometheus i can use post : http://prometheus/-/reload to reload.

what should i do in alertmanager?

Thks

fabxc:
You can send a `SIGHUP` to the Alertmanager process. For example `killall -HUP alertmanager`.

I renamed this issue so we add a `/-/reload` endpoint for Alertmanager as well.

ProfessorYang:
@fabxc Thank you very much for your reply to my question. Is the `/-/reload` api endpoint will appear in the near version?

fabxc:
Closed via #372

PR #369, opened 2016-06-06
Add /-/reload endpoint #353

ZhenyangZhao:
I had added "api/-/reload" endpoint to reload configuration file like prometheus server.
Hope my contribute can be used.
Thanks all.

### DIFF
--- a/api.go
+++ b/api.go
@@ -58,6 +58,7 @@ type API struct {
 	config         string
 	resolveTimeout time.Duration
 	uptime         time.Time
+	reloadCh       chan struct{}
 
 	groups func() AlertOverview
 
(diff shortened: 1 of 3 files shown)


fabxc:
Thanks! It's not quite there yet. You have to modify the part in `main.go` to make use of the `API.Reload()` channel. You can check how prometheus/prometheus does it for reference.

ZhenyangZhao:
@fabxc  Thanks for reply me. Did you mean to add the code like the prometheus server?

// Wait for reload or termination signals.
// Unblock SIGHUP handler.
close(hupReady)

fabxc:
Yes, we should have the `hupReader` functionality as well.
In particular I was talking about [this line](https://github.com/prometheus/prometheus/blob/master/cmd/prometheus/main.go#L134) which is necessary so that the reload channel from the API handler actually triggers a reload.

ZhenyangZhao:
@fabxc Thank you very much.I have learn so much knowledge.

Best Regards.

fabxc:
Thanks, can you run `make format` submit the changes and then squash all your commits?

ZhenyangZhao:
@fabxc sure :)

PR #372, opened 2016-06-06
Add api/-/reload endpoint #353

ZhenyangZhao:
Add api/-/reload endpoint #353 

@fabxc  Sorry for re-created the pull request. Because i cant fix the git conflict.

Thanks

### DIFF
--- a/api.go
+++ b/api.go
@@ -58,6 +58,7 @@ type API struct {
 	config         string
 	resolveTimeout time.Duration
 	uptime         time.Time
+	reloadCh       chan struct{}
 
 	groups func() AlertOverview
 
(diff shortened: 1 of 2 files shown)


fabxc:
Thanks 👍

juliusv:
Was it intentional that the path is `/api/-/reload` in AM, whereas Prometheus has the same under `/-/reload`?

ZhenyangZhao:
@juliusv Thankyou very much.  I add  this  reload  endpoint under the path "api/" in alertmanager,because i found almost all the alertmanager's api are under 'api/' path  not the root path like prometheus.

fabxc:
It should be the root path like Prometheus

juliusv:
Ok, I opened a PR to move it: https://github.com/prometheus/alertmanager/pull/376

ZhenyangZhao:
@fabxc @juliusv   :+1: thanks

PR #379, opened 2016-06-13
inhibit: reduce O(n^2) complexity problem

fabxc:
This change adds an inhibition cache. New alerts are checked against all
inhibition rules' source matchers. At notification time, we only have to
check the target matchers against the notified alerts. If there are
alerts in the source cache, we only have to verify the equal labels
against these.

### DIFF
--- a/inhibit.go
+++ b/inhibit.go
@@ -15,6 +15,7 @@ package main
 
 import (
 	"sync"
+	"time"
 
 	"github.com/prometheus/common/log"
 	"github.com/prometheus/common/model"
@@ -31,7 +32,8 @@ type Inhibitor struct {
 	rules  []*InhibitRule
 	marker types.Marker
 
-	mtx sync.RWMutex
+	mtx   sync.RWMutex
+	stopc chan struct{}
 }
 
 // NewInhibitor returns a new Inhibitor.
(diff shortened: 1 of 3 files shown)


fabxc:
Checking each incoming and outgoing alert against each inhibition rule is roughly equivalent to `metric_relabel_configs` just that our load is way lower here.

As we want to scale beyond several thousand alerts per second, we will want to work on lock contention a bit.

inhibit.go
```diff
@@ -125,20 +177,48 @@ func NewInhibitRule(cr *config.InhibitRule) *InhibitRule {
...
+
+Outer:
+	for _, a := range r.scache {
+		// The cache might be stale and contain resolved alerts.
+		if a.Resolved() {
+			continue
 		}
+		for n := range r.Equal {
+			if a.Labels[n] != lset[n] {
+				continue Outer
```
juliusv: To me, I guess a simple `return false` would be less confusing than doing label-continue business here.

fabxc: But it would be wrong. We have to check the subsequent alerts as well as one is sufficient to mute the target alert.

juliusv: Ah see... I got confused and thought the continue would end the outer loop. But of course it doesn't, sorry.

juliusv:
:+1:

fabxc:
Thanks

PR #390, opened 2016-06-20
provider/mesh: always store UTC

fabxc:


provider/mesh/state.go
```diff
@@ -235,9 +237,14 @@ func (st *silenceState) set(s *types.Silence) error {
 	st.mtx.Lock()
 	defer st.mtx.Unlock()
 
+	s.StartsAt = s.StartsAt.UTC()
+	s.EndsAt = s.EndsAt.UTC()
+
 	now := st.now()
 	s.UpdatedAt = now
 
+	// Truncate timestamps to seconds.
```
fabxc: Na, leftover. Wanted to solve the time zones before looking at anything different.

PR #392, opened 2016-06-20
Fixing the hipchat room number parsing issue #chefdeploy

sciffer:
It seems that the sprintf was trying to parse a %d from string type, so I just replaced the %d with %s.
Another option is to parse the string into a number (not necessary to my opinion).

Also adjusted the version according to the release.


beorn7:
Yeah, I broke that, because `go vet` believes the RoomID is an int. very weird.
Still looking.

Any any case, could you please change the version in a separate commit? Such a change should be obvious from the git log.

beorn7:
I simply have no explanation why `go vet` believes `n.conf.RoomID` is an `int`. @fabxc any idea?

Anyway, this is an important fix @sciffer . If you move the change in VERSION into a different commit, I'll merge, and then we have to release 0.2.1...

sciffer:
Reverted the version change - you can push it now.

beorn7:
Thanks, but that's not what I wanted. Now we have two confusing commits, the first includes a change to the version file despite its description saying something else, and the second reverts that version change although it is the right thing to do the change.

Could you rewrite your commits so that the one commit sets the version to 0.2.0, and the other (only) fixes the Sprintf problem?

(No problem if that's too much git gymnastics for you. I can create those commits myself. But we really want to avoid an unnecessarily confusing git history.)

sciffer:
Sure, I'll do that tomorrow.

PR #393, opened 2016-06-21
Handle UI timestamp drift

fabxc:
@beorn7


beorn7:
:+1: 

Looks like the tests are broken in fabx-soha, too. Or they are at least flaky. Would be good to get those fixed so that we don't have to check if the failing tests are the known ones each time.

fabxc:
The functional tests are flaking on CI since they exist.

beorn7:
They are also flaky on my laptop.

fabxc:
Never had that happen macbook pro. Timing is quite genrous - they are all
in parallel though.

I've plans on this but that's better for a later point in time.

PR #396, opened 2016-06-21
Adjusting version

sciffer:


VERSION
```diff
@@ -1 +1 @@
-0.1.1
+0.2
```
beorn7: Sorry for one last nit: This should be 0.2.0.

beorn7:
The commits should be squashed into one. But I'm not going to torture you anymore for a 2-byte change. I'll create a squashed commit myself and close this PR.

Thanks for your efforts.

PR #397, opened 2016-06-21
Add silence cache in front of BoltDB

fabxc:
@beorn7 @mattbostock 

This addresses the next iteration of too-much-serialization issues by simply adding an in-memory cache.
For the future, where we'll be looking at thousands of silences, we'll want to index silences in a smarter way.

We were also decoding all silences twice by accident – oops.

### DIFF
--- a/provider/boltmem/boltmem.go
+++ b/provider/boltmem/boltmem.go
@@ -217,6 +217,9 @@ func (a *Alerts) Put(alerts ...*types.Alert) error {
 type Silences struct {
 	db *bolt.DB
 	mk types.Marker
+
+	mtx   sync.RWMutex
+	cache map[uint64]*types.Silence
 }
 
 // NewSilences creates a new Silences provider.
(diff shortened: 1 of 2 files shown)


beorn7:
@fabxc On my review queue... Just busy with an offsite today, so latency is somewhat random.

provider/boltmem/boltmem.go
```diff
@@ -272,19 +288,14 @@ func (s *Silences) All() ([]*types.Silence, error) {
...
 				return err
 			}
-			ms.ID = binary.BigEndian.Uint64(k)
-
-			if err := json.Unmarshal(v, &ms); err != nil {
-				return err
-			}
-
-			res = append(res, types.NewSilence(&ms))
+			// The ID is is duplicated in the value and always equal
```
juliusv: "is is" -> "is"

juliusv:
:+1: otherwise

beorn7:
At least the deletion of silences has a race condition.

fabxc:
Sure – totally skipped that yesterday, thanks.

beorn7:
:+1:

PR #403, opened 2016-06-22
make: reduce number of parallel tests

fabxc:
Checking if it fixes the CI flakyness at least partially.


brian-brazil:
:+1:

fabxc:
Didn't actually help that much

PR #407, opened 2016-06-24
provider/mesh: fix end timestamp for unstarted silences

fabxc:
This changes the end timestamp for unstarted silences to the
start timestamp so the silence remains valid by not having the end
time before the start time.

@beorn7 @brian-brazil

### DIFF
--- a/provider/mesh/state.go
+++ b/provider/mesh/state.go
@@ -283,12 +283,20 @@ func (st *silenceState) del(id uuid.UUID) (*types.Silence, error) {
 	if !ok {
 		return nil, provider.ErrNotFound
 	}
+	now := st.now()
+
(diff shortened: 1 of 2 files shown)


provider/mesh/state.go
```diff
@@ -283,12 +283,20 @@ func (st *silenceState) del(id uuid.UUID) (*types.Silence, error) {
...
+
 	// Silences are immutable by contract so we create a
 	// shallow copy.
 	sil := *prev
-	now := st.now()
 	sil.UpdatedAt = now
-	sil.EndsAt = now
+
+	// If silence hasn't started yet, terminate it at
+	// it's starting time.
```
brian-brazil: its

brian-brazil:
:+1:

PR #408, opened 2016-06-24
HA AM silence adjustments in UI

fabxc:
@beorn7 @brian-brazil 

Please try it out – UI is really not my thing.


beorn7:
Not a UI person either…

This looks good in general. What worries me a bit is the `new Date` in the functions. This could give us an inconsistent representation of silences with unlucky timing (silences displayed twice or not at all). How about adding a "now" argument to each of those functions that now contain a `new Date` call so that we have a single notion of "now" throughout rendering of one page?

fabxc:
Totally agree on the  `new Date` part. It was a the safe route as I'm not 100% sure what's invoked when with the bi-directional data binding, which could even cause more weirdnesses.

fabxc:
Made some changes @beorn7

beorn7:
OK, thanks. I like it, and it all seemed to work perfectly when I tried it out.

I'm not really qualified about angular and such, but if nobody else wants to comment, then :+1:

PR #414, opened 2016-06-28
Updated webhook notifier config name in UnmarshalYAML.

diogogmt:
Going over the notifier's code noticed that the config name was set to slack in the webhook unmarshaler function.


fabxc:
Thanks 👍

PR #415, opened 2016-06-29
provider/mesh: implement snapshotting

fabxc:
This change adds snapshotting to the silences and notification infos
providers. Snapshots are created periodically and on shutdown. They are
loaded into memory on startup.
Periodic snapshotting is run right after garbage collection.

@beorn7

### DIFF
--- a/provider/mesh/peer.go
+++ b/provider/mesh/peer.go
@@ -13,6 +13,7 @@ import (
 	"github.com/weaveworks/mesh"
 )
 
+// replaceFile wraps a file that is moved to another filename on closing.
 type replaceFile struct {
 	*os.File
 	filename string
@@ -28,6 +29,7 @@ func (f *replaceFile) Close() error {
 	return os.Rename(f.File.Name(), f.filename)
 }
 
+// openReplace opens a new temporary file that is moved to filename on closing.
 func openReplace(filename string) (*replaceFile, error) {
 	tmpFilename := fmt.Sprintf("%s.%s", filename, utcNow().Format(time.RFC3339Nano))
 
@@ -54,6 +56,7 @@ type NotificationInfos struct {
 	stopc     chan struct{}
 }
 
+// NewNotificationInfos returns a new NotificationInfos object.
 func NewNotificationInfos(logger log.Logger, retention time.Duration, snapfile string) (*NotificationInfos, error) {
 	ni := &NotificationInfos{
 		logger:    logger,
@@ -79,6 +82,7 @@ func (ni *NotificationInfos) Register(g mesh.Gossip) {
 	ni.send = g
 }
 
+// TODO(fabxc): consider making this a flag.
 const maintenanceInterval = 15 * time.Minute
 
 // Run starts blocking background processing of the NotificationInfos.
@@ -204,6 +208,7 @@ type Silences struct {
 	snapfile  string
 }
 
+// NewSilences creates a new Silences object.
 func NewSilences(mk types.Marker, logger log.Logger, retention time.Duration, snapfile string) (*Silences, error) {
 	s := &Silences{
 		st:        newSilenceState(),
--- a/provider/mesh/peer.go
+++ b/provider/mesh/peer.go
@@ -2,6 +2,7 @@ package mesh
 
 import (
 	"fmt"
+	"os"
 	"time"
 
 	"github.com/prometheus/alertmanager/provider"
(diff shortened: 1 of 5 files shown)


beorn7:
I'll have a look ASAP. (Currently busy with SRECon talk..., deadline for slides today...)

beorn7:
This is at the top of my work queue, finally... but it won't happen today. I'm sorry.

provider/mesh/peer.go
```diff
@@ -12,35 +13,107 @@ import (
...
+	snapfile  string
+	logger    log.Logger
+	stopc     chan struct{}
 }
 
-func NewNotificationInfos(logger log.Logger) *NotificationInfos {
-	return &NotificationInfos{
-		logger: logger,
-		st:     newNotificationState(),
+func NewNotificationInfos(logger log.Logger, retention time.Duration, snapfile string) (*NotificationInfos, error) {
```
beorn7: Doc comment (still) missing.

provider/mesh/peer.go
```diff
@@ -12,35 +13,107 @@ import (
 	"github.com/weaveworks/mesh"
 )
 
+type replaceFile struct {
```
beorn7: Doc comment would be great to help the reader.

provider/mesh/peer.go
```diff
@@ -12,35 +13,107 @@ import (
...
+
+	return ni, ni.st.loadSnapshot(f)
 }
 
 // Register the given gossip channel.
 func (ni *NotificationInfos) Register(g mesh.Gossip) {
 	ni.send = g
 }
 
+const maintenanceInterval = 15 * time.Minute
```
beorn7: Should be a flag eventually. (Add a TODO if you don't want to add a flag right now.)

provider/mesh/peer.go
```diff
@@ -122,18 +195,34 @@ func (ni *NotificationInfos) Get(recv string, fps ...model.Fingerprint) ([]*type
...
+	retention time.Duration
+	snapfile  string
 }
 
-func NewSilences(mk types.Marker, logger log.Logger) *Silences {
-	return &Silences{
-		st:     newSilenceState(),
-		mk:     mk,
-		logger: logger,
+func NewSilences(mk types.Marker, logger log.Logger, retention time.Duration, snapfile string) (*Silences, error) {
```
beorn7: Doc comment (still) missing.

provider/mesh/state.go
```diff
@@ -70,6 +52,47 @@ func (st *notificationState) gc(retention time.Duration) {
...
+				break
+			}
+			return err
+		}
+		if err := dec.Decode(&n); err != nil {
+			return err
+		}
+		// TODO(fabxc): remove serialization of alert/receiver into
+		// string key.
+		st.set[k] = n
```
beorn7: I don't understand the context here. Shouldn't this TODO be in peer.go (line 162 or so)?

provider/mesh/state.go
```diff
@@ -70,6 +52,47 @@ func (st *notificationState) gc(retention time.Duration) {
 	}
 }
 
+func (st *notificationState) snapshot(w io.Writer) error {
```
beorn7: Shouldn't this use the same codepath as the `Encode` method? The latter writes to a byte slice, and this to a writer. But `Encode` could call this method with the bytes.Buffer as w.

However, `Encode` gobs the whole map, while this method sequentially goes through the map and encodes key and value separately. Is there a reason for that?

provider/mesh/state.go
```diff
@@ -70,6 +52,47 @@ func (st *notificationState) gc(retention time.Duration) {
...
+			return err
+		}
+		if err := enc.Encode(n); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (st *notificationState) loadSnapshot(r io.Reader) error {
```
beorn7: In the same spirit, this could use `decodeNotificationSet`.

beorn7:
Same remark as for encoding and decoding of the notification set applies to the silence set (and if I have missed something, I have missed the same here).

You haven't applied some of the (trivial) comments from https://github.com/prometheus/alertmanager/pull/365 . It would remove friction while reviewing if you did. (My brain could think about the real issues, not things like "missing doc comment".)

provider/mesh/state.go
fabxc: Serialization of state into messages for gossip communication and snapshotting is completely independent. Right now it's mostly the same (modulo your comment below) but I intend the encoding of either.

Think of this PR more to be about everything but the exact format.

provider/mesh/state.go
fabxc: See above, no relation between gossip state and snapshot encoding.

I'm not 100% sure how gob decodes a map internally, but I imagine it is not streaming as we are doing here. That'll at least make recovering partially corrupted snapshots hard at some point.
Separation of key/value is only because that's how the current types are.

Material for later PRs to not blow up this one.

beorn7:
I agree with both your points, (1) let's not define the exact on-disk format here and (2) let's not blow up this PR. However, my conclusion seems to be exactly different. Both points would make me use the same encoding/decoding codepath for mesh and snapshotting, and only split it up once a deviation actually occurs.

Your point (1) seems to be in contradiction with assumed optimizations for recoverability of corrupted snapshots. Making serialization code more complex based on that while the format is explicitly not finalized is premature optimization.

Having said all that, I don't feel strongly enough about those points to start a lengthy discussion. If you in turn feel strongly about keeping it the way it is, feel free to disregard my assessment.

fabxc:
The only thing that matters right now is that it's symmetric and works. It will literally be the next things being changed before being anywhere near alpha deployment.
Not much use in changing the entire PR and its tests now IMO.

Doc comments added. Would like to merge soon.

beorn7:
:+1:

PR #417, opened 2016-07-03
Add VictorOps Notifier

rhazdon:
Hi,

I upgraded the alertmanager for sending alerts to [VictorOps](https://victorops.com/). 
In my company we are using this feature already so please excuse I didn't open a discussion beforehand.

What do you think about it?


beorn7:
@brian-brazil @fabxc I guess you are more likely than me to have an opinion on this.

VERSION
```diff
@@ -1 +1 @@
-0.2.1
+0.2.2
```
beorn7: Version change will happen during release, not upon individual commits.

rhazdon: Of course. :)
I will fix that asap.

config/notifiers.go
```diff
@@ -92,6 +92,17 @@ var (
 		// TODO: Add a details field with all the alerts.
 	}
 
+	// DefaultVictorOpsConfig defines default values for VictorOps configurations.
+	DefaultVictorOpsConfig = VictorOpsConfig{
+		NotifierConfig: NotifierConfig{
+			VSendResolved: true,
+		},
+		MessageType:  `{{ if eq .Status "firing" }}CRITICAL{{ else }}RECOVERY{{ end }}`,
```
brian-brazil: Does Victorops treat messages differently depending on this? We may wish to hardcode resolved notifications to send a recovery message.

config/notifiers.go
```diff
@@ -306,6 +317,37 @@ func (c *OpsGenieConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
...
 
+// VictorOpsConfig configures notifications via VictorOps.
+type VictorOpsConfig struct {
+	NotifierConfig `yaml:",inline"`
+
+	APIKey       Secret `yaml:"api_key"`
+	APIURL       string `yaml:"api_url"`
+	RoutingKey   string `yaml:"routing_key"`
+	MessageType  string `yaml:"message_type"`
+	EntityID     string `yaml:"entity_id"`
```
brian-brazil: This should be hardcoded to avoid users changing it and inadvertently breaking things.

template/default.tmpl
```diff
@@ -31,6 +31,11 @@
 {{ define "pagerduty.default.instances" }}{{ template "__text_alert_list" . }}{{ end }}
 
 
+{{ define "victorops.default.entity_id" }}{{ .GroupLabels.SortedPairs.Values | join " " }} {{ if gt (len .CommonLabels) (len .GroupLabels) }} / {{ with .CommonLabels.Remove .GroupLabels.Names }}{{ .Values | join " " }}{{ end }}{{ end }}{{ end }}
```
brian-brazil: It's incorrect to include the common labels, this should only be the group labels.

template/default.tmpl
```diff
@@ -31,6 +31,11 @@
...
+{{ define "victorops.default.message" }}{{ template "__subject" . }} | {{ template "__alertmanagerURL" . }}{{ end }}
+{{ define "victorops.default.from" }}{{ template "__alertmanager" . }}{{ end }}
```
brian-brazil: We've used the URL for other similar fields

rhazdon: This field returneds the name of the tool from which the incident came from. 
VictorOps displays this information then. 
b

To avoid misunderstanding I can rename this field into the json key VictorOps expects (monitoring_tool), described here: [VictorOps API](http://victorops.force.com/knowledgebase/articles/Getting_Started/Incident-Fields-Glossary/?q=Create+Incident&l=en_US&fs=Search&pn=1).

config/notifiers.go
rhazdon: Yes. Depends on the value, VictorOps will create or close an incident.  
It expects one of the following keywords in this field: INFO, WARNING, ACKNOWLEDGEMENT, CRITICAL, RECOVERY. Available here: [API Docs](http://victorops.force.com/knowledgebase/articles/Integration/Alert-Ingestion-API-Documentation/). 

Is there maybe a better way to resolve the exact status/type of an alert?

template/default.tmpl
rhazdon: Removed. :)

config/notifiers.go
mirthy: I wonder whether or not this MessageType switching should occur instead impl.go.  Kind of like how the pagerduty and opsgenies ones do it. (PagerDuty): https://github.com/rhazdon/alertmanager/blob/001e0716bda50a9721b7c8b9cc32751969f6b8e6/notify/impl.go#L405 

But I can see an argument against it if we want to be able to use INFO, WARNING, etc.

config/notifiers.go
mirthy: From my understand in the [API docs](http://victorops.force.com/knowledgebase/articles/Integration/Alert-Ingestion-API-Documentation/), Entity Id is very much like the OpsGenie "Source," there we use the __alertmanagerURL but in this case we're using Group labels, which may be more readable and understandablet oo.

config/notifiers.go
brian-brazil: That's the interesting questino. Allowing the user to do it this way will lead to users breaking recovery notifications. So what we might want to do is leave this configurable but only apply to firing alerts, and have some code to substitute in emergency if an invalid value is used.

config/notifiers.go
brian-brazil: My read of the docs is that this is like Pagerduty's IncidentKey, and should be handled accordingly.

mirthy: ahh yes, I totally agree with that reading now.

mirthy:
Any more progress with this?  What do we need to do here?

rhazdon:
@mirthy Yes, in 2-3 days. I became a daddy two days ago. :)

rhazdon:
I rebased my fork and solved the merge conflicts.

mirthy:
Hi guys, looks like @rhazdon made some more changes, any chance someone could look at this again?  Sorry to be a pain.  Is there anything I can do to help?

rhazdon:
Hi Guys, 
is there anything I should change or improve? :)

fabxc:
I was on vacation. Am looking now.

notify/impl.go
```diff
@@ -730,6 +734,104 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) error {
...
+	if resp.StatusCode/100 != 2 {
+		body, _ := ioutil.ReadAll(resp.Body)
+
+		var responseMessage victorOpsErrorResponse
+		if err := json.Unmarshal(body, &responseMessage); err != nil {
+			return fmt.Errorf("could not parse error response %q", body)
+		}
+
+		fmt.Printf("unexpected VictorOps response from %s (POSTed %s), %s: %s",
+			apiURL, msg, resp.Status, body)
```
fabxc: No line break here.

Also this should be using `log.Debugf` rather than `fmt.Printf`.

config/notifiers.go
fabxc: Yes, this should not be configurable and still needs to be removed.
In the implementation it should use `GroupKey(ctx)`.

notify/impl.go
rhazdon: Hi @fabxc, thank you very much. You are right, I fixed that. :)

config/notifiers.go
brian-brazil: Is this visible to the end user in VictorOps? If so we probably want something prettier as is currently done.

fabxc: No we don't. In PagerDuty the key is visible to the user as well and we don't care. By default VictorOps creates a random ID for you, so even that it's not that different.

We can think about having something more descriptive in the future but not in the context of this PR. The current group key is also a fairly large digest of things, which would exceed a few thousands characters quite easily.

fabxc:
@rhazdon please also check the comment on the entity ID above.

rhazdon:
I did. Testing it currently on prodution.

fabxc:
Thanks. Can you squash that into one commit and then we should be good for merge.

fabxc:
Nope, compile error ahead.

fabxc:
👍

fuzzyami:
@rhazdon 
Thanks for your work on VO integration - its something we'd want to use too.

I'm trying it now - and my alerts have payloads ("StateMessage" in VO terms) that could be somewhat improved. The current format contains (among other things) a space-delimited list of values:

> "TooManyConfs1 AMS 3.3.3.3 GER-03  1.1.1.1"

This list originates in the key-value pairs that are part of the alert-data in the AlertManager. It would be awesome if that list contained *both* the keys and the values from the alert: 

> alertname:TooManyConfs1 datacetner:AMS external_ip:3.3.3.3 region:GER-03  internal_ip:1.1.1.1

Those key/value pairs provide important context to the alert. Without the key, the readability of the alert is greatly reduced. Perhaps this can be easily fixed?

Thanks,
-Ami.

brian-brazil:
That's a default for all receivers where the message is considered short, as usually you'll know what the values are from experience and there's things like SMS size limits to worry about. You can override it with notification templates.

fuzzyami:
Thanks, Brian.

For future readers of this thread, I'm pasting below our current VictorOps receiver config. This is a little crude, but it provides all the info we need (alert name is first, then summary and finally full 'payload'). In the future we might customize the summary text and drop the labels.
```
receivers:
- name: victorOps-receiver
  victorops_configs:
	- api_key: <our_secrect_api_key>
		routing_key: <our_routing_key>
		message: 'Alert: {{ .CommonLabels.alertname }}. Summary:{{ .CommonAnnotations.summary }}. RawData: {{ .CommonLabels }}'
```

PR #418, opened 2016-07-04
Add support for smtp AUTH LOGIN

tbouvet:
AUTH LOGIN is an authentication in smtp server (like PLAIN or MD5).

It would be nice to have this feature to use alertmanager email.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -17,6 +17,7 @@ import (
 	"bytes"
 	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
@@ -251,6 +252,12 @@ func (n *Email) auth(mechs string) (smtp.Auth, error) {
 				return nil, fmt.Errorf("invalid address: %s", err)
 			}
 			return smtp.PlainAuth(identity, username, password, host), nil
+		case "LOGIN":
+			password := string(n.conf.AuthPassword)
+			if password == "" {
+				continue
+			}
+			return LoginAuth(username, password), nil
 		}
 	}
 	return nil, nil
@@ -812,4 +819,31 @@ func tmplHTML(tmpl *template.Template, data *template.Data, err *error) func(str
 		s, *err = tmpl.ExecuteHTMLString(name, data)
 		return s
 	}
+}
+
+type loginAuth struct {
+	username, password string
+}
+
+func LoginAuth(username, password string) smtp.Auth {
+	return &loginAuth{username, password}
+}
+
+func (a *loginAuth) Start(server *smtp.ServerInfo) (string, []byte, error) {
+	return "LOGIN", []byte{}, nil
+}
+
+// Used for AUTH LOGIN. (Maybe password should be encrypted)
+func (a *loginAuth) Next(fromServer []byte, more bool) ([]byte, error) {
+	if more {
+		switch string(fromServer) {
+		case "Username:":
+			return []byte(a.username), nil
+		case "Password:":
+			return []byte(a.password), nil
+		default:
+			return nil, errors.New("unexpected server challenge")
+		}
+	}
+	return nil, nil
 }


notify/impl.go
```diff
@@ -251,6 +253,12 @@ func (n *Email) auth(mechs string) (smtp.Auth, error) {
 				return nil, fmt.Errorf("invalid address: %s", err)
 			}
 			return smtp.PlainAuth(identity, username, password, host), nil
+        case "LOGIN":
```
brian-brazil: Have you run `make format`?

notify/impl.go
```diff
@@ -813,3 +821,30 @@ func tmplHTML(tmpl *template.Template, data *template.Data, err *error) func(str
...
+
+func LoginAuth(username, password string) smtp.Auth {
+    return &loginAuth{username, password}
+}
+
+func (a *loginAuth) Start(server *smtp.ServerInfo) (string, []byte, error) {
+    return "LOGIN", []byte{}, nil
+}
+
+// Used for AUTH LOGIN. (Add base64 encrypt mode).
```
brian-brazil: base64 is encoding, not encrypting.

notify/impl.go
```diff
@@ -813,3 +821,30 @@ func tmplHTML(tmpl *template.Template, data *template.Data, err *error) func(str
...
+func (a *loginAuth) Next(fromServer []byte, more bool) ([]byte, error) {
+    if more {
+        switch string(fromServer) {
+        case "Username:":
```
brian-brazil: This will come base64 encoded from the server, is this handled for you already?

notify/impl.go
```diff
@@ -813,3 +821,30 @@ func tmplHTML(tmpl *template.Template, data *template.Data, err *error) func(str
...
+                return base64.StdEncoding.DecodeString(a.username)
```
brian-brazil: Should this be encode?

notify/impl.go
tbouvet: Update is done

notify/impl.go
tbouvet: Make format is done

notify/impl.go
tbouvet: yes

notify/impl.go
tbouvet: I think in the future, it will be good to encrypt all username/password. Because it's no secured to add some passwords in a config file without encrypted their.
Why do you think about ?

tbouvet:
Update is done. Thank you very much for your answer !

notify/impl.go
brian-brazil: That's pointless with a protocol such as this, as we need the plaintext anyway.

notify/impl.go
```diff
@@ -813,3 +821,30 @@ func tmplHTML(tmpl *template.Template, data *template.Data, err *error) func(str
...
+func (a *loginAuth) Start(server *smtp.ServerInfo) (string, []byte, error) {
+	return "LOGIN", []byte{}, nil
+}
+
+// Used for AUTH LOGIN. (Add base64 encode mode).
+func (a *loginAuth) Next(fromServer []byte, more bool) ([]byte, error) {
+	if more {
+		switch string(fromServer) {
+		case "Username:":
+			return base64.StdEncoding.DecodeString(a.username)
```
brian-brazil: I'm pretty sure this should be EncodeString. Have you tested this?

tbouvet: What do you mean ? .

brian-brazil: `a.username` is going to be a normal string, there should be no need to decode it.

notify/impl.go
tbouvet: The problem is not for the protocol but for the props file. Because everybody can look at the user/password in the file. I think it's not so secured. 
My question is just to secure configuration file.

notify/impl.go
tbouvet: done !

notify/impl.go
```diff
@@ -813,3 +821,30 @@ func tmplHTML(tmpl *template.Template, data *template.Data, err *error) func(str
...
+}
+
+// Used for AUTH LOGIN. (Add base64 encode mode).
+func (a *loginAuth) Next(fromServer []byte, more bool) ([]byte, error) {
+	if more {
+		switch string(fromServer) {
+		case "Username:":
+			return []byte(a.username), nil
+		case "Password:":
+			return base64.StdEncoding.DecodeString(a.password)
```
brian-brazil: Same here

tbouvet: Not the same because in this case, password is readable in the configuration file. To encode it is a little bit more secured. Not agree ?
In my opinion, it's not good to put a password neither encoded nor encrypted in a configuration file.

brian-brazil: It's not more secure to encode it, per Kerckhoffs's principle.

tbouvet: Agree with you. Maybe I can use a really crypted string (with a pem key for example) ?
It could be an option.

brian-brazil:
Thanks!

PR #419, opened 2016-07-05
Add crypted password for smtp AUTH LOGIN

tbouvet:
What do you think about to use an encrypted password for stmp auth ? In this case I use public key/private key mechanism (x509 certificate, very usefull in enterprise).

It is an **optional parameter**. 
If you put the parameter **smtp_auth_pem_file** in the configuration, you can put a crypted password in the **smtp_auth_password**.
Otherwise, it's the current use case.

Example:

```
smtp_from: 'xxx@domain.com'
smtp_auth_username: 'myuser'
smtp_auth_password: 'jJxNXYoVs0lQn/PyJO8zO3JtXmJ09wUx13T6rnY+kYWZnCfP/E+wmTJfCfWCpUXxxB5UyYWgt1w1S0LVo='
smtp_auth_pem_file: '/my/path/rsa.pem'
```

In this case, no more decrypted password in the configuration file.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -172,6 +172,9 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 			if ec.AuthIdentity == "" {
 				ec.AuthIdentity = c.Global.SMTPAuthIdentity
 			}
+			if ec.AuthPemFile == "" {
+				ec.AuthPemFile = c.Global.SMTPAuthPemFile
+			}
 		}
 		for _, sc := range rcv.SlackConfigs {
 			if sc.APIURL == "" {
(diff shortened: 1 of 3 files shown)


brian-brazil:
I appreciate that this may be useful for you, however this change won't be accepted. We generally try to avoid security features in Prometheus, as they'd be a never ending source of maintenance.

tbouvet:
I'm very sorry you don't want security features in Prometheus because security is mandatory in enterprise.We can't put a password into a configuration file without encrypt it. 

Really no ?

brian-brazil:
This adds no real security unless you're typing in the password at each launch, or pulling it off dedicated crypto hardware. That's a rabbit hole we don't want to go down.

sgissi:
Nice initiative but I have to agree with @brian-brazil , encrypting data using locally accessible key is as vulnerable as plain text (analogue to put something inside a locked chest and leave the key on top of it). A feature to enable using a private key as authentication in lieu of user/pass (still not more secure, just to be clear) or single-sign-on (e.g. Kerberos/GSSAPI) would be interesting for enterprises moving away from clear text.

PR #420, opened 2016-07-07
provider/boltmem: make alerts purely in-memory.

fabxc:
Initial testing has shown BoltDB in plain usage to be a bottleneck
at a few thousand alerts or more (especially JSON decoding.
This commit actually makes them purely memory as a temporary solution.

@mattbostock reopened

### DIFF
--- a/provider/boltmem/boltmem.go
+++ b/provider/boltmem/boltmem.go
@@ -16,9 +16,9 @@ package boltmem
 import (
 	"encoding/binary"
 	"encoding/json"
-	"fmt"
 	"path/filepath"
 	"sync"
+	"time"
 
 	"github.com/boltdb/bolt"
 	"github.com/prometheus/alertmanager/provider"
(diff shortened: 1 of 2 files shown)


test/acceptance/send_test.go
```diff
@@ -138,10 +139,11 @@ receivers:
...
 
-	at.Do(At(1.2), func() {
-		am.Terminate()
-		am.Start()
-	})
+	// XXX(fabxc): disabled as long as alerts are not persisted.
+	// at.Do(At(1.2), func() {
+	//	am.Terminate()
+	//	am.Start()
+	// })
```
mattbostock: Should we just remove this commented out code? It can always be pulled back out of Git's history when alerts are persisted again.

provider/boltmem/boltmem.go
```diff
@@ -30,38 +30,59 @@ import (
 var (
 	bktNotificationInfo = []byte("notification_info")
 	bktSilences         = []byte("silences")
-	bktAlerts           = []byte("alerts")
+	// bktAlerts           = []byte("alerts")
```
mattbostock: I think this could be removed. If we want to keep the comment, we should probably at least add an explanation as to why it's commented out.

fabxc: Yea, I know it's not exactly best practice. 
But I want to rework functional tests at some point and am worried to forget about it.

provider/boltmem/boltmem.go
```diff
@@ -128,89 +149,54 @@ func (a *Alerts) GetPending() provider.AlertIterator {
...
 
-		for k, v := c.First(); k != nil; k, v = c.Next() {
-			var a types.Alert
-			if err := json.Unmarshal(v, &a); err != nil {
-				return err
-			}
-			alerts = append(alerts, &a)
-		}
+	for _, alert := range a.alerts {
+		res = append(res, alert)
```
mattbostock: This could probably be simplified to `res = append(res, a.alerts...)`.

fabxc: Unfortunately not, `a.alerts` is a map.

mattbostock: Of course, thanks.

PR #423, opened 2016-07-08
Remove yaml js dependency

TheTincho:
This removes the need to parse YAML in JavaScript, by providing a pre-parsed configuration in the JSON data returned in `/api/v1/status`.


TheTincho:
Adding @fabxc in case it is needed. (sorry for the noise if it is not)

api.go
```diff
@@ -131,10 +139,12 @@ func (api *API) status(w http.ResponseWriter, req *http.Request) {
 
 	var status = struct {
 		Config      string            `json:"config"`
+		ConfigJson  config.Config     `json:"configJson"`
 		VersionInfo map[string]string `json:"versionInfo"`
 		Uptime      time.Time         `json:"uptime"`
 	}{
 		Config: api.config,
+		ConfigJson: api.configJson,
```
fabxc: Capitalized `JSON` for both.

TheTincho: Done. I also changed the visible JSON object member, please tell me if you prefer to keep the old one.

fabxc:
Okay, can you squash the commits and run `make assets` and `make format` before?
Good for merge then.

👍

TheTincho:
All is done now, but the CI seems to have some issues. I presume some test is being flaky.

PR #424, opened 2016-07-09
circleci : fix missing regex update for v-prefiexed tags

sdurrheimer:
I have also updated docker images `latest` tags to match the `v0.3.0`.

@fabxc


fabxc:
👍 Thanks

PR #426, opened 2016-07-12
OpsGenie Notifier, fixes and additions:

tcolgate:
- OpsGenie notifier confused the description and message fields
- Also added the notes field.
- Added verbose description template (taken from pushover


notify/impl.go
```diff
@@ -698,6 +702,7 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) error {
 		return err
 	}
 
+	log.With("incident", key).Debugf("post'ing: %v", buf.String())
```
juliusv: We probably don't want to keep this debug log line around?

juliusv:
:+1: otherwise.

notify/impl.go
tcolgate: Sorry, missed that one. Updated and pushed.

PR #429, opened 2016-07-15
Hide groups if all alerts within are silenced

mattbostock:
Currently, even if all alerts within a group are silenced, the group
summary still appears. This can take up a lot of space when many alerts
are silenced.

If all alerts in a group are silenced and 'Hide silences' is checked,
hide the empty group.


fabxc:
Again `make assets format` – sorry, we should probably make that happen automatically properly.

mattbostock:
Sure, updated.

fabxc:
Thanks 👍

PR #442, opened 2016-08-02
provider/mesh: set expiration time on create

fabxc:
@beorn7 as discussed, this sets an expiration timestamp on creation rather than every peer calculating the time at GC time.

For notification infos first to avoid duplicating mistakes in silences we might spot here.

---
This change is [Reviewable](https://reviewable.io/reviews/prometheus/alertmanager/442)


fabxc:
The merge tests just use a zero time as it doesn't really matter.
We could strip out expired records on merge (and sending for that matter). But this just duplicates GC logic essentially and should have negligible effect. Probably not worth the complexity.

beorn7:
On my review queue (just very very busy right now).

beorn7:
Forecast: Will review Friday on the train.

beorn7:
Didn't get enough done here on the train. Will try to use gaps in my vacation activities...

beorn7:
:+1: from the light review I did. After my vacations and conference marathon, I'll get deeper into HA AM...

PR #447, opened 2016-08-05
notify: eliminate select statement redundancy

tmrts:
cc @fabxc @brancz


fabxc:
FWIW, the PR referenced a long comment explaining why I was doing this.

tmrts:
@brancz @fabxc thanks, I missed the reference to that explanation :+1:

PR #449, opened 2016-08-08
notify: truncate HipChat from field

fabxc:
This commite truncates the `from` field in the HipChat REST call
to the maximum of 64 characters.

Closes #295 

@brian-brazil @brancz

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -584,7 +584,7 @@ func (n *Hipchat) Notify(ctx context.Context, as ...*types.Alert) error {
 	}
 
 	req := &hipchatReq{
-		From:          tmplText(n.conf.From),
+		From:          truncate(tmplText(n.conf.From), 64),
 		Notify:        n.conf.Notify,
 		Message:       msg,
 		MessageFormat: n.conf.MessageFormat,
@@ -956,4 +956,11 @@ func (a *loginAuth) Next(fromServer []byte, more bool) ([]byte, error) {
 		}
 	}
 	return nil, nil
+}
+
+func truncate(s string, l int) string {
+	if l <= 10 {
+		return s[:l]
+	}
+	return fmt.Sprintf("%s...", s[:l-3])
 }
--- a/notify/impl_test.go
+++ b/notify/impl_test.go
@@ -1,0 +1,32 @@
+// Copyright 2016 Prometheus Team
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package notify
+
+import "testing"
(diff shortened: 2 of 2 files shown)


notify/impl.go
```diff
@@ -957,3 +957,10 @@ func (a *loginAuth) Next(fromServer []byte, more bool) ([]byte, error) {
 	}
 	return nil, nil
 }
+
+func truncate(s string, l int) string {
```
brian-brazil: Didn't we already have a function for this?

fabxc: You requested it here https://github.com/prometheus/alertmanager/pull/295#discussion_r58917697

But I don't think we actually have it anywhere as this PR exists as the other one went stale.

notify/impl_test.go
```diff
@@ -0,0 +1,32 @@
...
+	cases := []struct {
+		in, out string
+		len     int
+	}{
+		{in: "a", len: 0, out: ""},
+		{in: "0123456789abcdef", len: 10, out: "0123456789"},
+		{in: "0123456789abcdef", len: 11, out: "01234567..."},
+	}
+	for _, c := range cases {
+		if have := truncate(c.in, c.len); have != c.out {
```
brancz: nit pick, but wdyt of `result` instead of `have`? And I know it's longer but maybe also `expectedLength` and `expectedOutput`? As I said, just nit pick, feel free to ignore...

notify/impl.go
```diff
@@ -584,7 +584,7 @@ func (n *Hipchat) Notify(ctx context.Context, as ...*types.Alert) error {
 	}
 
 	req := &hipchatReq{
-		From:          tmplText(n.conf.From),
+		From:          truncate(tmplText(n.conf.From), 64),
```
brancz: From just reading the code it's not completely clear why 64 chars. If I git blame I would find out. Is that the preferred way? Maybe either add a comment or move the 64 to a constant to describe it. Wdyt?

brian-brazil: It's in the spec: https://www.hipchat.com/docs/apiv2/method/send_room_notification

notify/impl_test.go
fabxc: Arguably, that's roughly the pattern we've been following for other unit tests (at least the ones I wrote :P).
The reason is simply that short names make the struct initialization above cleaner. This is more obvious in more complex and longer test cases.

I'd opt for sticking with the pattern from other parts of the code here.

notify/impl.go
fabxc: Yea, that should indeed be commented though.

fabxc: This makes me realize, we should also validate the other fields. We are going down quite the rabbit hole once we start getting into stuff like valid `color` fields etc.

I also cannot infer whether "length" refers to the number of (unicode) runes or the bytes in the encoded JSON string.

notify/impl.go
```diff
@@ -957,3 +957,10 @@ func (a *loginAuth) Next(fromServer []byte, more bool) ([]byte, error) {
 	}
 	return nil, nil
 }
+
+func truncate(s string, l int) string {
+	if l <= 10 {
+		return s[:l]
+	}
+	return fmt.Sprintf("%s...", s[:l-3])
```
sgissi: Won't `return s[:l-3]+"..."` be faster?

brian-brazil: I doubt it matters much, as after this we're going to be doing network traffic which will cost a lot more.

Issue #450, opened 2016-08-08
SMTP LOGIN authentication mechanism bug in V0.3.0


ZhenyangZhao:
Hi:
I found a bug in v0.3.0 when i testing. 
When AM send mail it always log "unexpected server challenge".
I think fromServer sometime like "username:", but the code compare like this "Username:"
Maybe password the same issue.
Testing mail server: "smtp.163.com:25"

Am i correct?

```
gofile: impl.go 
line: 949 
switch string(fromServer) {
case "Username:"
        return []byte(a.username), nil
case "Password:"
        return []byte(a.password), nil
default:
        return nil, errors.New("unexpected server challenge")
}
```

grobie:
Fixed with #451

PR #451, opened 2016-08-08
Modify SMTP login auth bug.

ZhenyangZhao:
Hi:
I found a bug in v0.3.0 when i testing. 
When AM send mail it always log "unexpected server challenge".
I think fromServer sometime like "username:", but the code compare like this "Username:"
Maybe password the same issue.
Testing mail server: "smtp.163.com:25"

Am i correct?

```
gofile: impl.go 
line: 949 
switch string(fromServer) {
case "Username:"
        return []byte(a.username), nil
case "Password:"
        return []byte(a.password), nil
default:
        return nil, errors.New("unexpected server challenge")
}
```

see #450

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -946,10 +946,10 @@ func (a *loginAuth) Start(server *smtp.ServerInfo) (string, []byte, error) {
 // Used for AUTH LOGIN. (Maybe password should be encrypted)
 func (a *loginAuth) Next(fromServer []byte, more bool) ([]byte, error) {
 	if more {
-		switch string(fromServer) {
-		case "Username:":
+		switch strings.ToLower(string(fromServer)) {
+		case "username:":
 			return []byte(a.username), nil
-		case "Password:":
+		case "password:":
 			return []byte(a.password), nil
 		default:
 			return nil, errors.New("unexpected server challenge")


fabxc:
Thanks.

I'm sure this works, but I spent a solid 15 minutes looking for an RFC or any formal documentation around `AUTH LOGIN` without success. This seems to be a Microsoft Exchange thing?

Anyway, all examples indicate it should be `Username` and no hints that it actually is case-insensitive. I would feel better about merging this if we had a good source on this.

ZhenyangZhao:
@fabxc 
Thanks a lot.
I test with different smtp server in my  country(China). Some smtp server "AUTH LOGIN" return "Username:" others return "username".

ZhenyangZhao:
@fabxc 
Thank you for reply.
As you said, i test with different smtp server in my country(China).I think you are right.
smtp.163.com:25 smtp.qq.com:25
When i send "AUTH LOGIN", 163 server return "username:", QQ server return "Username:".
The test pic as blow.
qq 20160808214034
qq 20160808214042
qq 20160808214049
qq 20160808214053

fabxc:
Thanks for the detail. Of course I believe you but a formal spec would have been great.
As this doesn't exist, let's do this based on real world application behavior.

👍

ZhenyangZhao:
@fabxc 
Thanks. :)

allenbang:
@ZhenyangZhao did you try to send mail by using QQ enterprise mail service?

ZhenyangZhao:
@allenbang  you mean exmail.qq.com? 

i just test mail.qq.com.

PR #454, opened 2016-08-09
*: restructure package tree

fabxc:
This commit packages up individual modules and removes the top-level
main package.

Aligns the rough structure with Prometheus and first step to allow us running several AMs in the same binary for better e2e tests.

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -38,12 +38,16 @@ import (
 	"github.com/prometheus/common/version"
 	"github.com/weaveworks/mesh"
 
+	"github.com/prometheus/alertmanager/api"
 	"github.com/prometheus/alertmanager/config"
+	"github.com/prometheus/alertmanager/dispatch"
+	"github.com/prometheus/alertmanager/inhibit"
 	"github.com/prometheus/alertmanager/notify"
 	"github.com/prometheus/alertmanager/provider/mem"
 	meshprov "github.com/prometheus/alertmanager/provider/mesh"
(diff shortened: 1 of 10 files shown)


cmd/alertmanager/main.go
```diff
@@ -216,8 +220,8 @@ func main() {
 		inhibitor.Stop()
 		disp.Stop()
 
-		inhibitor = NewInhibitor(alerts, conf.InhibitRules, marker)
-		disp = NewDispatcher(alerts, NewRoute(conf.Route, nil), build(conf.Receivers), marker)
+		inhibitor = inhibit.NewInhibitor(alerts, conf.InhibitRules, marker)
```
brancz: wdyt of applying the same scheme as for the `api` package here as in

```
inhibitor = inhibitor.New(...)
```

as the package doesn't contain anything but that..

brancz:
Looks good to me, and works for me :) 👍

cmd/alertmanager/main.go
fabxc: Technically that pattern is reserved if `lowercase(type) == package_name`. But I like verbs for package names :( What a Zwickmuehle.

brancz: I'm fine with either way, especially since it wouldn't work with the "Dispatcher" used above as there are multiple components to that package. And then mixing the two in consecutive lines of code would seem odd.

fabxc: Yea, I'll look into it again. This wasn't the last PR for this anyway.

PR #455, opened 2016-08-09
promu: readd build archs

fabxc:
We can build these binaries again with no longer depending on cgo.


sdurrheimer:
If i'm not mistaken we can remove the list completely from the file, isn't it ?

fabxc:
Even better. Thanks.

brancz:
:

Issue #293, opened 2016-04-04
Paginate and sort elapsed silences


grobie:
The elapsed silences are currently not paginated and sorted in the wrong order. This is both an UI problem (by having to scroll through an endless list of elapsed silences) and performance issue.

grobie:
Fixed with v0.4.0.

PR #456, opened 2016-08-10
Paginate silences

stuartnelson3:
~~somewhere between the first commit and the second commit I started getting nil pointer dereferences in `TestSilencesMutes`, which I'll look into. Also, there are no tests for this `lastID` thing.~~

~~Let me know what you think, and if the UI works. I added the `inflight` check to prevent duplicate requests being made to the backend -- apparently the throttling option in the code doesn't work very well.~~

Added:

- Pagination on backend
- Pagination on frontend
  - This added a couple [angular-ui dependencies](https://angular-ui.github.io/bootstrap/), which have been vendored.
  - This also added bootstrap.css, which has not been vendored. Pending your approval, I'll vendor it and plan on doing some more design work if desired.
- Restyling of silence create/update form

@fabxc 

**EDIT**

- Number of returned silences and page offset can be controlled through the url: `http://localhost:9093/#/silences?n=5&page=4`
- There's a default page size of 25
- ~~Don't be shy about closing this, it probably has little to do with how you're thinking of implementing paging + label searches~~


grobie:
I'm not a fan of infinite scrolling, it makes it hard to link people to a certain page.

stuartnelson3:
that makes sense to me, happy to remove it.

fabxc:
As silences are always changing, linking to pages doesn't seem to make too much sense.
Hard-linking a specific silence or an explicit list of them, seems to be the feature needed on top.

We certainly need pagination. Doesn't have to be infinite scrolling, but certainly something.

fabxc:
Also I merged the HA work into master just yesterday. Silences are no longer in Bolt or SQLite.
There also significant semantical changes around them. I'd recommend rebasing your changes on top, which shouldn't cause fundamental rework.

I'll take care of incorporating your the changes I'm working on.

stuartnelson3:
Is there a preference for pagination? Standard clicky-bar?

image

fabxc:
SGTM

grobie:
@fabxc Yes, being able to link to a specific silence (and alert!) is def. more important. I def. had the use case already for "hey, look at these 5 silences here <link>, do something like that". It's understood that these links are not stable, but better than "hey, open this link, scroll a lot, but not too much, and you'll see the 5 silences I want to show you"

stuartnelson3:
Things got a little out of hand. Adding pagination required bootstrap styling, so things started looking funny, so I updated a few forms .. Nothing major, but it definitely is looking different.

grobie:
Can you add a screenshot? Thanks for the work!

provider/mesh/peer.go
```diff
@@ -301,14 +301,45 @@ func (s *Silences) Mutes(lset model.LabelSet) bool {
 
 // All returns a list of all known silences.
 func (s *Silences) All() ([]*types.Silence, error) {
+	return s.Query(s.Count(), 0)
+}
+
+func (s *Silences) Count() int {
+	return len(s.st.m)
```
grobie: No read lock necessary?

api/api.go
```diff
@@ -359,15 +360,33 @@ func (api *API) delSilence(w http.ResponseWriter, r *http.Request) {
 }
 
 func (api *API) listSilences(w http.ResponseWriter, r *http.Request) {
-	sils, err := api.silences.All()
+	var (
+		num    = r.FormValue("n")
+		offset = r.FormValue("offset")
```
grobie: Offset based pagination has some disadvantages. Besides being quite slow with several relational-databases (which doesn't really matter here), it's not as stable as an after-id based approach. Using something like `?after=last-seen-id&limit=50` will ensure that the user will really see the next result when using pagination.

api/api.go
```diff
@@ -359,15 +360,33 @@ func (api *API) delSilence(w http.ResponseWriter, r *http.Request) {
...
 			typ: errorInternal,
 			err: err,
 		}, nil)
 		return
 	}
-	respond(w, sils)
+
+	respond(w, map[string]interface{}{
+		"silences":      sils,
+		"totalSilences": api.silences.Count(),
```
grobie: The silences and count get queried independently from each other with separate locks. This can result in different numbers being returned here. It should be better to return one type from `Query()` which includes both the list of requested silences as well as the total count.

provider/mesh/peer.go
stuartnelson3: ah so I guess I didn't think this part through, I assumed it was atomic without checking. I'll add the readlock.

api/api.go
stuartnelson3: That was my original approach with `lastID`, but then I wasn't sure how to handle users skipping ahead several pages in the pagination. I originally though it unlikely, but then moved towards calculating the next to be shown using `num * offset`. It's easy to re-introduce, but was unsure about supporting jumping forward `1+n` pages, and opted for that in this case. Happy to change it back if that's the vote.

As you pointed out, since it's not a relational db the offset is quick because it jumps to a position in a slice.

stuartnelson3:
image
image
image
image

api/api.go
grobie: You're right of course, we can't have the best of both worlds. What I described doesn't work for pagination by showing page numbers. Let's continue with offset then.

fabxc: I had this discussion with stuart in person. He made a point that for the order of a few thousand silences, this is unlikely to be a problem in the backend but gives huge benefits in the frontend right now.

Another iteration approach requires some fundamental changes in the backend storage, which isn't in its final form right now and doesn't support it.

ui/app/js/app.js
```diff
@@ -433,6 +493,14 @@ angular.module('am').config(
     when('/silences', {
       templateUrl: 'app/partials/silences.html',
       controller: 'SilencesCtrl',
+      resolve: {
+        silences: function(Silence) {
+          // Required to get the total number of silences before the controller
+          // loads. Without this, the user is forced to page 1 of the
+          // pagination.
+          return Silence.query({'n':0});
```
stuartnelson3: dummy request to just get the total number of silences, which is required to allow users to go to a page in the pagination that isn't 1.

stuartnelson3:
@fabxc I think this is ready for a review when you get the chance

provider/mesh/state.go
```diff
@@ -228,12 +230,17 @@ func (st *silenceState) gc(retention time.Duration) {
...
 	defer st.mtx.Unlock()
 
+	keys := []uuid.UUID{}
 	t := st.now().Add(-retention)
 	for k, v := range st.m {
 		if v.EndsAt.Before(t) {
 			delete(st.m, k)
+		} else {
+			// create new key cache from existing silences
+			keys = append(keys, v.ID)
```
fabxc: Note that iteration order over the map is random. Hence the key slice here will be random as well. I think it was intended to at least be sorted by creation time?

types/types.go
```diff
@@ -326,3 +327,15 @@ func (n *NotificationInfo) Fingerprint() model.Fingerprint {
 
 	return fp ^ n.Alert
 }
+
+// queryResponse is the data structure returned from the Query method.
+type QueryResponse struct {
```
fabxc: That's a very generic name for something entirely focused on Silences.

ui/app/index.html
```diff
@@ -9,12 +9,16 @@
...
 	<script src="lib/angular-moment.min.js"></script>
+	<script src="lib/ui-bootstrap-custom-tpls-2.0.1.min.js"></script>
 
 	<script src="app/js/app.js"></script>
 
 	<link rel="stylesheet" href="lib/kube.min.css">
 	<link rel="stylesheet" href="app/css/main.css">
 
+  <!-- vendor this -->
+  <link href="//netdna.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css" rel="stylesheet">
```
fabxc: Unrelated, but, should we just go back to a bootstrap UI entirely? @grobie would be happy I suppose.

stuartnelson3: I would be in favor of it. I'm not a designer, but it definitely helps things look more uniform when I use it.

types/types.go
stuartnelson3: This is very true. `SilenceQueryResponse`?

provider/mesh/state.go
stuartnelson3: Yeah, sorting needs to be added.

Is creation time appropriate for now?

fabxc:
In which browser are you testing this? Updated to the latest and trying to create a silence makes my tab crash on Chrome.

stuartnelson3:
that's some magic trick ...

I've been using chrome. I just regenerated assets one last time and pushed them. If this doesn't work, then your browser is probably caching the javascript. Go into developer tools, head to the network tab, and click "disable cache"

image

fabxc:
Just quickly, I haven't forgotten about you. I'm just rewriting the silencing layer in the same fashion as the notification log earlier this week. This is a good chance to figure out what we want now or in the close future.

We should probably tackle this directly based on the rewrite.
Do you want to get this in ASAP anyway? Otherwise we could grab lunch next week and talk through it in person a bit.

stuartnelson3:
Not a problem, I know you're busy :)

Let's do lunch next week and get this in based on how you want the backend to look. The frontend shouldn't change, and I have nothing against implementing the backend correctly. Plus, @grobie will be in Berlin, so we can use his sage advice as well.

I'm planning on hosting Mettwoch at SC if you're around at 1pm on Wednesday? Otherwise, Monday or Tuesday work for me as well. Thursday-Friday is PromCon.

grobie:
@fabxc Just as a note, the silence feature is pretty much unusable right now. And we only have a couple of tens of silences. I'd appreciate to not trying to rewrite everything first, but to solve the problems existing problems now.

grobie:
So if we're happy with the state of this PR, I'd vote to merge it now and rework it once the new backend is done.

stuartnelson3:
@grobie do you have any UX suggestions? I know those are your favorite

ui/app/index.html
grobie: I think this makes sense in general. My view is: half baked custom design < bootstrap < professional custom design. And given none of us is a frontender/designer, I don't see us ever beating bootstrap. So let's not even spend time on trying.

provider/mesh/state.go
grobie: Yes. All silences should be sorted descending.

That's also requested in https://github.com/prometheus/alertmanager/issues/293

fabxc:
Okay, so you want to cut a release with that then I suppose.
Because the current master is full-on development and won't be released before a few parts are rewritten. That's why it seemed more obvious to me.

In this case we should fork a `release-0.4` branch from `release-0.3` and PR the changes here onto that one.

grobie:
We won't necessarily need an official release, if that causes more issues.
We could just install some alpha version as well. I'm just afraid of
blocking these things even more, the current performance problems get
pretty much in the way when trying to silence pager storms during outages.

fabxc:
Yes, you are right.
Let's make it an official release then, though. The parts touched here shouldn't have drifted much from `release-0.3`.

ui/app/index.html
stuartnelson3: Shall I go ahead and vendor it in this PR then?

beorn7:
Also, if I got that right, a release from the master branch, would require porting the silences from the old to the new storage, which is (I assume) not trivial.

stuartnelson3:
Sorting is now occurring during gc, still an open question on if you all want bootstrap to be vendored in this PR. I added it in the latest commit, so feel free to merge this PR if you're happy, or revert the last commit if you want to use the CDN for now.

fabxc:
Yes, the JS stuff needs to be vendored. But as said above, if we want this in a stable release now, this has to be based on top of a `release-0.4` branch forked off of `release-0.3`.

fabxc:
Superseded by #468

PR #458, opened 2016-08-11
Pdbogen error on nonflag args

### LINKED ISSUE #457: alertmanager should fail when receiving unexpected command-line arguments
```
16:40 <      pdbogen> Hey guys, -web.external-url seems to be ignored for my alertmanager, at least for slack notifications with the default template. 
16:44 <@         jrv> pdbogen: hmm, weird, it shouldn't be: 
                      https://github.com/prometheus/alertmanager/blob/b822fe58b632ea3bae5c0b313030746b4ebb8e1c/template/default.tmpl#L2
16:44 <@         jrv> pdbogen: so maybe worth filing an issue
16:49 <@         jrv> pdbogen: hm yeah, just tracing the code from program initialization to notification templating, I don't see why it wouldn't use the 
                      external URL
16:49 <@         jrv> pdbogen: but I haven't run it
16:53 <@      grobie> this was broken in former versions of alertmanager. pdbogen, which version do you use?
17:08 <      pdbogen> It was user error; I build a docker container based on prom/alertmanager, but for some reason CMD with the "executable" syntax 
                      wasn't working for me. It ended up running /bin/alertmanager /bin/alertmanager <some args>. AM doesn't care about the non-flag 
                      argument, and flags doesn't parse past it. 
17:08 <      pdbogen> My actual settings were close enough to the default that it worked, except for ignoring my external-url. :) 
17:09 <@      grobie> can you create an issue please on alertmanager to fail on unexpected arguments
17:09 <      pdbogen> Sure. 
```

pdbogen:
Closes #457

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -88,6 +88,10 @@ func main() {
 	flag.Var(peers, "mesh.peer", "initial peers (may be repeated)")
 	flag.Parse()
 
+	if len(flag.Args()) > 0 {
+		log.Fatalln("Received unexpected and unparsed arguments: ", strings.Join(flag.Args(), ", "))
+	}
+
 	if *showVersion {
 		fmt.Fprintln(os.Stdout, version.Print("alertmanager"))
 		os.Exit(0)


pdbogen:
5684510ed65246509120c44a1afba2c1e09102bf is not directly related; it's some debug logging additions since this all came up because I was trying to trace an external-url that didn't seem to be honored. Please feel free to drop that commit if the additional debug logging is undesirable. (But I'm happy to amend if it it's close, just short of standards.)

fabxc:
Thanks!

I would indeed prefer to drop the debug logging.
I agree that debug logging is nice, but this particular case hopefully will no longer happen due to your fix.
In general, we just have one debug level and this is a good think as `-vvvvv` etc. are impossible to keep sane throughout a larger code base. But it also means we shouldn't establish doing debug logging everywhere. In particular not for user errors that should be minimized via validation.

fabxc:
Ah, thanks. I haven't seen that you updated.

Issue #382, opened 2016-06-14
Distinguish between backoff and permanent notification failure


fabxc:
Currently the Notifier interface can only error and we retry with exponential back off in any case.
This causes AM a misbehaving client (e.g
 400 HTTP errors).

Error inspection seems rather cumbersome.

I suggest introducing a new interface.

```
type NotificationFilter interface {
    Filter(alerts) (alerts, err)
}
```

For things like silencing, receiver routing, etc. and 

```
type Notifier interface {
    Notify(alerts) (bool, error)
}
```

For the actual notification integrations where the boolean indicates whether we should retry.

stuartnelson3:
fixed by #459

PR #459, opened 2016-08-11
refactor notification pipeline

brancz:
Move hard to read backwards declared pipeline approach to a more 
_in order_ approach with more detailed interfaces. Initially proposed by
@fabxc in #382.

Goal of this refactor is to have notification providers (pagerduty, slack, 
etc.) return errors that won't recover (eg. wrong api token) and the retry
functionality to stop trying completely as the error won't recover unless
the config is changed.

There is still a bunch of things to be moved, renamed and other cleanup,
comments, etc., but I wanted to get some early feedback, whether this
is the sort of direction we want to go into. All comments welcome 😃

### DIFF
--- a/inhibit/inhibit.go
+++ b/inhibit/inhibit.go
@@ -111,7 +111,7 @@ func (ih *Inhibitor) Stop() {
 	}
 }
 
-// Mutes returns true iff the given label set is muted.
+// Mutes returns true if the given label set is muted.
 func (ih *Inhibitor) Mutes(lset model.LabelSet) bool {
 	fp := lset.Fingerprint()
 
--- a/inhibit/inhibit.go
+++ b/inhibit/inhibit.go
@@ -111,7 +111,7 @@ func (ih *Inhibitor) Stop() {
 	}
 }
 
-// Mutes returns true if the given label set is muted.
+// Mutes returns true iff the given label set is muted.
 func (ih *Inhibitor) Mutes(lset model.LabelSet) bool {
 	fp := lset.Fingerprint()
 
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -37,6 +37,7 @@ import (
 	"golang.org/x/net/context/ctxhttp"
 
 	"github.com/prometheus/alertmanager/config"
+	meshprov "github.com/prometheus/alertmanager/provider/mesh"
 	"github.com/prometheus/alertmanager/template"
 	"github.com/prometheus/alertmanager/types"
 )
@@ -75,9 +76,9 @@ type integration interface {
 	name() string
 }
 
-// Build creates a fanout notifier for each receiver.
-func Build(confs []*config.Receiver, tmpl *template.Template) map[string]Fanout {
-	res := map[string]Fanout{}
+// BuildRouter builds fanouts and addts them to the router by receiver.
+func BuildRouter(confs []*config.Receiver, tmpl *template.Template, meshWait func() time.Duration, ni *meshprov.NotificationInfos) Router {
+	res := Router{}
 
 	filter := func(n integration, c notifierConfig) Notifier {
 		return NotifierFunc(func(ctx context.Context, alerts ...*types.Alert) error {
@@ -109,7 +110,13 @@ func Build(confs []*config.Receiver, tmpl *template.Template) map[string]Fanout 
 	for _, nc := range confs {
 		var (
 			fo  = Fanout{}
-			add = func(i int, on integration, n Notifier) { fo[fmt.Sprintf("%s/%d", on.name(), i)] = n }
+			add = func(i int, on integration, n Notifier) {
+				fo[fmt.Sprintf("%s/%d", on.name(), i)] = FanoutPipeline{
+					notificationInfos: ni,
+					meshWait:          meshWait,
+					notifier:          n,
+				}
+			}
 		)
 
 		for i, c := range nc.WebhookConfigs {
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -30,7 +30,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/log"
 	"github.com/prometheus/common/model"
 	"golang.org/x/net/context"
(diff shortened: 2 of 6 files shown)


inhibit/inhibit.go
```diff
@@ -111,7 +111,7 @@ func (ih *Inhibitor) Stop() {
 	}
 }
 
-// Mutes returns true iff the given label set is muted.
```
fabxc: This was supposed to be `iff` as in short for `if and only if` :)

notify/impl.go
```diff
@@ -75,9 +76,9 @@ type integration interface {
 	name() string
 }
 
-// Build creates a fanout notifier for each receiver.
-func Build(confs []*config.Receiver, tmpl *template.Template) map[string]Fanout {
-	res := map[string]Fanout{}
+// BuildRouter builds fanouts and addts them to the router by receiver.
```
fabxc: adds

notify/notify.go
```diff
@@ -242,21 +297,21 @@ func (n *DedupingNotifier) hasUpdate(alert *types.Alert, last *types.Notificatio
 	return true
 }
 
-// Notify implements the Notifier interface.
-func (n *DedupingNotifier) Notify(ctx context.Context, alerts ...*types.Alert) error {
+// ExtractNewNotifies filters out the notifications that shall be sent
```
fabxc: That rather describes the behavior this part had early on. Today it basically just checks whether even slightest thing changed in our alert batch and just resends everything if that's the case.

notify/notify.go
```diff
@@ -407,27 +440,5 @@ func (n *InhibitNotifier) Notify(ctx context.Context, alerts ...*types.Alert) er
 		}
 	}
 
-	return n.notifier.Notify(ctx, filtered...)
-}
-
-// LogNotifier logs the alerts to be notified about. It forwards to another Notifier
-// afterwards, if any is provided.
-type LogNotifier struct {
```
fabxc: Removed because no longer used?

Totally fair, this shouldn't exist anymore. If we want to do any debug logging between stages or at the very end, that should be done directly.

brancz: Yes no longer used. Just noticed myself while reviewing that I dropped it 😁 . I'll investigate where it makes most sense to add some debug logging.

notify/notify.go
brancz: I'm not completely happy with this design especially since we use returning `nil` as a special case to not continue the execution of the notification...

fabxc:
Okay that definitely goes into the right direction and now that this is fleshed out more clearly, I realize that I dropped some considerations in that initial issue.

My general idea was:

```go
stages := []NotificationFilter{Inhibit, Silence, ...}
err := StageSequence(stages...).Exec(ctx, alerts...)
// ...
```

This is, as you mentioned, broken by the route-fanout contraption in the middle, which only reveal at notification time which sequences of stages to take next (dedup, meshWait, retry, notify).

This seems unnecessary. The receiver is actually static for each [aggregation group](https://github.com/prometheus/alertmanager/blob/master/dispatch/dispatch.go#L248-L265). So we could materialize the stage pipeline as we create the aggregation group and remove this point of uncertainty.
We could assemble the stages above [here](https://github.com/prometheus/alertmanager/blob/master/dispatch/dispatch.go#L236-L242) and use them in the passed function. This would drag the assembling into `dispatch`, which seems better than  being shared between the `main` and `notify` package.

In total, with the fanout it would probably look something like this:

```go
type Stage interface {
    Exec(ctx context.Context, alerts ...*types.Alerts) ([]*types.Alerts, error)
}

// Implements Stage interface. Executes stages in sequence.
type MultiStage []Stage

// Implements Stage interface. Executes stages concurrenctly and returns
// after all have finished.
type FanoutStage []Stage

func createStage(receiverConfigurations, receiverName, ) Stage {
    var ms MultiStage
    // append fixed inhibit stage
    // append fixed silence stage

    // - formerly the router, can be statically resolved here...
    // - ... and we can jump right to the next stage

    var fs FanoutStage
    for integration := range receiverConfigurations[receiverName] {
        // create a MultiStage for each receiver finished by the retry stage
        // that directly calls the notifier.
        // Add the multi stage for the integration to the fanout stage.
    }
    // add fs to ms.
    return ms
}
```

This is roughly my idea after seeing the problems more clearly now.

I realize that this is mostly about making an existing working design "pretty", but I find it really important to make the overall code base straightforward to understand.

Thoughts?

brancz:
@fabxc makes a whole lot of sense. I like the idea of basically statically dispatching. Should even give a small performance improvement once a config is loaded. Huge advantage is IMO that it gives the pipeline immutability by config, makes it feel much more transparent.

brancz:
@fabxc applied your comment as discussed. I'm wondering if we actually want [this](https://github.com/brancz/alertmanager/blob/840dd7d2f5fef86ee493485ce834048bf156f5c4/notify/notify.go#L186-L194) or if it would be better to push that up to [here](https://github.com/prometheus/alertmanager/blob/master/dispatch/dispatch.go#L237). The current setup keeps the interface a bit cleaner imo. Wdyt?

notify/notify.go
```diff
@@ -293,72 +375,48 @@ func (n *DedupingNotifier) Notify(ctx context.Context, alerts ...*types.Alert) e
...
 }
 
-type WaitNotifier struct {
-	wait     func() time.Duration
-	notifier Notifier
+// WaitStage waits for a certain amount of time before continuing or until the
+// context is done.
+type WaitStage struct {
+	wait func() time.Duration
 }
```
fabxc: This can actually be `type WaitStage func() time.Duration`. 
But more of a taste thing than relevant :)

notify/impl.go
```diff
@@ -75,9 +75,9 @@ type integration interface {
...
 }
 
-// Build creates a fanout notifier for each receiver.
-func Build(confs []*config.Receiver, tmpl *template.Template) map[string]Fanout {
-	res := map[string]Fanout{}
+// BuildReceiverTree builds a tree of receivers and a unique id of notification providers.
+func BuildReceiverTree(confs []*config.Receiver, tmpl *template.Template) map[string]map[string]Notifier {
+	res := make(map[string]map[string]Notifier)
 
 	filter := func(n integration, c notifierConfig) Notifier {
```
fabxc: This function actually contains another 2 stages and it's hard to see when they run. The filtering of resolved notifications and the actual sending+instrumentation (which should probably stay together for sanity.

The filtering can go into its own stage. I think it's even wrong here. As we account for resolved alerts on deduplication but throw them away later. So we might send the same notification anyway.

notify/impl.go
```diff
@@ -108,8 +108,8 @@ func Build(confs []*config.Receiver, tmpl *template.Template) map[string]Fanout
 
 	for _, nc := range confs {
```
fabxc: When the above filtering is extracted, we can also process the logic of the loop below within createStage, I think, and don't have to build this tree out-of-band.

Might be overlooking something, though.

notify/impl.go
```diff
@@ -41,113 +40,77 @@ import (
...
+// given context.
+type Notifier interface {
+	Notify(context.Context, ...*types.Alert) error
 }
 
-type integration interface {
-	Notifier
-	name() string
+// An Integration wraps a notifier and it's config to be uniquely identified by
+// name and index from it's origin in the configuration
```
fabxc: Both `it's` to `its`. End comment with `.`

notify/impl.go
```diff
@@ -41,113 +40,77 @@ import (
...
-
 type notifierConfig interface {
 	SendResolved() bool
 }
 
-type NotifierFunc func(context.Context, ...*types.Alert) error
-
-func (f NotifierFunc) Notify(ctx context.Context, alerts ...*types.Alert) error {
-	return f(ctx, alerts...)
+// A Notifier is a type which notifies about alerts under constraints of the
```
fabxc: Omit "is a type which" – that's implied by the language semantics.

notify/impl.go
```diff
@@ -41,113 +40,77 @@ import (
...
 }
 
-type integration interface {
-	Notifier
-	name() string
+// An Integration wraps a notifier and it's config to be uniquely identified by
+// name and index from it's origin in the configuration
+type Integration struct {
+	notifier       Notifier
+	notifierConfig notifierConfig
```
fabxc: You've probably seen that the code base sticks with rather short identifier names as long as it's sufficiently clear by scope. I think `conf` is fine here and in other instances below.

notify/notify.go
```diff
@@ -119,109 +143,271 @@ func Now(ctx context.Context) (time.Time, bool) {
 	return v, ok
 }
 
-// A Notifier is a type which notifies about alerts under constraints of the
-// given context.
-type Notifier interface {
-	Notify(context.Context, ...*types.Alert) error
+// A Stage processes alerts constraints under the constraints of the given
+// context.
```
fabxc: Redundant "constraints".

notify/notify.go
```diff
@@ -119,109 +143,271 @@ func Now(ctx context.Context) (time.Time, bool) {
...
+type Stage interface {
+	Exec(ctx context.Context, alerts ...*types.Alert) ([]*types.Alert, error)
+}
+
+// Builds a RoutingStage as a map of receivers to Stages
+func BuildRoutingStage(confs []*config.Receiver, tmpl *template.Template, meshWait func() time.Duration, inhibitor *inhibit.Inhibitor, silences *meshprov.Silences, ni *meshprov.NotificationInfos, marker types.Marker) RoutingStage {
```
fabxc: This signature is quite long. That's okay for now but the style pattern for long arguments would be:

```go
+func BuildRoutingStage(
    confs []*config.Receiver,
    tmpl *template.Template,
    meshWait func() time.Duration,
    inhibitor *inhibit.Inhibitor,
    silences *meshprov.Silences,
    ni *meshprov.NotificationInfos,
    marker types.Marker,
) RoutingStage {
```

fabxc: I would argue that the name is a bit misleading here. In fact this function calls `createStage` and hence builds the entire nesting of stages. 

You call the result `pipeline` in `main()`. So maybe `BuildPipeline` might actually be a fitting name.

notify/notify_test.go
```diff
@@ -166,11 +166,10 @@ func TestDedupingNotifierHasUpdate(t *testing.T) {
 	}
 }
 
-func TestDedupingNotifier(t *testing.T) {
+func TestDedupingNewNotifiesExtraction(t *testing.T) {
```
fabxc: Leftover naming?

notify/notify.go
```diff
@@ -266,168 +452,101 @@ func (n *DedupingNotifier) Notify(ctx context.Context, alerts ...*types.Alert) e
...
-	case <-ctx.Done():
-		return ctx.Err()
-	}
-	return n.notifier.Notify(ctx, alerts...)
-}
-
-// Router dispatches the alerts to one of a set of
-// named notifiers based on the name value provided in the context.
-type Router map[string]Notifier
+// Exec implements the Stage interface
```
fabxc: Trailing `.`

notify/notify.go
```diff
@@ -119,109 +143,271 @@ func Now(ctx context.Context) (time.Time, bool) {
...
+
+	_, err := rs[receiver].Exec(ctx, alerts...)
+	return nil, err
+}
+
+// A MultiStage executes a series of stages sequencially.
+type MultiStage []Stage
+
+// Exec implements the Stage interface.
+func (ms MultiStage) Exec(ctx context.Context, alerts ...*types.Alert) ([]*types.Alert, error) {
```
fabxc: Test for this new stage required.

brancz: Yes I noticed as well and am already writing for the other new stages as well.

notify/notify.go
```diff
@@ -119,109 +143,271 @@ func Now(ctx context.Context) (time.Time, bool) {
...
+type RoutingStage map[string]Stage
+
+// Exec implements the Stage interface.
+func (rs RoutingStage) Exec(ctx context.Context, alerts ...*types.Alert) ([]*types.Alert, error) {
+	receiver, ok := Receiver(ctx)
+	if !ok {
+		return nil, fmt.Errorf("receiver missing")
+	}
+
+	_, err := rs[receiver].Exec(ctx, alerts...)
```
fabxc: Panics if there is no entry for the key. This shouldn't happen by design and panics are fair to occur for programmer errors. 
But I think the coupling between what ensures this invariant and this line is too loose to trust in that. I'd add a check here. 

And a test :)

notify/notify.go
```diff
@@ -119,109 +143,271 @@ func Now(ctx context.Context) (time.Time, bool) {
...
+		s = append(s, NewDedupStage(ni))
+		s = append(s, NewLogStage(log.With("step", "integration")))
+		s = append(s, NewRetryStage(i))
+		s = append(s, NewLogStage(log.With("step", "newNotifies")))
+		s = append(s, NewSetNotifiesStage(ni))
+		fs[fmt.Sprintf("%s/%d", i.name, i.idx)] = s
+	}
+
+	return append(ms, fs)
+}
```
fabxc: Yay, much cleaner look and feel and easy to understand :)

notify/notify.go
```diff
@@ -119,109 +143,271 @@ func Now(ctx context.Context) (time.Time, bool) {
...
+
+// createStage creates a MultiStage as a pipeline of stages for a receiver
+func createStage(receiverIntegrations []Integration, meshWait func() time.Duration, inhibitor *inhibit.Inhibitor, silences *meshprov.Silences, ni *meshprov.NotificationInfos, marker types.Marker) Stage {
+	var ms MultiStage
+	ms = append(ms, NewLogStage(log.With("step", "inhibit")))
+	ms = append(ms, NewInhibitStage(inhibitor, marker))
+	ms = append(ms, NewLogStage(log.With("step", "silence")))
+	ms = append(ms, NewSilenceStage(silences, marker))
+
+	var fs = make(FanoutStage)
```
fabxc: In Prometheus we standardized to avoid `make` whenever possible. So this would become "FanoutStage{}". Feels a bit ugly at first maybe, but standard > taste :)

notify/notify.go
```diff
@@ -119,109 +143,271 @@ func Now(ctx context.Context) (time.Time, bool) {
...
+	rs := make(RoutingStage)
+
+	for _, rc := range confs {
+		rs[rc.Name] = createStage(BuildReceiverIntegrations(rc, tmpl), meshWait, inhibitor, silences, ni, marker)
+	}
+
+	return rs
+}
+
+// createStage creates a MultiStage as a pipeline of stages for a receiver
```
fabxc: Trailing `.`

fabxc:
Great! A bunch of tiny nits and some missing tests, but looks good overall.

notify/notify.go
brancz: I don't have much of a preference, so will adapt :)

notify/notify.go
brancz: Added.

notify/notify.go
```diff
@@ -119,109 +143,283 @@ func Now(ctx context.Context) (time.Time, bool) {
...
-func (ns Fanout) Notify(ctx context.Context, alerts ...*types.Alert) error {
+	for _, rc := range confs {
+		rs[rc.Name] = createStage(BuildReceiverIntegrations(rc, tmpl), meshWait, inhibitor, silences, ni, marker)
+	}
+
+	return rs
+}
+
+// createStage creates a MultiStage as a pipeline of stages for a receiver.
+func createStage(receiverIntegrations []Integration, meshWait func() time.Duration, inhibitor *inhibit.Inhibitor, silences *meshprov.Silences, ni *meshprov.NotificationInfos, marker types.Marker) Stage {
```
fabxc: Same signature format as for `BuildPipeline` applies.

notify/notify.go
```diff
@@ -119,109 +143,283 @@ func Now(ctx context.Context) (time.Time, bool) {
...
-type Notifier interface {
-	Notify(context.Context, ...*types.Alert) error
+// A Stage processes alerts under the constraints of the given context.
+type Stage interface {
+	Exec(ctx context.Context, alerts ...*types.Alert) ([]*types.Alert, error)
 }
 
-// Fanout sends notifications through all notifiers it holds at once.
-type Fanout map[string]Notifier
+// BuildPipeline builds a RoutingStage as a map of receivers to Stages.
```
fabxc: Second `RoutingStage` should be dropped as well.

notify/notify_test.go
```diff
@@ -166,10 +170,78 @@ func TestDedupingNotifierHasUpdate(t *testing.T) {
 	}
 }
 
-func TestDedupingNewNotifiesExtraction(t *testing.T) {
+func TestMultiStage(t *testing.T) {
+	ctx := context.Background()
```
fabxc: Just used once below so we can just call it in the argument, no?

notify/notify_test.go
```diff
@@ -166,10 +170,78 @@ func TestDedupingNotifierHasUpdate(t *testing.T) {
...
+	stage := MultiStage{s1, s2}
+
+	out, err := stage.Exec(ctx, alerts...)
+	if err != nil {
+		t.Fatalf("Exec failed: %s", err)
+	}
+
+	if !reflect.DeepEqual(alerts, s1.alerts) ||
+		!reflect.DeepEqual(s1.alerts, s2.alerts) ||
+		!reflect.DeepEqual(s2.alerts, out) {
```
fabxc: Bit confused on this one. Aren't `alerts`, `s1.alerts`, and `s2.alerts` all static and not changed in `equalityStage` either? Why the comparisons?

I think what we really want to check here is a) that the stages are evaluated in the correct order, and b) that the output of the previous one is the input of the next one.

Maybe we can construct a `type StageFunc func(context.Context, ...*types.Alert) ([]*types.Alert, error)` that allows us to define new mock stages inline in this test.

Something like 

```go
var (
    alerts1 = []*types.Alert{{}}
    alerts2 = []*types.Alert{{}, {}}
    alerts3 = []*types.Alert{{}, {}, {}}
)

stage := MultiStage{
    StageFunc(func(context.Context, alerts ...*types.Alert) ([]*types.Alert, error) {
        if !equal(alerts, alerts1) {
            // something went wrong
        }
        return alerts2
    }),
}

fabxc:
Okay, last round :) ready after that.

notify/notify_test.go
brancz: Yes that makes more sense. I was a bit unsure how to test this...

fabxc:
🚢 🎉

PR #463, opened 2016-08-16
Reduce runtime context dependencies

fabxc:
@brancz 

This uses the first object of the new `nflog` package (separate PR which is the base of this one). As we built the pipeline per receiver on configuration load, there's no need to make these runtime dependencies.
After all, `context.Context` should not be used as a disguise for `interface{}` (or Python kwargs).
As a consequence, the fanout stage is now truly just executing stages in parallel, without knowledge about integrations.

Thought about where to draw the boundary. There are other values in the context too. We could also build a static pipeline as a new aggregation group gets created in the dispatcher. I figured a reasonably boundary would be that we build the pipeline on configuration load time and build static pipelines only based on the information we have at this point in time.

I also removed the logging stages, as they we didn't want to keep these around anyway.

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -60,16 +60,16 @@ const MinTimeout = 10 * time.Second
 type notifyKey int
 
 const (
-	keyReceiver notifyKey = iota
+	keyReceiverName notifyKey = iota
 	keyRepeatInterval
 	keyGroupLabels
 	keyGroupKey
 	keyNow
 )
 
-// WithReceiver populates a context with a receiver.
-func WithReceiver(ctx context.Context, rcv string) context.Context {
-	return context.WithValue(ctx, keyReceiver, rcv)
+// WithReceiverName populates a context with a receiver name.
+func WithReceiverName(ctx context.Context, rcv string) context.Context {
+	return context.WithValue(ctx, keyReceiverName, rcv)
 }
 
 // WithRepeatInterval populates a context with a repeat interval.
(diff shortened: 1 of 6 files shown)


notify/notify.go
```diff
@@ -160,53 +161,42 @@ func (f StageFunc) Exec(ctx context.Context, alerts ...*types.Alert) ([]*types.A
...
-	ms = append(ms, NewInhibitStage(inhibitor, marker))
-	ms = append(ms, NewLogStage(log.With("step", "silence")))
-	ms = append(ms, NewSilenceStage(silences, marker))
-
-	var fs = FanoutStage{}
-	for _, i := range receiverIntegrations {
+func createStage(rc *config.Receiver, tmpl *template.Template, wait func() time.Duration, ni *meshprov.NotificationInfos) Stage {
+	var fs FanoutStage
+	for _, i := range BuildReceiverIntegrations(rc, tmpl) {
+		recv := &nflogpb.Receiver{
```
brancz: This is nice, I was thinking of introducing an object like that as well.

notify/notify.go
```diff
@@ -248,42 +238,31 @@ func (ms MultiStage) Exec(ctx context.Context, alerts ...*types.Alert) ([]*types
 }
 
 // FanoutStage executes its stages concurrently
-type FanoutStage map[string]Stage
+type FanoutStage []Stage
 
-// Exec attempts to execute all stages concurrently. It returns a
-// types.MultiError if any of them fails.
+// Exec attempts to execute all stages concurrently and discards the results.
+// It returns its input alerts and a types.MultiError if one or more stages fails.
```
brancz: I believe it should be `... if one or more stages fail.`

brancz:
LGTM 👍

fabxc:
Thanks, holding off merge, as the base is still in review.

PR #464, opened 2016-08-17
notify: include context in Stage interface

fabxc:
This adds context.Context to the return arguments of a Stage.
This is necessary to propagate modified contexts.

@brancz

 depends on #463


brancz:
If I understand correctly then the important part here is that the `MultiStage` receives and forwards the context from one `Stage` to another. If so LGTM 👍

fabxc:
Test adjusted, PTAL.

brancz:
LGTM 👍

PR #467, opened 2016-08-18
*: integrate nflog

fabxc:
This commit replaces the previous NotifyInfo provider with the new
nflog package. It needs adjustments in the behavior of the deduping
stage.
The nflog stores notification digests per receiver per alert aggregation
group rather than one entry for alert per receiver. This drastically
reduces the number of entries and removes interference
across aggregation groups.

@brancz

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -14,11 +14,15 @@
 package notify
 
 import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/binary"
 	"fmt"
 	"sync"
 	"time"
 
 	"github.com/cenkalti/backoff"
+	"github.com/golang/protobuf/ptypes"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/log"
 	"github.com/prometheus/common/model"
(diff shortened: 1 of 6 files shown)


notify/notify.go
```diff
@@ -384,74 +401,127 @@ func (fr *FilterResolvedStage) Exec(ctx context.Context, alerts ...*types.Alert)
...
-			// the repeat interval has already passed.
-			if !now.After(last.Timestamp.Add(interval)) {
-				return false
-			}
+func utcNow() time.Time {
+	return time.Now().UTC()
+}
+
+// TODO(fabxc): this could get slow, but is fine for now. We may want to
+// have something mor sophisticated at some point.
```
brancz: more

notify/notify_test.go
```diff
@@ -235,86 +281,7 @@ func TestRoutingStage(t *testing.T) {
 }
 
 func TestSetNotifiesStage(t *testing.T) {
```
brancz: missing test?

fabxc: Ah, that was not intended.

fabxc: PTAL

brancz:
LGTM 👍

PR #468, opened 2016-08-22
Add pagination for silences

stuartnelson3:
@fabxc @grobie @beorn7 

for your viewing pleasure


api.go
```diff
@@ -373,15 +373,30 @@ func (api *API) delSilence(w http.ResponseWriter, r *http.Request) {
 }
 
 func (api *API) listSilences(w http.ResponseWriter, r *http.Request) {
-	sils, err := api.silences.All()
+	var (
+		num    = r.FormValue("n")
```
grobie: nitpick, but why is one parameter in the long form and the other short? I'd vote for consistency. Maybe use `page` instead?

grobie: Or page and limit. These are quiet common.

stuartnelson3: I couldn't think of a good short value for `offset`, but `n` seemed comfortable.

The original idea was to present `page` to the user in the URL, and to use `offset` in the API. Maybe it's confusing, but it made sense to me initially to view pages starting at 1, but using a zero based offset in the api.

- happy to change `n` to limit
- where do you stand on starting at 0 or 1? I don't really like showing `page=0` to the user, but I would prefer that to sending `page=1` to the API and then having to subtract one there for pulling IDs from the cached ID slice.

grobie: limit + offset would be fine as well. I like the idea of using page (starting at 1) externally and offset (starting at 0) internally.

Do we actually need a configurable limit to begin with? Could we just go with 25 for now and keep it simple until we have enough user requests to make it configurable?

stuartnelson3: Doesn't matter to me

stuartnelson3: I pushed a commit that removes the limit from the user's view. Let me know what you think.

provider/boltmem/boltmem.go
```diff
@@ -257,15 +260,47 @@ func (s *Silences) Mutes(lset model.LabelSet) bool {
...
-	for _, s := range s.cache {
-		res = append(res, s)
+	klen := len(s.k)
+
+	if klen < n {
+		n = klen
 	}
-	return res, nil
+
+	pageStart := n * o
```
fabxc: So `o` is the page-wise offset? That should be clarified in the doc string.

provider/boltmem/boltmem_test.go
```diff
@@ -559,3 +559,131 @@ func notifyInfoListEqual(n1, n2 []*types.NotifyInfo) bool {
...
+		}
+
+		s := res.Silences
+
+		start := p.offset * p.n
+		end := start + p.n
+		if end > n {
+			t.Fatalf("your test data doesn't include the range you're requesting: insert[%d:%d] (max index %d)", start, end, n)
+		}
+		expected := append([]*types.Silence{}, insert[start:end]...)
```
fabxc: can just be `expected := insert[start:end]`. The append would only help if we were potentially modifying the insert slice afterwards. But that's not happening here.

fabxc:
This seems to work fine, thanks.

provider/boltmem/boltmem.go
stuartnelson3: `// Query returns n silences starting at offset o.` is what's currently in there. Would this be a good clarification?
```
// Query returns n silences starting at page offset o.
```

fabxc: SGTM.

fabxc:
👍 after that.

PR #470, opened 2016-08-23
Correctly resolve silences query from server

stuartnelson3:
@fabxc @grobie this fixes the race condition that existed when moving between tabs within alertmanager. They didn't manifest for me locally because the response time was low enough that it wasn't an issue.


ui/app/js/app.js
```diff
@@ -266,8 +266,8 @@ angular.module('am.controllers').controller('SilenceCtrl',
   }
 );
 
-angular.module('am.controllers').controller('SilencesCtrl', function($scope, $location, Silence, silences) {
-  $scope.totalItems = silences.data.totalSilences;
+angular.module('am.controllers').controller('SilencesCtrl', function($scope, $location, Silence, totalSilences) {
+  $scope.totalItems = totalSilences;
```
grobie: Can you elaborate where that comes from and why we now that upfront?

grobie:
Yeah, I guess it's fine. I'd need to read more of the code than just looking at the snippet. LGTM if it works.

fabxc:
LGTM, but please remove the `release-0.4.1` branch. This still goes to `release-0.4` and the `v0.4.1` tag is attached to a commit in that branch.

ui/app/js/app.js
stuartnelson3: Discussed with grobie in person. This value is injected from the `$routeProvider`. The code has been fixed so that it properly waits for the promise in the router to resolve before loading the controller.

PR #471, opened 2016-08-24
Update silences page

stuartnelson3:
@fabxc @grobie This updates the silences page, and also returns the spacing on the alerts page to more or less what it was before. I didn't re-do the alerts page because fabian talked with me briefly about a few ideas to make it more usable, so I'll wait on updating that until we have a chance to talk about it more.


stuartnelson3:
Updated visuals:
image

grobie:
LGTM

PR #472, opened 2016-08-24
Add Plivo Notifier.

swerter:
This is a simple notifier, which rings your phone in case of an exception. 

Our use case is the following:
Receiving emails or other notifications are nice. However, in the middle of the night the best (worst?) way to get your attention is by ringing your phone. Plivio has a simple API to make your phone ring and gives you free credit. 

Let me know what you think of it.

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,0 +1,11 @@
+FROM golang:alpine
+
+RUN apk --update --no-cache add git make
+
+# RUN mkdir -p $GOPATH/src/github.com/prometheus &&\
+#     cd $GOPATH/src/github.com/prometheus &&\
+#     git clone https://github.com/prometheus/alertmanager.git
+
+WORKDIR $GOPATH/src/github.com/prometheus/alertmanager
+
+CMD ["make", "build"]
--- a/config/config.go
+++ b/config/config.go
@@ -228,6 +228,26 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 				voc.APIURL += "/"
 			}
 		}
+		for _, pli := range rcv.PlivoConfigs {
+			if pli.APIURL == "" {
+				if c.Global.PlivoAPIURL == "" {
+					return fmt.Errorf("no global Plivo APIURL set")
+				}
+				pli.APIURL = c.Global.PlivoAPIURL
+			}
+			if pli.AuthId == "" {
+				if c.Global.PlivoAuthId == "" {
+					return fmt.Errorf("no global plivo AuthId set")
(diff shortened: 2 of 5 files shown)


brian-brazil:
Thanks for the PR. There's so many services like this out there that I don't think it's practical for us to start supporting all of them. I'd suggest integrating via the Generic Webhook. I believe there's already a project started for the various SMS providers.

swerter:
I understand your concert, there are indeed many, many services. On the other hand, so far you have no single service that's only ringing your phone (for free), which is why I added it. 

Concerning your suggestion to use the webhook, how would you make such a call this via the generic webhook with basic authentication and specific parameters? 

    curl -u asdfasdf:asdf -H "Content-Type: application/json" -d '{"from":"123456","to":"123456","answer_url":"https://s3.amazonaws.com/static.plivo.com/answer.xml","answer_method":"GET"}' https://api.plivo.com/v1/Account/ABC/Call/

At least I didn't find a way how to do it. But I'm more than happy if you have a good idea.

And finally, where do you draw the lines which services to include and which ones not? But yes, I understand, this is finally your decision. In any case thanks for building prometheus, very nice set of applications.

brian-brazil:
The idea would be to run a small server that takes in the webhook and sends out a request to Plivo.

swerter:
I was thinking first about setting up a simply proxy, that does exactly as you say. Would have also been the easiest solution. I rejected it so I don't have to monitor yet another service. And it makes more sense to have it directly in the alertmanager, in my opinion. 

Now if you dislike the idea to merge it into the code, I can live with that, I just use my fork of the alertmanager. My suggestion in that case would rather be to implement a plugin system, because this issue will come up again with every additional notifier someone builds and you dislike, ie "why is notifier X supported natively and Y not?". This way you can support any notifier. At the end you also make alertmanager more compelling the more notification systems you support. But just my thoughts.

brian-brazil:
The generic webhook is intended as our plugin for the Alertmanager.

swerter:
I see. Then how about to make it a bit more flexible, eg add specific values to the data part of the, set headers etc.?

brian-brazil:
That's getting to reinventing XSLT but for JSON, which would be creating another Turing Complete language within the Prometheus ecosystem. One of those is enough.

swerter:
:) 
I see, we're getting into a Pattsituation :) 
Well, anyway, what shall I do? Close the PR? Or you still consider to merge it? Or just leave it open for now?

brian-brazil:
If you could close the PR that'd be good. We try to limit things to the most popular integrations.

PR #475, opened 2016-08-24
Return silences by bucket

stuartnelson3:
This PR introduces two things:

- Sorting silences by `CreatedAt`, `StartsAt`, and `EndsAt` (through `SilencesSorter`)
- Silences are sorted by `active`, `pending`, and `elapsed`, with each bucket being sorted by `EndsAt` (descending) right now.

So, before we go all O(n) on this PR, because it is being sorted like this on EVERY request, I'll say that it's massively faster than the non-paginated version :)

Given the high-availability future, I would prefer to wait for that branch to manage individual caches of silences in their various states. But, discuss.

Note: When looking at the order dropdown for search, I noticed that the value `$scope.order` was not updated by changing the dropdown. I mentioned it to @grobie and he said to delete it for now. It's easy to add back in, and re-work it so that instead of sorting the on-page silences, its value is appended to the API request.

@fabxc @beorn7


provider/boltmem/boltmem.go
```diff
@@ -284,25 +284,47 @@ func (s *Silences) Query(n, o int) (*types.SilencesQueryResponse, error) {
...
+
 	return &types.SilencesQueryResponse{
-		Silences:      silences,
-		TotalSilences: klen,
+		Silences:      copiedSils[pageStart:pageEnd],
+		TotalSilences: len(copiedSils),
 	}, nil
 }
 
+func (s *Silences) toBuckets() (active []*types.Silence, pending []*types.Silence, elapsed []*types.Silence) {
```
stuartnelson3: please suggest a better name if you want to stick with this method. happy with anything, maybe even just `s.buckets()`.

provider/boltmem/boltmem.go
```diff
@@ -261,15 +261,15 @@ func (s *Silences) Mutes(lset model.LabelSet) bool {
 
 // All returns all existing silences.
 func (s *Silences) All() (*types.SilencesQueryResponse, error) {
-	return s.Query(len(s.k), 0)
+	return s.Query(len(s.sils), 0, types.ByCreatedAt)
```
grobie: This still seems to be lacking the lock of the mutex.

stuartnelson3: It calls into `s.Query()` which immediately acquires the lock.

grobie: It accesses `s.sils` without having a lock acquired.

provider/boltmem/boltmem.go
```diff
@@ -261,15 +261,15 @@ func (s *Silences) Mutes(lset model.LabelSet) bool {
...
 }
 
 // Query returns n silences starting at page offset o.
-func (s *Silences) Query(n, o int) (*types.SilencesQueryResponse, error) {
+func (s *Silences) Query(n, o int, fn types.SilencesLessFunc) (*types.SilencesQueryResponse, error) {
```
grobie: Can be a follow-up commit, but `uint` would be the better type here. n/o are not descriptive variable names to be used in a exported function signature, I'd suggest `limit, offset uint, ...`

provider/boltmem/boltmem.go
grobie: `active, pending elapsed []*types.Silence`

provider/boltmem/boltmem.go
stuartnelson3: ahhh roger. I need to get that value inside a lock before calling into `Query` or it will deadlock, right? Since the lock is requested also within that function. So ..

```go
s.mtx.RLock()
n := len(s.sils)
s.mtx.RUnlock()
return s.Query(n, 0, types.ByCreatedAt)
```
?

provider/boltmem/boltmem.go
stuartnelson3: I understand the second point regarding descriptive names, but why would `uint` be better? Is it because we'll never need a negative value?

grobie: Yeah, so let's leverage the type system to guarantee that.

provider/boltmem/boltmem.go
stuartnelson3: That doesn't solve the problem either.

I'm personally inclined to just remove `All()` from the interface, since if a user is REALLY wanting all the silences, they can query the listSilences endpoint with some arbitrarily large `limit` value.

stuartnelson3:
I'm heading out for vacation for the next week and I won't have access to my computer. If any changes need to be made feel free to go for it and merge this.

Once this is merged into `release-0.4`, the new v0.4.1 should be ready to release.

grobie:
:+1: Looks good to me. I can take care of the release unless @fabxc has any comments.

PR #481, opened 2016-08-30
*: integrate new silence package

fabxc:
For now this just maps the current REST API into the new package around the semantic difference around comments.


beorn7:
:+1: (at a first glance).

PR #486, opened 2016-09-04
Rename simple.yml to alertmanager.yml

### LINKED ISSUE #484: simple.yml should be alertmanager.yml
Unlike other prometheus binaries with config files, simply running the alertmanager binary from the tarball doesn't work. We should rename simple.yml to alertmanager.yml so this works out of the box

juliusv:
Fixes #484


brian-brazil:
The promu config also need updating

fabxc:
The file was always just showing all the things you can configure. It's a
nonsensical out-of-the-box configuration that will give you a
non-functional AM instance.

Making it actually start without even looking at it makes it just worse for
users, I think.
If we want this, it should be a simpler configuration that generally makes
sense.

stuartnelson3:
Closing since it doesn't seem to be getting merged

PR #487, opened 2016-09-04
allow CORS for api GET and OPTIONS

vanesa:
Ideally I would also like to access POST and DELETE methods, but that probably doesn't make sense from a security perspective?


fabxc:
Thanks 👍

PR #489, opened 2016-09-05
*: consider mesh wait in notification timeouts

### LINKED ISSUE #488: HA notification failover collides with regular timeout
In an HA setup, Alertmanagers wait with notifying depending on their peer ordering.
For example, with 3 peers, the A send notification right away, B with 5 second delay, C with 10 second delay.

If `timeout := max(group_interval, notification_timeout); timeout > peer_wait` the context will time out before we have a chance to start sending a notification. In this scenario these Alertmanagers would never send a notification if the other peers are down.

The easiest solution to extend `timeout` by `peer_wait`. This will cause overlaps in the notification pipeline, i.e. while we still try to send the current notification, the next one is kicked off already.
For now this seems fine. Overwriting the contents of an in-progress notification  by more recent data introduces fragility. Generally the `group_wait` should be large enough for this never to happen.
If it's not, the difference is probably not large enough for the staleness to be even noticeable by the user.

I'd go ahead with the simple solution unless there are concerns.

@brian-brazil @brancz @beorn7

fabxc:
This adds the peer wait duration to the standard timeout to avoid
terminating a notification prematurely while being in failover
wait status.

Fixes #488 

@brancz @beorn7

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -183,6 +183,14 @@ func main() {
 		log.Fatal(err)
 	}
 
+	waitFunc := meshWait(mrouter, 5*time.Second)
+	timeoutFunc := func(d time.Duration) time.Duration {
+		if d < notify.MinTimeout {
+			d = notify.MinTimeout
+		}
+		return d + waitFunc()
+	}
+
 	reload := func() (err error) {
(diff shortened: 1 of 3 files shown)


brancz:
If no one can spot a problem with the general solution then LGTM in terms of implementation.

PR #492, opened 2016-09-06
add retry flag for notify providers

### LINKED ISSUE #311: using an unknown template makes AM flood your $alerttarget

brancz:
The retry flag allows an integration to specify whether a retry can
potentially be solved or if the error is likely not going to recover.
For example invalid authentication is likely a wrong configuration and
therefore a retry would not make sense, while a server error is likely
a temporary problem and can potentially be solved on the next retry.

It would be good to review all the added flags that enable/disable
retry, especially the ones in the email notifier.

@fabxc @beorn7 @brian-brazil 

Fixes #311

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -45,8 +45,10 @@ type notifierConfig interface {
 }
 
 // A Notifier notifies about alerts under constraints of the given context.
+// It returns an error if unsuccessful and a flag whether the error is
+// recoverable. This information is useful for a retry logic.
 type Notifier interface {
-	Notify(context.Context, ...*types.Alert) error
+	Notify(context.Context, ...*types.Alert) (bool, error)
 }
 
 // An Integration wraps a notifier and its config to be uniquely identified by
@@ -59,7 +61,7 @@ type Integration struct {
 }
 
 // Notify implements the Notifier interface.
-func (i *Integration) Notify(ctx context.Context, alerts ...*types.Alert) error {
+func (i *Integration) Notify(ctx context.Context, alerts ...*types.Alert) (bool, error) {
 	return i.notifier.Notify(ctx, alerts...)
 }
 
@@ -137,7 +139,7 @@ type WebhookMessage struct {
 }
 
 // Notify implements the Notifier interface.
-func (w *Webhook) Notify(ctx context.Context, alerts ...*types.Alert) error {
+func (w *Webhook) Notify(ctx context.Context, alerts ...*types.Alert) (bool, error) {
 	data := w.tmpl.Data(receiverName(ctx), groupLabels(ctx), alerts...)
 
 	groupKey, ok := GroupKey(ctx)
@@ -153,20 +155,26 @@ func (w *Webhook) Notify(ctx context.Context, alerts ...*types.Alert) error {
 
 	var buf bytes.Buffer
 	if err := json.NewEncoder(&buf).Encode(msg); err != nil {
-		return err
+		return false, err
 	}
 
 	resp, err := ctxhttp.Post(ctx, http.DefaultClient, w.URL, contentTypeJSON, &buf)
 	if err != nil {
-		return err
+		return true, err
 	}
 	resp.Body.Close()
 
-	if resp.StatusCode/100 != 2 {
-		return fmt.Errorf("unexpected status code %v from %s", resp.StatusCode, w.URL)
+	return w.retry(resp.StatusCode)
+}
+
+func (w *Webhook) retry(statusCode int) (bool, error) {
+	// Webhooks are assumed to respond with 2xx response codes on a successful
+	// request and 5xx response codes are assumed to be recoverable.
+	if statusCode/100 != 2 {
+		return (statusCode/100 == 5), fmt.Errorf("unexpected status code %v from %s", statusCode, w.URL)
 	}
 
-	return nil
+	return false, nil
 }
 
 // Email implements a Notifier for email notifications.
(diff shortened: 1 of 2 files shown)


notify/impl.go
```diff
@@ -153,20 +155,24 @@ func (w *Webhook) Notify(ctx context.Context, alerts ...*types.Alert) error {
...
 	var buf bytes.Buffer
 	if err := json.NewEncoder(&buf).Encode(msg); err != nil {
-		return err
+		return false, err
 	}
 
 	resp, err := ctxhttp.Post(ctx, http.DefaultClient, w.URL, contentTypeJSON, &buf)
 	if err != nil {
-		return err
+		return false, err
```
brancz: wdyt about these cases, recoverable or not?

fabxc: I went through the code chain again. This is mostly not recoverable (malformed requests etc.) – however, there are some points like the server returning an HTTP response to an HTTPS request etc., that could in theory be resolved on the side of the integration.

I'd vote to make it retry until someone reports a problem. Inspecting the error for further differentiation is probably not worth the effort. Retrying is not exactly costly for us.

notify/impl.go
```diff
@@ -153,20 +155,24 @@ func (w *Webhook) Notify(ctx context.Context, alerts ...*types.Alert) error {
...
 	}
 	resp.Body.Close()
 
-	if resp.StatusCode/100 != 2 {
-		return fmt.Errorf("unexpected status code %v from %s", resp.StatusCode, w.URL)
+	if resp.StatusCode/100 == 5 {
+		return true, fmt.Errorf("unexpected status code %v from %s", resp.StatusCode, w.URL)
+	}
+
+	if resp.StatusCode/100 != 2 && resp.StatusCode/100 != 3 {
```
fabxc: Redirects are resolved by the HTTP client, so we cannot encounter them at this point anymore. Also, if we had received a redirect at this point, we would not have sent anything and returns `true, nil` would not really be sane.

notify/impl.go
```diff
@@ -390,24 +399,31 @@ func (n *PagerDuty) Notify(ctx context.Context, as ...*types.Alert) error {
...
 	resp, err := ctxhttp.Post(ctx, http.DefaultClient, n.conf.URL, contentTypeJSON, &buf)
 	if err != nil {
-		return err
+		return false, err
 	}
 	resp.Body.Close()
 
+	// retry can solve the issue on 403 (rate limiting) and 5xx response codes
+	// https://v2.developer.pagerduty.com/docs/trigger-events
+	if resp.StatusCode == 403 || resp.StatusCode/100 == 5 {
```
fabxc: Oh my god – 403 for rate limiting?
I was just thinking about generalizing the retry inspection in its own function...

notify/impl.go
```diff
@@ -651,40 +680,45 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) error {
...
+		return false, fmt.Errorf("unexpected status code %v", resp.StatusCode)
 	} else if resp.StatusCode/100 != 2 {
 		body, _ := ioutil.ReadAll(resp.Body)
 		log.With("incident", key).Debugf("unexpected OpsGenie response from %s (POSTed %s), %s: %s",
 			apiURL, msg, resp.Status, body)
-		return fmt.Errorf("unexpected status code %v", resp.StatusCode)
+		return false, fmt.Errorf("unexpected status code %v", resp.StatusCode)
 	}
-	return nil
+	return true, nil
```
fabxc: Mh, shouldn't this default to false everywhere when everything went alright?

notify/impl.go
brancz: pd is even one of the better APIs if you ask me ... I guess a function per provider to decide whether to retry or not makes sense, but might not work for all providers..

notify/impl.go
brancz: yes, of course .. missed to remove that one, it was one of my first thoughts and removed it in all the other ones..

notify/notify.go
```diff
@@ -575,9 +575,13 @@ func (r RetryStage) Exec(ctx context.Context, alerts ...*types.Alert) (context.C
...
 		select {
 		case <-tick.C:
-			if err := r.integration.Notify(ctx, alerts...); err != nil {
+			if retry, err := r.integration.Notify(ctx, alerts...); err != nil {
 				numFailedNotifications.WithLabelValues(r.integration.name).Inc()
 				log.Warnf("Notify attempt %d failed: %s", i, err)
+				if !retry {
+					log.Warnf("Cancelling notify retry due to unrecoverable error: %s", err)
+					return ctx, alerts, nil
+				}
```
fabxc: We should probably make the first warning log a debug level log and show a final error after failing permanently. Either by no-retry or by timing out in retry (log last error then).

notify/impl.go
brancz: sgtm

beorn7:
This is on my review queue but I'm currently a bit overwhelmed with my day job. No ETA. Feel free to not wait for me.

fabxc:
All comments full sentences (upper-case and final period). LGTM after comments.

notify/impl.go
brancz: it doesn't make a difference in the implementation, but yes that makes sense

notify/impl.go
```diff
@@ -546,26 +571,33 @@ func (n *Hipchat) Notify(ctx context.Context, as ...*types.Alert) error {
...
 	defer resp.Body.Close()
 
-	if resp.StatusCode/100 != 2 {
-		return fmt.Errorf("unexpected status code %v", resp.StatusCode)
+	return n.retry(resp.StatusCode)
+}
+
+func (n *Hipchat) retry(statusCode int) (bool, error) {
+	// Response code 429 for rate limiting and 5xx can potentially recover.
+	// 2xx responce codes are indicate successful requests.
```
brancz: are indicate ..

notify/impl.go
```diff
@@ -651,40 +683,45 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) error {
...
 
 	resp, err := ctxhttp.Post(ctx, http.DefaultClient, apiURL, contentTypeJSON, &buf)
 	if err != nil {
-		return err
+		return true, err
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode == 400 && alerts.Status() == model.AlertResolved {
+	// missing documentation therefore assuming 5xx is recoverable
```
brancz: full sentence & upcase

notify/impl.go
```diff
@@ -227,38 +235,38 @@ func (n *Email) auth(mechs string) (smtp.Auth, error) {
...
 
 	if n.conf.RequireTLS {
 		if ok, _ := c.Extension("STARTTLS"); !ok {
-			return fmt.Errorf("require_tls: true (default), but %q does not advertise the STARTTLS extension", n.conf.Smarthost)
+			return false, fmt.Errorf("require_tls: true (default), but %q does not advertise the STARTTLS extension", n.conf.Smarthost)
 		}
 		tlsConf := &tls.Config{ServerName: host}
 		if err := c.StartTLS(tlsConf); err != nil {
-			return fmt.Errorf("starttls failed: %s", err)
+			return false, fmt.Errorf("starttls failed: %s", err)
```
fabxc: Mh, this is executing a command. Isn't that generally a request/response cycle and hence can always error due to a flaky network. I think that's recoverable at least in some cases.

fabxc: Mh, so are the `c.Extension()` calls I guess.

brancz: True, then I'll make all request/response calls retry. I missed that and only thought of "is a TLS likely to be recovered on a retry?".

fabxc: Well, not super likely but might just have been a network blip. Exponential backoff is still a relatively safe mechanism and if we fail at this stage, we are at least not spamming any humans.

brancz: as far as I understand the extension calls actually don't do a request/response (see [here](https://en.wikipedia.org/wiki/Extended_SMTP#Extensions) as a list of supported extensions is sent after a successful EHLO

brancz: Nevermind, `Extension` [actually calls `hello`](https://golang.org/src/net/smtp/smtp.go?s=10039:10092#L347) which in turn makes it a request/response game again. I propose to make all these retry then. Wdyt?

fabxc: Yep, let's err on the side of being noisy until proven otherwise.

fabxc:
👍 🚢

PR #493, opened 2016-09-06
Update README

fabxc:
This removes some things that no longer apply and syncs the installation intructions with the Prometheus repository

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,28 +4,27 @@
 [![Docker Repository on Quay](https://quay.io/repository/prometheus/alertmanager/status)][quay]
 [![Docker Pulls](https://img.shields.io/docker/pulls/prom/alertmanager.svg?maxAge=604800)][hub]
 
-The Alertmanager handles alerts sent by client applications such as the Prometheus server. It takes care of deduplicating, grouping, and routing them to the correct receiver integration such as email, PagerDuty, or OpsGenie. It also takes care of silencing and inhibition of alerts.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,29 +4,28 @@
...
+of the GitHub repository. Using the latest production release binary
+is the recommended way of installing Alertmanager.
 
-    sudo apt-get install build-essential libc6-dev
+### Container images
 
-Red Hat family:
-
-    sudo yum install glibc-static
+Container images are available on [Quay.io](https://quay.io/repository/prometheus/alertmanager).
```
grobie: I think @sdurrheimer has a point, "container images" is not very specific. It'd be helpful to know that this is about docker images.

README.md
```diff
@@ -4,29 +4,28 @@
...
 
-### Build dependencies
+There are various ways of installing Alertmanager.
 
-These dependencies are necessary for building Alertmanager. There are no runtime dependencies, as the resulting binary is statically linked.
+### Precompiled binaries
 
-Debian family:
+Precompiled binaries for released versions are available in the
+[*releases* section](https://github.com/prometheus/alertmanager/releases)
```
fabxc: Ah, this was just copying the pattern from the Prometheus README. This is from before we had your release section. Will update on both sides.

README.md
brancz: I agree, if we're some day going to add appc/oci images then we can still rename this

fabxc:
Both addressed.

brancz:
lgtm

PR #495, opened 2016-09-08
silence: add cache for silence matchers

brancz:
compiling regex silence matchers on every query is expensive, therefore
caching them as soon as they are gossiped through the mesh

@fabxc 

a microbenchmark showed roughly a 80x improvement for precompiled regex vs. compiled on every use

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -30,6 +30,7 @@ import (
 	"github.com/golang/protobuf/ptypes/timestamp"
 	"github.com/matttproud/golang_protobuf_extensions/pbutil"
 	pb "github.com/prometheus/alertmanager/silence/silencepb"
+	"github.com/prometheus/alertmanager/types"
 	"github.com/prometheus/common/log"
 	"github.com/prometheus/common/model"
 	"github.com/satori/go.uuid"
@@ -43,11 +44,57 @@ func utcNow() time.Time {
 	return time.Now().UTC()
 }
 
+type matcherCache map[*pb.Silence]types.Matchers
+
+// Get retrieves the matchers for a given silence. If it is a missed cache
+// access, it compiles and adds the matchers of the requested silence to the
+// cache.
+func (c matcherCache) Get(s *pb.Silence) (types.Matchers, error) {
+	if m, ok := c[s]; ok {
+		return m, nil
+	}
+	return c.add(s)
+}
+
+// add compiles a silences' matchers and adds them to the cache.
+// It returns the compiled matchers.
+func (c matcherCache) add(s *pb.Silence) (types.Matchers, error) {
+	var (
+		ms types.Matchers
+		mt *types.Matcher
+	)
+
+	for _, m := range s.Matchers {
+		mt = &types.Matcher{
+			Name:  m.Name,
+			Value: m.Pattern,
+		}
+		switch m.Type {
+		case pb.Matcher_EQUAL:
+			mt.IsRegex = false
+		case pb.Matcher_REGEXP:
+			mt.IsRegex = true
+		}
+		err := mt.Init()
+		if err != nil {
+			return nil, err
+		}
+
+		ms = append(ms, mt)
+	}
+
+	c[s] = ms
+
+	return ms, nil
+}
+
 // Silences holds a silence state that can be modified, queried, and snapshot.
 type Silences struct {
 	logger    log.Logger
 	now       func() time.Time
 	retention time.Duration
+
+	mc matcherCache
 
 	gossip mesh.Gossip // gossip channel for sharing silences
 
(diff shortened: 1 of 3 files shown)


silence/silence.go
```diff
@@ -447,33 +501,14 @@ func QTimeRange(start, end time.Time) QueryParam {
 }
 
 // QMatches returns silences that match the given label set.
-func QMatches(set map[string]string) QueryParam {
+func QMatches(set model.LabelSet) QueryParam {
```
fabxc: I never was much of a fan of these custom types. That's why i deliberately left it out of this API. We would drag in a package dependency without ever using anything from that package.

silence/silence.go
```diff
@@ -425,7 +479,7 @@ type query struct {
 
 // silenceFilter is a function that returns true if a silence
 // should be dropped from a result set for a given time.
-type silenceFilter func(*pb.Silence, *timestamp.Timestamp) (bool, error)
+type silenceFilter func(*pb.Silence, matcherCache, *timestamp.Timestamp) (bool, error)
```
fabxc: For being generic, let's pass in the entire `Silences` object here. Who knows what other caches or information we want to extend this by.

silence/silence.go
```diff
@@ -276,6 +325,11 @@ func (s *Silences) setSilence(sil *pb.Silence) error {
 	if err != nil {
 		return err
 	}
+
+	_, err = s.mc.Add(sil)
+	if err != nil {
+		return err
+	}
```
fabxc: This is not sufficient. This code path is executed when a silence is set in this Alertmanager instance.
When we receive one set in a peer Alertmanager, it is directly inserted into the state.

To work with the mesh protocol we call `gossipData.Merge()` with the update we construct below. Inserting into our own state (`gossipData`) also happens in that method. When we receive updates from other peers `gossipData.mergeDelta()` is called (via `GossipBroadcast()` and `OnGossip()`) which also changes our state. For the latter this code here is not executed and hence we end up with a silence that's not in our cache.
This would require access to the cache in `gossipData.mergeDelta()`, which we currently don't have. Maybe it would be wise to make `type gossiper` implement the `mesh.Gossiper` and `mesh.GossipData` interface at once?

Secondly, we store snapshots of silence data that we can restore from after a binary restart. In `loadSnapshot()` we have to populate the cache again for the silences we load.

fabxc: Ah, I see you basically circumvent this by lazily adding the silence to the cache when it doesn't exist yet?

fabxc: I'd vote for entirely skipping the `Add` then as it is not happening in them same place of the general adding of silences, which is what threw me off initially. There's no good reason to pre-cache some silences and on-demand others.

Then we can just move `Add()` entirely into `Get`.

silence/silence.go
brancz: Yep, that makes sense.

silence/silence.go
brancz: You you're saying, we skip initializing the cache all together and let it warm up while using?

brancz: I feel like there are lots of edge cases which can cause the cache to miss, but "pre-heating" the cache on `setSilence` and `loadSnapshot` still seems a good idea, wdyt @fabxc ?

silence/silence.go
brancz: we either use the `model.LabelSet` here or change all the types in `types/match.go`, `dispatch/route.go`, `inhibit/inhibit.go` .. having `map[string]string` was much easier before as we duplicated all the matching functionality... which trade-off should we take?

silence/silence.go
fabxc: `setSilence()` still seems inconsistent. Either we do it for every set silence or none. In an HA setup with N peers, generally you have to expect about 1/N silences being created in the instance itself. The rest is fed via the mesh network.
So either set the cache on `Merge()` and `mergeDelta()` as described in my first comment or nowhere.

Doing it in `loadSnapshot()` sounds good so the first notifications are not unnecessarily delayed potentially. When new silences are set, that's relatively speaking very rare and not a performance concern.

silence/silence.go
fabxc: I see. We probably have to think about that in terms of the entire repo at a later point. Fine with keeping it as it is for now.

brancz: Agreed 👍

fabxc:
You have to either hold the RLock in `query()` until after the filtering is done or give the cache its own mutex.
The first option makes probably more sense until we know whether the filtering is actually taking very long in some cases.

brancz:
Yes I felt like there was a missing lock in there, but wasn't sure how to handle it.

fabxc:
👍

brancz:
Fixed that failing build problem @fabxc

PR #502, opened 2016-09-09
add HA documentation to README.md

brancz:
@fabxc

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,38 @@ receivers:
   - service_key: <team-DB-key>
 ```
 
+## High Availability
+
+> Warning: High Availablility is under active development
+
+To create a highly available cluster of the Alertmanager the instances need to
+be configured to communicate with each other. This is configured using the
+`-mesh.*` flags.
+
+- `-mesh.hardware-address` string: MAC address, i.e. mesh peer ID (default "&lt;hardware-mac-address&gt;")
+- `-mesh.listen-address` string: mesh listen address (default "0.0.0.0:6783")
+- `-mesh.nickname` string: peer nickname (default "&lt;machine-hostname&gt;")
+- `-mesh.peer` value: initial peers (may be repeated)
+
+The `mesh.hardware-address` flag is used as a unique ID among the peers. It
+defaults to the MAC address, therefore the default value should typically be a
+good option. The same applies to the default of the `mesh.nickname` flag, as it
+defaults to the hostname. The chosen port in the `mesh.listen-address` flag is
+the port that needs to be specified in the `mesh.peer` flag of the other peers.
+
+To start a cluster of three peers on your local machine use `goreman` and the
+Procfile within this repository.
+
+	goreman start
+
+To point your prometheus instance to multiple Alertmanagers use the
+`-alertmanager.url` parameter. It allows passing in a comma separated list.
+Start your prometheus like this, for example:
+
+	./prometheus -config.file=prometheus.yml -alertmanager.url http://localhost:9095,http://localhost:9094,http://localhost:9093
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -166,6 +166,32 @@ receivers:
...
+
+	goreman start
+
+To point your prometheus instance to multiple alertmanagers use the
+`-alertmanager.url` parameter. It allows passing in a list. Start your
+prometheus like this, for example:
+
+	./prometheus -config.file=prometheus.yml -alertmanager.url http://localhost:9095,http://localhost:9094,http://localhost:9093
+
+> make sure to have a valid `prometheus.yml` in your current directory
```
fabxc: Is this `>` on purpose?

README.md
```diff
@@ -166,6 +166,32 @@ receivers:
   - service_key: <team-DB-key>
 ```
 
+## High Availability
+
+> Warning: High Availablility is under active development
+
+To create a highly available cluster of the alertmanager the instances need to
```
fabxc: "Alertmanager" everywhere.

README.md
brancz: Yes, I usually use it as "as a side note"

README.md
```diff
@@ -166,6 +166,32 @@ receivers:
...
+be started in a mesh network. To configure the mesh use the following
+parameters:
```
fabxc: Mesh network might be a bit too abstract. Maybe describe more freely like "... the Alertmanager instances need to be configured to communicate with each other. This is configured via the `-mesh.*` flags.

Clarify that nickname and hardware address do not have to be explicitly set. Clarify that the port used in `mesh.listen-address` is the one to be used in `mesh.peer`. We should say that `mesh.peer` has to point to at least one living Alertmanager at the time of startup.

brancz: sounds good

brancz: Rewrote it a bit, PTAL @fabxc

README.md
```diff
@@ -166,6 +166,38 @@ receivers:
...
+be configured to communicate with each other. This is configured using the
+`-mesh.*` flags.
+
+- `-mesh.hardware-address` string: MAC address, i.e. mesh peer ID (default "&lt;hardware-mac-address&gt;")
+- `-mesh.listen-address` string: mesh listen address (default "0.0.0.0:6783")
+- `-mesh.nickname` string: peer nickname (default "&lt;machine-hostname&gt;")
+- `-mesh.peer` value: initial peers (may be repeated)
+
+The `mesh.hardware-address` flag is used as a unique ID among the peers. It
+defaults to the mac address, therefore the default value should typically be a
```
fabxc: MAC

README.md
```diff
@@ -166,6 +166,38 @@ receivers:
...
+good option. The same applies to the default of the `mesh.nickname` flag, as it
+defaults to the hostname. The chosen port in the `mesh.listen-address` flag is
+the port that needs to be specified in the `mesh.peer` flag of the other peers.
```
fabxc: Thinking again, that's probably not necessarily true for prod setups. Guess it's fine for testing and people will get the drift if they have their own port mapping for containers etc.

fabxc:
That should generally be good for an alpha. We'll see where people struggle based on incoming questions and can then extend documentation.

README.md
brancz: Yes thought about that as well... but I think it's clear enough for now. The general relationship that those are the connection is there and that's the important part I guess.

README.md
brancz: fixed

README.md
brancz: On a second thought, it should be easy to rewrite it to match both cases.

PR #508, opened 2016-09-19
[UI Improvement] Save creator's e-mail address to cookie

tattsun:
At silence UI,
* Alertmanager saves creator's email address to cookie.
* Alertmanager automatically fill it from cookie.

Thanks.


ui/app/index.html
```diff
@@ -3,6 +3,7 @@
 <head>
 	<script src="lib/jquery.min.js"></script>
 	<script src="lib/moment.min.js"></script>
+  <script src="lib/js.cookie.js"></script>
```
fabxc: indentation is off

fabxc:
That seems good usability-wise.

tattsun:
@fabxc Thanks for reviewing! Please check new commit I pushed now.

brancz:
LGTM, wdyt @fabxc ?

grobie:
What's the reason to store the email in a cookie instead of the browser local storage?

tattsun:
@grobie, ah, it's simpler and better to use local storage. I fixed it to use local storage.

fabxc:
Yea, was thinking the same. Then thought that cookies are proven and reliable. But I suppose local storage is more ubiquitous and well working than when I worked with it last time about 5 years ago :)

fabxc:
Looks good, thank you! 👍

brancz:
I guess both would have worked, but yes local storage is the nicer solution.

PR #509, opened 2016-09-19
silence: add metrics

fabxc:
This adds metrics for querying as it is done internally and critical. The create/modification actions are basically covered by the API handler instrumentation and would just bloat here.

@discordianfish

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -31,6 +31,7 @@ import (
 	"github.com/matttproud/golang_protobuf_extensions/pbutil"
 	pb "github.com/prometheus/alertmanager/silence/silencepb"
 	"github.com/prometheus/alertmanager/types"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/log"
 	"github.com/prometheus/common/model"
 	"github.com/satori/go.uuid"
@@ -91,6 +92,7 @@ func (c matcherCache) add(s *pb.Silence) (types.Matchers, error) {
 // Silences holds a silence state that can be modified, queried, and snapshot.
 type Silences struct {
 	logger    log.Logger
+	metrics   *metrics
 	now       func() time.Time
 	retention time.Duration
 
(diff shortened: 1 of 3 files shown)


silence/silence.go
```diff
@@ -106,6 +108,50 @@ type Silences struct {
...
+
+func newMetrics(r prometheus.Registerer) *metrics {
+	m := &metrics{}
+
+	m.gcDuration = prometheus.NewSummary(prometheus.SummaryOpts{
+		Name: "alertmanager_silences_gc_duration_seconds",
+		Help: "Duration of the last garbage collection cycle.",
+	})
+	m.snapshotDuration = prometheus.NewSummary(prometheus.SummaryOpts{
+		Name: "alertmanager_silences_snapshot_duratio_seconds",
```
grobie: Typo

silence/silence.go
```diff
@@ -552,14 +605,22 @@ func QState(states ...SilenceState) QueryParam {
...
+	start := time.Now()
+	defer func() {
+		s.metrics.queryDuration.Observe(float64(time.Since(start)) / float64(time.Second))
+	}()
+	s.metrics.queriesTotal.Inc()
+
 	q := &query{}
 	for _, p := range params {
 		if err := p(q); err != nil {
+			s.metrics.queryErrorsTotal.Inc()
```
grobie: Might be better to move that into the defer func so that it can't be forgotten to increase the counter when returning a new error.

silence/silence.go
```diff
@@ -552,14 +605,22 @@ func QState(states ...SilenceState) QueryParam {
 
 // Query for silences based on the given query parameters.
 func (s *Silences) Query(params ...QueryParam) ([]*pb.Silence, error) {
+	start := time.Now()
+	defer func() {
+		s.metrics.queryDuration.Observe(float64(time.Since(start)) / float64(time.Second))
```
grobie: `time.Since(start).Seconds()` is a lot more readable.

silence/silence.go
fabxc: That's always pretty fragile when the return error is overwritten in a scope etc.
Went a different route that should solve it.

silence/silence.go
```diff
@@ -106,6 +108,50 @@ type Silences struct {
...
+	queryErrorsTotal prometheus.Counter
+	queryDuration    prometheus.Histogram
+}
+
+func newMetrics(r prometheus.Registerer) *metrics {
+	m := &metrics{}
+
+	m.gcDuration = prometheus.NewSummary(prometheus.SummaryOpts{
+		Name: "alertmanager_silences_gc_duration_seconds",
+		Help: "Duration of the last garbage collection cycle.",
```
juliusv: "last silence garbage collection cycle" for people to not confuse the HELP string with general Go garbage collection?

silence/silence.go
```diff
@@ -106,6 +108,50 @@ type Silences struct {
...
+	})
+	m.snapshotDuration = prometheus.NewSummary(prometheus.SummaryOpts{
+		Name: "alertmanager_silences_snapshot_duration_seconds",
+		Help: "Duration of the last snapshot.",
```
juliusv: "last silences snapshot"?

silence/silence.go
```diff
@@ -225,6 +273,9 @@ Loop:
 // GC runs a garbage collection that removes silences that have ended longer
 // than the configured retention time ago.
 func (s *Silences) GC() (int, error) {
+	start := time.Now()
+	defer func() { s.metrics.gcDuration.Observe(time.Since(start).Seconds()) }()
```
brancz: why defer here and not in `Query`?

brancz: Maybe I'm answering myself here, my guess is that defer makes it clearer but the performance is more important in the Query method. (I saw that you did the same defer statement in the other maintenance methods)

fabxc:
Everything addressed.

PR #513, opened 2016-09-25
Fixed the go install command

farcaller:
Removed the vendor experiment flag (it's obsolete now) and added the proper path to the main binary.


fabxc:
Thanks! Can we keep the `GOVENDOREXPERIMENT=1` around for a bit longer though?
It doesn't hurt if it is not necessary but users with go1.5.x that just want to try things out, won't run into problems.

farcaller:
Done.

PR #514, opened 2016-09-25
Added IRC receiver support

farcaller:
This adds a simple irc client support, so that AM can nag people over IRC too.


brian-brazil:
Thanks for taking the time to put this together. It doesn't make sense to have this in the AM, as unlike all other receivers this requires state in the form of a TCP connection to be maintained in one alertmanager. That won't work for a HA setup.

Instead a (single) separate binary that takes in notifications via the webhook would be best.

farcaller:
Noted.

PR #515, opened 2016-09-26
Adding TimeNow and FormattedTimeNow functions for templates

ipstatic:
It is useful for us to include the timestamp of when an alert is fired.
This change adds a TimeNow and FormattedTimeNow function that is
available inside a template. The Formatted function takes a string using
the values defined here: https://golang.org/pkg/time/#Time.Format

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -210,6 +210,11 @@ func (kv KV) Values() []string {
 	return kv.SortedPairs().Values()
 }
 
+// FormattedTimeNow returns a formatted time string of TimeNow
+func (data *Data) FormattedTimeNow(str string) string {
+  return data.TimeNow.Format(str)
+}
+
 // Data is the data passed to notification templates and webhook pushes.
 //
 // End-users should not be exposed to Go's type system, as this will confuse them and prevent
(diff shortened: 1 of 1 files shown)


juliusv:
@ipstatic That's the notification time, not the time when the alert started firing. The latter you can already get from the alert itself (`StartsAt` member of `Alert`). Is that what you want instead?

ipstatic:
We actually wanted the notification time as sometimes email can be delayed and it is helpful to know when the alert was actually triggered.

juliusv:
I see... wouldn't the mails still be correctly timestamped with the origination date by the sender though? https://tools.ietf.org/html/rfc2822#section-3.6.1. We add the notification time as the `Date:` header here: https://github.com/prometheus/alertmanager/blob/master/notify/impl.go#L320

ipstatic:
Not all clients reference the Date header. Gmail does not use the Date header field in its list and Apple's Mail.app uses the date stamp of the file on the server and not the header.

ipstatic:
After more thought, we can just use StartsAt. Thanks for the help!

PR #518, opened 2016-09-29
Nflog metrics

fabxc:
@brancz


nflog/nflog.go
```diff
@@ -113,6 +115,50 @@ type nlog struct {
...
+		Name: "alertmanager_nflog_gc_duration_seconds",
+		Help: "Duration of the last notification log garbage collection cycle.",
+	})
+	m.snapshotDuration = prometheus.NewSummary(prometheus.SummaryOpts{
+		Name: "alertmanager_nflog_snapshot_duration_seconds",
+		Help: "Duration of the last notification log snapshot.",
+	})
+	m.queriesTotal = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "alertmanager_nflog_queries_total",
+		Help: "How many notification log queries were received.",
```
brancz: Maybe "Number of notification log queries received." ?

nflog/nflog.go
```diff
@@ -113,6 +115,50 @@ type nlog struct {
...
+	})
+	m.queryErrorsTotal = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "alertmanager_nflog_query_errors_total",
+		Help: "How many notification log received queries did not succeed.",
```
brancz: Same as above

nflog/nflog.go
```diff
@@ -113,6 +115,50 @@ type nlog struct {
...
+	})
+	m.queryDuration = prometheus.NewHistogram(prometheus.HistogramOpts{
+		Name: "alertmanager_nflog_query_duration_seconds",
+		Help: "Duration of notification log query evaluation.",
```
brancz: Maybe it makes sense to mention that this is in the scope of the last evaluation, to not mistake it for a total. Maybe "Duration of *last* notification log query evaluation."

nflog/nflog.go
```diff
@@ -351,27 +408,37 @@ func (l *nlog) GC() (int, error) {
...
-	if q.recv == nil || q.groupKey == nil {
-		// TODO(fabxc): allow more complex queries in the future.
-		// How to enable pagination?
-		return nil, errors.New("no query parameters specified")
-	}
 
-	l.mtx.RLock()
-	defer l.mtx.RUnlock()
+		l.mtx.RLock()
+		defer l.mtx.RUnlock()
```
brancz: Typically I wouldn't try to avoid defer everywhere but in this case, since this is a method that is potentially called rather often it probably makes sense not to use it here. Plus there doesn't seem to be a possible error state within the 3 lines below, which would cause the mutex to stay locked. Please correct me if I'm wrong.

nflog/nflog.go
```diff
@@ -331,6 +385,9 @@ func (l *nlog) log(r *pb.Receiver, gkey, ghash []byte, resolved bool) error {
 
 // GC implements the Log interface.
 func (l *nlog) GC() (int, error) {
+	start := time.Now()
+	defer func() { l.metrics.gcDuration.Observe(time.Since(start).Seconds()) }()
```
brancz: Tests are failing here because the `Nflog` is not initialized `WithMetrics`.

brancz:
I don't think the "changes" in `template/internal/deftmpl/bindata.go` and `ui/bindata.go` actually changed anything, I'm just wondering how they got there.

Besides that tests are failing because the metrics are expected to be initialized, however, they only are if the `Nflog` is initialized `WithMetrics`.

nflog/nflog.go
fabxc: There are 3 returns points below where we'd have to repeat that. A defer has an overhead of something at the order of 20-60ns with a 2x improvement coming with go1.8.

Say we send out 10 notifications per second, which would even overload an organization with 1000 people on-call this will give us <0.0000005 of that one second being used for the defer. I'd argue this is fine for a load estimate about 100x above normal usage.

brancz: True. Let's improve if it actually becomes a problem.

nflog/nflog.go
fabxc: This is a histogram so it's not just about the last ones.

fabxc:
I fell back to always using the default registry as the client library currently doesn't support a noop-registry that would be easy to use.

brancz:
lgtm 👍

PR #520, opened 2016-09-30
Persist user options in browser local storage

cofyc:
It's convenient if user's choice is persistent.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -2,7 +2,7 @@ machine:
   environment:
     DOCKER_IMAGE_NAME: prom/alertmanager
     QUAY_IMAGE_NAME: quay.io/prometheus/alertmanager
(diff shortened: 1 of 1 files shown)


ui/app/index.html
```diff
@@ -10,6 +10,7 @@
 	<script src="lib/angular-resource.min.js"></script>
 	<script src="lib/angular-moment.min.js"></script>
 	<script src="lib/ui-bootstrap-custom-tpls-2.0.1.min.js"></script>
+	<script src="lib/ngStorage.min.js"></script>
```
fabxc: We are using local storage to store the most recent comment email address as will. There we didn't include another dependency. Are you sure we need this for the code below?

cofyc:
Hi, I removed ngStorage.min.js, and create a new pull request: #521

PR #525, opened 2016-10-05
notify: move resolved alert filtering to integration

### LINKED ISSUE #523: Notification log problems in master
Hi!

I tried to use AlertManager from the current HEAD of the master branch (a84fba776244fd6be464ffb7b6d1ce3e8c06ae9d), but there seems to be a big problem when SetNotifiesStage and FilterResolvedStage are used together.  Consider the following sequence of events:

  1. An alert fires, and an email is sent
  2. The same alert is resolved, but FilterResolvedStage prevents that fact from being propagated downwards
  3. The alert fires again

Now the fact that the alert is firing is propagated down the pipeline, but because SetNotifiesStage never saw the "resolved" event, DedupStage.needsUpdate thinks nothing's changed and holds the alert.  The problem continues until `repeat_interval`, and then the sequence starts all over again.

I'm not convinced that this is the correct fix, but the problem goes away if I apply the following change:

```Go
diff --git a/notify/notify.go b/notify/notify.go
index 264e494..31af6d8 100644
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -206,10 +206,10 @@ func createStage(rc *config.Receiver, tmpl *template.Template, wait func() time.
        }
        var s MultiStage
        s = append(s, NewWaitStage(wait))
-       s = append(s, NewFilterResolvedStage(i.conf))
        s = append(s, NewDedupStage(notificationLog, recv))
-       s = append(s, NewRetryStage(i))
        s = append(s, NewSetNotifiesStage(notificationLog, recv))
+       s = append(s, NewFilterResolvedStage(i.conf))
+       s = append(s, NewRetryStage(i))
 
        fs = append(fs, s)
    }
```

brancz:
notify: move resolved alert filtering to integration

Resolved alerts, even when filtered, have to end up in the
SetNotifiesStage, otherwise when an alert fires again it is ambiguous
whether it was resolved in between or not.
    
fixes #523

@fabxc @johto

### DIFF
--- a/notify/notify_test.go
+++ b/notify/notify_test.go
@@ -33,6 +33,18 @@ import (
 	"github.com/prometheus/alertmanager/types"
 )
 
+type notifierConfigFunc func() bool
+
+func (f notifierConfigFunc) SendResolved() bool {
+	return f()
+}
+
+type notifierFunc func(ctx context.Context, alerts ...*types.Alert) (bool, error)
+
+func (f notifierFunc) Notify(ctx context.Context, alerts ...*types.Alert) (bool, error) {
+	return f(ctx, alerts...)
+}
+
 type failStage struct{}
 
 func (s failStage) Exec(ctx context.Context, as ...*types.Alert) (context.Context, []*types.Alert, error) {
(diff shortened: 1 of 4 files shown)


notify/notify_test.go
```diff
@@ -283,6 +283,49 @@ func TestRoutingStage(t *testing.T) {
...
+}
+
+type RecordNotifier struct {
+	alerts []*types.Alert
+}
+
+func (n *RecordNotifier) Notify(ctx context.Context, alerts ...*types.Alert) (bool, error) {
+	for _, a := range alerts {
+		n.alerts = append(n.alerts, a)
+	}
```
fabxc: In general, this simplifies to `n.alerts = append(n.alerts, alerts...)`

notify/notify_test.go
```diff
@@ -283,6 +283,49 @@ func TestRoutingStage(t *testing.T) {
...
 
+type notifierConfigFunc func() bool
+
+func (f notifierConfigFunc) SendResolved() bool {
+	return f()
+}
+
+type RecordNotifier struct {
+	alerts []*types.Alert
+}
```
fabxc: I've found over time that having a single test/mock implementation of an interface is easier here. Usually, at some point there'll be further tests requiring a mock as well wis slightly different behavior.
It's then easier to have:

```go
type mockNotifier struct {
    notify func(context.Context, alerts ...*types.Alert) (bool, error)
}
```

And in your actual test you instantiate:

```go
var recorded []*types.Alert
n := &mockNotifier{
    notify: func(_ context.Context, alerts ...*types.Alert) (bool, error) {
        recorded = append(recorded, alerts...)
    },
}
// test code ...
```

It's more verbose but it more flexible and makes the entire behavior easily visible within the test function.

fabxc:
We should add a test case to the acceptance tests as well.

notify/notify_test.go
brancz: sgtm

brancz:
@fabxc addressed comments and added an integration test including a description of what it does.

PR #530, opened 2016-10-07
encrypt mesh communication

### LINKED ISSUE #529: Mesh config uses fixed password ""
`initMesh` supplies the empty string (and not `nil`) as the password when configuring the weave mesh in v0.5.0-beta.0. This should be configurable at runtime, preferably by reading the password from a filename-valued flag.

The problem with using an empty string is that it triggers the mesh code to negotiate an encrypted session on the network connections, giving the illusion that confidentiality is protected. Poor security is worse than no security. The newness of the mesh protocol already carries an administrative burden because IDS and other auditing mechanisms have to address it; let's not put deployers in the position of arguing for obscurity over security.

I suggest putting the password in a file and not in the config file because it will make password rotation easier. Rotating a password in a mesh of alertmanagers is a matter of restarting alertmanagers so that they are configured to join a new mesh with a new password, disjoint from the old mesh. The mesh peer flags will have to change for this to work, so it is no extra hardship to change the password flag at the same time. If the password is put in the config file, then something extra will have to be done to prevent the changed config file from being propagated to the old-mesh alertmanagers in the usual way.

Finally, consider recognizing a special flag value which will use `nil` as the `mesh.Config.Password` value. This will allow disabling the secure session setup entirely, which can be useful for a few reasons (namely, debugging, TLS-wrapping, and enabling auditing of traffic on sensitive network segments).

brancz:
Currently all mesh communication is unencrypted. Adding this parameter allows the operator to make use of this functionality of mesh. The operator has to take care of password rotations.

@fabxc @aecolley

fixes #529

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -84,6 +84,7 @@ func main() {
 		meshListen = flag.String("mesh.listen-address", net.JoinHostPort("0.0.0.0", strconv.Itoa(mesh.Port)), "mesh listen address")
 		hwaddr     = flag.String("mesh.hardware-address", mustHardwareAddr(), "MAC address, i.e. mesh peer ID")
 		nickname   = flag.String("mesh.nickname", mustHostname(), "peer nickname")
+		password   = flag.String("mesh.password", "", "password to join the peer network (empty password disables encryption)")
 	)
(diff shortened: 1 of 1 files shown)


fabxc:
👍

PR #531, opened 2016-10-08
inhibitrule: defer Unlock to fix race

chlunde:
There's a typo in InhibitRule.set, the defer keyword is missing.  Found with a -race build.


brian-brazil:
Thanks!

PR #532, opened 2016-10-13
Fix routing tree 0.4

stuartnelson3:
The routing tree regexp support was broken in https://github.com/prometheus/alertmanager/commit/7095c699521d8ede845206fa287c1ba3dd743f28, the `MatchRE` needs to be stringified before it gets marshaled, otherwise it gets marshaled as an empty object.

JSON response before:

```json
match_re: { service: {}}
```

JSON response after:
```json
match_re:{ service: "/some|regexp/" }
```

Pictures of the before and after:

Rendered tooltip before:
broken_routing_tree

Rendered tooltip after:
fixed_routing_tree


beorn7:
Wouldn't it be easier to add a ` MarshalJSON() ` method to the `Regexp` type similar to the one for YAML in https://github.com/stuartnelson3/alertmanager/blob/3d5d8c325f06071293932add27a2ffc54ca0e999/config/config.go#L483 ?

stuartnelson3:
Ah, way better. Didn't even think of that, didn't realize that `Regexp` was embedding `regexp.Regexp`.

api.go
```diff
@@ -57,7 +57,7 @@ type API struct {
 	alerts         provider.Alerts
 	silences       provider.Silences
 	config         string
-	configJSON     config.Config
+	configJSON     *config.Config
```
beorn7: Is this change (still/really) necessary?

stuartnelson3: The motivation was consistency with the definitions for marshaling/unmarshaling YAML. Doesn't bother me to change the marshal json to have a value receiver type instead of a pointer, and change this back.

config/config.go
```diff
@@ -455,3 +455,8 @@ func (re *Regexp) MarshalYAML() (interface{}, error) {
 	}
 	return nil, nil
 }
+
+// MarshalJSON implements the json.Marshaler interface.
+func (re *Regexp) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf(`"%s"`, re.String())), nil
```
beorn7: That won't handle the escaping correctly.

Use
```go
return json.Marshal(re.String())
```

stuartnelson3: way better than what I had :p

beorn7:
:+1:

fabxc:
👍

PR #537, opened 2016-10-22
Correctly render continue routing

stuartnelson3:
Nodes with `continue: true` were being ignored,
and only the first match in the routing tree was
being highlighted.

Additionally, each node was incorrectly being set
to continue=true if nothing was received from the
API. The API sends nothing if continue=false (the default).

This change should also be applied to the code in `master`.

See comments for redacted pictures.


stuartnelson3:
It's difficult to show, so perhaps it would be best to show this to someone in person like @beorn7, but I confirmed that the multi-routing with the `continue` option on a single "catch-all" route works successfully, and the other routes cease the tree search when they don't have `continue: true` in their config.

image

beorn7:
:+1:

beorn7:
Merge into master is not really trivial. I guess we have to cherrypick the change here.

stuartnelson3:
Yes, there's no way to merge into master.

PR #543, opened 2016-11-01
Cut 0.5.0

fabxc:
@brancz @beorn7 @brian-brazil @juliusv 

I think its time :)


brancz:
Do we have documentation prepared for this?

fabxc:
I think the basics are in the README – from the outside, nothing really changes aside from the storage wipe and the new flags to connect peers.

Given we didn't have general docs on AM operation aside from configuration so far, I don't see it as a release blocker.

I'll make sure the release email contains all necessary information.

brancz:
Alright sgtm. 👍

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
...
+
+This release requires a storage wipe. It contains fundamental internal
+changes that came with implementing the high availability mode.
+
+* [FEATURE] Alertmanager clustering for high availability
+* [FEATURE] Garbage collection of old silences and notification logs
+* [CHANGE] New storage format
+* [CHANGE] Stricter silence semantics for consistent historical view
+
 ## v0.5.0-beta.0 / 2016-10-06
```
beorn7: Shouldn't we merge the beta changelog into the changelog of the final version? In my understanding, we don't keep rc or beta changes separately, except for the current pre-release.

brancz: Generally I would agree but both of these pre releases were out there for a month each. So I would want a more detailed changelog in this case as a user. But also wouldn't merging them.

beorn7: We had pre releases out for quite some time for the Prometheus server and still had one consolidated block of changes for the final release.

You could point out in the announcement mail what changed compared to the last beta version.

beorn7: Not feeling strongly, just trying to document the practice so far. If you like to change it, go ahead. It should just be an informed decision.

brancz: Ok then let's stick with how it was done so far.

fabxc: Yea, I realized I was doing it differently and wondered whether the merging stuff was right to begin with.
But yea, I think it's fine as the GH releases still show the alpha and beta ones. I'll merge again to stay consistent for now.

beorn7:
:+1:

ledgr:
@fabxc, is it possible to convert data to new storage format? Is there a documented procedure?

brancz:
@ledgr part of a storage wipe are you concerned about? There is an (undocumented) rest api, where you could extract silences for example and then add them to the new instance using the rest api of the new version. This will require reading some code though as it is currently undocumented and doesn't have any stability guarantees.

fabxc:
Yes, silences are probably the interesting bit.
I recall @grobie having a script there that could help.

beorn7:
@grobie 's script read the silences directly from the (cold) sqlite data files and then uploaded them via the REST API. The former has to be replaced by retrieving the silences via the API. The latter needs to be updated to not try to set the original start date because that's something the new AM will prevent. (Start in the past, rewriting history, and all that jazz.)

PR #557, opened 2016-11-20
Enable display of dates in a named timezone (defaults to UTC).

aecolley:
This change imports the [moment-timezone sibling](http://momentjs.com/timezone/) of the moment javascript library, and uses it to convert times to a fixed timezone for display. Of course, the default is still UTC; but an edit to `uiTimezone` can customize it to, for instance, `America/Los_Angeles` for installations where there's great demand for that kind of thing.

The uptime field is now displayed as a "time ago" string, but the hovertext displays a full RFC-3339 time (in browser-local time, unfortunately). The datetime pickers used in the silence-creation form are also incorrigibly in browser local time, but at least they now indicate what timezone they're in.

The "calendar" displays of start and end times for alerts and silences now include a timezone abbreviation such as `PST`, achieved by adding the format `z` to the locale strings used by the `amCalendar` filter. On expanded rows, the full displayed times now consistently include time offsets.


fabxc:
@TheTincho how does this play out for you dependency wise for Debian releases?

TheTincho:
@fabxc, it is a bit tricky.

Moment-timezone is not yet packaged for Debian. There is a request for it since last year [#802262](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=802262), but nobody picked it up.

I don't really understand Grunt, but it does not seem to be too bad for packaging. I am more worried about the constant updates that the Olson database needs. Plus, I am not sure I can get it in time for the release (I will need to upload it before Christmas, and I am in the middle of a continent move :-))

Worst case, I will have to patch AM to revert this commit, or provide some degraded functionality.

Thanks for asking!

PR #563, opened 2016-11-22
Export parsed configuration as JSON and drop js-yaml dependency.

### LINKED ISSUE #556: Please re-merge #423
Hi, 

Back in July I prepared #423, to drop a dependency on a troublesome JS library. But I see that it got lost after 0.3.0. Could you re-merge it?

fabxc:
* Export parsed configuration as JSON in /api/v1/status.
* Avoid exporting the XXX fields.
* Drop js-yaml library and use already parsed configuration.
* Go fmt + go-bindata

Fixes #556 

@TheTincho

Sorry about dropping it before. Will cut a patch release with this one and some other fixes soon.


brancz:
lgtm 👍

PR #570, opened 2016-11-25
travis: update Go compiler version

fabxc:


brancz:
not too familiar with `promu` .. are there any other places where we need to change to this version to be consistent in building/testing/releasing/etc.

fabxc:
promu has a default which is up to date:
https://github.com/prometheus/promu/pull/51/files
It's just our Travis files that are not using promu. I updated Alertmanager
as well.

brancz:
👍

Issue #574, opened 2016-12-05
Add groupKey to /alerts/groups endpoint


soupdiver:
I currently work on an integration with the alertmanager.
I receive webhooks from alertmanager which includes the `groupKey`.

I wonder why the API endpoint `api/v1/alerts/groups` does not contain the same `groupKey` information so that I could match the earlier received key against a list of grouped alerts to silences them.

Am I missing something or does it make sense to add that information to the mentioned API endpoint?

mg03:
@soupdiver : Hi I was looking for documentation on doing something similar. Do you have a link to the docs ?

soupdiver:
@mg03 Not sure which docs you are looking for... I went through the source code and looked at the API responses.

brancz:
There are no API docs yet, as the API is not considered stable, and may change at any next release until it is explicitly marked as stable.

soupdiver:
@brancz yep, makes sense. Can you may comment on my suggestion about adding the `groupKey` to the mentioned API endpoint?

brancz:
I'm unsure why it's not in there seems fine to add it, maybe @fabxc knows more or has a stronger opinion.

fabxc:
If you are okay with the explicit instability, I don't see any reasons against accepting a PR adding that field.

soupdiver:
I don't mind the instability. It's communicated :)

brancz:
This can be closed right? @soupdiver @fabxc

soupdiver:
Solved by #576

PR #576, opened 2016-12-06
add groupKey to alerts/groups API endpoint

soupdiver:
Discussion started at #574 

Add the `groupKey` information to the `alerts/groups` endpoint.
When having configured a webhook receiver it contains the `groupKey`. Adding the `groupKey` to the overview of alertgroups makes it possible to match the incoming webhook with groups reported back from the API.

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -86,8 +86,9 @@ type APIAlert struct {
 
 // AlertGroup is a list of alert blocks grouped by the same label set.
 type AlertGroup struct {
-	Labels model.LabelSet `json:"labels"`
-	Blocks []*AlertBlock  `json:"blocks"`
+	Labels   model.LabelSet `json:"labels"`
+	GroupKey uint64         `json:"groupKey"`
+	Blocks   []*AlertBlock  `json:"blocks"`
 }
 
 // AlertOverview is a representation of all active alerts in the system.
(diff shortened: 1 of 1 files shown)


dispatch/dispatch.go
```diff
@@ -111,6 +112,7 @@ func (d *Dispatcher) Groups() AlertOverview {
 			alertGroup, ok := seen[ag.fingerprint()]
 			if !ok {
 				alertGroup = &AlertGroup{Labels: ag.labels}
+				alertGroup.GroupKey = uint64(ag.labels.Fingerprint())
```
fabxc: It's actually generated differently: https://github.com/prometheus/alertmanager/blob/master/dispatch/dispatch.go#L332.

Probably best to at a `GroupKey() uint64` method to the aggregation group to avoid getting out of sync.

soupdiver: Oh my bad 🙈

soupdiver:
Updated the PR as proposed in the comment

dispatch/dispatch.go
```diff
@@ -362,6 +364,10 @@ func (ag *aggrGroup) fingerprint() model.Fingerprint {
 	return ag.labels.Fingerprint()
 }
 
+func (ag *aggrGroup) GroupKey() uint64 {
+	return uint64(ag.labels.Fingerprint() ^ ag.routeFP)
```
brancz: Maybe I'm missing something, but we can just keep the `Fingerprint` type here all the way no? Then we don't have to do any casting along the way.

soupdiver: I was thinking the same about the Fingerprint, but sticked with the `grouopKey` as it was already in the webhook. This is up to you guys :)

brancz: Not sticking to the "Fingerprint" type is fine by me as well as long as we stick with the same type. But in the context of this change I guess the way it is is fine then. Nevermind me here :)

fabxc:
LGTM 👍

Issue #354, opened 2016-05-17
Feature request: "script receiver"


ahajnal:
Dear Developers,
I don't know if it is possible to implement but I would be interested in having a sort of "script" receiver, which would execute a user-defined (bash) script on alert events, like:

receivers:
- name: script
  script_configs:
  - script: /home/user/myalert.sh

If possible, working in a way that alert.sh either get alert json string as argument, or it would be even greater that the script could be templated (with like some {{label=''}} stuff).

I ask it because with using webhook I have to deploy another component to receive alert POSTs
(nor I can configure webhook to send GET/PUT/... with the desired content) and do something special I need.

In this way alertmanager would be more flexible. As I remember Zabbix can also run scripts on triggers.

What do you think?

Best regards,
Akos Hajnal

fabxc:
Hi Akos,

I get that the webhook is another hop, which is an inconvenience to setup.
Templating the script itself is not possible from a security perspective.

I'm open to discuss this and would like to hear more opinions and whether
more people feel a need for something similar.

Fabian

brian-brazil:
I'm against scripts, too much of a reliability threat (how well do you deal with 100 forks+exec of a script per second?) and there's the security risks too. There's nothing stopping development of a webhook utility that runs scripts.

ahajnal:
Hi Fabian,
To add some further details... In my special case, I would like not only to monitor something and get alerts from Prometheus-Alertmanager but to react actively on events - using as few components as possible. Script is something that can be quite active. What about a REST receiver, or a more generic receiver in which users can configure HTTP method/headers/payload? This avoids forks+exec and is similar to webhooks.
Akos

discordianfish:
> There's nothing stopping development of a webhook utility that runs scripts.

I've written exactly this. It takes a script as arg and listens on http for prometheus webhook. Will try to open source it and ping you.

ahajnal:
Hi discordianfish, 
That will be a definitely useful tool not just for alertmanager.
Akos

discordianfish:
@ahajnal Just opened it up: https://github.com/imgix/prometheus-am-executor

ahajnal:
Looks great!

fabxc:
Closing as a we now even have a webhook implementation solving this.

urashidmalik:
Not sure if you guys looked at monit.
https://mmonit.com/wiki/Monit/ConfigurationExamples

Which provides option for exec, pretty helpful.

PR #587, opened 2016-12-21
add runcommand receiver

rchicoli:
this implements the receiver `runcommand`, which enables the enduser to extend the functionality of a receiver with any script.
This might be useful for lots of people, which the notifiers are not added yet.

Configuration used for testing:

```yaml
global:

route:
  group_by: ['alertname']
  group_wait: 10s
  group_interval: 15s
  repeat_interval: 30s
  receiver: run-custom-scripts

receivers:
- name: 'run-custom-scripts'
  runcommand_configs:
  - script: 'sh /tmp/make-a-call.sh 911'
  - script: '/tmp/notify-zabbix.sh'

```

In addition is set some environment variable to the script:

```bash
PROMETHEUS_ANNOTATIONS_DESCRIPTION=This is a description of process fds
PROMETHEUS_ANNOTATIONS_SUMMARY=Process File Descriptor
PROMETHEUS_COMMONLABEL_ALERTNAME=ProcessFds Test
PROMETHEUS_COMMONLABEL_INSTANCE=172.17.0.3:9090
PROMETHEUS_COMMONLABEL_JOB=prometheus
PROMETHEUS_COMMONLABEL_MONITOR=codelab-monitor
PROMETHEUS_COMMONLABEL_SEVERITY=low
PROMETHEUS_EXTERNALURL=http://a298bedb1f71:9093
PROMETHEUS_GROUPLABEL_ALERTNAME=ProcessFds
PROMETHEUS_RECEIVER=runcommand
PROMETHEUS_STATUS=firing
```

hey @fabxc if you guys approve this, I am glad to provide the docs for it.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -430,14 +430,15 @@ type Receiver struct {
 	// A unique identifier for this receiver.
 	Name string `yaml:"name" json:"name"`
 
-	EmailConfigs     []*EmailConfig     `yaml:"email_configs,omitempty" json:"email_configs,omitempty"`
-	PagerdutyConfigs []*PagerdutyConfig `yaml:"pagerduty_configs,omitempty" json:"pagerduty_configs,omitempty"`
-	HipchatConfigs   []*HipchatConfig   `yaml:"hipchat_configs,omitempty" json:"hipchat_configs,omitempty"`
-	SlackConfigs     []*SlackConfig     `yaml:"slack_configs,omitempty" json:"slack_configs,omitempty"`
-	WebhookConfigs   []*WebhookConfig   `yaml:"webhook_configs,omitempty" json:"webhook_configs,omitempty"`
-	OpsGenieConfigs  []*OpsGenieConfig  `yaml:"opsgenie_configs,omitempty" json:"opsgenie_configs,omitempty"`
-	PushoverConfigs  []*PushoverConfig  `yaml:"pushover_configs,omitempty" json:"pushover_configs,omitempty"`
-	VictorOpsConfigs []*VictorOpsConfig `yaml:"victorops_configs,omitempty" json:"victorops_configs,omitempty"`
+	EmailConfigs      []*EmailConfig      `yaml:"email_configs,omitempty" json:"email_configs,omitempty"`
+	PagerdutyConfigs  []*PagerdutyConfig  `yaml:"pagerduty_configs,omitempty" json:"pagerduty_configs,omitempty"`
+	HipchatConfigs    []*HipchatConfig    `yaml:"hipchat_configs,omitempty" json:"hipchat_configs,omitempty"`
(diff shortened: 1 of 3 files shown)


juliusv:
Hi and thanks! Unfortunately it is (at least has been, haven't been following too closely) a non-goal to have the AM directly run scripts. The webhook is the way to go for building arbitrary integrations, see discussio in https://github.com/prometheus/alertmanager/issues/354

rchicoli:
I would like to ask for you guys to rethink about this decision.
The reason is that AM will never provide every single hooks that suits to everyone, some processes go beyond sending emails or notifications.

Personally I don't see any down sides, instead we could focus on making it reliable, for example by implementing timeouts, etc.
So the end-user does have the opportunity to choose what wants to use, even though he is aware of the disadvantages of running locally scripts.

I kinda like the idea of `prometheus-am-executor`, but then is another dependency I need to take care of if it's running.

On the other hand comparing AM with other tools, most out there provide a way to run scripts, for example `zabbix actions`, etc.
So we should not be scare of this feature, but ensure it does work and notify the users what could go wrong.

juliusv:
@rchicoli I can understand your sentiment, but given how often this has come up and rejected already, I guess it's unlikely to change. But you'd have to convince @brian-brazil and the other maintainers of the node exporter.

rchicoli:
Cool, I will give another try then. Olá amigo @brian-brazil could you write down your concerns again about this implementation?
It would be awesome if somehow we could sort this out in the best possible way.

brian-brazil:
As stated above and in other issues the webhook is the generic integration point with the Alertmanager. There's no plans to add a second one.

rchicoli:
Not even after a glass of caipirinha? Just kidding. That is a pity. So feel free to close this issue, if you don't change your mind. Sad news! I will have to keep using my own version of a compiled AM.

rchicoli:
Well, I would guess by now this feature is not wished. So I am closing this PR. Maybe next time I should ask which `receiver` is welcome before contributing.  

In addition, it makes sense to update this text.

https://prometheus.io/docs/alerting/configuration/
> Other receiver implementations available in version 0.0.4 of Alertmanager are not implemented yet. We are gladly accepting any contributions to add them to the new implementation.

juliusv:
@rchicoli That text was about receiver implementations that existed in version 0.0.4 (before a complete rewrite of Alertmanager which made it hard to just port the old receiver implementations). However, there was never a runcommand receiver, even in 0.0.4.

But it could be that all old receiver implementations now also exist in the current Alertmanager, which would still make that text outdated. Not sure though.

rchicoli:
No problem, I still think that this is a bit confusing, but I guess I misunderstand it. Anyways, next time I am up to writing a new receiver I will try to get in contact to you guys. Keep it up the good work.

Issue #586, opened 2016-12-21
Enable some form of Slack notifications parsing


wuub:
Hi,

right now mentioning slack users or teams is pretty tedious as it has to be done with slack IDs. Think
`<!subteam^ID|team>`  or `<!@U12345|nickname>` instead of normal `@team` or `@nickname`.

This is because slack does not parse messages received from webhooks (default `parse=none`). 
I think `parse=full` is unnecessary, but adding `link_names=1` here https://github.com/prometheus/alertmanager/blob/2433eebf481029d6f4b68c1874618d4cbe1d8d3c/notify/impl.go#L465 should be helpful.

What do you think? Are there any reasons why this is not there or will you accept a simple PR? :)

juliusv:
I think that makes sense. I only wonder whether it should be configurable (or always on), but probably configurable.

zerthimon:
is this gonna be merged any time soon ?

stuartnelson3:
Following up now, sorry for letting this slip

KirillTemnov:
We have same issue, can you fix it?

brian-brazil:
https://github.com/prometheus/alertmanager/pull/912 added this.

PaulAnnekov:
Looks like nowadays `link_names: true` + mentioning like `@username` stopped working. I tried even `<@username>` - no luck, it just prints this text as is. The only solution is to use user id, like `<@USERID>`. You can find user id using this manual https://help.workast.com/hc/en-us/articles/360027461274-How-to-find-a-Slack-user-ID.
I think this change is related to the blog post https://api.slack.com/changelog/2017-09-the-one-about-usernames.

PR #589, opened 2016-12-27
add option to automatically link named @user/@team in Slack notifications and turn on full slack message parsing

wuub:
link_names option is turned on by default as this is what most
people would expect, but can be disabled in configuration in those rare
instances when autolinking causes problems

should close: #586
cc: @juliusv

EDIT: This builds and passes rudimentary checks. I still need to  deploy it in our testing  to check how it behaves in real environment.

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -69,6 +69,7 @@ var (
 		Text:      `{{ template "slack.default.text" . }}`,
 		Fallback:  `{{ template "slack.default.fallback" . }}`,
 		LinkNames: true,
+		Parse:     "",
 	}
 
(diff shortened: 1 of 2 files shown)


wuub:
It works both in theory and in practice. But since no existing slack notification testing rig is in place adding one proved to a be a bit too time consuming. This is what I have now :)

stuartnelson3:
Sorry for such a long delay. Can you fix your conflicts? This should be an easy merge then.

wuub:
I think https://github.com/prometheus/alertmanager/pull/912 adds most of what this commit tried to do. Since it's already merged a new PR should be opened to add `parse` option.

PR #594, opened 2017-01-03
don't put ui/bindata.go in ui/bindata.go

alicebob:
No need to put Go code in bindata files.

For @fabxc ?


alicebob:
An other, aguably nicer, option would be to put the .go files in a different path. Up to you.

mxinden:
This looks fine to me. We can merge this, but it will be fixed with the new ui (See [ui-rewrite](https://github.com/prometheus/alertmanager/blob/ui-rewrite/Makefile#L63)).

If you don't mind I would close here and wait for the new ui to be merged soon.

@alicebob @fabxc What do you think?

alicebob:
Oh, there will be a new GUI, nice. I'll close this one then. Thanks!

PR #600, opened 2017-01-10
Add Matrix Notifier

brunoqc:
If I understand Matrix correctly, when you send an HTML message, you also send a plain text version so it can be shown in notifications and maybe to clients without HTML support.

I'm wondering if you find the default HTML message too complicated for the [configuration page](https://prometheus.io/docs/alerting/configuration/#slack-receiver-<slack_config>):

```
{{ define "matrix.default.html" }}<a href="{{ template "__alertmanagerURL" . }}"><font color="{{ if eq .Status "firing" }}red{{ else }}green{{ end }}"><b>{{ template "__subject" . }}</b></font></a>{{ end }}
```

A room id looks like `!AAffRRJHDD:matrix.org` so I used used `room_id: !<something>:matrix.org` in the simple.yml but maybe there's a better way.

Matrix support should allow to be able to change the server (i.e. matrix.org) but I'm thinking about doing it in the next PR.


brunoqc:
@fabxc

doc/examples/simple.yml
```diff
@@ -121,3 +121,9 @@ receivers:
     room_id: 85
     message_format: html
     notify: true
+- name: 'matrix'
+  matrix_configs:
+  - access_token: <access_token>
+    room_id: !<something>:matrix.org
```
rrix: Is there a way to make this not tied to matrix.org homeserver? I would like to use this with my homeserver without having to federate a room to matrix.org

notify/impl.go
```diff
@@ -542,6 +546,77 @@ func (n *Slack) retry(statusCode int) (bool, error) {
...
+}
+
+// Notify implements the Notifier interface.
+func (n *Matrix) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
+	var err error
+	var (
+		data     = n.tmpl.Data(receiverName(ctx), groupLabels(ctx), as...)
+		tmplText = tmplText(n.tmpl, data, &err)
+		tmplHTML = tmplHTML(n.tmpl, data, &err)
+		url      = fmt.Sprintf("https://matrix.org/_matrix/client/r0/rooms/%s/send/m.room.message/m%d.1?access_token=%s", n.conf.RoomID, time.Now().Unix(), n.conf.AccessToken)
```
freelock: Agree with rrix, URL of Matrix Homeserver should be configurable...

PR #602, opened 2017-01-11
Add proper JSON marshaling of config regexps.

devnev:
This fixes broken regexp rendering in the routing tree UI.

@beorn7

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -22,6 +22,7 @@ import (
 	"strings"
 	"time"
 
+	"encoding/json"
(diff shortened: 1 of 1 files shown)


beorn7:
@stuartnelson3 I vaguely remembered we dealt with exactly this issue in a PR before. Did that got lost somewhow?

beorn7:
OK, so we indeed fixed this before, but only in v0.4, never here.

I guess it makes sense to just merge this PR then, but I let @stuartnelson3 decide, who might be working on this very code.

PR #604, opened 2017-01-13
Include receiver name in notification metrics

mattbostock:
Add the receiver name to the `alertmanager_notifications_failed_total`
and `alertmanager_notifications_total` metrics so that it's possible to
identify which receiver is causing errors.

For example, I might have two webhook receivers, one for sending
notifications to [Jiralerts][] and another to send alerts to
Elasticsearch for reporting purposes. If the `webhook` integration
starts alerting, the receiver name allows you to determine which service
is causing the errors.

[Jiralerts]: https://github.com/fabxc/jiralerts

Example receiver configuration:

    receivers:
    - name: 'my_webhook'
      webhook_configs:

Example metrics produced:

    alertmanager_notifications_failed_total{integration="webhook",receiver="my_webhook"} 2
    alertmanager_notifications_total{integration="webhook",receiver="my_webhook"} 1

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -42,13 +42,13 @@ var (
 		Namespace: "alertmanager",
 		Name:      "notifications_total",
 		Help:      "The total number of attempted notifications.",
-	}, []string{"integration"})
+	}, []string{"integration", "receiver"})
 
 	numFailedNotifications = prometheus.NewCounterVec(prometheus.CounterOpts{
 		Namespace: "alertmanager",
 		Name:      "notifications_failed_total",
 		Help:      "The total number of failed notifications.",
-	}, []string{"integration"})
+	}, []string{"integration", "receiver"})
 )
 
 func init() {
@@ -207,7 +207,7 @@ func createStage(rc *config.Receiver, tmpl *template.Template, wait func() time.
 		var s MultiStage
 		s = append(s, NewWaitStage(wait))
 		s = append(s, NewDedupStage(notificationLog, recv))
-		s = append(s, NewRetryStage(i))
+		s = append(s, NewRetryStage(i, rc.Name))
 		s = append(s, NewSetNotifiesStage(notificationLog, recv))
 
 		fs = append(fs, s)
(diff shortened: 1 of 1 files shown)


brian-brazil:
In a large setup this could be thousands of timeseries, I think this is better suited to logs.

mattbostock:
Do setups exist where hundreds or thousands of receivers are configured?

That seems surprising; can you give an example of why one might have so many receivers?

brian-brazil:
When it comes to label dimensionality the mere possibility is the question. There are companies with thousands of teams, and who thus would have thousands of receivers.

mattbostock:
Thousands of teams sounds like an edge case, and I suspect that such a
company would likely not use one Prometheus instance to scrape Alertmanager
instance(s) that serve thousands of teams. I suspect that other limits
would be hit within Alertmanager first.

Having metrics that identify the receiver are useful if you want to alert
on erroring notifications but need the ability to silence a given receiver.
Without this ability, you'd have to silence by integration and run the risk
of silencing too much, or not silence at all and put up with unactionable
alerts.

brian-brazil:
> Thousands of teams sounds like an edge case, and I suspect that such a
company would likely not use one Prometheus instance to scrape Alertmanager
instance(s) that serve thousands of teams.

It's a case we need to support, and there would be one Prometheus monitoring it.

> I suspect that other limits would be hit within Alertmanager first.

That doesn't leave us free to add new limits in violation of our best practices.

> Having metrics that identify the receiver are useful if you want to alert
on erroring notifications but need the ability to silence a given receiver.

It's going to be a pretty weird case where that happens. Slack is unlikely to be half down for an extended period.

What you're proposing is too granular to be a label.

mattbostock:
> That doesn't leave us free to add new limits in violation of our best practices.

I agree. What I meant was that I think that the 'thousands of receivers' use case is unlikely enough that it's not worth worrying about.

> It's going to be a pretty weird case where that happens. Slack is unlikely to be half down for an extended period.

The use case I have in mind is webhooks, rather than Slack. Since webhooks are intended for generic use, it's quite likely that you may use more than one kind of webhook service.

> What you're proposing is too granular to be a label.

Understood, closing this PR.

mattbostock:
Following up on this, we're looking at applying this change as a patch to our Alertmanager build so that we can alert more usefully when alert notifications fail (some of our routes are less critical than others).

PR #615, opened 2017-01-27
Making slack color configurable via templates

shredder12:
The default value for color was not referring to a template variable.
This change ensures that colors can be easily overwritten via a custom template.

This is needed when one wants to use different color for low severity alerts.

@fabxc

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -59,7 +59,7 @@ var (
 		NotifierConfig: NotifierConfig{
 			VSendResolved: false,
 		},
-		Color:     `{{ if eq .Status "firing" }}danger{{ else }}good{{ end }}`,
+		Color:     `{{ template "slack.default.color" }}`,
 		Username:  `{{ template "slack.default.username" . }}`,
 		Title:     `{{ template "slack.default.title" . }}`,
 		TitleLink: `{{ template "slack.default.titlelink" . }}`,
--- a/template/default.tmpl
+++ b/template/default.tmpl
@@ -20,6 +20,7 @@
 {{ define "slack.default.iconemoji" }}{{ end }}
 {{ define "slack.default.iconurl" }}{{ end }}
 {{ define "slack.default.text" }}{{ end }}
+{{ define "slack.default.color" }}{{ if eq .Status "firing" }}danger{{ else }}good{{ end }}{{end}}
 
 
 {{ define "hipchat.default.from" }}{{ template "__alertmanager" . }}{{ end }}


brian-brazil:
This is already templatable.

shredder12:
@brian-brazil If you mean `color` parameter of slack_configs then yes its possible but in case of multiple slack_configs this would have to be done for each. This is the only way I know what would allow one to set a global default color for slack configs. If there is a different way, do let me know.

brian-brazil:
> If you mean color parameter of slack_configs then yes its possible but in case of multiple slack_configs this would have to be done for each.

Yes, that's the way it works.

shredder12:
This is very repetitive when you have multiple slack channels. Similar to all other slack_config variables, it would be good if we could overwrite `color` via templates too.

brian-brazil:
The presumption is that you have a configuration management system that can handle any repetitive parts of the configs for non-trivial configs.

shredder12:
The aim is to ensure that `color` parameter is also configurable through template files without every specifying it explicitly in the primary config, just like all other slack_config variables. I believe this is better than repeating configuration, irrespective of whether one is using a config management solution or not.

However, if you still feel that this is unnecessary, then feel free to close the PR. It was just an attempt to make things cleaner.

jangrewe:
@brian-brazil Why is `color` the only (as far as i can see) value that does NOT use the template value? I think this is very inconsistent, and quite frustrating to find out ;-)

brian-brazil:
This is already templated and documented as such. This change doesn't affect that.

jangrewe:
Well, you can override it in the `slack_config`, yes - but you can't define `slack.default.color` in a template, so i don't see how this is supposed to be "already templated"?

brian-brazil:
> Well, you can override it in the slack_config, yes

That's what being templated means, that you can override it with a template in the config.

jangrewe:
My definition of that would be to be able to use the templates, like i can do for all other values... if you'd called it "overrideable per slack config", fine...
But i don't get why this, and only this, value can't be set in a (separate) template file, like all other values?

brian-brazil:
The longer ones tend to be split out, whereas the simpler ones are kept inline. Overriding `slack.default.color` was never intended to be possible, that's only possible due to subsequent changes in Go. The idea is that you'll do whatever changes you need in the config file.

jangrewe:
Okay, i get the reason for it now, but: My color logic is definitely longer than some of the other values that are allowed to be externally templated ;-)

config/notifiers.go
```diff
@@ -59,7 +59,7 @@ var (
 		NotifierConfig: NotifierConfig{
 			VSendResolved: false,
 		},
-		Color:     `{{ if eq .Status "firing" }}danger{{ else }}good{{ end }}`,
+		Color:     `{{ template "slack.default.color" }}`,
```
nfirvine: For future spelunkers, this is wrong. It should be `{{ template "slack.default.color" .}}`. That last dot is very important.

PR #631, opened 2017-02-21
Remove AUTHORS.md and add MAINTAINERS.md

beorn7:
This implements the suggestion to name Stuart as the maintainer for
now. We can later revisit if this is a solution that works and is
sustainable for everyone involved.

@stuartnelson3 @fabxc


fabxc:
👍

Issue #567, opened 2016-11-22
Provide cli tool for interacting with alertmanager


Kellel:
I've been tossing around the idea of writing a cli tool for interacting with an alertmanager instance from cli. My primary interest is in being able to search/create/delete silences, but also being able to consume alerts via command line sounds mildly interesting. 

Are there any known tools that provide this functionality? Or would the community appreciate if I spent the time to toss something together? And if so would you like the contribution to be a part of this repo or a separate one?

Thanks!

brancz:
I think there are a lot of people who would be happy about such a tool (mainly for backup purposes), however, keep in mind that there are absolutely no stability guarantees regarding the Alertmanager API at this point. Even though it might not happen very soon, we have been thinking about cleaning up some parts of the API.

Just an idea, but I could see a tool similar to `promtool` that lives in the Alertmanager repo, but a separate binary, that can just like `promtool` be used to check configs, and perform other tasks such as the ones you described.

fabxc:
I also think it would be greatly appreciated by many!

Keeping it in this repo makes a lot of things simpler and also makes sure we release one with each AM version. Thus there'll be not much pain around API changes, as we are forced to fix the tool when changing our API.
Realistically, it won't be a big deal – API changes would be easy to swap and realistically, even if we replace the API, we'll probably keep it around for quite a while in parallel.

Kellel:
Ok. I'll get to work!

Kellel:
I've got a basic implementation working. I'm having some of my coworkers review it currently

brancz:
Cool, looking forward to it!

PR #636, opened 2017-02-25
Amtool implementation

Kellel:
As discussed in #567


Kellel:
Oops. Looks like I have some vendoring issues

beorn7:
Looks quite cool. But also very involved.

Before we dive into code-level review, it would be great if @fabxc and/or @brancz could comment if this is fundamentally sane and meeting their expectations. @stuartnelson3 might also have an opinion as he is getting into the AM API quite a bit recently.

cli/utils.go
```diff
@@ -0,0 +1,209 @@
...
+func GetAlertmanagerURL() (*url.URL, error) {
+	u, err := url.ParseRequestURI(viper.GetString("alertmanager.url"))
+	if err != nil {
+		return nil, errors.New("Invalid alertmanager url")
+	}
+	return u, nil
+}
+
+// Parse a list of labels (cli arguments)
+func parseMatchers(labels []string, resolve int) (types.Matchers, error) {
```
stuartnelson3: This is re-implementing functionality that already exists in PromQL, which seems like another vote for moving PromQL to a standalone package as mentioned in this PR https://github.com/prometheus/alertmanager/pull/633

cli/alert.go
```diff
@@ -0,0 +1,180 @@
...
+	displayAlerts := model.Alerts{}
+	for _, alert := range alerts {
+		// If we are only returning current alerts and this one has already expired skip it
+		if !expired {
+			if !alert.EndsAt.IsZero() && alert.EndsAt.Before(time.Now()) {
+				continue
+			}
+		}
+
+		if !showSilenced {
```
stuartnelson3: This filtering is something I've also implemented in the new alertmanager-ui. Seems like we should add it as a query string param to the url and do the filtering on the initial query, instead of duplicating our work in different consumers of the api?

stuartnelson3:
I think if we coordinate and add a few things to the alertmanager server it could remove a lot of logic from this pr (and others).

cli/alert.go
Kellel: Agreed. I kinda felt like it was a pain doing all this by hand in the cli tool

cli/utils.go
Kellel: Hmm. I looked at this, but it didn't immediately strike me as a fitting what I was trying to do. I'll have to take another look.

fabxc:
Yes, as discussed with @stuartnelson3 before, I'm happy to evaluate any API changes. The current API is officially not official for exactly that reason.

cli/format/format_extended.go
```diff
@@ -0,0 +1,94 @@
...
+func (formatter *ExtendedFormatter) SetOutput(writer io.Writer) {
+	formatter.writer = writer
+}
+
+func (formatter *ExtendedFormatter) FormatSilences(silences []types.Silence) error {
+	w := tabwriter.NewWriter(formatter.writer, 0, 0, 2, ' ', 0)
+	sort.Sort(ByEndAt(silences))
+	fmt.Fprintln(w, "ID\tMatchers\tStarts At\tEnds At\tUpdated At\tCreated By\tComment\t")
+	for _, silence := range silences {
+		line := fmt.Sprintf("%s\t%s\t%s\t%s\t%s\t%s\t%s\t", silence.ID, extendedFormatMatchers(silence.Matchers), FormatDate(silence.StartsAt), FormatDate(silence.EndsAt), FormatDate(silence.UpdatedAt), silence.CreatedBy, silence.Comment)
```
fabxc: Probably worth breaking out into multiple lines like 

```go
fmt.Sprintf("%s\t%s\t%s\t%s\t%s\t%s\t%s\t", 
    silence.ID,
    extendedFormatMatchers(silence.Matchers),
    FormatDate(silence.StartsAt),
    FormatDate(silence.EndsAt),
    FormatDate(silence.UpdatedAt),
    silence.CreatedBy,
    silence.Comment,
)
```

cli/format/format_simple.go
```diff
@@ -0,0 +1,70 @@
...
+	output := []string{}
+	for _, matcher := range matchers {
+		output = append(output, simpleFormatMatcher(*matcher))
+	}
+	return strings.Join(output, " ")
+}
+
+func simpleFormatMatcher(matcher types.Matcher) string {
+	if matcher.IsRegex {
+		return fmt.Sprintf("%s~=%s", matcher.Name, matcher.Value)
```
fabxc: `=~`

cli/root.go
```diff
@@ -0,0 +1,78 @@
...
+
+// RootCmd represents the base command when called without any subcommands
+var RootCmd = &cobra.Command{
+	Use:   "amtool",
+	Short: "Alertmanager CLI",
+	Long: `View and modify the current Alertmanager state.
+
+[Config File]
+
+The alertmanger tool will read a config file from $HOME/.amtool.yml or /etc/amtool.yml the options are as follows
```
fabxc: Would probably be better to read the path from `env AMTOOL_CONFIG` and default to `$HOME/.amtool.yml`.
Not sure whether we need `/etc/amtool.yml` to be generally considered but that could be the fallthrough option.

cli/utils.go
stuartnelson3: It's my hope to https://github.com/prometheus/alertmanager/pull/633 merged soon, so then the tool could accept a promql label matcher string (`{foo="bar", baz=~"qu.*"}`) and pass that on to the api.

this would be changing your api, however, so it's worth a discussion, and looping in @fabxc to see what he thinks. It looks like this might be expanding into generating a swagger api. Let's drop the discussion down to the main flow of the PR.

stuartnelson3:
There's a lot of moving pieces being proposed here, and I think it would be good to limit the scope of this PR.

It seems like we have three main points:

1. This PR as it is: it's user-interface, and if we should be using cobra/viper
2. Coordinating interaction with the API with https://github.com/prometheus/alertmanager/pull/633
3. Re-create the API with gRPC, generate a json api, visualize with swagger

I would say we focus first on 1, and then we can use any changes to inform its implementation and interaction with the API changes in 2. I would vote for 3 to be a separate PR.

Thoughts?

fabxc:
Yes, absolutely. This grew in scope a fair bit. We should focus on getting
this in and just being clear about a lot of its aspects changing a lot.
In particular, amtool and promtool potentially going into a single binary
and separate at some point.

cli/utils.go
Kellel: The other issue I had was that I needed to filter silences. Will that be covered in #633 ?

cli/alert.go
```diff
@@ -0,0 +1,180 @@
...
+  	If alertname is ommited and the first argument does not contain a '=' or a
+  	'=~' then it will be assumed to be the value of the alertname pair.
+
+  amtool alert query 'alertname=~foo.*'
+
+  	As well as direct equality, regex matching is also supported. The '=~' syntax
+  	(similar to prometheus) is used to represent a regex match. Regex matching
+  	can be used in combination with a direct match.
+
+  amtool alert query alertname=foo node={bar,baz}
```
stuartnelson3: This is a cool idea, but it's introducing an additional way to parse labels that isn't supported in prometheus (and is equivalent to `node=~"(bar|baz)"`). I would personally vote for maintaining the same label matching syntax as prometheus.

cli/alert.go
stuartnelson3: The functionality is already available here: https://github.com/prometheus/alertmanager/blob/master/dispatch/dispatch.go#L132-L134

All that needs to be done is parsing a `?silenced=true` to show silenced alerts, and by default don't return them in the response.

cli/config.go
```diff
@@ -0,0 +1,86 @@
...
+	"encoding/json"
+	"errors"
+	"net/http"
+	"path"
+	"time"
+
+	"github.com/prometheus/alertmanager/cli/format"
+	"github.com/prometheus/alertmanager/config"
+	"github.com/spf13/cobra"
+	//flag "github.com/spf13/pflag"
```
stuartnelson3: commented out import

cli/config.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/spf13/viper"
+)
+
+// Config is the response type of alertmanager config endpoint
+// Duped in cli/format needs to be moved to common/model
+type Config struct {
+	Config      string            `json:"config"`
+	ConfigJSON  config.Config     `json:configJSON`
+	VersionINFO map[string]string `json:"versionInfo"`
```
stuartnelson3: I think the go naming convention for non-acronyms would have this be `VersionInfo`

cli/alert.go
Kellel: The key here is that creating multiple silences is awkward without being able to use shell expansions. I understand that it's not a big win for searching, but I think it is a huge help when creating silences.

stuartnelson3: That's a good point. @fabxc ?

cli/utils.go
stuartnelson3: Filtering silences is available in that pr, in https://github.com/prometheus/alertmanager/pull/633/commits/0954959e821f76ef37b925a4f819d8ceff79233d

I'm going to try to get that PR pushed through, then it should just be a few easy changes here and we can get this merged.

stuartnelson3:
#633 has been merged. Can you rebase and see how the filtering built into the http api works for you? `/silences?filter=<url encoded label matchers>` and `/alerts?filter=<url encoded label matchers>`

cli/utils.go
Kellel: Cool. Looking at that.

Kellel: Quick question.
I implemented a feature where if a user doing a query (silence or alert) does something like the following:
    amtool alert query Host

With no matcher name specified then I attempt to match the alertname using a prefix match. Is this something we want to support? Moving to promql should do everything I want, but i'm curious if I should spend the time to re-implement this functionality?

stuartnelson3: hm, @fabxc ?

cli/alert.go
fabxc: Mh, how do you mean creating multiple silences? In the general case, there should be a single silence matching multiple labels if necessary rather than duplicated silences.

This comment is on alert querying, which is a different thing again.

I think I need a better example what's tried to be solved here.

In general, not growing representations for the same thing is preferable of course.

Kellel: Let me try and make this example clearer. I very well could be over complicating this.

Lets say I have an alert named `Http_Response_Slow` and an alert named `Http_Response_5xx` and I have some number of boxes reporting the metrics that compose this alert with their name in a `node` label.

If I want to silence both alerts for a box reporting as `node=www1` and `node=www2`, but not for the rest of the boxes I would have to create the following four silences.

`amtool silence add alertname=Http_Response_Slow node=www1`
`amtool silence add alertname=Http_Response_Slow node=www2`
`amtool silence add alertname=Http_Response_5xx node=www1`
`amtool silence add alertname=Http_Response_5xx node=www2`

With the ability to use the shell expansions I could make the same four silences with one command:

`amtool silence add alertname={Http_Response_{Slow,5xx} node=www{1,2}`

Does this make sense?

stuartnelson3: I guess the confusion is why doing something like

```
amtool silence add alertname=~"Http_Response_.*" node=~"www(1|2)"
```

and creating a single silence to cover all four alerts isn't preferable.

Kellel: I see what you are saying. I guess it depends on the use case. I would be willing to drop this if we thing that it is unnecessary. It would simplify a good deal of the code

Kellel:
Working on a refactor to use the filter endpoints introduced in #633. Should be ready in a day or two

Kellel:
I've opened #661 to help incorporate filtering changes into my code base

stuartnelson3:
Do you want to introduce the changes to the vendored file directly in the PR so we can see them in context?

Kellel:
Sure. Expect that update in a day or two. I've had other priorities at work. On another note would you like me to spin off a PR for removing silenced alerts in a query? Or have you already started one for that?

Kellel:
So I've got the code compiling with the changes to filtering, but if you try and run anything you will notice that I am getting "bad matcher format" on any query.

I've narrowed it down to this regex `(?:\s?)(\w+)(=|=~|!=|!~)\"((?:\W*|\w*)+)\"` which on the surface seems fine, but is going to require a bit of hand-holding to get to work in a cli tool (due to how bash works).

Example:

    `amtool alert alertname="foo"`
     ARGV:
         ["amtool", "alert", "alertname=foo"]

So my question is:

1. Should we relax this regex to allow omission of the quote characters?
2. Should I just parse the matchers by hand and not use github.com/prometheus/alertmanager/pkg/parse ?

vendor/github.com/prometheus/prometheus/pkg/labels/matcher.go
```diff
@@ -60,6 +60,11 @@ func (m *Matcher) String() string {
 	return fmt.Sprintf("%s%s%q", m.Name, m.Type, m.Value)
 }
 
+// Allow fetching of regex for conversion to internal types
+func (m *Matcher) Regexp() *regexp.Regexp {
```
Kellel: Code I want to add to the prometheus repo

cli/utils.go
```diff
@@ -0,0 +1,151 @@
...
+// Only valid for when you are going to add a silence
+// Doesn't allow negative operators
+func TypeMatcher(matcher labels.Matcher) (types.Matcher, error) {
+	var typeMatcher types.Matcher
+
+	switch matcher.Type {
+	case labels.MatchEqual:
+		typeMatcher = *types.NewMatcher(model.LabelName(matcher.Name), matcher.Value)
+	case labels.MatchRegexp:
+		typeMatcher = *types.NewRegexMatcher(model.LabelName(matcher.Name), matcher.Regexp())
```
Kellel: This is where I feel that I need the change to github.com/prometheus/prometheus/pkg/labels

stuartnelson3:
Go ahead and change the regex so the double quotes are optional

cli/utils.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// Expand a list of matchers into a list of all combinations of matchers for each matcher that has the same Name
+// ```
+// alertname=foo instance=bar instance=baz
+// ```
+// Goes t
+// ```
+// alertname=foo instance=bar
+// alertname=foo instance=baz
```
stuartnelson3: In this example, what's the reason for splitting this into multiple silences with a single `instance` label instead of a single silence with multiple instance labels? `alertname=foo instance=bar instance=baz` affects the same alerts, and produces (to me) the expected result from using the command line tool. I would be surprised if, for instance, this label set:

```
alertname=fooService alertname=bazService statusCode=400 statusCode=500
```

produced four silences:

```
alertname=fooService  statusCode=500
alertname=fooService statusCode=400
alertname=bazService statusCode=500
alertname=bazService statusCode=400
```

So I'm just curious what the use-case is that this is solving?

cli/utils.go
stuartnelson3: Actually, if you look at the struct definition: 
https://github.com/prometheus/alertmanager/blob/master/types/match.go#L25-L31
Only `Name`, `Value`, and `IsRegex` actually get sent when this is serialized.

You could simplify this to:

```go
func TypeMatcher(matcher labels.Matcher) (types.Matcher, error) {
	typeMatcher := types.NewMatcher(model.LabelName(matcher.Name), matcher.Value)

	switch matcher.Type {
	case labels.MatchEqual:
		// Not strictly necessary, because that's the initialization
		// value, but we need to exercise this code path.
		typeMatcher.IsRegex = false
	case labels.MatchRegexp:
		typeMatcher.IsRegex = true
	default:
		return types.Matcher{}, errors.New("invalid match type for creation operation: %s", m.Type)
	}
	return *typeMatcher, nil
}
```

cli/utils.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	return u, nil
+}
+
+// Parse a list of labels (cli arguments)
+func parseMatchers(inputLabels []string) ([]labels.Matcher, error) {
+	matchers := make([]labels.Matcher, 0)
+
+	for _, v := range inputLabels {
+		fmt.Println(v)
```
stuartnelson3: This can probably be removed?

cli/silence_query.go
```diff
@@ -0,0 +1,143 @@
...
+
+	err = decoder.Decode(&silenceResponse)
+	if err != nil {
+		return []types.Silence{}, err
+	}
+
+	return silenceResponse.Data, nil
+}
+
+func query(cmd *cobra.Command, args []string) error {
```
stuartnelson3: Given the filtering available at the silences endpoint, the merge code and multiple requests can be replaced by a single command line query:

```
amtool silence query alertname=~foo(a|b) node=~(bar|baz)
```

And then this can be sent to the server once parsed as:

```
http://alertmanager.cloudflare.example.biz/api/v1/silences?filter=%7Balertname%3D~%22foo(a%7Cb)%22%2C%20node%3D~%22(bar%7Cbaz)%22%7D
```

I think with this and my other comment the whole merge concept could be removed, which would reduce the number of web requests made to the backend and simplify the code being run in the tool.

cli/alert.go
stuartnelson3: I think right now we're leaning towards dropping this in favor of using a regex or having the user create their own e.g. bash loop

stuartnelson3: To expand:

I think the team tends to favor conservatively introducing new features. They're 100% reasonable and if this ends up being something that people request it'll happen, but to start out keeping it simple is the preference.

cli/alert.go
```diff
@@ -0,0 +1,189 @@
...
+
+	showSilenced, err := alertFlags.GetBool("silenced")
+	if err != nil {
+		return err
+	}
+
+	fetchedAlerts := model.Alerts{}
+	if len(groups) < 1 {
+		fmt.Println("No specified matchers")
+		fetchedAlerts, err = fetchAlerts(nil)
```
stuartnelson3: A slice contains a pointer to an internal array, so passing around a pointer to a slice isn't really something that happens in go.

The flow of this seems to be doing a lot of work:

1. Parse the user input into matchers
2. Pass matchers to fetch function
3. Stringify matchers for the query string

I think it would make sense to massage the string args into the format that gets sent to the server, and then let the server handle parsing and error handling:

```go
l := fmt.Sprintf("{%s}", strings.Join(os.Args[2:], ","))
fetched, err := fetchAlerts(l)

// in fetchAlerts
qs := url.QueryEscape("filter=" + l)
```

And just send that to the back end -- if there's an error, it'll be handled consistently from the server.

cli/alert.go
```diff
@@ -0,0 +1,189 @@
...
+		if err != nil {
+			return err
+		}
+	}
+
+	// Fetch all alerts that match each group of matchers
+	for _, matchers := range groups {
+		alerts, err := fetchAlerts(&matchers)
```
stuartnelson3: I think having a single query that gets sent will end up being easier to reason about

cli/format/format_extended.go
```diff
@@ -0,0 +1,108 @@
...
+		line := fmt.Sprintf("%s\t%s\t%s\t%s\t%s\t%s\t%s\t",
+			silence.ID,
+			extendedFormatMatchers(silence.Matchers),
+			FormatDate(silence.StartsAt),
+			FormatDate(silence.EndsAt),
+			FormatDate(silence.UpdatedAt),
+			silence.CreatedBy,
+			silence.Comment,
+		)
+		fmt.Fprintln(w, line)
```
stuartnelson3: This could be smooshed into a single

```
fmt.Fprintf(
	w, 
	"%s\t%s\t%s\t%s\t%s\t%s\t%s\t",
	silence.ID,
	extendedFormatMatchers(silence.Matchers),
	FormatDate(silence.StartsAt),
	FormatDate(silence.EndsAt),
	FormatDate(silence.UpdatedAt),
	silence.CreatedBy,
	silence.Comment,
)
```

cli/format/sort.go
```diff
@@ -0,0 +1,36 @@
...
+func (s ByEndAt) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
+func (s ByEndAt) Less(i, j int) bool { return s[i].EndsAt.Before(s[j].EndsAt) }
+
+type ByStartsAt model.Alerts
+
+func (s ByStartsAt) Len() int           { return len(s) }
+func (s ByStartsAt) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
+func (s ByStartsAt) Less(i, j int) bool { return s[i].StartsAt.Before(s[j].StartsAt) }
+
+func UniqueAlerts(alerts model.Alerts) model.Alerts {
```
stuartnelson3: This doesn't seem to be used

cli/format/sort.go
```diff
@@ -0,0 +1,36 @@
...
+		for j := 0; j < i; j++ {
+			if alerts[j] == alerts[i] {
+				exists = true
+				break
+			}
+		}
+		if !exists {
+			output = append(output, alerts[i])
+		}
+	}
```
stuartnelson3:
```go
output := model.Alerts{}
uniq := make(map[model.Alert]struct{}}
for _, a := range alerts {
	if _, prs := uniq[a]; prs {
		continue
	}    
	uniq[a] = struct{}{}
	output = append(output, a)
}
```

cli/silence_add.go
```diff
@@ -0,0 +1,172 @@
...
+
+		u, err := GetAlertmanagerURL()
+		if err != nil {
+			return err
+		}
+		u.Path = path.Join(u.Path, "/api/v1/silences")
+
+		buf := bytes.NewBuffer([]byte{})
+		enc := json.NewEncoder(buf)
+		err = enc.Encode(silence)
```
stuartnelson3: `json.NewEncoder(buf).Encode(silence)`

The encode and decode stuff doesn't get re-used, so no need to assign it to a var.

cli/silence_add.go
```diff
@@ -0,0 +1,172 @@
...
+		res, err := http.Post(u.String(), "application/json", buf)
+		if err != nil {
+			return err
+		}
+
+		defer res.Body.Close()
+		decoder := json.NewDecoder(res.Body)
+
+		response := addResponse{}
+		err = decoder.Decode(&response)
```
stuartnelson3: Same as above

cli/silence_add.go
```diff
@@ -0,0 +1,172 @@
...
+		if err != nil {
+			return errors.New(fmt.Sprintf("Unable to parse silence json response from %s", u.String()))
```
stuartnelson3: `errors.New(fmt.Sprintf("..", thing1, thing2)) == fmt.Errorf("..", thing1, thing2)`

cli/utils.go
Kellel: I was worried about the fact that we loose this: `"^(?:" + v + ")$"`

But I guess it is auto added for both alert and silences at the alertmanager api.

cli/utils.go
Kellel: Sorry debugging line. Was just trying to get it into a state where I could show the issues.

cli/silence_add.go
Kellel: TIL

cli/alert.go
Kellel: Sounds reasonable to me

Kellel:
Sorry for not having an update in quite a few days. Been busy with other things at work.

stuartnelson3:
Running the top level `amtool alert` is returning all alerts, but if I try a regex matcher I get an error:

```
# ./amtool alert query 'alertname=~MySQL.*'
rror: [bad_data] bad matcher format
Usage:
  amtool alert [flags]

Flags:
      --expired    Show expired alerts as well as active
  -s, --silenced   Show silenced alerts

Global Flags:
      --alertmanager.url string   Alertmanager to talk to
      --config string             config file (default is $HOME/.amtool.yml)
  -o, --output string             Output formatter (simple, extended, json) (default "simple")
  -v, --verbose                   Verbose running information

[bad_data] bad matcher format
```

The documentation seems to say this is valid though:

```
amtool alert query 'alertname=~foo.*'
```

Actually, it seems like every query I'm running is resulting in a bad matcher error?

```
# ./amtool alert query alertname=NodeHasMemoryErrors
Error: [bad_data] bad matcher format
```

Same deal with silences: Showing all silences works, but when I try to query for them it says bad matcher.

Kellel:
Looks like I forgot to register the `query` alias for the alert command. :rofl: 

Silence seems to be working in my testing environment. Did you rebuild your alertmanager and docker container?

stuartnelson3:
Ahhhh yeah I didn't rebuild the actual alertmanager binary :)

This is super cool!

Querying seems to be working, except for the bare alertname case:

```
# ./amtool --alertmanager.url $url alert query alertname=K8sCPURequestTooLow
Alertname            Starts At                Summary                 
K8sCPURequestTooLow  2017-04-10 06:27:06 UTC  Pod CPU use is too low 

# ./amtool --alertmanager.url $url alert query K8sCPURequestTooLow
Error: [bad_data] bad matcher format
Usage:
  amtool alert query [flags]

Flags:
      --expired    Show expired alerts as well as active
  -s, --silenced   Show silenced alerts

Global Flags:
      --alertmanager.url string   Alertmanager to talk to
      --config string             config file (default is $HOME/.amtool.yml)
  -o, --output string             Output formatter (simple, extended, json) (default "simple")
  -v, --verbose                   Verbose running information

[bad_data] bad matcher format
```

Other than that this looks pretty cool to me.

One thing that would be nice is also filtering by receiver as a sort of special key-value pair, but I think we can wait on that because it's also a desired feature in the UI, so doing that via a query param in the API is probably what will happen.

Kellel:
Oh I don't think I re-implemented the bare alertname functionality now that I'm not parsing any of the arguments. 

Should be reasonably simple to re-add

stuartnelson3:
seems to also be an error when using the config subcommand:

```
# ./amtool config -v
Using config file: /Users/stuartnelson/.amtool.yml
Error: json: cannot unmarshal string into Go struct field Route.match_re of type config.Regexp
```

expiring an already expired silence:

```
# ./amtool silence expire a6c18c6b-9cb9-4826-a657-f054bef0a121
Error: end time must not be before start time
Usage:
  amtool silence expire [flags]

Global Flags:
      --alertmanager.url string   Alertmanager to talk to
      --config string             config file (default is $HOME/.amtool.yml)
  -o, --output string             Output formatter (simple, extended, json) (default "simple")
  -v, --verbose                   Verbose running information

end time must not be before start time
```

it also seems that any error output is duplicated? I noticed this before but forgot to mention it.

One idea that I had while interacting with this:

It's awesome that I can e.g.:

```
# ./amtool silence expire $uuid1 $uuid2 $uuid3
```

But this requires me to do some bashery to grab the IDs:

```
# ./amtool silence expire (./amtool silence query 'alertname=~TestAmTool.*' | awk 'NR>1{print $1}')
```

Taking a cue from the docker cli, what do you think about adding a `-q` flag to show just IDs? So this would then collapse into

```
# ./amtool silence query foo=bar -q
c0c6cb44-cac5-4c26-a797-a7f688009890
c776bc72-7d88-438b-bed7-0485d1ab975b
8fa7a0e6-5723-4c9f-892e-8b8d99a93eb3
1795274c-555e-46dd-b861-da5da459bc22
76441da9-f583-4188-a7f7-e674f24092e0
6c4f4425-62f3-487a-bd83-92462662895b
d31b41fe-c1da-47d7-8dba-2b33438b87ef
e91409e4-e356-43c7-bd2f-422234d5b15a
82af6174-f098-47a9-b4e3-c6700b1ff5e7
0ce9c4c3-5001-4cf2-8d7d-67702a02a0d9
a368b6ce-5e27-4191-a315-18c649e3cce5
```

Seems to be `-q` and `--quiet`. What do you think about this?

stuartnelson3:
There's also an inconsistency between the UI and the cli tool when making a silence with a regex:

```
# ./amtool silence add 'alertname=~TestAmTool.*' --comment='test out creation'
dfbacf0a-a4ed-4bf2-80d5-26dad57f8b65
```

The created regex ends up being set as the name: 
image

But that's only supposed to happen internally when creating the regex. Creating the same thing through the ui results in `alertname=~"TestAmTool.*"` without the starting and ending anchors.

Kellel:
I think the `-q` `--quiet` would be a great addition. 

After a little bit of debugging it seems that the `"(?:TestAmTool.*)$"` bug seems to be coming directly from the way that `pkg/parse.go` works I was hoping to use that code to do the parsing of matchers.

vendor/github.com/prometheus/prometheus/pkg/labels/matcher.go
```diff
@@ -46,8 +46,7 @@ func NewMatcher(t MatchType, n, v string) (*Matcher, error) {
 		Value: v,
 	}
 	if t == MatchRegexp || t == MatchNotRegexp {
-		m.Value = "^(?:" + v + ")$"
-		re, err := regexp.Compile(m.Value)
+		re, err := regexp.Compile("^(?:" + m.Value + ")$")
```
Kellel: This seems to be the cause of the regex information showing up in the ui since I'm using the internal parser to parse silence matchers.

Is this change acceptable? All tests are still passing

stuartnelson3: I would prefer not to diverge from vendored packages.

I went through the code and made a couple changes, but I'll leave it to you to apply them to your branch.

First, reverting the change to the vendored package:
```
git revert 5d0bba4b3b3e682b4eae10381ce96660d5c3979a
```

Then here's a patch file that you can apply. Read through it to see what it does, but basically it separates out the actual parsing code in `pkg/parse` so that there's an exported function for parsing the raw input string, which the CLI then uses to send to the API.

[parse_regex.txt](https://github.com/prometheus/alertmanager/files/916562/parse_regex.txt)

Note: Apparently github won't let me upload a `.patch` file, so it says text, but you should be able to apply it all the same.

Kellel:
What are our next steps to get this merged?

stuartnelson3:
There are a few things that need to be fixed:

1. Config output is failing to marshal my config:
```
# ./amtool config -v
Using config file: /Users/stuartnelson/.amtool.yml
Error: json: cannot unmarshal string into Go struct field Route.match_re of type config.Regexp
Usage:
  amtool config [flags]

Global Flags:
      --alertmanager.url string   Alertmanager to talk to
      --config string             config file (default is $HOME/.amtool.yml)
  -o, --output string             Output formatter (simple, extended, json) (default "simple")
  -v, --verbose                   Verbose running information

json: cannot unmarshal string into Go struct field Route.match_re of type config.Regexp
# cat ~/.amtool.yml 
alertmanager.url: http://redacted.net:9093
author: stn@soundcloud.com
comment_required: true
output: simple
```
2. Output errors are printed twice (see above)
3. Can you move the `--quiet` flag to the top-level `silence` subcommand? So regardless of if I'm querying or looking at all silences, this will output the ids (currently it's just working for query)

After that this should be ready to merge. Thanks for you patience!

stuartnelson3:
Awesome. I'm going to go ahead and merge this because it's been so long in a branch. There's an issue I would like to address still, but I'll put that in issue #721

Issue #582, opened 2016-12-19
Alertmanager refueses to start on Centos 7 box due to incorrect "no valid network interface"


clarkezone:
[root@dev2 alertmanager-0.5.1.linux-amd64]# ./alertmanager 
panic: no valid network interfaces

goroutine 1 [running]:
panic(0x86ccc0, 0xc420142500)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
main.mustHardwareAddr(0x92a522, 0x13)
	/go/src/github.com/prometheus/alertmanager/cmd/alertmanager/main.go:408 +0x158
main.main()
	/go/src/github.com/prometheus/alertmanager/cmd/alertmanager/main.go:85 +0x335

despite a perfectly valid and working network (i've XX'd out the actual values below):

ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN 
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
2: venet0: <BROADCAST,POINTOPOINT,NOARP,UP,LOWER_UP> mtu 1500 qdisc noqueue state UNKNOWN 
    link/void 
    inet 127.0.0.1/32 scope host venet0
    inet XX.XX.XXX.XX/32 brd XX.XX.XXX.X3 scope global venet0:0
3: gre0: <NOARP> mtu 1476 qdisc noop state DOWN 
    link/gre 0.0.0.0 brd 0.0.0.0
4: gretap0: <BROADCAST,MULTICAST> mtu 1476 qdisc noop state DOWN qlen 1000
    link/ether 00:00:00:00:00:00 brd ff:ff:ff:ff:ff:ff

molecul:
I also encountered the same problem. But with operation system 'Ubuntu-14.04', on OpenVZ host hypervisor. 

root@ubuntu:/home/root/alertmanager-0.5.1.linux-386# ./alertmanager -config.file=simple.yml
panic: no valid network interfaces

goroutine 1 [running]:
panic(0x848ff00, 0x18a24df8)
        /usr/local/go/src/runtime/panic.go:500 +0x331
main.mustHardwareAddr(0x0, 0x0)
        /go/src/github.com/prometheus/alertmanager/cmd/alertmanager/main.go:408 +0x1ce
main.main()
[...]

PR #638, opened 2017-02-27
Only find MAC address if no command-line flag value given

stuartnelson3:
Defaulting to the machine's MAC address fails
sometimes fails and causes a panic. Allow the user to
specify custom address to skip this so they can
run AlertManager.

Addresses #582 

@beorn7 @brancz @fabxc this is more or less a quick fix so that users can manually enter something while we decide how to handle this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,12 +174,12 @@ To create a highly available cluster of the Alertmanager the instances need to
 be configured to communicate with each other. This is configured using the
 `-mesh.*` flags.
 
-- `-mesh.hardware-address` string: mesh peer ID (default "&lt;hardware-mac-address&gt;")
+- `-mesh.peer-id` string: mesh peer ID (default "&lt;hardware-mac-address&gt;")
 - `-mesh.listen-address` string: mesh listen address (default "0.0.0.0:6783")
 - `-mesh.nickname` string: mesh peer nickname (default "&lt;machine-hostname&gt;")
-- `-mesh.peer` value: initial initial peers (may be repeated)
+- `-mesh.peer` value: initial peers (may be repeated)
 
-The `mesh.hardware-address` flag is used as a unique ID among the peers. It
+The `mesh.peer-id` flag is used as a unique ID among the peers. It
 defaults to the MAC address, therefore the default value should typically be a
 good option. The same applies to the default of the `mesh.nickname` flag, as it
 defaults to the hostname. The chosen port in the `mesh.listen-address` flag is
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ route:
   # all alerts. It needs to have a receiver configured so alerts that do not
   # match any of the sub-routes are sent to someone.
   receiver: 'team-X-mails'
-  
+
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -174,10 +174,10 @@ To create a highly available cluster of the Alertmanager the instances need to
 be configured to communicate with each other. This is configured using the
 `-mesh.*` flags.
 
-- `-mesh.hardware-address` string: MAC address, i.e. mesh peer ID (default "&lt;hardware-mac-address&gt;")
+- `-mesh.hardware-address` string: mesh peer ID (default "&lt;hardware-mac-address&gt;")
 - `-mesh.listen-address` string: mesh listen address (default "0.0.0.0:6783")
-- `-mesh.nickname` string: peer nickname (default "&lt;machine-hostname&gt;")
-- `-mesh.peer` value: initial peers (may be repeated)
+- `-mesh.nickname` string: mesh peer nickname (default "&lt;machine-hostname&gt;")
+- `-mesh.peer` value: initial initial peers (may be repeated)
```
brancz: initial initial

README.md
```diff
@@ -174,10 +174,10 @@ To create a highly available cluster of the Alertmanager the instances need to
 be configured to communicate with each other. This is configured using the
 `-mesh.*` flags.
 
-- `-mesh.hardware-address` string: MAC address, i.e. mesh peer ID (default "&lt;hardware-mac-address&gt;")
+- `-mesh.hardware-address` string: mesh peer ID (default "&lt;hardware-mac-address&gt;")
```
brancz: I've thought about this before and was thinking that we should probably change this to `-mesh.peer-id` and have it default to the mac address, but not call it hardware address.

README.md
stuartnelson3: woops!

stuartnelson3:
Sounds good. I can leave this here and we can discuss the breaking change, and if we want to then add an alternative to the mac address as a default.

beorn7:
If we are reluctant to have a breaking change now, let's just merge this with the functional change but without the flag change.

But perhaps we are not worried about breaking changes at this pre-1.0 point?

stuartnelson3:
I think it's a valid point that this is experimental. We aren't cutting a release, so anyone using this would be compiling their own version.

brancz:
It's not that I'm reluctant to the breaking change, just want to agree on it :slightly_smiling_face:

brian-brazil:
I think a breaking change is okay, this feature is marked experimental and I can't imagine too many people setting this currently.

On the naming, this is very close to `-mesh.peer` which I can imagine causing confusion. Maybe something like unique-peer-id or my-peer-id?

fabxc:
I think `mesh.peer-id` is just fine. So is the breaking change here in my opinion.

stuartnelson3:
The naming is similar, but I say since it's experimental, we can change if we get user complaints. I'll figure out why the tests are failing and wait for a thumb from someone.

brancz:
lgtmog

PR #639, opened 2017-02-28
template: Add reReplaceAll function

jamesog:
This allows for using string replace inside templates.

e.g.:

```
{{ replace .GroupLabels.alertname "_" " " -1 }}
```

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -17,6 +17,7 @@ import (
 	"bytes"
 	"net/url"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 	"time"
@@ -129,6 +130,10 @@ var DefaultFuncs = FuncMap{
 	},
 	"safeHtml": func(text string) tmplhtml.HTML {
 		return tmplhtml.HTML(text)
+	},
+	"reReplaceAll": func(pattern, repl, text string) string {
+		re := regexp.MustCompile(pattern)
+		return re.ReplaceAllString(text, repl)
 	},
 }
 


jamesog:
I made this change from the v0.4.2 tag so submitted the PR against the release-0.4 branch, but not sure if this would be wanted against another (newer) branch.

brian-brazil:
This should go against master.

This should follow the naming and semantics of Prometheus templates: https://prometheus.io/docs/visualization/template_reference/#strings

stuartnelson3:
@jamesog can you branch from master and add this change? v0.4.x isn't being supported.

jamesog:
Yep, will do. I've been sick the last couple of days but I'll get to it ASAP.

Can you clarify what you mean about the naming? I was following the example of the `join` function.

brian-brazil:
`reReplaceAll` is the function name in Prometheus.

jamesog:
Would `stringReplace` be acceptable?

brian-brazil:
The naming should be consistent across both.

jamesog:
Right, but I assume the `re` in `reReplaceAll` refers to the fact that it's using `regexp.ReplaceAll`. In this case I'm using `strings.Replace`, so I'm just trying to clarify how you would want to name it in this case, or whether you'd prefer it to use `regexp` instead of `strings`.

brian-brazil:
Stick with the regex, if someone has a use case not covered by that we can consider it then.

jamesog:
Thanks very much, appreciate the clarification. I've made the change to use `reReplaceAll` as in Prometheus. I haven't been able to test this yet as we're still using 0.4.2 in production, but I'll see if I can do that now.

jamesog:
That's tested now and working as expected.

brian-brazil:
Thanks!

Zevran:
This feature is not included in the latest available release which is obvious but kind of confusing for newbies like me !

Issue #640, opened 2017-03-02
Make `-mesh.peer` and `-alertmanager.url` invocation uniform


stuartnelson3:
For HA alertmanager, specifying peers is done by repeating the `-mesh.peer` flag:

```
./alertmanager -mesh.peer=instanceA.net:6783 -mesh.peer=instanceB.net:6738
```

In prometheus, specifying multiple alertmanagers uses a comma separated list:

```
./prometheus -config.file=prometheus.yml -alertmanager.url http://localhost:9095,http://localhost:9094,http://localhost:9093
```
(taken from alertmanager readme)

This is made more confusing by the current format of the readme, which states that `-mesh.peer may be repeated`, and then has an example of the prometheus invocation.

As an immediate fix I'll add an example to the README.

I understand changing the commandline flags in prometheus would be a breaking change. I personally prefer the repeated flags and would be willing to wait for v2 to have it, but I would like your opinions on if we should change alertmanager to use a comma separated list like prometheus now, or just document the change and let users live with it.

brian-brazil:
`-alertmanager.url` will be going away in 2.0 as-is.

stuartnelson3:
In favor of using discovery in the config file?

brian-brazil:
Yes. The flags were only temporary until we had a better idea of the requirements on that end.

juliusv:
Yep. Closing, as there's nothing to be done on the AM side for this.

PR #641, opened 2017-03-02
Update README.md

stuartnelson3:
Make it more clear that -mesh.peer flag should be
repeated for each peer.

Addresses #640


brian-brazil:
:+1:

PR #644, opened 2017-03-07
api: Expose mesh status

mxinden:
The weaveworks mesh package reveals information about the current status
of the mesh network between alertmanager instances. This commit exposes
the current address and connection status of each instance connected to
the targeted alertmanager instance via the /status API endpoint.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -34,6 +34,7 @@ import (
 	"github.com/prometheus/alertmanager/silence"
 	"github.com/prometheus/alertmanager/silence/silencepb"
 	"github.com/prometheus/alertmanager/types"
+	"github.com/weaveworks/mesh"
 )
 
 var (
@@ -77,6 +78,7 @@ type API struct {
 	configJSON     config.Config
 	resolveTimeout time.Duration
 	uptime         time.Time
+	mrouter        *mesh.Router
 
 	groups func() dispatch.AlertOverview
 
@@ -86,13 +88,14 @@ type API struct {
 }
 
 // New returns a new API.
-func New(alerts provider.Alerts, silences *silence.Silences, gf func() dispatch.AlertOverview) *API {
+func New(alerts provider.Alerts, silences *silence.Silences, gf func() dispatch.AlertOverview, router *mesh.Router) *API {
 	return &API{
 		context:  route.Context,
 		alerts:   alerts,
 		silences: silences,
 		groups:   gf,
 		uptime:   time.Now(),
+		mrouter:  router,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


mxinden:
@fabxc Could you review this PR? What other [status](https://godoc.org/github.com/weaveworks/mesh#Status) info would you like to see included as well?

api/api.go
```diff
@@ -180,14 +184,40 @@ func (api *API) status(w http.ResponseWriter, req *http.Request) {
 			"buildDate": version.BuildDate,
 			"goVersion": version.GoVersion,
 		},
-		Uptime: api.uptime,
+		Uptime:     api.uptime,
+		MeshStatus: getStrippedDownMeshStatus(api),
```
brancz: you can remove the `getStrippedDown` ... `meshStatus` is perfect :)

api/api.go
```diff
@@ -180,14 +184,40 @@ func (api *API) status(w http.ResponseWriter, req *http.Request) {
...
 	}
 
 	api.mtx.RUnlock()
 
 	respond(w, status)
 }
 
+type strippedDownMeshStatus struct {
```
brancz: same here: ~~strippedDown~~

api/api.go
```diff
@@ -180,14 +184,40 @@ func (api *API) status(w http.ResponseWriter, req *http.Request) {
...
+	Connections []strippedDownLocalConnectionStatus `json:"connections"`
+}
+
+type strippedDownLocalConnectionStatus struct {
```
brancz: same here: ~~strippedDown~~

api/api.go
```diff
@@ -180,14 +184,40 @@ func (api *API) status(w http.ResponseWriter, req *http.Request) {
...
+	Address string `json:"address"`
+	State   string `json:"state"`
```
brancz: Maybe describe in a comment what `Address` and `State` will look like. While I can imagine that `Address` will be an IPv4 address, an IPv6 address or a hostname I have no idea what `State` could be (maybe "healthy"?).

api/api.go
```diff
@@ -180,14 +184,40 @@ func (api *API) status(w http.ResponseWriter, req *http.Request) {
...
+}
+
+func getStrippedDownMeshStatus(api *API) strippedDownMeshStatus {
+	status := mesh.NewStatus(api.mrouter)
+	strippedStatus := strippedDownMeshStatus{
```
brancz: Technically same here ~~stripped~~, but I can see that that will cause collisions, generally speaking though you can keep the variable names much shorter. For example:

```go
s := mesh.NewStatus(api.mrouter)
status := meshStatus{
...
}
```

fabxc:
Are we sure we want to use the information from `Status.Connections`? It seems from the user perspective we'd mostly care about connected peers, i.e. other Alertmanager instances.
There's a separate slice for that in the status object.

https://godoc.org/github.com/weaveworks/mesh#PeerStatus

Connections in the mesh will grow in complexity as we have more connected peers. The number of peers will remain overviewable.

stuartnelson3:
Exposing that information as some form of data visualization might be really cool (and infinitely more grokkable).

I agree with @fabxc, as a user I would rather see which peers are in "my" mesh than know which ones a particular node is actively connected to.

brancz:
I was thinking those are probably the same, but looked a bit more into it and agree that a user is probably more interested in the peers themselves then the connections.

fabxc:
👍

PR #649, opened 2017-03-08
Add footer and author, author_link fields to Slack notifications

dewey:
Just adding some additional fields.


beorn7:
Looks good to me in general. @stuartnelson3 will have a closer look.

stuartnelson3:
Have you tested this? Looking at the code it appears that while these fields are being defined and set, additional fields need to be added to the `slackAttachment` struct and then set when sending the notification:

https://github.com/prometheus/alertmanager/blob/master/notify/impl.go#L500-L508

PR #655, opened 2017-03-15
Adding support for templated URL on the webhook notifier

sciffer:
I ran into the case where I wanted to allow people to control the url definition using labels, so that a person can define an alert with a label that specifies the URL to send the webhook request to.
It seems that form a very old version this option did not exists - unlike the to and from of the email notification - so I thought I'll add it myself and share.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -17,6 +17,7 @@ import (
 	"bytes"
 	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


fabxc:
Looks like you made your changes against an old version or something similar. The commit delta definitely shows more than your changes would have been.

sciffer:
That's odd, I was certain I'm on master and everything was up to date from
a fork I just made.

I'll fight with it for while till I figure out what's wrong.

Sorry,
Sciffer.

PR #656, opened 2017-03-16
Second take on adding support for dynamic url for webhook notification

sciffer:
Previous pull request for whatever reason originated from on older version, based on the delta I see now it should be correct this time.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -170,12 +170,16 @@ func (w *Webhook) Notify(ctx context.Context, alerts ...*types.Alert) (bool, err
 		GroupKey: uint64(groupKey),
 	}
 
+  var err error
+  tmpl := tmplText(w.tmpl, data, &err)
+  url := tmpl(w.URL)
+
 	var buf bytes.Buffer
 	if err := json.NewEncoder(&buf).Encode(msg); err != nil {
 		return false, err
 	}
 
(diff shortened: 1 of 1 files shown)


stuartnelson3:
Could you add a test to verify this works as expected?

sciffer:
I'll do that as soon as I'll get some time for it.
Btw, I also added a specific rundeck integration as part of alertmanager -
but I'm not certain this is relevant to alertmanager roadmap - so I have
not made a pull request out of it.

I'll be happy to know that is something you would like me to push as
well(it is also missing the tests, although it works on our production env).

brian-brazil:
It doesn't make sense to template this, the expectation is that whatever is on the other end is designed to handle the message format that the AM sends.

sciffer:
We had a requirement to provide a url parameter that will allow the system
to choose a specific workflow(based on alert label).
Now, in theory - you are correct and it is not necessary, but sometimes
workflow engines require you to choose the workflow (reality requires more
flexibility).

A second option, which is far less elegant, is to create multiple receivers
with statically specified workflow (and whenever you add or remove a
workflow, you need to add/remove receiver) - not ideal.

Bryan, What do you think?

 Thanks,
Shalom

brian-brazil:
It's expected that you'll handle this within whatever is receiving the webhook JSON message.

PR #661, opened 2017-03-21
Allow export of regex from label parser

Kellel:
Small addition to filtering

This is useful when trying to translate a label.Matcher to a type.Matcher (Case of using the parser to create new silences). 

Currently I would need to duplicate the regex logic and that feels unsavory.


Kellel:
Not 100% sure that I've made this change in the correct place. A little bit confused with the vendored code that doesn't seem to come from the prometheus repo itself.

stuartnelson3:
The vendored prometheus code is from the branch `dev-2.0`, sorry for the confusion!

I'll check out the main PR so I can get the context for this change.

Kellel:
Oh the pr makes sense now. Closing this in favor of putting the change in my main PR

PR #677, opened 2017-03-31
Refresh every 30 seconds (fixes #524).

### LINKED ISSUE #524: Alert page should automatically refresh
The Alerts page should auto refresh so that it can be kept open and automatically update as alerts fire and clear.

stuartnelson3:
More of a short spike than anything, this commit
refreshes whichever list page a user is on once
every 30 seconds.

Not sure if we want to think about this more at the moment. One of the main issues is that silences aren't received from the API in a specific order, so if the user is scrolling through the list, a refresh will cause them to lose their spot.

If this is something to pursue, I would force a sort order on incoming data from the API and then we can try this out.

### DIFF
--- a/ui/app/src/Main.elm
+++ b/ui/app/src/Main.elm
@@ -26,6 +26,7 @@ import Views.SilenceList.Updates
 import Views.SilenceForm.Types exposing (initSilenceForm)
 import Views.Status.Types exposing (StatusModel, initStatusModel)
 import Updates exposing (update)
+import Subscriptions exposing (subscriptions)
 
 
 main : Program Never Model Msg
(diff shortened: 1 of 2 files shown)


mxinden:
I like the idea but I would prefer to stick with the manual user refresh for now. It reminds me a lot of the Jenkins *auto-refresh* feature. I think we should first focus on the basic features (tests, ui-library, API v2 requirements document) instead of adding additional complexity.

In the long run: Requesting an ordered list from the API sounds like a great idea. With the Elm virtual DOM feature the user should only see an actual refresh whenever there are new silences - cool. Maybe we could even just notify the user that he is viewing outdated data at the moment.

Good to see the *subscriptions* feature in practice! I thought there is a lot more magic to it.

Would you be ok if we keep it as a PR for now and not merge it? What are your thoughts?

stuartnelson3:
Yeah totally happy to keep around as a reference, no worries. It was a
Friday afternoon quick hack, I hadn't ever done anything with subscriptions
either and wanted to check them out.

fabxc:
Agreeing that this is something to consider after the ground work is solid.

Issue #674, opened 2017-03-30
Click label matcher -> update filter


stuartnelson3:
The actual UX for this is up for discussion, but I'm currently thinking:

- Click on a label matcher, it gets added to the filter text box
- cmd/ctrl+click will de-select a label and remove it from the filter text box
- **Bonus**: selected labels have a different CSS class applied to somehow "highlight" it. At first it seems difficult to do this when adding in regex label matchers, but the keys should be unique and we can just check for the existence of that?

stuartnelson3:
Note: Moving the discussion from the PR to here

@brian-brazil @brancz @mxinden 

>...from the mock ups we did on the whiteboard I would expect one click to add a label to a query in some kind of a search bar. Then a label in the query would have an "x" button to remove it from the query.

Right now the search bar is just a text area that contains label matchers (`{foo="bar", baz="quux"}`), so there's no fancy grouping like in the fiddle you linked to.

In the event of buttons like in the fiddle, I would expect:
1. Clicking on a label (without ctrl or any modifier key) adds additional labels
2. Clicking the same label would not add anything
3. Typing out a label matcher `alertname="HighLatencyWarning"` in an input field and hitting enter would:
    - Add the matcher to the filter list
    - Clear the input field
    - Filter the view (not sure on this one yet -- if someone wants to add 4 matchrs before executing an API call this would be wasteful)
4. Clicking the `X` on the matcher would remove the filter

For clarity, a **matcher** is:
![image](https://cloud.githubusercontent.com/assets/1398104/24656940/b9372094-1944-11e7-99d5-b9c9c590c7b3.png)

And a **label** is what is specifically applied to an alert:
![image](https://cloud.githubusercontent.com/assets/1398104/24657126/689795dc-1945-11e7-9178-c6b891372772.png)

If the above bullet-pointed list sounds like a good UX, I would propose that we accomplish this in 2 PRs:

1. Add click-to-modify-filter, with the click semantics described above, 1-2 (to be implemented in #690)
2. Change the textbox to a single input that creates the matcher buttons, 3-4

brancz:
All of the above sounds good to me, just one detail that was not elaborated on: there should be a separate action to actually perform the search, not as soon as a label gets added via the described actions, as that can change the list view one was previously clicking a label from (as we wanted to be able to add labels from alerts previously loaded).

w0rm:
I implemented 1. and 2. in #690

stuartnelson3:
[...]

PR #690, opened 2017-04-03
Click alert matchers

stuartnelson3:
On alert list view, clicking a matcher sets the filter to that value. Holding `ctrl` or `cmd` will append to the current filter, or, if the label already exists, it will remove it. The UX motivation behind this is how users interact with a standard filesystem GUI -- clicking single files highlights that file, cmd+click'ing will select multiples.

Play around with it to make sure it's working correctly, and that this UX is what we want.

addresses #674


stuartnelson3:
@w0rm

mxinden:
I like the idea of multi-label-selection but I don't know any web application that uses *ctrl* to do multi-select. This seems rather not intuitive for a web application. Do we need the `replace` option? Can't we just make the `append` option default without needing to press the *ctrl* button?

ui/app/src/Updates.elm
```diff
@@ -96,6 +99,64 @@ update msg model =
...
+            let
+                mode =
+                    if List.member code modifierKeys then
+                        Replace
+                    else
+                        Append
+            in
+                ( { model | mode = mode }, Cmd.none )
+
+        AddLabel msg ( key, value ) ->
```
mxinden: This is only relevant for the `SilenceList` and the `AlertList` right? Would it be possible to move it down into the respective `update` functions? Feels weird that this lives in the global `update` function.

brian-brazil:
I've seen many web-apps use crtl for multi-select, it has been the standard for desktop apps for decades.

ui/app/src/Updates.elm
stuartnelson3: I wasn't sure where to put it. It's definitely something only concerned with the `SilenceList` and `AlertList` views, but it is updating the top-level `filter`.

All the code for `labels` should be extracted into its own filter method, so it wouldn't be a problem to have the two different update functions implement this.

Thinking about this now, it would simplify the message being sent, and then the `update` function in question would know what follow-up message to use (fetch alerts/silences). I'll give it a try and see how it looks.

stuartnelson3:
the ctrl+click felt natural to me, but that is just my personal feeling.

If the group decides to have a default `Append` behavior and clicking the same label twice removes it, that's fine too. I'm not sure what the current expected web behavior is for something like this, but having talked briefly with @beorn7 this seemed to make sense to us both.

brancz:
I don't have the possibility to spin this up myself right now, but from the mock ups we did on the whiteboard I would expect one click to add a label to a query in some kind of a search bar. Then a label in the query would have an "x" button to remove it from the query. Something like [this](http://jsfiddle.net/7zkCU/30/).

brancz:
I don't think

> I've seen many web-apps use crtl for multi-select, it has been the standard for desktop apps for decades.

Is an argument for it from a usability perspective. Just because it's been done for a long time, it doesn't mean it's good in terms of usability.

ui/app/src/Updates.elm
```diff
@@ -96,6 +99,64 @@ update msg model =
...
+                                |> Regex.replace Regex.All (Regex.regex "{|}") (\_ -> "")
+                                |> (\cleaned ->
+                                        if String.isEmpty cleaned then
+                                            []
+                                        else
+                                            Regex.split Regex.All (Regex.regex "\\s*,\\s*") cleaned
+                                   )
+                                |> (\labels ->
+                                        if List.member label labels then
+                                            List.filter (\l -> l /= label) labels
```
w0rm: `List.filter ((/=) label) labels`

ui/app/src/Views/AlertList/Views.elm
```diff
@@ -68,21 +68,38 @@ alertView alert =
...
         b =
             if alert.silenced then
-                buttonLink "fa-deaf" ("#/silences/" ++ id) "blue" Types.Noop
+                buttonLink "fa-deaf" ("#/silences/" ++ id) "blue" Noop
             else
                 buttonLink "fa-exclamation-triangle" "#/silences/new?keep=1" "dark-red" (CreateSilenceFromAlert alert)
+
+        labels =
+            List.filterMap
+                (\( k, v ) ->
```
w0rm: `filterMap` is only needed when you want to filter and map. In this case `filter` is enough.

`labels = List.filter (Tuple.first >> (/=) "alertname") alert.labels`

ui/app/src/Updates.elm
```diff
@@ -96,6 +99,64 @@ update msg model =
...
+                labels =
+                    case model.mode of
+                        Append ->
+                            Maybe.withDefault "" filter.text
+                                |> Regex.replace Regex.All (Regex.regex "{|}") (\_ -> "")
+                                |> (\cleaned ->
+                                        if String.isEmpty cleaned then
+                                            []
+                                        else
+                                            Regex.split Regex.All (Regex.regex "\\s*,\\s*") cleaned
```
w0rm: What will happen if you call `Regex.split` on the empty string? Wouldn't this result with an empty list?

stuartnelson3: I check in the other if branch if it is an empty string and return an empty list

w0rm: @stuartnelson3 why do you need this check? Won't `Regex.split` work on the empty string?

ui/app/main.go
```diff
@@ -37,15 +37,14 @@ func main() {
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 
-		const elmMake = "elm-make"
-		elmMakeArgs := []string{"src/Main.elm", "--output", "script.js"}
+		elmMakeArgs := []string{"make", "src/Main.elm", "--yes", "--output", "script.js"}
```
w0rm: In the future there will only be a single `elm` binary.

stuartnelson3: That's cool. In general we try to bundle relevant code changes together in a PR, so try to do that in the future. That having been said, this `main.go` file is totally unimportant, so no worries on changing it here.

w0rm: Sorry, I should've opened a separate PR for this change. If you want, I can still do this.

stuartnelson3: Not a problem, this file will get deleted soon (I hope)

ui/app/src/Views/AlertList/Views.elm
```diff
@@ -59,30 +58,35 @@ alertView : Alert -> Html Msg
...
 
         b =
             if alert.silenced then
-                buttonLink "fa-deaf" ("#/silences/" ++ id) "blue" Types.Noop
+                buttonLink "fa-deaf" ("#/silences/" ++ id) "blue" Noop
             else
                 buttonLink "fa-exclamation-triangle" "#/silences/new?keep=1" "dark-red" (CreateSilenceFromAlert alert)
+
+        labels =
+            List.filter (Tuple.first >> (/=) "alertname") alert.labels
```
stuartnelson3: I can read and understand what it's doing, but how is blowing my mind.

PR #691, opened 2017-04-04
test: Add elm-format job to travis ci configuration

mxinden:
This runs elm-format inside a docker container serving the Elm
environment. Afterwards we check with git if a file changed and fail
accordingly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,16 @@
 sudo: false
 language: go
 
+services:
+  - docker
+
 go:
 - 1.7.3
 
+before_install:
+- docker build -t elm-env ui/.
+
 script:
 - make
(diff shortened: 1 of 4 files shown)


mxinden:
Ready to merge if you don't have objections.

w0rm:
Maybe we should try to compile the project as well.

w0rm:
I commented that we should run `elm make` and see if that passes, but actually the `make` command was already there. I'm not sure if it calls `elm make` though.

mxinden:
@stuartnelson3 added `--validate` flag

@w0rm added `elm-make` job

@stuartnelson3 @w0rm Please rereview. 

Just as a heads up, make sure you update your local `elm-format` version to the newest `0.6.0-alpha` version. Otherwise the CI will fail on your changes.

PR #695, opened 2017-04-05
Spike: Use vanilla Bootstrap as ui framework

mxinden:
This is a quick spike testing out Bootstrap as our UI framework.

@stuartnelson3  @w0rm What do you think?

This is not focussed on design but rather technical usability.

### DIFF
--- a/ui/app/index.html
+++ b/ui/app/index.html
@@ -20,10 +20,4 @@
     <script type="text/javascript">
      var app = Elm.Main.embed(document.body);
     </script>
-
-
-    <!-- Bootstrap requirements -->
-    <script src="https://code.jquery.com/jquery-3.1.1.slim.min.js" integrity="sha384-A7FZj7v+d/sdmMqp/nOQwliLvUsJfDHW+k9Omg/a/EheAdgtzNs3hpfag6Ed950n" crossorigin="anonymous"></script>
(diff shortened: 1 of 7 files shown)


ui/app/index.html
```diff
@@ -1,21 +1,29 @@
...
-    var app = Elm.Main.embed(document.body);
-  </script>
+    <!-- Removes splash and starts elm app. -->
+    <script type="text/javascript">
+     var app = Elm.Main.embed(document.body);
+    </script>
 
+
+    <!-- Bootstrap requirements -->
+    <script src="https://code.jquery.com/jquery-3.1.1.slim.min.js" integrity="sha384-A7FZj7v+d/sdmMqp/nOQwliLvUsJfDHW+k9Omg/a/EheAdgtzNs3hpfag6Ed950n" crossorigin="anonymous"></script>
```
w0rm: We should only use css from bootstrap. Js won't work with Elm.

mxinden: Done

stuartnelson3:
Looks cool to me. 

Eventually we'll want to serve up bootstrap from alertmanager itself. We could even embed it (and script.js) in the index.html when we build it. But, that's for a different PR.

mxinden:
Cool. Waiting for @w0rm final words. Then I will merge the PR if there are no other objections.

PR #698, opened 2017-04-07
[EXPERIMENT] minimal grpc example

fabxc:
@mxinden @stuartnelson3 @w0rm this generates a REST API with Swagger docs. Can you try whether this would work to generate clients for our new frontend?

@brian-brazil as you had strong feelings about all this in the past, can you take a look?

In general the benefits I'm seeing here:
* We have a typed definition of our API and no longer have `implementation == specification`
* One definition giving us a gRPC and REST API that are always in sync
* clients across many languages can be generated for either – no need for us to ever write client libs

I realize that the performance gRPC provides may not matter all that much here. But the framework it provides seems pretty neat.


mxinden:
Awesome! I will have a look later today or over the weekend.

brian-brazil:
From my standpoint, we'll need to continue to offer a JSON interface for the forseeable future. We'll also need to keep all prometheus<->am traffic and am<->am traffic off HTTP 2.0 by default for a few years yet, to give time for enterprises to be able to handle it.

As long as we're doing that, I'm not too bothered about the rest of the details.

mattbostock:
Brian: I'm not sure I understand your comment regarding HTTP/2. Is there a
reason enterprises couldn't rely on protocol negotiation? If intermediary
proxies don't support HTTP/2, HTTP/1.1 will be used.

In other words, HTTP/1.1 is the always the default unless both client and
server are capable and willing to upgrade to HTTP/2.

brian-brazil:
My understanding is that there are hard dependencies on HTTP/2 in grpc, and content negotiation isn't an option.

fabxc:
I think Brian's point was merely that there need to be an HTTP1+JSON option on top of whatever more modern approach we provide. gRPC indeed only works with HTTP2 to my knowledge.

mattbostock:
Thanks - makes sense, I need to read up on gRPC.

FWIW I'm very much in favour of retaining support for a JSON-based API as
it's so widely understood and performs well enough for Alertmanager under
relatively heavy traffic.

I'd be interested to know the use  cases in favour of gRPC for
Alertmanager. It would be good to document these architectural decisions.

fabxc:
Yea, we all agree on gRPC not being viable as the only option. That's what this whole test PR is about.

mxinden:
After having investigated a little into Elm libraries to generate api clients there are two out there:

- [Swagger-Elm](https://github.com/ahultgren/swagger-elm)
  - Generate Elm api client from `api.swagger.json` file
  - Lacks support for dict / maps (See [issue](https://github.com/ahultgren/swagger-elm/issues/10))
- [Elm-protobuf](https://github.com/tiziano88/elm-protobuf)
  - Generate Elm api client right from `api.proto` file
  - Lacks support for maps (See [issue](https://github.com/tiziano88/elm-protobuf/issues/3))
  - Plans to add map support this month (See [comment](https://github.com/tiziano88/elm-protobuf/issues/3#issuecomment-293199297))

Might be a good investment to contribute these features to one of them.

stuartnelson3:
Swagger-Elm looks like a good idea to me

PR #699, opened 2017-04-07
Make asset generation deterministic

mxinden:
By passing "-modtime 1" flag into go-bindata command it uses the same
unix timestamp (1) for every file in the fake filesystem. Thereby every
execution of "make assets" results in the same asset outputs if the
inputs were the same. This forces us to set the "Cache-Control"
attribute to "no-cache".


brancz:
Agreed. If no caching does become an issue, then we can solve it properly.

PR #700, opened 2017-04-09
Add Telegram notifier

leominov:
@stuartnelson3, Telegram – popular messenger in some countries, currently has 100m users, 15b messages sends daily.

All test passed on my laptop.

Additional:

* https://telegram.org
* https://core.telegram.org/bots/api
* http://telegram.me/BotFather

### global

```
[ telegram_api_url: <string> | default = "https://api.telegram.org/" ]
[ telegram_token: <string> ]
```

### <telegram_config>

Telegram notifications are sent via the [Telegram Bot API](https://core.telegram.org/bots/api).

```
# Whether or not to notify about resolved alerts.
[ send_resolved: <boolean> | default = true ]

# The API key to use when talking to the Telegram Bot API, can be obtained from `@BotFather`.
[ token: <string> | default = global.telegram_token ]

# The chat id or user id to send notifications to.
chat_id: <string>

# The url to send Telegram Bot API requests to.
[ api_url: <string> | default = global.telegram_api_url ]

# Disables link previews for links in this message.
[ disable_web_page_preview: <boolean> | default = false ]

# Sends the message silently. iOS users will not receive a notification, Android users will receive a notification with no sound.
[ disable_notification: <boolean> | default = false ]

# Send `Markdown` or `HTML`, if you want Telegram apps to show bold, italic, fixed-width text or inline URLs in your bot's message.
# https://core.telegram.org/bots/api#formatting-options
[ parse_mode: <string> | default = 'HTML' ]

# Notification message.
[ text: <tmpl_string> | default = '{{ template "telegram.default.text" . }}' ]
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"encoding/json"
+
 	"github.com/prometheus/common/model"
 	"gopkg.in/yaml.v2"
 )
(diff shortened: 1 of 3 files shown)


stuartnelson3:
Thanks for the contribution!

When adding new receivers, we require tests as well as code. 

For this case though we are recommending that you implement this as a separate webhook service, as we aren't sure of the desire from the community to use this, and as far as delivery guarantees go we would also recommend something a bit more robust.

We are talking about a generic JSON templated receiver as a solution, which would solve this and many other service's requirements. We haven't made any firm plans yet, though.

leominov:
@stuartnelson3, Oh, ok. Where I can find design-docs to start helping with some required things?

stuartnelson3:
https://github.com/prometheus/alertmanager/issues/701

vtolstov:
any news about telegram support without external bot system?

stuartnelson3:
There are no plans as of now to add telegram support

PR #702, opened 2017-04-10
Include notifier type in retry logs and errors

juliusv:


michaeljs1990:
Thanks for publishing this. I tested it out on one of the servers having an issue with too many alerts coming in from prometheus and was able to get a more detailed output!

notify/notify.go
```diff
@@ -543,9 +543,9 @@ func (r RetryStage) Exec(ctx context.Context, alerts ...*types.Alert) (context.C
 		case <-tick.C:
 			if retry, err := r.integration.Notify(ctx, alerts...); err != nil {
 				numFailedNotifications.WithLabelValues(r.integration.name).Inc()
-				log.Debugf("Notify attempt %d failed: %s", i, err)
+				log.Debugf("Notify attempt %d for %q failed: %s", i, r.integration.name, err)
```
brancz: Not sure this debug logging actually has any value as we ultimately always [log errors](https://github.com/prometheus/alertmanager/blob/7f1d11132473ace30b23b5cb4d04b742c2149654/dispatch/dispatch.go#L263-L266) happening in the notification pipeline, but I also don't mind keeping it as it's closer to the source here without many moving parts where it may get lost.

brancz:
lgtm :+1: , your call whether we want to keep the debug logging or not

fabxc:
👍

Issue #598, opened 2017-01-06
Incorrect needsUpdate in DedupStage for notifications results in spammy emails


ddewaele:
Alertmanager currently sends out notifications when it thinks the alert content has changed within a group and within the group_interval. It does so by comparing the hash of a group of alerts (firing and resolved alerts).

In my test case using 

```
scrape_interval: 5s
group_wait: 30s
group_interval: 2m
repeat_interval: 24h
```

and reproducible via

```
curl -XPOST -d'[{"labels":{"alertname":"a1"}}]' http://localhost:9093/alertmanager/api/v1/alerts; date ; sleep 15
curl -XPOST -d'[{"labels":{"alertname":"a2"}}]' http://localhost:9093/alertmanager/api/v1/alerts; date ; sleep 15
curl -XPOST -d'[{"labels":{"alertname":"a3"}}]' http://localhost:9093/alertmanager/api/v1/alerts; date ; sleep 15
curl -XPOST -d'[{"labels":{"alertname":"b1"}}]' http://localhost:9093/alertmanager/api/v1/alerts; date ; sleep 15
curl -XPOST -d'[{"labels":{"alertname":"b2"}}]' http://localhost:9093/alertmanager/api/v1/alerts; date ; sleep 15
curl -XPOST -d'[{"labels":{"alertname":"b3"}}]' http://localhost:9093/alertmanager/api/v1/alerts; date ; echo "done... now looping b3 for keepAlive"
while true; do  curl -XPOST -d'[{"labels":{"alertname":"b3"}}]' http://localhost:9093/alertmanager/api/v1/alerts; date  ; sleep 15; done
echo "done"
```

It is causing 4 notifications (where I think it should only be sending 2).

This comes from the _needsUpdate_ function in notify.go :
```
	if !bytes.Equal(entry.GroupHash, hash) {
		return true, nil
	}
```

I guess the hash will not be sufficient to determine if an alert should be sent.

If all alerts (including the resolved ones) within a group are hashed it could explain the 4 notifications:

My results : 
**Notification 1** 
```
 a1[7a091fb][active]
```
**Notification 2**
```
a2[4ad4244][active]
a3[7732e17][active]
b1[99488d8][active]
b2[4a429ad][active]
b3[3a21090][active]
a1[7a091fb][resolved]
```
**Notification 3**
```
b3[3a21090][active]
a2[4ad4244][resolved]
a3[7732e17][resolved]
b1[99488d8][resolved]
[...]

PR #703, opened 2017-04-11
Fix resolve notifications

### LINKED ISSUE #611: Alertmanager sends the resolved messages continuously
This problem is totally same as this closed issue: https://github.com/prometheus/alertmanager/issues/246

I changed the `resolve_timeout` parameter as you suggested but it did not help.

brancz:
This PR introduces a fundamental change to the protobufs gossiped, because the previously optimization we implemented for the 0.5 release did not turn out to work as expected. Instead we now gossip the entire list of previously firing and resolved alerts, so if the currently firing alerts are a subset of the alerts that were firing at the last notification, then we can deduplicate. It works the same way for resolved notifications, if the currently resolved alerts are a subset of those that have already been notified about then they are deduplcated. Consequently, if currently firing or resolved alerts are not a subset of the previous sets, then they need to be notified about (after the `group_interval`). This means the payload that is gossiped increases significantly, however, I decided against a premature optimization and rather benchmark this implementation if a load problem actually comes up.

I wanted to get this out for people to test, who have been struggling with this issue for a while. As this is quite a fundamental change and touches most of the critical parts of the notification pipeline I would like to test this a bit before having it ready to merge. /cc @ddewaele @tsboj17 @mattbostock

Fixes #611 and #598 

@fabxc @stuartnelson3 @mxinden


notify/notify.go
```diff
@@ -382,55 +382,47 @@ func (ws *WaitStage) Exec(ctx context.Context, alerts ...*types.Alert) (context.
...
-		binary.BigEndian.PutUint64(b, uint64(a.Fingerprint()))
-		// Resolved status is part of the identity.
-		if a.Resolved() {
-			b[8] = 1
-		}
-		for i, b := range sha256.Sum256(b) {
-			xsum[i] ^= b
-		}
+func hashAlert(a *types.Alert) uint64 {
+	b := make([]byte, 0, 1024)
```
fabxc: Can we get that one from a sync.Pool? Prometheus doesn't have it because mostly it does shortcut around the hash. (We probably still want to add it there eventually as well though as the QE hashes a lot.)

notify/notify.go
```diff
@@ -442,14 +434,18 @@ func allAlertsResolved(alerts []*types.Alert) bool {
 	return true
 }
 
-func (n *DedupStage) needsUpdate(entry *nflogpb.Entry, hash []byte, resolved bool, repeat time.Duration) (bool, error) {
+func (n *DedupStage) needsUpdate(entry *nflogpb.Entry, firing, resolved map[uint64]struct{}, repeat time.Duration) (bool, error) {
 	// If we haven't notified about the alert group before, notify right away
 	// unless we only have resolved alerts.
 	if entry == nil {
-		return !resolved, nil
+		return ((len(firing) > 0) || (n.sendResolved && len(resolved) > 0)), nil
```
fabxc: Wondering here. I remember providing explicit logic about not sending resolved notifications about stuff we have never notified about being firing in the first place.

But maybe just keep as it is here – might be better in doubt in case some other AM notified something as firing but couldn't propagate the entry via the mesh and then died.

notify/notify.go
```diff
@@ -476,10 +472,16 @@ func (n *DedupStage) Exec(ctx context.Context, alerts ...*types.Alert) (context.
 		return ctx, nil, fmt.Errorf("repeat interval missing")
 	}
 
-	hash := n.hash(alerts)
-	resolved := n.resolved(alerts)
+	firing := map[uint64]struct{}{}
+	resolved := map[uint64]struct{}{}
 
-	ctx = WithNotificationHash(ctx, hash)
```
fabxc: This context entry is completely gone now. The related function and const should be removed as well.

But we are redoing the whole hashing we have done before. Wouldn't it make sense to pass the hashes through the context to the SetNotifies stage?

notify/notify_test.go
```diff
@@ -91,44 +86,44 @@ func TestDedupStageNeedsUpdate(t *testing.T) {
...
-			res:   true,
+			entry:        nil,
+			firingAlerts: map[uint64]struct{}{2: struct{}{}, 3: struct{}{}, 4: struct{}{}},
+			res:          true,
 		}, {
-			entry: &nflogpb.Entry{GroupHash: []byte{1, 2, 3}},
-			hash:  []byte{2, 3, 4},
-			res:   true,
+			entry:        &nflogpb.Entry{FiringAlerts: []uint64{1, 2, 3}},
+			firingAlerts: map[uint64]struct{}{2: struct{}{}, 3: struct{}{}, 4: struct{}{}},
```
fabxc: Add a small constructor function (inlined in the test function or elsewhere) to construct those sets? Probably cannot hurt overall to make a `alertHashSet` helper type hiding the awkwardness of `map[uint64]struct{}`.

notify/notify_test.go
```diff
@@ -181,35 +184,41 @@ func TestDedupStage(t *testing.T) {
...
 	ctx, res, err = s.Exec(ctx, alerts...)
 	require.Contains(t, err.Error(), "result size")
 
 	// Must return no error and no alerts no need to update.
+	i = 0
 	s.nflog = &testNflog{
 		qerr: nflog.ErrNotFound,
+		qres: []*nflogpb.Entry{
+			{
+				FiringAlerts: []uint64{uint64(0), uint64(1), uint64(2)},
```
fabxc: The explicit `uint64()`s here should not be necessary.

nflog/nflogpb/set.go
```diff
@@ -0,0 +1,30 @@
+package nflogpb
```
brancz: Only reason I'm making this a separate file is so it doesn't get overridden when regenerating from the proto.

notify/notify.go
```diff
@@ -382,55 +382,47 @@ func (ws *WaitStage) Exec(ctx context.Context, alerts ...*types.Alert) (context.
...
+
+	for labelName, _ := range a.Labels {
+		labelNames = append(labelNames, labelName)
 	}
-	return xsum[:]
+	sort.Sort(labelNames)
+
+	for _, labelName := range labelNames {
+		b = append(b, string(labelName)...)
+		b = append(b, sep)
```
brancz: done

brancz:
Applied all your comments @fabxc. Care for a re-review? :slightly_smiling_face:

fabxc:
Currently pondering with releasing it now or waiting a bit more to try to get in a first version of the new UI. I suppose you prefer the fix sooner?

In general, nothing speaks against two releases in short proximity.

stuartnelson3:
I vote for getting the fix out, and then we just make another release when the UI is ready. Last release was back in November, and there have been many additions since then.

mattbostock:
@fabxc Sooner rather than later would be appreciated as this bug has the potential to create a lot of noise.

#717 (solving #609) is also an important fix, though I think the current implementation may preclude backporting it.

mattbostock:
Looks like this is included in 0.6.0 though not mentioned in the release description or the CHANGELOG:
https://github.com/prometheus/alertmanager/releases/tag/v0.6.0

Thanks!

damomurf:
Is it actually included in the 0.6.0 release? I'm seeing an infinitely firing "resolved" alert on 0.6.1?

fabxc:
@damomurf can you describe the scenario more precisely? Those things are hard to debug without a detailed timeline of when alerts were flowing in and what the exact notification behavior was.

A reproducible scenario is ideal, of course.

damomurf:
Sure @fabxc, I appreciate it's difficult to solve without being able to reproduce.

So currently, I have an "alert resolved" message being fired to Slack every three hours. The actual alert was resolved days ago and if I visit current alerts in either AlertManager or Prometheus, they're both empty.

`[RESOLVED] ElasticSearchClusterStatus account (/monitoring/es-exporter hostnane:port marathon-discovery-aws mesos-monitor)`

fabxc:
I assume 3 hours is your configured repeat interval. Restarting the AMs should solve it (but is not a viable solution obviously), will try to reproduce it later but it sounds like some edge case that went wrong and it could be hard to hit it manually again.

damomurf:
That's correct, 3 hours is the timeout. This is a single instance of Alertmanager, if that helps?

PR #704, opened 2017-04-11
Changes to send mail via smtp port 465

IamPrvn:
AlertManager currently is not able to send mails to smtp port 465. Adding this change that will take care of the above. In alertmanager conf file, you need to explicitly supply "require_tls" to false, in order to work with smtp port 465 

Refer issue #705 

@stuartnelson3  @beorn7

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -254,18 +254,31 @@ func (n *Email) auth(mechs string) (smtp.Auth, error) {
 
 // Notify implements the Notifier interface.
 func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
-	// Connect to the SMTP smarthost.
-	c, err := smtp.Dial(n.conf.Smarthost)
-	if err != nil {
-		return true, err
-	}
-	defer c.Quit()
-
 	// We need to know the hostname for both auth and TLS.
-	host, _, err := net.SplitHostPort(n.conf.Smarthost)
+	var c *smtp.Client
+	host, port, err := net.SplitHostPort(n.conf.Smarthost)
 	if err != nil {
 		return false, fmt.Errorf("invalid address: %s", err)
 	}
+
+	if port == "465" {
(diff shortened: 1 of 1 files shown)


IamPrvn:
@stuartnelson3 @beorn7 is there any update on this?

beorn7:
@stuartnelson3 Could you have a look? I have not much state on this.

stuartnelson3:
Thanks for the pr! Could you run your code through `goimports` and `gofmt`? Seems like some formatting is off.

IamPrvn:
Fixed formatting issues in the previous commit.

notify/impl.go
```diff
@@ -254,19 +254,32 @@ func (n *Email) auth(mechs string) (smtp.Auth, error) {
...
+	var c *smtp.Client
+	host, port, err := net.SplitHostPort(n.conf.Smarthost)
 	if err != nil {
 		return false, fmt.Errorf("invalid address: %s", err)
 	}
 
+	if port == "465" {
+		conn, err := tls.Dial("tcp", n.conf.Smarthost, &tls.Config{ServerName: host})
+		if err != nil {
+			return true, err
```
stuartnelson3: `true` should be `false` in all these error blocks

IamPrvn: AFAIK, true/false flag is tied to retry logic (whether the error is recoverable or not). For the communications happening over wire, multiple attempts shall be made(i.e. error shall be deemed to be recoverable, in this case we are trying to establish a connection to remote smtp server). Infact, even before my change most of the error blocks returned true.

stuartnelson3: Ah sorry, I should have looked further. Disregard.

IamPrvn: @fabxc @juliusv @beorn7 any updates?

beorn7:
I haven't worked with SMTP servers in Go either. @stuartnelson3 I'd say if you like it and nobody objects, go for it.

Issue #528, opened 2016-10-06
external location: alertmanager uses external locations for internal listener


kayrus:
For example prometheus configured with the `-alertmanager.url=http://alertmanager:9093`  option.

And alertmanager has `-web.external-url=https://my-external-ingress-controller.example.com/alertmanager` option.

For some reason alertmanager uses `/alertmanager` location even for internal listeners and as a result you receive `time="2016-10-06T11:20:37Z" level=error msg="Error sending alerts: bad response status 404 Not Found" alertmanager="http://alertmanager:9093" count=3 source="notifier.go:308"` error message in prometheus.

External and internal configuration options should be independent.

brian-brazil:
I believe this behaviour is correct. You told the alertmanager it'd only be accessed via a prefix, and then did not access it via that prefix.

kayrus:
@brian-brazil that's not correct. my prometheus ingress location has `my-external-ingress-controller.example.com` domain name. And this location serves several services including alertmanager. And when you define `external-url` you define _external_ url, not internal.

brian-brazil:
The URL provided is what all external access to the AM is via. There is no internal AM http.

groob:
I ran into this issue earlier today and wanted to document my usecase: 

I have prometheus and alertmanager deployed on a k8s cluster which is publicly accessible(not behind VPM/firewall). We use an ingress controller which
a) terminates TLS and 
b) optionally forces OAuth authentication for users  to access various dashboards, internal pages etc. 
So far it hasn't been a problem as most applications either have some form of robust auth on their own, or allow direct access to the backend unsecured(in which case we proxy it through oauth). 

For the setup to work, I need the  `-web.external-url` flag because I want the users to get the correct URL in their alerts, but I do not want to expose the alertmanager API directly to the public. Unfortunately, there's no way of pointing prometheus to a external URL which will force the request to go through a custom authentication proxy. 

What would be a good approach to resolve this issue? I'd be glad to submit a PR for separating the API and external URL if needed.

brian-brazil:
> Unfortunately, there's no way of pointing prometheus to a external URL which will force the request to go through a custom authentication proxy.

That's configurable on the Prometheus side.

mrwacky42:
Related...
[...]

PR #706, opened 2017-04-12
Add web.alert-url parameter so we can run behind reverse-proxy

alxmk:
In our deployment of prometheus we make the dashboards available through a reverse-proxy, this means that the alerts that come up in Slack or PagerDuty need to contain a different URL from the one that Alertmanager runs on.

Setting `web.external-url` to this externally visible URL doesn't work (see #528).

If the new `web.alert-url` parameter isn't set we default to the `web.external-url` value, so this change should be transparent.

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -91,6 +91,10 @@ func main() {
 	flag.Var(peers, "mesh.peer", "initial peers (may be repeated)")
 	flag.Parse()
 
+	if *alertURL == "" {
+		alertURL = externalURL
+	}
+
 	if *hwaddr == "" {
 		*hwaddr = mustHardwareAddr()
 	}
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -80,6 +80,7 @@ func main() {
 		retention  = flag.Duration("data.retention", 5*24*time.Hour, "How long to keep data for.")
 
 		externalURL   = flag.String("web.external-url", "", "The URL under which Alertmanager is externally reachable (for example, if Alertmanager is served via a reverse proxy). Used for generating relative and absolute links back to Alertmanager itself. If the URL has a path portion, it will be used to prefix all HTTP endpoints served by Alertmanager. If omitted, relevant URL components will be derived automatically.")
+		alertURL      = flag.String("web.alert-url", "", "The URL which is prefixed to the Alertmanager alert URLs, for example when accessing via a reverse proxy.")
 		listenAddress = flag.String("web.listen-address", ":9093", "Address to listen on for the web interface and API.")
 
 		meshListen = flag.String("mesh.listen-address", net.JoinHostPort("0.0.0.0", strconv.Itoa(mesh.Port)), "mesh listen address")
@@ -222,7 +223,11 @@ func main() {
 		if err != nil {
 			return err
 		}
-		tmpl.ExternalURL = amURL
+		alertingURL, err := extURL(*listenAddress, *alertURL)
+		if err != nil {
+			log.Fatal(err)
+		}
+		tmpl.ExternalURL = alertingURL
 
 		inhibitor.Stop()
 		disp.Stop()


brian-brazil:
As I asked in #528, can you explain why web.external-url isn't sufficient here? It appears to perfectly meet your needs.

alxmk:
@brian-brazil We don't have a route from Prometheus to Alertmanager via the reverse proxy, so we need Alertmanager to still be listening on its local address.

e.g. Alertmanager is on `server-01:9093`, reverse proxied as `monitoring.example.com`. 

Prometheus can't see that URL because it's private networked only so it needs to be configured with `-alertmanager.url=http://server-01:9093`.

If we set `-web.external-url=monitoring.example.com` then Prometheus can't talk to it any more as  it doesn't respond on `server-01:9093`.

brian-brazil:
Why wouldn't the AM respond on server-01:9093? It's still bound to that port.

alxmk:
```
alex@alex-VirtualBox:~/Workspaces/go/src/github.com/prometheus/alertmanager$ sudo docker run -v /tmp/alertmanager.yml:/etc/alertmanager/config.yml --net=host prom/alertmanager -config.file=/etc/alertmanager/config.yml -web.external-url=http://monitoring.example.com/some/path
time="2017-04-12T13:49:27Z" level=info msg="Starting alertmanager (version=0.5.1, branch=master, revision=0ea1cac51e6a620ec09d053f0484b97932b5c902)" source="main.go:101" 
time="2017-04-12T13:49:27Z" level=info msg="Build context (go=go1.7.3, user=root@fb407787b8bf, date=20161125-08:14:40)" source="main.go:102" 
time="2017-04-12T13:49:27Z" level=info msg="Loading configuration file" file="/etc/alertmanager/config.yml" source="main.go:195" 
time="2017-04-12T13:49:27Z" level=info msg="Listening on :9093" source="main.go:250" 
```

```
alex@alex-VirtualBox:~/Workspaces/go/src/github.com/prometheus/alertmanager$ curl localhost:9093
404 page not found
```

```
alex@alex-VirtualBox:~/Workspaces/go/src/github.com/prometheus/alertmanager$ curl http://localhost:9093/some/path
<a href="/some/path/">Moved Permanently</a>.
```

It looks like the problem happens around this line: https://github.com/prometheus/alertmanager/blob/1e34f29532a251b728ddf304ebccd9234fcbd81a/cmd/alertmanager/main.go#L255

Since we're accessing Alertmanager as part of a dashboard, we have some path after the host, we don't want that additional path on the local Alertmanager routing.

Apologies for the slightly shoddy explanation above, I forgot the path element was what was causing us the problem :)

sstarcher:
Using `-web.external-url` the alertmanager is correctly alerting with proper links that I can follow through my reverse proxy.  Prometheus is also able to correctly talk to my alertmanager and I can curl it perfectly fine from within the network using either localhost or over the local internal network.

alxmk:
@sstarcher As above, it works fine using a domain root e.g. `http://example.com`, but with additional path e.g. `http://example.com/dashboard/prod it does not, the path is moved to `http://localhost:9093/dashboard/prod` which is not the behaviour we want for our use case.

discordianfish:
I have the same problem: Setting `external-url` to something which includes a path, will make the AM only respond on that path.
So if you set that to http:/example.com/foo, the AM sets up handlers at /foo. Now if the reverse proxy strips the path (which is common), the AM sees a request to / and returns 404. I think this is essentially the same as https://github.com/prometheus/prometheus/issues/1191 but for the AM.

brancz:
Actually starting with the 0.7.0 release the Alertmanager has the `-web.route-prefix` flag in addition to the `-web.external-url` just like Prometheus, to be able to control the prefix of the internal http handlers and how Alertmanager is accessed.

brancz:
Given that those two are now decoupled, I think this PR is actually not necessary anymore.

discordianfish:
Oh I missed that, that's great. Yes I think then this PR can be closed.

Issue #157, opened 2015-11-24
Add sns notification support back


ecasti-gmail-com:
Hi all,
i'm trying to send notification to a SNS topic from alertmanager for trigger autoscaling of ECS,
i tried successfully slack notification, 
but when i try to enable SNS notification the alertmanager panic with this error :
time="2015-11-18T15:20:11Z" level=info msg="Sent 1 notifications" file=manager.go line=353 
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x20 pc=0x6b51d2]

goroutine 1 [running]:
github.com/aws/aws-sdk-go/service/sns.New(0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
        /app/.build/gopath/src/github.com/aws/aws-sdk-go/service/sns/service.go:53 +0x72
github.com/prometheus/alertmanager/manager.(_notifier).sendAmazonSnsNotification(0xc820194ae0, 0x1, 0xc820195560, 0xc82017c580, 0x0, 0x0)
        /app/.build/gopath/src/github.com/prometheus/alertmanager/manager/notifier.go:464 +0x55
github.com/prometheus/alertmanager/manager.(_notifier).handleNotification(0xc820194ae0, 0xc82017c580, 0x1, 0xc8200d9500)
        /app/.build/gopath/src/github.com/prometheus/alertmanager/manager/notifier.go:809 +0x853
github.com/prometheus/alertmanager/manager.(*notifier).Dispatch(0xc820194ae0)
        /app/.build/gopath/src/github.com/prometheus/alertmanager/manager/notifier.go:817 +0x86
main.main()
        /app/main.go:159 +0x1280

goroutine 17 [syscall, 1 minutes, locked to thread]:
runtime.goexit()
        /usr/local/go/src/runtime/asm_amd64.s:1696 +0x1

i"m using Docker images with this release:

branch        master
date        20151109-20:48:36
go_version        1.5.1
revision        86f5f80
user        @afbbf75ae75d
version        0.0.4

and alertmanager.conf is this:

notification_config {
  name: "alertmanager_test"
  amazon_sns_config {
    topic_arn: "arn:aws:sns:us-east-1:1234567890X:MyTopicName"
    send_resolved: true
  }
}

aggregation_rule {
  repeat_rate_seconds: 300
  notification_config_name: "alertmanager_test"
}

[...]

PR #391, opened 2016-06-20
Add AWS SNS Receiver

ericgerling:

PR #709, opened 2017-04-14
Reintroduce AWS SNS notifications

tyrken:
As requested in #157, though it was a bit more than a one liner after the re-write.

Requires AWS_REGION environment variable to be set to work currently, but wondering about making that a global setting (maybe also a default Topic_ARN) or default to the one in the ARN, not sure.

I see your comments about testing in #700, but as not an expert golang dev without a similar pattern from any existing notifier I'm at a loss where to start.  The only significant logic is in AmazonSns::Notify which if I could fake up the inputs could be called with a bad AWS_REGION to always fail, but that's more integration test than unit.


tyrken:
Damn, only just found #391 despite earlier searching - annoying.  Will review the feedback to that one & implement here (e.g. templating ARN also, which I was wondering about).

tyrken:
OK, looks ready to review to me - @stuartnelson3 do the tests included look sufficient?

tyrken:
Bump for @stuartnelson3 @fabxc - now you've got the release out the way, could I have some feedback on this?

Will rebase to pick up the latest master before final submission, obviously, but presume you have some comments first...

stuartnelson3:
Sorry for not getting to this! I'll give it a look

stuartnelson3:
I'm not going to have time to get to this for the next week at least -- I'll be on vacation for a bit. Maybe @fabxc or @brancz have some time to check this out. Sorry again for the delay!

tyrken:
*ping* - any chance of this being looked at yet?  And yes, I will rebase if there's any interest.

I was hoping to go on & produce a similar option to get AlertManager to periodically feed alert status to an AWS CloudWatch Metric, so I could have alarm associated that our enterprise monitoring system would see.  But if this relatively simple PR takes months to start review...

mxinden:
@tyrken I am sorry that this took so long. This looks good to me, but I don't feel confident approving this. I will make sure @fabxc or @brancz will have a look at it.

brancz:
The notification providers integrated into the Alertmanagers are meant to immediately notify someone. Enqueuing it in some other mechanism sort of defeats the purpose and additionally makes ensuring that your alerting pipeline works properly significantly more complicated if not impossible as you're inserting yet another fragile distributed system into the equation. I would suggest implementing this support via a webhook.

tyrken:
@brancz - respectfully I disagree.  SNS is being used to notify a system as our 'Enterprise alerting systems' which I'm required to use only practically accept input via SNS or CloudWatch Alarms.  I hope to trigger them as well as the Slack & other channels we will likely use in the local team.

You originally had SNS functionality but wrote it out during a refactor a year or so ago - several other people have commented on #157 and #391 wanting it back.

AFAIK you cannot trigger SNS (or CW Alarms) via a webhook directly, only by chaining together other AWS bits like API Gateway/Lambdas ... at which point it does become a fragile distributed system.

fabxc:
This is a discussion we have quite repeatedly. People have different requirements and expectations when it comes to their notifications and literally any system capable of delivering messages in one way or the other could be an integration.

Just considering how many pub/sub or message buses there are, it's obvious that we can neither provide all those, nor accept them all into our main code base. We've made our experiences with that for Prometheus service discovery integrations.  
There unfortunately no correlation between number of people asking for a feature and number of people willing to maintain it. If we change something in our code base, ultimately we have to adjust 20+ integrations that we don't know anything about. More often then not something breaks and no one to help fixing it is in sight.

The webhook format is not intended to be sent to arbitrary integrations directly but to run a small web server that provides integration with arbitrary systems.

I think the reasonable way here is to provide better examples or maybe a small framework to build such a webhook bridge.

I'm sorry that you put the effort to implement this direct integration, but I hope our reservations are understandable. They are not arbitrary but come from consistent experiences.
Your code should be easily extractable into a webhook bridge though. So definitely no need to throw it away.

tyrken:
Not the answer I was hoping for after 3 months, especially as you'd already implemented it once, the earlier PR merely wanted tests & wasn't actively rejected, but it's your project & I do recognise the issue.

How about you define a **plugin interface** (golang 1.8+) & allow those with differing opinions to bear their own burdens, yet be able to share them with the community in a way 100 user-specific webhook implementations never will be?

If this is at all possible then you or I can start a new issue to discuss, hopefully with a much quicker resolution.  Feel free to chuck out all the other integrations that break into the plugin in the future!

fabxc:
Personally I somewhat like the idea, but others feel differently. Practically it probably doesn't not make a difference. Whether we have 100 user specific plugins or webhooks ends up being roughly the same. Just that webhooks need a bit more initial deployment work.
The main question is where the code lives. And if those plugins live anywhere in our realm of responsibility, we as maintainers have the same problem as before.

Plugins also have remaining issues: https://github.com/golang/go/issues/19233
This would quite literally mean a plugin model would only work if we were the canonical source of distribution for them.

tyrken:
FYI the plugin issue claims to have been fixed in 1.9.2 so please reconsider allowing externally provided/maintained plugins.  Closing this PR as unacceptable.

kevinayres:
I know this is an old thread but https://github.com/DataReply/alertmanager-sns-forwarder 
This project appears to not have been updated in a couple of years. I haven't gotten it working yet.  It seems like one possible approach.

roidelapluie:
SNS will soon be supported natively: #2559

kevinayres:
> SNS will soon be supported natively: #2559

Julien, this is superb. Any idea of the timeline or when it will 'work' if not fully supported by the project yet? Thanks!

roidelapluie:
> > SNS will soon be supported natively: #2559
> 
> Julien, this is superb. Any idea of the timeline or when it will 'work' if not fully supported by the project yet? Thanks!

I have no insight of the planning of other contributors. I suggest you to bookmark the pull request page, I expect that you could have something to test in the coming weeks/months.


I guess users feedback is also welcome in the comments of #2559.

PR #711, opened 2017-04-14
Issue 609 2

prymitive:
Second attempt at solving #609, previous PR was #692 

This is still work in progress, changes here are complete but:
* [proper cleanup](https://github.com/prometheus/alertmanager/pull/692#issuecomment-293860372) still needs to be implemented
* naming could probably be improved (alert->status->status isn't perfect, and maybe use `new` instead of `unprocessed`?)


types/types.go
```diff
@@ -23,66 +24,160 @@ import (
...
+type state uint8
+
+const (
+	unprocessed state = iota
+	active
+	silenced
+	inhibited
+)
+
+// allows tracking of the status of an alert (active, silenced, )
```
stuartnelson3: Please follow go style for comments on exported package members https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences

stuartnelson3:
I finally got to looking at your changes and really like the direction. I hacked around on it a bit, see what you think. I opened a pr, but I've also attached a patch file that you can check out.

[711.txt](https://github.com/prometheus/alertmanager/files/930509/711.txt)

prymitive:
Great, I'll close this one as #717 is more complete

PR #714, opened 2017-04-16
Improve filter

### LINKED ISSUE #674: Click label matcher -> update filter

w0rm:
filter

This closes #674.

I implemented 3 and 4 from this list: https://github.com/prometheus/alertmanager/issues/674#issuecomment-291489942

What I did:

1. Introduced parsers for the filter text and the matcher text
2. Introduced a new `Matcher` type in the `Utils.Filter` module, because the old one had only supported two matching operators (`=` and `=~`).
3. Introduced AlertList model that keeps a list of matchers, alert groups and the current filter text
4. Moved the top level filtering messages into the `AlertList` component (`UpdateFilter` is left there because it is still used by the `SilenceList`)


w0rm:
backspace

Added a nice feature, pressing backspace in the empty field now erases the previous matcher.

mxinden:
@w0rm Re-triggered the Circle-CI tests. I think they are flaky from time to time.

w0rm:
@mxinden sorry, I just replaced the last commit with a better solution :)

w0rm:
@mxinden cool idea with editing, I'm very much fond of it! In fact, it is very easy to add it.

editing

ui/app/src/Views/AlertList/FilterBar.elm
```diff
@@ -1,20 +1,137 @@
...
+
+                    Nothing ->
+                        "has-danger"
+
+        maybeMatcher =
+            Utils.Filter.parseMatcher matcherText
+
+        onKeydown =
+            onKey "keydown" 8 <|
+                case ( matcherText, backspacePressed ) of
```
mxinden: @w0rm I am very confused around the entire backspacePressed idea. Don't quite get the code. Why do we need to save this in the state of the AlertList model. Why can't we let a backspacePressed event trigger a message which then removes the last element of all matchers? Wouldn't that be easier?

ui/app/src/Views/AlertList/Updates.elm
```diff
@@ -1,22 +1,74 @@
...
+                        else
+                            model.matchers ++ [ matcher ]
+                    , matcherText =
+                        if emptyMatcherText then
+                            ""
+                        else
+                            model.matcherText
+                }
+
+        DeleteFilterMatcher setMatcherText matcher ->
```
mxinden: @w0rm Is it possible to trigger two messages at once? If so, could we make `setMatcherText` a separate message? Then `DeleteFilterMatcher` would not need to care of this peace of logic. In my eyes this would separate concerns better?! Or instead of containing the new matcher text into the `DeleteFilterMatcher` message we could just trigger a `UpdateMatcherText` with the stringified matcher?

ui/app/src/Views/AlertList/Views.elm
```diff
@@ -1,44 +1,51 @@
...
+import Views.AlertList.Types exposing (AlertListMsg(..), Model)
+import Types exposing (Msg(Noop, CreateSilenceFromAlert, MsgForAlertList))
+
+
+view : Model -> Filter -> Html Msg
+view { alertGroups, matchers, matcherText, backspacePressed } filter =
+    div []
+        [ Views.AlertList.FilterBar.view matchers matcherText backspacePressed
+        , case alertGroups of
+            Success groups ->
```
mxinden: I very much like moving this peace of logic from the main view function into the AlertList view function 👍

ui/app/src/Views/AlertList/FilterBar.elm
stuartnelson3: We had a look at it together today. Holding down the backspace removes (one by one) each last character in the input, and the whole time sends multiple keyDown events. This is a guard so that if a user holds down backspace to remove the text in the input, they won't accidentally hold backspace too long and then delete the preceding matcher as well. So, once a user holds backspace to clear an input, they have to then lift up the key and press it again to proceed to deleting the next matcher. I think this makes sense to prevent users from deleting all their matchers at once accidentally.

ui/app/src/Views/AlertList/FilterBar.elm
```diff
@@ -1,20 +1,137 @@
...
+                            |> Maybe.withDefault Noop
+
+                    _ ->
+                        PressingBackspace True |> MsgForAlertList
+
+        onKeypress =
+            maybeMatcher
+                |> Maybe.map (AddFilterMatcher True >> MsgForAlertList)
+                |> Maybe.withDefault Noop
+                |> onKey "keypress" 13
```
stuartnelson3: Could we move the magic numbers into constants? I'll forget every time that `13` is the enter keycode :)

ui/app/src/Views/AlertList/FilterBar.elm
```diff
@@ -1,20 +1,137 @@
...
+
+        onKeyup =
+            onKey "keyup" 8 (PressingBackspace False |> MsgForAlertList)
```
stuartnelson3: here too (magic number -> constant)

ui/app/src/Views/AlertList/Updates.elm
stuartnelson3: From a dom event we can trigger only a single message.

I get where @mxinden is coming from -- there is logic in both of these messages that decides to either set the matcher text with X, or to use the matcher text in model.matcherText.

I think for right now we just stick with this, it's pretty simple as it as, and the idea I have for three messages doesn't really save us much.

Something like:

```elm
type AlertListMsg
    | AddFilterMatcher String Utils.Filter.Matcher
    | DeleteFilterMatcher String Utils.Filter.Matcher
    | UpdateFilterMatcher String
```

But I don't know if the extra message wins us much

ui/app/src/Views/AlertList/FilterBar.elm
```diff
@@ -1,20 +1,137 @@
 module Views.AlertList.FilterBar exposing (view)
 
-import Html exposing (Html, div, span, input, text, button, i)
-import Html.Attributes exposing (value, class)
-import Html.Events exposing (onClick, onInput)
+import Html exposing (Html, Attribute, div, span, input, text, button, i, small)
+import Html.Attributes exposing (value, class, style, disabled, id)
+import Html.Events exposing (onClick, onInput, on, keyCode)
+import Utils.Filter exposing (Matcher)
+import Views.AlertList.Types exposing (AlertListMsg(..))
```
w0rm: Changed to a qualified import.

w0rm:
@mxinden @stuartnelson3 I believe that all concerns have been addressed and this is ready to be merged

ui/app/src/Views/AlertList/FilterBar.elm
mxinden: Ok, I understand.  Thanks @stuartnelson3 

@w0rm Could you add this as a comment to the code? For me this was difficult to understand. Just to help my future me in 2 month.

w0rm: @mxinden added a comment to the `AlertList` model

stuartnelson3: To help all of our future selves :)

PR #717, opened 2017-04-18
Update status impl

stuartnelson3:
Adding on to #711


inhibit/inhibit.go
```diff
@@ -117,11 +117,11 @@ func (ih *Inhibitor) Mutes(lset model.LabelSet) bool {
 
 	for _, r := range ih.rules {
 		if r.TargetMatchers.Match(lset) && r.hasEqual(lset) {
-			ih.marker.SetInhibited(fp, true)
+			ih.marker.SetStatus(fp, types.Inhibited)
 			return true
 		}
 	}
-	ih.marker.SetInhibited(fp, false)
+	ih.marker.SetStatus(fp, types.Active)
```
stuartnelson3: I'm not sure if this is semantically correct. It appears that alerts will be marked in the notify pipeline, so setting them to active here would be temporary until they get checked as part of the pipeline. Can someone confirm? @brancz

prymitive: If we don't know the real status until notify can re-check it again then maybe we should reset it into `types.Unprocessed`, wouldn't that make more sense?

types/types.go
```diff
@@ -23,67 +24,158 @@ import (
 	"github.com/prometheus/common/model"
 )
 
+type state uint8
+
+const (
+	Unprocessed state = iota
```
prymitive: I'm not sure if `Unprocessed` is the best name for it, I was thinking about calling it `New` but that on the other hand might be mistaken for an alert that _just fired_ rather than indicate what we really mean here.

types/types.go
```diff
@@ -23,67 +24,158 @@ import (
...
+
+// custom serializer for alertStatus that handles providing default value
+// if alertStatus for given fingerprint is missing from the statusMap
+func (a *AlertStatus) MarshalJSON() ([]byte, error) {
+	status, found := statusMap[a.status]
+	if !found {
+		status = "unknown"
+	}
+
+	// TODO: Do we want to return all values?
```
prymitive: This would make sense for `Silenced` status as there might be multiple silences that match our alert. But this means that notify would need to check all, which I see that this PR is adding, so I would say yes, we should return everything.

types/types.go
stuartnelson3: Yeah I'm not sure what name to provide. I don't feel strongly one way or the other.

inhibit/inhibit.go
stuartnelson3: Yeah, it has to go through both the inhibit and silence exec stages, so it'll be set correctly.

stuartnelson3:
Does this seem sane to people? @fabxc @brancz @mxinden

types/types.go
```diff
@@ -23,67 +24,177 @@ import (
 	"github.com/prometheus/common/model"
 )
 
+type State uint8
+
+const (
+	Unprocessed State = iota
```
fabxc: Always prefix enums with the original type.

The type should be `AlertState` to know what it's even about. Thereby the values would be `AlertStateUnprocessed` etc.

provider/mem/mem_test.go
```diff
@@ -94,6 +96,86 @@ func TestAlertsPut(t *testing.T) {
 	}
 }
 
+func TestAlertsGC(t *testing.T) {
+	dir, err := ioutil.TempDir("", "alerts_test")
```
fabxc: `defer os.RemoveAll(dir)`

types/types.go
```diff
@@ -23,67 +24,177 @@ import (
...
+	}
+
+	return json.Marshal(status)
+}
+
+var statusMap = map[State]string{
+	Unprocessed: "unprocessed",
+	Active:      "active",
+	Silenced:    "silenced",
+	Inhibited:   "inhibited",
```
fabxc: An alert can be silenced and inhibited. More so once we move towards a model where this is shown immediately and not just after initial processing.

So "unprocessed", "active", and "suppressed" might be more accurate? @brian-brazil

types/types.go
```diff
@@ -23,67 +24,177 @@ import (
...
+	Unprocessed State = iota
+	Active
+	Silenced
+	Inhibited
+)
+
+// AlertStatus stores the state and values associated with an Alert.
+type AlertStatus struct {
+	status State
+	values []string
```
fabxc: Just having silence IDs here as a bunch of strings might fit with the UI currently offers, but it has semantically no real meaning from an API perspective.
What if we also want to show which alerts inhibited the given alert? How would we distinguish its ID from that of a silence.

I think the status should rather have "status", "silencedBy" (list of silence IDs), and "inhibitedBy" (list of alert IDs).

Also the attributes are both unexported so they shouldn't be marshalled by the JSON decoder.

types/types.go
brian-brazil: Sounds understandable to me

prymitive: I think there's a difference between silenced and inhibited from the user perspective and Alertmanager should distinguish between these two states. Inhibition and silencing is used for different purposes. One is "oh it's silenced, someone is working on it" the other is "oh it's caused by some other issue" (btw. there's no way to ACK an alert, so silences might be used as ACKs).
If alert is inhibited than it's clear that it's suppressed by another alert, so when that other alert is resolved our inhibited alert will fire. Silenced alert on the other hand will never fire while the silence is active. If you just see "suppressed" you won't know what that really means and whenever this alert will fire if there's an action to take in the near future (it won't if it's silenced).

stuartnelson3: True, but then from `AlertStatus` we could infer what sort of suppressed it is: if it has silenceIDs in the silencedBy field, it's silenced, if it has alertIDs in the inhibitedBy field it's inhibited. I suppose I'm not a big fan of having to infer the state, though.

types/types.go
stuartnelson3: `AlertStatus` implements the marshaler interface so we can convert the enum into a string representation, so that handles marshaling the unexported fields.

I flatten the `AlertStatus` inside `APIAlert` in `dispatch.go:139`, but I'm not sure if that's the "right" way to do it.

inhibit/inhibit.go
```diff
@@ -117,13 +117,17 @@ func (ih *Inhibitor) Mutes(lset model.LabelSet) bool {
 
 	for _, r := range ih.rules {
 		if r.TargetMatchers.Match(lset) && r.hasEqual(lset) {
-			ih.marker.SetInhibited(fp, true)
+			// We set a fake alert ID for now so that we can
+			// distinguish AlertStateSuppressed between silenced
+			// and/or inhibited. In the future there will be unique
+			// identifiers for alerts.
+			ih.marker.SetInhibited(fp, "123")
```
stuartnelson3: I put these semantics in place now even though we won't use them. This does lead to potential confusing behavior if we attempt to `SetInhibited` elsewhere before unique identifiers for alerts exist, but the symmetry with `SetSilenced` made sense to me.

inhibit/inhibit.go
```diff
@@ -116,12 +117,8 @@ func (ih *Inhibitor) Mutes(lset model.LabelSet) bool {
...
 
 	for _, r := range ih.rules {
-		if r.TargetMatchers.Match(lset) && r.hasEqual(lset) {
-			// We set a fake alert ID for now so that we can
-			// distinguish AlertStateSuppressed between silenced
-			// and/or inhibited. In the future there will be unique
-			// identifiers for alerts.
-			ih.marker.SetInhibited(fp, "123")
+		if inhibitedByFP, eq := r.hasEqual(lset); r.TargetMatchers.Match(lset) && eq {
+			ih.marker.SetInhibited(fp, fmt.Sprintf("%d", inhibitedByFP))
```
stuartnelson3: @fabxc Does this look correct?

fabxc:
👍

stuartnelson3:
thanks for getting this started @prymitive !!!

prymitive:
Is there any current estimate when would the next release be out? One that would contain this and possibly #748

brancz:
A patch release should go out very soon to release https://github.com/prometheus/alertmanager/pull/750

Issue #718, opened 2017-04-18
Add required name field to routing nodes


fabxc:
I'd like to propose a mandatory new `name` field to be added to each node in the configured routing tree.

Reasoning as follows:

We need some sort of node identity to store whether we sent a notification. By current idea this happens by taking a hash over all matchers of the final node and all its parents. This obviously drops relevant information that's part of how an alert ends up in a certain node. For example, the `continue` keyword means siblings of all nodes in the path an alert took could have played into the routing decision.  
Considering this in the hash would be very complex and would trigger unnecessary resending of notifications. Not doing it gives us some sort of reasonble mix of triggering and not triggering resends on minor config changes.   
It's not really possible to infer the user's semantic intent from the given matchers. It could cause accidental oscillations in large configs that will be nearly impossible to debug.

The most straightforward and easiest to reason about way of solving this is having a name for each node, that must be unique among its siblings. Any node can then be identified by the concatenation of the names, e.g. `team-infra/eu-west/monitoring`.

Cons:
* Adds another config field, which means mental overhead for the user.
* The name will often be self-explanatory for simple matchers like `team == monitoring` and seem superfluous.

Pros:
* `name` is an addition that is pretty easy to understand
* Easy to reason about and handy node names for debugging
* Completely avoids potential hash collisions
* Config changes do not trigger noisy notifications as the name clearly aligns the user's semantic intent rather than a hash preserving identity in a best-effort manner
* Forces users to provide a descriptive name for nodes more complex matchers such als multiple regexes
* No unexpected edge cases in large configs and generally straightforward internal semantics

Those are the arguments that came up on IRC with @brian-brazil.
Obviously, I'm leaning towards the pro side.

Opinions and further arguments to make a final call would be appreciated. 

@beorn7 @juliusv @mxinden @brancz

fabxc:
@mattbostock would be interesting to hear from a heavy user how adding of this field would impact your user experience.

stuartnelson3:
I'm in favor since the one-time cost of adding a name field doesn't come close to outweighing the benefits. Having a config linter would also help.

brian-brazil:
> I'm in favor since the one-time cost of adding a name field doesn't come close to outweighing the benefits. 

It's an ongoing cost, as all new routes also need it added.

stuartnelson3:
But you add it once to each route?

brian-brazil:
[...]

PR #725, opened 2017-04-21
*: switch group key to matcher serialization

fabxc:
Turn the GroupKey into a string that is composed of the matchers if the
path in the routing tree and the grouping labels.
Only hash it at the very end to ensure we don't exceed size limits of
integration APIs.

As described in #718 

@brian-brazil


brian-brazil:
:+1:

PR #727, opened 2017-04-21
test: Ensure bindata.go is always up to date

mxinden:
This adds three more steps to the Travis tests.
1. Compile Elm -> script.js
2. Bundle html.index & script.js -> bindata.go
3. Check if bindata.go hasn't changed


mxinden:
@stuartnelson3 @w0rm Make sure you have the current version of elm-format installed whenever you commit code.

w0rm:
@mxinden what about generating it on travis instead? It feels strange having to commit generated data.

stuartnelson3:
Non-pr comment: You should have access to the repo so don't have to open a PR against your own fork

stuartnelson3:
Hm, yeah this is something that's been an issue before, forgetting to generate the bindata.

Now might be a good time to just generate it in `circle.yml` and not have to worry about people committing it.

mxinden:
@w0rm valid point. But for now we want to enable anybody to build alertmanager without setting up elm on their machine. Thereby we need to check in `bindata.go` as well :/
In the long run we can do this inside docker containers.

Issue #689, opened 2017-04-03
Regex match on receiver query string


stuartnelson3:
Allow for a greedy regex, i.e. `backend-crit` and `backend-warn` would both be matched by `backend-.*`

stuartnelson3:
Closed by #729

PR #729, opened 2017-04-23
Allow filtering receiver with a regex

stuartnelson3:
e.g.:
http://localhost:5000/#/alerts?receiver=~(backend|ops)-warn

Most of the noise in this PR is because I moved the `Filter` type alias from `Utils.Types` into `Utils.Filter`. This was necessary because I am redefining `filter.receiver`, and it introduced a circular dependency if it remained in `Utils.Types`.

I tried to use the parser @w0rm wrote but it ended up being easier just looking for a `~` at the start of the query string value.

I'm not supporting `NotEq` or `NotRegexMatch` because how we represent that doesn't work with query string parameters:

```bash
# promQL
foo!=bar

# query string..?
foo=!bar
```

Requested by @beorn7 in #689. Does this match your requirements?


beorn7:
Regexp matching is great, but I don't think we can implement it this way in the HTTP API.

What happens if the receiver _actually_ starts with a "~"?

How about _always_ doing a regexp match? Then the user can escape as usual. It should rarely get into  the way of the user. We have to make sure it's handled correctly by the UI (if a receiver named "foo.bar" is selected by clicking, the UI needs to request "foo\.bar".)

beorn7:
General remarks:

The filtering is happening all in the browser, right? I guess that's the case right now anyway. With many alerts, this will become infeasible (as in: When i want to quickly look at my alerts because everything is on fire, I need to load the millions of alerts that are firing company wide first). But yeah, probably not the concern of this change in particular.

Different thing: To keep things consistent, the regexp should be anchored (AKA "full match", but that means different things to different people), i.e. as we do in everywhere in Prometheus. Not sure if that's the default in Elm.

ui/app/src/Views/AlertList/Filter.elm
```diff
@@ -2,14 +2,26 @@ module Views.AlertList.Filter exposing (receiver, silenced, matchers)
...
+        Just { key, op, value } ->
+            case op of
+                Eq ->
+                    by (filterAlertGroup ((==) value)) groups
+
+                RegexMatch ->
+                    by (filterAlertGroup ((regex >> contains) value)) groups
+
+                NotEq ->
+                    groups
```
mxinden: shouldn't his be:

```elm
by (filterAlertGroup ((/=) value)) groups
```

w0rm:
I don't like `=~` in the query string, we'd rather have another query param that controls the matching operation that. e.g. `receiver=(backend|ops)-warn&receiverOp=RegexMatch`.

ui/app/src/Views/AlertList/Filter.elm
stuartnelson3: I made the comment in the description that `NotEq` and `NotRegexMatch`aren't supported, but didn't want to have a `_ ->` match group to catch everything.

stuartnelson3:
@beorn7 
- Filtering is all happening in the client
- I don't think the regex is anchored

@w0rm 
Yeah I guess the `~` is kind of funny. I don't want to add more query parameters, though. I think beorn's suggestion makes sense: treat every receiver value as a (fully anchored) regex and anchor it. It will match any direct lookups (`receiver=backend-warn`), but then if a user adds anything then that will also work (`receiver=backend-.*`). Thoughts?

w0rm:
@stuartnelson3 yeah this is fine with me

ui/app/src/Views/AlertList/Filter.elm
stuartnelson3: I added the other functions if they somehow become used in the future..

stuartnelson3:
K, updated it so it's always an anchored regex match.

PR #751, opened 2017-04-27
Add config hash metric

### LINKED ISSUE #596: Metric for config consistency
In a highly available setup of the Alertmanager each instance has to be configured in the same way for the cluster to be working as expected, therefore it is a problem if any of the loaded configs differ.

The easiest way to figure out an inconsistency of such a case would be to hash the currently loaded config and expose the hash as a metric (gauge). Then we can alert on the hashes to be equal.

Thoughts?

@fabxc @brian-brazil @alexsomesan

brancz:
Closes #596

@fabxc @mxinden @brian-brazil

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -14,6 +14,8 @@
 package main
 
 import (
+	"crypto/md5"
+	"encoding/binary"
 	"flag"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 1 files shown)


cmd/alertmanager/main.go
```diff
@@ -423,3 +433,10 @@ func mustHostname() string {
 	}
 	return hostname
 }
+
+func md5HashAsMetricValue(data []byte) float64 {
+	sum := md5.Sum(data)
+	var bytes []byte = sum[0:8]
+	bits := binary.LittleEndian.Uint64(bytes)
+	return math.Float64frombits(bits)
```
brian-brazil: This can result in a NaN which will cause problems downstream. I'd suggest limiting to say 48 bits and sorting it as a value rather than bits.

brancz: 48 bits sounds reasonable regarding the NaN problem.

> storing it as a value rather than bits

What do you mean by this?

brian-brazil: Cast the int to a float

cmd/alertmanager/main.go
```diff
@@ -427,3 +436,11 @@ func mustHostname() string {
 	}
 	return hostname
 }
+
+func md5HashAsMetricValue(data []byte) float64 {
+	sum := md5.Sum(data)
+	smallSum := sum[0:6] // We only want 48 bits to avoid NaN
```
brian-brazil: ...as a float64 only has a 53 bit mantissa.

cmd/alertmanager/main.go
```diff
@@ -427,3 +436,12 @@ func mustHostname() string {
 	}
 	return hostname
 }
+
+func md5HashAsMetricValue(data []byte) float64 {
+	sum := md5.Sum(data)
+	// We only want 48 bits to avoid NaN as a float64 only has a 53 bit mantissa.
+	smallSum := sum[0:6]
+	var bytes = make([]byte, 8)
```
brian-brazil: Can you not pass smallSum to  Uint64 directly, as it takes a []byte?

cmd/alertmanager/main.go
```diff
@@ -427,3 +436,12 @@ func mustHostname() string {
 	}
 	return hostname
 }
+
+func md5HashAsMetricValue(data []byte) float64 {
+	sum := md5.Sum(data)
+	// We only want 48 bits to avoid NaN as a float64 only has a 53 bit mantissa.
```
brian-brazil: No need to mention NaN, as we're no longer messing with bit representations

cmd/alertmanager/main.go
brancz: it panics with index out of bound when we don't have a byte slice with length >= 8, and the type of `sum` is `[16]byte`, which we can't pass directly into the function.

brian-brazil:
:+1:

PR #754, opened 2017-04-28
Switch to /alerts endpoint and restlye alerts list page

### LINKED ISSUE #694: Use /api/v1/alerts instead of /api/v1/alerts/groups
I have been talking with @fabxc and @brancz about our current Alertmanager API consumption. We decided to keep `AlertGroups` and `AlertBlocks` a server-side implementation detail that the UI doesn't need to know of. This would flatten the UI representation of alerts and keep things simple. The UI would just consume `/api/v1/alerts` and would only know of the entity `Alerts` not `AlertGroups` or `Blocks`.

Steps needed:
1. Implement filter logic for endpoint `/alerts` like it already exists on `/alerts/groups`
2. Refactor UI to use `/alerts` instead of `/alerts/groups`
3. (Future) Implement groupBy logic for endpoint `/alerts`

@stuartnelson3 What do you think? As you implemented the filter logic for the `/alerts/groups` endpoint, is there any show stopper that I overlooked?

mxinden:
Instead of using the /alerts/groups endpoint to retriev alerts this
commit introduces usage of the /alerts endpoint. This bypasses a lot of
complexity by ignoring alert groups and alert blocks. Groups and blocks
are left as a pure back-end detail.

Fixes #694


mxinden:
@stuartnelson3 @w0rm This is a work in progress version. I am opening up this PR at this not yet finished state to get feedback as fast as possible. It introduces quite a lot of changes.

mxinden:
Small preview of the restyled AlertList page:

screenshot from 2017-04-28 15-24-56

ui/app/src/Utils/Date.elm
```diff
@@ -62,14 +65,19 @@ durationFormat time =
         |> String.trim
 
 
-dateFormat : ISO8601.Time -> String
```
w0rm: Maybe smth for the future, I think it makes sense to only have one datetime library

ui/app/src/Views/AlertList/AlertView.elm
```diff
@@ -0,0 +1,91 @@
...
+        , small [] [ text <| Utils.Date.dateFormat time ]
+        ]
+
+
+labelButtons : List ( String, String ) -> Html Msg
+labelButtons labels =
+    let
+        -- the alertname label should be first
+        sortedLabels =
+            List.append
```
w0rm: Looks like a good case for [`List.partition`](http://package.elm-lang.org/packages/elm-lang/core/latest/List#partition). Together with [`uncurry`](http://package.elm-lang.org/packages/elm-lang/core/5.1.1/Basics#uncurry) this looks nice and short:

```
labelButtons labels =
  labels
  |> List.partition (Tuple.first >> (==) "alertname") -- returns a tuple of two lists
  |> uncurry (++) -- joins two lists from a tuple
  |> List.map labelButton -- creates a list of lables
  |> div [] -- wraps labels in a div
```

You can then rewrite it in a point free style, but I wouldn't do it because it decreases readability:

```
labelButtons =
 List.partition (Tuple.first >> (==) "alertname")
 >> uncurry (++)
 >> List.map labelButton
 >> div []
```

ui/app/src/Utils/Api.elm
```diff
@@ -66,7 +66,7 @@ iso8601Time =
 
 baseUrl : String
 baseUrl =
-    "/api/v1"
+    "http://localhost:9093/api/v1"
```
w0rm: Would be nice to configure this from [the app flags](http://package.elm-lang.org/packages/elm-lang/html/1.1.0/Html-App#programWithFlags), so a `go` backend would pass it down to the html template.

w0rm:
Not a fan of the tag icons, using it many times on the page adds visual noise without being useful. Also more whitespace would look cleaner than table borders. If we figure out what information do we want to see and the importance of it, I'd like to put some effort into the design afterwards. I own a couple of [Edward Tufte](https://www.edwardtufte.com/tufte/books_vdqi) books :)

But for now I think this looks good enough!

ui/app/src/Utils/Views.elm
```diff
@@ -61,8 +65,8 @@ textField labelText content msg =
 
 buttonLink : String -> String -> String -> msg -> Html msg
 buttonLink icon link color msg =
-    a [ class <| "f6 link br1 ba mr1 ph3 pv2 mb2 dib " ++ color, href link, onClick msg ]
-        [ i [ class <| "fa fa-3 " ++ icon ] []
+    a [ class <| "" ++ color, href link, onClick msg ]
```
w0rm: Thanks for getting rid of the tachyons css classes, we should eventually rule it out

fabxc:
Agreeing with @w0rm on the icons on every labels for the same reason.
Also agree that we can address design concerns later.

ui/app/src/Utils/Date.elm
mxinden: For sure. I would prefer to use [rluiten/elm-date-extra](https://github.com/rluiten/elm-date-extra) over [jweir/elm-iso8601](https://github.com/jweir/elm-iso8601) as it offers more string formatting functionality. One issue for me was to have leading zeroes on single digit hours.

@stuartnelson3 What do you think?

If you are both fine with the first one I would replace the current one entirely with the new one.

ui/app/src/Views/AlertList/AlertView.elm
mxinden: Wow, this is beautiful. But @fabxc suggested to not render the label key if it is "alertname" to shorten the label list presentation. So the alertname label needs to have a different `labelButton` render function then any other label (See current implementation). I will use the partition function though.

But I will definitely

ui/app/src/Utils/Api.elm
mxinden: Ou, sorry - should not have been included in the commit.

Thought about this as well. But this would mean we need some kind of templating for the index.html file, right?

w0rm: @mxinden right, we would need to serve this page instead of including it as a static asset

mxinden: Both in the dev-server and the actual Alertmanager. I would like to not have two `index.html` files.

Happy for suggestions. Definitely a thing we can improve.

mxinden:
@w0rm @fabxc I have removed the label icons. Would love to see some Edward Tufte style design suggestions in the future!

mxinden:
@fabxc @stuartnelson3 What do you think of the back-end updates?

Is [this](https://github.com/prometheus/alertmanager/blob/master/types/types.go#L290) relevant for the /alerts endpoint? Should we really return an `time.Time{}` here?

fabxc:
Yes, we should. If we receive an alert we set a tentative endsAt value at which the alert ends UNLESS a the alert is sent again before that time. In that case the tentative end time is updated.
Exposing it to the user would just be confusing.

We should just get rid of this later when reworking the internal alert layer and replace it with an explicit `timesOutAt` field or something.

mxinden:
I would like to delay the decision of our favourite way to write css to issue #762.

Issue #743, opened 2017-04-26
Silencepage filter bar


mxinden:
The filter bar on the silence page should be updated like the one on the alert page.

stuartnelson3:
closed by #755

PR #755, opened 2017-04-28
Extract Alert Filter Bar

stuartnelson3:
Requested in #743 

~~This is currently a WIP, but I wanted to push it up now. No need to review yet.~~

@w0rm and I paired on extracting `FilterBar` from alertlist, and then integrating it back into alert list as an external module. I integrated it with silencelist, and added a few tests.


mxinden:
@stuartnelson3 In regards to eb50155: Is it possible to remove the `?` as well? Not very important!

ui/app/src/Views.elm
```diff
@@ -34,8 +34,9 @@ currentView model =
         AlertsRoute filter ->
             AlertList.view model.alertList filter
 
-        SilenceListRoute route ->
-            SilenceList.view model.silences model.silence model.currentTime model.filter
+        SilenceListRoute filter ->
+            -- Is this filter needed???
```
mxinden: @stuartnelson3 @w0rm As far as I can see we can remove both the comment and the `filter` payload from the `SilenceListRoute`.

In addition as the `filter` in the main model does not only hold the filter text but also all other kind of url parameters (e.g. `showSilenced`), what do you think of renaming it to `urlParameters`?

ui/app/src/Types.elm
```diff
@@ -13,7 +13,7 @@ import Time
 
 
 type alias Model =
-    { silences : ApiData (List Silence)
+    { silences : SilenceList.Model
```
mxinden: Can we call this `silenceList` (see `alertList`)?

w0rm:
Apart from what @stuartnelson3 mentioned, this looks good to me. Thanks for writing tests!

ui/app/src/Views/FilterBar/Views.elm
```diff
@@ -142,7 +141,7 @@ view matchers matcherText backspacePressed =
                             [ text "Custom matcher, e.g."
                             , button
                                 [ class "btn btn-link btn-sm align-baseline"
-                                , onClick (UpdateMatcherText "env=\"production\"" |> MsgForAlertList)
+                                , onClick (UpdateMatcherText "env=\"production\"")
```
mxinden: Small thing: Can we move `"env=\"production\"` to a constant so it stays in sync with line 146?

ui/app/src/Utils/Views.elm
```diff
@@ -6,12 +6,12 @@ import Html.Events exposing (onCheck, onInput, onClick)
 import Http exposing (Error(..))
 
 
-labelButton : Maybe msg -> ( String, String ) -> Html msg
-labelButton maybeMsg ( key, value ) =
+labelButton : Maybe msg -> String -> Html msg
+labelButton maybeMsg labelText =
```
mxinden: 👍 for `labelText`. I think I changed it to `(key, value)` in my previous PR, sorry about that.

ui/app/src/Views.elm
stuartnelson3: I would like to keep it as filter since the actual value is a `Filter`. If you check out the alertlist and silencelist parsing, the url values get parsed into this record type and then that is what is being passed around.

stuartnelson3:
I have no idea what is wrong with my generation of bindata. I'm copying and executing the docker commands on my local machine.

mxinden:
@stuartnelson3 I will give it a shot and update your last commit.

PR #757, opened 2017-04-28
new smtp conf parameters: skip certs verification & custom root ca

KarolBedkowski:
Two new configuration parameters related to smtp with tls:
smtp_tls_skip_verify / tls_skip_verify - do not verify certificates
smtp_tls_root_ca / tls_root_ca - use custom certificate for smtp server from pem file

@stuartnelson3

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -197,6 +197,9 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 				ec.RequireTLS = new(bool)
 				*ec.RequireTLS = c.Global.SMTPRequireTLS
 			}
+			if ec.TLSConfig == nil {
+				ec.TLSConfig = c.Global.SMTPTLSConfig
+			}
 		}
 		for _, sc := range rcv.SlackConfigs {
 			if sc.HTTPConfig == nil {
@@ -397,6 +400,8 @@ type GlobalConfig struct {
 	WeChatAPICorpID  string `yaml:"wechat_api_corp_id,omitempty" json:"wechat_api_corp_id,omitempty"`
 	VictorOpsAPIURL  string `yaml:"victorops_api_url,omitempty" json:"victorops_api_url,omitempty"`
 	VictorOpsAPIKey  Secret `yaml:"victorops_api_key,omitempty" json:"victorops_api_key,omitempty"`
+
+	SMTPTLSConfig *commoncfg.TLSConfig `yaml:"smtp_tls_config"`
 
 	// Catches all undefined fields and must be empty after parsing.
 	XXX map[string]interface{} `yaml:",inline" json:"-"`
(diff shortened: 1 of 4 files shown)


sgissi:
I have disabled TLS verification once because my SMTP provider would not
provide a correct SAN but it is in general a bad practice. I'm totally Ok
with custom CA in the other hand.

2017-04-28 9:41 GMT-07:00 Karol Będkowski <notifications@github.com>:

KarolBedkowski:
I agree, but disabling verification is quite useful in broken/test environments - when other methods fail.

KarolBedkowski:
Hi,
@stuartnelson3 - I've updated this path; now it use prometheus/common/config.TLSConfig for configuration.

stuartnelson3:
I'll bounce this to @brian-brazil and @fabxc since they should be the ones making this decision

notify/impl.go
```diff
@@ -272,7 +272,14 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
...
 		}
-		tlsConf := &tls.Config{ServerName: host}
+		tlsConf := &tls.Config{}
+		if n.conf.TLSConfig != nil {
+			tlsConf, err = n.conf.TLSConfig.GenerateConfig()
+			if err != nil {
+				return true, fmt.Errorf("generate TLS configuration failed: %s", err)
+			}
+		}
+		tlsConf.ServerName = host
```
brian-brazil: Why are you setting this?

config/config.go
```diff
@@ -317,6 +321,8 @@ type GlobalConfig struct {
 	OpsGenieAPIHost  string `yaml:"opsgenie_api_host" json:"opsgenie_api_host"`
 	VictorOpsAPIURL  string `yaml:"victorops_api_url" json:"victorops_api_url"`
 
+	SMTPTLSConfig *config.TLSConfig `yaml:"smtp_tls_config"`
```
brian-brazil: I don't think a struct like this works as a default, all the rest are simple types. This should possibly be per-receiver only.

KarolBedkowski:
@brian-brazil 
In my environment I have to set custom ca certificate for smtp server (or turn off whole certificate verification).
I don't want to setup this in per-receiver - I have one smtp server and many email receivers. Set it globally (like stmp_smarthost)  IMHO simplify configuration. Of course, change/set it in each receiver is possible (tls_config: ...) if you don't want setup it globally.  I think, this is ok for smtp and maybe webhook - I don't use other.

I've update patch - remove one unnecessary tls.Config allocation.

brian-brazil:
This is something you could handle in your configuration management system.

KarolBedkowski:
Unfortunately, I can't. My env is.. specific.
I need to use smtp directly, without smart host, I need accept custom ca but I can't add this to global system system certificates, etc.

brian-brazil:
We generally presume that our users have configuration management that includes some form of templating system, as that's pretty much essential for managing any non-trivial environment.

KarolBedkowski:
Well, there are also legacy system / corpo-world, where we can't change too much what we get and we try to survive.

Anyway, still Alertmanager can't talk with smtp when custom ca / client certificates are required. 

I propose simple patch for change this. It will be nice, if it be merged (I think, there are more people facing this problem). If not - no problem for me.

JeffHemmen:
+1

simonpasquier:
Closed by #1528

PR #759, opened 2017-04-28
Serialize AlertStatus as 'status'

prymitive:
AlertStatus doesn't have json tag with the field name, so it's serialized into 'Status', and it's the only uppercase field in the alert object. Tag it with 'status' name for consistency


prymitive:
@stuartnelson3 just noticed that this also fixes the UI - templates are testing for `alert.status` so silenced alerts are not marked correctly with _silenced_ labels

fabxc:
Thanks

Issue #744, opened 2017-04-26
Statuspage display more info


mxinden:
Currently the status page displays a very minimal amount of information. It should display all information retrieved by the /status endpoint (See old ui).

mxinden:
@stuartnelson3 @w0rm To prevent duplicate work: I will start working on this.

mxinden:
Closed by #761

PR #761, opened 2017-04-30
Display more information on status page

mxinden:
Adding more details of the `/status` endpoint to the status page. I have not put a lot of focus on design, I was more looking to be feature equal to the old Alertmanager UI status page (minus the routing tree).

I would suggest to delay the design work to a later PR and get the v0.7 milestone done. What do you think?

Closes issue #744


ui/app/src/Views.elm
```diff
@@ -26,7 +26,7 @@ currentView : Model -> Html Msg
 currentView model =
     case model.route of
         StatusRoute ->
-            Status.view model
+            Status.view model.status
```
w0rm: Thumbs up for extracting a status model

PR #763, opened 2017-05-02
Init result slice to prevent /alerts returning `null`

mxinden:
When there are no alerts to return the api endpoint `/alerts` should
return:
```
{
  "status": "success",
  "data": []
}
```
instead of
```
{
  "status": "success",
  "data": null
}
```

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -256,10 +256,12 @@ type APIAlert struct {
 func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 	var (
 		err error
-		res []*APIAlert
+		// Initialize result slice to prevent api returning `null` when there
+		// are no alerts present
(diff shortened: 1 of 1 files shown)


api/api.go
```diff
@@ -259,6 +259,10 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 		res []*APIAlert
 	)
 
+	// Initialize result slice to prevent api from returning `null` when there
+	// are no alerts present
+	res = []*APIAlert{}
```
fabxc: You can move this into the variabel declaration block above. The matchers too actually:

```go
var (
    err error
    res = []*APIAlert{}
    matchers = []*labels.Matcher{}
)
```

Issue #696, opened 2017-04-05
It should be possible to disable Mesh


mattbostock:
Alertmanager always listens on the Mesh port and cannot be disabled, even if you don't use the HA functionality.

It should be possible to turn off the HA functionality for deployments that don't use it or require it.

mxinden:
Closing here because #764 is merged.

PR #764, opened 2017-05-02
Added option to disable AM listening on mesh port

Conorbro:
Implements #696


cmd/alertmanager/main.go
```diff
@@ -117,7 +117,13 @@ func main() {
 	}
 
 	logger := log.NewLogger(os.Stderr)
-	mrouter := initMesh(*meshListen, *hwaddr, *nickname, *password)
+
+	var mrouter *mesh.Router
+	if *meshListen == "" {
+		mrouter = initMesh(*flag.String("flag-name", "0.0.0.0:0", "flag-msg"), *hwaddr, *nickname, *password)
```
fabxc: I assume this is to provide a dummy mesh router because things below would fail if it is nil?

So right now silence and nflog actually attempt to gossip their data to a dummy mesh router, that is not actually running. I'm not sure there are any guarantees on all methods against an un-started router returning etc.

As both can be configured to not gossip their data, I think we should rather do that by not providing the Gossip/Mesh config options to them. That seems generally safer/cleaner.

cmd/alertmanager/main.go
```diff
@@ -132,22 +138,29 @@ func main() {
 		nflog.WithMaintenance(15*time.Minute, stopc, wg.Done),
 		nflog.WithMetrics(prometheus.DefaultRegisterer),
 		nflog.WithLogger(logger.With("component", "nflog")),
-	)
+	}
+	if *meshListen == "" {
```
brian-brazil: It'd be clearer to conditionally append, rather than to remove the entry afterwards

cmd/alertmanager/main.go
fabxc: With the conditional options used now, we should no longer need to instantiate this dummy router.

cmd/alertmanager/main.go
```diff
@@ -168,7 +179,11 @@ func main() {
 		wg.Done()
 	}()
 
-	mrouter.Start()
+	// Disable mesh if empty string passed for mesh.listen-address flag
```
brian-brazil: Full stop at the end of comments.

Issue #356, opened 2016-05-23
continue:true not matching after the first match


mindprince:
I have the following route configuration:

```
route:
  continue: true

  # default
  receiver: hipchat-test

  routes:
    - match:
        notify: hipchat-test
      receiver: hipchat-test
      continue: true

    - match:
        notify: hipchat
      receiver: hipchat
      continue: true

    - match:
        notify: hipchat-net
      receiver: hipchat-net
      continue: true

    - match:
        notify: hipchat-data
      receiver: hipchat-data
      continue: true

    - match:
        notify_team: hipchat-data
      receiver: hipchat-data
      continue: true
```

My alert has these labels: `{notify_team="hipchat-data", notify="hipchat"}`
But it is only getting routed to `hipchat` and not to `hipchat-data`. I want to route to both of these.
Not sure if I am doing something wrong or if `continue` is broken.

mindprince:
I tested using: https://prometheus.io/webtools/alerting/routing-tree-editor/

dqminh:
i took a brief look and I think there should be a bug in https://github.com/prometheus/alertmanager/blob/master/ui/lib/routing-tree.js#L59-L62. In the cases of multiple matches, we only take the first match.

Also specified in the doc:

```
A route block defines a node in a routing tree and its children. Its optional configuration parameters are inherited from its parent node if not set.
```
However `continue` is not inherited https://github.com/prometheus/alertmanager/blob/master/route.go#L97. I'm not sure if this is a bug in documentation or code.

brian-brazil:
Continue should not be inherited.

raypettersen:
Stumbled across the same problem today. Is this a bug or what? How can I send an alert to multiple receivers??

rohit01:
Same problem. It's not working for me as well.

rohit01:
Gocha.. 
Continue works if specified for child routes. This attribute does not look like an inherited one.

ngu04:
@rohit01 if you don't mind can you please share route configuration . As it doesn't work for me .

mkrakowitzer:
Hopefully this helps:
```
route:
  group_by: ['cluster','alertname','host']
  group_wait: 30s
  group_interval: 2m
  repeat_interval: 10m
  receiver: slack_general
  routes:
  - match:
      severity: warning
    receiver: slack_general
    continue: true
  - match:
        severity: critical
    receiver: slack_general
    continue: true
  - match:
        severity: warning
[...]

PR #767, opened 2017-05-03
Throw error when continue in route root of Alertmanager config

Conorbro:
In relation to #356. 

`Continue` was incorrectly used in the Alertmanger configuration's root route. 
The lack of error here is misleading as `continue` is not inherited by the sibling routes.


fabxc:
👍

PR #768, opened 2017-05-03
Added unit test for continue in root route error

Conorbro:
Added unit test and comment for #767

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -37,3 +37,26 @@ route:
 		t.Errorf("\nexpected:\n%v\ngot:\n%v", expected, err.Error())
 	}
 }
+
+func TestContinueErrorInRouteRoot(t *testing.T) {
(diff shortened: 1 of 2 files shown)


config/config_test.go
```diff
@@ -37,3 +37,26 @@ route:
...
+
+	_, err := Load(in)
+
+	expected := "cannot have continue in root route"
+
+	if err == nil {
+		t.Fatalf("no error returned, expeceted:\n%v", expected)
+	}
+	if err.Error() != expected {
+		t.Errorf("\nexpected:\n%v\ngot:\n%v", expected, err.Error())
```
fabxc: Can you replace `%v` with `%q`?
This wraps the string in quotes, which makes clear where the expected/received string value starts and ends.

fabxc:
Thanks

PR #769, opened 2017-05-04
Add entity_display_name for VictorOps, use better state_message

### LINKED ISSUE #580: Add title support VictorOps receiver
In #417 support was added for VictorOps (PagerDuty alertnative).
However the receiver doesn't support [most fields listed in the docs](http://victorops.force.com/knowledgebase/articles/Integration/Alert-Ingestion-API-Documentation/), specificaly the `entity_display_name` field which serves as a title.

The result of this is the incidents created by Alertmanager have a random string as a title, forcing the user to open the incident page and search for the description body in order to understand what the incident it about.

The change should be pretty straight forward.

In addition, please allow to set the rest of the fields described in the docs above.

Thanks!

siavashs:
This PR fixes #580 and uses a better `state_message` since this field has a high character limit and is intended for a long, verbose, explanation of the problem.

More details about REST endpoint can be found [here](https://help.victorops.com/knowledge-base/victorops-restendpoint-integration/).


template/default.tmpl
```diff
@@ -46,9 +46,19 @@ Alerts Resolved:
 {{ define "opsgenie.default.source" }}{{ template "__alertmanagerURL" . }}{{ end }}
 
 
-{{ define "victorops.default.state_message" }}{{ template "__subject" . }} | {{ template "__alertmanagerURL" . }}{{ end }}
-{{ define "victorops.default.monitoring_tool" }}{{ template "__alertmanager" . }}{{ end }}
+{{ define "victorops.default.state_message" }}{{ template "__alertmanager" . }}: {{ template "__alertmanagerURL" . }}
```
brian-brazil: Please don't create additional default formats, use what we already have.

siavashs: I'm using `__subject` as `entity_display_name` now, and using a more verbose `state_message`(similar to `opsgenie.default.description`) instead of `__subject`.
In other words having `__subject` in 2 different fields is not very useful.

siavashs:
Please see `entity_display_name ` under [Required / Recommended Fields](https://help.victorops.com/knowledge-base/victorops-restendpoint-integration/#required-recommended-fields)

template/default.tmpl
brian-brazil: Having to maintain a format per notifier is not desirable, this should be a copy (preferably a template) of what we already have for this type of long-form field.

siavashs: Understood, I will use the same template used for `opsgenie.default.description`.

siavashs:
Not sure if this is the right place to mention this: having the generated binary files in repository causes problems with rebasing forks and PRs.

mxinden:
@siavashs In regards to the generated binary files: We plan to address this with the new UI in the future (See [ui-rewrite branch](https://github.com/prometheus/alertmanager/tree/ui-rewrite))

devinodaniel:
@siavashs Thanks for your work on this thus far. Is there anything I could do to help with getting this merged? Having `entity_display_name` would be sweet as we use VO but having the subject of an alert be a random hash is killing our mo-jo.

siavashs:
@devinodaniel I'm still waiting for maintainers to merge this. Regarding the `entity_id` the documentation says:
> ID of the Incident [String]
This field is the identity of the incident and must remain the same throughout the life-cycle of an incident. Make sure you use the same entity_id when triggering, acking, or resolving an incident. Example: "/", i.e... "diskspace/db01.mycompany.com"
If no entity_id is included, a random string will be used.

[alertmanager uses `sha256` hash of the alert group key](https://github.com/prometheus/alertmanager/blob/master/notify/impl.go#L810) for this, so it can later resolve the incident using the same `entity_id`.
It is possible to generate a string like the above example but I'd like the maintainers to comment on this.

ramiamar:
+1
We @ alooma.com would also love to be able to set the entity_display_name

stuartnelson3:
@siavashs we've talked about how to generate identifiers that would lend themselves to being used as `entity_id`, but I don't think we've made any active progress on it.

Could you fix the conflict with `config/notifiers.go`? I'll merge after that, sorry for the long delay.

siavashs:
@stuartnelson3 done, not sure why travis-ci is failing though.

mxinden:
@siavashs Your `/template/internal/deftmpl/bindata.go` is out of sync, just run `make assets` and add `bindata.go` to your commit.

siavashs:
@mxinden When I try to build the assets I get this error:
```
> make assets
>> writing assets
bindata: Failed to stat input path 'ui/app/script.js': lstat ui/app/script.js: no such file or directory
make: *** [assets] Error 1
```

mxinden:
@siavashs Ah, yes. This will be fixed with #792 but I haven't updated the PR yet. Can you comment out line 63 - 68 and then try again?

mxinden:
Cool, I guess that worked. I would still like @stuartnelson3 to look over this, as he was more involved so far.

stuartnelson3:
Hey, I'm on vacation for the next two weeks. If this looks good to @mxinden
then I say go for it, otherwise I will look when I am back

fabxc:
@siavashs, sorry about not looking at this earlier. This generally looks good to me. The documentation on `state_message` is pretty vague by saying "This field has a high character limit...".

We have had problems in the past by our generated messages causing receiver APIs to reject our notifications, which is pretty bad obviously. 
Given that these changes put a full list of all alerts into this field and there can be hundreds of alerts in a single notification, we should add some safety checks here.

Can we find out what the actual maximum length is and then truncate the generated text for this field to that length? Limiting to something like 4KB is probably sensible.

siavashs:
@fabxc good point, I sent a request to VictorOps support regarding the maximum length of `state_message`.

siavashs:
Response from VictorOps support:
>The maximum length for the state_message field is **20KB**.

I will add a check for `state_message` length and truncate it if exceeded **20KB**.

fabxc:
Thanks

devinodaniel:
Thanks all for getting this merged in, any chance an update to the docs https://prometheus.io/docs/alerting/configuration/#<victorops_config> can be made to reflect this capability?

Issue #770, opened 2017-05-04
v0.6.1 severe malfunction.


raypettersen:
We upgraded alertmanager today, and experienced severe problems in the GUI. For example, it seems the previous version had some javascripting cached so the rendering was way off. After testing in incognito, it was somewhat better, but we ran into some weird bugs. For example, when clicking "hide silences", the actual view updated and _showed_ the silences. The exact opposite behavior that was expected. Also, the silence icon was gone so we had no idea of knowing what alarm was silenced or not. We eventually downgraded as we were having so many issues that we didn't dare running this release in production.

Has anyone else experienced the same issues?

raypettersen:
`ime="2017-05-04T13:35:37+02:00" level=warning msg="Could not get file: AssetInfo ui/lib/angular-sanitize.min.js.map not found" source="web.go:31"`
`time="2017-05-04T13:35:37+02:00" level=warning msg="Could not get file: AssetInfo ui/lib/jquery.min.map not found" source="web.go:31" `
`time="2017-05-04T13:35:37+02:00" level=warning msg="Could not get file: AssetInfo ui/lib/angular.min.js.map not found" source="web.go:31" `
`time="2017-05-04T13:35:37+02:00" level=warning msg="Could not get file: AssetInfo ui/lib/angular-route.min.js.map not found" source="web.go:31" `
`time="2017-05-04T13:35:37+02:00" level=warning msg="Could not get file: AssetInfo ui/lib/angular-resource.min.js.map not found" source="web.go:31" `
`time="2017-05-04T13:35:37+02:00" level=warning msg="Could not get file: AssetInfo ui/lib/angular-moment.min.js.map not found" source="web.go:31" `
`time="2017-05-04T13:35:40+02:00" level=warning msg="Could not get file: AssetInfo ui/lib/jquery.min.map not found" source="web.go:31"`

![selection_804](https://cloud.githubusercontent.com/assets/7197374/25705243/ef4402a6-30dc-11e7-83b1-c17fa9f59aff.png)

![selection_802](https://cloud.githubusercontent.com/assets/7197374/25705275/0542e04a-30dd-11e7-88b8-14bbe00b3e58.png)

fabxc:
The logging of `.map` files not being found is normal. That usually happens when you open the dev console in your browser, which than tries to get a `.js.map` file for every `.js` file that was loaded.

Otherwise, I cannot report similar behavior. I've been switching between different versions multiple times.
[...]

PR #773, opened 2017-05-08
Fix alert status handling in UI

fabxc:
This fixes various problems with the recent change of the alert status. Some of the problems were outlined in #770 

@raypettersen could you try this branch out and report whether it fixes things for you?


types/types.go
```diff
@@ -24,58 +23,19 @@ import (
...
 )
 
-type AlertState uint8
+type AlertState string
 
 const (
-	AlertStateUnprocessed AlertState = iota
-	AlertStateActive
-	AlertStateSuppressed
+	AlertStateUnprocessed AlertState = "unprocessed"
```
mxinden: Can *comparing strings* be a performance issue for us in the long run?

fabxc: With pretty high confidence, I'd say no here. This is in probably in the low nanosecond range.

brancz:
code lgtm and had a glance over the UI with @mxinden which also looked good

Issue #742, opened 2017-04-26
Alertpage group by


mxinden:
A user should be able to group all displayed alerts by a set of labels. The grouping should be flat, not nested.

Auto complete would be nice to have. Could be similarly realised like the already existing filter bar by @w0rm.

stuartnelson3:
So we were thinking of having this be a separate set of autocomplete boxes, correct? the user would type in, e.g., `service`, `region`, and then the alerts would be sorted as such? example heading for a group being `service=foo_service` `region=eu-west-1`?

mxinden:
@stuartnelson3 Yes, I would suggest a separate auto-complete box. When I add a new label via the input, alerts should be automatically grouped by it.

> example heading for a group being service=foo_service region=eu-west-1?

Yes

mxinden:
Closing here as #775 is merged.

PR #775, opened 2017-05-08
GroupBy alerts

stuartnelson3:
This is still in progress, but it's a pretty big PR so it would be good to get some eyes whenever.

Implements #742 

TODO:

- Add `enter` and scrolling support for auto-complete
- Clean up all the filtering and business logic in `ui/app/src/Views/AlertList/Views.elm`, add tests for it
- More testing in general, probably

MAYBE:

- Highlight portions of matching text? Don't think this adds a ton to the experience, though ..


ui/app/src/Views/AutoComplete/Match.elm
```diff
@@ -0,0 +1,139 @@
...
+    -- to type Int?
+    list
+        |> List.filter
+            (\( x, y ) ->
+                x >= p - far && x <= p + far && y == q
+            )
+
+
+transposition : List ( Int, Int ) -> ( Int, Int ) -> Int
+transposition list ( p, q ) =
```
stuartnelson3: This is copied from the haskell implementation, but looking at the wikipedia description, it appears the implementation is missing a final step:

```
The number of matching (but different sequence order) characters divided by 2 defines
the number of transpositions.
```

But then it also says

```
Although 'C', 'T' appear in both strings, they are farther than 1,
i.e., floor(5/2)-1=1. Therefore, t=0 . 
```

So they're using a definition of `t = (displaced characters in string) / 2` but then further qualifying it by saying for the displaced characters to be included in the above calculation, they have to be within `floor( len_of_longer_str / 2 ) - 1` of their correct position.

https://en.wikipedia.org/wiki/Jaro%E2%80%93Winkler_distance#Jaro_distance

stuartnelson3: Disregard -- We limit our search for transpositions `t` to the bounded matches in `m`.

mxinden:
@stuartnelson3 This looks very cool! If it's ok with you I will delay my review, as this is not part of the 0.7 milestone which I would like to focus on first to get the new UI into master as soon as possible.

w0rm:
@stuartnelson3 I added keyboard controls for the autocomplete, and changed the style a little so more items fit on the screen.

autocomplete

stuartnelson3:
@mxinden sure thing. But maybe it'll make it if I get some free time to finish this :)

mxinden:
This looks really good. Small Todo: When clicking on an autocomplete element it should be added like when pressing enter.

ui/app/src/Views/AlertList/Types.elm
```diff
@@ -4,22 +4,27 @@ import Utils.Types exposing (ApiData, ApiResponse(Loading))
...
 type AlertListMsg
     = AlertsFetched (ApiData (List Alert))
     | FetchAlerts
     | MsgForFilterBar FilterBar.Msg
+    | MsgForAutoComplete AutoComplete.Msg
 
 
 type alias Model =
     { alerts : ApiData (List Alert)
+    , autoComplete : AutoComplete.Model
```
mxinden: Shouldn't this rather be called `groupBar` (see `filterBar`)?

ui/app/src/Views/AutoComplete/Views.elm
```diff
@@ -0,0 +1,169 @@
...
+import Html exposing (Html, Attribute, div, span, input, text, button, i, small, ul, li, a)
+import Html.Attributes exposing (value, class, style, disabled, id, href)
+import Html.Events exposing (onClick, onInput, on, keyCode, onFocus, onBlur)
+import Set
+import Utils.List
+import Utils.Keyboard exposing (keys, onKeyUp, onKeyDown)
+
+
+view : Model -> Html Msg
+view { list, fieldText, fields, matches, maybeSelectedMatch, backspacePressed, focused } =
```
mxinden: Could we break this function down into couple smaller functions?

ui/app/src/Views/AutoComplete/Views.elm
```diff
@@ -0,0 +1,169 @@
+module Views.AutoComplete.Views exposing (view)
```
w0rm: I wouldn't nest more than 3 levels deep. I think it is fine right now. I only renamed `AutoComplete` to `GroupBar`.

ui/app/src/Views/AutoComplete/Views.elm
w0rm: Done!

ui/app/src/Views/AlertList/Types.elm
w0rm: Renamed!

w0rm:
@stuartnelson3 I had to rebase because of the conflicts.

w0rm:
I synchronised the group bar with the location and implemented the default grouping by alert name. I also updated the design of the alerts page, it looks a little cleaner now.

w0rm:
@stuartnelson3 @mxinden I'm satisfied with the result now, I think this is ready for the final review. Would really like this to be merged into master.

mxinden:
@w0rm @stuartnelson3 Cool, sounds good. I will have a look at it tomorrow morning. Maybe we can merge this into ui-rewrite before merging #789. Whatever you guys prefer.

Sorry for all the spamming through changing the destiny branch. Forgot that #789 is not yet merged.

w0rm:
@mxinden I told a few colleagues about the UI rewrite and they were pretty excited to try it.

ui/app/src/Views/AlertList/Updates.elm
```diff
@@ -1,33 +1,76 @@
...
+                    GroupBar.setFields filter groupBar
+
+                newFilterBar =
+                    FilterBar.setMatchers filter filterBar
+
+                newModel =
+                    { model | filterBar = newFilterBar, groupBar = newGroupBar }
+            in
+                -- only fetch when filter changed
+                if newFilterBar /= model.filterBar || model.alerts == Initial then
```
stuartnelson3: This appears to be preventing my UI from re-querying for alerts when I remove a label matcher.

Trying to be efficient is cool, but I really don't think it's needed here. The actual load on any one alertmanager should be pretty small, and users should be round-robin'ing through their HA alertmanager instances. It seems that this new `Initial` type exists for the same reason? I would vote for removing this as I don't think the small increase in performance is worth adding the extra type and then having to worry about the sequence of events leading to the `FetchAlerts` message, as it's already caused a bug.

ui/app/src/Views/GroupBar/Updates.elm
```diff
@@ -0,0 +1,117 @@
...
+            else
+                -- TODO: Disallow adding spaces, or only check distance if
+                -- there are no spaces.
+                -- TODO: How many matches do we want to show?
+                -- NOTE: List.reverse is used because our scale is (0.0, 1.0),
+                -- but we want the higher values to be in the front of the
+                -- list.
+                Set.toList model.list
+                    |> List.filter ((flip List.member model.fields) >> not)
+                    |> List.sortBy (jaroWinkler model.fieldText)
```
stuartnelson3: I'm curious how this performs on an alertmanager that has a large set of possible labels.

ui/app/src/Views/AlertList/Views.elm
```diff
@@ -3,42 +3,139 @@ module Views.AlertList.Views exposing (view)
...
+                                            Just (alertList k alts filter)
+
+                                        Nothing ->
+                                            Nothing
+                            )
+                            (Dict.keys g
+                                |> List.partition ((/=) [])
+                                |> uncurry (++)
+                            )
+                        )
```
stuartnelson3: There's a lot going on here, would be nice to break it down into named functions so it's a bit easier to read and reason about.

stuartnelson3:
Would love to get this in, but there's a bug that needs fixing first.

I would also like to make the styling on silences and alerts a bit more similar before we release. I like the direction @w0rm is going in, but remembering to keep it compact is important. I think we're on the right track, and a lot of the requests for a "more compact" view are really just because there was no good filtering previously, but making sure it stays easy to scan over a large number of alerts is still important.

Also -- it seems like the silence view page lost the ability to edit or expire? What was the motivation for this? I think it's important to be able to hold on a link to a silence so that a user can expire/modify it without having to search through the whole silence list. Not sure where it was introduced, just noticing it now. There's also an inconsistency in displaying matchers (`instance="ip-1-2-3-4.com"`) and labels (`instance=ip-1-2-3-4.com`).

mxinden:
@stuartnelson3 Not being able to expire/edit a silence on the *SilenceView* page was introduced by me. Sorry about that. I have created an issue to get that back in (#794).

ui/app/src/Utils/Views.elm
```diff
@@ -8,18 +8,12 @@ import Http exposing (Error(..))
...
-            [ span [ class " badge badge-warning" ]
-                [ i [] [], text labelText ]
-            ]
-    in
-        case maybeMsg of
-            Nothing ->
-                span [ class "pl-2" ] label
+    case maybeMsg of
+        Nothing ->
+            Html.button [ class "btn btn-sm btn-outline-info ml-2 mb-2", disabled True ] [ text labelText ]
```
mxinden: `mb-2` makes `labelButtons` in the *SilenceList* view vertically off a little from the center.

w0rm:
@stuartnelson3 I think it's  pretty compact and scannable, but also clean and lightweight. The default grouping pulls out the alert names, but even if this grouping is removed, they are the first in the matchers list

ui/app/src/Utils/Views.elm
w0rm: Oh, sorry! I will put it back.  I could not check the silences because the endpoint that I'm using to test everything doesn't work with our new frontend

mxinden: No problem!

You can use the `./examples/ha/send_alerts.sh` script to generate sample alerts and I usually just create a couple of silences myself.

mxinden:
@brancz @fabxc liked the new design of the `AlertList` view as well. Only thing we would suggest visually separating each alert more e.g. with a line in between?

Other than that only blocking issue that I see right now is that the list of alerts is not refreshed when removing a filter, like described by @stuartnelson3 in this [comment](https://github.com/prometheus/alertmanager/pull/775#pullrequestreview-38020892).

It would be great if you could style the `SilenceList` view accordingly as well. But I would suggest to do this in a separate PR as this one is already pretty big.

brancz:
@mxinden sgtm

ui/app/src/Views/GroupBar/Updates.elm
w0rm: I tried it on the SC data, worked fine.

mxinden: Might be a valid concern. But I would push back the performance improvement until we actually face any performance issues.

ui/app/src/Utils/Views.elm
w0rm: @mxinden I rolled this back

w0rm:
@stuartnelson3 I changed the code to always fetch the alerts, but I want to keep the `Initial` state, because of my experience we may need it in the future. This is `NotAsked` state from http://blog.jenkster.com/2016/06/how-elm-slays-a-ui-antipattern.html

mxinden:
@w0rm As far as I can tell the `silenced` url query parameter is not yet implemented in the `/alerts` endpoint. Could we hide the checkbox for now and unhide it as soon as this is implemented on backend side?

ui/app/src/Views/AlertList/Views.elm
mxinden: 👍 to breaking this into smaller functions

ui/app/src/Views/GroupBar/Match.elm
```diff
@@ -0,0 +1,150 @@
...
+                    y :: ys ->
+                        if x == y then
+                            x :: cp xs ys acc
+                        else
+                            acc
+
+
+charMatch : Int -> List ( Int, Int ) -> ( Int, Int ) -> List ( Int, Int )
+charMatch matchRange list ( p, q ) =
+    -- TODO(w0rm): Is there a way to define this so it's not so strictly bound
```
mxinden: @w0rm Fyi, in case you haven't seen it.

ui/app/tests/Tests.elm
```diff
@@ -16,3 +18,10 @@ utils =
     describe "Utils"
         [ Filter.all
         ]
+
+
+match : Test
```
mxinden: For me this would rather be called `groupBar`. If you compare it to the example above: `Utils` is the folder and `Filter` is the actual file. Not very important.

stuartnelson3:
I would prefer to keep the silenced checkbox, since this is already implemented in the client side, and then we can just swap it out when we get to implementing it on the server side.

stuartnelson3:
Big thanks to @w0rm for pushing this through, I've been busy lately, and lack his awesome experience with elm and front-end design/ux. kudos!

ui/app/src/Views/GroupBar/Match.elm
stuartnelson3: It might not be important, but being able to use a generic comparable number would be good. I think I tried using number initially, but then it needed to be comparable, and I didn't know how to say `charMatch : <number, comparable>` etc etc

stuartnelson3:
And: It's not a huge burden to have `Initial`, and I remember removing that from our implementation of remote data because I had no reason to differentiate that state from `Loading`, but it's fine to leave it.

So, from my side, as soon as we get the silences page to look like the Alerts page, and maybe adding a light-gray 1px line between alerts, we're good to go!

ui/app/src/Views/AlertList/Views.elm
w0rm: Done

ui/app/tests/Tests.elm
w0rm: Renamed

ui/app/src/Views/GroupBar/Match.elm
w0rm: @stuartnelson3 still not sure what you want, but `number` is `comparable`, so I replaced `Int` with `number`. I also simplified the code a bit.

w0rm:
@stuartnelson3 I restored `showSilenced` checkbox

w0rm:
@mxinden @stuartnelson3 I addressed your comments (one was contradicting, so I chose to keep the checkbox) and made the ci green (had to bump my local elm-format, and compiled the bindata using the docker image instead of running `make assets` cause they produced different results).

PR #779, opened 2017-05-09
Sync master

mxinden:


stuartnelson3:
can you try rebasing? makes for a cleaner history.

RSamal:
Hi , on the create silencer UI , why I am seeing a different time zone. I am at PST hours but seeing the time as  2017-05-23T03:31:31.564Z for 2017-05-22T8:31.

Also if I have to call the API directly, what time format should I provide for thr startsAt value?

fabxc:
When talking to the API you can provide any RFC3339 timestamp, which
encodes the timezone in it.
The new UI shows everything in UTC. There are plans to point this out
explicitly. Any local time zone support will only be added after careful
consideration.

PR #780, opened 2017-05-09
Update mesh and add it to debug logging

fabxc:
A general update that's probably in order. Also mesh logs are now emitted at the debug level.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: false
 language: go
 
 go:
-- 1.7.3
+- 1.8.1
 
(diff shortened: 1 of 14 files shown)


stuartnelson3:
Looks cool once the build passes

fabxc:
Ah, tests are failing because the newest mesh version uses `time.Until` which is new in stdlib. Travis still runs stuff against 1.7. As there's really no reason to use old Go versions, I'll remove it from Travis and 1.8+ is then a requirement for building Alertmanager.

Issue #745, opened 2017-04-26
Silencepage quick filter by pending / active / expired


mxinden:
A user should be able to quickly filter all displayed silences by pending / active / expired ones with check boxes. The check boxes could be placed underneath the filter bar.

Style should be  similar to #688

mxinden:
@fabxc What do you think of returning pending / active / expired from the back-end instead of determining this on the client with timestamps? Would prevent time zone issues.  Either one is fine with me.

fabxc:
I'd not group them from the API, but attaching a status on pending/active/expired like we added for alerts would be good, yes.

We could follow the same pattern there of

```"status": {
  "state": <expired|pending|active>,
  "silences": [<alert ID/FP>, ...]
}
```

The latter can be added later. Just so we make `status` an object to begin with.

stuartnelson3:
+1 for setting state in the backend and passing it to the front end.

Regarding `suppresses` vs. `silences`: We use `suppressed` as an AlertState, but that struct still has a `silencedBy` field. I would say we stick with `silences` as the field name since that's the specific type of `suppressed` we're dealing with.

mxinden:
@stuartnelson3 @w0rm I am currently working on this. I will give the silence page a little bit of styling as well.

mxinden:
Closing here. #785 does not include the checkboxes but everything else. Let's see if we still need the filtering or if a hide/show button makes more sense.

PR #785, opened 2017-05-10
Silence State from API and UI

mxinden:
* Return Silence state on `/silences`
  silenceState = "expired" | "active" | "pending"
  ```
  "status": {
    "state": "expired"
  }
  ```
* Restyle SilenceList view
  * Group by Silence state
  * Adjust styling to AlertList view
* Adjust Silence view

---

screen shot 2017-05-10 at 16 32 49

Fixes issue #745 

This does not introduce the quick filter via three check-boxes. I would suggest that we don't really need those anymore? What do you think?


mxinden:
As this is the last issue in the milestone 4, we could merge the new UI into master once this is merged.
=> No more syncing between `master` and `ui-rewrite`

ui/app/src/Utils/String.elm
```diff
@@ -0,0 +1,19 @@
...
+    else
+        cons (firstChar string |> Char.toUpper) (slice 1 (length string) string)
+
+firstChar : String -> Char
+firstChar string =
+    case uncons string of
+        Nothing ->
+            ' '
+        Just (char, _) -> char
+
```
w0rm: Why not use `uncons` directly in `capitalizeFirst`?

```
capitalizeFirst : String -> String
capitalizeFirst string =
    case String.uncons string of
        Nothing ->
            string

        Just ( char, rest ) ->
            String.cons (Char.toUpper char) rest
```

mxinden: Thanks, that looks better.

ui/app/src/Views/Shared/SilenceBase.elm
```diff
@@ -63,3 +68,34 @@ matcherButton matcher =
...
+            String.join "/" [ "#/silences", silenceId, "edit" ]
+    in
+        a [ class "h-100 btn btn-success rounded-0", href editUrl ]
+            [ span [ class "fa fa-pencil" ] [] ]
+
+
+deleteButton : Silence -> Html Msg
+deleteButton silence =
+    if silence.status.state == "expired" then
+        i [] []
```
w0rm: Maybe change this to `text ""` or make all buttons return `Maybe (Html Msg)`, then you could render them with `List.filterMap identity maybeButtons`

ui/app/src/Views/SilenceList/Views.elm
```diff
@@ -1,62 +1,76 @@
...
-silences : List Silence -> FilterBarTypes.Model -> Html Msg -> Html Msg
-silences silences filterBar errorHtml =
+silenceGroupView : ( String, List Silence ) -> Html Msg
+silenceGroupView ( state, silences ) =
     let
-        html =
+        silencesView =
             if List.isEmpty silences then
-                div [ class "mt2" ] [ text "no silences found" ]
+                div [ class "" ] [ text "No silences found" ]
```
w0rm: Does it really need an empty class?

ui/app/src/Views/SilenceList/Views.elm
```diff
@@ -1,62 +1,76 @@
...
+
+
+states : List String
+states =
+    [ "active", "pending", "expired" ]
+
+
+filterSilencesByState : List Silence -> String -> List Silence
+filterSilencesByState silences state =
+    List.filter (\silence -> silence.status.state == state) silences
```
w0rm: May be written shorter `List.filter (\{status} -> status.state == state) silences`

w0rm: Or even shorter `List.filter (.status >> .state >> (==) state) silences`

w0rm: Or, you can change the order of the arguments,  which I think makes sense for the filter function. In this case you don't need to specify `silences`

```
filterSilencesByState: String -> List Silence ->  List Silence
filterSilencesByState state =
    List.filter (\{status} -> status.state == state)
```

ui/app/src/Views/SilenceList/Views.elm
mxinden: Good catch, thanks!

ui/app/src/Silences/Types.elm
```diff
@@ -37,6 +44,12 @@ type alias Silence =
     , updatedAt : Time
     , matchers : List Matcher
     , silencedAlerts : ApiData (List Alert)
+    , status : SilenceStatus
+    }
+
+
+type alias SilenceStatus =
```
stuartnelson3: Since this is already in `module Silences.Types`, I would prefer it just be `Status` to avoid some stutter.

stuartnelson3: Actually, looking at this, is there a reason it isn't a union type?

```elm
type Status
    = Active
    | Active
    | Pending
```

I think I would prefer this, as it saves us from having to do string matching later to check the status.

ui/app/src/Views/SilenceList/Updates.elm
```diff
@@ -28,13 +28,17 @@ update msg model silence filter =
...
-            -- TODO: Add DELETE to accepted CORS methods in alertmanager
-            -- TODO: Check why POST isn't there but is accepted
-            ( model, Loading, Navigation.newUrl "/#/silences" )
+        SilenceDestroyed statusCode ->
+            let
+                status =
+                    Debug.log "statusCode" statusCode
+            in
+                -- TODO: "Deleted id: ID" growl
+                -- TODO: Add DELETE to accepted CORS methods in alertmanager
```
stuartnelson3: `DELETE` has been added, so this can be deleted

ui/app/src/Views/SilenceList/Views.elm
```diff
@@ -1,62 +1,76 @@
...
+
+
+states : List String
+states =
+    [ "active", "pending", "expired" ]
+
+
+filterSilencesByState : String -> List Silence -> List Silence
+filterSilencesByState state =
+    List.filter (\{ status } -> status.state == state)
```
stuartnelson3: This has a big-O value of `O(n^x)`, where `x == len(states)`. Depending on the number of silences this could be pretty painful. Could you either make a note to replace it with `groupBy` (`O(n)`) in this pr https://github.com/prometheus/alertmanager/pull/775/files#diff-4b0e2e6b80ca6bbf40b7a26e66b231b9R76, or just port it over and make a note to make sure we only have one of them when the above PR gets merged?

ui/app/src/Silences/Types.elm
mxinden: Haven't thought of that. Sounds good. Changed 👍

ui/app/src/Views/SilenceList/Views.elm
mxinden: `groupBy` looks a lot better here, thanks. I will add a note so we can replace it once #775 is merged.

Btw: pretty sure it is `O(x*n)` which equals to `O(n)`, but still a good improvement 👍

mxinden:
@stuartnelson3 @w0rm I have addressed all of your comments, thanks for the review! Any other wanted changes?

@fabxc Do you agree with the changes on the `golang` side?

ui/app/src/Utils/String.elm
```diff
@@ -0,0 +1,14 @@
+module Utils.String exposing (capitalizeFirst)
+
+import String exposing (left, length, slice, cons, uncons, trim)
```
w0rm: If you're using `String.uncons` etc.,  then there is no need to import it in the scope.

mxinden: Addressed, thanks.

fabxc:
Yea, we can roll with this for now. At some point we might want to at least add a toggle/hide button to each group.

mxinden:
@stuartnelson3 Triggered [Circle CI](https://circleci.com/gh/prometheus/alertmanager/749)

@all Thanks for the review.

PR #786, opened 2017-05-10
Expose alert fingerprint in the API

prymitive:
Alert fingerprint is already provided as the value of status.inhibitedBy[] attribute that inhibited alerts have, but there's no way to get back to the alert that's inhibiting it as the fingerprint is not exposed.

I'm not sure if using fingerprints is desirable, but that's currently the value of `inhibitedBy` @stuartnelson3 
The goal here is to be able to link back from inhibited to inhibiting alert based on the API response, which isn't currently doable.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -346,7 +346,7 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 			Alert:       &a.Alert,
 			Status:      status,
 			Receivers:   receivers,
-			Fingerprint: fmt.Sprintf("%x", a.Fingerprint()),
+			Fingerprint: a.Fingerprint().String(),
 		}
 
 		res = append(res, apiAlert)
--- a/api/api.go
+++ b/api/api.go
@@ -346,6 +346,7 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 			Alert:     &a.Alert,
 			Status:    status,
 			Receivers: receivers,
+			ID:        fmt.Sprintf("%x", a.Fingerprint()),
 		}
 
 		res = append(res, apiAlert)
--- a/api/api.go
+++ b/api/api.go
@@ -343,10 +343,10 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 		}
 
 		apiAlert := &dispatch.APIAlert{
-			Alert:     &a.Alert,
-			Status:    status,
-			Receivers: receivers,
-			ID:        fmt.Sprintf("%x", a.Fingerprint()),
+			Alert:       &a.Alert,
+			Status:      status,
+			Receivers:   receivers,
+			Fingerprint: fmt.Sprintf("%x", a.Fingerprint()),
 		}
 
 		res = append(res, apiAlert)
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -80,9 +80,9 @@ type AlertBlock struct {
 // annotated with silencing and inhibition info.
 type APIAlert struct {
 	*model.Alert
-	Status    types.AlertStatus `json:"status"`
-	Receivers []string          `json:"receivers"`
-	ID        string            `json:"id"`
+	Status      types.AlertStatus `json:"status"`
+	Receivers   []string          `json:"receivers"`
+	Fingerprint string            `json:"fingerprint"`
 }
 
 // AlertGroup is a list of alert blocks grouped by the same label set.
@@ -137,9 +137,9 @@ func (d *Dispatcher) Groups(matchers []*labels.Matcher) AlertOverview {
 				}
 				status := d.marker.Status(a.Fingerprint())
 				aa := &APIAlert{
-					Alert:  a,
-					Status: status,
-					ID:     fmt.Sprintf("%d", a.Fingerprint()),
+					Alert:       a,
+					Status:      status,
+					Fingerprint: fmt.Sprintf("%x", a.Fingerprint()),
 				}
 
 				if !matchesFilterLabels(aa, matchers) {
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -139,7 +139,7 @@ func (d *Dispatcher) Groups(matchers []*labels.Matcher) AlertOverview {
 				aa := &APIAlert{
 					Alert:       a,
 					Status:      status,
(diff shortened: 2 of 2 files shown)


stuartnelson3:
I think fingerprints are the method of identification we're going to be using for now, but with this method you're required to filter through all the alerts being sent from the current api endpoint. I think it would make more sense to have a new endpoint added, like `/api/v0/alerts/abc-fingerprint-123`, and then maintain a `map[Fingerprint]Alert` to query against. This map would have to have access coordinated to it for additions and deletions.

@fabxc is this new endpoing something that should be added?

prymitive:
Is there a reason to hide fingerprints in the `/api/v1/alerts/groups` endpoint?
A dedicated endpoint to get an alert by id seems like a good idea, but you should still be able to get an id of any alert, not just those that are inhibiting others.
Use case example: I have an alert, I want to see if it's inhibiting any other alert - for that I need to know the id.

stuartnelson3:
Makes sense to me, but waiting on a response from @fabxc

fabxc:
Probably makes sense to add them to our API. We can make individual alerts gettable once there's a use case.

prymitive:
Added to the list endpoint

api/api.go
```diff
@@ -314,6 +315,7 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 		apiAlert := &APIAlert{
 			Alert:  &a.Alert,
 			Status: status,
+			ID:     fmt.Sprintf("%d", a.Fingerprint()),
```
mxinden: Could we print this not as base10 but as base16 with `%x`?

dispatch/dispatch.go
```diff
@@ -81,6 +81,7 @@ type AlertBlock struct {
 type APIAlert struct {
 	*model.Alert
 	Status types.AlertStatus `json:"status"`
+	ID     string            `json:"id"`
```
mxinden: As the fingerprint is not really an id, I would prefer renaming this to `HASH`. What do you think?

stuartnelson3: I'm not a fan of hash since that's naming it after its implementation. I'm fine with `id` for now since the fingerprint is how we're identifying the alerts.

api/api.go
stuartnelson3: Is this to reduce the length of the ID?

mxinden: No, it's just the `base 16, with lower-case letters for a-f`. See https://golang.org/pkg/fmt/#hdr-Printing

dispatch/dispatch.go
mxinden: I am not sure if fingerprints are entirely unique, so thereby `id` might be misleading. @fabxc What do you think?

brancz: I do not think `id` is the right naming. An `id` would give it the notion of uniqueness and uniqueness over time. An alert can appear/disappear and even alerting rules get changed with or without the labels of the alert changing.

Since we all refer to it as this, why not `fingerprint`?

stuartnelson3: fingerprint is fine by me

api/api.go
stuartnelson3: Right, I'm just wondering why you want base 16 instead of base 10

mxinden: I think this was a suggestion by @fabxc. @fabxc can you answer this?

mxinden: I am sorry for the wrong answer here @stuartnelson3. Yes, changing to base16 is due to reducing the length. In addition we prevent someone of interpreting the fingerprint as an increasing integer, and we enforce exposing the fingerprint as a string in a JSON data structure.

mxinden:
@prymitive Do you have some time to finish this PR? In case you are too busy I can take over.

prymitive:
I should have, I just need to remind myself what needs to be done here, will update soon

prymitive:
Renamed `ID`->`Fingerprint` and using `%x` for printing instead of `%d`. I think that's all that was requested here

api/api.go
```diff
@@ -343,9 +343,10 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 		}
 
 		apiAlert := &dispatch.APIAlert{
-			Alert:     &a.Alert,
-			Status:    status,
-			Receivers: receivers,
+			Alert:       &a.Alert,
+			Status:      status,
+			Receivers:   receivers,
+			Fingerprint: fmt.Sprintf("%x", a.Fingerprint()),
```
mxinden: @prymitive When I curl Alertmanager it does not expose the Fingerprint base16 encoded. I looked through the `common` package and found the  [`String`](https://github.com/prometheus/alertmanager/blob/master/vendor/github.com/prometheus/common/model/fingerprinting.go#L40) function. Would you mind changing it again to `Fingerprint: a.Fingerprint().String(),` then we are consistent across projects. I know there are plans to get rid of the `commons` repo in the future, but as we are using `Fingerprint` from the repo anyways, we might as well use `Fingerprint.String()`. Let me know if this sounds reasonable to you.

//cc @fabxc

dispatch/dispatch.go
```diff
@@ -136,8 +137,9 @@ func (d *Dispatcher) Groups(matchers []*labels.Matcher) AlertOverview {
 				}
 				status := d.marker.Status(a.Fingerprint())
 				aa := &APIAlert{
-					Alert:  a,
-					Status: status,
+					Alert:       a,
+					Status:      status,
+					Fingerprint: fmt.Sprintf("%x", a.Fingerprint()),
```
mxinden: See comment in `api/api.go` in regards to using `Fingerprint.String()`.

api/api.go
prymitive: That makes sense, thanks

`"fingerprint": "6f02aa263df152bc"`

prymitive:
Anything else left here?

prymitive:
Should I squash those commits?

mxinden:
@prymitive No need to squash, can do so via Github. What is the change in the `template/internal/deftmpl/bindata.go` in your last commit?

prymitive:
Not sure, probably some dangling changes from building alertmanager locally for testing. Will nuke it.

prymitive:
It's gone now

mxinden:
@stuartnelson3  concern is addressed. I will merge here. Thanks @prymitive for the contribution and for the quick changes.

PR #668, opened 2017-03-24
Add Elm UI rewrite from github.com/stuartnelson3/am-ui with history to "ui-rewrite" branch

mxinden:
There has been some discussions around rewriting the Alertmanager and Prometheus UI. @stuartnelson3 , @w0rm  and I have recently written a prototype for the Alertmanager UI in Elm. We would like the code to live in the branch `ui-rewrite` until it is feature equal to the current Alertmanager UI. See further information in the [Google Group Topic](https://groups.google.com/forum/#!topic/prometheus-developers/gJ0eLtweJIM).

* Including glue between Alertmanager server and Elm UI.
* Rebuilding assets (ui/bindata.go).
* Exact commit: 'bd78de6b16bcefaacf4229304d439b33aa09cc72'
* Subdirectory: ui/app

PR #789, opened 2017-05-12
Rewrite UI in Elm

mxinden:
As milestone [0.7](https://github.com/prometheus/alertmanager/milestone/4) is done now I would like to propose merging the new ELM-UI into `master`. For further details please see PR #668.

There are still a lot of steps to take, but as far as I can tell it is already a good replacement for the existing Angular-UI.

What do you think? Happy for any feedback.


brancz:
I spun up an instance, and all looks good to me. One thing that was sort of unexpected is that I had to do a hard refresh because I had previously had a 0.6.2 Alertmanager instance on the same host port combination. As this is a one time thing it's fine, just thought we may want to include it in the release notes.

brancz:
The remaining conflicts are mainly generated and deleted files, so I'd say we manually resolve on merge. lgtm

stuartnelson3:
Assuming the compiled binary runs, I'm good with this. I think from an operational perspective one of the things I would like to do is move compilation + embedding the assets in a go file into a CI step, and not rely on us remembering to do that. But that doesn't prevent this from going out!

:rocket:

mxinden:
@stuartnelson3 I have a PR coming for improving the build process.

fabxc:
Needs a rebase.

brancz:
:tada:

PR #799, opened 2017-05-16
Add 'silenced' query param to /alerts endpoint

### LINKED ISSUE #797: Implement silenced query parameter in /alerts endpoint
The 'silenced' url query parameter is not yet implemented in the /alerts endpoint. We already show a checkbox on the AlertList view and change the query parameter accordingly, but the back-end ignores it.

mxinden:
This adds the `silenced` query parameter to the `/alerts` API endpoint.

@stuartnelson3 @w0rm In addition it changes the default in the UI. Previously silenced alerts were displayed by default. This PR hides silenced alerts by default. @fabxc and I thougth this would improve the UX as most users are probably looking for firing alerts. What do you think? Is the way I implemented it in `AlertList/Parsing.elm` ok?

Fixes #797


stuartnelson3:
I'm cool with initially hiding silenced alerts

api/api.go
```diff
@@ -271,6 +272,21 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	if silencedParam := r.FormValue("silenced"); silencedParam != "" {
```
stuartnelson3: You could potentially use `strconv.ParseBool`, but I would ask for @fabxc 's preference here, as I'm not sure what's common in the go community. It's a pretty easy check when it's a string.

api/api.go
```diff
@@ -286,6 +302,10 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 
 		status := api.getAlertStatus(a.Fingerprint())
 
+		if !showSilenced && len(status.SilencedBy) != 0 {
```
stuartnelson3: This reminds me that we need to add some indication in the UI if an alert is inhibited

ui/app/src/Views/AlertList/Parsing.elm
```diff
@@ -4,19 +4,23 @@ import UrlParser exposing ((</>), (<?>), Parser, int, map, oneOf, parseHash, s,
 import Utils.Filter exposing (Filter, parseMatcher, MatchOperator(RegexMatch))
 
 
-boolParam : String -> UrlParser.QueryParser (Maybe Bool -> a) a
-boolParam name =
+parseSilenced : String -> UrlParser.QueryParser (Maybe Bool -> a) a
+parseSilenced name =
```
stuartnelson3: I'm a bit conflicted on this.

On one hand, this is the only place it's being used, so having it be specifically for parsing the silenced param is fine.

On the other hand we had a generic bool parser that we could then just throw a `(parseSilenced >> Maybe.withDefault (Just False)) silencedParam` at (assuming this compiles, haven't tried it :))

Not sure. I'll take either, up to you.

ui/app/src/Views/AlertList/Parsing.elm
```diff
@@ -4,19 +4,23 @@ import UrlParser exposing ((</>), (<?>), Parser, int, map, oneOf, parseHash, s,
...
     UrlParser.customParam name
         (\x ->
             case x of
                 Nothing ->
-                    Nothing
```
w0rm: I think it is fine to have `Nothing` here to indicate that it was unset. Then we can later decide on what does it mean. Otherwise, it shouldn't be a `Maybe bool`.

api/api.go
fabxc: We'd commit to accepting all true/false variations `ParseBool` provides. Keeping it simple for now is a safe bet.

api/api.go
fabxc: Yea, practically, one probably just wants to hide everything non-firing. But lets stick to silencing only for now. Few people are using inhibition anyway.

We should definitely have an inhibition indicator in the UI for feature parity.

mxinden: For documentation: Linking #801 here.

mxinden:
@stuartnelson3 @w0rm I have addressed your comments. Thanks for the input. I have reverted the changes on the `boolParam` parser and instead added it to the `generateQueryString` function.

In the long run, what do you think of the idea of having the `filter` not in the global modal but in the model of each view. E.g. the `SilenceView` does not need the `showSilenced` parameter in it's filter object. Sorry if I brought this up before.

ui/app/src/Utils/Filter.elm
```diff
@@ -48,7 +48,7 @@ generateQueryString { receiver, showSilenced, text, group } =
     let
         -- TODO: Re-add receiver once it is parsed on the server side.
         parts =
-            [ ( "silenced", Maybe.map (toString >> String.toLower) showSilenced )
+            [ ( "silenced", Maybe.withDefault False showSilenced |> toString |> String.toLower |> Just )
```
stuartnelson3: Is there a reason to force this to be `Just True` or `Just False`? The server code checks to see if the query string exists, so I don't think there's anything wrong with this being `Nothing` and just having it drop out from the api call.

mxinden: If it is `Nothing` we will not add the `silenced` query parameter to the API call. If we don't send the `silenced` parameter to the API, the back-end will send all alerts (silenced & !silenced).

When the user opens up Alertmanager without any parameters, `filter/showSilenced` will be `Nothing`. We want to hide silenced alerts by default, so the `generateQueryString` has to add the `silenced=false` query parameter in our case, where `showSilenced` is `Nothing`.

On the `/alerts` API endpoint returning all alerts should be the default.
On the `/alerts` UI page displaying only !silenced alerts should be the default.

Happy for any feedback or alternative solutions.

stuartnelson3: Ah, k. So it's because the default behavior is to send everything. That's fine, no worries.

w0rm:
Also, if the alerts are filtered on the backend, should we remove the [filtering on the frontend](https://github.com/mxinden/alertmanager/blob/9d2c113c2eb1d05a1026752bb164e6db2f0527cb/ui/app/src/Views/AlertList/Views.elm#L97)? cc @stuartnelson3

stuartnelson3:
Good point -- that filtering code still exists. Let's get rid of that.

mxinden:
@stuartnelson3 @w0rm Removed the browser side filtering code.

Issue #800, opened 2017-05-16
Filtering alerts by receivers


w0rm:
Alerts could be filtered by receivers in the old UI. I think receivers make sense, from a perspective of being on call, I can quickly get all alerts that I am responsible for, without having to memorize all the labels and manually entering them into the filter bar.

It also allows me to filter out the alerts I am responsible for, even if they have totally different labels, so it is some kind of `OR` operation for the matchers that is impossible with the new filter bar.

We stopped filtering by receivers when we switched away from grouped endpoint to the `/alerts` endpoint. Since this affects the way the old UI is used at the moment, I wonder what is the plan for this feature or is there something that we can offer instead?

stuartnelson3:
I remember we discussed moving away from receivers the last time we were all together talking about this, but I don't remember the reasoning any more. I think the main point here that makes sense is being able to easily filter alerts with disparate labelsets but that I (as an operator) am interested in because I'm on-call.

What was the reason for getting rid of receivers? @fabxc @mxinden @brancz

fabxc:
It's a consequence of switching to a flat view of all alerts.
We can add it back later. Probably by annotating alerts with this information or correlating it with the notification log.

fabxc:
I'm somewhat worried about an explosion of options for views and clicking around them will not be really good UX.

Textual options were brought up in the past. I think that would be a good idea here.
Can we maybe filter by `silenced:true`, `silenced:false`, and `receiver:<name>` inline with the label selectors? Potentially even `group(a,b)` later.
Less clicky-pointy but not really complex and more powerful once learned.

brancz:
Yes, I think this is very powerful. I imagine it like GitHub style issue/PR search.

w0rm:
Can we support the `receiver` param in the `/alerts` endpoint, and add another endpoint to get the list of receivers for the dropdown in the UI?

beorn7:
Talked to @w0rm about it. Conclusion was to have an endpoint to find out about all configured receivers (for a UI component to select them).
The alerts endpoint needs an additional parameter, which is a regexp of the receiver (so you can match things like `teamX.*`, which would include `teamX-warn`, `teamX-crit`, and `teamX-info`, or lists, e.g. `foo|bar|baz`).

It would then be nice to give each alert a field `receivers` with a list of receivers this alert goes to. This could be displayed in the _Info_ section to easily see who got this alert.

In ever later iterations of the UI, each receiver could be linked to information how this receiver is configured.

[...]

PR #803, opened 2017-05-17
Remove receiver filtering from backlink

fabxc:
As it's currently not supported, we can remove it.
One could probably add label matchers there, but I'd first just see how people end up using the new structure.


mxinden:
Linking issue #800 here for documentation.

👍  Looks good to me.

PR #804, opened 2017-05-18
updated logrus so alertmanager can build on Solaris/Illumos

fazalmajid:
Same deal as https://github.com/prometheus/prometheus/pull/2733 but for alertmanager this time.

### DIFF
--- a/vendor/github.com/Sirupsen/logrus/CHANGELOG.md
+++ b/vendor/github.com/Sirupsen/logrus/CHANGELOG.md
@@ -1,3 +1,50 @@
+# 0.11.5
+
+* feature: add writer and writerlevel to entry (#372)
+
+# 0.11.4
+
+* bug: fix undefined variable on solaris (#493)
+
+# 0.11.3
+
+* formatter: configure quoting of empty values (#484)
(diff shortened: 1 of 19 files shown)


fazalmajid:
Unfortunately that is not enough, as github.com/stuartnelson2/guac depends on github.com/howeyc/fsnotify rather than its newer fork github.com/fsnotify/fsnotify (which is a candidate for eventual inclusion in the Go standard library, but makes incompatible API changes). howeyc does not support Solaris event ports/file event notifications but fsnotify does.

I created an upstream PR with stuartnelson3 to integrate the new fsnotify: https://github.com/stuartnelson3/guac/pull/1

stuartnelson3:
The guac code isn't required for anything necessary in alertmanager. I'm surprised it's being checked since it's in a nested `vendor/` folder. You can just delete `ui/app/vendor` and `ui/app/main.go` if it's for some reason showing up in your compilation ...

fazalmajid:
It's probably because I was building using;

```
env PATH=${PATH}:/usr/local/go/bin GOPATH=`pwd` GO15VENDOREXPERIMENT=1 go get -f -t -u -v github.com/prometheus/alertmanager/...
```

Issue #798, opened 2017-05-16
Rework SilenceForm


mxinden:
@fabxc and I went over some improvements needed in the *SilenceForm* before v0.7:

* ~~Style SilenceForm and remove *tachyons* classes.~~
* ~~Show an initial matcher example~~
* ~~Instead of editing an expired silence, the UI should open the "new silence" dialog with the matchers from the old silence instead~~
* Make form without matchers invalid. If no matchers are specified "Create silence" button should be deactivated.

stuartnelson3:
I would add that we should have some visual indication that the creator and comment field need to be filled in. Currently, a silence cannot be created without those being filled in (in the UI), but some indication would clarify that.

w0rm:
@mxinden Editing duration puts the wrong value in the end time field

![screen shot 2017-05-19 at 19 57 07](https://cloud.githubusercontent.com/assets/43472/26260771/c6296a16-3cce-11e7-8d93-648a593870b0.png)
 and end time broke.

w0rm:
@mxinden If you don't mind I can fix this and maybe address the initial validation state. 

`type alias ValidatedField a = { value: String, validation: Result String a }` instead of `Result (String, String) (String, a)`

mxinden:
Sounds like a good idea. Thanks, and sorry for introducing the bug.

w0rm:
@mxinden it's fine. I could've probably noticed this earlier! My plan is to complete this early next week, so that I can demo the new UI at work on Friday.

mxinden:
Updated issue. Most things are fixed by #806 & #827. Only things missing is:

> Make form without matchers invalid. If no matchers are specified "Create silence" button should be deactivated.

w0rm:
@mxinden it is impossible to have a form without matchers now, so this is fixed.

Issue #805, opened 2017-05-18
Duplicate alerts in list


ipstatic:
When testing 0.6.2 we noticed duplicate alerts showing in the listing when more than one receiver is configured. Using Git bisect, it appears to be introduced in this [commit](https://github.com/prometheus/alertmanager/commit/1e34f29532a251b728ddf304ebccd9234fcbd81a). If you run the same configuration before this commit, you do not see the duplication issue.

![screen shot 2017-05-18 at 10 17 11 am](https://cloud.githubusercontent.com/assets/332617/26208478/2f337974-3bb8-11e7-88a2-09e9d555bae6.png)

![screen shot 2017-05-18 at 10 17 19 am](https://cloud.githubusercontent.com/assets/332617/26207663/2a23f58c-3bb6-11e7-90ea-8e5ea3648141.png)

Here is the JSON response from `api/v1/alerts/groups` for pre commit:

[pre.txt](https://github.com/prometheus/alertmanager/files/1011239/pre.txt)

and post:

[post.txt](https://github.com/prometheus/alertmanager/files/1011240/post.txt)

Finally here is the configuration I used to replicate the issue:

[prometheus.txt](https://github.com/prometheus/alertmanager/files/1011251/prometheus.txt)
[alertmanager.txt](https://github.com/prometheus/alertmanager/files/1011252/alertmanager.txt)

jjneely:
So, there is a known issue #539 that causes duplicate alerts to appear on the UI when routing alerts to two different receivers.  This should be fixed with the UI design.  But testing this version....yes now I have 4 alerts in the UI when routing alerts to 2 receivers.

This is not something I can expose my operations teams to.

mxinden:
@ipstatic as @jjneely pointed out, this will be fixed with the UI rewrite. The only thing holding us back from releasing the new UI right now is #806, which should be merged in the upcoming days. 

Thanks for reporting the issue. I will close here. Feel free to reopen if you have any further questions.

ipstatic:
@mxinden Actually @jjneely was pointing out this is not the same issue. The API is duplicating data per "group" after this filtering was added. Please look at my pre and post dumps of this data.

jjneely:
Yes, this is now significantly worse.  We've introduced more bugs, not kept the same.

mxinden:
@ipstatic I am sorry for closing so quickly here.  I will look into it tomorrow.

r4j4h:
[...]

PR #806, opened 2017-05-18
Rework SilenceForm

mxinden:
* Style the SilenceForm
* Validate the SilenceForm
  Instead of saving each value of the SilenceForm as a plain string I save it as a `ValidatedString` which caries the plain string as well as validation information which I then display underneath each input field.
  There are **definitely** better ways and prettier ways to do this. So I am very much looking forward to your feedback. (Turns out, validating a form is harder than I thought.)
* Redirect to creating a new Silence when trying to edit an expired one

Fixes issue #798.


ui/app/src/Views/SilenceForm/Updates.elm
```diff
@@ -172,5 +191,17 @@ update msg model =
 
                 newSilence =
                     toSilence newForm
+
+                -- In case the new silence is valid, update the affected alerts list
+                cmd =
+                    if isFormValid newForm then
+                        case newSilence of
+                            Ok silence ->
+                                Task.perform (PreviewSilence >> MsgForSilenceForm) (Task.succeed silence)
```
w0rm: Does it mean that on each keystroke we will refetch alerts?

ui/app/src/Views/SilenceForm/Types.elm
```diff
@@ -128,3 +153,78 @@ type SilenceFormFieldMsg
...
+
+
+areMatcherValid : List ValidatedMatcher -> Bool
+areMatcherValid matchers =
+    List.foldl (\m acc -> (isMatcherValid m) && acc) True matchers
+
+
+isMatcherValid : ValidatedMatcher -> Bool
+isMatcherValid matcher =
+    if matcher.name.result == Valid && matcher.value.result == Valid then
```
w0rm: You don't need an `if` for a boolean expression

```
isMatcherValid : ValidatedMatcher -> Bool
isMatcherValid {name, value} =
    name.result == Valid && value.result == Valid
```

ui/app/src/Utils/Date.elm
```diff
@@ -92,6 +92,7 @@ timeFromString =
         >> Maybe.map (ISO8601.toTime >> toFloat)
 
 
+
```
w0rm: Does it come from elm-format?

ui/app/src/Utils/FormValidation.elm
```diff
@@ -0,0 +1,62 @@
...
+
+
+areValidResults : List ValidationResult -> Bool
+areValidResults results =
+    List.foldr (\r acc -> (isValidResult r) && acc) True results
+
+
+isValidResult : ValidationResult -> Bool
+isValidResult result =
+    case result of
```
w0rm:
```
isValidResult : ValidationResult -> Bool
isValidResult =  
    (==) Valid
```

ui/app/src/Utils/FormValidation.elm
```diff
@@ -0,0 +1,62 @@
...
+
+type alias ValidatedString =
+    { value : String
+    , result : ValidationResult
+    }
+
+
+validateDate : String -> ValidatedString
+validateDate date =
+    case timeFromString date of
```
w0rm: What bothers me in this solution is that we discard the result of validation. Why don't we store the time in the `Valid` state?

ui/app/src/Views/SilenceForm/Views.elm
```diff
@@ -81,45 +90,41 @@ createSilence maybeId silenceResult =
                 Nothing ->
                     "Create"
     in
-        case silenceResult of
-            Ok silence ->
-                a
+        case ( isFormValid form, silenceResult ) of
```
w0rm: If form is invalid, then the `silenceResult` should be `Err`, we don't need two indicators.

ui/app/src/Utils/Date.elm
stuartnelson3: We'll find out if the tests pass :)

ui/app/src/Views/SilenceForm/Types.elm
```diff
@@ -45,46 +63,53 @@ toSilence { id, createdBy, comment, startsAt, endsAt, matchers } =
...
+    , startsAt = validateStartsAt ""
+    , endsAt = validateEndsAt ""
+    , duration = validateDuration ""
     , matchers = []
     }
 
 
 fromMatchersAndTime : List Matcher -> Time -> SilenceForm
-fromMatchersAndTime matchers now =
+fromMatchersAndTime maybeEmptyMatchers now =
```
stuartnelson3: I would vote for not using `maybe` in the variable name, as that signals the `Maybe` type to me when first reading it

ui/app/src/Utils/FormValidation.elm
```diff
@@ -0,0 +1,62 @@
...
+
+type ValidationResult
+    = Valid
+    | NotValid String
+
+
+type alias ValidatedString =
+    { value : String
+    , result : ValidationResult
+    }
```
stuartnelson3: This is piggy-backing on @w0rm's comment: I think we can get rid of the type alias, and squash all information into the union type:

```
type ValidationResult
    = Valid String
    | NotValid String
```

That way when you do `case validation of`, you can just access the value directly.

stuartnelson3: I would need to play with this / ask for the wisdom of the @w0rm, but potentially it could also be

```elm
type ValidationResult a
    = Valid a
    | NotValid String
```
I think then it can capture different types, like when using a `Maybe`. So you would have a function signature like:

```elm
isMatcherValid : ValidationResult Matcher -> Bool
```

ui/app/src/Views/SilenceForm/Updates.elm
stuartnelson3: I think that's what's happening.

I think for now we should keep the preview a manual feature, and create a new issue for adding a debounce on keystrokes and refreshing after, e.g., a 500ms delay since the last keystroke (if the form is valid)

ui/app/src/Views/SilenceForm/Views.elm
stuartnelson3: Right, like in `preview` below

ui/app/src/Utils/Date.elm
stuartnelson3: We'll find out if the tests pass :)

ui/app/src/Utils/FormValidation.elm
w0rm: Such type already exists, it is called Result, that is parametrized by success and error types and provides a few nice functions.

mxinden: My only issue with `Result` is, that, in case of `Err` I still want to access a. Should I just create a tuple:
```
Result a (a, String)
```
Where `a` is my actual text input (e.g. Silence creator name) and String is my error message.

@w0rm What do you think?

w0rm: @mxinden you would still need to store the raw input value + a result of validated input separately. The idea is to keep the validation result, because it's type may be not the raw string type, e.g. for a time input. Then, because each field stores the validation result, we can get all values from there and create a silence out of them, maybe via mapping all results with a silence constructor using smth like http://package.elm-lang.org/packages/elm-lang/core/latest/Result#map5

fabxc:
I think this is a really important improvement. How does everyone feel about getting this merged and releasing a `0.7-rc.0` today. Code nits could be addressed in a subsequent PR as long as this functionally works.

#805 shows again that we really have to move forward with all of this.

w0rm:
@fabxc sure, in this case we can merge this and work on top.

ui/app/src/Views/SilenceForm/Updates.elm
mxinden: Reverted 👍

ui/app/src/Views/SilenceForm/Types.elm
mxinden: Thanks, changed.

ui/app/src/Utils/Date.elm
mxinden: Will format it. Thanks.

ui/app/src/Utils/FormValidation.elm
mxinden: Removed `isValidResult`

ui/app/src/Utils/FormValidation.elm
mxinden: I have rewritten it entirely. We now safe the result so we don't have to recalculate it later. Thanks for the hint!

ui/app/src/Views/SilenceForm/Views.elm
mxinden: True, adjusted.

ui/app/src/Views/SilenceForm/Types.elm
mxinden: Good point. Changed.

ui/app/src/Utils/FormValidation.elm
mxinden: I have rewritten it, this time with the help of Elm `Result`.

Each value in the `SilenceForm` is now a `Result`. The `Result` for `startsAt` e.g. is: 
```
Result (String, String) (String, Time)
```
If  `startsAt` can be parsed it will be `Ok (String, Time)` where `String` is the string representation of `startsAt` and `Time` is the parsed time. This way whenever we want to access the parsed representation we don't have to parse it again.

If  `startsAt` can not be parsed it will be `Err (String, String)` where the first `String` is the string representation of `startsAt` and the second `String` is the error message explaining why it could not be parsed.

@w0rm @stuartnelson3 What do you think?
(@w0rm using `Result.map5` as suggested by you. Cool hint)

mxinden:
@w0rm @stuartnelson3 I have replied to all of your comments and changed the code accordingly. Could you re-review? Either I will address the next comments in this PR, or, if you don't mind, push it off to a new issue so we can do the release candidate.

ui/app/src/Utils/FormValidation.elm
```diff
@@ -0,0 +1,72 @@
...
+                Err err ->
+                    Err err
+
+                Ok m ->
+                    Ok (m :: matchers)
+
+
+validatedMatcherToMatcher : ValidatedMatcher -> Result String Matcher
+validatedMatcherToMatcher matcher =
+    case ( matcher.name, matcher.value ) of
```
w0rm: You can use `Result.map2` here

ui/app/src/Utils/FormValidation.elm
```diff
@@ -0,0 +1,72 @@
...
+        [] ->
+            agg
+
+        m :: list ->
+            fold list (aggregateValidatedMatchers agg m)
+
+
+aggregateValidatedMatchers : Result String (List Matcher) -> ValidatedMatcher -> Result String (List Matcher)
+aggregateValidatedMatchers agg matcher =
+    case agg of
```
w0rm: You can compose with `Result.andThen` and `Result.map` here.

mxinden: @w0rm I am down to this now. How do I compose the two?
```
    Result.andThen
        (\matchers ->
             Result.map (\m -> m::matchers) (validatedMatcherToMatcher matcher)
        )
        agg
```

ui/app/src/Utils/FormValidation.elm
mxinden: 👍

ui/app/src/Utils/FormValidation.elm
w0rm: Yeah, looks fine

PR #811, opened 2017-05-22
silence: fix and test expiration behavior

### LINKED ISSUE #783: Silences placed in future creates a dead-lock.
When creating a future-silence, the silence is created with a delete button that doesn't function. I see no clear way of disarming or deactivating the silence.

fabxc:
Fixes #783

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -446,12 +446,17 @@ func (s *Silences) expire(id string) error {
 	if !ok {
 		return ErrNotFound
 	}
-	if getState(sil, s.now()) == StateExpired {
-		return errors.Errorf("silence %s already expired", id)
-	}
-
 	sil = cloneSilence(sil)
-	sil.EndsAt = s.now()
+	now := s.now()
+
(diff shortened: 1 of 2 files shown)


RSamal:
For pending state please update the below code . Otherwise if I try to expire/delete the Silencer from UI. The alert still stays there. It will go to the expired state only after that time arrive. Logic wise works just bit of confusing to see through ui. Setting the startAt to now expire it right away.

`	case StatePending:
		sil.StartsAt = now
		sil.EndsAt = now`

fabxc:
Yes, that confusing display behavior is what we discussed in #783.
@mxinden we haven't considered also modifying the start timestamp to now, which will naturally sort it under expired. Do we prefer that or just evaluating any `startsAt == endsAt` as expired regardless of the current time?

mxinden:
@RSamal @fabxc This sounds like the best solution, as it does not require any frontend changes and no additional backend API code. I will move the discussion to issue #783 as this is merged now.

Issue #815, opened 2017-05-22
Prometheus stops firing but alerts stay active


mxinden:
This issue was reported by Jack and Julius [here](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/prometheus-users/QBTPLmB5c2o/RGDKytieAwAJ):

> When the metrics behind my test alerts go back to normal, Prometheus stops firing, but the alert stays active in AM.  The PagerDuty incident is not resolved.  Yikes.

> Can confirm. A simple "up == 0" alert sticks around in Alertmanager (current master) after it gets resolved in Prometheus, and also doesn't disappear after 5 minutes (the resolve timeout). It stays in the UI and it keeps sending "firing" webhook notifications.

juliusv:
Prometheus built from current master:

```
./prometheus -alertmanager.url=http://localhost:9093/
INFO[0000] Starting prometheus (version=1.6.2, branch=master, revision=5405a4724fbc7307786a65c176168a8948c0cbbf)  source="main.go:88"
INFO[0000] Build context (go=go1.8, user=julius@julius-samsung, date=20170523-13:45:49)  source="main.go:89"
```

Alertmanager built from current master:

```
./alertmanager
INFO[0000] Starting alertmanager (version=0.7.0-rc.0, branch=master, revision=6be3fbcb68af2ceb4d6bf77049a3390a1d6ad32f)  source="main.go:109"
INFO[0000] Build context (go=go1.8, user=julius@julius-samsung, date=20170523-13:47:45)  source="main.go:110"
```

Prometheus config:

```
global:
  scrape_interval: 5s

rule_files:
  - "prometheus.rules"

scrape_configs:
  - job_name: 'prometheus'
    static_configs:
      - targets: ['localhost:9090']
  - job_name: 'node'
    static_configs:
      - targets: ['localhost:9100']
```

`prometheus.rules` file:

```
ALERT SomethingIsDown IF up == 0
```

Alertmanager config:

```
route:
  group_by: ['alertname']
  group_wait: 30s
  group_interval: 5m
  repeat_interval: 3h

  receiver: webhook

receivers:
- name: 'webhook'
  webhook_configs:
    - url: 'http://localhost:12345/'
```

1. Start Prometheus without node exporter running
2. Wait for alert to fire - it appears on AM and notifications to the webhook (listener started with netcat) are sent
3. Bring up node exporter

[...]

PR #820, opened 2017-05-23
Do not expose resolved alerts & do not send resolved if never firing

mxinden:
Do not expose resolved alerts on the /alerts endpoint. Do not send
resolved alerts to receivers if the alerts have never been fired before.

This is related to #815, but does not resolve the specified issue.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -300,6 +300,11 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 			continue
 		}
 
+		// Continue if alert is resolved
+		if !a.Alert.EndsAt.IsZero() && a.Alert.EndsAt.Before(time.Now()) {
+			continue
+		}
+
 		status := api.getAlertStatus(a.Fingerprint())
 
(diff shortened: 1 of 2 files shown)


juliusv:
Travis is failing with:

```
/app/ui/app/tests/Main.elm has an invalid module declaration. Check the first line of the file and make sure it has a valid module declaration there!
Makefile:6: recipe for target 'test' failed
make: *** [test] Error 1
```

juliusv:
I can confirm that this fixes the UI issues I observed in https://github.com/prometheus/alertmanager/issues/815

stuartnelson3:
Try rebasing against master, @w0rm just updated our tests to run on CI

w0rm:
@stuartnelson3 the problem is because https://github.com/prometheus/alertmanager/blob/master/ui/Dockerfile doesn't fix the versions of `elm-format` and `elm-test`, it should be `elm-format@0.6.1-alpha` and `elm-test@0.18.3` now.

mxinden:
Rebased PR. Tests are all green.

PR #827, opened 2017-05-24
Improved Validation

w0rm:
1. Extracted affected alerts out of the silence type
2. Unified ApiData
3. Added separate validation messages for each field to validate on blur
4. Changed Update/Preview messages so they create the silence from the form or trigger validation errors
5. If there was an error from the backend when saving a silence, it will be displayed

I thought of changing the form styles in this PR, but I think it's too much.


ui/app/src/Main.elm
```diff
@@ -8,7 +8,7 @@ import Types
         ( Route(..)
         , Msg
             ( NavigateToSilenceList
-            , NavigateToSilence
+            , NavigateToSilenceView
```
mxinden: Not a big thing: Why adding `View` to `NavigateToSilence` msg? We don't do it with the other ones.

ui/app/src/Main.elm
```diff
@@ -54,7 +54,7 @@ init location =
                 _ ->
                     nullFilter
     in
-        update (urlUpdate location) (Model initSilenceList Loading initSilenceForm initAlertList route filter initStatusModel)
+        update (urlUpdate location) (Model initSilenceList initSilenceView initSilenceForm initAlertList route filter initStatusModel)
```
mxinden: 👍

ui/app/src/Utils/Types.elm
```diff
@@ -1,20 +1,19 @@
...
-import Http
 import Time
 
 
-type ApiResponse e a
+type ApiData a
     = Initial
     | Loading
-    | Failure e
+    | Failure String
```
mxinden: Having a string instead of a type variable forces us to drop information, doesn't it? Can't we keep the type variable here and convert an error to a string in the view functions?

ui/app/src/Views/SilenceForm/Types.elm
```diff
@@ -4,209 +4,198 @@ module Views.SilenceForm.Types
...
 import Time exposing (Time)
 import Utils.FormValidation
     exposing
-        ( validateDate
-        , ValidatedMatcher
-        , validatedMatchersToMatchers
+        ( initialField
+        , ValidationState(..)
+        , ValidatedField
+        , validate
```
mxinden: 👍

ui/app/src/Views/SilenceForm/Types.elm
```diff
@@ -4,209 +4,198 @@ module Views.SilenceForm.Types
...
+                | id = id
+                , comment = nonEmptyComment
+                , matchers = validMatchers
+                , createdBy = nonEmptyCreatedBy
+                , startsAt = parsedStartsAt
+                , endsAt = parsedEndsAt
+            }
+        )
+        (stringNotEmpty comment.value)
+        (List.foldr appendMatcher (Ok []) matchers)
```
mxinden: When I don't specify any matchers and create a silence, I get redirected to the `/silences` page without any error messages in the UI. The API call fails with `silence invalid: at least one matcher required`. Should we enforce having at least one matcher in this PR or in an upcoming one? Sorry, I missed this requirement in my last changes as well.

ui/app/src/Utils/FormValidation.elm
```diff
@@ -1,60 +1,57 @@
...
 
-aggregateValidatedMatchers : Result String (List Matcher) -> ValidatedMatcher -> Result String (List Matcher)
-aggregateValidatedMatchers agg matcher =
-    Result.andThen
-        (\matchers ->
-            Result.map (\m -> m :: matchers) (validatedMatcherToMatcher matcher)
-        )
-        agg
+validate : (String -> Result String a) -> ValidatedField -> ValidatedField
+validate validate field =
```
mxinden: Could we do `validate validator field =` instead? A little confusing to see the same term twice.

ui/app/src/Utils/FormValidation.elm
```diff
@@ -1,60 +1,57 @@
...
+    { field | validationState = fromResult (validate field.value) }
 
 
-validatedMatcherToMatcher : ValidatedMatcher -> Result String Matcher
-validatedMatcherToMatcher matcher =
-    Result.map2 (\n v -> Matcher n v matcher.isRegex) matcher.name matcher.value
-        |> Result.mapError second
+stringNotEmpty : String -> Result String String
```
mxinden: In the long run we should trim the string as well so  a space is not valid either.

ui/app/src/Utils/Types.elm
```diff
@@ -1,20 +1,19 @@
 module Utils.Types exposing (..)
 
-import Http
 import Time
 
 
-type ApiResponse e a
```
mxinden: If I see it correctly we get rid of `ApiResponse`? 👍 for removing an extra layer of complexity.

ui/app/src/Utils/Types.elm
w0rm: I don't see where we require this information. I need to create the error myself, when e.g. validation fails. It is easier to create it from `String`, rather than manually create `Http.Error`, I don't think this is even possible.

w0rm: We can always change the `Failure String` to be a `Failure` of our own union type if we need to extract additional information.

ui/app/src/Views/SilenceForm/Types.elm
w0rm: I forgot to remove the redirect. Now it should display the error from the backend. It is not pretty, we should address this later.
 
screen shot 2017-05-24 at 18 08 37

ui/app/src/Utils/FormValidation.elm
w0rm: Renamed

ui/app/src/Utils/FormValidation.elm
w0rm: Added `String.trim`

ui/app/src/Utils/Types.elm
w0rm: Yeah, I had no idea of why both were there. It was pretty confusing

ui/app/src/Main.elm
w0rm: Because I renamed `Silence` to `SilenceView`, which represents a page. I didn't want to confuse it with a Silence type.

mxinden:
Related to #798.

ui/app/src/Utils/FormValidation.elm
mxinden: Thanks.

ui/app/src/Views/SilenceForm/Types.elm
mxinden: This is definitely good enough for now. Thanks.

ui/app/src/Utils/Types.elm
mxinden: 👍

ui/app/src/Views/SilenceView/Views.elm
```diff
@@ -27,8 +28,8 @@ view model =
             error msg
 
 
-silence2 : Silence -> Html Msg
-silence2 silence =
+silence2 : ApiData (List Alert) -> Silence -> Html SilenceViewMsg
```
mxinden: Just a random question: Is there a particular reason we called this `silence2`?

ui/app/src/Views/SilenceForm/Types.elm
w0rm: Oh, I've got an idea. We should just hide the delete button, when there is only one matcher left. Will do it in another pull request after this is merged.

ui/app/src/Views/SilenceView/Views.elm
w0rm: Maybe because `silence` is taken. I guess it's a question to @stuartnelson3

w0rm: I renamed it to `viewSilence`

ui/app/src/Utils/Types.elm
```diff
@@ -30,10 +29,6 @@ type alias Label =
     ( String, String )
 
 
-type alias ApiData a =
-    ApiResponse Http.Error a
```
stuartnelson3: I think I originally brought this in from reading the RemoteData blog post. I think the other had a generic type, and then made a specific type alias for dealing with http requests.

ui/app/src/Utils/Views.elm
```diff
@@ -47,19 +47,43 @@ checkbox name status msg =
         ]
 
 
-validatedFormField : String -> Result ( String, String ) String -> String -> (String -> msg) -> Html msg
-validatedFormField labelText validatedString classes msg =
-    case validatedString of
-        Ok inputValue ->
+validatedField : (List (Attribute msg) -> List (Html msg) -> Html msg) -> String -> String -> (String -> msg) -> msg -> ValidatedField -> Html msg
+validatedField htmlField labelText classes inputMsg blurMsg field =
+    case field.validationState of
```
stuartnelson3: The three different cases in this function differ only by the css classes `has-success`, `has-danger`, or the absence of these two. Would it be possible to just write the HTML once and switch on the css class?

ui/app/src/Utils/Views.elm
```diff
@@ -117,27 +124,8 @@ loading =
...
-
-                BadUrl url ->
-                    "malformed url: " ++ url
-    in
-        div []
-            [ p [] [ text <| "Error: " ++ msg ]
-            ]
+    div []
+        [ p [] [ text <| "Error: " ++ err ]
+        ]
```
stuartnelson3: +1 for moving this logic out of the view

ui/app/src/Views/SilenceForm/Types.elm
stuartnelson3: I smell a library forming here

ui/app/src/Views/SilenceForm/Types.elm
```diff
@@ -4,209 +4,198 @@ module Views.SilenceForm.Types
...
+    }
 
 
-validateMatcher : Matcher -> ValidatedMatcher
-validateMatcher matcher =
-    { name = validateMatcherName matcher.name, value = validateMatcherValue matcher.value, isRegex = matcher.isRegex }
+validateForm : SilenceForm -> SilenceForm
+validateForm { id, createdBy, comment, startsAt, endsAt, duration, matchers } =
+    { id = id
+    , createdBy = validate stringNotEmpty createdBy
```
stuartnelson3: mmm love that `validate` function

ui/app/src/Views/SilenceForm/Types.elm
```diff
@@ -4,209 +4,198 @@ module Views.SilenceForm.Types
...
-            , matchers = List.map validateMatcher enrichedMatchers
-        }
-
-
 type alias Model =
-    { silence : Result String Silence
-    , form : SilenceForm
+    { form : SilenceForm
+    , silenceId : ApiData String
+    , alerts : ApiData (List Alert)
```
stuartnelson3: I like having alerts exist independently on this and the silence view model, +1

ui/app/src/Views/SilenceForm/Updates.elm
```diff
@@ -115,72 +169,84 @@ updateForm msg form =
...
+                        , form = validateForm model.form
+                      }
+                    , Cmd.none
+                    )
 
-        SilenceCreate silence ->
-            case silence of
-                Success id ->
-                    ( model, Navigation.newUrl ("/#/silences/" ++ id) )
+        SilenceCreate silenceId ->
```
stuartnelson3: We can definitely re-think these names I came up with. `SilenceCreate` and `CreateSilence` could definitely be better.

(note for the future)

ui/app/src/Utils/Views.elm
w0rm: @stuartnelson3 another class is applied to the input field (`form-control-success` or `form-control-danger`), plus there is an error message in the error case. I think this is fine.

w0rm:
@stuartnelson3 I made it impossible to delete the last matcher. But accidentally committed `True` there instead of using `showDeleteButton`. Thanks, fixed now!

ui/app/src/Utils/Types.elm
w0rm: Since that thing was a library, it had to be more generic. We don't need more than String for error at the moment. Changing this to smth else later won't be a problem.

ui/app/src/Utils/Views.elm
stuartnelson3: mk

PR #832, opened 2017-05-29
Expire pending silence and move to expired state

### LINKED ISSUE #783: Silences placed in future creates a dead-lock.

mxinden:
Instead of setting endsAt to startsAt we can set both to now. Thereby
the Silence will get the expired state by default.

Fixes #783.


fabxc:
👍

Issue #681, opened 2017-04-02
Add ID for alerts and add alert ID's to silences


HyperDevil:
Hi,

I am trying to parse the API output, for alerts and silences. (in PHP)
The silences seem to have an ID, but does not have the explicit alarm ID attached in the JSON output (if alarm has an ID?).
Also the alarms do not have an unique ID so that i cannot match which alarms are being silences by a certain rule.
Is there an internal connection between the alarms and the silences?

stuartnelson3:
In the current data model, alerts are ephemeral and do not have an ID assigned to them.

Internally, the first silence that matches (and silences) an alert has its ID added to the alert's data structure in the `silenced` field.

You can do two things:

1. You can iterate through the alerts and check for the `silenced` field. This only shows the first silence that affects the alert, though, and won't show any additional silences that also affect it.
2. You can take the label matchers defining a silence, and make a request to the alert groups endpoint:

```
http://alertmanager.example.biz/api/v1/alerts/groups?filter=%7Bfoo%3D%22bar%22%2C%20baz%3D~%22quux%22%7D
```

(filter == `{foo="bar", baz=~"quux"}`)

And that will return the alerts that match the matchers in `filter`.

stuartnelson3:
That having been said, we are adding alert ID's as part of the UI/API update that is currently underway. Assigning to @mxinden since he's so keen on it.

fabxc:
I'm curious about those alert IDs. What does the current plan look like?

mxinden:
@stuartnelson3: I talked to @fabxc in regards to adding Alert ids with the API update. There are no actions (create, update, delete) needed in the UI where we have to address a specific alert. Thereby adding a specific Alert id is not on the agenda. Doing so would open up many questions in regards to the realisation of eventual consistency inside a mesh network of Alertmanagers.

@HyperDevil Let us know if @stuartnelson3 answer solves your problem.

HyperDevil:
@mxinden and @stuartnelson3 
Thank you for the answers so far.
I am currently using version 0.4.2 of Alertmanager, and i have silences active. But i cannot seem to find any alerts with silences attached when using the /api/v1/alerts, am i using an old version that does not have this?

The point for me is that i need to make a list of alerts outside of Alertmanager using the API, and i want to filter out the alarms that i silcenced (i dont care about which silence rule, as long as it is silenced)

stuartnelson3:
[...]

PR #833, opened 2017-05-29
Show alert annotations

### LINKED ISSUE #821: Display alert annotations
The old UI used to contain current alerts listed by labels with silence buttons, as does the new UI.

To the left of the labels list in the old UI there used to be a Source link with a `+` button. Upon clicking the plus, a "panel" would pop out beneath showing the actual "hydrated" alert text.

The new UI seems to be missing this functionality.

For example, if the alert was `Low Quorom: There's only {{value}} things!` in Prometheus' config view here it would show up as `Low Quorom: There's only 3 things!` if the represented time series' value was 3.

Here's some pictures for a better illustration:

Prometheus example:

prom

0.6 example, with + button already clicked:

0.7 of the same alert, notice there is a Source button but if there is an expander I do not see it :(

w0rm:
Closes  #821

annotations


w0rm:
Please don't merge yet, I noticed a small issue, will fix before noon.

w0rm:
@mxinden this bindata stuff is annoying. It produces different file locally, even when I pull the latest golang docker and do `docker run --rm -t -v "$(pwd):/usr/src/app" -w /usr/src/app golang make assets`

w0rm:
@stuartnelson3 oh I see. it would be nice if alerts had ids so we didn't have to rely on index

stuartnelson3:
Well, they DO have a unique fingerprint.

@mxinden @fabxc @brancz thoughts on returning the alert's fingerprint from the `/alerts` endpoint?

w0rm:
@stuartnelson3 if we're going to add this field, will it be named `id`? I can mock it in the decoder to be index for now, and then we can use the real value.

mxinden:
Is it possible to finish this PR without adding *id* to the alert object for now? If so, can we move the *AlertID* discussion to this issue #681

w0rm:
@mxinden I think it's fine to enhance alerts with local ids. We will just have to fix the decoder, and it will still work. Please re-review my changes.

w0rm:
If @stuartnelson3 approves, I'd like to ask someone of you to regenerate the bindata, cause I can't make it work.

w0rm:
@stuartnelson3 @mxinden sorry for the misinformation, it was the same problem of updated `elm-test`. I fixed the versions in the Dockerfile and it works now.

Issue #323, opened 2016-04-26
-web.external-url does it work?


mokshpooja:
Hi Guys,

I am using proxy thus both prometheus and alert manager are both behind proxy. I used -web.external-url in prometheus which is working fine but this doesn't seem to be happening for alert manager.
I can reach the main page alert page but not the css or any other styling files thus i see a very primitively styled page.

juliusv:
Hi, this should have gotten fixed in https://github.com/prometheus/alertmanager/pull/289, but the newest release doesn't contain that fix yet (0.1.1 is from March 15, the fix is from March 30).

/cc @fabxc for maybe doing a release already, but I know he's quite busy right now...

juliusv:
Closing the issue here, since most likely this is fixed in master.

mokshpooja:
buildng from master works. Thanks. Hope the release is out soon.

fabxc:
@juliusv please see prometheus/prometheus#1191 and @chancez's suggestion.
`-web.external-url` currently doesn't solve all cases and should we change it, we shouldn't release it for AM in the first place.

jcollie:
Just tried out 0.7.0-rc.0 and it's failing because the main page has an absolute reference to "/script.js".

stuartnelson3:
The path is correct. I just downloaded and ran the release candidate, and experienced no failure.

jcollie:
The path "/script.js" is only correct if you have AlertManager at the root of the URL path. If, like me, you are using a proxy in front of AlertManager to put it into a subdirectory it's not correct. I have the proxy set up so that https://www.example.com/prometheus forwards to my Prometheus installation and https://www.example.com/alertmanager forwards to my AlertManager installation.

psalaberria002:
I can confirm this is also happening here with v0.7.0-rc.0

stuartnelson3:
Ah, thanks for the clarification

psalaberria002:
@mxinden I think this should be added to v0.7 Milestone. It should be an easy fix.

mxinden:
@psalaberria002 @jcollie We have fixed this with #836. You can try out current master or wait a little. We are planning to release again hopefully tomorrow. Let me know if you are having any further issues with this.

PR #836, opened 2017-05-31
Fix external web url 

stuartnelson3:
Seems to be working from initial tests, but still
need to try it out a bit more.

EDIT:

A consequence of this change is now we can use elm's built-in `elm-reactor` to compile and auto-recompile code during development. This means I was able to remove the hacky `main.go` file and its vendored dependencies I was using.

I also added `uglifyjs` as part of a script.js make target. It reduced the file size by 70%, so it seemed worth it. `gzip` at the default compression value further reduces this to 45kb, but that's something that can be added in a later PR.

Addresses #323


brancz:
I've been having some problems with this, so I'll drop this patch in my cluster to test it, thanks a lot for addressing it! I'm guessing the test failure needs you to re-generate the `bindata.go` file.

brancz:
Whatever we do, we should be consistent with Prometheus, in which case we need to introduce the path prefix at this point, and also have the API prefix default to the external URL path for its prefix.

I realize this is marked as WIP, but it does not currently work for me.

I'm running this in a Kubernetes cluster and have the external URL set to the address the Alertmanager is routed to via an ingress object.

stuartnelson3:
Thanks for the feedback! I don't have access to a kubernetes cluster, unfortunately, so I'm doing my best with a local nginx.

Did you regenerate the bindata? A large part of this change comes from the front-end code.

w0rm:
@stuartnelson3 sorry for the merge conflict

stuartnelson3:
Not a problem!

stuartnelson3:
I caught a few redirects I hadn't correctly updated, and it seems to be working when using nginx as a proxy on my local machine. Here's the config I'm using:

```nginx
worker_processes  5;  ## Default: 1
worker_rlimit_nofile 8192;

events {
  worker_connections  4096;  ## Default: 1024
}

http {
  index    index.html index.htm index.php;

  default_type application/octet-stream;
  log_format   main '$remote_addr - $remote_user [$time_local]  $status '
    '"$request" $body_bytes_sent "$http_referer" '
    '"$http_user_agent" "$http_x_forwarded_for"';
  sendfile     on;
  tcp_nopush   on;
  server_names_hash_bucket_size 128;

  server { # simple reverse-proxy
    listen       3000;
    server_name  domain2.com www.domain2.com;

    location /arbitrary/alertmanager/path {
      proxy_pass      http://localhost:9093;
    }
  }
}
```

Where I'm also setting the external url when I start alertmanager to `/arbitrary/alertmanager/path`. Let me know if this is still not working on kubernetes.

mxinden:
@w0rm I might be wrong but as far as I can tell we can't run `elm reactor` with an `index.html` file. Thereby bootstrap would be missing. So we will still need our golang dev server, right?

w0rm:
@mxinden we can insert the link tag from Elm.

stuartnelson3:
The change I made removes index.html and adds a template, because it needs to have the external url written into it when we serve up the front-end.

mxinden:
Ok, great. I remember me looking for a solution, but without success.

mxinden:
General design question: Instead of passing the external URL from the golang backend to the frontend, why don't we let the frontend figure its current url out by itself via the [`Location`](http://package.elm-lang.org/packages/elm-lang/navigation/2.1.0/Navigation#Location) type in the [*Navigation*](http://package.elm-lang.org/packages/elm-lang/navigation/2.1.0/Navigation#Location) package?

Benefit would be that the UI could be available under two different URLs, and that the `externalURL` parameter does not need to be configured in order for the frontend to work correctly.

Only issue left would be to relatively import the `app.js` file in our `index.html`. 

If I am not mistaking, this worked with the old Angular UI?! (See [Prometheus-Opertaor docs](https://github.com/coreos/prometheus-operator/blame/master/Documentation/user-guides/exposing-prometheus-and-alertmanager.md#L130))

Let me know what you think (/ if this idea is sane)?

stuartnelson3:
Really cool ideas from both of you.

@mxinden using location seems to be working, so that just gets set internally in the first `init` function instead of being passed in.

@w0rm I moved to just generating a single index.html instead of making a script and then having that be loaded by some semi-static HTML. This solves the problem of having to template the index.html so we can properly load the script.js. There's a new problem, though: I need to move the `<style>` tag to the top of the document, but I don't know how to do that. The style tag with bootstrap is currently being loaded after elm loads, so there's a delay where we show the page without any css. Any ideas?

stuartnelson3:
`elm-reactor` is "working" but I need your help @w0rm clarifying what you mean by having a default `Value` that we parse to have a different apiUrl value for the dev environment

ui/app/src/Views.elm
```diff
@@ -16,12 +16,31 @@ import Views.NavBar.Views exposing (navBar)
...
+        , navBar model.route
         , div [ class "container pb-4" ]
             [ currentView model ]
         ]
 
 
+renderLink : Html msg
+renderLink =
+    node "link"
+        [ href "https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-alpha.6/css/bootstrap.min.css"
```
stuartnelson3: We need to somehow move these into the `<head>` tags, as there's a delay when initially loading the page before the CSS is available.

w0rm: Hmm, maybe we can try to add `on "load"` to the link. I'm not sure how to avoid the flash of unstyled content.

w0rm:
@stuartnelson3 I meant this https://github.com/zalando/elm-street-404/blob/master/src/Main.elm#L53

ui/app/src/Views.elm
mxinden: Would it help (improve delay) to just serve the css file ourselves from the backend? As a side effect, one could use the Alertmanager without Internet connection.

stuartnelson3: +1 to @mxinden's suggestion, we should be doing that regardless. But perhaps in a different pr :)

stuartnelson3:
Ah ok, I get it. Is there a way to tell elm-make to add flag values to the output index.html? That's the thing I'm trying to figure out, I'm not sure how to differentiate between what elm-reactor creates and what we're going to be packaging into the binary for release, without doing either manual editing of the file or some weird command line hackery.

w0rm:
@stuartnelson3 why not use the dev setting as the default value from the elm reactor, and pass the one from the index.html template when running the compiled binary?

stuartnelson3:
@brancz If you get a chance, can you test out these changes in kubernetes? They are working for me locally with nginx.

stuartnelson3:
@w0rm @mxinden updated so that we can use `elm-reactor` locally instead of that go server. Let me know how this works for you all.

ui/app/src/Views.elm
stuartnelson3: Hm, wasn't able to figure out what to do with `on "load"`, it seems like that's an event that needs a message attached to it?

w0rm: @stuartnelson3  yeah and the state (cssLoaded) in the model

w0rm:
@stuartnelson3 this looks great! I added bootstrap loading state to prevent the flash of unstyled content. If bootstrap is not available, then it will show this error:

screen shot 2017-06-02 at 11 21 28

ui/app/src/Main.elm
```diff
@@ -53,8 +59,32 @@ init location =
 
                 _ ->
                     nullFilter
+
+        apiUrl =
+            flags
```
mxinden: If we get the url from the *Navigation* package, parsing flags should not be necessary anymore, right?

ui/app/src/Main.elm
```diff
@@ -53,8 +59,32 @@ init location =
...
+            (Model
+                initSilenceList
+                initSilenceView
+                initSilenceForm
+                initAlertList
+                route
+                filter
+                initStatusModel
+                baseUrl
+                apiUrl
```
mxinden: Why do we need both `apiUrl` and `baseUrl`? I can't think of a case where they are not the same.

ui/app/src/Views/AlertList/Updates.elm
```diff
@@ -12,61 +12,65 @@ import Utils.Filter exposing (generateQueryString)
...
-                    FilterBar.setMatchers filter filterBar
-            in
-                ( { model | alerts = Loading, filterBar = newFilterBar, groupBar = newGroupBar, activeId = Nothing }
-                , Api.fetchAlerts filter |> Cmd.map (AlertsFetched >> MsgForAlertList)
-                )
+                    newFilterBar =
+                        FilterBar.setMatchers filter filterBar
+                in
+                    ( { model | alerts = Loading, filterBar = newFilterBar, groupBar = newGroupBar, activeId = Nothing }
+                    , Api.fetchAlerts apiUrl filter |> Cmd.map (AlertsFetched >> MsgForAlertList)
```
mxinden: We should use `baseUrl` from the *Navigation* package here, right?

ui/app/src/Main.elm
```diff
@@ -25,21 +25,27 @@ import Views.AlertList.Types exposing (initAlertList)
...
         }
 
 
-init : Navigation.Location -> ( Model, Cmd Msg )
-init location =
+init : Json.Value -> Navigation.Location -> ( Model, Cmd Msg )
+init flags location =
     let
+        baseUrl =
+            location.pathname
```
mxinden: Can we change this to `basePath` or something similar? I think according to the [Wiki spec](https://en.wikipedia.org/wiki/URL#Syntax) the *url* is the entire string containing the *host* string as well (e.g. `github.com`), but we only want to get the *path* here (e.g. `prometheus/alertmanager`), right?

ui/app/src/Main.elm
stuartnelson3: In local development with elm-reactor, the baseUrl is something like `localhost:8000/src/Main.elm/#/alerts`, while the `apiUrl` would be wherever alert manager is running, in this case defaulting to `localhost:9093`

ui/app/src/Main.elm
stuartnelson3: I need a way to differentiate between using the template and using elm-reactor. It could be as simple as a boolean that says `development=true`, but if we want to use elm-reactor (which I think we should) we need a way to differentiate

ui/app/src/Views/AlertList/Updates.elm
stuartnelson3: since we're interacting with the api, we want the apiUrl. that includes the `/api/v1` stuff.

stuartnelson3:
I saw your local error, which I am no longer seeing after a couple changes.

regarding kubernetes: how is alertmanager being started? You need to pass it the external url:

```
./alertmanager -web.external-url /arbitrary/alertmanager/path
```

So my browser then makes requests to `http://localhost:4444/arbitrary/alertmanager/path/api/v1/alerts?silenced=false`

mxinden:
@stuartnelson3 I might be missing something, but why do we still need to specify the `-web.external-url` parameter in order for the frontend to work properly? The `script.js` file is loaded properly without the parameter in my k8s environment and all API calls can be routed correctly with the information obtained via the *Navigation* package. Sorry if this has been answered before, or if I am just missing something obvious.

brancz:
I believe for this to work properly we have to introduce some kind route prefix flag(s) at this point, just like in Prometheus. That would also be good in terms of consistency between the projects.

stuartnelson3:
When I use nginx (and based on how `-web.external-url` was working), requests for some route `/proxy` get forwarded to a particular backend. So if nginx is running in localhost:3000, `localhost:3000/proxy/#/alerts` gets sent to my alertmanager server, localhost:5000, without the path being modified, as `localhost:5000/proxy/#/alerts`. So it's the backend that requires this flag.

Kubernetes is only forward only the part pertaining to alertmanager (`/api/v1/alerts?silenced=false`), not the whole thing (`/api/v1/proxy/namespaces/default/pods/alertmanager-654384031-36n74/api/v1/alerts?silenced=false`), as indicated by `script.js` loading properly. So our two testing cases were different, which is probably the source of the confusion.

I pushed a change so that all url stuff is inferred, let me know if it's right ...

brancz:
I'll make sure to try this out on Tuesday. Why this will probably not suffice is that many reverse proxies or configurations of those rewrite the URL to trim the incoming path. For example Kubernetes' proxy does this which renders alermanager unusable in that case. In that case the external URL is necessary to be set to the proxy URL, but the server needs to actually be listening on root without a prefix.

ui/templates.go
```diff
@@ -9,7 +11,7 @@
     </head>
     <body>
         <!-- Your source after making -->
-        <script src="/script.js"></script>
-        <script>Elm.Main.embed(document.body)</script>
+        <script src="{{ .ExternalURL }}/script.js"></script>
```
mxinden: Is this still needed? In an k8s cluster with `kubectl proxy` `script.js` is loaded properly without specifying the `-web.external-url` parameter.

stuartnelson3: I believe this will still be needed for the simple reverse proxy situation I tested using nginx

.gitignore
```diff
@@ -7,6 +7,8 @@
 /.release
 /.tarballs
 
+.vimlocal
```
fabxc: I removed editor specific ignores various repos some time ago. There's just too many of them and they get more over time. These should be in every users global `.gitignore`.

ui/app/src/Views.elm
```diff
@@ -16,10 +19,44 @@ import Views.NavBar.Views exposing (navBar)
 view : Model -> Html Msg
 view model =
     div []
-        [ navBar model.route
-        , div [ class "container pb-4" ]
-            [ currentView model ]
+        [ renderLink "https://maxcdn.bootstrapcdn.com/bootstrap/4.0.0-alpha.6/css/bootstrap.min.css"
```
fabxc: We should be hosting that file ourselves.

mxinden: @fabxc Yes, for sure. See *outdated* [comment](https://github.com/prometheus/alertmanager/pull/836#discussion_r119600791) above, but if you don't mind we would do it in a separate PR.

.gitignore
stuartnelson3: ah, sorry about that

stuartnelson3:
@fabxc Fixing the URL allows us to rely on elm's built-in server/automatic recompiler for local development instead of the hacky go server I had, which is responsible for most of the diff.

ui/app/src/Views.elm
mxinden: See #846

brancz:
As in functionality that the `-web.external-url` flag works as in Prometheus, besides a small odd thing. When I type in `http://localhost:9093/alertmanager` for an Alertmanager instance I get redirected to `http://localhost:9093/alertmanager//#/alerts` (note double `/` after the path), that seems like a bug. All of this runs in a Kubernetes cluster and this behavior is consistent with a `kubectl port-forward` as well as via an Ingress.

In addition to that I have a Prometheus instance firing alerts against this instance and I noticed that it's not hitting the right endpoint, as this also automatically prefixes the API. This is why I mentioned above that we should have the ability to set the route prefix independently, to be able to maintain no prefix for traffic within the cluster, and only use the external url for redirecting and templating. Defaulting is still ok, but it should be possible to override to what the server binds the routes to.

stuartnelson3:
I see what your point is. Right now we either:

- Rewrite the routes for the API (by setting `-web.external-url`), or
- We don't re-write any routes

For proxies that do path translation (`/some/proxy/#/alerts` -> `/#/alerts`), there shouldn't be any need to set `-web.external-url`. The frontend infers its location from where it is loaded, and will just append the correct `/#/alerts` (or whatever route) to its requests. So, I don't think users have to do anything in the event of this sort of proxy (this is what @mxinden saw).

As you noticed, we run into problems when we set `-web.external-url` but also have a prometheus that is sending alerts to the default path. Because of how the frontend is working, though, I think a user needs to set `-web.external-url` only when interactions between other services and alertmanager (not alertmanager the backend and alertmanager the frontend) have to hit an altered path. The templating that currently exists can be removed. It only exists so that the CSS is loaded first, instead of after the main code is loaded.

I'll look at the double `//`, I thought I had fixed that.

Since I don't have a kubernetes cluster to test this against, and the feedback cycle is fairly slow, would it be possible to hand this off to @mxinden to finish? I think that'll get this merged much quicker.

mxinden:
@stuartnelson3 Sounds like a good idea. I will add the needed changes to the PR.

brancz:
I guess whatever the result of a discussion in https://github.com/prometheus/prometheus/issues/2813 is is what we should do here in terms of route prefix handling. It should definitely exist, but what a sensible default is still to be determined.

mxinden:
I fixed all the issues discussed above. @stuartnelson3 @w0rm Feel free to try it. @brancz is going to test it as well.

mxinden:
@stuartnelson3 @fabxc @brancz I have tested this fix with the following scenarios:

1. `kubectl port-forward  alertmanager-2171812374-crqfh 9093`
2. `kubectl proxy`
3.  ```
     - "-web.route-prefix=/alertmanager"
     - "-web.external-url=https://mxinden.dev.coreos.systems/alertmanager"
     # ingress.kubernetes.io/rewrite-target: "/"
     ```
4.  ```
     - "-web.route-prefix=/alertmanager"
     # - "-web.external-url=https://mxinden.dev.coreos.systems/alertmanager"
     # ingress.kubernetes.io/rewrite-target: "/"
     ```
5.  ```
    # - "-web.route-prefix=/alertmanager"
     - "-web.external-url=https://mxinden.dev.coreos.systems/alertmanager"
     # ingress.kubernetes.io/rewrite-target: "/"
     ```
6.  ```
     - "-web.route-prefix=/"
     - "-web.external-url=https://mxinden.dev.coreos.systems/alertmanager"
     ingress.kubernetes.io/rewrite-target: "/"
     ```
7.  ```
     - "-web.route-prefix=/"
    # - "-web.external-url=https://mxinden.dev.coreos.systems/alertmanager"
     ingress.kubernetes.io/rewrite-target: "/"
     ```
8.  ```
     # - "-web.route-prefix=/"
     # - "-web.external-url=https://mxinden.dev.coreos.systems/alertmanager"
     ingress.kubernetes.io/rewrite-target: "/"
     ```
9. On my local machine
10. On my local machine with the dev server

PR #839, opened 2017-05-31
Add favicon

### LINKED ISSUE #825: Add favicon
A small thing that will make the UI look complete.

w0rm:
screen shot 2017-06-01 at 00 27 06

Closes #825


kfdm:
A very small nitpick from a random passerby, but would it be useful to have the color be different, so that at a glance you could tell what tab is Prometheus and what tab is Alertmanager ?

fabxc:
Ha, this came up many times. The result was funnily enough always that the
discussion completely stalled over disagreement which alternative color
that should be :)
But you have a very valid point of course. I would say let's merge this so
there's something and then we can adjust a color later once we reached
agreement.

fabxc:
@w0rm I already retrigged Travis but the same issue remains. Not sure what that one is about.

kfdm:
Yeah, I seem to recall some discussion in prometheus/prometheus#1238 and color choice can be quite difficult for someone like me that lacks good design sense :)

w0rm:
@fabxc I needed to fix the node version in the Dockerfile.

Issue #840, opened 2017-06-01
Properly elide secrets in config in UI


brian-brazil:
Similarly to what we did in Prometheus.

@Conorbro

PR #841, opened 2017-06-02
Elide secrets in alertmanager config

Conorbro:
Implements #840

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -28,14 +28,21 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-var patAuthLine = regexp.MustCompile(`((?:api_key|service_key|api_url|token|user_key|password|secret):\s+)(".+"|'.+'|[^\s]+)`)
-
 // Secret is a string that must not be revealed on marshaling.
 type Secret string
 
 // MarshalYAML implements the yaml.Marshaler interface.
 func (s Secret) MarshalYAML() (interface{}, error) {
-	return "<hidden>", nil
+	if s != "" {
+		return "<secret>", nil
+	}
+	return nil, nil
+}
+
+//UnmarshalYAML implements the yaml.Unmarshaler interface for Secrets.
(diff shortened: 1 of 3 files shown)


config/testdata/conf.good.yml
```diff
@@ -0,0 +1,125 @@
+global:
```
brian-brazil: Also slack's api url, pushover's token, opsgenic api_key, victorops api_key,

Conorbro: They've been added now to the bottom of the yml.

config/config_test.go
```diff
@@ -60,3 +62,20 @@ receivers:
...
+	c, err := LoadFile("testdata/conf.good.yml")
+	if err != nil {
+		t.Errorf("Error parsing %s: %s", "testdata/good.yml", err)
+	}
+
+	// String method must not reveal authentication credentials.
+	s := c.String()
+	secretRe := regexp.MustCompile("<secret>")
+	matches := secretRe.FindAllStringIndex(s, -1)
+	if len(matches) != 18 || strings.Contains(s, "mysecret") {
```
brian-brazil: I count 10, how are you getting 18?

Conorbro: Remarshalling adds in fields that have secrets omitted.

brian-brazil: We should `omit_empty` any of those that are optional then.

config/config.go
```diff
@@ -28,14 +28,18 @@ import (
...
 
-var patAuthLine = regexp.MustCompile(`((?:api_key|service_key|api_url|token|user_key|password|secret):\s+)(".+"|'.+'|[^\s]+)`)
-
 // Secret is a string that must not be revealed on marshaling.
 type Secret string
 
 // MarshalYAML implements the yaml.Marshaler interface.
 func (s Secret) MarshalYAML() (interface{}, error) {
-	return "<hidden>", nil
+	return "<secret>", nil
```
brian-brazil: Return empty if the actual value is empty

PR #850, opened 2017-06-08
fix config JSON marshaling

brancz:
As in the code we marshaled/unmarshaled the config a few times (in cmd, api, cli), the elided secrets caused some parsing errors when marshaling/unmarshaling JSON. Unnecessary marshaling and unmarshaling is removed in this PR. `amtool` still does it, but it is necessary there.

@mxinden @stuartnelson3 @w0rm @fabxc 

@Kellel I fixed the usage/changes in `amtool` to the best of my ability, and made sure that the `amtool config` command works.

/cc @Conorbro


config/config.go
```diff
@@ -496,7 +501,7 @@ func (re *Regexp) UnmarshalJSON(data []byte) error {
 	if err := json.Unmarshal(data, &s); err != nil {
 		return err
 	}
-	regex, err := regexp.Compile(s)
+	regex, err := regexp.Compile("^(?:" + s + ")$")
```
mxinden: @brancz Is this just to anchor the regex?

brian-brazil: Yes

mxinden: @brian-brazil thanks! That was quick!

Issue #855, opened 2017-06-08
0.7.0 - filtering doesn't work in Chrome


jhorwit2:
I'm trying to add a simple filter `namespace=tyrell-test` and i receive the following error. 

```
script.js:1 Uncaught DOMException: Failed to execute 'pushState' on 'History': A history state object with URL 'http:/#/alerts?silenced=false&filter=%7Bnamespace%3D%22tyrell-test%22%7D' cannot be created in a document with origin 'http://asx001.clearspring.local:9093' and URL 'http://asx001.clearspring.local:9093/#/alerts'.
```

mxinden:
@jhorwit2 Thanks for reporting! I have created PR #856 with a fix. Does this version resolve your issue?

jhorwit2:
I'll do a build tomorrow and test it out. 👍

mxinden:
@jhorwit2 You can check out release v0.7.1 here: https://github.com/prometheus/alertmanager/releases/tag/v0.7.1

w0rm:
This is fixed

PR #849, opened 2017-06-07
*: cut v0.7.0

mxinden:
I would like to cut the v0.7.0 release. This should be delayed till #847 is merged.

I have copied and extended the release notes of v0.7.0-rc.0. Let me know if this is not the usual way to do it.

Thanks in advance for any feedback.

PR #856, opened 2017-06-08
Fix filtering by label

mxinden:
When clicking on a label on the AlertList or SilenceList view, the
changed Alert/Silence query is appended to the current url. This Postfix
should not start with a slash. If it does, the url is not appended but
replaced by the browser instead resulting in a `pushState` exception.

Fixed issue #855 
Addresses [comment](https://github.com/prometheus/alertmanager/pull/849#issuecomment-307099529) in #849


stuartnelson3:
Can you push your branches directly to the alertmanager repo? It saves us the trouble of tracking both this remote and yours.

brancz:
I strongly disagree with that practice @stuartnelson3. It clutters up the whole repo and when someone wants to actually go and browse the repository all they find are stale, unmerged, old outdated branches that no one cleans up.

If you want to checkout a PR use:

```
git fetch upstream pull/<PR-number>/head:<local-branch-name>
```

Or even better install [`hub`](https://github.com/github/hub) and use

```
git checkout https://github.com/github/hub/pull/134
```

Other than that, as you mentioned, it is mostly the same people working on these upstream development branches, then it should also not be much hassle to add each others remotes once.

Aside from that, but a lot more important, it also makes contributions feel a lot more welcome, when maintainer patches are done the same way everybody else has to.

mxinden:
@w0rm Thanks for the review. Tested this inside a Kubernetes cluster behind an Nginx as well.

stuartnelson3:
>I strongly disagree with that practice @stuartnelson3. It clutters up the whole repo and when someone wants to actually go and browse the repository all they find are stale, unmerged, old outdated branches that no one cleans up.

@brancz I think this statement and your following one are somewhat contradictory:

>Other than that, as you mentioned, it is mostly the same people working on these upstream development branches, then it should also not be much hassle to add each others remotes once.

If it's only a few people working on the repo, then any clutter is caused by just a handful of people. The cleanup for this is minimal, and as simple as deleting the branch on merge (which is our standard practice). Looking at the alertmanager and prometheus repos, there are fewer than 10 non-release branches currently open per repo.

Looking at a random sampling of PRs coming from org members, it seems that you and @mxinden are the only two following this practice. I haven't heard this mentioned before by anyone else in the org, but it seems you feel strongly about this. Perhaps it's something we should bring up to the group? Otherwise it's going to continue being you and max doing this, and no one else.

brian-brazil:
It's pretty normal for -team to push to the Prometheus repos, and to do a cleanup run afterwards. It's handy that Github also now offers the option of deleting a branch when merging.

brancz:
Yes I do feel this practice is not correct and I have brought it up before. I don't have the feeling that this "cleanup" everyone talks about actually happens, as on each repository I see multiple branches that have been abandoned, merged, or stale.

I don't seem to be alone with my opinion, but I can't force anyone into my practice if it's not the common consent. So I practice the way I believe is correct.

PR #859, opened 2017-06-09
Fix template color head table in email template

agareev:


template/email.html
```diff
@@ -323,7 +323,12 @@
       <div class="content">
         <table class="main" width="100%" cellpadding="0" cellspacing="0">
           <tr>
+            {{ if gt (len .Alerts.Firing) 0 }}
             <td class="alert alert-warning">
+            {{ end }}
+            {{ if gt (len .Alerts.Resolved) 0 }}
```
brian-brazil: Both of these can be true, you want an else here.

agareev:
I've fixed conditionals.

agareev:
@brian-brazil

brian-brazil:
This logic doesn't make sense. Some alerts being resolved does not mean things are okay, as there are still alerts firing.

agareev:
I haven't understand what did you mean (sorry for my bad english).
Isn't my pull request need?

mxinden:
@brian-brazil Could you help @agareev with this PR again?

PR #872, opened 2017-06-18
Stn/add receiver support

stuartnelson3:
@w0rm updated the backend code to support receiver filtering.

```
curl -sS 'http://localhost:9093/api/v1/alerts?silenced=false&receiver=webhook\d' | jq .data[].annotations.receiver
"webhook1"
"webhook1"
"webhook2"
"webhook2"
"webhook1"
"webhook1"
"webhook2"
"webhook2"
```

I updated `examples/ha/alertmanager.yml` to have additional receivers:

```
global:
  resolve_timeout: 5m

route:
  group_by: ['alertname']
  group_wait: 10s
  group_interval: 10s
  repeat_interval: 1h
  receiver: 'webhook'
  routes:
  - receiver: 'webhook1'
    match:
      instance: 'example1'
  - receiver: 'webhook2'
    match:
      instance: 'example2'

receivers:
- name: 'webhook'
  webhook_configs:
- name: 'webhook1'
  webhook_configs:
- name: 'webhook2'
  webhook_configs:
inhibit_rules:
  - source_match:
      severity: 'critical'
    target_match:
      severity: 'warning'
```

So, have fun with the front end!

addresses #800


api/api.go
```diff
@@ -278,46 +305,32 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
...
-		if !showSilenced && len(status.SilencedBy) != 0 {
-			continue
-		}
+	respond(w, flattenGroups(api.groups(matchers, showSilenced, re)))
+}
 
-		apiAlert := &APIAlert{
-			Alert:  &a.Alert,
-			Status: status,
+func flattenGroups(groups dispatch.AlertOverview) []*dispatch.APIAlert {
```
mxinden: Doesn't this result in duplicate alerts, because one alert can be present in multiple groups?
//CC @fabxc as I think we have talked about this before.

stuartnelson3: ah, I wasn't aware. Once @fabxc confirms I'll de-duplicate the alerts.

dispatch/dispatch.go
```diff
@@ -133,7 +138,17 @@ func (d *Dispatcher) Groups(matchers []*labels.Matcher) AlertOverview {
...
 				}
+
 				status := d.marker.Status(a.Fingerprint())
+
+				if !showSilenced && len(status.SilencedBy) != 0 {
+					continue
+				}
+
+				a.Annotations = a.Annotations.Merge(
+					model.LabelSet{"receiver": model.LabelValue(route.RouteOpts.Receiver)},
```
brancz: I believe due to the `continue` field in the config, an alert can be sent by multiple receivers, this value should be a list.

It makes me question whether we should be doing it this way, or maybe go the other way and have an index of the receivers to the leaf routes/aggregation groups and filter the alerts by the series of matchers to that leaf. This makes selecting multiple receivers more difficult though.

stuartnelson3: Hm, would that be an alert from a different aggrGroup, which would have a different route (and route receiver)? I'm reading the code and trying to make sense of it -- if an aggrGroup is defined by its set of labels, then that would mean the alert is at a different point in the tree, and potentially included?

In this case it would be a duplicate (except for the different `receiver` field), so yeah a list would be right. I totally forgot about `continue`. So we'll want to de-duplicate outside this method inside the poorly named `flattenGroups` function in `api.go`, I think.

>maybe go the other way and have an index of the receivers to the leaf routes/aggregation groups

It seems like from line 120 we already have that:

```go
for route, ags := range d.aggrGroups {
```

>This makes selecting multiple receivers more difficult though.

I think this is making an assumption about the UX, that we're going to have a multi-select html element. But even if we go this route, I think we could just squish all the selected receivers into a regexp `(receiverA|receiverB|receiverF)` when making the API call, and the current regexp match method would work.

brancz: The snippet you posted is the reverse of what I was thinking. I was thinking you have an index/map from receiver to the possible leaf routes that have that would send notifications to that receiver.

Maybe we're making this too complicated. I'm thinking the receivers could just be a short-hand to add values to the search query. As in you click on the receiver in the UI and it just adds the possible label queries that can lead to that receiver to the search, because in the end a receiver is just an abstraction/union of routes. I feel that would be the ideal UX, but I'm unsure about the complexity of this in terms of changes in the backend.

template/default.tmpl
```diff
@@ -1,5 +1,5 @@
 {{ define "__alertmanager" }}AlertManager{{ end }}
-{{ define "__alertmanagerURL" }}{{ .ExternalURL }}/#/alerts{{ end }}
+{{ define "__alertmanagerURL" }}{{ .ExternalURL }}/#/alerts?receiver={{ .Receiver }}{{ end }}
```
beorn7: This needs escaping of regexp special characters.

mxinden:
@stuartnelson3 After talking to @fabxc and @brancz I would like to propose a different approach on how to enrich alerts with the receiver information. 

Instead of using the `api.groups` function to retrieve the alerts I would like to stick with the `api.alerts.GetPending()` function. This ensures that we don't have duplicate alerts like mentioned [above](https://github.com/prometheus/alertmanager/pull/872#discussion_r122592149) and that we don't have to bother with `groups` and `blocks`. This was the reason why we switched to the `/alerts` endpoint in the first place, if I remember correctly. In addition it solves [this](https://github.com/prometheus/alertmanager/pull/872/files#r122640780) discussion as well. 

In order to enrich this with the receiver information I would like to use the `dispatch/route.go/Match` function. You can think of it as *taking the routing tree and pushing every alert through it once, before returning it to the UI*. Then we can both filter the alerts by receiver, if a `receiverParam` was specified, and return them attached to each alert.

What do you think? Hope I did a decent job at explaining my approach.

@fabxc I hope I summarized this correctly?

stuartnelson3:
That sounds cool to me. I was originally eyeing that function for the implementation, but then (incorrectly) thought that just flattening out the alerts from `dispatch.Groups` would be enough.

I'll make this change tomorrow and then probably have to hand the rest of the PR off to you, since I'll be on vacation for two weeks starting saturday and it sounds like some folks really want this PR.

w0rm:
I'm not sure how long the proposed changes would take. Only wondering if it is possible to merge the solution (even a suboptimal one), and then iterate on top to improve it? This feature is highly anticipated at SC.

stuartnelson3:
While this is very desired at SC, I would rather do one right solution than push out a bad one because of one company

mxinden:
@stuartnelson3 Cool, thanks! Sounds good to me.

@w0rm I agree with @stuartnelson3 here.

beorn7:
Thanks, everybody, for looking into this so carefully. While this feature is very important for our workflow (and arguably for other's, too), we should not rush something we need to change shortly after again.

w0rm:
@stuartnelson3 @mxinden alright, no pressure! Let me know if you are happy with the UI part of it or you need me to change anything.

template/default.tmpl
stuartnelson3: The template sends out receiver strings for a firing/resolved alert as configured in the alertmanager.yml config file. These are not influenced by using regexp's on the front end for filtering, so I don't think this is necessary.

stuartnelson3:
Just pushed my latest changes, seemed to be working for me using the sample config listed in the PR description.

api/api.go
```diff
@@ -287,6 +344,16 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 			break
 		}
 
+		routes := api.route.Match(a.Labels)
```
stuartnelson3: @mxinden is this what you meant?

api/api.go
```diff
@@ -320,6 +388,16 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 	respond(w, res)
 }
 
+func regexpAny(re *regexp.Regexp, ss []string) bool {
```
stuartnelson3: I'm assuming `continue` in the route config is used somewhat infrequently, so this function will probably only have a single element in `ss` for the majority of cases.

api/api.go
mxinden: @stuartnelson3 Yes, exactly. Thanks!

api/api.go
```diff
@@ -217,10 +233,15 @@ func getMeshStatus(api *API) meshStatus {
 	return strippedStatus
 }
 
-func (api *API) alertGroups(w http.ResponseWriter, req *http.Request) {
-	var err error
-	matchers := []*labels.Matcher{}
-	if filter := req.FormValue("filter"); filter != "" {
+func (api *API) alertGroups(w http.ResponseWriter, r *http.Request) {
```
mxinden: @stuartnelson3 Are the changes on the `alertGroups` function and all subsequent changes still needed?

stuartnelson3: They aren't needed, but I figured as long as the route still exists, it doesn't hurt to offer equivalent functionality. Once this route is deleted, all the code goes anyway.

mxinden: I would prefer to keep this PR as lean as possible without adding unneeded functionality on an endpoint that we want to get rid of anyways. Especially as it not only touches `api.go` but `dispatch.go` and `main.go` as well.

template/default.tmpl
beorn7: Or even if the filtering is done in the frontend, the URL param for `receiver=...` should be a regexp in any case, so the above is true in any case.

beorn7: Sorry for my confusion, but the initial example in the PR's description says it all, actually:

```
curl -sS 'http://localhost:9093/api/v1/alerts?silenced=false&receiver=webhook\d' | jq .data[].annotations.receiver
"webhook1"
"webhook1"
"webhook2"
"webhook2"
"webhook1"
"webhook1"
"webhook2"
"webhook2"
```

Thus, the backend does the filtering with a regexp (which is good), so at some point between the receiver as specified in `alertmanager.yml` and the receiver as added as a URL parameter in the template, we need to do the regexp escaping.

w0rm: Yeah, this should be escaped to create a proper link. UI does this when choosing options from a dropdown.

stuartnelson3: Ok, so to clarify: does anything need to be done, or is the regexp escaping on the front end when creating the URLs sufficient?

stuartnelson3: using receiver `web.hook` in the UI creates this url: http://localhost:9093/#/alerts?silenced=false&receiver=web%5C.hook

api/api.go
stuartnelson3: k, removed

template/default.tmpl
beorn7: A URL is created here in the template and sent out to other places (email, slack, PD), and the link will be rendeered there, so the link generated in the templating must be regexp-escaped. If that's not easy with Go templating, I guess we should have a regexp-escaped version of the receiver string ready to insert here.

beorn7: > using receiver `web.hook` in the UI creates this url: http://localhost:9093/#/alerts?silenced=false&receiver=web%5C.hook

Neat. Is this some Go templating magic? Or do we do the regexp escaping somewhere already?

beorn7: Wait a sec, that's not generated in the template, isn't it? No magic can know that the receiver param is meant to be regexp escaped.
You mean, that's just in the UI. But the template is escaped for the notification, as said above.
I think we just need `regexp.QuoteMeta` before creating the URL in the template.

w0rm: @stuartnelson3 we should escape the receiver before passing it to the template.

stuartnelson3: Ok thanks for the clarification!

stuartnelson3:
Ok, pushed the final regexp escaping change, I think this is officially "ready"

stuartnelson3:
The CI error seems to be a docker issue on circle ci:

```
Status: Downloaded newer image for quay.io/prometheus/golang-builder:1.8.3-base
Error deleting container: Error response from daemon: Driver btrfs failed to remove root filesystem 8f6731d3d7c8ec3386f6f1e76d8c785f463793db18098456ca6b4e46a48b201a: Failed to destroy btrfs snapshot /var/lib/docker/btrfs/subvolumes for 8f6731d3d7c8ec3386f6f1e76d8c785f463793db18098456ca6b4e46a48b201a: operation not permitted
!! The base builder docker image exited unexpectedly: [255] Error deleting container: Error response from daemon: Driver btrfs failed to remove root filesystem 8f6731d3d7c8ec3386f6f1e76d8c785f463793db18098456ca6b4e46a48b201a: Failed to destroy btrfs snapshot /var/lib/docker/btrfs/subvolumes for 8f6731d3d7c8ec3386f6f1e76d8c785f463793db18098456ca6b4e46a48b201a: operation not permitted
```

w0rm:
@stuartnelson3 yeah this is what I got
screen shot 2017-06-23 at 16 08 54

@mxinden are you happy with the changes?

stuartnelson3:
Tests are green, circle-ci seems to have fixed it's stuff

ui/app/src/Views/AlertList/Updates.elm
```diff
@@ -47,9 +48,26 @@ update msg ({ groupBar, filterBar } as model) filter apiUrl basePath =
...
+                        ]
                     )
 
+            ReceiversFetched (Success receivers) ->
+                ( { model | receivers = receivers }, Cmd.none )
+
+            ToggleReceivers show ->
+                ( { model | showRecievers = show }, Cmd.none )
+
+            ReceiversFetched _ ->
```
mxinden: Can we move this underneath the first `ReceiversFetched` definition?

template/default.tmpl
mxinden: For documentation purposes. This is addressed in bb2d1b5716d1d19511ed323495b5417de5716451

ui/app/src/Views/AlertList/Updates.elm
w0rm: Sure, done!

w0rm:
@mxinden yeah, I believe this is good enough to bring back the missing feature. We can do another iteration over the UI in a separate pull request.

mxinden:
@w0rm Perfect.

I would like to give @fabxc and @beorn7 a chance to give their approval as well. I will merge later today if there are no objections.

beorn7:
I don't have the cycles for a code-level review, but happy to try this out here at SC.
I'd say, let's merge but not yet release.

mxinden:
@beorn7 Thanks, that would be great, if you could test this out at SC. Looking forward to the feedback!

beorn7:
Bleeding edge HEAD build running at SC. Looks good so far.

PR #885, opened 2017-06-26
Add notification for ChatWork

griffin-stewie:
@stuartnelson3 

I added [ChatWork](http://www.chatwork.com/) integration based on HipChat notification.


griffin-stewie:
It works fine if I run `make test` locally. Does anybody figure out the reason of failure?

mxinden:
@griffin-stewie Tests seemed to have timed out. Haven't seen that before. I have restarted them.

mxinden:
@griffin-stewie I am sorry for not further following up on this PR. Thanks a lot for your contribution.

@fabxc @brian-brazil Is ChatWork an integration we want to maintain? If so I would move on here and help @griffin-stewie get this up, running and green.

stuartnelson3:
I'm not familiar with ChatWork, but I think our general recommendation is to first use a webhook. Do we have an officially published position on this? A quick scan in the documentation only shows:

>Other receiver implementations available in version 0.0.4 of Alertmanager are not implemented yet. We are gladly accepting any contributions to add them to the new implementation.

https://prometheus.io/docs/alerting/configuration/#receiver

We might want to provide some additional guidance as to how we choose receivers, as the most common PRs we get are implementing new ones.

brancz:
IIRC someone wanted to work on a webhook framework to easily implement webhook servers.

griffin-stewie:
How about this? AlertManager executes registered commands. I have no idea how to register but like writing configuration file or specific directory, defines naming rule `alertmanager-XXXX` just like git subcommands.

maybe you can use same input as webhook for command pass it via STDIN.

This plan is handy and easy to use for user that can't have server to receive webhook, and less management.

brancz:
Personally I have no problem with ABI's, but I don't think this is the right application for it, it would cause an explosion of unmaintained container images floating around.

PR #892, opened 2017-07-03
Support for custom SMTP hello string

ideaship:
Some MTAs insist that they be greeted with a fully qualified domain
name. The default provided by the net/smtp library, "HELLO localhost",
is not sufficient and will result in rejected messages.

This changeset adds a new configuration option that allows the
alertmanager to do its job in such an environment.

Considering that this is a rather trivial PR, I am following the guidelines for contributing by addressing @stuartnelson3.


ideaship:
@stuartnelson3 Per request, I added a test that will fail if the file is not parsed correctly.

PR #893, opened 2017-07-04
Add User-Agent for webhook requests

kfdm:
Using https://github.com/prometheus/prometheus/pull/2447/files as an example, this adds a user agent for Webhook requests.

This is the first time I've written code in go, so please let me know what I can fix up.
It's likely that the same user agent could be used for the other notifiers but for now I have only added this to the webhook notifier.

@stuartnelson3

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -33,6 +33,7 @@ import (
 
 	"github.com/prometheus/common/log"
 	"github.com/prometheus/common/model"
+	"github.com/prometheus/common/version"
 	"golang.org/x/net/context"
 	"golang.org/x/net/context/ctxhttp"
 
@@ -135,6 +136,8 @@ func BuildReceiverIntegrations(nc *config.Receiver, tmpl *template.Template) []I
 
 const contentTypeJSON = "application/json"
 
+var userAgentHeader = fmt.Sprintf("Alertmanager/%s", version.Version)
+
 // Webhook implements a Notifier for generic webhooks.
 type Webhook struct {
 	// The URL to which notifications are sent.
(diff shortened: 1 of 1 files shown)


notify/impl.go
```diff
@@ -135,6 +136,8 @@ func BuildReceiverIntegrations(nc *config.Receiver, tmpl *template.Template) []I
 
 const contentTypeJSON = "application/json"
 
+var userAgentHeader = fmt.Sprintf("Alertmanager/%s", version.Version)
```
stuartnelson3: Could you put this and the preceeding line into a `const` declaration block? 

```go
const (
        contentTypeJSON = "application/json"
        userAgentHeader = fmt.Sprintf("Alertmanager/%s", version.Version)
)
```

kfdm: I tired that but I got this error

```
# github.com/prometheus/alertmanager/notify
notify/impl.go:139: const initializer fmt.Sprintf("Alertmanager/%s", version.Version) is not a constant
!! command failed: build -o /Users/st20638/.golang/src/github.com/prometheus/alertmanager/alertmanager -ldflags -X github.com/prometheus/alertmanager/vendor/github.com/prometheus/common/version.Version=0.7.1 -X github.com/prometheus/alertmanager/vendor/github.com/prometheus/common/version.Revision=e60ffbe6fb8c58f45864fccd1c741219402c2617 -X github.com/prometheus/alertmanager/vendor/github.com/prometheus/common/version.Branch=user-agent -X github.com/prometheus/alertmanager/vendor/github.com/prometheus/common/version.BuildUser=st20638@LF02Q331602A96M.local -X github.com/prometheus/alertmanager/vendor/github.com/prometheus/common/version.BuildDate=20170710-23:55:18  -s -a -tags netgo github.com/prometheus/alertmanager/cmd/alertmanager: exit status 2
make: *** [build] Error 1
```

I'm not familiar with Go, so I'm not sure how to fix that. When I tried this, it seems to work though

```
const (
	contentTypeJSON = "application/json"
	userAgentHeader = `Alertmanager/{{version.Version}}`
)
```

but again, I'm not familiar enough with Go to know if that's acceptable or not.

kfdm: Tried to read up on things, but since const is checked at compile time, but it seems like version.Version is filled in by the linker, it may not be possible to have userAgentHeader as a const (and in the linked PR for Prometheus it's a var there as well)

brian-brazil: Yes, consts in Go can't be calculated unfortunately.

stuartnelson3: Oh my mistake, sorry.

PR #897, opened 2017-07-06
Added possibility to have a global victorops api_key

caarlos0:
If the idea is to setup one victorops_receiver per team or something like that, this would allow the user to not repeat the API key for each of them.

Since the API key is the same for all integrations, this seemed like a good idea.

I didn't added a test for that because there is none already, but I can add new tests if you want :)

Thanks for the awesome tools btw 🍻

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -131,7 +131,6 @@ func TestJSONUnmarshalMarshaled(t *testing.T) {
 }
 
 func TestEmptyFieldsAndRegex(t *testing.T) {
-
 	boolFoo := true
 	var regexpFoo Regexp
 	regexpFoo.Regexp, _ = regexp.Compile("^(?:^(foo1|foo2|baz)$)$")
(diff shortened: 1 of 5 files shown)


stuartnelson3:
Looks great! Could you add a quick test that parses an example config.yaml and verifies the behavior?

caarlos0:
@stuartnelson3 thanks, done!

Found a bug while doing that , BTW :P

config/notifiers.go
```diff
@@ -340,9 +340,6 @@ func (c *VictorOpsConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
 	if err := unmarshal((*plain)(c)); err != nil {
 		return err
 	}
-	if c.APIKey == "" {
-		return fmt.Errorf("missing API key in VictorOps config")
-	}
```
caarlos0: this is now being validated when the global key is set.

config/config_test.go
```diff
@@ -205,3 +204,28 @@ func TestEmptyFieldsAndRegex(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("Error parsing %s: %s", "testdata/conf.victorops-default-apikey.yml", err)
+	}
+
+	var defaultKey = conf.Global.VictorOpsAPIKey
+	if defaultKey != conf.Receivers[0].VictorOpsConfigs[0].APIKey {
+		t.Errorf("Invalid victorops key: %s\nExpected: %s", conf.Receivers[0].VictorOpsConfigs[0].APIKey, defaultKey)
+	}
+	if defaultKey == conf.Receivers[1].VictorOpsConfigs[0].APIKey {
+		t.Errorf("Invalid victorops key: %s\nExpected: %s", conf.Receivers[0].VictorOpsConfigs[0].APIKey, "qwe456")
```
stuartnelson3: perfect, thanks for testing that a team can override the global!

caarlos0:
Thanks @stuartnelson3 !

Any ETA for a release?

stuartnelson3:
Unfortunately I don't know :/ We have a milestone for the new release, but it's mostly frontend stuff: https://github.com/prometheus/alertmanager/milestone/5

The best thing to do would be to compile the project yourself right now. If you have go installed, you should be able to clone the repo and create your binary without too much trouble.

caarlos0:
ok, np :) 

thanks!

mxinden:
@caarlos0 See release PR #906

caarlos0:
Neat!

PR #900, opened 2017-07-10
Receiver autocomplete

w0rm:
The receiver dropdown had too many results at SC and it was not possible to enter any arbitrary regex into the field. This pull request adds an autocomplete with a possibility to input any regex.


ui/app/src/Alerts/Types.elm
```diff
@@ -18,17 +15,7 @@ type alias Alert =
     }
 
 
-type alias AlertGroup =
```
w0rm: `AlertGroup` and `Block` were not used anywhere

mxinden: Thanks for cleaning this up. This is a left over from switching to the /alerts endpoint. My bad.

ui/app/src/Views/AlertList/Views.elm
```diff
@@ -91,6 +91,12 @@ alertGroups activeId filter groupBar alerts =
                         (alertList activeId labels filter)
                         (Dict.get labels grouped)
                 )
+            |> (\list ->
+                    if List.isEmpty list then
+                        [ Utils.Views.error "No alerts found" ]
```
mxinden: 👍

ui/app/src/Utils/Match.elm
```diff
@@ -1,4 +1,4 @@
-module Views.GroupBar.Match exposing (jaro, jaroWinkler, commonPrefix)
+module Utils.Match exposing (jaro, jaroWinkler, commonPrefix)
```
stuartnelson3: great!

w0rm:
@mxinden the dropdown only shows top 10 results, the same as in group by.

ui/app/src/Views/AlertList/Views.elm
w0rm: Yeah, we had this message in different place and it was never triggered. Now it is

w0rm:
@mxinden feel free to create an issue to display receivers, but maybe we should hide this in the "info" collapsible section.

PR #909, opened 2017-07-16
Add current time to go webhook standard output

mxinden:
This is just a small addition for better debugging of notifications send
by Alertmanager.


mxinden:
@stuartnelson3 Good catch. Thanks. Adjusted accordingly.

PR #906, opened 2017-07-15
*: cut v0.8.0

mxinden:
* [FEATURE] Add ability to filter alerts by receiver in the UI (#890)
* [FEATURE] Add User-Agent for webhook requests (#893)
* [ENHANCEMENT] Add possibility to have a global victorops api_key (#897)
* [ENHANCEMENT] Add EntityDisplayName and improve StateMessage for Victorops
  (#769)
* [ENHANCEMENT] Omit empty config fields and show regex upon re-marshalling to
  elide secrets (#864)
* [ENHANCEMENT] Parse API error messages in UI (#866)
* [ENHANCEMENT] Enable sending mail via smtp port 465 (#704)
* [BUGFIX] Prevent duplicate notifications by sorting matchers (#882)
* [BUGFIX] Remove timeout for UI requests (#890)
* [BUGFIX] Update config file location of CLI in flag usage text (#895)

PR #910, opened 2017-07-16
Update matching algorithm

stuartnelson3:
Discussion around this emerged in #906

This does a few things:

1. The algorithm wasn't appending search matches to the accumulator `acc` in the `cp` method. Strings were being searched through and values were being returned, but `acc` wasn't being used for anything. Now it's storing the values as we recurse through the string, and ending recursion when a character in the search term isn't found consecutively.
2. Consecutive character calculation is no longer limited to 4 characters. The more consecutive characters are, the value should (in my opinion) keep being ranked more highly.
3. I weighted consecutive values more highly in the overall calculation of a match (`0.1` -> `0.25`). It seems to "work", but this value is up for tweaking.


w0rm:
Works better! I could type in `warn` and got all the receivers that end with `-warn`

mxinden:
I am facing difficulties with this.

With the following config:
```
global:
  resolve_timeout: 5m

route:
  group_by: ['alertname']
  group_wait: 10s
  group_interval: 10s
  repeat_interval: 1h
  receiver: 'web.hook2'
  routes:
  - match:
      dev: 'sda1'
    receiver: 'web.hook1'
receivers:
- name: 'web.hook2'
  webhook_configs:
- name: 'web.hook1'
  webhook_configs:
inhibit_rules:
  - source_match:
      severity: 'critical'
    target_match:
      severity: 'warning'
```
If I type `1` into the edit field, `All`, `web.hook1` and `web.hook2` are displayed, instead of just `web.hook1`.

Am I misunderstanding something here? Are you able to reproduce this?

stuartnelson3:
The values are sorted by how well they match, they aren't eliminated if they have no matching character. The changes in this PR fiddle with the weighting of search results, but the functionality remains unchanged.

PR #912, opened 2017-07-20
When I notify alert to slack, I want to use this option.

9to6:
Hi.
Thank you for your effort.

I use Alertmanager with Slack.
When I received messages, I wanna use this option for slack.

**link_names**

You can check this option what is.

https://api.slack.com/methods/chat.postMessage

I updated some codes a little bit.

**Usage**

```yaml
slack_configs: 
   link_names: true
```

I think this options is so important for alerting. 
I think most people want to notify them to their owners when they receive messages on Slack.

Thanks.

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -218,6 +218,7 @@ type SlackConfig struct {
 	Fallback  string `yaml:"fallback,omitempty" json:"fallback,omitempty"`
 	IconEmoji string `yaml:"icon_emoji,omitempty" json:"icon_emoji,omitempty"`
 	IconURL   string `yaml:"icon_url,omitempty" json:"icon_url,omitempty"`
+	LinkNames string `yaml:"link_names,omitempty" json:"link_names,omitempty"`
 
 	// Catches all undefined fields and must be empty after parsing.
 	XXX map[string]interface{} `yaml:",inline" json:"-"`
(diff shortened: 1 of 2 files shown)


9to6:
Hi. Where is a question?

stuartnelson3:
Sorry about that! Must have disappeared somehow. Is `link_names` a string or a bool? From the documentation it looks like it might be a bool?

9to6:
image
This option's default value is false. 
When I use this option, I can input value, number 0 or 1, boolean true or false,
but when I test this option, It cannot allow **"false"**, type is String
When this option set the value, string value **"false"**, I think this option is regarded set **True**.

stuartnelson3:
In that case, can you change the type to boolean? It probably interprets any string value as truthy.

9to6:
I changed type to bool.
But I included invalid commit log inadvertently.

mxinden:
@9to6 Thank you very much for your contribution! Can you create a new branch right off of master and cherry-pick your two commits onto it?

Something along the lines of: (No guarantee, please backup your changes before)
```
git checkout master
git pull upstream master
git checkout -b slack_mention2
git cherry-pick fc2ab29b6dee17f7612958345c1d22e5d311eaab
git cherry-pick 18041521610b56f6f4eed39024d56a27ec1c967b
git push <your github repo alias> slack_mention2:slack_mention -f
```

9to6:
@mxinden Thank you for describing detailed.
I applied that.

mxinden:
@9to6 Thanks for keeping the history clean. @stuartnelson3 Feel free to merge if you have no objections.

Issue #914, opened 2017-07-20
alertmanager crashes when receive alert.


pytker:
CentOS release 6.8   2.6.32-642.4.2.el6.x86_64

prometheus 1.5.2
alertmanager 0.7.1

prometheus config:
```

global:
  external_labels:
    env: "test"

rule_files:
  - "/opt/prometheus/prometheus-1.5.2.linux-amd64/configs/alerts.yml"

scrape_configs:
  - job_name: nodes
    static_configs:
      - targets:
        - *******************************
        - *******************************
        - *******************************
        - *******************************
        labels:
          exporters:  node
```

alerts:
```

ALERT low_disk_space
  IF (node_filesystem_avail * 100)/node_filesystem_size <= 20
  FOR 5m
  LABELS { severity = "warning" }
  ANNOTATIONS {
    summary = "Less than 20% left on disk {{$labels.mountpoint}} on {{$labels.instance}}"
  }

```
alertmanager config

```
global:
  smtp_smarthost: 'smtp.domain.ru:25'
  smtp_from: 'alertmanager@domain.ru'

route:
  group_by: ['alertname', 'service', 'severity']
  group_wait: 30s
  group_interval: 5m
  repeat_interval: 3h
  receiver: omni_adm
  routes:

inhibit_rules:
- source_match:
    severity: 'critical'
  target_match:
    severity: 'warning'
  equal: ['alertname', 'service']

receivers:
- name: 'adm'
  email_configs:
  - to: 'admin.domain.ru'
```

log

```
time="2017-07-20T11:14:57+03:00" level=info msg="Listening on :9084" source="main.go:308"
time="2017-07-20T11:15:37+03:00" level=debug msg="Received alert" alert=low_disk_space[0426069][active] component=dispatcher source="dispatch.go:183"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x78 pc=0x8c762f]

goroutine 117 [running]:
main.meshWait.func1(0x18)
        /go/src/github.com/prometheus/alertmanager/cmd/alertmanager/main.go:349 +0x3f
main.main.func6(0x45d964b800, 0xc4201aa960)
        /go/src/github.com/prometheus/alertmanager/cmd/alertmanager/main.go:229 +0x49
github.com/prometheus/alertmanager/dispatch.(*aggrGroup).run(0xc4200f6e10, 0xc4201b1710)
[...]

PR #919, opened 2017-07-22
Fix crash when no mesh router is configured

juliusv:
This adds a check for `meshListen != ""` around the waitFunc code as we
have around the other mesh-related code parts above.

Fixes https://github.com/prometheus/alertmanager/issues/914


stuartnelson3:
Looking into the bindata failure ..

stuartnelson3:
I had to clean and completely remake the project in order to get these autogenerated changes

Issue #918, opened 2017-07-21
Crash on Status page when mesh is disabled


roman-vynar:
This commit https://github.com/prometheus/alertmanager/commit/02484d18b012f4e7d03bacb80d8fb4c454427e8c allowed to disable mesh stuff with `-mesh.listen-address=""`.
However, if you do that and go to Status web page, alertmanager will crash:
```
time="2017-07-21T21:19:06Z" level=info msg="Starting alertmanager (version=0.8.0, branch=HEAD, revision=74e7e48d24bddd2e2a80c7840af9b2de271cc74c)" source="main.go:109" 
time="2017-07-21T21:19:06Z" level=info msg="Build context (go=go1.8.3, user=root@439065dc2905, date=20170720-14:14:06)" source="main.go:110" 
time="2017-07-21T21:19:06Z" level=info msg="Loading configuration file" file="/etc/alertmanager.yml" source="main.go:234" 
time="2017-07-21T21:19:06Z" level=info msg="Listening on 127.0.0.1:9093" source="main.go:308" 
2017/07/21 21:20:20 http: panic serving 127.0.0.1:55424: runtime error: invalid memory address or nil pointer dereference
goroutine 1426 [running]:
net/http.(*conn).serve.func1(0xc420074640)
	/usr/local/go/src/net/http/server.go:1721 +0xd0
panic(0x952ae0, 0xd59b50)
	/usr/local/go/src/runtime/panic.go:489 +0x2cf
github.com/prometheus/alertmanager/vendor/github.com/weaveworks/mesh.NewStatus(0x0, 0xc42058c000)
	/go/src/github.com/prometheus/alertmanager/vendor/github.com/weaveworks/mesh/status.go:37 +0x37
github.com/prometheus/alertmanager/api.getMeshStatus(0xc420113000, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/go/src/github.com/prometheus/alertmanager/api/api.go:218 +0x72
github.com/prometheus/alertmanager/api.(*API).status(0xc420113000, 0xc83d80, 0xc4200fc028, 0xc4202ca500)
	/go/src/github.com/prometheus/alertmanager/api/api.go:197 +0xa5
github.com/prometheus/alertmanager/api.(*API).(github.com/prometheus/alertmanager/api.status)-fm(0xc83d80, 0xc4200fc028, 0xc4202ca500)
[...]

PR #920, opened 2017-07-22
Stn/status page ohne mesh

stuartnelson3:
How does this look?

image


stuartnelson3:
I'm not sure, but I believe the default is for a server to start a mesh with itself. That way, other servers can be brought up that will connect to this initial server.

mxinden:
@stuartnelson3 Sounds reasonable. Thanks.

mxinden:
Fixes #918

PR #927, opened 2017-07-24
Add production users list

mxinden:
As discussed in [1] I would like to add a list of production users of
Alertmanager to this repo. On the one hand this would show the level of trust people
put into the project. On the other hand it gives us the ability to reach
out to this user group for advice and experience.

Please suggest other possible metrics that you would be interested in, or mention those that you would rather like to have removed.

Once we have settled on a set of metrics I will reach out to those companies that we are aware of.

[1]
https://github.com/coreos/etcd/blob/master/Documentation/production-users.md

//CC @brancz

### DIFF
--- a/doc/production-users.md
+++ b/doc/production-users.md
@@ -1,0 +1,23 @@
+# Production Users
+
+This document lists usages of the Alertmanager in production environments. The
+purpose of the list is to illustrate possible use case scenarios of
+Alertmanager. In addition it enables contributors to reach out to this user
+group for advice and experiences.
+
+Please feel free to contact us, if you would like to be listed here.
+
+
+## Name
+
+- **Monitored applications**: *e.g.* etcd cluster & Ruby on Rails web application
+- **Number of monitored machines**: *e.g.* 100
+- **Alertmanager cluster size**: *e.g.* 3
+- **Prometheus cluster size**: *e.g.* 2
+- **Launched**: *e.g.* 2017-07-24
+- **Platform**: *e.g.* AWS, bare metal, or ...
+- **Environment**: *e.g.* Kubernetes, self-build, or ...
+- **Integrations with**: *e.g.* Slack, PagerDuty, or ...
+- **Number of silences**: *e.g.* 20
+
+Further description ...


brancz:
As a lot of these questions are about the Prometheus setup I wonder whether we should just have a document in the Prometheus repo regarding this. Or maybe even on prometheus.io, as it might be easier to explain the usage across multiple tools?

If you want to keep it repository focused I'd suggest to scope the questions more towards the actual Alertmanager usage.

brian-brazil:
There is already a user list on the front page of prometheus.io, and for more detailed reports we have the use case blogs.

brancz:
Agreed, but I do miss a document that is somewhere in-between those two. Users with a few main bullet points on how their setup looks like, I'm sure we can extract a lot of data from the detailed reports we have in form of blog posts.

mxinden:
I agree with @brian-brazil that we already have a lot of information around production use cases. What I am missing is a simple table or list of use cases where it is easy to compare different setups with a basic set of metrics.

@brancz I am happy to make this more open and move it to a different place. But I would like to keep a basic set of information around Alertmanager in it.

brancz:
I think it would be great to have a quick concise overview of users' setups.

fabxc:
The users list on the website is merely a very incomplete collection of logos. Anyone seriously about evaluating using Prometheus components will not gain anything from a cloud of images.

In other news, our website should fan in docs from our component repos. I think that is part of the effort @Gouthamve and @grobie are currently doing towards versioned documentation.
The repository itself is the right place for this and a textual overview of users with some background info on deployment size etc. is the right thing as well.

mxinden:
Adding a *production users list* here which is then displayed on the *Prometheus.io* website sounds good to me.

@fabxc What do you think of my proposed facts that we would collect from each user? @brancz You said you had some that you would like to add?!

Instead of collection these metrics for each Prometheus project separately, we could have them in one place. Would reduce the amount of repetition. E.g. 'How many nodes are you monitoring?', which is probably interesting for both the Alertmanager production users list, as well as the Prometheus production users list.

fabxc:
I think we'll boil down to Prometheus and Alertmanager for this. Exporters do not need those user summaries.
I think with focus on the right questions, the data we want to gather will have little overlap. E.g. nodes monitored does have little impact on Alertmanager. The performance benchmark is rather sent alerts – even that is probably secondary to across how many teams and channels notifications are organized.

Anyway, the little duplication we'll have doesn't hurt IMO – especially as Prometheus and AM are and always will be relatively independent tools.

brancz:
What @fabxc mentioned is exactly the data I was thinking of. In general agreed with all what @fabxc said.

brian-brazil:
If we're going to collect this sort of information, I'd suggest having one central list that covers all the components.

The wiki is probably the best place, as that's the lowest friction.

brian-brazil:
> Nobody watches the wiki – like ever. It's a completely distinct work flow from our usual repositories. I don't know when the last changes to anything in the wiki were made or what they would've been.

People actively watching changes to these docs is not required, people being able to find them and change them easily is what's needed here. It works for the port list, and I think it could also work well here.

> All docs should go into the repo – developer as well as user ones.

This is neither user nor developer docs, it's a list of users.

brancz:
> This is neither user nor developer docs, it's a list of users.

I disagree. This is very much user documentation, mainly for people who are evaluating Prometheus and are looking into whether it fits their need and/or whether it works out at other organizations and at which scale. All of that in a simple concise list on the website seems like the best place to me.

I'm not a particularly huge fan of the port list or see it's importance and I am not able to follow the changes that happen to it reliably.

brian-brazil:
> All of that in a simple concise list on the website seems like the best place to me.

The issue with something like this being directly on the website is that we'll get much fewer contributions, as the user has to go to the effort of creating a PR. You'll note how rarely we get new logos on the frontpage.

> I'm not a particularly huge fan of the port list or see it's importance and I am not able to follow the changes that happen to it reliably.

It comes from a time when the official components had clashing port numbers, and inadvertently grew from there to a comprehensive list of exporters and integrations. It's useful in practice, and if it ain't broke don't fix it.

mxinden:
Closing here due to missing consensus.

PR #930, opened 2017-07-27
*: configure http client from config

### LINKED ISSUE #853: What's the correct way to config Alertmanager to use a forward proxy?
In this [thread](https://github.com/coreos/prometheus-operator/issues/399#issuecomment-307114123) with @brancz, I was seeking a way to config Alertmanager to use a forward proxy. One way is to use the `https_proxy` env var detailed [here](https://github.com/prometheus/alertmanager/issues/628).

Is this the correct way to do it in the long haul, or should it be in Alertmanager's config. Thanks!

brancz:
It seems that many people are interested in configuring the HTTP client used to perform requests to the notification providers. As @Conorbro recently added support for this in the common/config package, I made use of it here.

Closes #853

@stuartnelson3 @fabxc @mxinden @Gouthamve @brian-brazil

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -17,6 +17,8 @@ import (
 	"fmt"
 	"strings"
 	"time"
+
+	commoncfg "github.com/prometheus/common/config"
 )
 
 var (
@@ -176,6 +178,8 @@ func (c *EmailConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 type PagerdutyConfig struct {
 	NotifierConfig `yaml:",inline" json:",inline"`
 
+	HTTPConfig *commoncfg.HTTPClientConfig `yaml:"http_config,omitempty" json:"http_config,omitempty"`
+
 	ServiceKey  Secret            `yaml:"service_key,omitempty" json:"service_key,omitempty"`
 	URL         string            `yaml:"url,omitempty" json:"url,omitempty"`
 	Client      string            `yaml:"client,omitempty" json:"client,omitempty"`
@@ -203,6 +207,8 @@ func (c *PagerdutyConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
 // SlackConfig configures notifications via Slack.
 type SlackConfig struct {
 	NotifierConfig `yaml:",inline" json:",inline"`
+
+	HTTPConfig *commoncfg.HTTPClientConfig `yaml:"http_config,omitempty" json:"http_config,omitempty"`
 
 	APIURL Secret `yaml:"api_url,omitempty" json:"api_url,omitempty"`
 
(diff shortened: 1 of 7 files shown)


config/notifiers.go
```diff
@@ -176,6 +178,8 @@ func (c *EmailConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 type PagerdutyConfig struct {
 	NotifierConfig `yaml:",inline" json:",inline"`
 
+	HTTPConfig *commoncfg.HTTPClientConfig `yaml:"http_config,omitempty" json:"http_config,omitempty"`
```
brian-brazil: Should this go in NotifierConfig considering we use it everywhere?

notify/impl.go
```diff
@@ -141,13 +142,13 @@ var userAgentHeader = fmt.Sprintf("Alertmanager/%s", version.Version)
 // Webhook implements a Notifier for generic webhooks.
 type Webhook struct {
 	// The URL to which notifications are sent.
```
brian-brazil: Comment needs an update

brancz: Not sure why the URL was exported before, I moved the `Config` -> `conf` to be consistent with all the other implementations.

config/notifiers.go
brancz: We have it everywhere except email configs, it would be odd to be able to have it there, no?

brian-brazil: That's true.

brancz:
If there are no general concerns anymore I'll start working on docs.

brancz:
I just talked to @fabxc, he has some comments on the general topic of the common package. I'll hold off on merging util he has a look.

fabxc:
I've plenty of concerns on this – but that's a longer discussion I hope to find time for soonish.

Where else is this struct being used outside of AM?

brian-brazil:
Right now the blackbox exporter. We likely also need it in the consul exporter.

brancz:
@fabxc @brian-brazil how should we continue on this, as it should be possible for people to configure this somehow.

brian-brazil:
I've been proposing code in common for this for 18 months now, which is what the blackbox exporter uses. @fabxc is against us having common packages for this, while I strongly believe code like this (particularly security related) must come from a common library.

brancz:
> I've plenty of concerns on this – but that's a longer discussion I hope to find time for soonish.

I guess this would be the start of having this discussion, so we can come to an agreement of how to proceed.

simonpasquier:
To set the discussion, here is a recap of the concrete use cases that call for more flexibility on the HTTP client configuration:

* Use bearer tokens for web-hook services running behind proxies like [kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) and [oauth-proxy](https://github.com/openshift/oauth-proxy).
* Override TLS parameter (skip certificate verification, custom CA, ...).
* Use HTTP proxies.

brian-brazil:
We're all good with the use cases, it's exactly how we share the common code that there is disagreement on.

ddewar2k:
Moving my question to this more recent thread. I still don't see code that supports http proxy for receivers and it looks like all the threads on this closed. Is my understanding correct?

simonpasquier:
@brancz @brian-brazil @fabxc Now that Prometheus itself uses http client from the common library  (see https://github.com/prometheus/prometheus/pull/3674), can we reconsider this PR?
FWIW I've rebased the PR on the master branch in my local fork and after resolving the conflicts, it seems to be working fine.

brancz:
I'm happy with going forward with this.

simonpasquier:
@brancz, cool! This is your PR rebased on master: https://github.com/prometheus/alertmanager/compare/master...simonpasquier:httpcfg
If you don"t have time, I can carry on the work.

brancz:
@simonpasquier I don't know when I would get to it, so if you can, go for it!

simonpasquier:
@brancz I've created #1222.

Issue #887, opened 2017-06-26
Make label text selectable


beorn7:
After discussing it with @w0rm , this seems to be a hard one…

Let me explain:

A very common workflow is the following:
- I'm looking at an alert with an instance label, e.g. `instance="10.20.30.40:9100"`.
- Now I want to ssh into the instance, so I select `10.20.30.40` and use my favorite copy&paste key sequences to get `ssh 10.20.30.40` into my command line.

The problem with the new UI is that all the labels are buttons, and text on buttons is not selectable.

My work-around is the following:
- Click on the label, e.g. `instance="10.20.30.40:9100"`, which pushes that label into the filter box.
- Click on the filter item, which makes it editable.
- Select the now editable text.

This is not only cumbersome, it is also very hard to discover for new users.

Is there an HTML hack to make text on buttons selectable, or work around the problem in some other way?
In principle, clicking a button is a distinctly different stream of events from selecting text on a button (button down - no movement - button release vs. button down - some movement within button text - button release). Fundamentally, it should be possible.

w0rm:
What if we rendered each label as a label + button, where a label would not be clickable? We could also limit the length and show three dots (plus the full label in a tooltip).

fabxc:
Is it not possible to just make it a styled `<span>` with an on-click action attached?

w0rm:
@fabxc this should be possible, but I'm not sure if it affects accessibility.

w0rm:
Closed via #932

PR #932, opened 2017-07-28
Selectable labels

### LINKED ISSUE #886: Make the grouping label clickable, too (to move them into the filter box).
In the view of alerts, if I click on a grey label, it is pushed into the filter box, which is quite handy.
However, I would expect the same to happen if I click on a blue label (i.e. a label that is currently set as a grouping label). But nothing happens then.

w0rm:
I think that the best way to address #887 is to have split buttons, with a text portion and a clickable element. This way we don't mess with accessibility.

I also fixed #886, because the grouping labels are very similar to the ones from the alert list.


stuartnelson3:
Would be good to get some other opinions:

@beorn7 @grobie @mxinden @fabxc @brancz @brian-brazil if any of you have time to check out the new button interaction, would be cool

mxinden:
@w0rm I think this is a very sound solution to #887.

As a usability suggestion, can we replace the *plus* symbol by a *filter* symbol like e.g. http://fontawesome.io/icon/filter/? In addition we could put the symbol next to the string *Filter* in the tab component. This connects the two, so users know if they click the filter icon on a label, it will be added to the filter tab on the top. What do you think?

w0rm:
@mxinden I don't like this icon, I think it doesn't represent an action of adding a matcher to the filter list. Alternatively we could use a magnifying glass with a plus, but I think a tooltip is enough.

grobie:
I didn't immediately grasp the elements, but it seems to be a pretty clear interface actually. I find the plus sign a clear icon for the action. In case there is a more common one for that scenario, I wouldn't mind though.

Feature request: It would be helpful to easily select the label value with a single click/double click. Maybe by changing the selection behavior so that a double click inside the value string selects everything between the quotes.

stuartnelson3:
@grobie feature requests belong in issues :)

stuartnelson3:
Unless we here a strong "no" from someone, I'm going to merge this in 24 hours. I think we agree that it's an improvement, and if we decide that a different icon would be preferable, we can easily change that.

Issue #290, opened 2016-03-31
Feature: Support multipart mail


RichiH:
@fabxc as per IRC

RichiH:
When this is touched, supporting custom headers would also make sense.

brian-brazil:
Custom headers are already supported.

RichiH:
For documentation if anyone finds this issue: https://prometheus.io/docs/alerting/configuration/#email_config

doctorlector:
That documentation for creating e-mail custom templates isn't very helpful at all.

brian-brazil:
Text parts are support these days.

PR #934, opened 2017-07-31
added email notification text content support

awaragi:
added email notification text content support (configuration text)
added default email notification text content default value empty
converted email notification from html to multipart/alternatives with support to both html (first) and text (last)
ignore intellij IDE .idea working folder

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -33,6 +33,7 @@ var (
 			VSendResolved: false,
 		},
 		HTML: `{{ template "email.default.html" . }}`,
+		TEXT: ``,
 	}
 
 	// DefaultEmailSubject defines the default Subject header of an Email.
@@ -142,6 +143,7 @@ type EmailConfig struct {
 	AuthIdentity string            `yaml:"auth_identity,omitempty" json:"auth_identity,omitempty"`
 	Headers      map[string]string `yaml:"headers,omitempty" json:"headers,omitempty"`
 	HTML         string            `yaml:"html,omitempty" json:"html,omitempty"`
+	TEXT         string            `yaml:"text,omitempty" json:"text,omitempty"`
 	RequireTLS   *bool             `yaml:"require_tls,omitempty" json:"require_tls,omitempty"`
 
 	// Catches all undefined fields and must be empty after parsing.
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -365,14 +365,18 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 		fmt.Fprintf(wc, "%s: %s\r\n", header, mime.QEncoding.Encode("utf-8", value))
 	}
 
-	fmt.Fprintf(wc, "Content-Type: text/html; charset=UTF-8\r\n")
 	fmt.Fprintf(wc, "Date: %s\r\n", time.Now().Format(time.RFC1123Z))
+	boundary := "=_NextPart_57bb889e057d551696ff582664f96f502da361c88cdc32a9e603f0af4b1c2b17" // SHA-256 of random string
+	fmt.Fprintf(wc, "Content-Type: multipart/alternative;  boundary=\"%s\"\r\n", boundary)
+
 
 	// TODO: Add some useful headers here, such as URL of the alertmanager
 	// and active/resolved.
 	fmt.Fprintf(wc, "\r\n")
 
-	// TODO(fabxc): do a multipart write that considers the plain template.
+	// Html template
+	fmt.Fprintf(wc, "--%s\r\n", boundary)
+	fmt.Fprintf(wc, "Content-Type: text/html; charset=UTF-8\r\n")
 	body, err := n.tmpl.ExecuteHTMLString(n.conf.HTML, data)
 	if err != nil {
 		return false, fmt.Errorf("executing email html template: %s", err)
@@ -381,6 +385,23 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 	if err != nil {
 		return true, err
 	}
+
+	// Text template
+	// Last alternative based on recommendation in section 7.2.3
+	// https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html
+	fmt.Fprintf(wc, "--%s\r\n", boundary)
+	fmt.Fprintf(wc, "Content-Type: text/plain; charset=UTF-8\r\n\r\n")
+	body, err = n.tmpl.ExecuteTextString(n.conf.TEXT, data)
+	if err != nil {
+		return false, fmt.Errorf("executing email text template: %s", err)
+	}
+	_, err = io.WriteString(wc, body)
+	if err != nil {
+		return true, err
+	}
+
+	// closing multi-part content
+	fmt.Fprintf(wc, "--%s--\r\n", boundary)
 
 	return false, nil
 }
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /.build
 /.release
 /.tarballs
+/.idea
 
 !/doc/examples/simple.yml
 !/circle.yml


config/notifiers.go
```diff
@@ -142,6 +143,7 @@ type EmailConfig struct {
 	AuthIdentity string            `yaml:"auth_identity,omitempty" json:"auth_identity,omitempty"`
 	Headers      map[string]string `yaml:"headers,omitempty" json:"headers,omitempty"`
 	HTML         string            `yaml:"html,omitempty" json:"html,omitempty"`
+	TEXT         string            `yaml:"text,omitempty" json:"text,omitempty"`
```
fabxc: Just `Text` as it's not an acronym.

fabxc:
Thanks for picking this up!

Go's standard library actually has a package for multipart: https://godoc.org/mime/multipart
It seems cleaner to use this rather than constructing everything ourselves. Could you switch to using that?

config/notifiers.go
awaragi: Hi @fabxc 

I am not sure I understand your comment. I am not a Go dev (yet) so I need more details about what you are requesting as a change.

Also regarding the package for multi-part. The original code of email notification was constructed manually from smtp handshake to email construction. I am not at the level to rework this entire section of code. The smtp protocole is old dated and is not likely to change any time soon. may I recommend adding //TODO refactor to use standard multipart library https://godoc.org/mime/multipart?

awaragi:
@fabxc The original code of email notification was constructed manually from smtp handshake to email construction. I am not at the level to rework this entire section of code. The smtp protocole is old dated and is not likely to change any time soon. may I recommend adding //TODO refactor to use standard multipart library https://godoc.org/mime/multipart?

config/notifiers.go
Kellel: I think the key here is that `HTML` is all caps since it it an initialism, but `TEXT` would be written as `Text` since it's just a word.

Does that make sense?
https://github.com/golang/go/wiki/CodeReviewComments#initialisms gives move info on this case

awaragi:
Hi, any progress with this pull request? I see that the flag for changes requested is still there but I have removed the offending file (.gitignore).

mxinden:
@awaragi For sure. Thanks for the reminder.

@fabxc Can you have a second look at this?

fabxc:
> @fabxc The original code of email notification was constructed manually from smtp handshake to email construction. I am not at the level to rework this entire section of code. The smtp protocole is old dated and is not likely to change any time soon. may I recommend adding //TODO refactor to use standard multipart library https://godoc.org/mime/multipart?

Sorry, I've to raise a very clear "NO" on this. Every new feature is a maintenance liability and already having pending TODOs when adding them is not sustainable. 
The question is, who is going to refactor this? The answer is typically no one.

I hope my concern is understandable.

I'm happy to review any attempts to use the proper standard library package. If you don't feel like trying it yourself, maybe someone else who's keen on this feature wants to give it a shot.

awaragi:
@fabxc I understand your point. My liking to Go is increasing so I've attempted to use the mime/multipart library. Take another look and let me know if it works. 

By the way, your original TODO about adding support text template is now removed so one less to go.

mxinden:
As far as I can tell (for documentation purposes): Fixes #290

awaragi:
Hey Guys @fabxc @mxinden , any other suggestions, modifications or can you accept this PR?

awaragi:
Any chance someone can take a look at this PR?

notify/impl.go
```diff
@@ -365,23 +366,47 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 		fmt.Fprintf(wc, "%s: %s\r\n", header, mime.QEncoding.Encode("utf-8", value))
 	}
 
-	fmt.Fprintf(wc, "Content-Type: text/html; charset=UTF-8\r\n")
+	buffer := &bytes.Buffer{}
+	multipartWriter := multipart.NewWriter(buffer)
+
 	fmt.Fprintf(wc, "Date: %s\r\n", time.Now().Format(time.RFC1123Z))
+	fmt.Fprintf(wc, "Content-Type: multipart/alternative;  boundary=\"%s\"\r\n", multipartWriter.Boundary())
```
fabxc: Looks like this can be replaced with this convenience function: https://godoc.org/mime/multipart#Writer.FormDataContentType

notify/impl.go
```diff
@@ -365,23 +366,47 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
...
+
 
 	// TODO: Add some useful headers here, such as URL of the alertmanager
 	// and active/resolved.
 	fmt.Fprintf(wc, "\r\n")
 
-	// TODO(fabxc): do a multipart write that considers the plain template.
+	// Html template
+	multipartWriter.CreatePart(textproto.MIMEHeader{"Content-Type": {"text/html; charset=UTF-8"}})
```
fabxc: According to the documentation of `CreatePart` the `io.Writer` it returns should be used to write the output of the template into.
Same below when creating the part for the text template.

https://godoc.org/mime/multipart#Writer.CreatePart

The returned error should be checked as well.

awaragi:
Thanks for the feedback. I will update accordingly.

notify/impl.go
awaragi: Hi Fabian @fabxc, your initial thought is correct however for multipart email content, you need the multipart/alternative instead of multipart/form-data which is what the FormDataContentType return. I did however notice an additional quotes that were problematic that I took out from around the boundary string.

notify/impl.go
awaragi: @fabxc implemented as suggested and added basic error handling.

notify/impl.go
```diff
@@ -365,23 +366,53 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
...
+	body, err = n.tmpl.ExecuteTextString(n.conf.Text, data)
+	if err != nil {
+		return false, fmt.Errorf("executing email text template: %s", err)
+	}
+	_, err = w.Write([]byte(body))
 	if err != nil {
 		return true, err
 	}
 
+	// closing multi-part content
```
stuartnelson3: This comment probably isn't necessary, since the code says essentially the same thing.

notify/impl.go
```diff
@@ -365,23 +366,53 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
...
+	multipartWriter.Close()
+
+	// writing multipart content
```
stuartnelson3: Same as above comment.

stuartnelson3:
@fabxc any chance to take another look?

awaragi:
@fabxc any possibility to move forward with this pull request?

awaragi:
Anyone?

awaragi:
Any possibility of merging this pull request?

stuartnelson3:
Sorry for the delay, I'll try to poke @fabxc again

fabxc:
Thanks, LGTM! Sorry about the long delay.

alin-amana:
This causes a regression.

I've built alertmanager from head and now I'm getting multipart emails with the old html content and empty text content. If your email client chooses the text format (which Gmail does for me, for some reason) then the content of the alert email is empty.

The code should check for non-empty html and/or text templates and skip the empty one, if any.

alin-amana:
Also, a `MIME-Version: 1.0` header must be added: http://www.rfc-base.org/txt/rfc-2045.txt

alin-amana:
And, per [RFC 2046](http://www.rfc-base.org/txt/rfc-2046.txt) section 5.1.4, the parts should be in increasing preference order. Meaning text should come before HTML. This explains why I'm seeing the empty plain text version in Gmail.

Never mind, I'll submit a PR.

stuartnelson3:
@alin-amana Are you working on this currently? This is a big enough issue that a new release will go out when it's fixed.

alin-amana:
Yup, working as we speak.

alin-amana:
Done, PTAL at #1009.

RichiH:
Huh, missed that one. Thanks to all involved!

Issue #884, opened 2017-06-24
Display URLs as links in labels and annotations


grobie:
It's common to include links in annotations and labels, e.g. to runbooks or dashboards. Detect them and render them as real link.

@w0rm

beorn7:
That's also one of the items on my wish list.

mxinden:
@grobie @beorn7 links in annotations are now clickable thanks to @w0rm (#946).

I do understand @w0rm concern of having links in a label. From your point of view, are rendered links in the Alert annotations good enough, or are links in labels a very common use case which we should support?
> I'm not sure that the links in labels should be clickable, I think it would be too much and might affect the rendering performance of the alert list.

grobie:
I think starting with links in annotations should be enough. I just checked and couldn't find any links in in labels at SoundCloud. If someone has a need for that, they should add a new issue and we can discuss alternative solutions or whether it's worth to add such feature.

Thanks a lot for the feature @w0rm!

PR #946, opened 2017-08-13
Linkify annotations

w0rm:
This pr partially addresses #884 by making the links in annotations clickable. I'm not sure that the links in labels should be clickable, I think it would be too much and might affect the rendering performance of the alert list.

The implementation is rather simple: I'm splitting the text by words, then checking if a word starts with `http(s)://`. This means each link should be surrounded by a whitespace.

screen shot 2017-08-13 at 01 56 14


ui/app/src/Utils/Views.elm
```diff
@@ -33,6 +33,20 @@ labelButton maybeMsg labelText =
                 [ span [ class "text-muted" ] [ text labelText ] ]
 
 
+linkifyText : String -> List (Html msg)
+linkifyText str =
+    List.map
+        (\result ->
+            case result of
+                Ok link ->
+                    a [ href link ] [ text link ]
```
mxinden: Could you add `target "blank"` here?

PR #959, opened 2017-08-22
Add xMatters notifier

xMTinkerer:
This is a PR for adding an xMatters notifier.


xMTinkerer:
Doh, need to write some tests. I'll be back later.

stuartnelson3:
At this point, we're recommending implementing your notifier using the webhook + external service, and if we find enough interest + someone willing to maintain the integration, promoting it to being natively implemented.

xMTinkerer:
I'm willing to maintain it. If we get enough customers using it, then we'll build it in as a native integration on our side.

PR #963, opened 2017-08-28
Add dockerfile for ppc64le

hkshaw1990:
Adds a dockerfile for ppc64le and a check for architecture in the Makefile for target `docker`.

### DIFF
--- a/Dockerfile.ppc64le
+++ b/Dockerfile.ppc64le
@@ -1,0 +1,11 @@
+FROM ppc64le/busybox:latest
+
+COPY alertmanager               /bin/alertmanager
+COPY doc/examples/simple.yml    /etc/alertmanager/config.yml
+
+EXPOSE     9093
+VOLUME     [ "/alertmanager" ]
+WORKDIR    /alertmanager
+ENTRYPOINT [ "/bin/alertmanager" ]
+CMD        [ "-config.file=/etc/alertmanager/config.yml", \
+             "-storage.path=/alertmanager" ]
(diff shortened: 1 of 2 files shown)


stuartnelson3:
Thanks for the contribution. What reason do you have for wanting to run a docker container based on ppc64le?

hkshaw1990:
> Thanks for the contribution. What reason do you have for wanting to run a docker container based on ppc64le?

@stuartnelson3 If I understand the question correctly, ppc64le is a supported architecture for docker just like amd64 and arm. And with docker being the preferred way to deploy cloud-native apps across arch, we want to run docker container on ppc64le so that enterprises/users can leverage the benefits of power platform with docker.

stuartnelson3:
I personally would prefer that interested users maintain their own separate dockerfile if they want to use a different arch to run the container, as we already provide and support one.

hkshaw1990:
>I personally would prefer that interested users maintain their own separate dockerfile if they want to use a different arch to run the container, as we already provide and support one.

@stuartnelson3 What do you suggest the best way forward to ensure `alertmanager` is available for multiple architectures?  
We have followed very similar approaches in case of other projects like [node-exporter](https://github.com/prometheus/node_exporter/blob/master/Dockerfile.ppc64le) where
architecture specific dockerfiles are part of the main project.

While users can maintain their own separate dockerfiles, for the project to support different architectures I feel that dockerfiles and any associated helper scripts (if required) should be part of the main project, IMHO.  This will ensure that interested users can just pull this repo and build the docker images irrespective of whatever architecture they are running on.

stuartnelson3:
Each release is currently built for several architectures. If a user decides that they want to run a specific arch in a docker container, they can download and unpack the .tar.gz via their dockerfile.

Issue #812, opened 2017-05-22
Alert sorting in 0.7 is not consistent.


raypettersen:
When testing the RC, issues are rendered in different order on each refresh. Can be reproduced simply by reloading the page, or toggling the silence switch.

Not a big issue (more an annoyance), but a bit confusing for the end-user. Suggest sorting by start-date, perhaps with a per-alertname toggle so the end user can sort by newest or oldest.

w0rm:
Agree! I wonder if the sorting (most recent first) can be done in the api.

stuartnelson3:
This will be addressed in #872

mxinden:
This issue still persists.

stuartnelson3:
closed by #969

PR #969, opened 2017-09-02
Sort alerts by their fingerprint

stuartnelson3:
addresses #812 

@fabxc @brancz @mxinden opinion on using `Fingerprint`?

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 	"net/http"
 	"regexp"
+	"sort"
 	"sync"
 	"time"
 
@@ -359,6 +360,9 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 		}, nil)
 		return
 	}
+	sort.Slice(res, func(i, j int) bool {
+		return res[i].Fingerprint < res[j].Fingerprint
+	})
 	respond(w, res)
 }
 


brancz:
Not sure how to message this or whether that is necessary at all, but we need to make sure that this is just a default sorting order, that happens to be consistent at the moment, and that this behavior may change at any moment. Those things are easy for things like config files, that have documentation, but for front-end it's harder, but we also don't want to confuse people who start using it, however, given the current behavior, this is already much better, so probably nothing to worry about.

mxinden:
@stuartnelson3 @brancz Why not use `StartsAt`? Both `StartsAt` and `Fingerprint` would serve a consistent ordering, but `StartsAt` would additionally give a human useful sorting?

brancz:
I can't find right now in which issue I commented this, but I think sorting by fingerprint is the default that the backend should always apply, and then additional sorting should be configured by the front-end explicitly, but sorting by `StartAt` could be a useful default selection in the front-end. Just my two cents, don't really have a very strong opinion.

brancz:
code lgtm, but there is something wrong with `bindata.go` on Travis CI. Also we'll need to keep in mind what to do if we want to have pagination as we probably don't want to sort all records, but we can think of that when the time comes, just want to make sure this is not set in stone, if we encounter problems with it, I think we can be free to change this behavior.

mxinden:
The failure on Travis was just a flake. I have re-triggered it, all green now.

mxinden:
Thanks @stuartnelson3!

PR #970, opened 2017-09-02
Stn/edit on silence view page

### LINKED ISSUE #794: Edit / Expire silence on SilenceView page
When viewing the details of a silence, a user should be able to jump to the silence edit page, or expire the silence without going back to the *SilenceList* page.

stuartnelson3:
fixes #794 

@w0rm @mxinden  I thought about moving the buttons into their own `Views/Shared/SilenceButtons.elm` but decided to wait to see what you thought. The only thing I'm not really happy about is having the `refresh` bool to indicate "navigate to all silences if on a silence view page, but don't if on the main silences page".


ui/app/src/Views/SilenceView/Updates.elm
```diff
@@ -8,11 +8,11 @@ import Utils.Types exposing (ApiData(..))
 import Utils.Filter exposing (nullFilter)
 
 
-update : SilenceViewMsg -> Model -> String -> ( Model, Cmd SilenceViewMsg )
-update msg model basePath =
+update : SilenceViewMsg -> Model -> String -> String -> ( Model, Cmd SilenceViewMsg )
+update msg model basePath apiUrl =
```
w0rm: Looks like `basePath` argument isn't used.

stuartnelson3:
@w0rm can you take a look at the style on the silence view page? the buttons don't really line up with the header, and I'm not sure what to do about it.

w0rm:
@stuartnelson3 sure, I will check this tomorrow!

w0rm:
@stuartnelson3  this would make it look nicer

```elm
h1 []
    [ text "Silence"
    , span
        [ class "ml-3" ]
        [ editButton silence
        , deleteButton silence True
        ]
    ]
```

stuartnelson3:
looks great, updated

mxinden:
@stuartnelson3 Cool, thanks for the fix!

I am fine with the `refresh` boolean. What we could do as well is just always redirect to the `/#silences` page on delete just for simpler code, as performance is probably not that important. No strong opinion, I am totally fine with this.

stuartnelson3:
That's what I initially did, but if you have an active filter, redirecting will erase that. If we preserve the filter, we still have to deal with having an unstable silence order currently :/ happy to change this once we have stable silence order.

Issue #947, opened 2017-08-14
Switch to Kingpin?


carlpett:
Hi,
Prometheus is switching to Kingpin [as part of 2.0](https://github.com/prometheus/prometheus/issues/2455). I've migrated all the Go exporters to do the same, and could do AM as well. However, the amtool/cli parts of Alertmanager uses Viper/Cobra (while the main Alertmanager binary is still using `flag`).

I'd just like to check beforehand that switching to Kingpin is also the desired path for Alertmanager?

@stuartnelson3 @fabxc

carlpett:
Also, while I was looking over the source, I found that `-mesh.peers` is a bit peculiar: it both allows repetitions as well as doing comma-splitting of arguments. The latter behaviour doesn't seem to be documented, is this a remnant of a previous approach or a deliberate shorthand?

fabxc:
Yes, we want to switch everything to it for consistency.

The `-mesh.peers` flag is the way it is because I thought it's convenient to allow either and didn't see much harm in the ambiguity. But if there are arguments to just pick one of them, that's fine with me as well.

stuartnelson3:
implemented

PR #974, opened 2017-09-04
Switch alertmanager to kingpin

carlpett:
This PR contains the first (easy) part of the migration to Kingpin, as discussed in #947, namely converting the alertmanager binary.

I have converted amtool as well, but since that becomes quite a bit messier, I thought we'd start here and take that in a separate PR.


stuartnelson3:
Thanks for the pr! I'll try to give this a look soon, and I hope some of the other folks will too :)

carlpett:
@stuartnelson3 Apparently this PR, related to #976, also wasn't merged :)
Rebased onto master now.

PR #975, opened 2017-09-04
Remove superfluous trailing whitespace

johto:


johto:
Oh.  Actually, it looks like the space is intentional as it's meant to be a separator between the key=value strings.  Sorry for the noise.

PR #976, opened 2017-09-04
Switch amtool to kingpin

carlpett:
Related to #974, this PR switches amtool to use kingpin.

This PR is a bit more involved, since cobra/viper does not translate quite as easily as the core flag package. There are some remaining points before this can be merged that I would like feedback on:
1. There were quite nice long format help texts written previously. There is no "long help" field in kingpin, but maybe it would be a good idea to make the man page template (`--help-man`) use it?
1. Is the config file an important feature? Kingpin does not have any file loading capabilities, so we'll need to built it ourselves in that case, and ensure it falls back in the correct order. If so, maybe the number of file formats could be reduced to avoid all too much complexity?
1. What is the purpose of `date.format`? It is only exposed through the config file from what I understand of the code, and it is not documented. 
1. `comment-required` is not implemented since it is only supported through config-file at the moment.

Let me know if there are any other changes you think are necessary! @stuartnelson3 @fabx


stuartnelson3:
1. :+1: 
2. It would be good to maintain the config file functionality, since there is a lot of additional metadata that regularly needs to be added since the cli tool does affect alertmanager. I would say stick with .yml and support the two location that are currently supported.
3. Not sure .. @Kellel ? I've never used it, and I think we're limited by the date format that alertmanager parses.
4. I think it would be best to require a comment, since we require one through the UI.

Thanks for doing this!

Kellel:
The date.format option is only used for the output formatting.

It's not super important. I mean, I even forgot about it and I am the one who wrote it.

Kellel:
I am happy with the comment being required, but if so we should enforce this at the api level as well (Out of scope for this PR).

Kellel:
Also as far as the config file goes, viper (the config file library) can be used without cobra (the flag parsing library). So you can feasibly still use viper for all the config file processing and kingpin for the flags. I don't know if that will be easier, but it might be worth exploring.

carlpett:
Thanks for the input!

@stuartnelson3 Sounds good. I'll implement some config file loading.

@Kellel Regarding using viper, I looked into that a bit, but from what I can see in the [related issue](https://github.com/alecthomas/kingpin/issues/184), it doesn't look trivial at least.

carlpett:
So, I found this PR lying around on my laptop, thought I had finished it long ago... Sorry about that. I finished up the small work that remained (which mainly consisted of getting a PR into kingpin, done since many months now), and now I'm working through the many merge conflicts... Looks like it could take a while longer than I have available today, but I hope to finish it tomorrow.

stuartnelson3:
cool, thanks for the update!

carlpett:
So, merging old branches is a pain (no news there). I'm testing it a bit manually now to make sure parsing etc does what it should, but I'm also seeing the acceptance tests fail. I don't understand why, though, as far as I see, amtool is not used for this?

carlpett:
Ok, acceptance tests are suddenly green again after some last fixes of other things. Are they known to be flakey, or could I have broken something?

Anyway, I think I've addressed all the points I had in the original post:
1. Long-form help-texts are given with `--help-long`.
2. The config file support is implemented (although only in the default locations, and only with yaml format)
3. date.format has been given a flag, as well as works from config file
4. require-comment is supported in the config file, and has a (hidden) flag as well

Let me know what you think!

stuartnelson3:
Taking a look, thanks!

And yeah, the acceptance tests are flakey :/

cli/root.go
```diff
@@ -1,83 +1,177 @@
...
+
+	app.Version(version.Print("amtool"))
+	app.GetFlag("help").Short('h')
+	app.UsageTemplate(kingpin.CompactUsageTemplate)
+	app.Flag("long-help", "Give more detailed help output").UsageAction(&kingpin.UsageContext{
+		Template: longHelpTemplate,
+		Vars:     map[string]interface{}{"LongHelp": longHelpText},
+	}).Bool()
+
+	configResolver := newConfigResolver()
```
stuartnelson3: I would prefer to have `newConfigResolver` return an error, and instead of panicking on a file read error or a yaml unmarshal error, provide a shorter and clearer message. `panic()`'s stack trace probably won't aid a user much, and generate a github issue when they just need to fix their yaml.

cli/silence_add.go
```diff
@@ -7,14 +7,10 @@ import (
 	"fmt"
 	"net/http"
 	"os/user"
-	"path"
 	"time"
 
+	"github.com/alecthomas/kingpin"
```
stuartnelson3: import

cli/silence_expire.go
```diff
@@ -6,29 +6,27 @@ import (
...
+	if len(*expireIds) < 1 {
 		return errors.New("no silence IDs specified")
 	}
 
-	for _, arg := range args {
-		u.Path = path.Join(basePath, arg)
+	basePath := "/api/v1/silence"
+	for _, id := range *expireIds {
+		u := GetAlertmanagerURL(path.Join(basePath, id))
 		req, err := http.NewRequest("DELETE", u.String(), nil)
```
stuartnelson3: again, not your fault, but could you do the `if err != nil` dance here?

cli/alert.go
```diff
@@ -6,17 +6,15 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
-	"path"
 	"strings"
 	"time"
 
+	"github.com/alecthomas/kingpin"
```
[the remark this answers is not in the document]
carlpett: Actually, no :) We need to use the v3-branch, since the pull request I made to kingpin to add support for config files is only present on that branch.

cli/root.go
carlpett: Agreed.

cli/silence_expire.go
carlpett: :+1:

carlpett:
@stuartnelson3 Pushed fixes for all the comments except the import path, because of the reason I replied with there. I think there might be a bit of time until v3 is released, given the number of remaining items on the roadmap, but I haven't had any stability issues in our usage of "v3-unstable" in our work projects.

stuartnelson3:
Looks great, works great. I have one final, tiny request (or question, I guess), and then this is merged. You mentioned that extended help is available via `--help-long`, and that's the same flag that's mentioned in the kingpin documentation. It's implemented here as `--long-help`. I'm fine with either since it shows up as a flag when a user does `-h`, but just wanted to see if this was done on purpose or not.

carlpett:
Oh, uh, that is just a silly mistake on my part... I don't know how I miswrote that. Pushing in a sec!

stuartnelson3:
No worries :)

PR #978, opened 2017-09-05
amtool check-config

### LINKED ISSUE #333: Validate config
Is there a way to check if the config is valid? (similar to promtool with Prometheus)

Thanks.

iksaif:
This is similar to `promtool check-config` and allows one
to validate the alertmanager configuration (as a git presubmit for example).

`govendor fetch github.com/spf13/{cobra,pflag}` was needed to
have support for `Args`.

This fixes #333 

```
$ ./amtool check-config alertmanager.yml ./doc/examples/simple.yml 
Checking 'alertmanager.yml'  SUCCESS
Found 1 templates:   FAILED: template: foo.tmpl:3: unexpected {{end}}

Checking './doc/examples/simple.yml'  SUCCESS
Found 1 templates:   SUCCESS

Error: Failed to validate 1 file(s).
```


mxinden:
@iksaif Thanks a lot for the contribution. This was requested very often.

I am not familiar with the amtool testing story (//CC @stuartnelson3), but I think we should have a basic test. @iksaif would you be able to add one for the configuration check?

stuartnelson3:
Ah I forgot to comment! I tested this manually this morning, worked great.

A really high level smoke test should be enough if it's not too much trouble -- does it recognize a good config.yml and an bad one.

iksaif:
Added tests

cli/check_config_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+func TestCheckConfig(t *testing.T) {
+	err := CheckConfig([]string{"testdata/conf.good.yml"})
+	if err != nil {
+		t.Fatalf("error: %v", err)
+	}
+
+	err = CheckConfig([]string{"testdata/conf.bad.yml"})
+	if err == nil {
+		t.Fatalf("Failed to detect missing file")
```
mxinden: I am confused by the word `missing` here. How about `failed to detect invalid file`?

cli/check_config_test.go
```diff
@@ -0,0 +1,17 @@
+package cli
+
+import (
+	"testing"
+)
+
+func TestCheckConfig(t *testing.T) {
+	err := CheckConfig([]string{"testdata/conf.good.yml"})
+	if err != nil {
+		t.Fatalf("error: %v", err)
```
mxinden: Instead of `error: ` how about `checking valid config file failed with:  `?

stuartnelson3:
couple small wording changes on the test failure output and then we'll merge

cli/check_config_test.go
iksaif: Forgot to change this when I changed the test

cli/check_config_test.go
iksaif: done

iksaif:
done

PR #984, opened 2017-09-08
Add a mutex to silences.go:gossipData

iksaif:
This should fix silence/silence.go #982


iksaif:
I'm not super happy about naming.
It needs to be carefully reviewed too, mutex can get hard.

stuartnelson3:
Thanks for the quick PR! I'll look at it this weekend

silence/silence.go
```diff
@@ -771,17 +790,28 @@ func (gd gossipData) Encode() [][]byte {
...
-		res[id] = s
+func (gd *gossipData) clone() *gossipData {
+	gd.mtx.RLock()
+	defer gd.mtx.RUnlock()
+
+	data := make(silenceMap, len(gd.data))
+	for id, s := range gd.data {
+		data[id] = s
 	}
+	res := &gossipData{data: data}
```
stuartnelson3: you can just `return &gossipData{data: data}`

silence/silence.go
```diff
@@ -281,12 +281,12 @@ func (s *Silences) GC() (int, error) {
...
 	defer s.mtx.Unlock()
 
-	for id, sil := range s.st {
+	for id, sil := range s.st.data {
 		if sil.ExpiresAt.IsZero() {
 			return n, errors.New("unexpected zero expiration timestamp")
 		}
 		if !sil.ExpiresAt.After(now) {
-			delete(s.st, id)
+			delete(s.st.data, id)
```
stuartnelson3: Reading this now, this is probably what caused the concurrent read/write. If you look at the trace you provided us, it starts with:

```
time="2017-09-08T12:38:05Z" level=info msg="running maintenance" component=silences source="silence.go:232" 
time="2017-09-08T12:38:05Z" level=info msg="running maintenance" component=nflog source="nflog.go:286" 
time="2017-09-08T12:38:05Z" level=info msg="maintenance done" component=nflog duration=137.709µs source="nflog.go:303" 
time="2017-09-08T12:38:05Z" level=info msg="maintenance done" component=silences duration=84.744µs source="silence.go:249" 
fatal error: concurrent map iteration and map write
```

Access to `s.st.data` here also needs to be locked.

I'm a bit curious on the internals of how go works, because access to silences was definitely locked, but it seems the internal map was still concurrently available. Maybe it's something to do with the map being a type alias.

@fabxc @beorn7 any ideas or light you can shed on this?

iksaif: I don't think this was caused by the maintainance, I think it was caused by two `Encode` functions simulatneously encoding the same data:
```
github.com/prometheus/alertmanager/silence.gossipData.Encode(0xc424e69c20, 0xc82580, 0xc424e69c20, 0xc4213f37c0)
github.com/prometheus/alertmanager/silence.gossipData.Encode(0xc424e69c20, 0xc82580, 0xc424e69c20, 0xc4233ab960)
```

Here s.st.data is protected by s.mtx (it's never being accessed without this mutex being held)

silence/silence.go
```diff
@@ -724,10 +726,23 @@ func (g gossiper) OnGossipUnicast(src mesh.PeerName, msg []byte) error {
...
 
-type gossipData map[string]*pb.MeshSilence
+type silenceMap map[string]*pb.MeshSilence
+
+type gossipData struct {
+	data silenceMap
+	mtx  sync.RWMutex
+}
+
+var _ mesh.GossipData = &gossipData{}
```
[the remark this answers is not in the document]
iksaif: It's something that I got from the example. It's purpose is to validate early that the interface is correctly implemented.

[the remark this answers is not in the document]
stuartnelson3: This can probably be deleted? Compilation would fail if something not implementing an interface were being passed to a function.

iksaif:
This got deployed on a setup with ~2000 silences and 7 alertmanagers. Seems stable.

PR #985, opened 2017-09-08
Adds PagerTree native integration

armiiller:
Adds PagerTree native integration

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.8.1 / 2017-09-07
+
+* [FEATURE] Add PagerTree integration notifications.
+
 ## 0.8.0 / 2017-07-20
 
 * [FEATURE] Add ability to filter alerts by receiver in the UI (#890)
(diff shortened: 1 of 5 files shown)


armiiller:
Is there anyway we can make an exception for this? The [CONTRIBUTING.md](https://github.com/prometheus/alertmanager/blob/master/CONTRIBUTING.md) doesn't mention this as the direction of the project, and looking at the [pull requests](https://github.com/prometheus/alertmanager/pulls) there still seems to be a lot of notifiers in consideration. Seeing as how PagerDuty, OpsGenie, and VictorOps are all implemented, it would seem partial to not include this change.

stuartnelson3:
Those integrations were added a long time ago, and our policy regarding adding notifiers has since changed. Sorry that contributing.md hasn't been updated. Unfortunately, we cannot accept and maintain integrations with every possible notifier.

Turning this pull request into a webhook server should be a fairly straightforward task.

PR #990, opened 2017-09-12
allow a binary name to be passed to make build

cstyan:
While working on promtool and prometheus I often went back and forth commenting out binaries in the `.promu.yml` file in order to only rebuild the binary I wanted to rebuild. The alertmanager repo also builds two binaries. 

In this pr: https://github.com/prometheus/promu/pull/77 I added the option to pass a binary name to promu's build function. This is controlled by setting the `BINARIES` var.

For example: 
```
$make build BINARIES=amtool
>> building binaries
 >   amtool
```

In this PR we simply pass `BINARIES` to promu. If it's not set, promu will build all binaries from the YAML file as normal.
```
$make build
>> building binaries
 >   alertmanager
 >   amtool
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ vet:
 # Will only build the back-end
 build: promu
 	@echo ">> building binaries"
-	@$(PROMU) build --prefix $(PREFIX)
+	@$(PROMU) build --prefix $(PREFIX) $(BINARIES)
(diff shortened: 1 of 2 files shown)


fabxc:
Neat, that was bugging me as well for quite a while. 👍

PR #993, opened 2017-09-14
Added confirmation dialog when expiring silences

### LINKED ISSUE #883: Confirmation dialog when deleting or expiring the silence
I guess this should be implemented, in order to make it impossible to accidentally expire the silence

Conorbro:
Closes #883 

Added to silence list view and individual silence view.

See:
test


stuartnelson3:
Thanks for the contribution! I assigned @w0rm to take a look, he should have some time soon.

Regarding the test failure: Once we're ready to merge, the last step is to run `make build-all` from the root dir. This compiles the elm code into javascript, and embeds the javascript into the alertmanager binary.

Conorbro:
@stuartnelson3 Strange it's failing as I ran `make build-all` as per the contribution.md for the Alertmanager ui.

w0rm:
@Conorbro thanks for working on this issue!

I have two concerns:

do we really need to use a library for this or we can have our own dialog component?
showing a confirmation instead of a button or in a tooltip would be a better ux, because it doesn't block the whole ui
Because it's an improvement over what we have now, I think it can be merged, and then we could iterate on top.

Conorbro:
@w0rm 

Yes I agree, I wasn't sure whether introducing another library would be a good idea or not.

It's a fairly simple library so we could probably implement something similar ourselves with a confirmation/tooltip as you have suggested in the future.

PR #994, opened 2017-09-14
Show total number of silenced alerts in preview

### LINKED ISSUE #953: Show total number of silenced alerts
Just realized that when sitting in @mxinden 's talk.

When clicking Preview in Alertmanager, it would be nice to see the total number of alerts hit over the list.

Use case is to see quickly if you're roughly silencing as many alerts as you're expecting to.

Conorbro:
Closes #953 

Show number of affected alerts in create Silence preview:

test2


stuartnelson3:
This looks good to me, can you run `make build-all` from the repo root to regenerate bindata.go, and commit and push that?

w0rm:
@Conorbro thanks for working on this issue! 

I have two concerns:

1. do we really need to use a library for this or we can have our own dialog component?  
2. showing a confirmation instead of a button or in a tooltip would've been a better ux, because it wouldn't block the whole ui

Because it's an improvement over what we have now, I think it can be merged, and then we could iterate on top.

Conorbro:
@w0rm I think your comment was meant for #993 !

w0rm:
@Conorbro sorry for this! :)

brancz:
What are y'all using to create those gorgeous GIFs showing your front-end changes?

w0rm:
@brancz I'm using https://www.cockos.com/licecap/ on OSX

PR #995, opened 2017-09-15
silences: avoid deadlock

iksaif:
Calling gossip.GossipBroadcast() will cause a deadlock if
there is a currently executing OnBroadcast* function.

See #982

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -370,7 +370,9 @@ func (s *Silences) setSilence(sil *pb.Silence) error {
 	}
 
 	s.st.Merge(st)
-	s.gossip.GossipBroadcast(st)
+	// setSilence() is called with s.mtx locked, which can produce
+	// a deadlock if we call GossipBroadcast from here.
+	go s.gossip.GossipBroadcast(st)
 
 	return nil
 }
--- a/silence/silence_test.go
+++ b/silence/silence_test.go
@@ -197,19 +197,24 @@ func TestSilencesSetSilence(t *testing.T) {
 		},
 	}
 
-	var called bool
+	done := make(chan bool)
 	s.gossip = &mockGossip{
 		broadcast: func(d mesh.GossipData) {
 			data, ok := d.(*gossipData)
 			require.True(t, ok, "gossip data of unknown type")
 			require.Equal(t, want, data, "unexpected gossip broadcast data")
-
-			called = true
+			close(done)
 		},
 	}
 	require.NoError(t, s.setSilence(sil))
-	require.True(t, called, "GossipBroadcast was not called")
 	require.Equal(t, want.data, s.st.data, "Unexpected silence state")
+
+	// GossipBroadcast is called in a goroutine.
+	select {
+	case <-done:
+	case <-time.After(1 * time.Second):
+		t.Fatal("GossipBroadcast was not called")
+	}
 }
 
 func TestSilenceSet(t *testing.T) {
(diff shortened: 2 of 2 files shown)


silence/silence_test.go
```diff
@@ -197,19 +197,24 @@ func TestSilencesSetSilence(t *testing.T) {
...
-
-			called = true
+			close(done)
 		},
 	}
 	require.NoError(t, s.setSilence(sil))
-	require.True(t, called, "GossipBroadcast was not called")
 	require.Equal(t, want.data, s.st.data, "Unexpected silence state")
+
+	// GossipBroadcast is called in a gorountine.
```
olive42: typo: goroutine

iksaif: fixed

olive42:
LGTM

silence/silence.go
```diff
@@ -370,7 +370,9 @@ func (s *Silences) setSilence(sil *pb.Silence) error {
 	}
 
 	s.st.Merge(st)
-	s.gossip.GossipBroadcast(st)
+	// setSilence() is called with s.mtx locked, which can produce
+	// a deadlock if we call GossipBroadcast from here.
+	go s.gossip.GossipBroadcast(st)
```
stuartnelson3: The tests still pass when I remove the `go` here. The tests should fail without the inclusion of this.

iksaif: fixed

silence/silence_test.go
```diff
@@ -197,19 +197,34 @@ func TestSilencesSetSilence(t *testing.T) {
...
+			require.Equal(t, want.data, data.data, "unexpected gossip broadcast data")
+			close(done)
 		},
 	}
+
+	// setSilence() is always called with s.mtx locked()
+	s.mtx.Lock()
 	require.NoError(t, s.setSilence(sil))
-	require.True(t, called, "GossipBroadcast was not called")
+	s.mtx.Unlock()
```
stuartnelson3: While this test passes because of the `go`, if someone were to remove the `go`, this deadlocks and the timeout in the `select` below is never called. If you re-arrange the code like below, then we have a passing test with your code :raised_hands: , and if someone removes the `go`, it will fail instead of locking up:

```patch
diff --git a/silence/silence_test.go b/silence/silence_test.go
index 420a78c..eed4060 100644
--- a/silence/silence_test.go
+++ b/silence/silence_test.go
@@ -213,11 +213,11 @@ func TestSilencesSetSilence(t *testing.T) {
 	}
 
 	// setSilence() is always called with s.mtx locked()
-	s.mtx.Lock()
-	require.NoError(t, s.setSilence(sil))
-	s.mtx.Unlock()
-
-	require.Equal(t, want.data, s.st.data, "Unexpected silence state")
+	go func() {
+		s.mtx.Lock()
+		require.NoError(t, s.setSilence(sil))
+		s.mtx.Unlock()
+	}()
 
 	// GossipBroadcast is called in a goroutine.
 	select {
@@ -225,6 +225,8 @@ func TestSilencesSetSilence(t *testing.T) {
 	case <-time.After(1 * time.Second):
 		t.Fatal("GossipBroadcast was not called")
 	}
+
+	require.Equal(t, want.data, s.st.data, "Unexpected silence state")
 }
 
 func TestSilenceSet(t *testing.T) {
```

iksaif: good point !

PR #998, opened 2017-09-15
silence|alerts: add metrics about current silences and alerts

iksaif:
This adds metrics that look like this:
```
alertmanager_alerts_count{state="active"} 6
alertmanager_alerts_count{state="suppressed"} 0
alertmanager_silences_count{state="active"} 1
alertmanager_silences_count{state="expired"} 1
alertmanager_silences_count{state="pending"} 0
```

This can be used to monitor alertmanager's usage and validate that
alertmanagers in a mesh have a similar number of silences and alerts.

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -63,6 +63,8 @@ var (
 		Name: "alertmanager_config_last_reload_success_timestamp_seconds",
 		Help: "Timestamp of the last successful configuration reload.",
 	})
+	alertsActive     prometheus.GaugeFunc
+	alertsSuppressed prometheus.GaugeFunc
 )
 
 func init() {
@@ -70,6 +72,27 @@ func init() {
 	prometheus.MustRegister(configSuccessTime)
 	prometheus.MustRegister(configHash)
 	prometheus.MustRegister(version.NewCollector("alertmanager"))
+}
+
+func newAlertMetricByState(marker types.Marker, st types.AlertState) prometheus.GaugeFunc {
+	return prometheus.NewGaugeFunc(
+		prometheus.GaugeOpts{
+			Name:        "alertmanager_alerts",
+			Help:        "How many alerts by state.",
+			ConstLabels: prometheus.Labels{"state": string(st)},
+		},
+		func() float64 {
+			return float64(marker.Count(st))
+		},
+	)
+}
+
+func newMarkerMetrics(marker types.Marker) {
+	alertsActive = newAlertMetricByState(marker, types.AlertStateActive)
+	alertsSuppressed = newAlertMetricByState(marker, types.AlertStateSuppressed)
+
+	prometheus.MustRegister(alertsActive)
+	prometheus.MustRegister(alertsSuppressed)
 }
 
 func main() {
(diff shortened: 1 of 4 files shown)


cmd/alertmanager/main.go
```diff
@@ -72,6 +74,27 @@ func init() {
 	prometheus.MustRegister(version.NewCollector("alertmanager"))
 }
 
+func newAlertMetricByState(marker types.Marker, st types.AlertState) prometheus.GaugeFunc {
+	return prometheus.NewGaugeFunc(
+		prometheus.GaugeOpts{
+			Name:        "alertmanager_alerts_count",
```
brian-brazil: The _count suffix is for summaries/histograms time series

iksaif: Removed.

iksaif:
count removed, tests fixed

types/types.go
```diff
@@ -67,6 +69,27 @@ type memMarker struct {
 	mtx sync.RWMutex
 }
 
+// Count alerts of a given state.
+func (m *memMarker) Count(states ...AlertState) int {
+	count := 0
+
+	m.mtx.Lock()
```
brancz: This just needs a read lock.

silence/silence.go
```diff
@@ -113,9 +113,26 @@ type metrics struct {
...
+
+func newSilenceMetricByState(s *Silences, st SilenceState) prometheus.GaugeFunc {
+	return prometheus.NewGaugeFunc(
+		prometheus.GaugeOpts{
+			Name:        "alertmanager_silences",
+			Help:        "How many silences by state.",
+			ConstLabels: prometheus.Labels{"state": string(st)},
+		},
+		func() float64 {
+			count, _ := s.CountState(st)
```
brancz: We should probably at least log this error.

silence/silence.go
```diff
@@ -584,6 +610,16 @@ func (s *Silences) Query(params ...QueryParam) ([]*pb.Silence, error) {
 	return sils, err
 }
 
+// Count silences by state.
+func (s *Silences) CountState(states ...SilenceState) (int, error) {
+	// This could probably be optimized.
```
brancz: I'm hoping that we'll one day we'll be able to index the silence and alert in-memory databases, so we can search, count, query a bit more efficiently.

silence/silence.go
iksaif: done

types/types.go
iksaif: done

iksaif:
Should be good now

types/types.go
```diff
@@ -67,6 +69,27 @@ type memMarker struct {
 	mtx sync.RWMutex
 }
 
+// Count alerts of a given state.
+func (m *memMarker) Count(states ...AlertState) int {
+	count := 0
+
+	m.mtx.RLock()
+	defer m.mtx.Unlock()
```
brancz: This should be `RUnlock`, when using `RLock`.

iksaif: True, sorry for that, fixed

iksaif:
>  I'm unsure about the performance implications the locking for counting may have

Should not be worse than accessing the API or the UI.

brancz:
Fair enough. Thanks for the contributions!

Issue #996, opened 2017-09-15
Expired silences are not garbage collected after -data.retention has passed.


zarkoc:
Im using the api to create a silence, and afterwards delete it.
The deleted silence goes to expired.
So potentially I could have hundreds of expired alerts.
Is there a way to delete expired silences, or some other way to handle this case.
I'm on version 0.7.1.
Thanks,
Zarko

fabxc:
Expired silences will be automatically garbage collected after ~14 days.
The reason is that it's good to still see them afterwards. For example to
analyse why certain alerts were not sent or to re-create a silence.

zarkoc:
Thanks! Didnt know that.

stuartnelson3:
@beorn7 you reported that soundcloud has several thousand expired silences -- can you check to see if they're being GC'd? The retention is set by `-data.retention`, with a default value of 5 days.

@fabxc Looking at the code it looks like 5 days is the default time, did I miss something that sets it to 14 days?

fabxc:
No, I was honestly just guessing – has been a while since writing that code.

stuartnelson3:
Not a problem, just wanted to make sure I wasn't crazy

beorn7:
The flag `-data.retention` has indeed a default value of 120h (== 5d). We're running AM 0.8.0 with the default flage, and we have accumulated 4341 expired silences by now, most of them definitely expired longer ago than 5d.

josedonizetti:
Hey, is it okay if I take stab at this one?

beorn7:
I think that would be great, @josedonizetti . @stuartnelson3 might be able to give more context.

josedonizetti:
I run a few tests on this, and:

1) Standalone test

All tests I did here worked perfectly, the silences will always be GC. But one point caught my attention if for some reason one of the expires has a zero value time it will break the GC loop and stop the Maintenance go routine that runs it and log the error. 

https://github.com/prometheus/alertmanager/blob/master/silence/silence.go#L286
https://github.com/prometheus/alertmanager/blob/master/silence/silence.go#L235

I was thinking that instead of stopping the Maintenance routine for good in case an error value appears, we should log the error, but keep GCing stuff. Basically changing the L286 to log the error, and continue. What do you think?

2) HA test

[...]

PR #999, opened 2017-09-19
Ignore expired silences OnGossip

josedonizetti:
This will fix the bug of resync deleted silences due to the state of other peers. More context on https://github.com/prometheus/alertmanager/issues/996

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -291,6 +291,7 @@ func (s *Silences) GC() (int, error) {
 			n++
 		}
 	}
+
 	return n, nil
 }
 
@@ -831,8 +832,8 @@ func (gd *gossipData) Merge(other mesh.GossipData) mesh.GossipData {
 	return gd
 }
 
-// mergeDelta behaves like Merge but returns a gossipData only
-// containing things that have changed.
+// mergeDelta behaves like Merge but ignores expired silences, and
+// returns a gossipData only containing things that have changed.
 func (gd *gossipData) mergeDelta(od *gossipData) *gossipData {
 	delta := newGossipData()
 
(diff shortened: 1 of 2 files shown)


stuartnelson3:
Thanks for the PR! I will have to think about this, as it opens up new issues: some people like to configure their data retention time high so that they can view and recreate silences from expired ones. By not gossiping expired silences, there will be inconsistency between the alertmanagers in a mesh.

josedonizetti:
@stuartnelson3 

This isn't about ignoring gossiped expired silences, those will still exist in all HA, no inconsistency on this part. The PR ignores only GCed expired silences. I've started to test this after seen this issue https://github.com/prometheus/alertmanager/issues/996.

The test was done on current master. 

What I did:
1) decrease the data retention to a smaller window (5 min)
2) decrease the GC ticker to a smaller window (3 min)
3) start 3 instances of HA
4) send a few alerts
5) silence those alerts
6) expire those silences
7) wait to see if the GC would clean them

What I saw:

The GC would run on a specific instance 'A' and would clean the expired silences. But just a few seconds back the expired alerts would reappear in 'A' due to a gossiped of another instance 'B' or 'C', where the GC hadn't run.

Let me know if there more information I can gather to help out on this issue.

beorn7:
@stuartnelson3 are you still up to reviewing this?

The not-happening GC is hurting us quite a lot by now, so I'm tempted to work on this myself, but I don't want to re-do work already done or work redundantly on something that is already being worked on behind the scenes.

stuartnelson3:
Feel free to grab it, I was forced to have a weekend to myself and haven't touched it :)

silence/silence.go
```diff
@@ -831,8 +832,8 @@ func (gd *gossipData) Merge(other mesh.GossipData) mesh.GossipData {
 	return gd
 }
 
-// mergeDelta behaves like Merge but returns a gossipData only
-// containing things that have changed.
+// mergeDelta behaves like Merge but ignore expired silences, and
```
stuartnelson3: `/s/ignore/ignores/`

silence/silence.go
```diff
@@ -843,6 +844,11 @@ func (gd *gossipData) mergeDelta(od *gossipData) *gossipData {
 	defer gd.mtx.Unlock()
 
 	for id, s := range od.data {
+		// Only merge silence if not expired
```
stuartnelson3: Could you expand on this comment a bit? I know I'm going to come back to it in the future and need to refresh myself. Something along the lines of:

>If a gossiped silence is expired, skip it. For a given silence duration exceeding a few minutes, active silences will have already been gossiped. Once the active silence is gossiped, its expiration should happen more or less simultaneously on the different alertmanager nodes. Preventing the gossiping of expired silences allows them to be GC'd, and doesn't affect consistency across the mesh.

PR #1001, opened 2017-09-23
[bugfix] fix default email template rendering

xginn8:
emails sent from builds from HEAD are missing the HTML template,
which is a result of the missing Close() call on the HTML Part.
@stuartnelson3


stuartnelson3:
Closed in favor of #1009

Issue #701, opened 2017-04-10
Arbitrary notifier integrations with Jsonnet


fabxc:
Similar to Prometheus SD integrations, Alertmanager struggles with an increasing demand for more notifier integrations. We can only maintain so many of them ourselves, contributed integrations are often not great in quality and/or have no long-term maintainers.

Many requested integrations are not designed for reliable alert delivery, yet there's an understandable demand by people with low-risk environments (personal websites etc.). As Alertmanager really is only about making single API calls, which are generally REST APIs using JSON, it's not too crazy think about templating such calls.

[Jsonnet](http://jsonnet.org/) is a JSON data templating language by Google. In the past there was no Go implementation but now there's one [open sourced](https://github.com/google/go-jsonnet).

From my understanding it has reasonably matured semantics (certainly better than us providing something on top of Go templating).
I'd like to consider it to allow people templating of JSON payloads with it. Combined with templating of request URLs, this should be sufficient to let people integrate with almost anything today.

I realize its throwing another templating language into the game and even nesting them if some JSON fields should be templated additionally.
But generally those would be write-once per integration and are easy to share among users without us having any responsibility burden. Ultimately, we could probably even implement a lot of the existing integrations with it.

I'd like to investigate this. Would definitely need a proposal doc though.

@stuartnelson3  @leominov  @brian-brazil

brian-brazil:
One thing to consider is that errors are very inconsistently signalled. There's a good mix of 4xx, 5xx and message content - sometimes within the same integration.

xMTinkerer:
It seems that accepting pull requests for new notifiers is no longer happening as the burden of maintaining them rests finally on the maintainers of alertmanager. In addition to the Telegram and Jira notifiers, AmazonSNS was not accepted (#709). 
Having some kind of plugin system for these will help alleviate the burden off the alertmanager maintainers and back onto the plugin maintainer where it belongs. There will be an ever growing list of new notifiers or snowflake things users want so I would encourage moving that burden to those who can and want to have it.

stuartnelson3:
https://github.com/hashicorp/go-plugin was recently brought to my attention, which seems like a well-used solution

brian-brazil:
How would that differ from the existing webhook? It looks like a slightly different way of doing the same thing.

stuartnelson3:
[...]

PR #1003, opened 2017-09-25
Add etcd notifier

dtnaylor:
Add notifier that puts/deletes keys in an etcd KV store using etcd's grpc-gateway.

I’m not sure how much demand for this there is, so if you don’t think it’s worth supporting, then consider this a +1 for #701.

A sample configuration:

```
# sample etcd notifier config that writes a KV pair when an alert fires
# and deletes it when the alert is resolved
etcd_configs:
 - firing:
     url: http://localhost:2379/v3alpha/kv/put
     key_annotation: 'etcd_key'
     value_annotation: 'etcd_value'
   resolved:
     url: http://localhost:2379/v3alpha/kv/deleterange
     key_annotation: 'etcd_key'
     value_annotation: 'etcd_value'
```

And a sample rule:

```
ALERT QueueFull
  IF queue_occupancy > 0.9
  FOR 5s
  LABELS { severity = “warning” }
  ANNOTATIONS {
    summary = “…”,
    etcd_key = "/queue_occupancy/{{ $labels.queue_name }}",
    etcd_value = "{{ $value }}",
  }
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -458,6 +458,7 @@ type Receiver struct {
 	HipchatConfigs   []*HipchatConfig   `yaml:"hipchat_configs,omitempty" json:"hipchat_configs,omitempty"`
 	SlackConfigs     []*SlackConfig     `yaml:"slack_configs,omitempty" json:"slack_configs,omitempty"`
 	WebhookConfigs   []*WebhookConfig   `yaml:"webhook_configs,omitempty" json:"webhook_configs,omitempty"`
+	EtcdConfigs      []*EtcdConfig      `yaml:"etcd_configs,omitempty" json:"etcd_configs,omitempty"`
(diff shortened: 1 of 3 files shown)


stuartnelson3:
Thanks for the pr! As you noticed, we're not actively adding notifiers at the moment as we try to figure out something more flexible. For the time being, we recommend implementing custom notifiers via the webhook notifier.

dtnaylor:
Great, we'll stay tuned to see what you come up with :)

In case it helps your planning, the two things we need in this case that the webhook notifier doesn't (currently) offer are (1) the ability to add custom keys to the POSTed JSON and (2) the ability to base64 encode the values (maybe base64encode could be added as a template function for Prometheus alerts?).

PR #1008, opened 2017-09-29
Inconsistent use of baseUrl and apiUrl arguments

### LINKED ISSUE #1007: [Bug] Affected alerts are not returned on the page of a silence in 0.9.0
The 'Affected alerts' field of an active silence is displaying 'Network error' rather than the alerts affected by this silence. 

It looks like the request to get affected alerts has dropped the web.external-url value from the request. 

Version 0.8.0 was requesting: http://alertmanager.foo.bar:9093/api/v1/alerts?silenced=true&filter=action%3D%22scale-in%22

Version 0.9.0 is requesting: http://alerts?silenced=true&filter=action%3D%22scale-in%22

stuartnelson3:
Both are Strings, so flipping the order between
one invocation and another didn't raise any
argument from the compiler. The baseUrl argument
is actually unnecessary, and has been removed.

fixes #1007


stuartnelson3:
@benclapp do you have the time to build off of this branch and confirm that the fix works?

benclapp:
@stuartnelson3 seems like that's fixed it!

PR #1011, opened 2017-09-29
Initialise notifications_total and notifications_failed_total

### LINKED ISSUE #928: alertmanager_notifications_total and alertmanager_notifications_failed_total should be initialised
These should be initialised to the integrations used.

Conorbro:
Closes #928 

This works but perhaps we could add a utility method of some kind to avoid duplication?

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -50,6 +50,23 @@ var (
 )
 
 func init() {
+	numNotifications.WithLabelValues("email")
+	numNotifications.WithLabelValues("hipchat")
+	numNotifications.WithLabelValues("pagerduty")
+	numNotifications.WithLabelValues("pushover")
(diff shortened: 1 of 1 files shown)


brancz:
I'm wondering whether we should also include the receiver name in this metric, and generally whether we should be initializing the metrics for _all_ available notification providers or just those configured.

brian-brazil:
There could be too many receiver names, potentially hundreds to thousands.

I see the use case for this as spotting when PagerDuty is down, not that an individual receiver is misconfigured.

> generally whether we should be initializing the metrics for all available notification providers or just those configured.

All is simplest, and it may always be configured later on.

brancz:
Makes sense, lgtm :+1:

Issue #666, opened 2017-03-23
Alerts can inhibit themselves


lmb:
I have a bunch of alerts in the form of Prefix_Some_Name. I've set up the following inhibition rule:

```
- source_match_re:
    alertname: Prefix_.+
  target_match:
    alertname: Prefix_Failsafe_Alert
  equal:
    - label1
    - label2
```

As far as I can tell, this alert will inhibit itself, and never be delivered. Depending on how you look at it this is a bug in the alertmanager, or an important gotcha that should be documented.

Happy to contribute a patch either way.

Edit: I should add that I empirically verified this against an older alertmanager. I looked through the current master branch, and AFAICT the behaviour is still the same.

juliusv:
> As far as I can tell, this alert will inhibit itself

Hm yes, that is the behavior I would expect. I guess you expected that alerts generally cannot inhibit themselves?

I think adding a clarfiying sentence to https://github.com/prometheus/docs/blob/master/content/docs/alerting/configuration.md#inhibit_rule would be generally welcome if you think this is something that is confusing.

threemachines:
I also discovered this behavior the hard way. I agree that it's strictly reasonable within the context of how inhibitions are currently structured, but is this behavior considered desirable?

If alerts never inhibited themselves, certain types of alerting noise could be reduced in a way that's currently infeasible. The only use case I can imagine for self-inhibiting alerts is ensuring they never notify anybody, which would surely be better handled by configuring a matching route with no receivers. But maybe I'm just not thinking big enough?

juliusv:
Yeah, I can't come up with a reason why an alert should be able to inhibit itself either. The question is, is it worth introducing an explicit exception for this case? That's ultimately up to the Alertmanager maintainers (@stuartnelson3 @fabxc).

brian-brazil:
I think the general expectation here is that there's some other label you can work off to distinguish these, such as a "dependency" label or similar.

The "Prefix_" being the same on both is also a bit of a smell, that should probably be a label.

lmb:
@brian-brazil that might be correct, but doesn't change the underlying issue.

fabxc:
This should probably a documented caveat. I think it doesn't really happen for typical inhibition usage as there should be some clearly distinguishing labels. Probably an edge case not worth complicating implementation for.

mxinden:
[...]

PR #1012, opened 2017-09-29
Add Jira notifier

alin-amana:
This PR is heavily based on #601 (with ideas from [github.com/fabxc/jiralerts](https://github.com/fabxc/jiralerts)), addressing a number of the concerns raised in #601.

For one, it doesn't open a new issue every `group_interval`. It sets a label of the form `ALERT{alertname="FooTooHigh",job="bar"}` (using the group key labels) on each issue it creates and, if found, it will reopen that issue if it's closed (by transitioning to a configurable `reopen_state`) unless the resolution is *Won't Fix* (or some other, configurable value).

The basic ideas behind it are:
1. Flapping alerts should cause as little noise as possible. As long as the issue is open, there are no automated updates to it.
1. You don't want issues to resolve automatically when the alert stops firing. Someone needs to look at them and explicitly resolve them (and hopefully either fix the issue or tweak the alert threshold in the process).
1. If the issue is simply dismissed and no action is taken, a new alert will reopen it.
1. But optionally (and disabled by default, because it may swallow notifications unintentionally) you can define a "Won't Fix" resolution and any issues closed with that resolution will never be automatically reopened.
1. All fields of an automatically generated issue may be freely changed as long as the `ALERT{...}` label is kept in place. And if it disappears for any reason then you simply get another issue.

In the same way as #601, it allows custom fields to be set on creation and all of them are treated as templates. Tests are on the way.


alin-amana:
Can't seem to get the bindata to match whatever CircleCI ends up generating.

stuartnelson3:
Don't worry about it, we can fix that part later.

@fabxc @brian-brazil thoughts? We've been advising users to create their own integrations via webhooks for a while, but jira is something that I think would have much wider adoption than those. Is this something you all want to consider?

brian-brazil:
You're the maintainer, but I believe all new acceptances have been stopped and the concerns in #601 remain.

alin-amana:
If I might venture an opinion: I don't mind owners/maintainers deciding they don't want a feature, it's a completely reasonable response.

But, assuming there exists a corresponding feature request, it would be nice to close it when the decision is made, with a closing comment clearly stating why, Or, in situations like #666, renamed to "Document that alerts can inhibit themselves". It would save a lot of effort and frustration on everyone's part.

config/notifiers.go
```diff
@@ -395,3 +407,59 @@ func (c *PushoverConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
 	}
 	return checkOverflow(c.XXX, "pushover config")
 }
+
+type JiraConfig struct {
+	NotifierConfig `yaml:",inline" json:",inline"`
+
+	APIURL   string `yaml:"api_url" json:"api_url"`
```
stuartnelson3: this should be type `Secret`

alin-amana:
Great, I'll rework this into a webhook then. I'll be on vacation for the following 2 weeks, though, and I'm not sure how bored I'll be so it may take a while. But I will do it eventually, because we do need it over here.

alin-amana:
So. I finally got around to doing it and here it is: https://github.com/alin-sinpalean/jiralert

Apologies to @fabxc for ripping off the name of his Jira webhook. I dropped an "s" so there's no chance of confusion. :o)

Should I add the link to https://github.com/prometheus/docs/blob/master/content/docs/operating/integrations.md ? And if so, should it replace @fabxc's jiralerts?

alin-amana:
The functionality in this PR has now been released as a fully separate webhook receiver, referenced from the Prometheus documentatoin. Thanks for the feedback and support.

PR #1015, opened 2017-09-29
Sort silences based on current status

### LINKED ISSUE #957: Silences are randomly sorted
On the `/#/silences` endpoint, it seems like silences are in a different order every time the page is refreshed. Is it possible to have them sorted e.g. by start or expiration time?

Conorbro:
Closes #957 

I see mention of a configurable UI to sort silences. 
Does feature this really make sense after filtering when there'll only be a few silences shown?

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -593,7 +593,34 @@ func (api *API) listSilences(w http.ResponseWriter, r *http.Request) {
 		sils = append(sils, s)
 	}
 
-	respond(w, sils)
+	var active, pending, expired, silences []*types.Silence
+
+	for _, s := range sils {
+		switch s.Status.State {
+		case "active":
+			active = append(active, s)
+		case "pending":
+			pending = append(pending, s)
+		case "expired":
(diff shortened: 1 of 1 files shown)


api/api.go
```diff
@@ -593,7 +607,28 @@ func (api *API) listSilences(w http.ResponseWriter, r *http.Request) {
...
+			active = append(active, s)
+		case "pending":
+			pending = append(pending, s)
+		case "expired":
+			expired = append(expired, s)
+		}
+	}
+
+	sort.Sort(ByEndTime(active))
+	sort.Sort(ByStartTime(pending))
```
stuartnelson3: These could be replaced with https://golang.org/pkg/sort/#Slice

brancz:
For a generic `listSilences` call I don't think we should be bucketing here.

api/api.go
Conorbro: I've replaced with your suggestion 👍

PR #1024, opened 2017-10-04
add metric of alertmanager position in mesh

brancz:
The metric that this PR introduces is helpful in order to understand which position in the mesh an Alertmanager instance believes to be in. If a number is non unique, that means that there is a problem with the peers connections in the mesh.

@stuartnelson3 @fabxc


fabxc:
👍

PR #1025, opened 2017-10-04
dispatch: fix race condition in dispatch test

brancz:
This fixes one of the (many) race conditions in the dispatch tests. As there are quite a few I was going to tackle them one by one. This one was a fairly obvious one.

@stuartnelson3 @fabxc

### DIFF
--- a/dispatch/dispatch_test.go
+++ b/dispatch/dispatch_test.go
@@ -3,6 +3,7 @@ package dispatch
 import (
 	"reflect"
 	"sort"
+	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


brancz:
It's just a race in the test itself, this is not actually a race in the Alertmanager code itself. But in order to one day run all tests with the `-race` flag, we'll also need to fix these :)

brancz:
Sorry I don't think I understood your question in my previous comment. We're guarding reading/writing the `last` and `current` time variables.

Issue #1027, opened 2017-10-05
Make matchers and labels in silence view and pre-view selectable


beorn7:
If you view a particular silence (e.g. http://alertmanager/#/silences/<some-hash>), the matchers of the silence and the labels of the silenced alerts are not selectable. That would be handy (as you can copy&paste them somewhere else). They are not UI elements of other kind (as was the problem with the alert view, where labels were also buttons, clickable to add them to the current filter), so it should be quite easy to make them selectable.

@w0rm ?

stuartnelson3:
Closed by #1030

w0rm:
@beorn7 sorry I'm on vacation until tomorrow. @stuartnelson3 thanks for fixing this!

w0rm:
Actually closing this issue.

PR #1030, opened 2017-10-06
Allow selectable matchers on silence view

josedonizetti:
When a label is not used as a button, it allows the text to be copied. https://github.com/prometheus/alertmanager/issues/1027


ui/app/src/Utils/Views.elm
```diff
@@ -22,7 +22,9 @@ labelButton maybeMsg labelText =
     case maybeMsg of
         Nothing ->
             span
-                [ class "btn btn-sm bg-faded btn-secondary mr-2 mb-2" ]
+                [ style [ ( "user-select", "text" ) ]
```
stuartnelson3: This works in chrome, but not firefox. Check out https://github.com/prometheus/alertmanager/blob/master/ui/app/src/Views/AlertList/AlertView.elm#L98-L102.

PR #1031, opened 2017-10-06
cli/silence_add: don't ingore errors when getting current user

iksaif:
I had a system where amtool would segfault on startup because of
that.

### DIFF
--- a/cli/silence_add.go
+++ b/cli/silence_add.go
@@ -56,8 +56,16 @@ var addCmd = &cobra.Command{
 }
 
 func init() {
-	user, _ := user.Current()
-	addCmd.Flags().StringP("author", "a", user.Username, "Username for CreatedBy field")
+	var username string
+
+	user, err := user.Current()
+	if err != nil {
+		fmt.Printf("failed to get the current user, specify one with --author: %v\n", err)
+	} else {
+		username = user.Username
+	}
+
+	addCmd.Flags().StringP("author", "a", username, "Username for CreatedBy field")
 	addCmd.Flags().StringP("expires", "e", "1h", "Duration of silence (100h)")
 	addCmd.Flags().String("expire-on", "", "Expire at a certain time (Overwrites expires) RFC3339 format 2006-01-02T15:04:05Z07:00")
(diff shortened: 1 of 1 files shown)


cli/silence_add.go
```diff
@@ -56,8 +56,16 @@ var addCmd = &cobra.Command{
 }
 
 func init() {
-	user, _ := user.Current()
-	addCmd.Flags().StringP("author", "a", user.Username, "Username for CreatedBy field")
+	user, err := user.Current()
+	username := ""
```
josedonizetti: as a zero value string using `var username string` reads better, what do think?

iksaif: done

cli/silence_add.go
```diff
@@ -56,8 +56,16 @@ var addCmd = &cobra.Command{
...
+
+	if err != nil {
+		fmt.Errorf("Failed to get current user")
```
stuartnelson3: This doesn't actually do anything -- it creates an `error` but doesn't indicate it to the user. It would be useful to let the user know that they don't have a default username available, and that they'll have to provide one (since we require an `author` when creating silences).

iksaif: right, fixed.

iksaif:
Ok, the error message should be correctly printed now :)

cli/silence_add.go
```diff
@@ -56,8 +56,16 @@ var addCmd = &cobra.Command{
 }
 
 func init() {
-	user, _ := user.Current()
-	addCmd.Flags().StringP("author", "a", user.Username, "Username for CreatedBy field")
+	var username string
+
+	user, err := user.Current()
+	if err != nil {
+		fmt.Printf("Failed to get the current user, specify one with --author: %s\n", err)
```
stuartnelson3: It's typical in go to have the first letter of a log or print statement be lowercase, so `failed` instead of `Failed`. Also, err is of type `error`, which is printed out with `%v`. Once that's done, good to go.

iksaif:
fixed

PR #1032, opened 2017-10-06
Fix inhibit race

### LINKED ISSUE #1014: Race in inhibitor shutdown
Same testing round as #1013, another race that was detected.

```
11:43:04 a1 | ==================
11:43:04 a1 | WARNING: DATA RACE
11:43:04 a1 | Write at 0x00c4202984d0 by goroutine 90:
11:43:04 a1 |   github.com/prometheus/alertmanager/inhibit.(*Inhibitor).Stop()
11:43:04 a1 |       /home/brancz/go/src/github.com/prometheus/alertmanager/inhibit/inhibit.go:111 +0xe9
11:43:04 a1 |   main.main.func8()
11:43:04 a1 |       /home/brancz/go/src/github.com/prometheus/alertmanager/cmd/alertmanager/main.go:270 +0x440
11:43:04 a1 |   main.main.func9()
11:43:04 a1 |       /home/brancz/go/src/github.com/prometheus/alertmanager/cmd/alertmanager/main.go:332 +0x68
11:43:04 a1 | Previous read at 0x00c4202984d0 by goroutine 40:
11:43:04 a1 |   github.com/prometheus/alertmanager/inhibit.(*Inhibitor).Run()
11:43:04 a1 |       /home/brancz/go/src/github.com/prometheus/alertmanager/inhibit/inhibit.go:79 +0x173
11:43:04 a1 | Goroutine 90 (running) created at:
11:43:04 a1 |   main.main()
11:43:04 a1 |       /home/brancz/go/src/github.com/prometheus/alertmanager/cmd/alertmanager/main.go:325 +0x22ea
11:43:04 a1 | Goroutine 40 (running) created at:
11:43:04 a1 |   main.main.func8()
11:43:04 a1 |       /home/brancz/go/src/github.com/prometheus/alertmanager/cmd/alertmanager/main.go:286 +0x7a3
11:43:04 a1 |   main.main()
11:43:04 a1 |       /home/brancz/go/src/github.com/prometheus/alertmanager/cmd/alertmanager/main.go:291 +0x1aa3
11:43:04 a1 | ==================
```

This race only showed when shutting down the Alertmanager, and it seems there is a race between [closing and clearing the `stopc` channel](https://github.com/prometheus/alertmanager/blob/9f5f4b2a516d35cfaf196530b277f1d109254569/inhibit/inhibit.go#L110-L111) and possibly [reading from it](https://github.com/prometheus/alertmanager/blob/master/inhibit/inhibit.go#L79). If I understand correctly, this can cause the Alertmanager to deadlock as receiving from a nil channel blocks forever, on shutdown of the inhibitor which happens at shutdown and reload of the Alertmanager.

@stuartnelson3 @fabxc

brancz:
Fixes #1014

@fabxc @stuartnelson3

### DIFF
--- a/inhibit/inhibit.go
+++ b/inhibit/inhibit.go
@@ -14,10 +14,12 @@
 package inhibit
 
 import (
+	"context"
 	"fmt"
 	"sync"
 	"time"
 
+	"github.com/oklog/oklog/pkg/group"
 	"github.com/prometheus/common/log"
 	"github.com/prometheus/common/model"
 
@@ -33,8 +35,8 @@ type Inhibitor struct {
 	rules  []*InhibitRule
 	marker types.Marker
 
-	mtx   sync.RWMutex
-	stopc chan struct{}
+	mtx    sync.RWMutex
+	cancel func()
 }
 
 // NewInhibitor returns a new Inhibitor.
@@ -50,33 +52,26 @@ func NewInhibitor(ap provider.Alerts, rs []*config.InhibitRule, mk types.Marker)
 	return ih
 }
 
-func (ih *Inhibitor) runGC() {
+func (ih *Inhibitor) runGC(ctx context.Context) {
 	for {
 		select {
 		case <-time.After(15 * time.Minute):
 			for _, r := range ih.rules {
 				r.gc()
 			}
-		case <-ih.stopc:
+		case <-ctx.Done():
 			return
 		}
 	}
 }
 
-// Run the Inihibitor's background processing.
-func (ih *Inhibitor) Run() {
-	ih.mtx.Lock()
-	ih.stopc = make(chan struct{})
-	ih.mtx.Unlock()
-
-	go ih.runGC()
-
+func (ih *Inhibitor) run(ctx context.Context) {
 	it := ih.alerts.Subscribe()
 	defer it.Close()
 
 	for {
 		select {
-		case <-ih.stopc:
+		case <-ctx.Done():
 			return
 		case a := <-it.Next():
 			if err := it.Err(); err != nil {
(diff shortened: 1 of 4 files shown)


inhibit/inhibit.go
```diff
@@ -50,33 +52,26 @@ func NewInhibitor(ap provider.Alerts, rs []*config.InhibitRule, mk types.Marker)
...
+func (ih *Inhibitor) runGC(ctx context.Context) {
 	for {
 		select {
 		case <-time.After(15 * time.Minute):
 			for _, r := range ih.rules {
 				r.gc()
 			}
-		case <-ih.stopc:
+		case <-ctx.Done():
 			return
```
josedonizetti: Think you should return `ctx.Err()`

inhibit/inhibit.go
```diff
@@ -50,33 +52,26 @@ func NewInhibitor(ap provider.Alerts, rs []*config.InhibitRule, mk types.Marker)
...
-
+func (ih *Inhibitor) run(ctx context.Context) {
 	it := ih.alerts.Subscribe()
 	defer it.Close()
 
 	for {
 		select {
-		case <-ih.stopc:
+		case <-ctx.Done():
 			return
```
josedonizetti: Also `return ctx.Err()` here

inhibit/inhibit.go
```diff
@@ -98,17 +93,41 @@ func (ih *Inhibitor) Run() {
...
+		ctx context.Context
+	)
+
+	ctx, ih.cancel = context.WithCancel(context.Background())
+	gcCtx, gcCancel := context.WithCancel(ctx)
+	runCtx, runCancel := context.WithCancel(ctx)
+
+	g.Add(func() error {
+		ih.runGC(gcCtx)
+		return nil
```
josedonizetti: If you return the context error, you can return it here. And the same for the ih.run below.

inhibit/inhibit.go
brancz: `ctx.Err()` just returns why a context was canceled or deadline exceeded. It doesn't actually tell us anything here.

```
        // If Done is not yet closed, Err returns nil.
        // If Done is closed, Err returns a non-nil error explaining why:
        // Canceled if the context was canceled
        // or DeadlineExceeded if the context's deadline passed.
        // After Err returns a non-nil error, successive calls to Err return the same error.
        Err() error
```

josedonizetti: true, but if non nil isn't worth logging?

brancz: I don't think so, because we explicitly use the context to cancel the execution in the `Inhibitor.Stop` method.

josedonizetti: Gotcha 👍

fabxc: Yes, the error is always `!=` nil if the context was canceled. If we expect this to happen, there's generally no point in checking.

PR #1033, opened 2017-10-08
Add usage question note to CONTRIBUTING.md

mxinden:
Usage questions should not be discussed via GitHub issues but instead in
the Prometheus users Google group. Redirecting users from GitHub issues
to the Prometheus users Google group is happening quite often. To make
things easier for maintainers and users this patch adds a note to
CONTRIBUTING.md in the repo. When opening up a new issue in GitHub,
there is a link to the CONTRIBUTING.md file at the top of the dialog.

@stuartnelson3 what do you think?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # Contributing
 
-Prometheus uses GitHub to manage reviews of pull requests.
+Alertmanager uses GitHub to manage reviews of pull requests.
 
 * If you have a trivial fix or improvement, go ahead and create a pull request,
   addressing (with `@...`) the maintainer of this repository (see


josedonizetti:
This is awesome, but I don't think people do click on that section, they just ignore it and create the issue directly (i'm one of those people. :/). Maybe this same text could be used in the issue template, some repos (like kubernetes) do have a template for issues, with questions the user should answer before opening it, using your idea to have such a template maybe would be more effective. Not sure if it is better or not, just an opinion.

What you think @mxinden @stuartnelson3?

CONTRIBUTING.md
```diff
@@ -1,6 +1,6 @@
 # Contributing
 
-Prometheus uses GitHub to manage reviews of pull requests.
```
stuartnelson3: Yeah, that's a good change :)

mxinden:
@stuartnelson3 @josedonizetti Adding an issue template is a great idea. I don't have admin access, @stuartnelson3 could you configure it?

@stuartnelson3 I will update the PR to only contain the s/Prometheus/Alertmanager.

Issue #1034, opened 2017-10-09
UI: Option to hide inhibited alerts


beorn7:
Similar to silenced alerts, there should be an option to hide inhibited alerts.
After a short discussion with @w0rm , we concluded that it's most likely the same situation, i.e. we don't need to bloat the UI with two separate checkboxes, but rather have a checkbox _Show Silenced/Inhibited_ rather than the current _Show Silenced_. We can do this as a first attempt and see if it is sufficient. If people then have legitimate need for separate checkboxes, we can still reconsider.

@stuartnelson3 would you be able to implement the necessary backend changes? Perhaps have the flexible separate filtering in the API so that the UI is free to decide about separate or joined checkboxes.

caarlos0:
Working on it...

PR #1039, opened 2017-10-12
also hide inhibited alerts

caarlos0:
refs #1034

I feel like it would be better to rename the form parameter to something else, since it's not only silenced alerts anymore... what do you folks think?

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -342,6 +342,9 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 		if !showSilenced && len(status.SilencedBy) != 0 {
 			continue
 		}
+		if !showSilenced && len(status.InhibitedBy) != 0 {
+			continue
+		}
 
 		apiAlert := &dispatch.APIAlert{
 			Alert:       &a.Alert,
--- a/api/api.go
+++ b/api/api.go
@@ -264,9 +264,10 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 		re  *regexp.Regexp
 		// Initialize result slice to prevent api returning `null` when there
 		// are no alerts present
-		res          = []*dispatch.APIAlert{}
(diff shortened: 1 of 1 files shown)


brancz:
Internally this is often referred to as "muted".

caarlos0:
@beorn7 @stuartnelson3 what do you guys think about renaming this form param to `muted`?

beorn7:
We could do that. Or we create a new parameter `showInhibited` to give the UI the option to select those separately even if we wouldn't use it for now.

brancz:
Yes I think it would be good to be able to view inhibited, silenced, and firing alerts in all combinations.

stuartnelson3:
+1 to beorn's suggestion. Keep it flexibile in the API even if we don't make it an option from the UI.

caarlos0:
@stuartnelson3 @beorn7 @brancz done!

beorn7:
@stuartnelson3 I'd better leave the code level review to you. (One day, I'll get familiar with Alertmanager internals, but not today…)

api/api.go
```diff
@@ -343,6 +359,10 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 			continue
 		}
 
+		if !showInhibited && len(status.InhibitedBy) != 0 {
```
brancz: still hoping that we'll one day have a real index around this :slightly_smiling_face:

caarlos0:
> Generally looks good so far, but I would expect that I can toggle the front end to show me the inhibited alerts. As far as I can tell from this PR, this only handles the API filtering part. (if that is meant to be done in a subsequent PR, I'm fine with that as well)

Yeah, as far as I understood the original issue was about the backend only. I can change the frontend as well if required...

beorn7:
@w0rm will take care of the frontend part (IIRC).

brancz:
In that case, this lgtm

PR #1043, opened 2017-10-18
Add support for configuring and sending contexts to PagerDuty when triggering alerts.

kadaan:
When Alertmanager creates PagerDuty incidents it appends all of the alert data as custom details.  Links within the custom details are clickable, but they are not easily discoverable.  Alertmanager should support populating the Contexts element (from [PagerDuty Event API V1 ](https://v2.developer.pagerduty.com/docs/trigger-events))  and Links/Images when [PagerDuty Event API V2](https://v2.developer.pagerduty.com/docs/send-an-event-events-api-v2) supported.

This PR is still a work in progress, but the goal is to populate the Links section in the PagerDuty UI as shown below.
screen shot 2017-10-19 at 1 44 40 am

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 /.build
 /.release
 /.tarballs
+/.idea
 
 !/doc/examples/simple.yml
 !/circle.yml
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -178,12 +178,22 @@ func (c *EmailConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 type PagerdutyConfig struct {
 	NotifierConfig `yaml:",inline" json:",inline"`
 
-	ServiceKey  Secret            `yaml:"service_key,omitempty" json:"service_key,omitempty"`
-	URL         string            `yaml:"url,omitempty" json:"url,omitempty"`
-	Client      string            `yaml:"client,omitempty" json:"client,omitempty"`
-	ClientURL   string            `yaml:"client_url,omitempty" json:"client_url,omitempty"`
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -6,6 +6,7 @@
 /.build
 /.release
 /.tarballs
+/.idea
```
stuartnelson3: please remove user-specific gitignore changes

brian-brazil:
With this PR, how would one send one context per alert?

kadaan:
As is it won't.  Your comments in IRC showed that.  To make that work we could loop over all alerts and for each alert expand all context.  That should work, but it isn't really that graceful.

As an alternative, could we just allow the entire PagerDuty (or Webhook) request body to be a template?  Then I could have full control and do the looping via Range.

kadaan:
@stuartnelson3 Updated the description and provided an image of what the end goal in PagerDuty is.

.gitignore
kadaan: Why?  Isn't to good to ignore common things?

kadaan:
@stuartnelson3
> Is there something about the current configuration that makes you want to add the link context?
The link in the context is much more discoverable.

> Is there a reason you just wanted link, and not image?
Ideally, this PR would support specifying images as well as the PagerDuty API supports that.

.gitignore
stuartnelson3: By what definition of "common" ? :) we use our gitignore for project specific ignores; if you have your own files you need to ignore because of your environment, please use your global gitignore.

kadaan: Fair enough.  I'll remove this.

PR #1045, opened 2017-10-19
Add limit to OpsGenie message

josedonizetti:
[OpsGenie API](https://docs.opsgenie.com/docs/alert-api-deprecated) has a limit of 130 characters to a message.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -762,10 +762,16 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 		apiURL = n.conf.APIHost + "v1/json/alert/close"
 		msg = &opsGenieCloseMessage{&apiMsg}
 	default:
+		message := tmpl(n.conf.Message)
+		if len(message) > 130 {
+			message = message[:127]+"..."
+			level.Debug(n.logger).Log("msg", "Truncated message to %q due to OpsGenie message limit", "truncated_message", message, "incident", key)
+		}
+
 		apiURL = n.conf.APIHost + "v1/json/alert"
 		msg = &opsGenieCreateMessage{
 			opsGenieMessage: &apiMsg,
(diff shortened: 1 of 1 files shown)


notify/impl.go
```diff
@@ -753,10 +753,16 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 		apiURL = n.conf.APIHost + "v1/json/alert/close"
 		msg = &opsGenieCloseMessage{&apiMsg}
 	default:
+		message := tmpl(n.conf.Message)
+		if len(message) > 130 {
+			message = message[:130]
```
stuartnelson3: From an end-user standpoint, I think doing an ellipsis would be a good clue to the alertreceiver that something is being truncated, so they can update their config message. Something like

```go
if len(message) > 130 {
    message = message[:127]+"..."
    log.With...
}
```

Thoughts?

I think this is being done somewhere else but I'm not sure..
@brancz @brian-brazil

brian-brazil: In 2 places we just truncate, in 1 place we add an ellipsis.

josedonizetti: I like the ellipsis as well. Should we change all other msgs to be consistent using it?

stuartnelson3: Let's do the ellipsis here, and if it makes sense in the other two places, move those to ellipses as well.

brancz: ellipsis is appropriate I think

stuartnelson3: (the ellipses changes for the other places should be in a different pr, if there was any confusion)

josedonizetti:
@stuartnelson3 done. Let me know if you want me to squash the commits, or it's fine separated.

josedonizetti:
@stuartnelson3 I've rebased it from master to update logging to use the go-kit way. Also squashed it in 1 commit.

PR #1047, opened 2017-10-19
Add ellipsis to truncated messages

josedonizetti:
UX change, ellipsis gives context to the user that the msg was truncated.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -953,11 +953,11 @@ func (n *Pushover) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 	message := tmpl(n.conf.Message)
 	parameters.Add("title", title)
 	if len(title) > 512 {
-		title = title[:512]
+		title = title[:509] + "..."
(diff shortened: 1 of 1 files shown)


notify/impl.go
```diff
@@ -953,11 +953,11 @@ func (n *Pushover) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 	message := tmpl(n.conf.Message)
 	parameters.Add("title", title)
 	if len(title) > 512 {
-		title = title[:512]
+		title = title[:509] + "..."
 		log.With("incident", key).Debugf("Truncated title to %q due to Pushover message limit", title)
 	}
 	if len(title)+len(message) > 512 {
-		message = message[:512-len(title)]
+		message = message[:509-len(title)] + "..."
```
stuartnelson3: If `len(title) == 512`, this will result in `message[:-3]`, which will panic

josedonizetti:
Close it in favor of https://github.com/prometheus/alertmanager/pull/1055

Issue #619, opened 2017-02-02
OpsGenie api_key should be template-expanded


aecolley:
The PagerDuty service key is template-expanded inside Notify, which is useful for keeping access keys out of the config repository. Please do the same for OpsGenie.

linkinpark342:
The HipChat [https://github.com/prometheus/alertmanager/blob/v0.5.1/notify/impl.go#L681](implementation) is similarly in need of this:

mxinden:
@aecolley Is this still relevant? If so, would you mind creating a PR for this?

brian-brazil:
> which is useful for keeping access keys out of the config repository

They need to go in the configuration one way or the other, templating doesn't help that.

linkinpark342:
Templating helps in that, with it, you can check in your prometheus configuration for all to see, but have your keys managed elsewhere (ansible-vault or some such). I think this is still relevant for the product, especially for consistency as it looks like the PagerDuty integration has it. However our org (both aecolley and myself) gave up on opsgenie so we're probably not going to spend too much time on this :P. Though did referenced commit #1081 take care of this already?

brian-brazil:
> You can check in your prometheus configuration for all to see, but have your keys managed elsewhere (ansible-vault or some such).

If you're using ansible, then you can pull the secret out of vault with the `template` module. This is the sort of solution that would be recommended.

The pagerduty service keys are templatable because they are usable for routing. The OpsGenie api_key is not.

PR #1054, opened 2017-10-24
Add support for PagerDuty API v2

### LINKED ISSUE #877: PagerDuty v1 API will be unsupported from February 2018
See:
https://v2.developer.pagerduty.com/v2/docs/api-v2-frequently-asked-questions

> API v1 will be supported until February 6, 2018

We should migrate to v2.

scarvel8:
This closes #877 and maintains backwards compatibility with current configurations

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -283,7 +283,7 @@ func TestEmptyFieldsAndRegex(t *testing.T) {
 			HipchatURL:       "https://hipchat.foobar.org/",
 			SlackAPIURL:      "mysecret",
 			SMTPRequireTLS:   true,
-			PagerdutyURL:     "https://events.pagerduty.com/generic/2010-04-15/create_event.json",
+			PagerdutyURL:     "https://events.pagerduty.com/v2/enqueue",
 			OpsGenieAPIHost:  "https://api.opsgenie.com/",
 			VictorOpsAPIURL:  "https://alert.victorops.com/integrations/generic/20131114/alert/",
 		},
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -438,18 +438,116 @@ const (
 )
 
 type pagerDutyMessage struct {
-	ServiceKey  string            `json:"service_key"`
-	IncidentKey string            `json:"incident_key"`
-	EventType   string            `json:"event_type"`
-	Description string            `json:"description"`
+	RoutingKey  string            `json:"routing_key,omitempty"`
+	ServiceKey  string            `json:"service_key,omitempty"`
+	DedupKey    string            `json:"dedup_key,omitempty"`
+	IncidentKey string            `json:"incident_key,omitempty"`
+	EventType   string            `json:"event_type,omitempty"`
+	Description string            `json:"description,omitempty"`
+	EventAction string            `json:"event_action"`
+	Payload     *pagerDutyPayload `json:"payload"`
 	Client      string            `json:"client,omitempty"`
 	ClientURL   string            `json:"client_url,omitempty"`
 	Details     map[string]string `json:"details,omitempty"`
 }
 
+type pagerDutyPayload struct {
+	Summary       string            `json:"summary"`
+	Source        string            `json:"source"`
+	Severity      string            `json:"severity"`
+	Timestamp     string            `json:"timestamp,omitempty"`
+	Component     string            `json:"component,omitempty"`
+	Group         string            `json:"group,omitempty"`
+	CustomDetails map[string]string `json:"custom_details,omitempty"`
+}
+
+func (n *PagerDuty) notifyV1(ctx context.Context, eventType string, key string, tmpl func(string) string, details map[string]string, as ...*types.Alert) (bool, error) {
+
+	msg := &pagerDutyMessage{}
+
+	var err error
+
+	level.Info(n.logger).Log("msg", "PagerDuty v1 API will no longer be supported: https://v2.developer.pagerduty.com/v2/docs/api-v2-frequently-asked-questions")
+	msg.ServiceKey = string(n.conf.ServiceKey)
+	msg.EventType = eventType
+	msg.IncidentKey = hashKey(key)
+	msg.Description = tmpl(n.conf.Description)
+	msg.Details = details
+	n.conf.URL = "https://events.pagerduty.com/generic/2010-04-15/create_event.json"
+
+	if eventType == pagerDutyEventTrigger {
+		msg.Client = tmpl(n.conf.Client)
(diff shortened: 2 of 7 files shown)


notify/impl.go
```diff
@@ -438,18 +438,32 @@ const (
 )
 
 type pagerDutyMessage struct {
-	ServiceKey  string            `json:"service_key"`
-	IncidentKey string            `json:"incident_key"`
-	EventType   string            `json:"event_type"`
-	Description string            `json:"description"`
+	RoutingKey  string            `json:"routing_key,omitempty"`
+	ServiceKey  string            `json:"service_key,omitempty"`
+	DedupKey    string            `json:"dedup_key"`
```
brian-brazil: What is this and the incident key?

I suspect neither should be user configurable.

config/config_test.go
```diff
@@ -276,16 +276,17 @@ func TestEmptyFieldsAndRegex(t *testing.T) {
...
-			VictorOpsAPIURL:  "https://alert.victorops.com/integrations/generic/20131114/alert/",
+			ResolveTimeout:    model.Duration(5 * time.Minute),
+			SMTPSmarthost:     "localhost:25",
+			SMTPFrom:          "alertmanager@example.org",
+			HipchatAuthToken:  "mysecret",
+			HipchatURL:        "https://hipchat.foobar.org/",
+			SlackAPIURL:       "mysecret",
+			SMTPRequireTLS:    true,
+			PagerdutyURL:      "https://events.pagerduty.com/v2/enqueue",
+			PagerdutySeverity: "error",
```
brian-brazil: We don't have similar settings as global config options for other notifiers.

notify/impl.go
```diff
@@ -474,17 +488,41 @@ func (n *PagerDuty) Notify(ctx context.Context, as ...*types.Alert) (bool, error
...
+
+	if n.conf.ServiceKey != "" {
+		msg.ServiceKey = tmpl(string(n.conf.ServiceKey))
+		msg.EventType = eventType
+		msg.IncidentKey = hashKey(key)
+		msg.Description = tmpl(n.conf.Description)
+		msg.Details = details
+		n.conf.URL = "https://events.pagerduty.com/generic/2010-04-15/create_event.json"
+	} else {
+		msg.RoutingKey = tmpl(string(n.conf.RoutingKey))
```
josedonizetti: RoutingKey is an integration key provided by pagerduty, should not be processed like a template.

notify/impl.go
```diff
@@ -474,17 +488,41 @@ func (n *PagerDuty) Notify(ctx context.Context, as ...*types.Alert) (bool, error
...
+		msg.EventAction = eventType
+		msg.DedupKey = hashKey(key)
+		if eventType == pagerDutyEventTrigger {
+			msgpayload := &pagerDutyPayload{
+				Summary:       tmpl(n.conf.Description),
+				Source:        n.conf.Client,
+				Severity:      n.conf.Severity,
+				CustomDetails: details,
+			}
+			if n.conf.Component == "" {
```
josedonizetti: You setting it on the payload if empty?

notify/impl.go
```diff
@@ -474,17 +488,41 @@ func (n *PagerDuty) Notify(ctx context.Context, as ...*types.Alert) (bool, error
...
+				msgpayload.Component = n.conf.Component
+			}
+			if n.conf.Group == "" {
```
josedonizetti: You setting it on the payload if empty?

notify/impl.go
```diff
@@ -474,17 +488,41 @@ func (n *PagerDuty) Notify(ctx context.Context, as ...*types.Alert) (bool, error
...
-	msg := &pagerDutyMessage{
-		ServiceKey:  tmpl(string(n.conf.ServiceKey)),
-		EventType:   eventType,
-		IncidentKey: hashKey(key),
-		Description: tmpl(n.conf.Description),
-		Details:     details,
+	msg := &pagerDutyMessage{}
+
+	if n.conf.ServiceKey != "" {
+		msg.ServiceKey = tmpl(string(n.conf.ServiceKey))
```
josedonizetti: Add some logging here saying that the v1 support is deprecated. Maybe add the user to the pagerduty faq where they day the date that they still stop supporting it.

config/config_test.go
scarvel8: removing this option from the global config

notify/impl.go
scarvel8: :+1:

notify/impl.go
scarvel8: good catch, thanks

notify/impl.go
scarvel8: Strangely, ServiceKey (which is the v1 equivalent) was being processed as a template as well in the past

notify/impl.go
scarvel8: they are not

notify/impl.go
```diff
@@ -474,17 +488,42 @@ func (n *PagerDuty) Notify(ctx context.Context, as ...*types.Alert) (bool, error
...
 
-	msg := &pagerDutyMessage{
-		ServiceKey:  tmpl(string(n.conf.ServiceKey)),
-		EventType:   eventType,
-		IncidentKey: hashKey(key),
-		Description: tmpl(n.conf.Description),
-		Details:     details,
+	msg := &pagerDutyMessage{}
+
+	if n.conf.ServiceKey != "" {
```
stuartnelson3: Could you branch here into two separate functions, something like 
```go
if n.conf.ServiceKey != "" {
	level.Info(n.logger).Log("msg", "PagerDuty v1 API will no longer be supported: https://v2.developer.pagerduty.com/v2/docs/api-v2-frequently-asked-questions")
	return n.notifyV1(ctx, as...)
else {
        return n.notifyV2(ctx, as...)
}
```

And have the same for the retry logic?

scarvel8:
@josedonizetti retry logic added
@stuartnelson3 added the branching logic for v1 vs v2

Squashed commits

josedonizetti:
@scarvel8 I've mistakenly asked you to not use templates for the PD configuration keys (RoutingKey, and ServiceKey), but apparently, this is a good way to keep the keys outside of the source code repository. See https://github.com/prometheus/alertmanager/issues/619

@stuartnelson3 Does it make sense?

brian-brazil:
> this is a good way to keep the keys outside of the source code repository. See #619 (sorry :()

This is incorrect.

brian-brazil:
To expand, we currently template things used for routing but not things which are purely used as credentials.

josedonizetti:
Thanks for the answer @brian-brazil

PR #1067, opened 2017-11-01
Clarify variable names in Dispatcher.processAlert()

juliusv:
A single entry in aggrGroups is just a single group, not plural.


brancz:
lgtm :+1:

Issue #1063, opened 2017-10-31
For flapping alerts, FIRING notifications are only sent once when `send_resolved: false` is set


juliusv:
Given:

- Prometheus 2.0 at current master *or* Prometheus 1.8.1 (doesn't matter)
- Alertmanager at current master

...and the intentionally flapping alerting rule:

```
groups:
- name: demo-service-alerts
  interval: 15s
  rules:
  - alert: TimeBasedTestAlert
    expr: vector(time()) % 600 < 300
    for: 1m
    labels:
      severity: test
```

...and the AM config:

```
route:
  receiver: foo
  group_by: ['alertname']
  group_wait: 30s
  group_interval: 5m
  repeat_interval: 3h

receivers:
- name: 'foo'
  slack_configs:
  - api_url: 'https://hooks.slack.com/services/XXX/XXX'
    channel: '#test-alerts'
    send_resolved: false
```

...the initial FIRING notification always makes it to Slack, but for further on/off cycles of the alert, no further FIRING notifications get sent to Slack even after the `group_interval` has passed multiple times.

When setting `send_resolved: true`, both RESOLVED and FIRING notifications appear at regular intervals as expected (every 10 minutes).

It doesn't seem like this always has been the case. At least in Cortex, which still vendors old AM packages, this is handled correctly. I noticed this when upgrading Cortex to current AM packages. But I haven't pinpointed the exact commit that caused this yet in AM.

juliusv:
A `git bisect` suggests this problem was introduced by https://github.com/prometheus/alertmanager/commit/309c6af4b2a05fb8a90236859c847ec6f2a6f2de. /cc @fabxc

josedonizetti:
@juliusv I wonder what is the expected behavior here? Because I had the impression that an alert successfully sent would only be resent after the repeat_interval had elapsed. At least that what I got from the documentation.

```
# How long to wait before sending a notification again if it has already
# been sent successfully for an alert. (Usually ~3h or more).
[ repeat_interval: <duration> | default = 4h ]
```

[...]

PR #1068, opened 2017-11-01
Fix alerting when send_resolved false

josedonizetti:
Fixes https://github.com/prometheus/alertmanager/issues/1063

The inconsistency when alerting with send_resolved set to false happens because the notification logs are not updated when send_resolved is false and a 'resolved' alert arrives. 
When the send_resolved is set to true and a resolved alert arrives the logs will be updated because the pipeline is not canceled at the DedupStage, it will move on until the SetNotifiesStage which will update the notification log here -> https://github.com/prometheus/alertmanager/blob/master/notify/notify.go#L652

The solution will only update the nflog if there are new resolved alerts, the sendResolved is set to false, and the pipeline would be canceled because `needsUpdate` returned false.
The check to validate if new resolved alerts exist is the same done when send_resolved is true and will or not move on with the pipeline https://github.com/prometheus/alertmanager/blob/master/notify/notify.go#L497

@juliusv @stuartnelson3

### DIFF
--- a/config/notifiers_test.go
+++ b/config/notifiers_test.go
@@ -1,8 +1,8 @@
 package config
 
 import (
-	"testing"
 	"gopkg.in/yaml.v2"
+	"testing"
 )
 
 func TestEmailToIsPresent(t *testing.T) {
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -552,6 +552,18 @@ func (n *DedupStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.Al
 	} else if ok {
 		return ctx, alerts, nil
 	}
+
+	// If the pipeline is not configured to send_resolved, but the resolvedSet changed,
+	// we need to write it on the log to update the firing status existent on it, which
+	// would avoid subsequent firing alerts to be send because the status would still be
+	// stored as active.
+	// In case the send_resolved is true, this will happen at the end of the pipeline
+	// (SetNotifiesStage) but when it is false the pipeline stops here.
+	if !n.sendResolved && !entry.IsResolvedSubset(resolvedSet) {
+		level.Debug(l).Log("msg", "Updating the notification log", "stage", "DedupStage")
+		n.nflog.Log(n.recv, gkey, firing, resolved)
+	}
+
 	return ctx, nil, nil
 }
 
@@ -649,5 +661,6 @@ func (n SetNotifiesStage) Exec(ctx context.Context, l log.Logger, alerts ...*typ
 		return ctx, nil, fmt.Errorf("resolved alerts missing")
 	}
 
+	level.Debug(l).Log("msg", "Updating the notification log", "stage", "SetNotifiesStage")
 	return ctx, alerts, n.nflog.Log(n.recv, gkey, firing, resolved)
 }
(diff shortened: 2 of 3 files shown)


juliusv:
Ah thanks. I was studying this code too today to learn more about it. An idea I got was to just remove any notion of `SendResolved()` in the `Dedup` stage and leave it up to the final notification integration whether to actually send out a notification or not (as in, don't send a notification if there are only resolved alerts). That way resolved alerts would still be recorded in the nflog. This comment here in the integration even suggests that this late-stage filtering is the right way:

https://github.com/prometheus/alertmanager/blob/master/notify/impl.go#L71-L72

But to be honest, I have not tested that yet (want to, though). Will also study your approach in a bit.

juliusv:
Sketched down my idea in code (untested so far): https://github.com/prometheus/alertmanager/commit/168be980a8726e84119dbbe17b8e0e5fae028c48

brancz:
Thanks for having a look at this you two! I've been wanting to fix this (and a couple of other things) for a while but haven't gotten around doing so.

I believe the approach that @juliusv skeched out is approximately what we should do here. The gossip implementation is already hard enough to debug, we should limit ourselves to where we call out to trigger gossiping of data. I initially developed the retry logic, but looking at it today, I feel what we call the `RetryStage` is really the `SendStage` and this knowledge seems like it would belong in the `SendStage`. What do you think?

juliusv:
So I think this PR here would work, but I wonder if the other approach is nicer (if it works) since it keeps any writes to the nflog at the end of the pipeline where it happens anyway.

juliusv:
Whoops, didn't see your other replies before my last one, sorry (missing page reload). Will reply later!

brancz:
> @juliusv Your approach is way better, definitely the way to go. But there still something we need to track on it, because testing to send the alerts didn't work, even after a resolved the subsequent alert wasn't dispatched. It also broked an acceptance test.

I'm not sure I understand, can you elaborate?

> So I think this PR here would work, but I wonder if the other approach is nicer (if it works) since it keeps any writes to the nflog at the end of the pipeline where it happens anyway.

Agreed.

josedonizetti:
@juliusv Let's fix it using your approach, will simplify the code. And the comment you mentioned can be found in other places in the code (about resolved to go all the way to SetNotifiesStage), so this is what is expected. https://github.com/prometheus/alertmanager/blob/master/test/acceptance/send_test.go#L332

juliusv:
> I feel what we call the `RetryStage` is really the `SendStage` and this knowledge seems like it would belong in the `SendStage`. What do you think?

@brancz Yep yep. So in my version it is in there (although I think we might want to do the renaming of the stage separately if we want that). Will complete my change then and submit a PR!

juliusv:
@josedonizetti 

> But there still something we need to track on it, because testing to send the alerts didn't work, even after a resolved the subsequent alert wasn't dispatched. It also broked an acceptance test.

Yep, will look into both issues.

> Gonna close this PR here, k? :)

Sounds good, but many thanks!

josedonizetti:
@juliusv Your branch does work, I messed up on the first test. :/
Sorry for all the messages.

juliusv:
@josedonizetti Even better, good to know :)

PR #1071, opened 2017-11-01
Fix notifications for flapping alerts

juliusv:
Fixes https://github.com/prometheus/alertmanager/issues/1063

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -236,7 +236,7 @@ func createStage(rc *config.Receiver, tmpl *template.Template, wait func() time.
 		}
 		var s MultiStage
 		s = append(s, NewWaitStage(wait))
-		s = append(s, NewDedupStage(notificationLog, recv, i.conf.SendResolved()))
+		s = append(s, NewDedupStage(notificationLog, recv))
 		s = append(s, NewRetryStage(i))
 		s = append(s, NewSetNotifiesStage(notificationLog, recv))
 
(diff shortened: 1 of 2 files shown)


juliusv:
@brancz @josedonizetti

notify/notify.go
```diff
@@ -561,6 +559,19 @@ func NewRetryStage(i Integration) *RetryStage {
 
 // Exec implements the Stage interface.
 func (r RetryStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.Alert) (context.Context, []*types.Alert, error) {
+	// If we shouldn't send notifications for resolved alerts, but there are only
+	// resolved alerts, report them all as successfully notified (we still want the
+	// notification log to log them).
+	if !r.integration.conf.SendResolved() {
+		firing, ok := FiringAlerts(ctx)
+		if !ok {
+			return ctx, alerts, fmt.Errorf("firing alerts missing")
```
brancz: If we error out here the `MultiStage` will abort the further execution, therefore the resolved subset of this set of alerts will not be gossiped. My understanding is that if there are no firing alerts in this set of alerts, but there are resolved alerts, we still want to gossip that.

juliusv: Yep. OTOH isn't this the same thing we are doing here? https://github.com/prometheus/alertmanager/blob/master/notify/notify.go#L627

AFAICS if this errors it means there's a bug in the code that should be fixed, as the context key should always be set earlier in the pipeline?

brancz: You're absolutely right.

Issue #504, opened 2016-09-10
Support for secrets set in ENV variables


kaihendry:
Currently I don't see any support for SMTP_AUTH_PASSWORD to be set via typically Docker env variables. Am I mistaken?

brian-brazil:
The chosen approach is to put them in the config file.

There's many many possible ways to provide configuration, for sanity we have to choose just one of them.

kaihendry:
It's fairly common practice now to provide secrets via ENV variables & it would be good if you could support this approach together with your configuration. Ideally your configuration can be set by an ENV variable like shell.

brian-brazil:
You're free to put scripts around our configs to provide this for yourself.

kaihendry:
But that would mean forking the Docker image, right? So I can add a little m4 substitution or something.

Andor:
@kaihendry you can mount directory with your config into container with alertmanager.

pydolan:
@brian-brazil -- Is there a thread somewhere that has the discussion to end support for environment variables? I'm trying to understand why this design decision was made.

From a programming standpoint, it would be fairly trivial to modify `configuration/configuration.go` to loop over all configuration attributes, looking for a matching environment variable, and overwriting the stored value if one is found.

Given that many organizations store configuration files in source repositories (e.g., git) and that is a highly-recommended practice to not store secrets in source repositories, allowing prometheus/alertmanager users to store secrets in environment variables (without having to use `awk` hacks) seems like a great idea.

Thank you in advance for any clarification you can provide.

brian-brazil:
It was a matter of consistency and allowing for many many secrets to be provided.

We want to have just one way to configure a given setting, and users can then build on that.

fvigotti:
If possible I would like to vote for some kind of env interpolation for config files, this is a common best practice and would be a great feature in prometheus alertmanager..

seanhoughton:
If environment variables aren't going to be supported what about encrypted secrets in the config? The eyaml (https://github.com/TomPoulton/hiera-eyaml) project works well for Puppet.

brian-brazil:
That is not something we plan on supporting. Once again, too many ways to do it. You're free to use your own tooling around this.

groob:
I'd like to chime in and say that ENV variables should be supported as well. given that so many of us are using prometheus with docker/k8s deployment, it's very odd to have to deal with the config file and not have builtin support for interpolating secrets/env variables.

shaneramey:
[...]

PR #1075, opened 2017-11-02
Add env & expandenv template functions

balboah:
Using environment variables is a common practice when deploying
docker containers. This allows for using environment variables in notifications.

This can also enable functionality described in https://github.com/prometheus/alertmanager/issues/504 after adding parsing to the `Secret` types.

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -16,6 +16,7 @@ package template
 import (
 	"bytes"
 	"net/url"
+	"os"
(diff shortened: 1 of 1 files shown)


brian-brazil:
As was explained in #504, we do not allow configuration from the environment. If you want this it is recommended that you use your configuration management system to do so.

This would also go against our security model.

PR #1076, opened 2017-11-03
Refactor silence constants

josedonizetti:
Clean up silence constants duplicated; I chose to keep the ones declared inside the types package because it's where the silence struct is also declared. 

@stuartnelson3 @brancz


josedonizetti:
@stuartnelson3 It appears that not only the State constants are duplicated, there is a few parts duplicated for pb.Silence and types.Silence. Maybe I should stay the constants duplicated as well, and only change the code that is using the hardcoded strings. Let me know if you what you prefer.

PR #1080, opened 2017-11-05
Remove old silence code

josedonizetti:
The `(silence) Delete()` method was added at this commit https://github.com/prometheus/alertmanager/commit/c0103dd8c61a9e9e57a3c888a9ae579e24a09b8d a code related to the old silence provider, that was deleted at this commit https://github.com/prometheus/alertmanager/commit/3b5dfa8aba9db19c4c175360d69cce8f0c3d9b74, and now it not used anymore.

But more importantly for me is that the semantically now this method should be called `Expired()` https://github.com/prometheus/alertmanager/blob/master/silence/silence.go#L493, maybe instead of removing I should just rename it.

what do you think?
@stuartnelson3 @brancz

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -14,8 +14,6 @@
 package types
 
 import (
-	"fmt"
-	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


josedonizetti:
The `Delete()` method is not the only part not used of types.Silence, also:

- `Mutes()` - https://github.com/prometheus/alertmanager/blob/master/types/types.go#L433
- `Init()` https://github.com/prometheus/alertmanager/blob/master/types/types.go#L419
- `Validate()` https://github.com/prometheus/alertmanager/blob/master/types/types.go#L385

The `Validate()` method specifically is duplicated inside pb.Silence. Do you think it's better to keep them or just remove then all?

stuartnelson3:
I'm fine with you removing it if we're not using it

josedonizetti:
@stuartnelson3 Done.

Not sure why every time I amend code circleci fails. Is there a way to for me to rerun the tests when? Thanks.

stuartnelson3:
I believe they need to be re-run by someone with access to the account. No worries, seems to have passed.

PR #1081, opened 2017-11-05
Template-expand OpsGenie api_key

### LINKED ISSUE #619: OpsGenie api_key should be template-expanded

Tom-Fawcett:
Closes #619 

Allows keeping the api_key in a separate template file, useful for secret management.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -779,7 +779,10 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 		return true, err
 	}
 	req.Header.Set("Content-Type", contentTypeJSON)
-	req.Header.Set("Authorization", fmt.Sprintf("GenieKey %s", n.conf.APIKey))
+	req.Header.Set("Authorization", fmt.Sprintf("GenieKey %s", tmpl(string(n.conf.APIKey))))
+	if err != nil {
+		return false, err
+	}
 
 	resp, err := ctxhttp.Do(ctx, http.DefaultClient, req)
 


brian-brazil:
As far as I'm aware, this is not used for routing so there's no need to have it templatable. What is your use case?

Tom-Fawcett:
Hi Brian,

With Kubernetes I would like to keep the main Alertmanager config in a config object, but keep the secrets (in my case the OpsGenie api_key) in a secret object.

It would be ideal to be able to achieve this without:
1. Increasing the complexity of my setup.
2. More importantly, increasing the complexity of Alertmanager.

I believe this PR achieves this.

Using a templated secret is a preexisting concept (see Pushover and Pagerduty notifiers), and therefore, does not increase maintainer burden, nor does it introduce new concepts to end users.

I note from your comment that this PR uses templated secrets for a slightly different purpose than Pagerduty. I admit I didn't realise this from the documentation. I believe the difference is slight enough that my previous point on having used a preexisting concept still stands.

brian-brazil:
> I believe this PR achieves this.

Why do you believe this? I don't see any way in which this helps you.

Tom-Fawcett:
With this PR I can do the following:

alertmanager.yml (extract)
```
templates: 
- 'templates/*.tmpl'

receivers:
- name: opsgenie
  opsgenie_configs:
  - api_key: '{{ template "opsgenie.xxx.api_key" . }}'
```

secrets.tmpl
```
{{ define "opsgenie.xxx.api_key" }}XXXXXX-XXXX-XXXX-XXXX-XXXXXXXXX{{ end}}
```

brian-brazil:
I see what you're doing now. Thinking on it, I don't think the practice you're proposing is a good idea security-wise.

Templates loaded from files are available to all templated fields, not just the ones that are meant to hold secrets. It is expected that large multi-team setups will have relatively non-trusted users that are meant to be writing just their team's configuration with no impact on any other team. If the secrets are sitting in templates loaded from files a malicious actor only needs to add all the template expansions to a message body and all secrets are then exposed to them. If the `tmpl` function is ever copied over from alert templating, there are additional potential attack vectors.

Secrets should be put directly in the file to avoid this. The secrets that are templated are intended to be used in a specific advanced use case where it's obvious that their secrecy is lost.

Tom-Fawcett:
I completely agree that this method does not allow me to have weaker write access controls on alertmanager.yml than on secrets.tmpl. As you said, there would be nothing stopping a person with write access on alertmanager.yml from extracting secrets out of secrets.tmpl. This would appear to be no worse than just having the secret straight in alertmanager.yml.

However, it does allow different levels of read access control. Pre-deploy, I can safely store alertmanager.yml in git, with few limits on read access, as long as secrets.tmpl is securely stored elsewhere. This would seem to be an advantage over having the secret straight in alertmanager.yml.

Possibly I have not fully understood you?

----
(Unrelated - congratulations on the Prometheus 2.0.0 release!)

brian-brazil:
> This would appear to be no worse than just having the secret straight in alertmanager.yml.

You're assuming that everyone with write access to some part of alertmanager.yml is also someone with access to the secrets. This is not a good assumption once you get into multi-team setups, where for example the config is assembled from multiple files where a team only has write to their file.

Tom-Fawcett:
So you could summarise as follows - 

This PR/functionality could be misused by users in multi-team setups, where they should instead be using config management to assemble the finished alertmanager.yml.

It could also be used correctly in single team setups, where it offers a simplified config process.

So is it reasonable to merge this PR when it could be misused? I believe it is, however, I appreciate such an answer is subjective.

brian-brazil:
> This PR/functionality could be misused by users in multi-team setups, where they should instead be using config management to assemble the finished alertmanager.yml.

Yes.

> It could also be used correctly in single team setups, where it offers a simplified config process.

The word to use is safely, not correctly. If a user is very careful in such a setup, and the setup never changes, and the company never expands, they can avoid a security breach.

> So is it reasonable to merge this PR when it could be misused? I believe it is, however, I appreciate such an answer is subjective.

The multi-team setup is what is also recommended for single teams.

It is presumed that you have configuration management when using Prometheus, and we don't add features merely because users lack that. Nor is it wise to add features that are fragile in security terms, merely because it makes some users doing things in a non-recommended manner easier.

Tom-Fawcett:
It was interesting discussing this with you. As we aren't really going anywhere, I'll close this.

PR #1082, opened 2017-11-06
[amtool] - Add a new `silence import` command

### LINKED ISSUE #1000: Provide silence backup and restore functionality in amtool cli tool
Since amtool is setup for querying and managing silences it would be great if you could perhaps dump the json output to a file and load it back into the api.

This would be useful when attempting to migrate existing silences to another instance or in the event of disaster recovery.

lebinh:
This command read silences data from a query JSON output and import to
alertmanager. It allows `amtool` to be used as a backup/restore tool for
silences, i.e. fix #1000

Backup / export:

```
amtool silence -o json > silences.json
```

Restore / import:

```
amtool silence import silences.json
```

This adds new silence one by one and will be slow if you have thousands of silences but alertmanager API doesn't allow multiple silences in POST so ¯\\_(ツ)_/¯

### DIFF
--- a/cli/silence_import.go
+++ b/cli/silence_import.go
@@ -9,6 +9,7 @@ import (
 	"github.com/prometheus/alertmanager/types"
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
+	"sync"
 )
 
 var importFlags *flag.FlagSet
@@ -35,8 +36,8 @@ func init() {
 	importFlags = importCmd.Flags()
 }
 
-func addSilenceWorker(silences <-chan *types.Silence, errs chan<- error) {
-	for s := range silences {
+func addSilenceWorker(silencec <-chan *types.Silence, errc chan<- error) {
+	for s := range silencec {
 		silenceId, err := addSilence(s)
 		sid := s.ID
 		if err != nil && err.Error() == "[bad_data] not found" {
@@ -50,12 +51,17 @@ func addSilenceWorker(silences <-chan *types.Silence, errs chan<- error) {
 		} else {
 			fmt.Println(silenceId)
 		}
-		errs <- err
+		errc <- err
 	}
 }
 
 func bulkImport(cmd *cobra.Command, args []string) error {
 	force, err := importFlags.GetBool("force")
+	if err != nil {
+		return err
+	}
+
+	workers, err := importFlags.GetInt("worker")
 	if err != nil {
 		return err
 	}
@@ -70,22 +76,31 @@ func bulkImport(cmd *cobra.Command, args []string) error {
 	}
 
 	dec := json.NewDecoder(input)
-
 	// read open square bracket
 	_, err = dec.Token()
 	if err != nil {
 		return errors.Wrap(err, "couldn't unmarshal input data, is it JSON?")
 	}
 
-	silences := make(chan *types.Silence, 100)
-	errs := make(chan error, 100)
-	workers, err := importFlags.GetInt("worker")
-	if err != nil {
-		return err
-	}
+	silencec := make(chan *types.Silence, 100)
+	errc := make(chan error, 100)
+	var wg sync.WaitGroup
 	for w := 0; w < workers; w++ {
-		go addSilenceWorker(silences, errs)
+		go func() {
+			wg.Add(1)
+			addSilenceWorker(silencec, errc)
+			wg.Done()
+		}()
 	}
+
+	errCount := 0
+	go func() {
+		for err := range errc {
+			if err != nil {
+				errCount++
+			}
+		}
+	}()
 
 	count := 0
 	for dec.More() {
@@ -100,28 +115,16 @@ func bulkImport(cmd *cobra.Command, args []string) error {
 			s.ID = ""
 		}
 
-		silences <- &s
+		silencec <- &s
 		count++
 	}
-	close(silences)
 
-	// read closing bracket
-	_, err = dec.Token()
-	if err != nil {
-		return errors.Wrap(err, "invalid JSON")
-	}
-
-	errCount := 0
-	for i := 0; i < count; i++ {
-		err = <-errs
-		if err != nil {
-			errCount++
-		}
-	}
+	close(silencec)
+	wg.Wait()
+	close(errc)
 
 	if errCount > 0 {
 		return fmt.Errorf("couldn't import %v out of %v silences", errCount, count)
 	}
-
 	return nil
 }
--- a/cli/silence_import.go
+++ b/cli/silence_import.go
@@ -1,0 +1,127 @@
+package cli
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
+	"github.com/pkg/errors"
+	"github.com/prometheus/alertmanager/types"
+	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
+)
+
+var importFlags *flag.FlagSet
+var importCmd = &cobra.Command{
+	Use:   "import [JSON file]",
+	Short: "Import silences",
+	Long: `Import alertmanager silences from JSON file or stdin
+
+  This command can be used to bulk import silences from a JSON file
+  created by query command. For example:
+
+  amtool silence query -o json foo > foo.json
+  amtool silence import foo.json
+
+  JSON data can also come from stdin if no param is specified.
+	`,
+	Args: cobra.MaximumNArgs(1),
+	Run:  CommandWrapper(bulkImport),
+}
+
+func init() {
+	importCmd.Flags().BoolP("force", "f", false, "Force adding new silences even if it already exists")
+	importCmd.Flags().IntP("worker", "w", 8, "Number of concurrent workers to use for import")
+	importFlags = importCmd.Flags()
+}
+
+func addSilenceWorker(silences <-chan *types.Silence, errs chan<- error) {
+	for s := range silences {
+		silenceId, err := addSilence(s)
+		sid := s.ID
+		if err != nil && err.Error() == "[bad_data] not found" {
+			// silence doesn't exists yet, retry to create as a new one
+			s.ID = ""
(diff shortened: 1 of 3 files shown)


cli/silence_import.go
```diff
@@ -0,0 +1,67 @@
...
+	var silences []types.Silence
+	err = json.Unmarshal(data, &silences)
+	if err != nil {
+		return errors.Wrap(err, "couldn't unmarshal input data, is it in JSON?")
+	}
+
+	for _, silence := range silences {
+		// reset the ID, otherwise alertmanager API will try to replace an existing silence
+		// this *might* create duplicated silences but that should generally be OK
+		silence.ID = ""
```
Kellel: I think we wan't to avoid duplicate silences if possible. Does the alertmanager API accept posting a new silence with the ID field present?

lebinh: It will accept if the silence ID exists and will replace it with the new one. But if the ID doesn't exist then it will return an error, not found. We can check and only remove ID after receive "not found" error but that will create a lot of extra round trip in the backup / restore use case (as every add will have to check and wait for a not found error first).

lebinh:
Hi @stuartnelson3, @Kellel can you guys help take a look and see if we can merge this PR?

stuartnelson3:
Hey sorry for the long delay, I've been away for the last few weeks and have been behind on reviewing things. I'll take a look at this in the next few days, thanks for your patience!

lebinh:
Sure @stuartnelson3 no problem, thank you for maintaining this.

cli/silence_import.go
```diff
@@ -0,0 +1,67 @@
...
+	Args: cobra.MaximumNArgs(1),
+	Run:  CommandWrapper(bulkImport),
+}
+
+func bulkImport(cmd *cobra.Command, args []string) error {
+	input := os.Stdin
+	var err error
+
+	if len(args) == 1 {
+		input, err = os.Open(args[0])
```
josedonizetti: You need to close the opened fd.

cli/silence_import.go
josedonizetti: Yeah, don't think we should have duplicated silences as well. Maybe skip the import if silence already exist?

stuartnelson3: Since this is for disaster recovery/data migration, I'm assuming maintaining the IDs wouldn't be important (and wouldn't expect there to be any pre-existing silences in the new instance of AM).

cli/silence_import.go
stuartnelson3: This should be added after the `err` is checked below

cli/silence_import.go
```diff
@@ -0,0 +1,67 @@
...
+
+		if err != nil {
+			return err
+		}
+	}
+
+	data, err := ioutil.ReadAll(input)
```
stuartnelson3: `ioutil.ReadAll` will load the entire input into memory. It would be better to process the information as a stream.

See https://golang.org/pkg/encoding/json/#example_Decoder_Decode_stream

stuartnelson3: Actually it ended up being a super easy change:

```go
dec := json.NewDecoder(input)

t, err := dec.Token()
if err != nil {
	return err
}

for dec.More() {
	var s types.Silence
	err := dec.Decode(&s)
	if err != nil {
		log.Fatal(err)
	}

	s.ID = ""

	err = addSilence(s)
	if err != nil {
		msg := fmt.Sprintf("couldn't add silence: %s", s)
		return errors.Wrap(err, msg)
	}
}

t, err = dec.Token()
if err != nil {
	return err
}

return nil
```

cli/silence_import.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return errors.Wrap(err, "couldn't unmarshal input data, is it in JSON?")
+	}
+
+	for _, silence := range silences {
+		// reset the ID, otherwise alertmanager API will try to replace an existing silence
+		// this *might* create duplicated silences but that should generally be OK
+		silence.ID = ""
+
+		err = addSilence(silence)
```
stuartnelson3: Since this is being done in serial, it will be relatively slow. We can always create a small worker pool (https://gobyexample.com/worker-pools) without much trouble. I'm fine adding it later, but if you want to add it now, feel free.

cli/silence_import.go
```diff
@@ -0,0 +1,67 @@
...
+  amtool silence import foo.json
+
+  JSON data can also come from stdin if no param is specified.
+	`,
+	Args: cobra.MaximumNArgs(1),
+	Run:  CommandWrapper(bulkImport),
+}
+
+func bulkImport(cmd *cobra.Command, args []string) error {
+	input := os.Stdin
```
stuartnelson3: also accepting a json stream on stdin is 👌

cli/silence_import.go
lebinh: I will add a flag to `force` this, it is need as we are running Alertmanager 0.6 and thus cannot update an existing silences. I will keep the silence ID by default as it appear Alertmanager will only update existing one without creating any new silence in that case.

lebinh:
Sorry for the close / re-open as I pushed the wrong commit 😞 
I added a new flag, `force` to remove the silence ID before calling add API. The default behavior, i.e. without `--force`, is to only remove that ID and re-try if we received "not found" error result.

There is also a new flag `--worker` to specify the number of concurrent workers calling Alertmanager API.

lebinh:
Probably worth mention that the`force` flag can also be used to improve speed of import in case of new silences:

```
$ time ./amtool silence import -w 4 silences.json
...
real	0m6.188s
user	0m0.247s
sys	0m0.087s

$ time ./amtool silence import -w 4 -f silences.json
...
real	0m3.549s
user	0m0.218s
sys	0m0.068s
```

The end results (with or without `-f`) are the same but the difference in speed is because we don't have to wait for a network roundtrip of "not found" error and then re-try.

stuartnelson3:
Awesome work! I'll give this a try tomorrow morning.

cli/silence_import.go
```diff
@@ -0,0 +1,127 @@
...
+	dec := json.NewDecoder(input)
+
+	// read open square bracket
+	_, err = dec.Token()
+	if err != nil {
+		return errors.Wrap(err, "couldn't unmarshal input data, is it JSON?")
+	}
+
+	silences := make(chan *types.Silence, 100)
+	errs := make(chan error, 100)
```
stuartnelson3: can you rename this to `errc`?

cli/silence_import.go
```diff
@@ -0,0 +1,127 @@
...
+
+	dec := json.NewDecoder(input)
+
+	// read open square bracket
+	_, err = dec.Token()
+	if err != nil {
+		return errors.Wrap(err, "couldn't unmarshal input data, is it JSON?")
+	}
+
+	silences := make(chan *types.Silence, 100)
```
stuartnelson3: go convention for naming channels is typically to name something and then put `c` after it (to indicate channel). could you change this to `silencec`?

cli/silence_import.go
```diff
@@ -0,0 +1,127 @@
...
+	errs := make(chan error, 100)
+	workers, err := importFlags.GetInt("worker")
+	if err != nil {
+		return err
+	}
+	for w := 0; w < workers; w++ {
+		go addSilenceWorker(silences, errs)
+	}
```
stuartnelson3: These three lines can be updated to help prevent a deadlock:

```go
var wg sync.WaitGroup
for w := 0; w < workers; w++ {
	go func(w int) {
		wg.Add(1)
		addSilenceWorker(silc, errc)
		wg.Done()
	}(w)
}

go func() {
	for err := range errc {
		if err != nil {
			errCount++
		}
	}
}()
```

Check out down below where we `wg.Wait()`, which indicates that all the silence workers are finished, and then we can `close(errc)`.

cli/silence_import.go
```diff
@@ -0,0 +1,127 @@
...
+
+		if force {
+			// reset the silence ID so Alertmanager will always create new silence
+			s.ID = ""
+		}
+
+		silences <- &s
+		count++
+	}
+	close(silences)
```
stuartnelson3: The final piece is here:

```go
close(silencec)
wg.Wait()
close(errc)
```

once we're done sending all the parsed silences, we can close the channel we're ranging over (anything still in the channel waiting to be processed will be processed). from there, we `wg.Wait()` to know all the silences have been created and their errors sent on the error channel, and then we can `close(errc)`.

cli/silence_import.go
```diff
@@ -0,0 +1,127 @@
...
+		return errors.Wrap(err, "invalid JSON")
+	}
+
+	errCount := 0
+	for i := 0; i < count; i++ {
+		err = <-errs
+		if err != nil {
+			errCount++
+		}
+	}
```
stuartnelson3: This code should be removed, it has been placed into the goroutine further up.

lebinh:
Thanks @stuartnelson3 these feedbacks are very helpful as I only started with Go recently. I've updated the PR as suggested.

cli/silence_import.go
```diff
@@ -100,28 +115,16 @@ func bulkImport(cmd *cobra.Command, args []string) error {
...
-		silences <- &s
+		silencec <- &s
 		count++
 	}
-	close(silences)
 
-	// read closing bracket
-	_, err = dec.Token()
-	if err != nil {
-		return errors.Wrap(err, "invalid JSON")
```
lebinh: Removed this as I think it doesn't really make sense to throw an error over JSON format here, the silences are added already at this point.

stuartnelson3: I would actually prefer to keep it in. Even though all the silences have been added, a user should still be informed if they are attempting to use invalid json.

stuartnelson3:
I'll take a look at this tomorrow to confirm it's working, thanks for the work on this

lebinh:
Great! Thanks @stuartnelson3

Issue #875, opened 2017-06-19
hipchat_url is not consistent with api_url


brian-brazil:
The global default for the Hipchat url and its corresponding per-receiver setting are inconsistenly named. One has `api`, the other doesn't. This should be cleaned up.

Given that users shouldn't have to touch the global, I suggest renaming it to `hipchat_api_url`.

josedonizetti:
@stuartnelson3 This can be closed.

Issue #680, opened 2017-04-01
`opsgenie_api_host` in the config should be `opsgenie_api_url`


juliusv:
It's a URL, not just a hostname. This is confusing if one doesn't look at the default value (which is a URL). I don't know what our current policy towards feature breakage is in AM, but maybe break it together with the UI revamp?

brancz:
Agreed. Currently breaking is ok. :+1: for grouping with UI revamp.

PR #1087, opened 2017-11-08
Fix config name inconsistency

josedonizetti:
This PR renames configuration from 3 different notifiers to make it consistent with others. Basically, global configurations will be named `<notifier>_api_url`, and local configurations per notifier `api_url`. 

This are **breaking changes**.

Fixes https://github.com/prometheus/alertmanager/issues/680 and https://github.com/prometheus/alertmanager/issues/875

@stuartnelson3 @brancz @brian-brazil @juliusv

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -202,10 +202,10 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		}
 		for _, hc := range rcv.HipchatConfigs {
 			if hc.APIURL == "" {
-				if c.Global.HipchatURL == "" {
+				if c.Global.HipchatAPIURL == "" {
 					return fmt.Errorf("no global Hipchat API URL set")
 				}
-				hc.APIURL = c.Global.HipchatURL
+				hc.APIURL = c.Global.HipchatAPIURL
 			}
 			if !strings.HasSuffix(hc.APIURL, "/") {
 				hc.APIURL += "/"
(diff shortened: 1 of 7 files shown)


brian-brazil:
We've generally followed the name that the provider uses in their docs when it comes to these things, as a user cares about only 1 of these which they already know - not that they're consistent across the alertmanager.

 Do these all match up with that?

brancz:
I wonder how much it's worth breaking these, as I'm pretty sure _a lot_ of configurations are going to break when people upgrade to these changes.

josedonizetti:
@brian-brazil I've changed the PR to rename only two configs:

- `hipchat_url` to `hipchat_api_url`, this fixes https://github.com/prometheus/alertmanager/issues/875 and checking the hipchat documentation the term 'API' does match more than just 'URL'

- `opsgenie_api_host` to `opsgenie_api_url`, this fixes https://github.com/prometheus/alertmanager/issues/680, the opsgenie documentation does not use the term 'host' but 'API' and 'URL'

josedonizetti:
@brancz Yup, agree that it will break. My idea when I first send this PR was to have it merged for the version **0.10.0**, was considering at the time that the first double number in the version would be a good moment to such a break.

But i totally understand if you think that is not worth, let's just add a note about it on the issues so no other PR is created for it for now, maybe in a future 1.0 release the break might make more sense.

brancz:
> in a future 1.0 release

I think this would be more appropriate. Similar breaking re-namings were done from Prometheus 0.x to 1.x.

brian-brazil:
I don't think that this is a breaking change in practice, as these settings should ~never need to be changed from the defaults by users.

aecolley:
You lose a karma point for this because my employer's private HipChat server means that this was a setting that most definitely needed to be different from the default.

stuartnelson3:
Thank you for bringing that to my attention. I was not aware hipchat could be run privately.

PR #1094, opened 2017-11-11
Support OpsGenie Priority field

Tom-Fawcett:
Now that Alertmanager is on the OpsGenie v2 API we can use the "Priority" field.

This is a non-mandatory field [described as follows](https://docs.opsgenie.com/docs/alert-api#section-create-alert):

> Priority level of the alert. Possible values are P1, P2, P3, P4 and P5. Default value is P3.

Further information about this field can be found [here](https://docs.opsgenie.com/docs/priority-field), including information on validation:

> If an unknown value is assigned or the tool does not carry a priority information, the alert will have the default priority value, P3.

With the above in mind, this PR does not validate the value for this field. Though I could add validation if so desired.


Tom-Fawcett:
Thanks :)

Issue #1060, opened 2017-10-27
UI: Negative filtering filters out empty/missing labels.


beorn7:
**What did you do?**

Filtered in the UI for alerts with unequality matcher: `foo!="bar"`

**What did you expect to see?**

I still want to see alerts that do not have a `foo` label at all, as "missing label" implies "label with empty label value", and `foo=""` clearly fulfills `foo!="bar"`.

**What did you see instead? Under which circumstances?**

An alert without `foo` label was not displayed with above filter.

**Environment**

* Alertmanager version:

0.9.1

stuartnelson3:
Ah, this wouldn't have been the expected behavior for me. When filtering by a label `foo!="bar"`, I would expect to be returned timeseries with a label foo, but that label not equalling bar.

Your expectation is consistent with prometheus, so it makes sense to fix this in alertmanager.

beorn7:
Indeed, we should definitely keep the semantics consistent throughout the Prometheus stack. Empty label and no label has to be semantically equivalent.

PR #1095, opened 2017-11-12
Fix silences negative filtering

josedonizetti:
We changed the alerting filtering logic to be consistent with prometheus https://github.com/prometheus/alertmanager/issues/1060. 

This PR makes the same change to the silence filtering, and extract the filtering logic which is equal no matter if an alert, or silence. This way is easier for a new change to be effective cross alertmanager, without having to change two different places.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -618,7 +618,7 @@ func (api *API) listSilences(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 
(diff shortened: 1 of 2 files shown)


stuartnelson3:
fix the merge conflicts and it's good to go

josedonizetti:
@stuartnelson3 Done. I've rebased the master commits. Thanks.

PR #1096, opened 2017-11-12
Add expired silence validation

josedonizetti:
At first, I wanted to validate if a silence is expired on https://github.com/prometheus/alertmanager/blob/master/silence/silence.go#L347, but expired silences are part of the internal logic https://github.com/prometheus/alertmanager/blob/master/silence/silence.go#L492, so I've just added a validation to the set silences API. I don't think it's useful for someone to create an expired silence from the UI/amtool.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -15,6 +15,7 @@ package api
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net/http"
 	"regexp"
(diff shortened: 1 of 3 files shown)


stuartnelson3:
While it's good to prevent people from accidentally creating a silence with start and end times being equal, they might also create a valid silence but with an ends at time in the past. probably worth checking that too.

josedonizetti:
@stuartnelson3 sorry for not following up sooner, had a family problem and spend couple days without time for anything. I'll get back to it today.

josedonizetti:
@stuartnelson3 Added the validation to end time in the past, and also rebased from master.

PR #1061, opened 2017-10-29
Upgrade OpsGenie notifier to v2 API

Tom-Fawcett:
Closes #902

The [OpsGenie Web API](https://docs.opsgenie.com/docs/web-api) (referred to as 'old' below) is deprecated.

This upgrades Alertmanager to the [OpsGenie REST Alert API](https://docs.opsgenie.com/docs/alert-api) (referred to as 'new' below).

----

## Key changes made
* Use new paths.
* Use new auth header.
* Adjust response code logic.
  * [Docs](https://docs.opsgenie.com/docs/response#section-response-codes)
  * Expand retry to include new "429 - Too Many Requests" throttling.
  * We no longer need to handle closing an already closed alert, as such a request is now asynchronous.

## Transition
* No changes in config.
* New integration still uses alert alias, so it is backwards compatible.
  * i.e. you can open a request on the old API and close it on the new API.

## Tests
* Open an alert with Alertmanager, then close with Alertmanager. ✓
* Open an alert with Alertmanager, close elsewhere, then observe Alertmanager attempting to close without erroring. ✓

## Asynchronous
Unlike the old API, the new API is asynchronous.

> Alert creation, deletion and action requests are processed asynchronously to provide higher availability and scalability, therefore valid requests for those endpoints are responded with HTTP status 202 - Accepted.

[link](https://docs.opsgenie.com/docs/alert-api)

* It seems reasonable to keep things simple, and not attempt to write a load of logic around requesting the status of previous requests.
* This works to our advantage when attempting to close an already closed alert, as we are returned 202.

PR #1101, opened 2017-11-13
Fix OpsGenie Teams field

Tom-Fawcett:
#1061 broke the OpsGenie Teams field.

The structure of this field has in fact changed between v1 and v2.

[OpsGenie Documentation](https://docs.opsgenie.com/docs/alert-api#section-create-alert)

## Tests
* No team config set. ✓
* One team value set. ✓
* Multiple team values set. ✓

----

I apologise for missing this in #1061.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -784,19 +784,15 @@ func NewOpsGenie(c *config.OpsGenieConfig, t *template.Template, l log.Logger) *
 }
 
 type opsGenieCreateMessage struct {
-	Alias       string            `json:"alias"`
-	Message     string            `json:"message"`
-	Description string            `json:"description,omitempty"`
-	Details     map[string]string `json:"details"`
-	Source      string            `json:"source"`
-	Teams       []opsGenieTeam    `json:"teams,omitempty"`
-	Tags        string            `json:"tags,omitempty"`
-	Note        string            `json:"note,omitempty"`
-	Priority    string            `json:"priority,omitempty"`
-}
-
-type opsGenieTeam struct {
-	Name string `json:"name,omitempty"`
+	Alias       string              `json:"alias"`
+	Message     string              `json:"message"`
+	Description string              `json:"description,omitempty"`
+	Details     map[string]string   `json:"details"`
+	Source      string              `json:"source"`
+	Teams       []map[string]string `json:"teams,omitempty"`
+	Tags        string              `json:"tags,omitempty"`
+	Note        string              `json:"note,omitempty"`
+	Priority    string              `json:"priority,omitempty"`
 }
 
 type opsGenieCloseMessage struct {
@@ -839,9 +835,9 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 		}
 
 		apiURL = n.conf.APIURL + "v2/alerts"
-		var teams []opsGenieTeam
+		var teams []map[string]string
 		for _, t := range strings.Split(string(tmpl(n.conf.Teams)), ",") {
-			teams = append(teams, opsGenieTeam{Name: t})
+			teams = append(teams, map[string]string{"name": t})
 		}
 		msg = &opsGenieCreateMessage{
 			Alias:       alias,
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -789,10 +789,14 @@ type opsGenieCreateMessage struct {
 	Description string            `json:"description,omitempty"`
 	Details     map[string]string `json:"details"`
 	Source      string            `json:"source"`
-	Teams       string            `json:"teams,omitempty"`
+	Teams       []opsGenieTeam    `json:"teams,omitempty"`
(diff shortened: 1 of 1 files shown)


notify/impl.go
```diff
@@ -835,13 +839,17 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 		}
 
 		apiURL = n.conf.APIURL + "v2/alerts"
+		var teams []opsGenieTeam
+		for _, t := range strings.Split(string(tmpl(n.conf.Teams)), ",") {
```
brian-brazil: This mix of templating, splitting and a list of single-field structs seems odd. Why not keep the config as-is, and split?

Tom-Fawcett: Thank you for the review.

I am probably going to show myself up here - but this was the only way I could think of to produce the desired JSON array of objects. ie:

```
"teams":[{"name":"foo_team"}, {"name":"bar_team"}], 
```

josedonizetti: @Tom-Fawcett Think that instead of having the `teams` config be a string https://github.com/prometheus/alertmanager/blob/master/config/notifiers.go#L308 on the alertmanager yaml, it should be a list of key values, this way we can also support the team's id, and not only the name.

eg:
```
teams: 
   - id: 'team_id1'
   - id: 'team_id2'
```
or 
```
teams:
  - name: 'team_name1'
  - name: 'team_name2'
```

@brian-brazil what do you think?

brian-brazil: The problem with that is that you have to know the number of teams in advance, as we can only template within a field - not across a list.

stuartnelson3: Is there a reason we don't want to do this at the config level? It would break configs for users upgrading, so we would have to do another release.

```yaml
# A set of arbitrary key/value pairs that provide further detail
# about the incident.
[ teams: { <id|name>: <tmpl_string>, ... } ]
```

The docs support either a `name` or `id` as the key; do we want to make the assumption for users that they're providing team names?

stuartnelson3: Race condition on writing apparently

josedonizetti: @brian-brazil is it diff than what is done here? https://github.com/prometheus/alertmanager/blob/master/notify/impl.go#L815

stuartnelson3: We can drop the `opsGenieTeam` struct and just append a `map[string]string{"name": t}`

josedonizetti: maybe to avoid a new release, and also fix the problem we can for now do

1) what @stuartnelson3 said `We can drop the opsGenieTeam struct and just append a map[string]string{"name": t}`
2) only support one team name for now, I don't like much the 'strings.Split' by commas

Then we discuss a better approach for a next release.

Tom-Fawcett: Thank you for the feedback.

The use of Split was to try and preserve functionality without changing config.

I can push the suggestion of using `map[string]string{"name": t}` in place of `opsGenieTeam`.

Could I please get some further guidance on the race condition.

And a decision on what I should do about config, if I am removing Split.

stuartnelson3: Sorry my comment wasn't clear, there is no race condition in the code you've written. @josedonizetti and I suggested the same thing within minutes of eachother, that's what I was referring to.

Only supporting a single team name would also be breaking the interface we present to users, as we explicitly state that it's a comma separated list of teams.

```go
t := strings.Split(n.conf.Teams, ",")
teams := make([]map[string]string, 0, len(t))
for _, name := range t {
	teams = append(teams, map[string]string{"name": tmpl(name)})
}
```

I would be fine with the above for now.

josedonizetti: @Tom-Fawcett mind tackling the `tags` problem as well?

Tom-Fawcett: @stuartnelson3 ah, yes I missed that joke!

The main difference between your example snippet and this PR (after my latest commit), is you split and then use tmpl. That doesn't seem like it would work?

@josedonizetti yep 👍

notify/impl.go
```diff
@@ -784,15 +784,15 @@ func NewOpsGenie(c *config.OpsGenieConfig, t *template.Template, l log.Logger) *
...
-	Tags        string            `json:"tags,omitempty"`
-	Note        string            `json:"note,omitempty"`
-	Priority    string            `json:"priority,omitempty"`
+	Alias       string              `json:"alias"`
+	Message     string              `json:"message"`
+	Description string              `json:"description,omitempty"`
+	Details     map[string]string   `json:"details"`
+	Source      string              `json:"source"`
+	Teams       []map[string]string `json:"teams,omitempty"`
+	Tags        string              `json:"tags,omitempty"`
```
stuartnelson3: Can you update this to `[]string`, as specified in the API docs, and split `n.conf.Tags` on `,` and pass that to the create message?

stuartnelson3: disregard, #1108 covers it

PR #1103, opened 2017-11-14
Fix typo

josedonizetti:


brancz:
Why didn't this fail before?

josedonizetti:
@brancz The test is correct, this is only the description shown whenever a test fails!

brancz:
I see now. Sorry about that, lgtm

PR #1104, opened 2017-11-14
Add tests to memory provider

josedonizetti:
@stuartnelson3 Take a special look into `TestAlertsSubscribe` as it involves concurrency. Think I was able to validate the behavior well without writing a flap test.

Thanks.


josedonizetti:
@stuartnelson3 ping. Any feedback here?

PR #1108, opened 2017-11-15
Fix OpsGenie Tags field

Tom-Fawcett:
#1061 broke the OpsGenie Tags field.

The structure of this field has in fact changed between v1 and v2.

[OpsGenie Documentation](https://docs.opsgenie.com/docs/alert-api#section-create-alert)

I have used strings.Split to preserve functionality without changing config.

## Tests
* No tag config set. ✓
* One tag value set. ✓
* Multiple tags set. ✓

----

I apologise for missing this in #1061.


Tom-Fawcett:
@stuartnelson3 would you like me to resolve the conflict and push again?

stuartnelson3:
that would be great

Tom-Fawcett:
Done

Issue #1116, opened 2017-11-22
PagerDuty v2 Source not execute template


wutaizeng:
https://github.com/prometheus/alertmanager/blob/55a40d1d3af5d1f8f52b0251d4f78a60a152d495/notify/impl.go#L504

stuartnelson3:
Closed by #1117

PR #1117, opened 2017-11-22
Send the prom gen url as alert source

stuartnelson3:
From the documentation at
https://v2.developer.pagerduty.com/docs/send-an-event-events-api-v2
```
payload.source: required string

The unique location of the affected system, preferably a hostname or FQDN.
```
Sending the generatorURL isn't technically the
affected system, but an event sent to PD can
encompass multiple alerts. The generatorURL seems
like the correct way to help users debug this.

addresses #1116

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -501,14 +501,11 @@ func (n *PagerDuty) notifyV2(ctx context.Context, eventType, key string, tmpl fu
 	if eventType == pagerDutyEventTrigger {
 		payload = &pagerDutyPayload{
 			Summary:       tmpl(n.conf.Description),
+			Source:        tmpl(n.conf.Client),
 			Severity:      n.conf.Severity,
 			CustomDetails: details,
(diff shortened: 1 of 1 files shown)


stuartnelson3:
looking for opinions: @brancz @grobie @brian-brazil @beorn7

brian-brazil:
I believe our current behaviour is correct.

The generator URL isn't right here, as a notification can come from multiple Prometheus servers (and technically generator URLs are per-alert). Randomly picking one alert doesn't seem right to me.

I think the sane default is a link to the alertmanager, and allowing templating if a user knows something about their system that lets them select a better option. Which is what we already have.

stuartnelson3:
`Source` wasn't being templated, so that change is now made in this pr.

brian-brazil:
:+1: 

Watch the commit message.

stuartnelson3:
Updated the commit message to make sense, will merge once tests are green

PR #1120, opened 2017-11-24
Added ability to query for silences that will expire soon

sinkingpoint:
As part of our workflow we find it quite useful to be able to query alerts that will expire soon, so that we can investigate whether they should stay silenced or be extended before we get a flood of alerts. 

This PR adds such functionality to amtool so that a user can do something like:

```
amtool silence query --within 8h
```

which will give all the silences that expire within the next 8h.

### DIFF
--- a/cli/silence_query.go
+++ b/cli/silence_query.go
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/prometheus/alertmanager/cli/format"
+	"github.com/prometheus/common/model"
 	"github.com/prometheus/alertmanager/pkg/parse"
 	"github.com/prometheus/alertmanager/types"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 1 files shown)


stuartnelson3:
sorry for the delay, I've been on vacation and will be getting back to work soon. I'll review this soon!

cli/silence_query.go
```diff
@@ -117,6 +123,11 @@ func query(cmd *cobra.Command, args []string) error {
 		if !expired && silence.EndsAt.Before(time.Now()) {
 			continue
 		}
+
+		if int64(within) > 0 && silence.EndsAt.After(time.Now().Add(within)) {
+			continue
```
josedonizetti: EndsAt is UTC on the AM, and calling time.Now() on amtool will use the user time zone, better to use UTC as well `time.Now().UTC().Add ...`

cli/silence_query.go
```diff
@@ -89,6 +90,11 @@ func query(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	within, err := queryFlags.GetDuration("within")
```
stuartnelson3: For the sake of consistency, would it be possible to get the flag value as a string, and then parse it with the prometheus duration parser?

https://github.com/prometheus/alertmanager/commit/ea9a584e

It seems unlikely folks are interested in filtering down to silences that are expiring in weeks or even years, but we should offer the same duration parsing functionality.

sinkingpoint:
@stuartnelson3 @josedonizetti I've addressed both of your comments

stuartnelson3:
sorry for the delay, I'll review this tomorrow. Thanks for your patience!

stuartnelson3:
Looks good! Can you add a note in the command's long description for the new flag? Once that's in, it's all good.

sinkingpoint:
Done @stuartnelson3 , just waiting on CI. I've also rebased for cleanliness.

PR #1121, opened 2017-11-24
nflog: Copy and replace gossipData instead of modifying it in place

### LINKED ISSUE #1037: race in gossip
After some further testing, the go race detector has found possibly multiple races in the gossip protocol.

As the race detector was very elaborate with its output on this find, I put it in a gist, rather than including it here in the issue.

https://gist.github.com/brancz/db2b2a8a3a60617c07ceee84e5e0b445

@fabxc @stuartnelson3

brancz:
It's very hard to prove, but I think this may fix #1037. Even if not, this makes it a lot easier to reason about the modification of `gossipData`, as it is not modified in place, but rather copied and then modified, and then the entire variable is replaced. That way we can be sure that nothing within the `Merge` and `mergeDelta` functions modify the receiver itself.

What I do not know is how this may impact performance of the system. Fundamentally we don't know what Alertmanager can handle today, so it's hard to tell how much worse this is.

I have tried to reproduce the race originally reported in #1037, but haven't been able to. It's questionable what this means, as it was not easy to reproduce it even without this patch.

@stuartnelson3 @grobie @fabxc

### DIFF
--- a/nflog/nflog.go
+++ b/nflog/nflog.go
@@ -491,7 +491,8 @@ func (l *nlog) OnGossip(msg []byte) (mesh.GossipData, error) {
 	l.mtx.Lock()
 	defer l.mtx.Unlock()
 
-	if delta := l.st.mergeDelta(gd); len(delta) > 0 {
+	var delta gossipData
+	if l.st, delta = l.st.mergeDelta(gd); len(delta) > 0 {
 		return delta, nil
 	}
 	return nil, nil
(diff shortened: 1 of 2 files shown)


stuartnelson3:
I'm in favor of making the system easier to reason about, and then worrying about performance implications later. At some point we'll have to benchmark the system, and also figure out what "acceptable" performance is for an alertmanager instance.

The performance penalty probably grows exponentially as users add more mesh members, but we'll see when we actually measure.

brancz:
Agreed

PR #1123, opened 2017-11-27
[amtool] - Add new command to update silence

lebinh:
This adds a new command, `update` (and also its alias, `extend`), to update
existing silence in Alertmanager. User can use this command to update the
expiration or comment on existing silences. The API already support this
so I only expose the same functionality to amtool.

Don't allow update CreatedBy field as it is "Created" not "Updated", so
we should keep the original author.

Sample usage:

```
$ ./amtool silence add -e 2h -c test foo=bar
2c9ff0c4-9133-4c82-b4a1-01806c3a6673

$ ./amtool silence query foo=bar
ID                                    Matchers  Ends At                  Created By  Comment
2c9ff0c4-9133-4c82-b4a1-01806c3a6673  foo=bar   2017-11-27 08:49:04 UTC  @binh       test

$ ./amtool silence extend -e 4h -c 'updated test' 2c9ff0c4-9133-4c82-b4a1-01806c3a6673
ID                                    Matchers  Ends At                  Created By  Comment
2c9ff0c4-9133-4c82-b4a1-01806c3a6673  foo=bar   2017-11-27 10:49:28 UTC  @binh       updated test

$ ./amtool silence query foo=bar
ID                                    Matchers  Ends At                  Created By  Comment
2c9ff0c4-9133-4c82-b4a1-01806c3a6673  foo=bar   2017-11-27 10:49:28 UTC  @binh       updated test
```

### DIFF
--- a/cli/silence_update.go
+++ b/cli/silence_update.go
@@ -1,0 +1,147 @@
+package cli
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"path"
+	"time"
+
+	"github.com/prometheus/alertmanager/cli/format"
+	"github.com/prometheus/alertmanager/types"
+	"github.com/spf13/cobra"
+	flag "github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+type getResponse struct {
+	Status    string        `json:"status"`
+	Data      types.Silence `json:"data,omitempty"`
+	ErrorType string        `json:"errorType,omitempty"`
+	Error     string        `json:"error,omitempty"`
+}
+
+var updateFlags *flag.FlagSet
+var updateCmd = &cobra.Command{
+	Use:     "update <id> ...",
+	Aliases: []string{"extend"},
+	Args:    cobra.MinimumNArgs(1),
+	Short:   "Update silences",
+	Long:    `Extend or update existing silence in Alertmanager.`,
+	Run:     CommandWrapper(update),
+}
+
+func init() {
+	updateCmd.Flags().StringP("expires", "e", "", "Duration of silence (100h)")
+	updateCmd.Flags().String("expire-on", "", "Expire at a certain time (Overwrites expires) RFC3339 format 2006-01-02T15:04:05Z07:00")
+	updateCmd.Flags().StringP("comment", "c", "", "A comment to help describe the silence")
+	updateFlags = updateCmd.Flags()
+}
+
+func update(cmd *cobra.Command, args []string) error {
+	if len(args) < 1 {
+		return fmt.Errorf("no silence IDs specified")
+	}
+
+	alertmanagerUrl, err := GetAlertmanagerURL()
+	if err != nil {
+		return err
+	}
+
+	var updatedSilences []types.Silence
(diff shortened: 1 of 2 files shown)


josedonizetti:
@lebinh I know most of amtool code isn't tested. But I really would like to change that. Especially on new code. Do you mind writing tests here?

lebinh:
Okay, I'll try to add some test soon

stuartnelson3:
I'll take a look at this tomorrow. Can you resolve the conflicts?

lebinh:
Resolved, although I haven't got around to add any test yet.

cli/silence_update.go
```diff
@@ -0,0 +1,147 @@
...
+	Error     string        `json:"error,omitempty"`
+}
+
+var updateFlags *flag.FlagSet
+var updateCmd = &cobra.Command{
+	Use:     "update <id> ...",
+	Aliases: []string{"extend"},
+	Args:    cobra.MinimumNArgs(1),
+	Short:   "Update silences",
+	Long:    `Extend or update existing silence in Alertmanager.`,
```
stuartnelson3: Could you update this with a longer description, with a usage example? It would also be good to note that `expires-on` will override `expires`.

cli/silence_update.go
```diff
@@ -0,0 +1,147 @@
...
+	Run:     CommandWrapper(update),
+}
+
+func init() {
+	updateCmd.Flags().StringP("expires", "e", "", "Duration of silence (100h)")
```
stuartnelson3: I'm thinking we should rename these. Maybe `expires` should become `duration`, with `d` as the short flag?

cli/silence_update.go
```diff
@@ -0,0 +1,147 @@
...
+	updateCmd.Flags().String("expire-on", "", "Expire at a certain time (Overwrites expires) RFC3339 format 2006-01-02T15:04:05Z07:00")
```
stuartnelson3: In alertmanager, you can also update the starting time when editing a silence.

How about adding that as well, and naming the flags `date.start` and `date.end`?

stuartnelson3: hm, thinking more about the name .. not sure i like `date`. @grobie , naming guru, any recommendation?

cli/silence_update.go
lebinh: I was just following the parameter names in `silence add` command here. `duration` does make more sense than `expires` though, but if we are going to change we need to update `add` command as well. That will be a breaking change though.

cli/silence_update.go
lebinh: Should we just follow the field names in Alertmanager API, `startsAt` and `endsAt`?

cli/silence_update.go
lebinh: I'dd add that in once we decided on the parameter names :)

cli/silence_update.go
stuartnelson3: hrm, those are more json-idiomatic names. for a CLI I think something different would be more appropriate.

cli/silence_update.go
stuartnelson3: good point

PR #1129, opened 2017-12-05
Adding check for webhook's URL formatting

xginn8:
Since alertmanager will fail silently when trying to send to a schemaless URL, provide a way to check that a URL is properly formatted in alertmanager

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -15,6 +15,7 @@ package config
 
 import (
 	"fmt"
+	"net/url"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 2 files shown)


config/notifiers.go
```diff
@@ -292,6 +293,14 @@ func (c *WebhookConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if c.URL == "" {
 		return fmt.Errorf("missing URL in webhook config")
 	}
+	url, err := url.Parse(c.URL)
+	if err != nil {
+		return err
+	}
+	if url.Scheme != "https" && url.Scheme != "http" {
+		return fmt.Errorf("non-absolute URL in webhook config")
```
stuartnelson3: Can you change this to `scheme required for webhook url`?

Issue #731, opened 2017-04-24
[Request] Add Slack fields to notifications


taharah:
The Slack notification API has added a fields parameter that allows for some additional formatting [1]. It looks like the work was started to add this additional parameter [2], but the feature was never finished so that it was available. 

[1] https://api.slack.com/docs/message-attachments#fields
[2] https://github.com/prometheus/alertmanager/blob/v0.5.1/notify/impl.go#L485-L490

mxinden:
Hi @taharah do you have some time creating a PR for this? We are here to help in case you have any questions.

chhetripradeep:
@mxinden @taharah  I would like to pick this up. Will create a PR soon with my basic Golang knowledge. Thanks.

PR #1135, opened 2017-12-10
Add Slack additional "fields" to notifications 

rbtr:
Had a go at implementing https://github.com/prometheus/alertmanager/issues/731

> The Slack notification API has added a fields parameter that allows for some additional formatting [1]. It looks like the work was started to add this additional parameter [2], but the feature was never finished so that it was available.
>[1] https://api.slack.com/docs/message-attachments#fields
>[2] https://github.com/prometheus/alertmanager/blob/v0.5.1/notify/impl.go#L485-L490

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -626,12 +626,13 @@ type slackReq struct {
 
 // slackAttachment is used to display a richly-formatted message block.
 type slackAttachment struct {
-	Title     string `json:"title,omitempty"`
-	TitleLink string `json:"title_link,omitempty"`
-	Pretext   string `json:"pretext,omitempty"`
-	Text      string `json:"text"`
-	Fallback  string `json:"fallback"`
-	Footer    string `json:"footer"`
+	Title     string                 `json:"title,omitempty"`
+	TitleLink string                 `json:"title_link,omitempty"`
+	Pretext   string                 `json:"pretext,omitempty"`
+	Text      string                 `json:"text"`
+	Fallback  string                 `json:"fallback"`
+	Fields    []slackAttachmentField `json:"fields"`
+	Footer    string                 `json:"footer"`
 
 	Color    string   `json:"color,omitempty"`
 	MrkdwnIn []string `json:"mrkdwn_in,omitempty"`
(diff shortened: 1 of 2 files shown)


notify/impl.go
```diff
@@ -660,6 +661,20 @@ func (n *Slack) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 		Color:     tmplText(n.conf.Color),
 		MrkdwnIn:  []string{"fallback", "pretext", "text"},
 	}
+
+	var numFields = len(n.conf.Fields)
+	if numFields > 0 {
+		var fields = make([]slackAttachmentField, numFields)
+		for k, v := range n.conf.Fields {
+			fields[k] = slackAttachmentField{
+				v["title"],
```
stuartnelson3: let's add `tmplText()` around both `title` and `value`, so the users can optionally define a template

notify/impl.go
```diff
@@ -662,6 +663,20 @@ func (n *Slack) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 		Color:     tmplText(n.conf.Color),
 		MrkdwnIn:  []string{"fallback", "pretext", "text"},
 	}
+
+	var numFields = len(n.conf.Fields)
+	if numFields > 0 {
+		var fields = make([]slackAttachmentField, numFields)
+		for k, v := range n.conf.Fields {
+			fields[k] = slackAttachmentField{
+				v[tmplText("title")],
```
rbtr: @stuartnelson3 is this what you meant?

notify/impl.go
```diff
@@ -669,8 +669,8 @@ func (n *Slack) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 		var fields = make([]slackAttachmentField, numFields)
 		for k, v := range n.conf.Fields {
 			fields[k] = slackAttachmentField{
-				v["title"],
-				v["value"],
+				tmplText(v["title"]),
```
rbtr: @stuartnelson3 is this what you meant?

stuartnelson3: yeah, exactly

rbtr:
Closes #731

JeanMertz:
This capability isn't documented anywhere on the Prometheus website, despite the latest release being from 13 days ago.

see: https://prometheus.io/docs/alerting/configuration/#%3Cslack_config%3E

Am I looking in the wrong place, or should the documentation be updated?

rbtr:
@JeanMertz i think the doc just needs to be updated. i'll see if i can contrib that

stuartnelson3:
@JeanMertz @rbtr yeah the docs need to be updated

PR #1140, opened 2017-12-11
make tls option InsecureSkipVerify configurable by alertmanager

victoryang:
When we try to send mails to notify clients, and if it requires TLS, the TLS option InsecureSkipVerify should be configurable with alertmanager's configuration. This PR is to add InsecureSkipVerify into alertmanager's global configuration.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -191,6 +191,10 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 				ec.RequireTLS = new(bool)
 				*ec.RequireTLS = c.Global.SMTPRequireTLS
 			}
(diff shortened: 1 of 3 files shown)


stuartnelson3:
PLEASE don't close and open a new pull request because CI fails. You can restart CI. This creates a ton of noise and is very annoying.

victoryang:
Sorry for that...

brancz:
I was very much hoping we could get the entire HTTP config a la [this](https://github.com/prometheus/alertmanager/pull/930), to make sure everything we know being a valid use case can be enabled.

But we never go to agreement on how to continue. /cc @fabxc @brian-brazil

simonpasquier:
Closed by #1528

PR #1151, opened 2017-12-19
Make alertGC interval configurable

stuartnelson3:
@brancz and I were discussing this earlier, and it seemed to make sense to make this configurable.

I'm happy to change the flag name, so any suggestions are welcome.


brancz:
lgtm :+1:

Issue #625, opened 2017-02-14
Improve debuggability of HA and mesh


brian-brazil:
If HA isn't working correctly there's very little to help you debug it.

It would be useful to have metrics showing the amount of data the mesh is handling, how many notifications were inhibited by it, how big the mesh is etc.

Log wise, info logs when the mesh is joined and its members change would be useful. Debug logs on how the HA features affect notifications would help.

Information on the status page relating to the mesh would be helpful.

fabxc:
Did you hit a debugging case or is this a general issue?

brian-brazil:
I'm currently helping a user debug an issue and everything looks to be setup correctly thus far, however there are duplicate notifications.

Rudd-O:
I have HA problems too.  Silences set on one of the nodes propagate to one other node, but not to the third.  This is causing alerts to be sent when we don't want them sent.  Nothing in the debug logs to fix or even to track down the issue either.

stuartnelson3:
@Rudd-O can you write an email to the prometheus-users google group with the following:

- Describing your situation
- Your command line invocation for the 3 nodes
- Basic machine information

It might just be a configuration mistake, and if it's an actual bug then we can write it up as an separate issue.

Rudd-O:
On 03/30/2017 10:53 AM, stuart nelson wrote:
>
> @Rudd-O <https://github.com/Rudd-O> can you write an email to the
> prometheus-users google group with the following:
>
>   * Describing your situation
>   * Your command line invocation for the 3 nodes
>   * Basic machine information
>

I'm using the latest Docker container release of alertmanager.  I will
give you the command lines tomorrow as I am not at work right now. 
Basically, if I set a silence in one of my alertmanagers, it propagates
to one other, but the third doesn't get it.  Which one propagates to
which other one is really random.  I am confident this isn't a
configuration mistake.

I perused the source code and I do not see any logging statements that
could signal what the health of the mesh really is, not even in debug mode.

-- 
    Rudd-O
    http://rudd-o.com/

stuartnelson3:
>Can you write an email to the prometheus-users google group...

aecolley:
At the risk of metooing, I also have a great need for mesh metrics. I have three long-running alertmanagers in a mesh, and today I found that one of them had lost its connections to the other two. Nothing appears in the log and there are no metrics which I could use to drive an alert for the condition. I found the problem only because an alert notification was sent which should have been silenced.

[...]

PR #1156, opened 2017-12-20
Log mesh router status on SIGUSR1 signal

simonpasquier:
This is an experiment aiming at improving the troubleshooting of the mesh router. If it isn't an acceptable approach for the project, I'll be fine with it :-)
Examples of logs

```
level=info ts=2017-12-20T17:22:14.315435721Z caller=main.go:414 msg="Mesh router status"
level=info ts=2017-12-20T17:22:14.321914642Z caller=main.go:416 msg="Local peer status" name=00:00:00:00:00:02 connections_nb=1 peers_nb=1 termination_count=0
level=info ts=2017-12-20T17:22:14.322267377Z caller=main.go:418 msg="Connection status" address=127.0.0.1:8001 outbound=true state=failed info="dial tcp4 :0->127.0.0.1:8001: getsockopt: connection refused, retry: 2017-12-20 18:22:16.265975795 +0100 CET m=+10.003780206"
[...]
level=info ts=2017-12-20T17:23:26.724653333Z caller=main.go:414 msg="Mesh router status"
level=info ts=2017-12-20T17:23:26.724983428Z caller=main.go:416 msg="Local peer status" name=00:00:00:00:00:02 connections_nb=1 peers_nb=2 termination_count=0
level=info ts=2017-12-20T17:23:26.72512933Z caller=main.go:418 msg="Connection status" address=127.0.0.1:8001 outbound=true state=pending info="none   00:00:00:00:00:01(a)"
```

It somehow relates to #625.

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -372,9 +372,11 @@ func main() {
 		hup      = make(chan os.Signal)
 		hupReady = make(chan bool)
 		term     = make(chan os.Signal)
+		usr1     = make(chan os.Signal)
 	)
 	signal.Notify(hup, syscall.SIGHUP)
 	signal.Notify(term, os.Interrupt, syscall.SIGTERM)
+	signal.Notify(usr1, syscall.SIGUSR1)
 
 	go func() {
 		<-hupReady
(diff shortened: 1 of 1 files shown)


brancz:
Why SIGUSR1? Since this is only reading I don't see a reason not to put this in the existing status output (maybe it even is already?).

As a side note, we've been looking into replacing weaveworks/mesh as it is not actively being maintained/developed and there are a number of problems with it. This investigation is in its very first steps however, so for the time being it is probably a good idea to have more insight into the existing mesh.

simonpasquier:
@brancz only because I've been using SIGUSR1 for the same purpose with other tools. Personally I find it handy to have the information in the logs rather than open the UI. It can also be useful when helping users since you can ask them to signal their processes and provide the logs. Agree that the information should also be added to the Status page (I'll take care of it in another PR).

Finally thanks for the news on the future of the mesh part. Looking forward to it!

simonpasquier:
I've created #1164 which adds more information to the Status page. Closing this one for now.

PR #1158, opened 2017-12-21
Lock around variables used in Update()

stuartnelson3:
Found two places where struct members being
updated in api.Update() where being accessed
elsewhere without locks.

testing this at the moment.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -331,6 +331,7 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 	alerts := api.alerts.GetPending()
 	defer alerts.Close()
 
+	api.mtx.RLock()
 	// TODO(fabxc): enforce a sensible timeout.
 	for a := range alerts.Next() {
 		if err = alerts.Err(); err != nil {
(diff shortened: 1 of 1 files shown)


api/api.go
```diff
@@ -331,6 +331,7 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 	alerts := api.alerts.GetPending()
 	defer alerts.Close()
 
+	api.mtx.RLock()
```
brancz: What are we locking against here?

stuartnelson3: there's a `routes := api.route.Match(a.Labels)` a few lines below, in a loop. This is one of the fields that gets updated in `Update()`

brancz: Got it. I missed that.

brancz:
lgtm

api/api.go
stuartnelson3: no worries, github didn't bother showing it 🤷‍♂️

brancz: I actually expanded but missed the `api.route` part. So it's just me needing coffee.

Issue #991, opened 2017-09-13
Provide a HTTP Health / Readiness Check


iksaif:
Similar to https://github.com/prometheus/pushgateway/issues/105

I might work on it soon, but feel free to take it.

Everybody agrees with  `/-/healthy` and `/-/ready` ?

stuartnelson3:
The nature of the mesh is other am's can join and leave, and there's no real manifest to know how many are supposed to be connected, and similarly, I'm not sure how we can know we've received most of the data, since an alertmanager instance doesn't know how much there is.

How would you define most of the data?

iksaif:
I guess we could wait at least `gossip_settle_min_wait_ms`  (1000) up to `gossip_settle_max_wait_ms` (10000) as long as there is active traffic on the mesh (for silences and peers).

Having simply an option to wait "some" time would be a good start.

grobie:
@iksaif Have you had a chance to look into this?

iksaif:
Nop, I implemented it for push_gateway though, so this could be used as an example

simonpasquier:
I've submitted #1159  for adding `/-/healthy`. Any other idea regarding readiness apart from a fixed wait time after (re)start?

simonpasquier:
This issue can be closed thanks to #1209.

cc @stuartnelson3

PR #1159, opened 2017-12-21
Add /-/healthy endpoint

simonpasquier:
This partially addresses #991. Regarding the readiness probe, I'm not sure that we can reliably evaluate whether the mesh is ready or not. Thoughts?

### DIFF
--- a/ui/web.go
+++ b/ui/web.go
@@ -15,6 +15,7 @@ package ui
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"net/http"
 	_ "net/http/pprof" // Comment this line to disable pprof endpoint.
@@ -74,6 +75,11 @@ func Register(r *route.Router, reloadCh chan<- struct{}, logger log.Logger) {
 	r.Post("/-/reload", func(w http.ResponseWriter, req *http.Request) {
 		w.Write([]byte("Reloading configuration file..."))
 		reloadCh <- struct{}{}
+	})
+
+	r.Get("/-/healthy", func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		fmt.Fprintf(w, "OK")
 	})
 
 	r.Get("/debug/*subpath", http.DefaultServeMux.ServeHTTP)


ui/web.go
```diff
@@ -76,6 +77,11 @@ func Register(r *route.Router, reloadCh chan<- struct{}, logger log.Logger) {
 		reloadCh <- struct{}{}
 	})
 
+	r.Get("/-/healthy", func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
```
iksaif: could /healthy check that we can:
* read data from memory
* read data from disk

brancz:
> I'm not sure that we can reliably evaluate whether the mesh is ready or not. Thoughts?

If mesh is enabled, then I would expect readiness to express that at least a successful connection to at least one peer has been established, if possible even that an initial sync of data has been performed.

ui/web.go
brancz: Healthiness and readiness is often hard to distinguish, but I in my opinion you nailed it: anything regarding it's local operability defines it's healthiness. Readiness describes whether requests can be successfully served, meaning depending services can be reached, or in the mesh case, a mesh network was able to be established.

iksaif: wow, my comment was for prometheus, not alertmanager, sorry for the confusion.

For alertmanager we could just make sure that nothing is horribly wrong with the nflog or silences ?

brancz: both of those points are applicable for Alertmanager as well though :)

simonpasquier: > For alertmanager we could just make sure that nothing is horribly wrong with the nflog or silences ?
Looking at the code I don't see anything that would stop silences and nflog from working once they have been instantiated. We could check that the maintenance hasn't failed for the last X calls but it seems a bit overkill to me.

simonpasquier:
> If mesh is enabled, then I would expect readiness to express that at least a successful connection to at least one peer has been established

hmm I imagine that it is ok for a mesh with at least 3 nodes where you tolerate one node to be down. But with 2 nodes cluster, it wouldn't work, right?
Maybe we can check the mesh status only when number of configured mesh peers > 2?

ui/web.go
brancz: > We could check that the maintenance hasn't failed for the last X calls but it seems a bit overkill to me.

I want real metrics to alert on something like that, that happens at runtime. If instantiation is successful and there is nothing standing in the way of setting up the HTTP server to serve requests I'd say it's healthy. Readiness is to express that operationally we assume requests will actually successfully served including, basically if all other dependencies are "ready".

stuartnelson3:
post holiday poke to @brancz, if you have time to follow up on this

brancz:
As the Alertmanager process exists if the nflog or silencelog from disk is not loadable, this lgtm for healthiness.

PR #1164, opened 2017-12-22
Display connections in the Status page

simonpasquier:
Relates to #625.

## _Mesh status with 2 nodes and one being down_

image

## _Same cluster with the 2 nodes up_

image


brancz:
lgtm, @stuartnelson3 should probably glance over the elm parts

stuartnelson3:
Interesting, have you seen the state leave `pending`? The alertmanagers are definitely gossiping silences, but the status is never transitioning from `pending` to `established`. Interesting.

simonpasquier:
Thx @stuartnelson3! Indeed I've spotted the `pending` state but since the info comes from the mesh library and @brancz said that the lib might be replaced eventually, I didn't investigate further.

brancz:
Still worth an investigation I would say, replacing weaveworks/mesh is a big task, it will take a long time.

PR #1167, opened 2017-12-24
Fix error message

### LINKED ISSUE #1165: UI error outputs html as text
**What did you do?**
Navigated within the UI after the backend had become unavailable
A nicely formatted error message
HTML as text
**Environment**

* System information:

* Alertmanager version:
v0.12.0

image

w0rm:
closes #1165

Changed to not render an html body, only a status code and a status message, e.g.:

screen shot 2017-12-24 at 15 23 08


Conorbro:
Looks good, thanks!

PR #1171, opened 2017-12-29
Fix expire buttons on the silences page

### LINKED ISSUE #1170: UI: silence's expire button always expires the last silence in list (not the one the button is clicked for)
**What did you do?**
Go to "Silences" view, "Active" tab. Click "Expire" button on the first silence on the list.

The first silence got expired.

The last silence onlist got expired. It always happens when there is more than 1 silence on the list.

**Environment**

* System information:

Chrome 63, MacOS 10.13.2

* Alertmanager version:

alertmanager, version 0.12.0 (branch: HEAD, revision: fc33cc78036f82ef8d4734c197a96f7cb6c952a3)
  build user:       root@c9169eb10d06
  build date:       20171215-14:13:20
  go version:       go1.9.2

w0rm:
Fixes #1170

We were rendering many confirmation dialogs for all silences from the list. The last one was always on top. I fixed this by keeping the silence id and only showing confirmation for this only silence.


Conorbro:
Looks good to me.

PR #1172, opened 2017-12-29
UI: Fix JavaScript error in MSIE due to endswith() usage

hoffie:
MSIE does not support endswith() [1]. substr() can be used to work around this limitation.

@stuartnelson3: Could this be merged?

[1] https://docs.microsoft.com/en-us/scripting/javascript/reference/endswith-method-string-javascript


w0rm:
@hoffie thanks for your contribution!

PR #1175, opened 2018-01-02
Add --expired-within option to list only recently expired silences

tbregolin:


stuartnelson3:
thanks for the submission! sorry for the delay, was off on vacation. I'll give this API a think, I can understand your desire for this feature.

tbregolin:
No worries, thanks for looking.

I thought of making '--expired' invert the value of '--within' so that it looks backwards and shows only the expired silences instead of all of them plus the ones about to expire, but that would modify existing behaviour, hence the new option.

sinkingpoint:
I agree with you @stuartnelson3 - I like the idea. I can't really think of a use case for the flag where we would care that much about active silences and we couldn't just use a second command to pull them. 

I think the main usecase for us around this flag is for finding whether the alert that just fired is something that just happened or a silence that just came unsilenced, so for that, I say go ahead with @tbregolin 's idea of 

> making '--expired' invert the value of '--within' 

I'll ask around the team though to make sure.

stuartnelson3:
Cool, let us know what you find

sinkingpoint:
Yeah I think we're fine to change it here. Internally we actually log all silence expires so most people just check that. Thus this flag is basically not used by us.

stuartnelson3:
@tbregolin do you want to make that change? `--expired` only returns expired silences, and inverts `--within` so that `amtool silences --expired --within 2h` will show just the silences that expired within the last two hours?

tbregolin:
@stuartnelson3 sure. That was my initial patch, so I just have to push it.

Shall we close this PR and open another one?

stuartnelson3:
sure, might as well

Issue #1152, opened 2017-12-20
Reloading configuration via HTTP-POST fails but returns 200


anthraxn8b:
**What did you do?**

1. I intentionally provided a bad config file!
2. Reload configuration via POST on "http://myhostname:9093/-/reload".

**What did you expect to see?**

An HTTP-400 return code.

**What did you see instead? Under which circumstances?**

An HTTP-200 return code:

Response (0.054s)
200 OK
HEADER:
* Date: Wed, 20 Dec 2017 08:04:18 GMT
* Content-Length: 31
* Content-Type: text/plain; charset=utf-8

CONTENT:
"Reloading configuration file..."

**Environment**

* System information:

    Docker image "prom/alertmanager:v0.11.0"
    Linux 3.10.0-514.6.1.el7.x86_64 x86_64

* Alertmanager version:

    alertmanager, version 0.11.0 (branch: HEAD, revision: 30dd0426c08b6479d9a26259ea5efd63bc1ee273)
    build user:       root@3e103e3fc918
    build date:       20171116-17:43:56
    go version:       go1.9.2

* Prometheus version:

    NOT REQUIRED FOR THIS BUG

* Alertmanager configuration file:

    NOT REQUIRED FOR THIS BUG

* Prometheus configuration file:

    NOT REQUIRED FOR THIS BUG

* Logs:
```
level=info ts=2017-12-20T08:04:18.396762583Z caller=main.go:293 msg="Loading configuration file" file=/etc/alertmanager/config.yml
level=error ts=2017-12-20T08:04:18.397019223Z caller=main.go:296 msg="Loading configuration file failed" file=/etc/alertmanager/config.yml err="unknown fields in global: opsgenie_api_host, hipchat_url"

```

PR #1180, opened 2018-01-06
Return reload status from http endpoint (#1152)

stuartnelson3:


ui/web.go
```diff
@@ -72,8 +73,16 @@ func Register(r *route.Router, reloadCh chan<- struct{}, logger log.Logger) {
...
 
 	r.Post("/-/reload", func(w http.ResponseWriter, req *http.Request) {
-		w.Write([]byte("Reloading configuration file..."))
-		reloadCh <- struct{}{}
+		errc := make(chan error)
+		defer close(errc)
+
+		reloadCh <- errc
+		if err := <-errc; err != nil {
+			http.Error(w, fmt.Sprintf("config file reload failed: %v", err), http.StatusInternalServerError)
```
stuartnelson3: not sure if it's necessary to provide the error via http, or it's enough to just say the reload failed. Open to opinions.

ui/web.go
```diff
@@ -72,8 +73,16 @@ func Register(r *route.Router, reloadCh chan<- struct{}, logger log.Logger) {
 	))
 
 	r.Post("/-/reload", func(w http.ResponseWriter, req *http.Request) {
-		w.Write([]byte("Reloading configuration file..."))
-		reloadCh <- struct{}{}
+		errc := make(chan error)
+		defer close(errc)
+
+		reloadCh <- errc
```
brancz: I don't think passing around channels like this is appropriate here, it's only ever going to be a single error received from `errc`.

ui/web.go
```diff
@@ -72,8 +73,16 @@ func Register(r *route.Router, reloadCh chan<- struct{}, logger log.Logger) {
...
+		if err := <-errc; err != nil {
+			http.Error(w, fmt.Sprintf("config file reload failed: %v", err), http.StatusInternalServerError)
+			return
+		}
+
+		w.Write([]byte("config file reloaded"))
```
brancz: An empty 200 response is probably just as good. What does the Prometheus server do? I would out of consistency do the same.

brian-brazil:
I wouldn't worry about exposing information here, all sorts of information is available via the web interface already including the ability to create and delete alerts. The only thing that'd be a problem would be secrets, and it's a bug to expose those in logs too.

I think this feature is a good idea.

ui/web.go
stuartnelson3: Coincidentally, it seems prometheus is using this pattern: https://github.com/prometheus/prometheus/blob/master/web/web.go#L638-L644

ui/web.go
stuartnelson3: Same link from above: https://github.com/prometheus/prometheus/blob/master/web/web.go#L638-L644

Looks like we're only writing something on failure

ui/web.go
brancz: I've been trying to think of a solution, but can't off the top of my head, so I think it's fine to go with it then.

brancz:
lgtm on green :+1:

PR #1185, opened 2018-01-09
Re-introduce prometheus durations in amtool silence creation

### LINKED ISSUE #1183: amtool no longer takes duration units greater than hours
**What did you do?**

    amtool silence add -e 1w -c 'Testing' instance=somethingthatdoesntexist

A silence being created

    amtool: error: time: unknown unit w in duration 1w

We believe this is a regression introduced by https://github.com/prometheus/alertmanager/pull/976 which changes the type of the cli arguments. Before this commit we took Strings and them parsed to them to Durations (The internal model, not the std lib one) ourselves. This PR instead takes std lib Durations which, according to [the docs](https://golang.org/pkg/time/#ParseDuration) only support units up to hours. 

Considering it's reasonably common to want to silence an alert for a day, or a week, or more, having to convert this into hours is tedious.

stuartnelson3:
The expanded duration parsing (support for w/d/y) was removed in the migration to kingpin. This adds it back.

Fixes #1183

### DIFF
--- a/cli/silence_add.go
+++ b/cli/silence_add.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/alecthomas/kingpin"
 	"github.com/prometheus/alertmanager/types"
+	"github.com/prometheus/common/model"
 )
 
 type addResponse struct {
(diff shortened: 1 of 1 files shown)


stuartnelson3:
@sinkingpoint let me know if this works for you (seemed to for me)

sinkingpoint:
Yup this works for me. Thanks for picking it up @stuartnelson3

PR #1186, opened 2018-01-09
Add insecure_skip_verify to skip TLS verification

AndreaGiardini:
Hello!

I added the `insecure_skip_verify` option to skip TLS verification. The same option is present already in prometheus. I tried the patch and it seems to work correctly.

First time go contributor, waiting for feedback :)


brancz:
There are some efforts to get all the HTTP configurations into Alertmanager, but there was some contention around it, so I'm hoping we can resolve that soon and finally have them. I don't think this will be necessary then.

brian-brazil:
This PR is about SMTP, not HTTP. We already have a setting for not requiring TLS, do we really need this? An SMTP server presenting an invalid cert seems odd.

AndreaGiardini:
In my case I need to send email to the SMTP server using an ssh tunnel (I can't connect directly from the k8s cluster to the mailserver). The mailserver requires tls, but alertmanager complains about the certificate since i am using a different hostname.

AndreaGiardini:
Can I have an update on this pull request?

As said above, from GCP is impossible to reach external mail services (except very few, but for example office365 is not included) so i need to forward emails using a proxy service.
The mail certificate shown is valid for the office365 domain, but not for our internal k8s service. This is why we need this patch in alertmanager.

brancz:
Looks like it's been picked up again, with the full config: https://github.com/prometheus/alertmanager/pull/1222

AndreaGiardini:
Does it include also smtp? It doesn't look like to me...

brancz:
This is embarrassing ... :facepalm: don’t mind me, I’m still in post fosdem recovery.

ramonskie:
any update on this?

AndreaGiardini:
@ramonskie I can rebase it and resolve the conflicts, but I would like to understand from one of the maintainers if they intend to merge this feature or not. I see several people interested in this.

ramonskie:
@AndreaGiardini i thought this was for amtool.
but at a closer look it seems that this is not the case

c0ff3e:
Hi,  this is something i need.
as we run a internal DNS and none public CA.

Issue #427, opened 2016-07-13
Flaky test


TheTincho:
Debian users are already starting to report bugs :-)

I can confirm this one. I have seen it once or twice and never again.

This seems like a timer is firing a few µs too early, or a rounding error?

```
The arm64 build of prometheus-alertmanager failed:

  === RUN   TestAggrGroup
  --- FAIL: TestAggrGroup (1.30s)
    dispatch_test.go:132: received batch to early after 299.982ms
  === RUN   TestInhibitRuleHasEqual
  --- PASS: TestInhibitRuleHasEqual (0.00s)
  === RUN   TestInhibitRuleGC
  --- PASS: TestInhibitRuleGC (0.00s)
  === RUN   TestRouteMatch
  --- PASS: TestRouteMatch (0.00s)
  FAIL
  FAIL  github.com/prometheus/alertmanager  1.314s
```

TheTincho:
Another flaky test due to timing issues: `TestBatching` sometimes complains it received 7 instead of 12 alerts. It seems it is highly dependent on timing, so I am disabling it in Debian.

TheTincho:
Found two more flaky tests: TestResolved and TestResolvedFilter, which you can see in this build log: https://buildd.debian.org/status/fetch.php?pkg=prometheus-alertmanager&arch=s390x&ver=0.5.1%2Bds-2&stamp=1481542495

TheTincho:
The timing issues in `dispatch_test.go` get pretty serious under contention. I had already many build failures due to this, and went on to increase the tolerance to 10%, which seems to defeat the purpose of that test. Sadly, I can't find the root cause for these discrepancies.

Also, today I got another build failure test due to the test server not being up fast enough in the acceptance tests. I produced a patch that I think could be merged, I will send a PR.

simonpasquier:
I'm having a look at it since I've experienced the same problems while submitting a PR. I think I can at least explain the problem with `TestBatching` (has `repeat_interval = 5s` and `group_interval = 1s`):
* The dispatcher flushes the alerts at `t0`, sends the notification and marks the notification log at `t0+epsilon`.
* The dispatches flushes the alerts at `t1`, `t2`, `t3` and `t4` and doesn't send the notifications as expected.
* At `t5`, the dispatcher flushes the alerts but then `current_time - (t0+epsilon)` is a bit less then `repeat_interval` so nothing is sent.

Now the notification should be sent again at `t6` for sure.

[...]

PR #1193, opened 2018-01-11
Fix flaky TestBatching acceptance test

simonpasquier:
Fix for #427.

This change decreases the repeat_interval parameter from 5s to 4.9s to
make sure that the alerts are effectively sent after 5 seconds.

The workflow is:
- The dispatcher flushes the alerts at t0, sends the notification and
marks the notification log at t0+epsilon.
- The dispatcher flushes the alerts at t1, t2, t3 and t4 and doesn't
send the notifications as expected.
- At t5, the dispatcher flushes the alerts because current_time - (t0+epsilon)
is less then repeat_interval.

If repeat_interval is exactly 5s, there is a little chance that it is
greater than current_time - (t0+epsilon).

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -347,7 +347,7 @@ func (ag *aggrGroup) run(nf notifyFunc) {
 	for {
 		select {
 		case now := <-ag.next.C:
-			// Give the notifcations time until the next flush to
+			// Give the notifications time until the next flush to
 			// finish before terminating them.
 			ctx, cancel := context.WithTimeout(ag.ctx, ag.timeout(ag.opts.GroupInterval))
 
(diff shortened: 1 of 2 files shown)


dispatch/dispatch.go
```diff
@@ -347,7 +347,7 @@ func (ag *aggrGroup) run(nf notifyFunc) {
 	for {
 		select {
 		case now := <-ag.next.C:
-			// Give the notifcations time until the next flush to
+			// Give the notifications time until the next flush to
```
stuartnelson3: 🥇

simonpasquier:
@stuartnelson3 forgot to mention that I ran the test a hundred times on my machine and it didn't fail so I'm quite confident that it will do the job. If you're aware of other tests which are flaky, I can have a look too.

PR #1205, opened 2018-01-22
Notify only when new firing alerts are added

### LINKED ISSUE #1195: AlertManager Sends alert notification when the alert in a group gets resolved
Whenever an alert in a group gets resolved alertmanager sends a notification of an alert with the remaining alerts in a group. This notification is not of a resolved alert but it is of the remaining alerts.

For example if i have 4 alerts(A, B, C, D) firing in a group. I get a notification with 4 alerts firing. Now if an alert 'C' gets resolved. I get a notification with 3 alerts(A, B, D) firing after group_interval period. Ideally this shouldn't happen since i have already received a notification of these alerts (A, B, D) firing. We should only get a notification if any new alert is added in a group.

* Alertmanager version:
0.12.0

* Prometheus version:
2.0

For config file:
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/prometheus-users/4nibd10rJ_0/eqKZ_1gSAQAJ

simonpasquier:
After the initial notification has been sent, AlertManager shouldn't notify the
receiver again when no new alerts have been added to the group during
group_interval.

This change also modifies the acceptance test framework to assert that no
notification has been received in a given interval.

Fixes #1195.

cc @brian-brazil, @stuartnelson3, @brancz. Since https://github.com/prometheus/alertmanager/pull/703 (meaning 0.6.0), AlertManager sends notifications either when a new alert is firing or when a new alert gets resolved. Brian suggested in #1195 that it shouldn't in the latter case.

### DIFF
--- a/nflog/nflogpb/set.go
+++ b/nflog/nflogpb/set.go
@@ -21,21 +21,6 @@ func (m *Entry) IsFiringSubset(subset map[uint64]struct{}) bool {
 		set[m.FiringAlerts[i]] = struct{}{}
 	}
 
-	return isSubset(set, subset)
-}
-
-// IsResolvedSubset returns whether the given subset is a subset of the alerts
-// that were resolved at the time of the last notification.
-func (m *Entry) IsResolvedSubset(subset map[uint64]struct{}) bool {
-	set := map[uint64]struct{}{}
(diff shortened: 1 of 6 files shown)


brancz:
That sounds reasonable to me. lgtm

PR #1207, opened 2018-01-25
improved error when victorops fails

caarlos0:
I am getting errors like this:

```
level=error ts=2018-01-25T11:14:09.509893874Z caller=dispatch.go:266 component=dispatcher msg="Notify for alerts failed" num_alerts=1 err="cancelling notify retry for \"victorops\" due to unrecoverable error: unexpected status code 404"
level=error ts=2018-01-25T11:14:11.027745774Z caller=notify.go:302 component=dispatcher msg="Error on notify" err="cancelling notify retry for \"victorops\" due to unrecoverable error: unexpected status code 404"
level=error ts=2018-01-25T11:14:11.027791108Z caller=dispatch.go:266 component=dispatcher msg="Notify for alerts failed" num_alerts=1 err="cancelling notify retry for \"victorops\" due to unrecoverable error: unexpected status code 404"
level=error ts=2018-01-25T11:14:11.038292052Z caller=notify.go:302 component=dispatcher msg="Error on notify" err="cancelling notify retry for \"victorops\" due to unrecoverable error: unexpected status code 404"
level=error ts=2018-01-25T11:14:11.038324649Z caller=dispatch.go:266 component=dispatcher msg="Notify for alerts failed" num_alerts=1 err="cancelling notify retry for \"victorops\" due to unrecoverable error: unexpected status code 404"
```

its hard to find out which alerts are erroring, I thought that maybe improve the error adding the alert names might help...

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -16,7 +16,6 @@ package notify
 import (
 	"fmt"
 	"sort"
-	"strings"
 	"sync"
 	"time"
 
@@ -601,13 +600,13 @@ func (r RetryStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.Ale
 		case <-tick.C:
 			if retry, err := r.integration.Notify(ctx, alerts...); err != nil {
 				numFailedNotifications.WithLabelValues(r.integration.name).Inc()
-				level.Debug(l).Log("msg", "Notify attempt failed", "attempt", i, "integration", r.integration.name, "err", err)
+				var alertnames []string
+				for _, al := range alerts {
+					alertnames = append(alertnames, al.Name())
+				}
+				level.Debug(l).Log("msg", "Notify attempt failed", "attempt", i, "integration", r.integration.name, "alerts", alertnames, "err", err)
 				if !retry {
-					var alertnames []string
-					for _, al := range alerts {
-						alertnames = append(alertnames, al.Name())
-					}
-					return ctx, alerts, fmt.Errorf("cancelling notify retry for %q (alerts: %s) due to unrecoverable error: %s", r.integration.name, strings.Join(alertnames, ", "), err)
+					return ctx, alerts, fmt.Errorf("cancelling notify retry for %q due to unrecoverable error: %s", r.integration.name, err)
 				}
 
 				// Save this error to be able to return the last seen error by an
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -16,7 +16,6 @@ package notify
 import (
 	"fmt"
 	"sort"
-	"strings"
 	"sync"
 	"time"
 
@@ -239,7 +238,7 @@ func createStage(rc *config.Receiver, tmpl *template.Template, wait func() time.
 		var s MultiStage
 		s = append(s, NewWaitStage(wait))
 		s = append(s, NewDedupStage(notificationLog, recv))
-		s = append(s, NewRetryStage(i))
+		s = append(s, NewRetryStage(i, recv))
 		s = append(s, NewSetNotifiesStage(notificationLog, recv))
 
 		fs = append(fs, s)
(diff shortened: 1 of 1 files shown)


stuartnelson3:
I would prefer to leave the error message as is, and add the alert names to the debug level logging above.

caarlos0:
@stuartnelson3 done

notify/notify.go
```diff
@@ -600,7 +600,11 @@ func (r RetryStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.Ale
 		case <-tick.C:
 			if retry, err := r.integration.Notify(ctx, alerts...); err != nil {
 				numFailedNotifications.WithLabelValues(r.integration.name).Inc()
-				level.Debug(l).Log("msg", "Notify attempt failed", "attempt", i, "integration", r.integration.name, "err", err)
+				var alertnames []string
```
stuartnelson3: one final nit .. could you change this to
```go
alertnames := make([]string, 0, len(alerts))
```

makes sure we only allocate memory once.

caarlos0: sure! :)

caarlos0:
@stuartnelson3 done

caarlos0:
yeah, that didn't seem to work:

```
level=debug ts=2018-01-25T11:51:09.010972223Z caller=notify.go:607 component=dispatcher msg="Notify attempt failed" attempt=1 integration=victorops alerts="unsupported value type" err="unexpected status code 404"
```

caarlos0:
joining strings work as expected.

stuartnelson3:
Actually, now that I'm thinking about this, would it really be the fault of a particular alert for failure? I think it would be your configuration of a particular receiver, rather than the alert.

caarlos0:
> Actually, now that I'm thinking about this, would it really be the fault of a particular alert for failure? I think it would be your configuration of a particular receiver, rather than the alert.

yeah, but I have dozens of different receivers on 2 different accounts.

I tried to access the receiver from the alert but wasn't able to... can I do that?

Anyway, by the alertnames I was able to found the issue, I think.

Another question: alerts are "bucketed" to be sent to victorops, right? It seems like one of the accounts api_key was revoked (still looking into that), and some alerts from the other api_key were not routed to victorops as well. Is it possible that a failing alert  in a bucket abort the entire bucket (thus not even trying to send the other alerts)?

caarlos0:
actually seems like one of the accounts was removed... so that explains the 404

stuartnelson3:
The receiver (and its name) is available in the `createStage` function. You could add that to the creation on the retry stage.

I'm not sure what you mean by failing alert, but I'm assuming that if the connection info to victor ops is correct, it will receive the alerts successfully.

caarlos0:
(sorry)

caarlos0:
the problem was the following:

- we had 2 accounts set up (lets say engineering and sales)
- several alerts, some going to engineering and others to sales
- the sales account was deleted (on victorops), but alerts were still routed to it
- some alerts that should have gone to engineering account didn't and there were lots of 404 as the description of this PR
- some other alerts were routed to engineering account just fine

so my question is: is it possible that one bad config affects other alerts?

as far as the stage thing goes, I'll try to do that, thanks! :)

stuartnelson3:
The two receivers should be totally separate and with separate API keys. Since you were sending lots of failing alerts for sales (and retrying), potentially you were being rate-limited based on your alertmanager IP address, which caused the engineering alerts to get blocked. I don't know unfortunately. Typically there's rate limiting per API token, but they might also have a more general IP rate limiting in place. This would be a question for victor ops.

notify/notify.go
```diff
@@ -238,7 +238,7 @@ func createStage(rc *config.Receiver, tmpl *template.Template, wait func() time.
 		var s MultiStage
 		s = append(s, NewWaitStage(wait))
 		s = append(s, NewDedupStage(notificationLog, recv))
-		s = append(s, NewRetryStage(i))
+		s = append(s, NewRetryStage(i, recv))
```
stuartnelson3: could you update this to only pass in the `GroupName`? No reason to pass around more data than is necessary.

caarlos0: done

caarlos0:
> Thanks! Sorry about all the back and forth :)

no problem, thanks!

Issue #1200, opened 2018-01-16
Replace weaveworks/mesh


Gouthamve:
Quoting @brancz :

> It's effectively unmaintained, and we have various known problems; initial investigation shows hashicorp/memberlist might be a good candidate, as it has many similar primitives in its library

brancz:
I've started this, but it's in a very very early stage.

brb:
Out of curiosity, could you link to problems you encountered with weaveworks/mesh? Thanks.

simonpasquier:
FWIW this is closed by #1232.

brancz:
@simonpasquier thanks for catching, yes!

cstyan:
the GSoC page still lists this as a project: https://github.com/cncf/soc#replace-weaveworksmesh

brancz:
Thanks for noticing @cstyan! I opened a PR to remove it.

PR #1209, opened 2018-01-25
Add /-/ready based on mesh status

iksaif:
This is not a final proposal, merely an attempt to get an idea of what should be done.

As an heuristic for mesh readyness, try to see if
the mesh looks stable (the number of peers isn't changing too much).
This implementation always mark the altermanager as ready after a maximum of 60s.

It was briefly discussed in 991, would be affected by #1200.

It's important not to start processing requests ASAP when you have thousands of alerts and silences to gossip and alertmanager instances in other continents.


stuartnelson3:
Sorry for the delay on this, I'll take a look next week

cmd/alertmanager/main.go
```diff
@@ -388,6 +396,45 @@ func (s peerDescSlice) Len() int           { return len(s) }
...
+	for {
+		time.Sleep(PollInterval)
+		if nOkay >= NumOkayRequired {
+			level.Info(logger).Log("msg", "No gossip backlog; proceeding")
+			break
+		}
+		n := len(r.Peers.Descriptions())
+		if n == nPeers {
+			nOkay++
+			level.Debug(logger).Log("msg", "gossip looks settled.")
```
stuartnelson3: can remove the punctuation

cmd/alertmanager/main.go
```diff
@@ -388,6 +396,45 @@ func (s peerDescSlice) Len() int           { return len(s) }
...
+		} else {
+			nOkay = 0
+			level.Info(logger).Log("msg", "Gossip not settled", "polls", totalPolls, "before", nPeers, "now", n)
```
stuartnelson3: have `Gossip` be `gossip`

cmd/alertmanager/main.go
```diff
@@ -388,6 +396,45 @@ func (s peerDescSlice) Len() int           { return len(s) }
...
+		}
+		nPeers = n
+		totalPolls++
+		if totalPolls >= ForceAfter {
+			level.Info(logger).Log("msg", "Gossip not settled but continuing anyway", "polls", totalPolls)
```
stuartnelson3: should there be a `break` here if we've polled more times than `ForceAfter`?

cmd/alertmanager/main.go
```diff
@@ -388,6 +396,45 @@ func (s peerDescSlice) Len() int           { return len(s) }
 func (s peerDescSlice) Less(i, j int) bool { return s[i].UID < s[j].UID }
 func (s peerDescSlice) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
 
+// meshSettle set isReady to 1 once the mesh is considered settle (or after a
+// configurable timeout). The idea is that we don't want to start "working" before
+// we get a chance to know most of the alerts and/or silences.
+func meshSettle(r *mesh.Router, isReady *int32, timeout time.Duration, logger log.Logger) {
```
stuartnelson3: `timeout` doesn't seem to be used

cmd/alertmanager/main.go
```diff
@@ -388,6 +396,45 @@ func (s peerDescSlice) Len() int           { return len(s) }
...
+	// Inspired from https://github.com/apache/cassandra/blob/7a40abb6a5108688fb1b10c375bb751cbb782ea4/src/java/org/apache/cassandra/gms/Gossiper.java
+	// This is clearly not perfect or strictly correct but should prevent the alertmanager
+	// to be used before it is obviously not read.
+	const PollInterval = 1 * time.Second
```
stuartnelson3: this should probably be configurable via command line, with the default being 1 sec

cmd/alertmanager/main.go
```diff
@@ -388,6 +396,45 @@ func (s peerDescSlice) Len() int           { return len(s) }
...
+	const NumOkayRequired = 3
+	const ForceAfter = 60 // Limit this to 10 * PollInterval, this needs to be higher than the mesh connection timeout.
```
stuartnelson3: actually, all three of these should be configurable

cmd/alertmanager/main.go
```diff
@@ -388,6 +396,45 @@ func (s peerDescSlice) Len() int           { return len(s) }
...
+	level.Info(logger).Log("msg", "Waiting for gossip to settle...")
+	time.Sleep(PollInterval)
```
stuartnelson3: is this sleep necessary if the first thing happening in the `for` loop is to sleep?

stuartnelson3:
When sd is properly checking the alertmanager instances and not returning unhealthy instances to prometheus, this seems like a fair start/restart strategy.

There's nothing preventing an alertmanager from restarting and sending alerts based on its on-disk stored state. If we are preventing an alertmanager from receiving any new alerts, maybe we should also gate the sending of alerts with this same variable until the instance is considered "settled"?

brian-brazil:
> When sd is properly checking the alertmanager instances and not returning unhealthy instances to prometheus, this seems like a fair start/restart strategy.

That doesn't sound in line with how the AM is intended to work, it needs to get all alerts - especially if the mesh is unstable.

stuartnelson3:
>That doesn't sound in line with how the AM is intended to work, it needs to get all alerts - especially if the mesh is unstable.

This is true. Is a configurable (~60sec) timeout acceptable for a booting alertmanager not sending out any alerts, in the hopes it makes a stable mesh connection? Maybe it makes sense to move this implementation from preventing AM being in sd (and getting all alerts) to waiting initially before sending alerts. 

60 seconds probably wouldn't be the end of the world for an initial delay, and the alertmanager would have all alerts and be ready to act on them if it cannot make the mesh connection (this current implementation would potentially have no alertmanagers available via SD if they all were restarted at once).

iksaif:
@stuartnelson3 do you mean delaying most things (not only accepting traffic) until the mesh is considered ready (and maximum after 1 minute or so ?)

I'll update the patch according to your comments during the day.

brian-brazil:
> This is true. Is a configurable (~60sec) timeout acceptable for a booting alertmanager not sending out any alerts, in the hopes it makes a stable mesh connection?

That's not unreasonable, if you've an AM crashlooping that fast you have bigger problems. This might also help a bit with https://github.com/prometheus/prometheus/issues/2585.

iksaif:
>  Maybe it makes sense to move this implementation from preventing AM being in sd (and getting all alerts)

For frameworks supporting `/-/ready` it also makes it clear for the administrator that the instance is healthy but not ready. So I think this bit is important.

> this current implementation would potentially have no alertmanagers available via SD if they all were restarted at once

To be fair that's already the case during the restart. This is just making the restart a little bit longer.

cmd/alertmanager/main.go
```diff
@@ -388,6 +396,45 @@ func (s peerDescSlice) Len() int           { return len(s) }
 func (s peerDescSlice) Less(i, j int) bool { return s[i].UID < s[j].UID }
 func (s peerDescSlice) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
 
+// meshSettle set isReady to 1 once the mesh is considered settle (or after a
+// configurable timeout). The idea is that we don't want to start "working" before
+// we get a chance to know most of the alerts and/or silences.
+func meshSettle(r *mesh.Router, isReady *int32, timeout time.Duration, logger log.Logger) {
+	// Inspired from https://github.com/apache/cassandra/blob/7a40abb6a5108688fb1b10c375bb751cbb782ea4/src/java/org/apache/cassandra/gms/Gossiper.java
+	// This is clearly not perfect or strictly correct but should prevent the alertmanager
+	// to be used before it is obviously not read.
```
simonpasquier: (nit) s/read/ready/

cmd/alertmanager/main.go
```diff
@@ -347,7 +348,14 @@ func main() {
 
 	webReload := make(chan chan error)
 
-	ui.Register(router, webReload, logger)
+	var isReady int32 = 0
+	if *meshListen != "" {
+		go meshSettle(mrouter, &isReady, 5*time.Second, logger)
```
simonpasquier: Replace `logger` by `log.With(logger, "component", "mesh")`?

cmd/alertmanager/main.go
```diff
@@ -388,6 +396,45 @@ func (s peerDescSlice) Len() int           { return len(s) }
...
+	level.Info(logger).Log("msg", "Waiting for gossip to settle...")
+	time.Sleep(PollInterval)
+	
+	nPeers := 0
+	nOkay := 0
+	totalPolls := 0
+	for {
+		time.Sleep(PollInterval)
+		if nOkay >= NumOkayRequired {
+			level.Info(logger).Log("msg", "No gossip backlog; proceeding")
```
simonpasquier: Not sure about this comment. Shouldn't it be `"gossip settled, proceeding"`?

stuartnelson3:
@iksaif which frameworks are you referencing? In kubernetes, readiness is defined as should or should not receive traffic, but alertmanager should always be receiving incoming prometheus alerts.

>@stuartnelson3  do you mean delaying most things (not only accepting traffic) until the  mesh is considered ready (and maximum after 1 minute or so ?)

delaying sending any notifications. AM flushes its state to disk periodically; when an instance starts up and doesn't receive mesh information from a peer it might start sending out duplicate alerts.

iksaif:
> @iksaif which frameworks are you referencing? In kubernetes, readiness is defined as should or should not receive traffic, but alertmanager should always be receiving incoming prometheus alerts.

I'm not sure I understand what you think should be done. Do you think we should accept traffic ASAP but delay sending notifications ? That sounds like it would be more complex to implement and to understand for users/operators.

In your opinion what is the value of accepting traffic ASAP ? (If you restart all alertmanager at once, you're going to have downtime anyway since it probably takes more time to start a new container than to wait for the mesh to settle).

In our usecase, we don't use persistent disk, so it's even more likely that the alertmanager will send duplicate notification (or worse, silenced ones). It means that if we delay notification, this could delay alerts that would otherwise have been sent if this instance hadn't registered to SD right away.

(Note: I'm happy to implement any of the two solutions as both would be an improvement)

stuartnelson3:
>I'm not sure I understand what you think should be done. Do you think we should accept traffic ASAP but delay sending notifications ? That sounds like it would be more complex to implement and to understand for users/operators.

That was my desire, yes. Why is it more difficult to understand? It should result in all alertmanagers receiving alerts as soon as possible, but delay sending any notifications until the mesh has settled. From a user/operator perspective, an AM instance has an initial delay when it attempts to join the mesh before sending out notifications.

>In your opinion what is the value of accepting traffic ASAP ? (If you restart all alertmanager at once, you're going to have downtime anyway since it probably takes more time to start a new container than to wait for the mesh to settle).

Piggy-backing off of @brian-brazil's original comment. The model is for a prometheus server to send alerts to all AM instances, and then for them to coordinate sending notifications. Marking an AM instance as unready so as to avoid receiving alerts breaks that model. What do you see as a benefit in not accepting alerts from prometheus?

>In our usecase, we don't use persistent disk, so it's even more likely that the alertmanager will send duplicate notification (or worse, silenced ones). It means that if we delay notification, this could delay alerts that would otherwise have been sent if this instance hadn't registered to SD right away.

Many places still have alertmanager running on baremetal, which will flush state to disk. If an instance is restarted and takes a while to join the mesh, it may send out alerts before receiving confirmation that another instance has already sent the messages. Waiting for the mesh to settle should give a new instance a current view of the world.

iksaif:
> The model is for a prometheus server to send alerts to all AM instances, and then for them to coordinate sending notifications.

I did not know that, this makes much more sense. I'll come up with a patch doing this.

cmd/alertmanager/main.go
iksaif: done

cmd/alertmanager/main.go
iksaif: done

cmd/alertmanager/main.go
iksaif: yes ! done

cmd/alertmanager/main.go
iksaif: fixed

cmd/alertmanager/main.go
iksaif: I made the two intervals configurable, this should be enough to control the full behavior.

cmd/alertmanager/main.go
iksaif: removed

cmd/alertmanager/main.go
iksaif: done

cmd/alertmanager/main.go
iksaif: done

cmd/alertmanager/main.go
iksaif: done

iksaif:
New version, the important part isn't done yet, I'm not sure how I'll implement the delay itself. I was thinking about adding another stage, that would probably block on a channel. The channel would replace meshReady. Does that look ok to you ?

###  Wait for the mesh to settle before sending notifications

See #1209 for details.

As an heuristic for mesh readyness, try to see if
the mesh looks stable (the number of peers isn't changing too much).
This implementation always mark the altermanager as ready after a maximum of 60s.

This adds two new flags to control this behavior:
```
      --mesh.poll-interval=60s   mesh poll interval during startup.
      --mesh.settle-timeout=10s  mesh settling timeout. Do not wait more than this duration on startup.
```

It also adds `/-/ready` which always return 200 (in order to make it clear
that we are ready as soon as we can receive requests).

The mesh status is exposed in `/api/v1/status` and visible on `/#/status`.

stuartnelson3:
```
--mesh.poll-interval=60s   mesh poll interval during startup.
--mesh.settle-timeout=10s  mesh settling timeout. Do not wait more than this duration on startup.
```

Would `poll-interval` be something like 5s (check mesh connection count against previous connection count every 5s), and `settle-timeout` be 60s (continue booting after 60s regardless of changing mesh connection count)? The values seem flipped in your example and I just wanted to be clear I understood.

stuartnelson3:
I like the settle stage, but whether it aborts the pipeline execution or blocks is something I'll think about. I think blocking makes sense so that any pipelines that attempted to execute will as soon as the mesh has settled.

iksaif:
Yes, the example was supposed to be the other way around. 10s is `notify.MinTimeout`, and I think this kind of makes sense. But I could put 5s instead. I'll try to implement a settling stage to see how it looks.

stuartnelson3:
we can worry about the details on default timings later, no worries

iksaif:
New patch sent with code moved to mesh/mesh.go and a new stage to inhibit alerts.

mesh/mesh.go
```diff
@@ -0,0 +1,271 @@
+// Copyright 2015 Prometheus Team
```
stuartnelson3: 2015 -> 2018

mesh/mesh.go
```diff
@@ -0,0 +1,271 @@
...
+import (
+	"fmt"
+	"net"
+	"sort"
+	"strconv"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/prometheus/client_golang/prometheus"
```
stuartnelson3: this should be dropped into it's own separate import block (we separate stdlib, 3rd party, and then prometheus imports)

mesh/mesh.go
```diff
@@ -0,0 +1,271 @@
...
+	return mesh, nil
+}
+
+// Start the mesh.
+func (m *Mesh) Start(peers []string) {
+	m.Router.Start()
+	m.Router.ConnectionMaker.InitiateConnections(peers, true)
+}
+
+// Settle wait until the mesh is ready (and set the appropriate internal state when it is).
```
stuartnelson3: `waits`, `sets`

mesh/mesh.go
iksaif: done

mesh/mesh.go
iksaif: done

mesh/mesh.go
iksaif: oops

brancz:
this highly clashes with changes proposed in #1232 (cc @fabxc )

stuartnelson3:
Is there an estimate on when that will be ready to merge? I think the logic here around settling and having a stage that prevents notifications before being settled will be able to remain mostly unchanged. If the memberlist PR will be ready in a week we could put this on hold until then?

mesh/mesh.go
```diff
@@ -0,0 +1,272 @@
...
+	// This is especially important for those that do not have persistent storage.
+	const NumOkayRequired = 3
+	level.Info(m.logger).Log("msg", "Waiting for gossip to settle...")
+	start := time.Now()
+	nPeers := 0
+	nOkay := 0
+	totalPolls := 0
+	for {
+		time.Sleep(interval)
+		elapsed := time.Since(start)
```
brancz: sleep and timeout should be handled with a select statement here and a passed context into the `Settle` function call

mesh/mesh.go
```diff
@@ -0,0 +1,272 @@
...
+		n := len(m.Router.Peers.Descriptions())
```
brancz: I wonder if this can have a race where a deployment doesn't continue because the first peer has discovered where other peers are supposed to live, but can't discover them in mesh.

iksaif: There is a global timeout for Settle(), so we should be fine.

mesh/mesh.go
iksaif: pushing a patch to do that

iksaif:
https://github.com/prometheus/alertmanager/pull/1232 has been merged, I'll rebase

iksaif:
Rebased

cmd/alertmanager/main.go
```diff
@@ -172,6 +169,12 @@ func main() {
 			level.Error(logger).Log("msg", "Unable to initialize gossip mesh", "err", err)
 			os.Exit(1)
 		}
+		ctx, cancel := context.WithTimeout(context.Background(), *settleTimeout)
+		defer func() {
+			cancel()
+			peer.Leave(10 * time.Second)
+		}()
+		go peer.Settle(ctx, *pushPullInterval / 2)
```
simonpasquier: `*pushPullInterval / 2` means that by default the interval is 30s while at the same time the context time-out is 60s. In practice the Settle() function always terminates on context time-out since it waits for 4 intervals.

cluster/cluster.go
```diff
@@ -191,6 +221,46 @@ func (p *Peer) Peers() []*memberlist.Node {
...
+	// This is clearly not perfect or strictly correct but should prevent the alertmanager to send notification before it is obviously not ready.
+	// This is especially important for those that do not have persistent storage.
+	const NumOkayRequired = 3
+	level.Info(p.logger).Log("msg", "Waiting for gossip to settle...", "interval", interval)
+	start := time.Now()
+	nPeers := 0
+	nOkay := 0
+	totalPolls := 0
+	for {
+		select {
```
simonpasquier: Any reason why the select block isn't at the end of the for block? Currently it takes 4 `interval` for the function to return if everything's ok.

cluster/cluster.go
```diff
@@ -169,6 +175,30 @@ func (p *Peer) ClusterSize() int {
...
+func (p *Peer) Ready() bool {
+	select {
+	case <-p.readyc:
+		return true
+	default:
+	}
+	return false
+}
+
+// Wait until Settle() as finished.
```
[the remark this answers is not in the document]
iksaif: done

notify/notify.go
```diff
@@ -318,6 +321,23 @@ func (fs FanoutStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.A
 	return ctx, alerts, nil
 }
 
+// GossipSettleStage waits until the Gossip has settle to forward alerts.
```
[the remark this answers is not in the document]
iksaif: done

notify/notify.go
```diff
@@ -318,6 +321,23 @@ func (fs FanoutStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.A
...
+type GossipSettleStage struct {
+	peer *cluster.Peer
+}
+
+// NewGossipSettleStage return a new GossipSettleStage
```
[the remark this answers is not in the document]
iksaif: done

cmd/alertmanager/main.go
iksaif: You are right, I maybe let's use `gossipItnerval * 10` instead. I know it's a magic value, but it should mostly work.

cluster/cluster.go
iksaif: yes, it is at the beginning because we start with a `Sleep()` in order to have "some" data when we look at `Peers()` for the first time. As you pointed out it means that we always wait once, then we wait enough to have the thing stabilized for three rounds.

iksaif:
@simonpasquier  / @stuartnelson3  do you think it looks ok now ? Thanks !

simonpasquier:
LGTM, just need to rebase on master.

iksaif:
rebased

PR #1213, opened 2018-01-26
Jomenxiao/support push to kafka

jomenxiao:
We want to archive alert messages in Kafka cluster forever.
so code it to push alert messages to Kafka.


stuartnelson3:
This would be better handled via the webhook receiver. Define your first route as your webhook, set the route to `continue` so it will match your webhook but then continue searching the routing tree, and then define the rest of your routes as normal.

https://prometheus.io/docs/alerting/configuration/#route

Issue #1211, opened 2018-01-25
Alert severity not correctly set on PagerDuty Alert


dbonatto:
I'm trying to migrate to the PagerDuty API v2. I already changed to use the routing_key in the configuration so that AM would use the API v2. What I can not find how to do is passing on to PagerDuty the severity that is declared as a label in the alert.

```
curl -XPOST -k https://somehost/api/v1/alerts -d '[{
        "status": "firing",
        "labels": {
                "alertname": "TestAlertWarning-13127",
                "service": "my-service",
                "team": "C Team",
                "severity":"warning",
                "instance": "TestAlertWarning-13127.example.net"
        },
        "annotations": {
                "summary": "High latency is high!"
        },
        "generatorURL": "http://prometheus.int.example.net/<generating_expression>"
}]'
```
What is happening now is that AM is sending a "error" severity to PagerDuty, which looking in the code I could see that is the default behavior when no severity is set.

Not sure if it is a code or documentation issue, would appreciate some help.

* Alertmanager version:

Branch: HEAD
BuildDate: 20180112-10:32:46
BuildUser: root@d83981af1d3d
GoVersion: go1.9.2
Revision: fb713f6d8239b57c646cae30f78e8b4b8861a1aa
Version: 0.13.0

* Prometheus version:

Version 	1.8.2
Revision 	5211b96d4d1291c3dd1a569f711d3b301b635ecb
Branch 	HEAD
BuildUser 	root@1412e937e4ad
BuildDate 	20171104-16:09:14
GoVersion 	go1.9.2

* Alertmanager configuration file:
```
  - receiver: CTeam-pagerduty-alert
    match:
      team: C Team
...
- name: CTeam-pagerduty-alert
  pagerduty_configs:
  - send_resolved: true
    routing_key: <secret>
    url: https://events.pagerduty.com/v2/enqueue
    client: '{{ template "pagerduty.default.client" . }}'
    client_url: '{{ template "pagerduty.default.clientURL" . }}'
    description: '{{ template "pagerduty.default.description" .}}'
    details:
      firing: '{{ template "pagerduty.default.instances" .Alerts.Firing }}'
      num_firing: '{{ .Alerts.Firing | len }}'
      num_resolved: '{{ .Alerts.Resolved | len }}'
      resolved: '{{ template "pagerduty.default.instances" .Alerts.Resolved }}'
```

stuartnelson3:
Severity is set in `pagerduty_configs`:

```yaml
[...]

PR #1214, opened 2018-01-26
 Fixes prometheus/alertmanager#1211

dbonatto:
@stuartnelson3 here is the proposal, as we discussed in the issue #1211 in case you agree that this change makes sense to be part of the AM official release.


stuartnelson3:
Looks good. If you want, you could update the documentation here: https://github.com/prometheus/docs/blob/23e294832437b8bf17b238cb3711a43963ee85a8/content/docs/alerting/configuration.md#pagerduty_config

Severity can be added, noting that it's templated, and its default value is "error".

Issue #1223, opened 2018-02-01
OpsGenie returns 422 when team is not specified


iksaif:
I'm still unsure if this is due to something on our side, but to help others users potentially affected by this I'm creating this issue.

Using the opsgenie notifier, we started to get the following error (with both 0.11.0 and 0.13.0).

{code}
{"message":"Request body is not processable. Please check the errors.","errors":{"teams[0]":"id or name should be provided"},"took":0.009,"requestId":"xxxxxxxxxxxxxxxxxxxxxxxxxxxx"}
{code}

this happens because the query looks like:

{code}
{..."teams":[{"name":""}],...}
{code}

we made the following very quick and dirty patch to fix the issue

{code}
$ git diff
diff --git a/notify/impl.go b/notify/impl.go
index 7eddc3e..c519039 100644
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -30,6 +30,7 @@ import (
        "net/url"
        "strings"
        "time"
+       "io/ioutil"
 
        "github.com/go-kit/kit/log"
        "github.com/go-kit/kit/log/level"
@@ -837,8 +838,13 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
                apiURL = n.conf.APIURL + "v2/alerts"
                var teams []map[string]string
                for _, t := range strings.Split(string(tmpl(n.conf.Teams)), ",") {
-                       teams = append(teams, map[string]string{"name": t})
+                       if len(t) != 0 {
+                               teams = append(teams, map[string]string{"name": t})
+                       }
                }
+               if len(teams) == 0 {
+                        teams = nil
+                }
                msg = &opsGenieCreateMessage{
                        Alias:       alias,
                        Message:     message,
@@ -859,6 +865,7 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
        if err := json.NewEncoder(&buf).Encode(msg); err != nil {
                return false, err
        }
+       fmt.Printf("JSON: %v\n", buf.String())
 
        req, err := http.NewRequest("POST", apiURL, &buf)
        if err != nil {
@@ -868,7 +875,12 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
        req.Header.Set("Authorization", fmt.Sprintf("GenieKey %s", n.conf.APIKey))
 
        resp, err := ctxhttp.Do(ctx, http.DefaultClient, req)
[...]

PR #1224, opened 2018-02-05
Fix OpsGenie notifier and add unit tests

iksaif:
See #1223, looks like OpsGenie now sometimes returns a 422 when you
don't specify a team. This change cleans up the JSON output and
add a few unit tests.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -952,9 +952,38 @@ type opsGenieCloseMessage struct {
 
 // Notify implements the Notifier interface.
 func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
+	req, retry, err := n.createRequest(ctx, as...)
+	if err != nil {
+		return retry, err
+	}
+
+	resp, err := ctxhttp.Do(ctx, http.DefaultClient, req)
+
+	if err != nil {
+		return true, err
+	}
+	defer resp.Body.Close()
+
+	return n.retry(resp.StatusCode)
+}
+
+// Like Split but filter out empty strings.
+func safeSplit(s string, sep string) []string {
+	a := strings.Split(strings.TrimSpace(s), sep)
+	b := a[:0]
+	for _, x := range a {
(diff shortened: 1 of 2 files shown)


notify/impl.go
```diff
@@ -987,45 +1013,39 @@ func (n *OpsGenie) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 
 		apiURL = n.conf.APIURL + "v2/alerts"
 		var teams []map[string]string
-		for _, t := range strings.Split(string(tmpl(n.conf.Teams)), ",") {
+		for _, t := range safeSplit(string(tmpl(n.conf.Teams)), ",") {
```
stuartnelson3: I think this should be returned to just `strings.Split` and then check each `t` for being an empty string. If someone enters `"team1,"` in their config, it'll still end up adding `"name":""` in the json.

iksaif: done

notify/impl.go
```diff
@@ -952,9 +952,39 @@ type opsGenieCloseMessage struct {
...
+		return true, err
+	}
+	defer resp.Body.Close()
+
+	return n.retry(resp.StatusCode)
+}
+
+// Like Split but filter out empty strings.
+func safeSplit(s string, sep string) []string {
+	// https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating
```
stuartnelson3: this can be removed

iksaif: done

dswarbrick:
Is there some reason why Alertmanager doesn't simply use https://github.com/opsgenie/opsgenie-go-sdk?

stuartnelson3:
Not that I'm aware of. Potentially it didn't exist when this was written.

Also, the entire implementation in our codebase is ~150 lines of code. Maybe it wasn't viewed as necessary to add the dependency.

brian-brazil:
I don't believe we were aware of it. We had problems even finding what API docs there were.

PR #1225, opened 2018-02-05
Add mesh metrics

### LINKED ISSUE #1219: Expose gossip status via /metrics
Would be handy for alerting.

simonpasquier:
This change adds 2 new metrics for the mesh:

* alertmanager_peer_connection, state of the connection between the
  Alertmanager instance and a peer.
* alertmanager_peer_terminations_total, total number of terminated
  connection.

It also moves the gathering of the alertmanager_peer_position metric
outside of the meshWait() function so that the metric is computed
accurately even when no alerting group fires.

Fixes #1219.

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -54,10 +54,6 @@ import (
 )
 
 var (
-	peerPosition = prometheus.NewGauge(prometheus.GaugeOpts{
-		Name: "alertmanager_peer_position",
-		Help: "Position the Alertmanager instance believes it's in. The position determines a peer's behavior in the cluster.",
-	})
 	configHash = prometheus.NewGauge(prometheus.GaugeOpts{
 		Name: "alertmanager_config_hash",
 		Help: "Hash of the currently loaded alertmanager configuration.",
@@ -75,7 +71,6 @@ var (
 )
 
 func init() {
-	prometheus.MustRegister(peerPosition)
 	prometheus.MustRegister(configSuccess)
 	prometheus.MustRegister(configSuccessTime)
 	prometheus.MustRegister(configHash)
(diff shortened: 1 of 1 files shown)


cmd/alertmanager/main.go
```diff
@@ -382,31 +378,106 @@ func main() {
...
+	termDesc *prometheus.Desc
+}
+
+func NewMeshCollector(router *mesh.Router) *meshCollector {
+	return &meshCollector{
+		router: router,
+		connDesc: prometheus.NewDesc(
+			"alertmanager_peer_connection",
+			"State of the connection between the Alertmanager instance and a peer.",
+			[]string{"peer", "nick"},
```
brian-brazil: Are both of these labels needed?

simonpasquier: My reasoning is that the nickname makes it easier to identify the peer:

```
alertmanager_peer_connection{instance="prometheus-0",job="alertmanager",nick="prometheus-1",peer="02:42:ac:11:00:03"}  1
alertmanager_peer_connection{instance="prometheus-1",job="alertmanager",nick="prometheus-0",peer="02:42:ac:11:00:02"}  1
```

But we can't get rid of of `peer` altogether since `nick`'s uniqueness isn't enforced by the mesh library.

stuartnelson3: these are both belonging to the connected AM, I imagine only one is necessary

stuartnelson3: I would vote for only using `peer`. If a user wants to make it easier to identify they can set a custom peer-id.

simonpasquier: Fine with me. I'll update the PR accordingly.

Issue #645, opened 2017-03-07
Add service discover for -mesh.peers


iksaif:
Currently -mesh.peers takes a static list of host:port. If you want to use service discovery to discover your peers you need to do that before the command line invocation.

It more or less work with consul because you can use something like dig +noall +answer foo.service.dc.consul SRV, but that is far from optimal.

Do you think it would make sense to add service discovery support for this in the alertmanager ?

stuartnelson3:
I think this sounds like a good addition, but it is worth noting that the current gossip model will discover additional peers. The list you provide at the start is not the "set in stone" list of your peers. For example, if you start peerA, and then start the following peers B, C, and D, all configured with peerA as their initial peer list, they will discover each other eventually.

As far as finding an additional list of peers, though, I do think supporting service discovery makes sense.

brancz:
I've spoken to @fabxc about this and we also came up with something in the direction of what @stuartnelson3 wrote. It is mainly interesting for the initial discovery, and for that reason we didn't think taking on the huge dependency, would be worth it. Actually dynamically discovering and changing peers could disrupt the mesh network, so it may actually harm it - or seem like it doesn't work because it would in fact not dynamically reconfigure like Prometheus does.

The way we handle the mesh network initialization it in the Prometheus Operator is that we don't connect the Alertmanagers through an IP address but through a stable DNS name. All instances deployed then have all other instances as arguments. While this may look odd when scaling up and down it actually always works out as all initial instances are always part of the mesh and by the mesh sharing all connections all peers are eventually connected to each other.

matthiasr:
Since it's only used at startup, I think relying on scripting around for cases where DNS is not available is good enough, compared to the complexity of having full blown SD but then using it very differently. Keep in mind that many Prometheus SDs (all but DNS and file?) rely on relabeling to actually select targets from the global lost, so we'd have to pull that in too.

I would reconsider if someone can demonstrate a case that is trivial with SD and impossible to solve otherwise, but for now I'd say no.

brian-brazil:
I think we'll ultimately have to add it, but let's not do it earlier than we need to.

fabxc:
I don't think we ever have to add it. There is really no point as it is only needed during bootstrapping.

Agreeing with @matthiasr on all ends.

There any script that populates the startup flags with results from querying service discovery serves the purpose just fine. All the complexity of relabeling makes the cost/benefit of adding this underwhelming.
[...]

Issue #1206, opened 2018-01-24
Lost Mesh peer due DNS to IP caching


mjudeikis:
Running Alertmanager in  Statefulset where Peers is being defined as POD IP's using POD dns name:
https://github.com/mjudeikis/origin/blob/prometheus-ha/examples/prometheus/prometheus.yaml#L283-L288

When mesh network is established after startup if one of the members is restarted POD gets different IP address (next from the pool). But survival member is caching IP address of dead pod and trying to reconnect.

This ends up in broken mesh and each alert manager is firing all alerts as they would be in standalone mode. 

I suspect mesh library doing this. But because its already being replaced https://github.com/prometheus/alertmanager/issues/1200 just raising this for awareness.

```
-bash-4.4# oc get pods -o wide                                                                                                                                                                                     
NAME           READY     STATUS    RESTARTS   AGE       IP           NODE                                
alerts-0       2/2       Running   0          7m        172.17.0.5   default                             
prometheus-0   4/4       Running   0          7m        172.17.0.3   default                             
prometheus-1   4/4       Running   0          7m        172.17.0.2   default  
```
Peers:
```
Peers:
Name: 02:42:ac:11:00:02
Nick Name: prometheus-1
UID: 10375467427516783000
Name: 02:42:ac:11:00:03
Nick Name: prometheus-0
UID: 5968773859996667000
```

Now restart prometheus-0 (which will change IP address) and prometheus-1 will try to reconnect to old pod IP. 

prometheus-1:
```
ts=2018-01-24T18:12:40.163695152Z caller=<autogenerated>:1 component=mesh level=debug msg="->[172.17.0.3:6783] error during connection attempt: dial tcp4 :0->172.17.0.3:6783: getsockopt: connection refused"
```
and Pod status now:
```
-bash-4.4# oc get pods -o wide                      
NAME           READY     STATUS    RESTARTS   AGE       IP           NODE                                
alerts-0       2/2       Running   0          33m       172.17.0.5   default                             
prometheus-0   4/4       Running   0          5m        172.17.0.6   default                             
prometheus-1   4/4       Running   0          17m       172.17.0.2   default  
```
I know its trivial problem to solve with different deployment types, but it exist :)

mjudeikis:
closing in favor of https://github.com/prometheus/alertmanager/pull/1232

PR #1228, opened 2018-02-06
[WIP] Resolve peer names directly

### LINKED ISSUE #929: `-mesh.peer` only uses one record returned by DNS
When using the `-mesh.peer` flag with a DNS name value, only one of the records would be used. It'd be nice to add *all* records. For example, if all alertmanagers were listed in Consul's service catalog, a single DNS name could be used. 

This appears to be a limitation of the weaveworks/mesh library:
https://github.com/prometheus/alertmanager/blob/master/vendor/github.com/weaveworks/mesh/connection_maker.go#L96
which uses `net.ResolveTCPAddr`.

See for discussion: 
https://groups.google.com/forum/#!topic/prometheus-developers/176Pl1uK0aE

simonpasquier:
With this change, AlertManager resolves the peers instead of relying on the mesh library. If a peer's resolution fails be resolved, AlertManager retries it for a minute at most.

This fixes a couple of issues/gaps:

* Avoid dropping peers if they aren't resolved immediately at startup.
* Support for A records returning multiple addresses.

Fixes #929 and #645 (partially for now since it doesn't work with SRV records). Before doing further changes, I wanted to get a feeling whether this is a) useful and b) in the right direction.

cc @stuartnelson3 @iksaif @discordianfish

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -214,8 +214,11 @@ func main() {
 
 	// Disable mesh if empty string passed for mesh.listen-address flag.
 	if *meshListen != "" {
-		mrouter.Start()
-		mrouter.ConnectionMaker.InitiateConnections(*peers, true)
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			startRouter(mrouter, peers, stopc, log.With(logger, "component", "mesh"))
+		}()
 	}
 
 	defer func() {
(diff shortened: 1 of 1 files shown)


stuartnelson3:
From re-reading the discussion, it seems like we settled on a minimal implementation that supports this and srv lookups. (cf. https://github.com/oklog/oklog/blob/master/cmd/oklog/forward.go#L145-L187). I'm personally in favor of scripting around it with e.g. config management, but see where this would make peoples' lives easier with little maintenance penalty.

I'm not sure we need to add all the parallelization code, but that can be discussed later if others still think minimal sd is something to add here.

@brancz @brian-brazil

iksaif:
SRV looking would be great, it would work out the box with things such as Consul. I think if we add more code related to the mesh here it might be a good idea to move all of it in a new file, and maybe wrap it a little bit (will help when somebody starts changing the mesh implementation with something else).

simonpasquier:
@iksaif I've seen your other PR and agree that eventually it should be splitted into a separate file.

brian-brazil:
I'm strongly against SRV lookups, we already have a full service discovery system and I don't believe we should start developing a second parallel system.

mjudeikis:
DNS resolution would solve issues when running it in k8s statefulsets where IP address changes constantly behind DNS
 (https://github.com/prometheus/alertmanager/issues/1206) . 
+1

iksaif:
:+1: , having to script stuff to be able to discover the port isn't super friendly (just getting the IP is easy)

simonpasquier:
Putting this PR on hold as https://github.com/prometheus/alertmanager/pull/1232 revamps the whole mesh thing.

discordianfish:
> From re-reading the discussion, it seems like we settled on a minimal implementation that supports this and srv lookups. (cf. https://github.com/oklog/oklog/blob/master/cmd/oklog/forward.go#L145-L187).

Yes, that's what I strongly want too. And only for initial peer discovery, e.g when creating the mesh the first time or when a new member joins. In it's current form the AM is really hard to setup in dynamic environments.

simonpasquier:
Superseded by the refactoring of the clustering part.

Issue #1218, opened 2018-01-30
missing agentid


tom20180101:
System information:

Linux ha-1 3.10.0-693.11.1.el7.x86_64 #1 SMP Mon Dec 4 23:52:40 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

 Alertmanager version:latest

 Prometheus version:2.0
 WeChat:latest
Alertmanager configuration file:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: alertmanager
  namespace: default
data:
  config.yml: |-
    global:
      smtp_smarthost: 'smtp.qq.com:25'
      smtp_from: '***********@qq.com'
      smtp_auth_username: '1027155512@qq.com'
      smtp_auth_password: '*************'
      wechat_api_secret: '********************************'
      wechat_api_corp_id: '**************************'
    templates:
      - '/etc/alert/default.tmpl'
    route:
      receiver: 'default-receiver'
      group_by: ['alertname', 'cluster', 'service']
      group_wait: 30s
      group_interval: 5m
      repeat_interval: 15m
    receivers:
    - name: 'email_alert'
      email_configs:
      - to: '***********@139.com'
    - name: 'default-receiver'
      email_configs:
      - to: '***********@139.com'
      wechat_configs:
      - agent_id: '1000002'
        to_user: '*******'
        to_party: '2'
        corp_id: '**********'
```
my alertmanager on k8s 1.8.1 . when i use wechat alarm, i can`t  receive any message,  then, I found the following error through the dashboard
```
level=debug ts=2018-01-30T09:20:46.481211313Z caller=impl.go:910 msg="response: {\"errcode\":41011,\"errmsg\":\"missing agentid, hint: [1517304046_5_3f158368fdce3d37301495e4190372f1], more info at https://open.work.weixin.qq.com/devtool/query?e=41011\"}" incident="{}:{alertname=\"node_memory_low\"}"
```
but if i change this place ,all every look likes ok

notify/impl.go

before
```
        default:
                msg = &weChatCreateMessage{
                        Text: weChatMessage{
                                Content: tmpl(n.conf.Message),
                        },
                        ToUser:  tmpl(n.conf.ToUser),
                        ToParty: tmpl(n.conf.ToParty),
                        Totag:   tmpl(n.conf.ToTag),
                        AgentID: tmpl(n.conf.AgentID),
                        Type:    "text",
                        Safe:    "0",
```
after
```  
      default:
[...]

PR #1229, opened 2018-02-07
Fix WeChat issue

songjiayang:
For issue #1218

Changes:

1. change `weChatCreateMessage` and `weChatCloseMessage ` to  `weChatMessage `
2. cache WeChat AccessToken with` accessTokenAt`
3. code refactor & format.

Test config:
```
global:
  resolve_timeout: 5m
  wechat_api_secret: 'xxxx'

route:
  group_by: ['alertname']
  receiver: 'wechat'

receivers:
- name: 'wechat'
  wechat_configs:
  - corp_id: 'xxxx'
    to_party: '1'
    agent_id: 'xxx'
```
 Work with this config.

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -102,13 +102,11 @@ var (
 			VSendResolved: true,
 		},
 		Message:   `{{ template "wechat.default.message" . }}`,
-		APIURL:    `{{ template "wechat.default.api_url" . }}`,
 		APISecret: `{{ template "wechat.default.api_secret" . }}`,
 		ToUser:    `{{ template "wechat.default.to_user" . }}`,
 		ToParty:   `{{ template "wechat.default.to_party" . }}`,
 		ToTag:     `{{ template "wechat.default.to_tag" . }}`,
 		AgentID:   `{{ template "wechat.default.agent_id" . }}`,
-		// TODO: Add a details field with all the alerts.
 	}
 
 	// DefaultVictorOpsConfig defines default values for VictorOps configurations.
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -823,7 +823,7 @@ type weChatMessageContent struct {
 	Content string `json:"content"`
 }
 
-type weChatErrorResponse struct {
+type weChatResponse struct {
 	Code  int    `json:"code"`
 	Error string `json:"error"`
 }
@@ -849,11 +849,8 @@ func (n *Wechat) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 		return false, err
 	}
 
-	var accessToken string
 	// Refresh AccessToken over 2 hours
-	if n.accessToken != "" && time.Now().Sub(n.accessTokenAt) < 110*time.Minute {
-		accessToken = n.accessToken
-	} else {
+	if n.accessToken == "" || time.Now().Sub(n.accessTokenAt) > 2*time.Hour {
 		parameters := url.Values{}
 		parameters.Add("corpsecret", tmpl(string(n.conf.APISecret)))
 		parameters.Add("corpid", tmpl(string(n.conf.CorpID)))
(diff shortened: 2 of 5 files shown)


notify/impl.go
```diff
@@ -842,75 +839,86 @@ func (n *Wechat) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
...
-			ToUser:  tmpl(n.conf.ToUser),
-			ToParty: tmpl(n.conf.ToParty),
-			Totag:   tmpl(n.conf.ToTag),
-			AgentID: tmpl(n.conf.AgentID),
-			Type:    "text",
-			Safe:    "0",
+
+	var accessToken string
+	// Refresh AccessToken over 2 hours
+	if n.accessToken != "" && time.Now().Sub(n.accessTokenAt) < 110*time.Minute {
```
stuartnelson3: You can change `110*time.Minute` to `2*time.Hour`

notify/impl.go
```diff
@@ -842,75 +839,86 @@ func (n *Wechat) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
...
+		apiURL := n.conf.APIURL + "gettoken"
+
+		u, err := url.Parse(apiURL)
+		if err != nil {
+			return false, err
+		}
+
+		u.RawQuery = parameters.Encode()
+
+		level.Debug(n.logger).Log("msg", "Sending Wechat  message", "incident", key, "url", u.String())
```
stuartnelson3: `s/Sending/sending/`
There is an extra space between `Wechat` and `message` that can be removed.

notify/impl.go
```diff
@@ -842,75 +839,86 @@ func (n *Wechat) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
...
+
+		resp, err := ctxhttp.Get(ctx, http.DefaultClient, u.String())
```
stuartnelson3: Can you replace this with
```go
res, err := http.NewRequest("GET", u.String(), nil)
if err != nil {
        return true, err
}

resp, err := http.DefaulClient.Do(req.WithContext(ctx))
```

I want to start removing the `ctxhttp` library :)

notify/impl.go
stuartnelson3: This could be changed to

```go
if n.accessToken == "" || time.Now().Sub(n.accessTokenAt) > 2*time.Hour {
        // refresh token
        // Set n.accessToken and n.AccessTokenAt
}

// Down below, use n.accessToken directly
```

notify/impl_test.go
```diff
@@ -10,12 +10,12 @@ import (
...
 
-	"github.com/prometheus/common/model"
-	"github.com/prometheus/alertmanager/types"
 	"github.com/prometheus/alertmanager/config"
 	"github.com/prometheus/alertmanager/template"
-	"net/url"
+	"github.com/prometheus/alertmanager/types"
+	"github.com/prometheus/common/model"
 	"io/ioutil"
+	"net/url"
```
stuartnelson3: Both `"io/ioutil"` and `"net/url"` should be in the first block of imports in this list, with the other stdlib imports.

stuartnelson3:
Thanks for the contribution! Few easy changes and then this should be good to be merged

songjiayang:
Show Demo:
img_1829

config/notifiers.go
```diff
@@ -101,14 +101,8 @@ var (
 		NotifierConfig: NotifierConfig{
 			VSendResolved: true,
 		},
-		Message:   `{{ template "wechat.default.message" . }}`,
-		APIURL:    `{{ template "wechat.default.api_url" . }}`,
-		APISecret: `{{ template "wechat.default.api_secret" . }}`,
-		ToUser:    `{{ template "wechat.default.to_user" . }}`,
-		ToParty:   `{{ template "wechat.default.to_party" . }}`,
-		ToTag:     `{{ template "wechat.default.to_tag" . }}`,
-		AgentID:   `{{ template "wechat.default.agent_id" . }}`,
```
stuartnelson3: Why were all these fields removed?

config/notifiers.go
```diff
@@ -101,14 +101,8 @@ var (
...
-		// TODO: Add a details field with all the alerts.
+		Message: `{{ template "wechat.default.message" . }}`,
+		APIURL:  DefaultGlobalConfig.WeChatAPIURL,
```
stuartnelson3: It would be good to leave this as a templated value for users. There is the check in https://github.com/prometheus/alertmanager/pull/1229/files#diff-3baf47c64847a8fb8aaa8cc2e088513bR250 that will set the api url to the global default.

config/notifiers.go
songjiayang: I think these fields should be a determined value.

config/notifiers.go
songjiayang: ok, I will check it again.

songjiayang:
fixed

PR #1232, opened 2018-02-07
*: move to memberlist for clustering

fabxc:
@brancz working in manual testing. Some tests are failing now though – will debug that in the meantime.

It was quite smooth in general and our code actually got a lot simpler. The only friction was that memberlist doesn't natively support multiple independent "data channels". So that is implemented manually in `cluster` now.

This now also allows to pass DNS names as cluster peers, which will then be resolved for easier cluster initialization.


silence/silence.go
```diff
@@ -671,252 +663,149 @@ func (s *Silences) query(q *query, now time.Time) ([]*pb.Silence, error) {
...
 
-var _ mesh.GossipData = &gossipData{}
+type state map[string]*pb.MeshSilence
 
-func newGossipData() *gossipData {
-	return &gossipData{
-		data: silenceMap{},
+func (s state) clone() state {
+	c := make(state, len(s))
+	for k, v := range s {
```
brancz: I can't tell from this PR, but if the caller is not protecting this call, then it seems like this needs a read lock

api/api.go
```diff
@@ -228,32 +235,21 @@ type connectionStatus struct {
...
-	if api.mrouter == nil {
+	if api.peer == nil {
 		return nil
 	}
 
-	status := mesh.NewStatus(api.mrouter)
 	strippedStatus := &meshStatus{
-		Name:        status.Name,
-		NickName:    status.NickName,
-		Peers:       make([]peerStatus, len(status.Peers)),
```
brancz: I think one of these was rendered in the frontend, so we need to make sure that is either removed or replaced with something similar

cluster/cluster.go
```diff
@@ -0,0 +1,437 @@
...
+		return nil, err
+	}
+
+	p := &Peer{
+		states: map[string]State{},
+		stopc:  make(chan struct{}),
+	}
+	p.delegate = newDelegate(l, reg, p)
+
+	cfg := memberlist.DefaultLANConfig()
```
brancz: I don't remember entirely how to configure this, but memberlist exposes a bunch of metrics, but I think it has to be configured so it registers them with the Prometheus metrics registry. See: https://github.com/armon/go-metrics#sinks

cluster/cluster.go
```diff
@@ -0,0 +1,437 @@
...
+func (p *Peer) warnIfAlone(logger log.Logger, d time.Duration) {
+	tick := time.NewTicker(d)
+	defer tick.Stop()
+
+	for {
+		select {
+		case <-p.stopc:
+			return
+		case <-tick.C:
+			if n := p.mlist.NumMembers(); n <= 1 {
```
brancz: I would argue this is a readiness condition if HA/gossiping is enabled

cluster/cluster.go
```diff
@@ -0,0 +1,437 @@
...
+	gossipMsgsReceived prometheus.Counter
+}
+
+func newDelegate(l log.Logger, reg prometheus.Registerer, p *Peer) *delegate {
+	bcast := &memberlist.TransmitLimitedQueue{
+		NumNodes:       p.ClusterSize,
+		RetransmitMult: 3,
+	}
+	gossipMsgsReceived := prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "thanos_gossip_messages_received_total",
```
brancz: thanos -> alertmanager

cluster/cluster.go
```diff
@@ -0,0 +1,437 @@
...
+		Help: "Total gossip NotifyMsg calls.",
+	})
+	gossipClusterMembers := prometheus.NewGaugeFunc(prometheus.GaugeOpts{
+		Name: "thanos_cluster_members",
```
brancz: thanos -> alertmanager

silence/silence.go
fabxc: It's just some sugar for a map – so yea, the caller has to lock.

cluster/cluster.go
fabxc: memberlist just writes to a global registry though :(

cluster/cluster.go
fabxc: Readyness in what sense? For a readyness endpoint, sure. Not for blocking actually though.
Since AM doesn't seem to have a readiness endpoint still, I guess that's for another PR.

cluster/cluster.go
fabxc: After looking a bit more I'm against actually using their metrics. I doubt they are that useful for the average user and for the important ones we may be better of using our own ones.

The config has quite a lot of paramters which all scream that nothing about this is in line with Prometheus's semantics: https://github.com/armon/go-metrics/blob/master/start.go#L48
It may expose something well-formatted if configured right but it's probably going to be quite confusing semantically.

api/api.go
simonpasquier: @brancz you're correct. It was added by #1164.

cluster/cluster.go
brancz: Right I had a feeling like that, I stumbled upon the same findings last time I looked into this. Let's add our own metrics as applicable.

brancz: One on "bytes_received/sent_total" seems useful though.

cluster/cluster.go
brancz: Just discussed this offline and decided that this is better served with a metric/alert, as otherwise we have a problem with the initial peer. Plus there is no readiness handler yet anyways.

Procfile
```diff
@@ -1,5 +1,5 @@
-a1: ./alertmanager --log.level=debug --storage.path=$TMPDIR/a1 --web.listen-address=:9093 --mesh.peer-id=00:00:00:00:00:01 --mesh.nickname=a --mesh.listen-address=:8001 --config.file=examples/ha/alertmanager.yaml
-a2: ./alertmanager --log.level=debug --storage.path=$TMPDIR/a2 --web.listen-address=:9094 --mesh.peer-id=00:00:00:00:00:02 --mesh.nickname=b --mesh.listen-address=:8002 --mesh.peer=127.0.0.1:8001 --config.file=examples/ha/alertmanager.yaml
-a3: ./alertmanager --log.level=debug --storage.path=$TMPDIR/a3 --web.listen-address=:9095 --mesh.peer-id=00:00:00:00:00:03 --mesh.nickname=c --mesh.listen-address=:8003 --mesh.peer=127.0.0.1:8001 --config.file=examples/ha/alertmanager.yaml
+a1: ./alertmanager --log.level=debug --storage.path=$TMPDIR/a1 --web.listen-address=:9093  --cluster.address=127.0.0.1:8001 --config.file=examples/ha/alertmanager.yaml
```
brancz: Do you think we should keep the flags, for backward compatibility?

fabxc: Mh, we could add a hidden flag with the old name. But some flags were removed entirely – we could just blackhole these but then people will have mis-aligned expectations and we might get complaints.

In particular I removed the ability to set a password. Memberlist provides similar functionality – but our general policy on security features is that we aren't able to support them. I felt this is a similar case and it would be better to be consistent. /cc @brian-brazil 

Ultimately, people have to switch anyway. As it's not a 1:1 mapping, maybe it's better to be explicit and just fail hard?

brancz: Makes sense. In that case I'm definitely for a beta/rc release though, just so people get the notion that larger things are changing here.

brian-brazil: I'm a little unclear exactly what the change is here from the PR description, if we have to break I'm generally in favour of hard breaks as it's easier to debug.

For consistency we should keep away from security features and I'm unsure exactly what properties this password offered. When Ben's TLS plans come to fruition we can reconsider, which will be a different approach anyway.

cmd/alertmanager/main.go
```diff
@@ -119,11 +125,18 @@ func main() {
 		routePrefix   = kingpin.Flag("web.route-prefix", "Prefix for the internal routes of web endpoints. Defaults to path of --web.external-url.").String()
 		listenAddress = kingpin.Flag("web.listen-address", "Address to listen on for the web interface and API.").Default(":9093").String()
 
-		meshListen = kingpin.Flag("mesh.listen-address", "mesh listen address. Pass an empty string to disable.").Default(net.JoinHostPort("0.0.0.0", strconv.Itoa(mesh.Port))).String()
-		hwaddr     = kingpin.Flag("mesh.peer-id", "mesh peer ID").Default(mustHardwareAddr()).String()
-		nickname   = kingpin.Flag("mesh.nickname", "mesh peer nickname").Default(mustHostname()).String()
-		password   = kingpin.Flag("mesh.password", "password to join the peer network (empty password disables encryption)").Default("").String()
-		peers      = kingpin.Flag("mesh.peer", "initial peers (may be repeated)").Strings()
+		clusterBindAddr = kingpin.Flag("cluster.address", "listen address for cluster").
```
brian-brazil: cluster.listen.address would be clearer, this current reads as the cluster to talk to

cmd/alertmanager/main.go
```diff
@@ -98,6 +102,8 @@ func newMarkerMetrics(marker types.Marker) {
 	prometheus.MustRegister(alertsSuppressed)
 }
 
+const defaultClusterAddr = "0.0.0.0:19093"
```
brian-brazil: 9094 is the next free port in the core components range

fabxc:
Added message size metrics and dimension that distinguishes between update and full state exchange messages.

@stuartnelson3 updated the API response for the cluster stauts. Just needs integration with the UI still.

Procfile
fabxc: Basically it's about encrypting the peer-to-peer traffic with a secret. FWIW, I don't think that's really needed for a while.

brancz: Any idea what that might look like? Seems like TLS is not really a natural fit for the way memberlist communication works. Maybe we should think this through a bit more just to make sure we're not setting ourselves road blocks for the future in this regard.

brian-brazil: Encryption isn't much use without signing. TLS is fine for it I'd think, every cluster endpoint would have a cert signed by some CA and they'd be checked at at connection time. ServerName might be a bit interesting here if we only have IPs (not so much a problem if we use service discovery), but it's in line with the way @SuperQ and I have been talking about TLS.

fabxc: In mesh and memberlist alike it is a builtin feature – we can take it or leave it

stuartnelson3:
@fabxc updated

`--cluster.address=""`
image

with `goreman start`
image

fabxc:
Nice, thanks! Can you resolve the bindata conflict?

Procfile
brian-brazil: It's not in line with future plans (and a secret on the command line is a bad idea anyway), so I'd say remove it while we're breaking things.

stuartnelson3:
there were recent changes to bindata for improving rendering silences. rebased on top of master, but now github is saying i'm a non-author comitter on all your commits. sorry about that :/

cmd/alertmanager/main.go
fabxc: I felt like this was mostly for user-visible stuff. In mesh we just used the default as well.

brian-brazil: The default port list has expanded quite a bit since we first created it for a handful of official components. All our ports are there, including things like statsd ingestion so all things being equal we should stick to it.

stuartnelson3:
When do you think this will be ready to merge into a release client? I would like to do one release before we start that process.

brancz:
This seems very close to finished to me. Once @brian-brazil's comments are resolved this lgtm.

stuartnelson3:
Cool, I'll put out an AM release today then

fabxc:
Addressed all comments. @stuartnelson3 bindata is having conflicts again though.

stuartnelson3:
@fabxc just gotta rebase on master and then regenerate it

brancz:
let's do this

Procfile
giganteous: > It's not in line with future plans (and a secret on the command line is a bad idea anyway), so I'd say remove it while we're breaking things.

I disagree: A secret on the commandline is more secure than an unencrypted message that is flowing unencrypted between 2 datacenters.

For the next feature that you drop: please put it in the release notes. This was hard to find.

PR #1233, opened 2018-02-07
Fixes AM wrongly counting alerts with EndTimes in the future as resolved

Conorbro:
As far as I can tell Prometheus may send a time stamp in the future for the end time of an alert meaning AM shouldn't automatically mark this alert.


brancz:
lgtm

PR #1238, opened 2018-02-13
api: support zero StartsAt for alerts

### LINKED ISSUE #1191: Alertmanager incorrectly handling newly resolved alert default startTime when zero
This is a continuation from my findings on https://github.com/prometheus/prometheus/issues/1871

**What did you do?**

Sent following request to the AM where the `startTime` is zero and the `endTime` is an arbitrary time in the past:

Endpoint: `http://127.0.0.1:9093/api/v1/alerts`

```
[
  {
    "labels": {
      "alertname": "InstanceUp",
      "instance": "127.0.0.1:9115",
      "job": "node",
      "monitor": "test_prod",
      "severity": "page"
    },
    "annotations": {
      "description": "127.0.0.1:9115 of job node has been down for more than 5 minutes.",
      "summary": "Instance 127.0.0.1:9115 down"
    },
    "startsAt": "0001-01-01T00:00:00Z",
    "endsAt": "2018-01-10T12:42:30.584249Z",
    "generatorURL": "http://Conors-MacBook-Pro-3.local:9090/graph?g0.expr=up+%3D%3D+0\\u0026g0.tab=1"
  }
]
```

What seems to be happening:

1. The endTime is set to the current time and the [startTime set to zero in Prometheus](https://github.com/prometheus/prometheus/blob/master/rules/alerting.go#L269).

2. The Alertmanager then [sets the received alert's startTime to the current time](https://github.com/prometheus/alertmanager/blob/master/api/api.go#L488) which will be ahead of the endTime and result in the "start time must be before end time" error.

Looking at the problem it doesn't make sense to alter the start or end times when this occurs. Perhaps we should only set a default start time when the the start time is both zero and not resolved?

```
{
    "status": "success"
}
```

```
{
    "status": "error",
    "errorType": "bad_data",
    "error": "start time must be before end time"
}
```

**Environment**

* System information:

	Darwin 17.3.0 x86_64

* Alertmanager version:

	HEAD

* Prometheus version:

	HEAD

 N/A

 N/A

* Logs:

`level=error ts=2018-01-10T17:35:57.233839Z caller=api.go:803 msg="API error" err="bad_data: start time must be before end time"`

simonpasquier:
Fixes #1191.

cc @Conorbro @stuartnelson3 @brian-brazil

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -103,6 +103,10 @@ func New(
 	peer *cluster.Peer,
 	l log.Logger,
 ) *API {
+	if l == nil {
+		l = log.NewNopLogger()
+	}
+
 	return &API{
 		alerts:         alerts,
 		silences:       silences,
@@ -466,7 +470,11 @@ func (api *API) insertAlerts(w http.ResponseWriter, r *http.Request, alerts ...*
 
 		// Ensure StartsAt is set.
 		if alert.StartsAt.IsZero() {
-			alert.StartsAt = now
+			if alert.EndsAt.IsZero() {
+				alert.StartsAt = now
+			} else {
+				alert.StartsAt = alert.EndsAt
+			}
 		}
 		// If no end time is defined, set a timeout after which an alert
 		// is marked resolved if it is not updated.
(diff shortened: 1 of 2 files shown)


brian-brazil:
:+1:

PR #1242, opened 2018-02-13
cluster: Make peer timeout configurable

roidelapluie:


fabxc:
Yes, this makes sense since the configurable gossip/push-pull interval will impact message propagation speed.

There's of course a [bunch of other internal timings](https://github.com/hashicorp/memberlist/blob/master/config.go#L220-L278) in the memberlist library. I hope we get reasonably far with those that we provide as flags here though.

PR #1251, opened 2018-02-16
Remove unused functions for mesh

stuartnelson3:
These functions were used with weaveworks/mesh,
but are no longer needed with memberlist.


brian-brazil:
:+1:

PR #1256, opened 2018-02-22
Fix miscellaneous issues revealed by Go 1.10

simonpasquier:
Spotted at https://travis-ci.org/prometheus/alertmanager/builds/344780795#L594


simonpasquier:
Travis CI now runs with Go 1.10. And [by default](https://golang.org/doc/go1.10#test) Go 1.10 runs `go vet` on the package being tested which highlighted a couple of issues.

cc @brian-brazil @stuartnelson3

brian-brazil:
Thanks!

PR #1264, opened 2018-02-28
Run staticcheck as part of the build process

simonpasquier:
This change also fixes potential issues highlighted by running
staticcheck.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,9 @@
 # limitations under the License.
 
 GO    := GO15VENDOREXPERIMENT=1 go
-PROMU := $(GOPATH)/bin/promu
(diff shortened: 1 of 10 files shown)


stuartnelson3:
build seems to be failing on the makefile

simonpasquier:
Apparently Circle CI sets the GOPATH with [multiple directories](https://discuss.circleci.com/t/gopath-is-set-to-multiple-directories/7174). To fix it, I've applied the [trick](https://github.com/prometheus/prometheus/blob/master/Makefile#L19) from  the Prometheus's Makefile.

stuartnelson3:
Very cool. I'm actually not familiar with staticcheck, but it seems to be done here in the same way as in prometheus.

PR #1279, opened 2018-03-08
Set User object in alertmanager url.

### LINKED ISSUE #1028: amtool basic auth support
Currently if you put alertmanager behind a basic http authentication url amtool is unable to specify credentials (at least looking at https://github.com/prometheus/alertmanager/blob/master/cli/alert.go)
Is it possible for amtool to support basic auth in the config file / flags ?

limscoder:
Sets User object in alertmanager url to enable amtool to work with basic auth for use with proxies. Fixes #1028.

### DIFF
--- a/cli/utils.go
+++ b/cli/utils.go
@@ -27,11 +27,9 @@ func (s ByAlphabetical) Less(i, j int) bool {
 }
 
 func GetAlertmanagerURL(p string) url.URL {
-	return url.URL{
-		Scheme: (*alertmanagerUrl).Scheme,
-		Host:   (*alertmanagerUrl).Host,
-		Path:   path.Join((*alertmanagerUrl).Path, p),
-	}
+	amURL := **alertmanagerUrl
+	amURL.Path = path.Join((*alertmanagerUrl).Path, p)
+	return amURL
 }
 
 // Parse a list of labels (cli arguments)


stuartnelson3:
thanks for the contribution! looking at this code, it seems to me that we just want to copy the amURL we're getting from the config, and then update the path. Can you check to see if this code works for you?

```diff
diff --git a/cli/utils.go b/cli/utils.go
index 256c677..0828373 100644
--- a/cli/utils.go
+++ b/cli/utils.go
@@ -27,11 +27,9 @@ func (s ByAlphabetical) Less(i, j int) bool {
 }
 
 func GetAlertmanagerURL(p string) url.URL {
-	return url.URL{
-		Scheme: (*alertmanagerUrl).Scheme,
-		Host:   (*alertmanagerUrl).Host,
-		Path:   path.Join((*alertmanagerUrl).Path, p),
-	}
+	amURL := **alertmanagerUrl
+	amURL.Path = path.Join((*alertmanagerUrl).Path, p)
+	return amURL
 }
 
 // Parse a list of labels (cli arguments)
```

limscoder:
@stuartnelson3 I'm not familiar with the `**` operator. Does that copy a struct?

stuartnelson3:
`alertmanagerUrl` is a reference to a reference (`**url.URL`), so the double `**` de-references to the actual struct and makes a copy. Check out the playground:
https://play.golang.org/p/MjVJheu_7vQ

limscoder:
@stuartnelson3 that code works for me, and makes sense to copy the whole thing instead of selected properties. I updated the PR.

stuartnelson3:
cool. the code worked for me in my tests, can you double check that it works for you with your basic auth setup?

limscoder:
Yes, it works for me with basic auth

PR #1287, opened 2018-03-16
The default group_by is meant to be no labels.

brian-brazil:
This is what the intended default is, and what
the documentation says.

@stuartnelson3


simonpasquier:
Closes https://github.com/prometheus/alertmanager/issues/1288

stuartnelson3:
@w0rm are there other places in the code that make special assumptions about a default grouping of `alertname` we need to do something about?

w0rm:
Why do we change the default UI behavior? As far as I remember it was a conscious decision to group alerts by alertname by default to improve usability.

brian-brazil:
I don't see that we have to have the UI in sync with this default.

stuartnelson3:
Ah my mistake, sorry about that

Issue #1240, opened 2018-02-13
GUI: new silence form without matchers


r0bj:
**What did you do?**

There is active alert `MySQLReplicationNotRunning`:
<img width="1121" alt="screen shot 2018-02-13 at 14 54 08" src="https://user-images.githubusercontent.com/1297747/36153223-d54b9f22-10cd-11e8-8efd-4122acea50c2.png">

I'm clicking `Silence` next to this alert:
```
http://alertmanager.example.com/#/silences/new?filter=%7Bseverity%3D%22critical%22%2C%20service%3D%22mysqld-exporter-db-centities-s1%22%2C%20prometheus_cluster%3D%22prometheus-ops%22%2C%20pod%3D%22mysqld-exporter-db-centities-s1-86db8744bf-nlxmb%22%2C%20namespace%3D%22ops%22%2C%20master_uuid%3D%228efcb402-5429-11e6-b7aa-00163e3acc15%22%2C%20master_host%3D%22db-centities-s2%22%2C%20job%3D%22db-centities-s1%22%2C%20instance%3D%2210.200.0.125%3A9104%22%2C%20endpoint%3D%22http-metrics%22%2C%20connection_name%3D%22%22%2C%20channel_name%3D%22%22%2C%20alertname%3D%22MySQLReplicationNotRunning%22%7D
```

New silence form is opened with empty `Matchers Alerts affected by this silence.`:
<img width="1191" alt="screen shot 2018-02-13 at 14 57 39" src="https://user-images.githubusercontent.com/1297747/36153375-54baf550-10ce-11e8-8de7-2dd2efd1f16b.png">

**What did you expect to see?**

New silence form is opened with filled matchers, eg:
<img width="1183" alt="screen shot 2018-02-13 at 14 59 17" src="https://user-images.githubusercontent.com/1297747/36153430-82050d70-10ce-11e8-8f0f-f2b207e515eb.png">

**What did you see instead? Under which circumstances?**

New silence form is opened with empty `Matchers Alerts affected by this silence.`

**Environment**

* Alertmanager version: 0.14.0
* Prometheus version: 2.0.0
* Alertmanager configuration: generated by prometheus-operator
* Prometheus configuration file: generated by prometheus-operator

roidelapluie:
I can not reproduce this (Alertmanager version: 0.14.0 too)

stuartnelson3:
[...]

PR #1289, opened 2018-03-16
Allow empty matchers

w0rm:
@stuartnelson3 I changed the parser to allow empty matchers, I also removed the validation rule for empty matcher values in the silence form. Right now I am getting errors from the backend when I submit empty matchers:

screen shot 2018-03-16 at 11 16 53

screen shot 2018-03-16 at 11 17 31

Relates to #1240


stuartnelson3:
Awesome, thanks for doing this. The backend errors are expected, and will be fixed now that I know about the correct empty matcher behavior. I'll add that to this PR.

stuartnelson3:
@w0rm backend code should be working now. I think we can remove the `stringNotEmpty` validation on the matcher values in the front end on the silence form, since that's preventing creation of silences with empty matchers. Right?

w0rm:
@stuartnelson3 I've already done that in my pr. Check the changes to ` ui/app/src/Views/SilenceForm/Types.elm`

stuartnelson3:
we're inconsistent in how we are searching for alerts between the alert list view and the alert preview on the silence form. When updating the filter bar in the alert list view, the filter query string has curly braces `filter={foo%3D"bar"}`, while on the silence form there are no curly braces `filter=foo="bar",baz=""`. Would be good to make this consistent (trying to look now)

stuartnelson3:
Changes also needed to be made here:

```diff
diff --git a/ui/app/src/Views/SilenceForm/Types.elm b/ui/app/src/Views/SilenceForm/Types.elm
index da09c31..e098a10 100644
--- a/ui/app/src/Views/SilenceForm/Types.elm
+++ b/ui/app/src/Views/SilenceForm/Types.elm
@@ -204,7 +204,7 @@ fromMatchersAndTime defaultCreator matchers now =
 appendMatcher : MatcherForm -> Result String (List Matcher) -> Result String (List Matcher)
 appendMatcher { isRegex, name, value } =
     Result.map2 (::)
-        (Result.map2 (Matcher isRegex) (stringNotEmpty name.value) (stringNotEmpty value.value))
+        (Result.map2 (Matcher isRegex) (stringNotEmpty name.value) (Ok value.value))
 
 
 filterMatcherToMatcher : Utils.Filter.Matcher -> Maybe Matcher
diff --git a/ui/app/src/Views/SilenceForm/Updates.elm b/ui/app/src/Views/SilenceForm/Updates.elm
index 137a5ff..0113f3b 100644
--- a/ui/app/src/Views/SilenceForm/Updates.elm
+++ b/ui/app/src/Views/SilenceForm/Updates.elm
@@ -158,7 +158,7 @@ updateForm msg form =
             let
                 matchers =
                     Utils.List.replaceIndex index
-                        (\matcher -> { matcher | value = validate stringNotEmpty matcher.value })
+                        (\matcher -> { matcher | value = matcher.value })
                         form.matchers
             in
                 { form | matchers = matchers }
```

w0rm:
@stuartnelson3 oh, sorry, I missed that. Thanks!

api/api.go
```diff
@@ -695,11 +705,31 @@ func matchFilterLabels(matchers []*labels.Matcher, sms map[string]string) bool {
 	for _, m := range matchers {
 		v, prs := sms[m.Name]
 		switch m.Type {
-		case labels.MatchNotEqual, labels.MatchNotRegexp:
+		case labels.MatchNotRegexp:
```
stuartnelson3: I couldn't think of a "clever" way to do this, and didn't want to nest a bunch of if statements. Any suggestions are welcome, but this is pretty clear to read so I'm ok with it.

stuartnelson3:
@brancz do you have a chance to review the go code in this PR?

api/api.go
```diff
@@ -695,11 +705,31 @@ func matchFilterLabels(matchers []*labels.Matcher, sms map[string]string) bool {
...
+			if string(m.Value) == "^(?:)$" && prs {
```
simonpasquier: The `string(m.Value) == "^(?:)$"` test seems brittle to me. Also this will break if we update the vendored `github.com/prometheus/prometheus/pkg/labels` package because in the latest version, the original `m.Value` is kept (no anchoring added, see [this](https://github.com/prometheus/prometheus/blob/5b962c57489f6995f24c56fc4ac8b97faec753fd/pkg/labels/matcher.go#L55-L69)). Maybe updating `github.com/prometheus/prometheus/pkg/labels` and matching againt `""` then?

stuartnelson3: yeah, 100%. Good to hear that the latest version doesn't change `m.Value`, will update that and simplify this.

stuartnelson3:
@simonpasquier if you have time, feel free to review the code further :)

simonpasquier:
LGTM

Issue #1297, opened 2018-03-23
Make `amtool` code modular


stuartnelson3:
The code in the `cli` package is using entirely global variables.

Following the example here: https://github.com/alecthomas/kingpin/blob/master/_examples/modular/main.go

Turn the code in `cli/` into a modular cli application. Each sub command should have its own struct storing the values passed to the command's flags.

This will make it easy to contribute new subcommands, as well as test the different pieces.

stuartnelson3:
closed by #1321

PR #1298, opened 2018-03-23
[amtool] update silence add and update flags

stuartnelson3:
- Change --expires/-e to --duration/-d
- Change --expires-on to --end
- Add --start
- `silence update` has no default duration value
- `silence update` now returns the ID of the newly created silence, instead of printing the ID for the old silence.

### DIFF
--- a/cli/silence_update.go
+++ b/cli/silence_update.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"net/http"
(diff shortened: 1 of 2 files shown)


stuartnelson3:
@sinkingpoint I think you folks are the biggest users of `amtool`. Do any of the changes here (other than updating flag names) help/hurt your workflows?

stuartnelson3:
@brian-brazil @simonpasquier any comments/objections? my plan is to do #1297 afterwards and have automated tests. So far I've only manually tested this code.

brian-brazil:
Sounds fine by me.

cli/silence_update.go
```diff
@@ -98,30 +100,43 @@ func getSilenceById(silenceId string, baseUrl url.URL) (*types.Silence, error) {
...
+	} else if *updateDuration != "" {
+		d, err := model.ParseDuration(*updateDuration)
+		if err != nil {
+			return nil, err
+		}
+		if d == 0 {
 			return nil, fmt.Errorf("silence duration must be greater than 0")
 		}
-		silence.EndsAt = time.Now().UTC().Add(time.Duration(duration))
+		silence.EndsAt = silence.EndsAt.UTC().Add(time.Duration(d))
```
simonpasquier: It is now adding the passed duration to the existing `EndsAt`. I suppose this is intentional but it wasn't obvious from the change description.

sinkingpoint:
Sorry for the late reply. This is fine by us, but is there a justification for a breaking change like this?

stuartnelson3:
It more closely maps to the field names used in the alertmanager frontend, and the flag name for the starting time `--starts` (or `--begin`) seemed to pair in my mind better with `--ends` than `--expires-on`. This was entirely subjective though.

cli/silence_update.go
stuartnelson3: Hm, yeah, the original behavior was confusing to me. Now that I'm looking at this again, I'm thinking the duration should be added to `StartsAt`...

With `amtool update <id> --start=<time> --duration=2h`, I would expect this silence to be active for 2 hours after the start time.
With `amtool update <id> --duration=2h` I think it's less clear what the behavior should be. I would take this to mean "set the duration to `start_time+2h`, but I'm not sure if other folks agree?

simonpasquier: Yeah my gut feeling is that `amtool update <id> --duration=<x>` is ambiguous for users whichever time reference is chosen. Using `StartsAt` is probably the less confusing option but this is just my opinion ;-)

PR #1309, opened 2018-04-03
inhibit: update inhibition cache when alerts resolve

simonpasquier:
@stuartnelson3 this is the fix for #1153.

I've added another acceptance scenario on inhibition where inhibiting and inhibited alerts fire and resolve together.

### DIFF
--- a/inhibit/inhibit.go
+++ b/inhibit/inhibit.go
@@ -42,9 +42,6 @@ type Inhibitor struct {
 
 // NewInhibitor returns a new Inhibitor.
 func NewInhibitor(ap provider.Alerts, rs []*config.InhibitRule, mk types.Marker, logger log.Logger) *Inhibitor {
-	if logger == nil {
-		logger = log.NewNopLogger()
-	}
 	ih := &Inhibitor{
 		alerts: ap,
 		marker: mk,
--- a/inhibit/inhibit.go
+++ b/inhibit/inhibit.go
@@ -43,6 +43,9 @@ type Inhibitor struct {
 
 // NewInhibitor returns a new Inhibitor.
 func NewInhibitor(ap provider.Alerts, rs []*config.InhibitRule, mk types.Marker, logger log.Logger) *Inhibitor {
+	if logger == nil {
+		logger = log.NewNopLogger()
+	}
 	ih := &Inhibitor{
 		alerts: ap,
 		marker: mk,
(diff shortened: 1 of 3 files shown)


inhibit/inhibit.go
```diff
@@ -81,14 +84,9 @@ func (ih *Inhibitor) run(ctx context.Context) {
...
-			if a.Resolved() {
-				// As alerts can also time out without an update, we never
-				// handle new resolved alerts but invalidate the cache on read.
-				continue
-			}
-			// Populate the inhibition rules' cache.
+			// Update the inhibition rules' cache.
 			for _, r := range ih.rules {
-				if r.SourceMatchers.Match(a.Labels) {
+				if r.exists(a) || r.SourceMatchers.Match(a.Labels) {
```
stuartnelson3: what is the purpose of `r.exists(a)`? It seems like from the code, if it exists, then it gets set (which doesn't accomplish anything, since it already exists?)

Running the tests without this code still passes.

inhibit/inhibit.go
```diff
@@ -145,7 +143,7 @@ func (ih *Inhibitor) Mutes(lset model.LabelSet) bool {
 	for _, r := range ih.rules {
 		// Only inhibit if target matchers match but source matchers don't.
 		if inhibitedByFP, eq := r.hasEqual(lset); !r.SourceMatchers.Match(lset) && r.TargetMatchers.Match(lset) && eq {
-			ih.marker.SetInhibited(fp, fmt.Sprintf("%d", inhibitedByFP))
+			ih.marker.SetInhibited(fp, fmt.Sprintf("%s", inhibitedByFP.String()))
```
stuartnelson3: this can just be `ih.marker.SetInhibited(fp, inhibitedByFP.String())`

stuartnelson3: It also appears that the original code works as well?

simonpasquier: Done. Yes the original code works but the API returns different fingerprint formats between silenced and inhibited alerts:

```
      ...snip...
      "status": {
        "state": "suppressed",
        "silencedBy": [
          "95c9789c-f7f6-4d57-99d3-66c9eb25a4b8"
        ],
        "inhibitedBy": []
      },
      ...snip..
```

```
      ...snip...
      "status": {
        "state": "suppressed",
        "silencedBy": [],
        "inhibitedBy": [
          "12124264373735012735"
        ]
      },
      ...snip...
```

inhibit/inhibit.go
simonpasquier: IIUC the only case where `r.exists(a)` would be false is when AlertManager processes a resolved alert that hasn't been seen as firing previously (eg after a restart). Not sure it is worth to keep it then.

inhibit/inhibit.go
```diff
@@ -43,6 +42,9 @@ type Inhibitor struct {
 
 // NewInhibitor returns a new Inhibitor.
 func NewInhibitor(ap provider.Alerts, rs []*config.InhibitRule, mk types.Marker, logger log.Logger) *Inhibitor {
+	if logger == nil {
```
mxinden: @simonpasquier As far as I can tell, `logger` is only `nil` for tests, right? I would prefer to create a proper `logger` in the test code and pass it to `NewInhibitor` instead of bloating the real code here. What do you think?

simonpasquier: > logger is only nil for tests

yes. Agree with your recommendation.

stuartnelson3:
fixed in #1331

PR #1313, opened 2018-04-05
Provide default working config for artifacts

### LINKED ISSUE #484: simple.yml should be alertmanager.yml

stuartnelson3:
Fixes #484


brian-brazil:
:+1:

mxinden:
@stuartnelson3 @brian-brazil What is the reasoning for changing `WORKDIR` from `/alertmanager` to `/etc/alertmanager`? [Prometheus](https://github.com/prometheus/prometheus/blob/master/Dockerfile#L17) uses `/prometheus`. I would expect `etc` for configuration files, but not for alerting data. If not `/alertmanager`, maybe `/var/alertmanager`?

This came up here: https://github.com/prometheus/alertmanager/issues/1433

discordianfish:
Yeah that broke our setup too. It's also not mentioned in the changelog.

mxinden:
@discordianfish yes, sorry about the missing changelog entry. I fixed that and updated `CHANGELOG.md` in both release and master branch.

Once @stuartnelson3 is back, he can probably give us more details here for the initial change.

mxinden:
@stuartnelson3 ok. Let's move this discussion to https://github.com/prometheus/alertmanager/pull/1435 then.

PR #1314, opened 2018-04-05
Split cli package

simonpasquier:
This PR modifies the cli commands to use the newly added API client.

cc @stuartnelson3


cli/client.go
```diff
@@ -153,7 +154,7 @@ func (h *httpStatusAPI) Get(ctx context.Context) (*ServerStatus, error) {
 // AlertAPI provides bindings for the Alertmanager's alert API.
 type AlertAPI interface {
 	// List returns all the active alerts.
-	List(ctx context.Context) ([]*ExtendedAlert, error)
+	List(ctx context.Context, filter string, silenced bool, inhibited bool) ([]*ExtendedAlert, error)
```
stuartnelson3: `List(ctx context.Context, filter string, silenced, inhibited bool) ([]*ExtendedAlert, error)`

cli/client.go
```diff
@@ -193,8 +194,15 @@ type httpAlertAPI struct {
 	client api.Client
 }
 
-func (h *httpAlertAPI) List(ctx context.Context) ([]*ExtendedAlert, error) {
+func (h *httpAlertAPI) List(ctx context.Context, filter string, silenced bool, inhibited bool) ([]*ExtendedAlert, error) {
```
stuartnelson3: same as above, you only need to declare `bool` the one time

cli/cmd/silence_import.go
```diff
@@ -31,20 +36,29 @@ amtool silence import foo.json
 JSON data can also come from stdin if no param is specified.`
 }
 
+func addSilence(silence *types.Silence) (string, error) {
+	client, err := api.NewClient(api.Config{Address: (*alertmanagerUrl).String()})
```
stuartnelson3: I think most of this function can disappear. It makes sense to me to create the client + silence client once, and then pass the silence client into each `addSilenceWorker` function. A bulk import could potentially have thousands of silences, which in its current form would create thousands of clients when only one is needed.

With this, the line `silenceID, err := addSilence(s)` in `addSilenceWorker` could just be `silenceID, err := silenceAPI.Set(context.Background(), *silence)`

simonpasquier:
> It might make more sense to move those files back to cli/, and then move cli/client.go to its own client/ dir. I'm guessing that was the separation you wanted, between "cli stuff" and "the client that interacts with the api".

:+1: I'll update accordingly. Thanks for the other tips.

PR #1319, opened 2018-04-11
Rename silence API Delete() -> Expire()

stuartnelson3:
Within alertmanager, expire is the term used,
since silences still "exist" but aren't in effect.


simonpasquier:
LGTM except you need to sign your commit to please the DCO check ;-)

PR #1321, opened 2018-04-11
Move amtool to modular structure

stuartnelson3:
implements #1297


cli/alert.go
```diff
@@ -14,16 +14,21 @@ import (
...
+	matcherGroups     []string
+}
 
-func init() {
-	alertQueryCmd.Action(queryAlerts)
+func configureAlertCmd(app *kingpin.Application, longHelpText map[string]string) {
+	var (
+		a             = &alertQueryCmd{}
+		alertCmd      = app.Command("alert", "View and search through current alerts")
+		alertQueryCmd = alertCmd.Command("query", "View and search through current alerts").Default()
```
simonpasquier: I find it a bit confusing to have both `alertQueryCmd` as a local variable here and as a global struct.

cli/config_resolver.go
```diff
@@ -0,0 +1,57 @@
+package cli
```
simonpasquier: I've already noticed that most of the files in the `cli` package are missing the Apache 2 license header. This shouldn't be addressed in this PR but rather tackled once https://github.com/prometheus/prometheus/issues/3904 is fixed.

stuartnelson3:
@simonpasquier any more comments/fixes?

cli/check_config.go
```diff
@@ -8,23 +8,28 @@ import (
...
-)
+// TODO: This can just be a type that is []string, doesn't have to be a struct
+type checkConfigCmd struct {
+	files []string
+}
+
+func configureCheckConfigCmd(app *kingpin.Application, longHelpText map[string]string) {
+	var (
+		c              = &checkConfigCmd{}
+		checkConfigCmd = app.Command("check-config", "Validate alertmanager config files")
```
simonpasquier: Same confusing names with the struct declared at L12.

simonpasquier:
:+1:

PR #1324, opened 2018-04-14
Show AM health based on mesh status

stuartnelson3:
Conditionally return OK or NOT HEALTHY based on
whether the mesh is settled or not.

This is a coarse indicator. In the future, a
more detailed check of the health of an instance
can be implemented.

Discussed this briefly with @grobie 

This check wasn't added to the ready endpoint because a running alertmanager should always be ready to accept alerts (taking the kubernetes terminology), but health indicates its current running status.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -186,6 +186,12 @@ func (p *Peer) Ready() bool {
 	return false
 }
 
+// Healthy implements the web.healthIndicator interface. It is used to indicate
+// whether the mesh is settled or not.
+func (p *Peer) Healthy() bool {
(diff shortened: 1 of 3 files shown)


brian-brazil:
I don't think an unhealthy mesh should cause the Alertmanager to be killed, during a network partition you want it to keep running.

roidelapluie:
I agree with @brian-brazil. This is even the opposite: those are the moments you want AM to work.

stuartnelson3:
that's fair. I'm wondering if there's any benefit to exposing this directly through an endpoint rather than just through metrics.

PR #1327, opened 2018-04-16
tell users opening issues to use alertmanager  --version

tzz:
`-version` doesn't work as of 0.15-rc.1 so users should run `alertmanager  --version`

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -25,7 +25,7 @@
 
 * Alertmanager version:
 
-	insert output of `alertmanager -version` here
+	insert output of `alertmanager --version` here
 
 * Prometheus version:
 


mxinden:
@tzz Thanks for the fix. We have introduced [DCO](https://github.com/probot/dco#how-it-works) for the entire Prometheus Github org. That is why the _DCO_ bot failed on your PR.

**TLDR: Could you add `-s` to your `git commit` command?**

tzz:
@mxinden I think it adds a lot of friction to ask for `-s` for **documentation** changes, especially tiny ones like this one-liner. Can you please reconsider?

To be clear, I **can** use `-s`. I'm asking to consider that generally documentation changes are much easier from the web interface, where signing off (as far as I know) is not available.

Let me know and I'll proceed accordingly. Thank you.

mxinden:
I understand the extra work this puts on you @tzz. I am sorry about that. The DCO is there to protect us from legal issues. This is both important for actual code, as well as its documentation.

You can still use the web-interface. You only need to make sure that you add "Signed-off-by: Random J Developer <random@developer.example.org>" to the commit message. Does this address your concern?

Thanks to @beorn7 we now have a wiki page giving you more details: https://github.com/prometheus/prometheus/wiki/DCO-signing

jamtur01:
@tzz There's also a number of ways you can automate this - for repos that need -s you could use:

* A git alias that automatically adds -s.
* A commit message template that automatically appends the Signed-off-by line.
* In very recent Git versions use `git-interpret-trailers` (see https://stackoverflow.com/a/46536244/1048941)

PR #1329, opened 2018-04-16
Templating the Slack API URL

hairyhenderson:
I'd like to be able to template parts of the Slack API URL. This PR allows that.

/cc @stuartnelson3

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -711,7 +711,7 @@ func (n *Slack) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 		return false, err
 	}
 
-	resp, err := ctxhttp.Post(ctx, c, string(n.conf.APIURL), contentTypeJSON, &buf)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Can you explain why you want to template this?

hairyhenderson:
@brian-brazil The URL is effectively a secret, and I want to be able to store it separate from the AM config. This can be done for other secrets (such as `pagerduty`'s `routing_key`), since they're templated.

brian-brazil:
Templating of things like this is for routing, not for secrets and it'd generally a bad idea to use the templating like this as anyone who can change the config will be able to exfiltrate them.
See https://prometheus.io/docs/operating/security/#alertmanager

`routing_key` is templatable in spite of being a secret, not because it is one.

hairyhenderson:
Ok, understood. I'll have to approach this differently, then 👍

Issue #1199, opened 2018-01-15
amtool requires --alertmanager.url for check-config in 0.13.0


mitchellrj:
**What did you do?**
```
amtool check-config /path/to/alertmanager.yml
```

**What did you expect to see?**
```
Checking '/path/to/alertmanager.yml'  SUCCESS
Found 1 templates:   SUCCESS
```

**What did you see instead? Under which circumstances?**
```
amtool: error: required flag --alertmanager.url not provided
```

**Environment**

* System information:
Linux 4.9.49-moby x86_64

* Alertmanager version:
```
alertmanager, version 0.13.0 (branch: HEAD, revision: fb713f6d8239b57c646cae30f78e8b4b8861a1aa)
  build user:       root@d83981af1d3d
  build date:       20180112-10:32:46
  go version:       go1.9.2
```

carlpett:
Sorry for the slow reply, been travelling last week. Also, sorry for
causing this bug :(
I don't see a nice way around this, unfortunately. The two workarounds I
can think of is to either remove the `Required()` and do the check
explicitly for all the other sub-commands, or move the flag to the
sub-commands.
Not sure which is the least bad - both seem to invite subtle bugs or
behavior mismatches over time.

stuartnelson3:
Neither of those options seems very desirable. I guess we'll leave it as it is for now, and just let users know that they need to pass something as the alertmanager url.

stuartnelson3:
fixed by #1330

PR #1330, opened 2018-04-17
[amtool] use kingpin.v2

### LINKED ISSUE #1328: amtool doesn't depend on a stable version of kingpin
`amtool` depends on a version of the kingpin package (commit [75c6058](https://github.com/alecthomas/kingpin/commit/75c60582c55399763d53eeb59078d63d3616f58c) to be precise) that isn't the [current stable](https://gopkg.in/alecthomas/kingpin.v2) version (v2).

It has at least 2 drawbacks:
* Package maintainers' life gets harder.
* There's no guarantee on the API stability outside of v1/v2 so updating the `kingpin` package might break the build (and in fact it does currently).

The (only?) reason for using the "unstable" version is that `amtool` already supported reading its configuration from YAML files before moving to kingpin but `v2` kingpin didn't offer this out-of-the-box (more details [here](https://github.com/prometheus/alertmanager/pull/976#discussion_r157977731)).

simonpasquier:
This is an attempt to fix #1328. To workaround the lack of configuration file support in kingpin v2, it injects environment variables after reading the flags from the YAML files. This isn't pretty but at least seems to work (this would need more testing obviously).

As noted in #976, there's no way to have longer help descriptions when using the `--help-long` flag. I've tried several approaches but failed. I don't think it is even possible. So switching to v2 would mean to get rid of either the short helps or the long helps.

Waiting for @stuartnelson3 &  @TheTincho's comments...


cli/config_resolver.go
```diff
@@ -3,55 +3,56 @@ package cli
...
+var (
+	// kingpin transforms flags to <APPNAME>_<FLAG>.
+	envarTransformRegexp = regexp.MustCompile(`[^a-zA-Z0-9_]+`)
+	oldFlags             = map[string]string{
+		"comment_required": "require-comment",
 	}
+)
 
-	resolver := configResolver{}
+// loadConfig loads flags from YAML files and injects them as environment variables.
```
brian-brazil: Could any of these be secrets? The environment is not a safe place for secrets.

TheTincho: An idea would be to store this in a public map at init time and use it to populate defaults for flags.. but maybe it is too convoluted

TheTincho:
Thanks for this!

I have no comments on the PR, except my small suggestion.

cli/config_resolver.go
stuartnelson3: The alertmanager.url can potentially contain basic auth credentials.

Any ideas on how else to implement this? @TheTincho's map idea might work. Parse the file and populate a struct/map, and set those as the default values for their appropriate flags?

simonpasquier: I've just tried and it doesn't work because of `alertmanager.url` (at least) which is a required flag . As stated in the kingpin [documentation](https://godoc.org/gopkg.in/alecthomas/kingpin.v2#FlagClause.Required): `You can not provide a Default() value to a Required() flag`.

I've tried the suggestion from https://github.com/alecthomas/kingpin/issues/208#issuecomment-327407428 but it doesn't work either because some flags are only valid for some commands and you can't have a flag present both in the configuration file and in the program's arguments (unless the flag can be repeated).

stuartnelson3: Could we move it from being a `Required` flag, provide the value in the config file, and validate after the fact ourselves?

simonpasquier: @stuartnelson3 good idea, it seems to work...

stuartnelson3:
This has the fantastic side-effect of allowing `check-config`to no longer require `--alertmanager.url`

https://github.com/prometheus/alertmanager/issues/1199

cli/root.go
```diff
@@ -4,17 +4,40 @@ import (
...
 
+func requireAlertManagerURL(pc *kingpin.ParseContext) error {
+	// Return no error if an help flag is set.
+	for _, elem := range pc.Elements {
+		f, ok := elem.Clause.(*kingpin.FlagClause)
+		if !ok {
+			continue
+		}
+		name := f.Model().Name
+		if name == "help" || name == "help-long" || name == "help(man" {
```
stuartnelson3: what's `help(man`?

cli/config/config.go
```diff
@@ -0,0 +1,74 @@
...
+			return nil, err
+		}
+
+		var m map[string]string
+		err = yaml.Unmarshal(b, &m)
+		if err != nil {
+			return nil, err
+		}
+		for k, v := range m {
+			if new, ok := legacyFlags[k]; ok {
```
stuartnelson3: it appears to be compiling here, but `new` is a keyword in golang and should be renamed

simonpasquier:
Thanks @stuartnelson3, from my testing, the precedence between command line flags and configuration files works as expected but I plan to increase the coverage in the unit tests.

There's still a question around short vs. long help messages though.

cli/config/config.go
simonpasquier: Thanks for the catch-up!

cli/root.go
simonpasquier: Oops typo.

stuartnelson3:
>As noted in #976, there's no way to have longer help descriptions when using the --help-long flag. I've tried several approaches but failed. I don't think it is even possible. So switching to v2 would mean to get rid of either the short helps or the long helps.

I would prefer to keep the longer help message. The documentation isn't too long, and it covers the use cases. We would have to figure out where to add the long help information on prometheus.io if we were to get rid of it.

simonpasquier:
> I would prefer to keep the longer help message.

This is my feeling too: having it buried under `--help-long` isn't very user-friendly. I'll update the PR in this direction then.

cli/root.go
```diff
@@ -4,64 +4,77 @@ import (
...
+		kingpin.Fatalf("required flag --alertmanager.url not provided")
+	}
+	return nil
+}
+
 func Execute() {
 	var (
-		longHelpText = map[string]string{}
-		app          = kingpin.New("amtool", "Alertmanager CLI").DefaultEnvars()
+		app = kingpin.New("amtool", helpRoot).DefaultEnvars()
```
stuartnelson3: I think `DefaultEnvars()` can be removed, right?

stuartnelson3: k. I doubt anyone is relying on it, but I suppose if someone is we can leave it for now.

Issue #1002, opened 2017-09-25
send_resolved:true not take effect when I setup two email addresses.


smallfish01:
Hi Support,
When I setup two emails in my simply.yml and enable send_resolved, I just received the resolved email in my second email (smallfish@abc.com).
My simply.yml configurations as below:

receivers:
- name: 'default-receiver'
  email_configs:
  - to: '18927374488@189.cn'
  - to: 'smallfish@abc.com'
    html: '{{ template "email.html" .  }}'
    send_resolved: true
And I check http://localhost:9093/#/status:
- name: default-receiver
  email_configs:
  - send_resolved: false
    to: 118927374488@189.cn
    from: sys-admin@abc.com
    smarthost: smtp.abc.com:587

When I change my second emall (smallfish@abc.com) to the first, I cannot receive resolved email in smallfish@abc.com but I can receive resolved email in 18927374488@189.cn .

Thank you very much!
    auth_username: sys-admin@abc.com
    auth_password: <secret>
    auth_secret: null
    auth_identity: ""
    headers:
      From: smallfish@abc.com
      Subject: '{{ template "email.default.subject" . }}'
      To: 18927374488@189.cn
    html: '{{ template "email.default.html" . }}'
    require_tls: true
  - send_resolved: true
    to: smallfish@abc.com

stuartnelson3:
`email_configs` only supports a single `to` field.

https://prometheus.io/docs/alerting/configuration/#<email_config>

stuartnelson3:
To expand: You can comma separate the email names.

smallfish01:
The issue has been fixed. I found I need setup send_resolved: true for every email address, such as:
name: 'default-receiver'
email_configs:

    to: '18927374488@189.cn'
    html: '{{ template "email.html" . }}'
    send_resolved: true
    to: 'smallfish@abc.com'
    send_resolved: true

PR #1334, opened 2018-04-20
Validate Slack field config and only allow the necessary input

taharah:
Resolves the comments on the PR [#1002](https://github.com/prometheus/docs/pull/1002) to update the documentation for Slack fields. Currently, any arbitrary key/value pairs can be defined; however, per the Slack documentation for defining fields [[1](https://api.slack.com/docs/message-attachments#fields)] only the properties `title`, `value`, and `short` can be defined with the first two being required for each field. This updates the Slack field configuration to only allow the aforementioned three properties as well as validates that both a title and value are defined for each property in the `yaml.Unmarshaler` interface. Moreover, this allows the value for `short` to be defined on a per-field basis.

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -220,6 +220,31 @@ func (c *PagerdutyConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
 	return nil
 }
 
+// SlackField configures a single Slack field that is sent with each notification.
+// Each field must contain a title, value, and optionally, a boolean value to indicate if the field
+// is short enough to be displayed next to other fields designated as short.
+// See https://api.slack.com/docs/message-attachments#fields for more information.
+type SlackField struct {
+	Title string `yaml:"title,omitempty" json:"title,omitempty"`
+	Value string `yaml:"value,omitempty" json:"value,omitempty"`
+	Short *bool  `yaml:"short,omitempty" json:"short,omitempty"`
+}
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface for SlackField.
+func (c *SlackField) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type plain SlackField
+	if err := unmarshal((*plain)(c)); err != nil {
+		return err
+	}
+	if c.Title == "" {
+		return fmt.Errorf("missing title in Slack field configuration")
+	}
+	if c.Value == "" {
+		return fmt.Errorf("missing value in Slack field configuration")
+	}
+	return nil
+}
+
 // SlackConfig configures notifications via Slack.
 type SlackConfig struct {
 	NotifierConfig `yaml:",inline" json:",inline"`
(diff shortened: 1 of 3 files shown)


config/notifiers.go
```diff
@@ -220,6 +220,31 @@ func (c *PagerdutyConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
...
+}
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface for SlackField.
+func (c *SlackField) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	type plain SlackField
+	if err := unmarshal((*plain)(c)); err != nil {
+		return err
+	}
+	if c.Title == "" {
+		return fmt.Errorf("missing title in Slack field configuration, value=%v", c.Value)
```
simonpasquier: I wouldn't log `c.Value` in the error.

config/notifiers.go
```diff
@@ -220,6 +220,31 @@ func (c *PagerdutyConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
...
+	}
+	if c.Value == "" {
+		return fmt.Errorf("missing value in Slack field configuration, title=%v", c.Title)
```
simonpasquier: Same for `c.Title`.

config/notifiers_test.go
```diff
@@ -253,3 +253,43 @@ token: ''
 		t.Errorf("\nexpected:\n%v\ngot:\n%v", expected, err.Error())
 	}
 }
+
+func TestSlackFieldConfigValidation(t *testing.T) {
```
simonpasquier: Can you add a test case with a valid YAML connfiguration?

notify/impl.go
```diff
@@ -678,12 +671,21 @@ func (n *Slack) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
...
-		for k, v := range n.conf.Fields {
-			fields[k] = slackAttachmentField{
-				tmplText(v["title"]),
-				tmplText(v["value"]),
-				n.conf.ShortFields,
+		var fields = make([]config.SlackField, numFields)
+		for index, field := range n.conf.Fields {
+			// Check if short was defined for the field otherwise fallback to the global setting
+			var short bool
+			if field.Short {
```
simonpasquier: This won't work as expected when `field.Short` is set to `false` on purpose. `Short` needs to be a `*bool` instead.

taharah:
@simonpasquier I've resolved all of the comments you made. Thank you for reviewing!

config/notifiers_test.go
```diff
@@ -253,3 +253,60 @@ token: ''
...
+			expected: "missing title in Slack field configuration",
+		},
+		{
+			in: `
+fields:
+- title: first
+  value: hello
+  short: true
+- title: second
+  value: world
```
simonpasquier: It would be good to validate that the unmarshaled data is matching what is expected when there's no error.

notify/impl.go
```diff
@@ -678,12 +671,21 @@ func (n *Slack) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
...
-		var fields = make([]slackAttachmentField, numFields)
-		for k, v := range n.conf.Fields {
-			fields[k] = slackAttachmentField{
-				tmplText(v["title"]),
-				tmplText(v["value"]),
-				n.conf.ShortFields,
+		var fields = make([]config.SlackField, numFields)
+		for index, field := range n.conf.Fields {
+			// Check if short was defined for the field otherwise fallback to the global setting
+			var short *bool
```
simonpasquier: nit but I would do:

```go
                        var short bool
			if field.Short != nil {
				short = *field.Short
			} else {
				short = n.conf.ShortFields
			}
```

taharah: The main reason it is being set as it is currently is that I reused `SlackField` from the config. package, which has 'Short` defined as a `*bool` now. I initially had it like that, but changed it after the tests failed and caught the error. I can change the if statement to be as above, and then pass a pointer to the variable `short` when building the `SlackField`.

config/notifiers_test.go
taharah: :+1: I'll get this added.

taharah:
@simonpasquier I resolved your latest comments. If there's anything else you see in the latest batches of changes let me know and I'll get them pushed up. Thanks!

simonpasquier:
LGTM

PR #1335, opened 2018-04-20
Improve notification instrumentation

stuartnelson3:
- Add the groupName on notification
success/failure counter, to help users debug
quicker which group is failing (in case they use
different configs)
- Add notificationLatencySeconds histogram to
debug duplicate messages. This can help rule out
if duplicate messages are being caused by
excessive latency when sending a notification.

The buckets are very coarse. The current default delay is 15s, set by `--cluster.peer-timeout`, and I'm mostly interested in seeing if this threshhold is causing duplicate messages that have been reported.

I wanted to add the groupName to the success/failure counter for integrations, but now this makes it difficult to initialize the metrics with values. I'm not sure if this is worth it (I think it will help reduce the time taken to fix a broken receiver's config), but if so then I'm not sure what would be the best way to handle initialization.


notify/notify.go
```diff
@@ -624,8 +623,11 @@ func (r RetryStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.Ale
 
 		select {
 		case <-tick.C:
-			if retry, err := r.integration.Notify(ctx, alerts...); err != nil {
-				numFailedNotifications.WithLabelValues(r.integration.name).Inc()
+			now := time.Now()
+			retry, err := r.integration.Notify(ctx, alerts...)
+			notificationLatencySeconds.WithLabelValues(r.integration.name).Observe(time.Since(now).Seconds())
+			if err != nil {
+				numFailedNotifications.WithLabelValues(r.integration.name, r.groupName).Inc()
```
simonpasquier: The definition of `numFailedNotifications` hasn't been updated with the new label.

notify/notify.go
```diff
@@ -635,7 +637,7 @@ func (r RetryStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.Ale
 				// integration upon context timeout.
 				iErr = err
 			} else {
-				numNotifications.WithLabelValues(r.integration.name).Inc()
+				numNotifications.WithLabelValues(r.integration.name, r.groupName).Inc()
```
simonpasquier: ditto

brian-brazil:
There can be thousands of groupnames inside a single alertmanager, which is too high cardinality wise to use in a label. That's why we only break out by notifier currently.

stuartnelson3:
@simonpasquier could you take a look?

simonpasquier:
:+1:

Issue #1178, opened 2018-01-02
allow user to request consistent ordering of grouped alerts


tzz:
alertmanager currently groups alerts without a consistent ordering, which is tough to follow for large groups.

The user should be able to specify a sort order on some label or annotation, which would give consistency and locally meaningful sorting. This could be done with a Go sort function in the template or with a alertmanager configuration entry perhaps.

brian-brazil:
I propose that at a minimum the `Alerts` in the template structure should come with some ordering on it, like we have done elsewhere.

stuartnelson3:
Are you referring to the `/api/v1/alerts/groups` endpoint? we're currently sorting by internal identifier in `/api/v1/alerts` (https://github.com/prometheus/alertmanager/blob/master/api/api.go#L405-L407)

brian-brazil:
This is about notification templates.

tzz:
I don't mind writing some code if it will help move this forward. For me it would be enough to have a function or alert manager option to simply sort alerts lexicographically by label or annotation.

brian-brazil:
I think we should start with a consistent ordering on the Alerts we provide to notification templates. That way users get something okay without having to do extra work.

tzz:
For a default ordering, how about lexicographical by `(alert.annotations.summary, alert.creation_time)`? That's the simplest thing I can see that could be generally useful.

brian-brazil:
You can't presume that any particular annotation exists, nor that creation times are stable. I'd suggest working entirely off alert labels. This should not be configurable at this level.

tzz:
OK, what alert labels would make for a sensible default sort key? `(job, instance)`?

brian-brazil:
You'll need to use all of them, otherwise it won't be consistent. Moving job and instance to the front is probably wise.

tzz:
Looking at `dispatch/dispatch.go`, there seem to be two ways to go: either modify `aggrGroup` to always have a sorted list of `Alert` structs instead of a map, or sort the `alertsSlice` when alerts are flushed. The `alertsSlice` sort seems easier to implement.

Does that make sense or am I misunderstanding your intention or the code?

brian-brazil:
Yes, the slice would be the one to sort somewhere along that codepath.

tzz:
I'm having trouble compiling the code because of some strange unrelated error `vendor/github.com/weaveworks/mesh/token_bucket.go:33: undefined: time.Until
[...]

PR #1336, opened 2018-04-20
Allow to determine which alerts' label would be sorted

tiagoapimenta:
This allow user to determine which label will be used for sorting


brian-brazil:
Things affecting notification templating shouldn't leak outside notification templating. Why do you want this?

tiagoapimenta:
I don't believe it is fair to be hardcoded the labels that alerts is being sorted

tiagoapimenta:
I forgot to say that the sorting is reversed on master, in my humble option the way I did is the most appropriate, but if you believe it is worthwhile the sorting could be configurable as well.

tzz:
@tiagoapimenta the sorting is fixed in #1339

tzz:
this would resolve #1178 I think

SuperQ:

PR #1337, opened 2018-04-20
Bringing new feature that allow user to determine sorting labels

tiagoapimenta:
This allow user to determine which label will be used for sorting


mxinden:
@tiagoapimenta What is your use case to configure sorting by route?

tiagoapimenta:
@mxinden As we are able to group by route, I feel it would be fair to be possible to configure the sorting of this list by route as well, as template on go does not allow functional sorting, it could be at least configurable by label.

PR #1339, opened 2018-04-23
Sort dispatched alerts by job+instance in the correct order (#1178)

tzz:
Alert order was wrongly implemented, trivial change. Followup fix to #1234

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -430,15 +430,15 @@ func (ag *aggrGroup) flush(notify func(...*types.Alert) bool) {
 		for _, override_key := range [...]model.LabelName{"job", "instance"} {
 			key_i, ok_i := alertsSlice[i].Labels[override_key]
 			if !ok_i {
-				return true
+				return false
 			}
 			key_j, ok_j := alertsSlice[j].Labels[override_key]
 			if !ok_j {
-				return false
+				return true
 			}
 
 			if key_i != key_j {
-				return key_i > key_j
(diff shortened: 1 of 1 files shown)


tiagoapimenta:
@tzz Don't you believe you should switch the true/false on lines 433 and 437 as well?

simonpasquier:
> This really should have an acceptance test to avoid the hassle but I don't know how to write it.

AFAICT the existing sort algorithm on `types.Alert` is only used by `dispatch/dispatch_test.go`.

https://github.com/prometheus/alertmanager/blob/cecfe5b2f5e76174957f0ae441affe535c77d728/types/types.go#L269-L271

I guess that you can replace it by what you have implemented in `dispatch/dispatch.go`. Then you can add the unit tests to `types/types_test.go` which is simpler than doing it in `dispatch/dispatch_test.go`.

tzz:
@simonpasquier thanks for the advice. Is the test a blocker for this PR or can we add it later (so it won't block the 0.15 release)? I definitely can't work on it in the next 2 weeks.

simonpasquier:
I would rather have some unit tests because it may surface other sorting problems. I could give it a try if you don't mind.

tzz:
Please do! I would really appreciate it.

mxinden:
https://github.com/prometheus/alertmanager/pull/1349/ included this patch with additional tests. I will close here. @tzz please feel free to reopen in case I am missing something.

Thanks a lot for the fix!

PR #1349, opened 2018-04-30
Sort alerts in correct order

simonpasquier:
This PR adds unit tests on top of #1339.


simonpasquier:
cc @tzz

tzz:
That's at least 10 times better than my code :) Thanks for giving it so much thought.

Issue #344, opened 2016-05-08
Be able to silence based on generatorURL


brian-brazil:
The primary purpose of generatorURL is to be able to silence on it, this is important when you've a broken/inaccessible Prometheus server that continues to send incorrect alerts. This should be an additional option for silences, potentially hidden in an advanced section.

matthiasr:
+1 I needed exactly this the other day.

juliusv:
Not objecting to being able to silence on it, but to the statement that that is the primary purpose of the generatorURL :) My intent when sending this field to the Alertmanager was to have a nice link back from alerts/notifications to the Prometheus server graphing interface, so that one can easily play with the alerting expression.

brian-brazil:
The silence is the primary purpose, and quite some time was spent arguing that it should be a separate field rather than a label or annotation due to the required semantics of the silence and de-duplication. 

Technically it should be two fields however with that and the link being combined into one it's much less likely that other am client implementations would skip over it not realising why it's so important.

Kellel:
Since generatorURL is outside of labels (which is what are used for matching currently) do we want to expand this feature request to be able to silence based on any part of a alert? 

I can see that receivers might be useful to silence against, maybe fingerprint(?), but to be honest i'm struggling to come up with a silence usecase for any of the other fields.

Also in my experience I haven't ever hit something that would make silencing by generatorURL useful. Usually I will have labels that convey enough of the same information that it's not needed.

brian-brazil:
I don't see a need to silence on anything else.

> Usually I will have labels that convey enough of the same information that it's not needed.

The whole point is for cases where labels aren't sufficient, such as a HA pair which are producing identical alert labels but one of them is dodgy.

sunilvasnata:
I have similar requirement, want to inhibit notification from second replica of Prometheus. Any update on this issue ??

zecke:
I recently had a case where silencing based on the generatorURL(or anything related peer) would have made a big difference. Is there still interest to implement it? What do you think of the approach taken here?

base698:
Was expecting a `__source__` or something.  The alert seems to have the source, if you could write a regex to it it would allow for a pretty useful silence.

crockk:
Bump on this very old issue. Does this functionality exist in some other way? Being able to silence based on the source of the alert would be really useful.

juliusv:
> Does this functionality exist in some other way? Being able to silence based on the source of the alert would be really useful.

[...]

PR #1352, opened 2018-04-30
api: Introduce OpenAPI generated Alertmanager API V2

mxinden:
The current Alertmanager API v1 is undocumented and written by hand.
This patch introduces a new Alertmanager API - v2. The API is fully
generated via an OpenAPI 2.0 [1] specification (see
`api/v2/openapi.yaml`) with the exception of the http handlers itself.

Pros:
- Generated server code
- Ability to generate clients in all major languages
  (Go, Java, JS, Python, Ruby, Haskell, *elm* [3] ...)
    - Strict contract (OpenAPI spec) between server and clients.
    - Instant feedback on frontend-breaking changes, due to strictly
      typed frontend language elm.
- Generated documentation (See Alertmanager online Swagger UI [4])

Cons:
- Dependency on open api ecosystem including go-swagger [2]

In addition this patch includes the following changes.

- README.md: Add API section

- test: Duplicate acceptance test to API v1 & API v2 version

  The Alertmanager acceptance test framework has a decent test coverage
  on the Alertmanager API. Introducing the Alertmanager API v2 does not go
  hand in hand with deprecating API v1. They should live alongside each
  other for a couple of minor Alertmanager versions.

  Instead of porting the acceptance test framework to use the new API v2,
  this patch duplicates the acceptance tests, one using the API v1, the
  other API v2.

  Once API v1 is removed we can simply remove `test/with_api_v1` and bring
  `test/with_api_v2` to `test/`.

[1]
https://github.com/OAI/OpenAPI-Specification/blob/master/versions/2.0.md

[2] https://github.com/go-swagger/go-swagger/

[3] https://github.com/ahultgren/swagger-elm

[4]
http://petstore.swagger.io/?url=https://raw.githubusercontent.com/mxinden/alertmanager/apiv2/api/v2/openapi.yaml


mxinden:
Link to Prometheus developers Google group discussion: https://groups.google.com/forum/#!topic/prometheus-developers/2CCuFTMbmAg

mxinden:
@stuartnelson3 @simonpasquier @brian-brazil @brancz @squat would you mind taking a look at this proposal? Most of these changes are either vendored or auto-generated code.

The first commit builds up the basic skeleton of the API v2 and makes it _near_ feature complete with v1. All follow up commits are related to documentation, testing and basic refactoring.

Unfortunately GitHub collapses `api/v2/api.go` by default, which is the heart of the new API and not auto-generated.

Very much looking forward to your feedback. Let me know if you need a more detailed description for a code review. To understand the structure, the [online Swagger UI of the Alertmanager API Spec](http://petstore.swagger.io/?url=https://raw.githubusercontent.com/mxinden/alertmanager/apiv2/api/v2/openapi.yaml) is probably a good start.

stuartnelson3:
>Very much looking forward to your feedback. Let me know if you need a more detailed description for a code review. To understand the structure, the online Swagger UI of the Alertmanager API Spec is probably a good start.

This alone instantly makes me like it :)

Will take a look soon. I haven't looked at the schedule, but this might also be something worth discussing at the dev summit the day after promcon

mxinden:
> Will take a look soon. I haven't looked at the schedule, but this might also be something worth discussing at the dev summit the day after promcon

Already added it to the list @stuartnelson3.

mxinden:
Thanks @brian-brazil for looking over this!

---

> Might silencedby/inhibited by be something we wish to remove on performance grounds in the future?

Have you hit concrete performance issues in the `Marker` logic? Can you add some more details?

---

> On the other hand apparently there's extra XSRF-related protections for PUT/DELETE that don't apply to POST

Can you reference some further documentation on this matter here?

---

> The generatorURL may not actually be a URI, I'd leave it as a string.

In which cases is the _generatorURL_ not a URI? If so, we should rename the parameter?

---

> Matchers are strings in input, and structs as output. This is [in]consistent.

Are you referring to the `GET /alerts?filter` parameter?

According to the [OpenAPI 2.0 spec](https://swagger.io/docs/specification/2-0/describing-parameters/#query-parameters):

> Query parameters only support primitive types. You can have an array, but the items must be a primitive value type. Objects are not supported.  

So we could represent matchers as an array of strings, but not as objects. What do you think? Representation as an array of strings sounds better than a single string to me.

brian-brazil:
> Have you hit concrete performance issues in the Marker logic? Can you add some more details?

No, but I can imagine us wanting to remove this in the future if we changed the implementation. Once this is here removing it could be difficult.

> Can you reference some further documentation on this matter here?

https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#Functional_overview

> In which cases is the generatorURL not a URI? If so, we should rename the parameter?

When the alertmanager client doesn't pass a URL. It shouldn't happen, but it's not as if we're validating it when it comes in. Renaming it would be a breaking change, and it is usually a URL.

> Are you referring to the GET /alerts?filter parameter?

Yes.

mxinden:
>> Have you hit concrete performance issues in the Marker logic? Can you add some more details?

> No, but I can imagine us wanting to remove this in the future if we changed the implementation. Once this is here removing it could be difficult.

I am voting for keeping the `silencedby` and `inhibited` properties for now with the risk that we might have to make a breaking change at some point when we refactor the *Marker* logic. Much of the current UI depends on it, which I would like to port over to API v2 as a follow up to this.

Any thoughts by others on this?

mxinden:
>>   In which cases is the generatorURL not a URI? If so, we should rename the parameter?

> When the alertmanager client doesn't pass a URL. It shouldn't happen, but it's not as if we're validating it when it comes in. Renaming it would be a breaking change, and it is usually a URL.

As this is only enforced by the API v2, we can include the validations once we port Alertmanager client and UI over to API v2, if the validations are not already auto-generated in the first place.

brian-brazil:
> As this is only enforced by the API v2, we can include the validations once we port Alertmanager client and UI over to API v2, if the validations are not already auto-generated in the first place.

My point is that adding validation would be a breaking change, and I think that we shouldn't break it.

mxinden:
@stuartnelson3 @simonpasquier @brancz @squat friendly ping - PR looking for review.

brancz:
Generally this looks good to me and a step forward in all directions. (I didn't review all the generated code, but parts of it and it seems reasonably readable should there ever be a problem)

That said there are still some instrumentation and structural TODOs, so I'll wait for that for a final review, but this is going in a great direction.

knweiss:
In case anyone else (like me) is wondering why this new API is based on the OpenAPI 2.0 spec when the 3.0.0 spec was released a year ago and the latest spec revision is [OpenAPI 3.0.1](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.1.md):

Unfortunately, `go-swagger`currently does only support the 2.0 spec. See https://github.com/go-swagger/go-swagger/issues/1122 for details.

@mxinden Apart from that: Were there any other reasons that kept you from using the latest spec? Would you mind to share them? (Honest question, not a complaint!)

mxinden:
In case there are no objections I would like to delay adding proper instrumentation of the Alertmanager API v2 to a follow up PR. This PR is already big enough and there is some deeper refactoring necessary as `prometheus/common/route.Router` doesn't allow me to register a full handler (generated by Swagger) instead of individual endpoints.

mxinden:
I have addressed all outstanding `TODO`s.

@stuartnelson3 @simonpasquier @brian-brazil @brancz I would like to move forward here and merge into `master`. What are your thoughts?

This patch also touches the `README.md`, stating:

> API v2 is still under heavy development and thereby subject to change.

mxinden:
Before merging I would like to squash these commits into one and rework the commit message.

api/v2/openapi.yaml
```diff
@@ -0,0 +1,345 @@
...
+      tags:
+        - silence
+      operationId: postSilences
+      description: Post a new silence or update an existing one
+      consumes:
+        - application/json
+      parameters:
+        - in: body
+          name: silence
+          description: the silence to create
```
brian-brazil: The

api/v2/openapi.yaml
```diff
@@ -0,0 +1,345 @@
...
+      tags:
+        - silence
+      operationId: getSilences
+      description: Get a list of silences
+      responses:
+        '200':
+          description: List of silences response
+          schema:
+            $ref: '#/definitions/silences'
+        '500':
```
brian-brazil: Why does this explicitly have 500 called out, and others don't?

api/v2/openapi.yaml
```diff
@@ -0,0 +1,345 @@
...
+        format: date-time
+      updatedAt:
+        type: string
+        format: date-time
+      endsAt:
+        type: string
+        format: date-time
+      generatorURL:
+        type: string
+        format: uri
```
brian-brazil: This is an arbitrary string.

api/v2/openapi.yaml
```diff
@@ -0,0 +1,345 @@
...
+  silences:
+    type: array
+    items:
+      $ref: '#/definitions/silence'
+  matchers:
+    type: array
+    items:
+      $ref: '#/definitions/matcher'
+    minItems: 1
+  matcher:
```
brian-brazil: How is a!="b" indicated?

api/v2/openapi.yaml
```diff
@@ -0,0 +1,345 @@
...
+          required: false
+          type: string
+      responses:
+        '200':
+          description: Get alerts response
+          schema:
+            type: array
+            items:
+              '$ref': '#/definitions/alert'
+        '400':
```
kyoto: Maybe define references for default 400 and 500 error responses, since they are used in a few places?

```
responses:
  400:
    description: Bad request
    schema:
      type: string
  500:
    description: Internal server error
    schema:
      type: string
```

api/v2/openapi.yaml
```diff
@@ -0,0 +1,345 @@
...
+swagger: '2.0'
+
+info:
+  version: 0.0.1
+  title: Alertmanager API
+  description: API of the Prometheus Alertmanager (https://github.com/prometheus/alertmanager)
+  license:
+    name: Apache 2.0
+    url: http://www.apache.org/licenses/LICENSE-2.0.html
+
```
kyoto: Would it make sense to define `consumes` and `produces` at the root level?
```
consumes:
   - "application/json"
produces:
  - "application/json"
```

api/v2/openapi.yaml
```diff
@@ -0,0 +1,345 @@
...
+          description: Bad request
+          schema:
+            type: string
+  /silence/{silenceID}:
+    get:
+      tags:
+        - silence
+      operationId: getSilence
+      description: Get a silence by its id
+      parameters:
```
kyoto: Since `silenceID` is a `path` parameter, you could move this `parameters` block under `/silence/{silenceID}` so that it applies to all methods.

api/v2/openapi.yaml
```diff
@@ -0,0 +1,345 @@
...
+          description: A silence with the specified id was not found
+        '500':
+          description: Internal server error
+          schema:
+            type: string
+    delete:
+      tags:
+        - silence
+      operationId: deleteSilence
+      description: Delete a silence by its id
```
kyoto: `id` → `ID`

api/v2/openapi.yaml
```diff
@@ -0,0 +1,345 @@
...
+        '400':
+          description: Bad request
+          schema:
+            type: string
+  /silence/{silenceID}:
+    get:
+      tags:
+        - silence
+      operationId: getSilence
+      description: Get a silence by its id
```
kyoto: `id` → `ID`

api/v2/openapi.yaml
```diff
@@ -0,0 +1,345 @@
...
+          format: uuid
+          required: true
+          description: ID of the silence to get
+      responses:
+        '200':
+          description: Get silence response
+          schema:
+            $ref: '#/definitions/silence'
+        '404':
+          description: A silence with the specified id was not found
```
kyoto: `id` → `ID`

api/v2/openapi.yaml
```diff
@@ -0,0 +1,345 @@
...
+            type: string
+        '500':
+          description: Internal server error
+          schema:
+            type: string
+    post:
+      tags:
+        - alert
+      operationId: postAlerts
+      description: Create a new Alert
```
kyoto: Looks like this allows multiple alerts to be created at once, so should change description to indicate that.

api/v2/openapi.yaml
```diff
@@ -0,0 +1,345 @@
...
+      parameters:
+        - in: body
+          name: alerts
+          description: the alerts to create
+          required: true
+          schema:
+            $ref: '#/definitions/alerts'
+      responses:
+        '200':
+          description: Get alerts response
```
kyoto: `Get` → `Create`

README.md
```diff
@@ -171,6 +171,25 @@ receivers:
   - routing_key: <team-DB-key>
 ```
 
+## API
+
+Alertmanager ships with two APIs, v1 and v2.
+
+- **v1**: At this point in time the API is undocumented. Best way to get started
```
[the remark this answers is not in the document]
mxinden: Sounds reasonable. Removed.

api/v2/openapi.yaml
mxinden: Whenever a handler e.g. `getSilencesHandler` needs to return a specific HTTP error code, it needs to be mentioned in the respective `path` definition in the OpenAPI specification.

Others don't have 500 explicitly called out, because their handlers do not return a 500 error code.

api/v2/openapi.yaml
mxinden: As far as I can tell as of today Alertmanager only supports equal or RegEx matching (See [internal matcher representation](https://github.com/prometheus/alertmanager/blob/master/types/match.go#L27)).

A user that would still like to input the semantics of a not-equal would need to use a RegEx.

Is this related to the API @brian-brazil?

api/v2/openapi.yaml
mxinden: :+1:

api/v2/openapi.yaml
mxinden: Good catch.

brian-brazil:
> Given the fact that the major version of the API changes, this breaking change seems acceptable to me.

This could break existing non-Prometheus clients of the AM, and some of them may not be running http servers to link to. It's intended primarily as a unique identifier of the source of an alert, that it is often also a URL that the user finds useful is secondary. What is the benefit of breaking this?

brancz:
Introducing a v2 API sounds like the right opportunity to set this kind of thing straight. I think if we continue to call it "generatorURL" then it should be a valid URI, if we don't want this validation, then we should call it "generator" instead.

brian-brazil:
Also keep in mind that the old API, which we will continue to have to support for some time, has no such validation so non-URLs will get into the system.

brancz:
I think generally I prefer the un-validated "generator", I would prefer a broken backlink over a non-accepted alert fired. And for non-Prometheus generators there could be non-URI cases as well for the generator I think (not sure this is something we would want to take into consideration though).

brian-brazil:
Something else is that this field name is exposed in notification templates and webhooks.

mxinden:
> This could break existing non-Prometheus clients of the AM

We are not going to remove API v1 right away, thereby giving non-Prometheus clients time to migrate.

> Also keep in mind that the old API, which we will continue to have to support for some time, has no such validation so non-URLs will get into the system.

The API v2 only validates client input, not server output. Thereby alerts with invalid generatorURLs created via API v1 can still be retrieved via API v2. 

> What is the benefit of breaking this?

It would remove the confusion of a user expecting a valid URI in a property called `generatorURL`. In addition when one clicks the `Source` button on an alert in the UI with an invalid `generatorURL`, that person is directed to an invalid page.

A user inserting an invalid `generatorURL` would get instant feedback via the API v2:

```json
"message":"generatorURL in body must be of type uri: \"1234\""
```

> Something else is that this field name is exposed in notification templates and webhooks.

Which makes me favour keeping `generatorURL` instead of using `generator`.

stuartnelson3:
I'm favoring @brancz's position -- moving to `generator` would remove potential semantic confusion about a client using `generatorURL` for a different purpose, and not validating it would support that.

however, @mxinden does raise a good point, that the current UI assumes that the `generatorURL` is a URI. Maybe we can on the frontend parse `generator` into either a URI (which we link to), or if it's invalid, render it in the expanded 'Info' section?

mxinden:
Validating the `generatorURL` is not a breaking-change for any upstream Prometheus projects. It might be a breaking change for existing non-Prometheus clients of Alertmanager, but so far we are not aware of any (please correct me if I am wrong).

I am still advocating for introducing this maybe-breaking-change with the new Alertmanager API v2. It is a major API version bump, thereby a breaking change should not struck anyone by surprise. Given the scenario that there is a non-Prometheus client of Alertmanager that depends on `generatorURL` not being a valid URI, for short-term, that project could still use the API v1, for long term that project could leverage alert annotations instead.

Moving from `generatorURL` to `generator` would introduce two breaking changes. One in the ingestion structure, second in notification templates and webhooks. In addition it forces us to include extra logic in the UI to detect whether it is a valid URI.

simonpasquier:
Re the `generatorURL` discussion, I'd agree with @mxinden that we should keep `generatorURL` (as it doesn't require additional change to the Prometheus projects) and add validation for it. Also IIUC the API definition allows for `generatorURL` to be empty so clients that don't pass a valid URI value today can just drop the field and potentially use Annotations instead.

brian-brazil:
In principle all clients should be setting that field, it has a difference purpose to an annotation as the ultimate goal is to allow silencing based on it. I'm against things that would result in clients not setting that field.

Makefile
```diff
@@ -54,6 +55,19 @@ ui/app/script.js: $(shell find ui/app/src -iname *.elm)
...
+	cd api/v2 && swagger generate server  -f openapi.yaml --exclude-main -A alertmanager
+
+test/with_api_v2/api_v2_client/models test/with_api_v2/api_v2_client/client: $(SWAGGER) api/v2/openapi.yaml
+	-rm -r test/with_api_v1/api_v2_client; mkdir -p test/with_api_v2/api_v2_client
+	cd test/with_api_v2/api_v2_client/ && swagger generate client -f ../../../api/v2/openapi.yaml
+
+$(SWAGGER):
+	go get -u -v -d github.com/go-swagger/go-swagger/cmd/swagger
+	cd $(FIRST_GOPATH)/src/github.com/go-swagger/go-swagger; git checkout 0.15.0
+	go install github.com/go-swagger/go-swagger/cmd/swagger
```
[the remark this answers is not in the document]
stuartnelson3: I would prefer it be contained within the project dir

api/v2/api.go
```diff
@@ -0,0 +1,577 @@
...
+		getAlertStatus: sf,
+		peer:           peer,
+		silences:       silences,
+		logger:         l,
+	}
+
+	// load embedded swagger file
+	swaggerSpec, err := loads.Analyzed(restapi.SwaggerJSON, "")
+	if err != nil {
+		level.Error(api.logger).Log("err", err)
```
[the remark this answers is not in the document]
mxinden: Great catch. Instead Alertmanager `os.Exit(1)` now. I think this is the most save option as starting with failed components could cause inconsistencies. Let me know what you think.

api/v2/openapi.yaml
```diff
@@ -0,0 +1,341 @@
...
+          name: inhibited
+          type: boolean
+          description: Show inhibited alerts
+          default: true
+        - in: query
+          name: unprocessed
+          type: boolean
+          description: Show unprocessed alerts
+          default: true
+        - name: filter
```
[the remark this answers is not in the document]
mxinden: Yes, see https://github.com/prometheus/alertmanager/pull/1352#issuecomment-405292279. Let me know if you have further suggestions.

cmd/alertmanager/main.go
```diff
@@ -389,10 +403,11 @@ func main() {
 
 	ui.Register(router, webReload, logger)
 
-	apiv.Register(router.WithPrefix("/api/v1"))
+	apiV1.Register(router.WithPrefix("/api/v1"))
 
-	level.Info(logger).Log("msg", "Listening", "address", *listenAddress)
-	go listen(*listenAddress, router, logger)
+	// TODO: How about having a http.handler for each (web, apiv1, apiv2) and
+	// combine them all together in `listen()`
```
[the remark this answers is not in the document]
mxinden: Cool. I would like to realise this along with the instrumentation follow up, outside of this PR (See https://github.com/prometheus/alertmanager/pull/1352#issuecomment-414334987).

Makefile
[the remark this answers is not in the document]
mxinden: Fair point.

In order for it to live inside `/vendor` it would need to be imported somewhere. I see you have solved this in https://github.com/prometheus/prometheus/pull/4430 via `go generate`. This works under the prerequisite that _go-swagger_ works as a library as _vfsgen_ does. I have not yet found an easy way how to accomplish that.

_Go-swagger_ publishes a docker image with the CLI on each release. I am suggesting using that instead. As the generated API code is checked in, running docker is not a prerequisite to build Alertmanager, but only needed whenever `openapi.yaml` is touched.

@simonpasquier @stuartnelson3 what do you think?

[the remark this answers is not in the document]
simonpasquier: Yes that would work. It is also consistent with the way we generate Javascript code from Elm files so it doesn't impose more constraints on contributors (eg requiring a container runtime).

simonpasquier:
> as the ultimate goal is to allow silencing based on it.

For the record, this is tracked in #344. In that case, it is a free-form field and should be called `generator`. Templates could continue to expose `GeneratorURL` as an alias for `Generator` under the hood.

mxinden:
> In that case, it is a free-form field and should be called generator.

@simonpasquier can you elaborate on that? Why can it not be a valid URI in that case?

mxinden:
I have addressed all outstanding comments.

Are there any objections or concerns left?

@brian-brazil with the above discussion, is the validation of `generatorURL` acceptable for you?

brian-brazil:
What exactly is the proposal to handle it? I can't quite figure out from the conversation.

mxinden:
The proposal is to validate that `generatorURL` is a valid URI in the  `POST /alerts` endpoint.

brian-brazil:
That breaks existing users, so we shouldn't do that.

mxinden:
@brian-brazil do you have any specific users in mind? Would it be possible for them to migrate to a valid `generatorURL` between the deprecation and the deletion date (multiple minor versions) of the API v1?

brian-brazil:
I'm not aware of anyone in particular, but as long as the v1 API exists I think we should keep them working on that api.

mxinden:
Maybe I have not been clear, sorry. This does not enforce validation on the API v1ingestion endpoint, only on the API v2 ingestion endpoint. The API v1 endpoint will still accept invalid `generatorURL`s.

@brian-brazil with that in mind, is the proposal acceptable for you?

mxinden:
@simonpasquier @stuartnelson3 does this PR look good to you as well?

simonpasquier:
:+1: for merging this version and iterate from there. Thanks!

brancz:
:+1:

stuartnelson3:
:+1:

stuartnelson3:
merge at will, @mxinden !

s-urbaniak:
:tada: :+1:

mxinden:
Thanks everyone for the repeated feedback and help!

Issue #1353, opened 2018-05-02
Invalid configuration template in WechatConfig


RogerYuQian:
It is invalid to configurate template in the section of WechatConfig.
Code show as below:
```
        // "notify/impl.go" 1384 lines
	msg := &weChatMessage{
		Text: weChatMessageContent{
			Content: tmpl(n.conf.Message),
		},
                 // The following fields have not been forcibly converted by tmpl. 
                 // As a result, the template does not take effect.
		ToUser:  n.conf.ToUser,
		ToParty: n.conf.ToParty,
		Totag:   n.conf.ToTag,
		AgentID: n.conf.AgentID,
		Type:    "text",
		Safe:    "0",
	}
```

stuartnelson3:
fixed by #1356

PR #1356, opened 2018-05-02
fix wechat issue (#1353)

RogerYuQian:
Invalid configuration template in WechatConfig


RogerYuQian:
Thank you for your approval，is it necessary to backport to previous version？

stuartnelson3:
Not necessary, since we don't officially support older versions in our current pre-1.0 state.

PR #1360, opened 2018-05-03
Update alerts argument order, rename expired to inhibited

stuartnelson3:
- rename `expired` to `inhibited`
- order for `silenced` and `inhibited` was reversed (cf. https://github.com/prometheus/alertmanager/blob/master/client/client.go#L155-L160)

the API is not properly filtering responses, but that's for a different PR.

@simonpasquier


simonpasquier:
:+1:

Issue #1357, opened 2018-05-02
can't set e-mail CC or Bcc headers


tzz:
**What did you do?**

I want to send copies of alert e-mails. I specify `CC` and `Bcc` under `headers`.

**What did you expect to see?**

CC or Bcc going to the specified addresses.

**What did you see instead? Under which circumstances?**

CC or Bcc get dropped and they do not get used. The addresses are valid and if they are added to the `to` header, they get the e-mail. In either case there's nothing in the `info` level logs.

**Environment**

* Alertmanager version:

	0.15.0-rc.1

* Alertmanager configuration file:

Working:

```
- name: my-receiver
  email_configs:
  - to: 'contact@mydomain.invalid, {{ template "alert.recipient.tzz" }}'
    headers:
      subject: 'my subject'
      Reply-To: '{{ template "alert.recipient.tzz" }}'
    html: '{{ template "quota" . }}'
```
Not receiving the CC (tried `CC`, `cc`):

```
- name: my-receiver
  email_configs:
  - to: 'contact@mydomain.invalid'
    headers:
      subject: 'my subject'
      CC: '{{ template "alert.recipient.tzz" }}'
      Reply-To: '{{ template "alert.recipient.tzz" }}'
    html: '{{ template "quota" . }}'
```

brian-brazil:
This is how SMTP has always worked. The headers are just for display, the envelope to is where the emails actually go.

It makes more sense to ask questions like this on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) rather than in a GitHub issue. On the mailing list, more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

tzz:
@brian-brazil thanks for explaining how SMTP works.

Sorry if my issue statement was confusingly phrased. I'm saying I can't do CC or Bcc with alertmanager, not that `headers` specifically should set them. There is no way, that I can, see, of requesting those in the `email_configs`. The only envelope settings I see are `to` and `from`. Are you saying CC and Bcc won't be supported by closing this issue?

roidelapluie:
To send an email with a CC: in alertmanager, it is not sufficient to add a CC: header.

```
receivers:
- name: my-receiver
  email_configs:
  - to: 'one@foo.com'
    headers:
      subject: 'my subject'
      CC: 'two@bar.com'
```

You also need to add the CC: address to the to: field and explicitely add a to:
field.

```
receivers:
- name: my-receiver
[...]

PR #1361, opened 2018-05-03
doc/examples/simple.yml: add email_configs CC and Bcc examples

tzz:
Adding CC and Bcc examples following discussion in #1357 thanks to advice from @brian-brazil and @roidelapluie 

@simonpasquier I hope you agree the examples directory is a good place for this.

### DIFF
--- a/doc/examples/simple.yml
+++ b/doc/examples/simple.yml
@@ -108,6 +108,27 @@ receivers:
   email_configs:
   - to: 'team-Y+alerts@example.org'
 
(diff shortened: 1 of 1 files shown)


perrefe:
What about more than one Cc or Bcc ? It's possible?

roidelapluie:
Thank you, this is nice. I'd like to move the comments closer to the example, e.g. explictly say that the second snippet is about bcc.

PR #1362, opened 2018-05-04
api: Deprecate `api/alerts` endpoint

mxinden:
With prometheus/prometheus commit
e114ce0ff7a1ae06b24fdc479ffc7422074c1ebe [1] Prometheus switches from
using `api/alerts` to `api/v1/alerts`. This commit is included starting
from Prometheus v0.17.0. As discussed on the prometheus-developers
mailing list [2] the deprecation period is long over.

[1] github.com/prometheus/prometheus/commit/e114ce0ff7a1ae06b24fdc479ffc7422074c1ebe
[2]
https://groups.google.com/d/msg/prometheus-developers/2CCuFTMbmAg/Qg58rvyzAQAJ

//CC @stuartnelson3  @brian-brazil

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -133,16 +133,10 @@ func (api *API) Register(r *route.Router) {
 
 	r.Options("/*path", wrap(func(w http.ResponseWriter, r *http.Request) {}))
 
-	// Register legacy forwarder for alert pushing.
-	r.Post("/alerts", wrap(api.legacyAddAlerts))
-
-	// Register actual API.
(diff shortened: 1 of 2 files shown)


api/api.go
```diff
@@ -133,16 +133,10 @@ func (api *API) Register(r *route.Router) {
 
 	r.Options("/*path", wrap(func(w http.ResponseWriter, r *http.Request) {}))
```
mxinden: This will now stretch not `api/*` but only `api/v1/*`.

PR #1366, opened 2018-05-07
api: support more query filters

### LINKED ISSUE #1359: silenced and inhibited filters for listAlerts endpoint don't work
**What did you do?**

Attempted to list only inhibited or only silenced alerts from the api:

http://alertmanager.url.biz/api/v1/alerts?silenced=false&inhibited=true
http://alertmanager.url.biz/api/v1/alerts?silenced=true&inhibited=false

Only inhibited or silenced alerts
`inhibited=true`: Active, inhibited, and unprocessed alerts
`silenced=true`: Active, silenced, and unprocessed alerts

* Alertmanager version: current master

simonpasquier:
Fixes #1359.

This change adds 2 new query filters to the /api/v1/alerts endpoint.

- active, filter out active alerts when set to 'false' (default: 'true').
- unprocessed, filter out unprocessed alerts when set to 'false'
 (default: 'true').

The default values ensure that the API behavior remains the same as
before when the query filters aren't provided.

cc @stuartnelson3

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -225,7 +225,7 @@ func TestListAlerts(t *testing.T) {
 			false,
 			map[string]string{"filter": "{alertname"},
 			400,
-			[]string{"alert1", "alert2", "alert3", "alert4"},
+			[]string{},
 		},
 		{
 			false,
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -12,6 +12,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/prometheus/alertmanager/config"
 	"github.com/prometheus/alertmanager/dispatch"
 	"github.com/prometheus/alertmanager/provider"
 	"github.com/prometheus/alertmanager/types"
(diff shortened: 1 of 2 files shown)


api/api_test.go
```diff
@@ -20,26 +21,72 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// fakeAlerts is a struct implementing the provider.Alerts interface for tests.
```
stuartnelson3: :100:

api/api_test.go
```diff
@@ -88,6 +136,167 @@ func TestAddAlerts(t *testing.T) {
...
+			false,
+			map[string]string{"filter": "{alertname=\"alert3\""},
+			200,
+			[]string{"alert3"},
+		},
+		{
+			false,
+			map[string]string{"filter": "{alertname"},
+			400,
+			[]string{"alert1", "alert2", "alert3", "alert4"},
```
stuartnelson3: this `[]string{}` should probably be empty, since it's not being executed anyway because of the `400`.

api/api_test.go
```diff
@@ -88,6 +136,167 @@ func TestAddAlerts(t *testing.T) {
 	}
 }
 
+func TestListAlerts(t *testing.T) {
```
stuartnelson3: :heart_eyes:

PR #1376, opened 2018-05-10
cli: print more config details

adamdecaf:
Not everything about a config was printed before. 

Example output:

```
$ amtool check-config alertmanager.yaml
Checking 'alertmanager.yaml'  SUCCESS
Found:
 - global config
 - route
 - 0 inhibit rules
 - 13 receivers
 - 0 templates
```


adamdecaf:
@stuartnelson3 Good call. Updated.

PR #1379, opened 2018-05-13
Add additional cluster configuration flags

rhysm:
The cluster configuration uses DefaultLANConfig which seems to be quite sensitive to WAN conditions. 

Allowing the tuning of these 3 parameters (TCP Timeout, Probe Interval and Probe Timeout) makes
clustering more robust across WAN connections.

Default values remain the same as the DefaultLANConfig.

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,6 +301,10 @@ be configured to communicate with each other. This is configured using the
   convergence speeds at expense of bandwidth (default "1m0s")
 - `--cluster.settle-timeout` value: maximum time to wait for cluster
   connections to settle before evaluating notifications.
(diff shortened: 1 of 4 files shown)


brian-brazil:
The defaults seem pretty low for a WAN, we should probably bump them.

rhysm:
The defaults from memberlist for WAN config are:

- TCP timeout - 30s
- Probe timeout - 3s
- Probe interval - 5s

Previously I was seeing dead peers within 24 hours, with these values it's been over a week with no problems so far.

brian-brazil:
Those sound better to me. @stuartnelson3?

PR #1364, opened 2018-05-05
*: Cut v0.15.0-rc.2

mxinden:
Update `CHANGELOG.md` and bump `VERSION` to `v0.15.0-rc.2`.

PR #1363, opened 2018-05-05
cherry-pick bug fixes and cli patches into release-0.15

mxinden:
I have cherry-picked all bug fixes as well as all CLI related patches into the `release-0.15` branch. I would like to not include any breaking changes or new features between two release candidates. Thereby I have not included the following commits:

- b95b328 ui: replace deprecated InstrumentHandler() (#1302)
- f53b247 api: initialize alerts_received_total labels (#1310)
- 300a87e Removed file changes to resolve conflict. (#1318)
- bc263d3 Improve notification instrumentation (#1335)
- af5dd74 tell users opening issues to use alertmanager  --version (#1327)
- cecfe5b Validate Slack field config and only allow the necessary input (#1334)
- b04e9ad #1346: move maintenance messages to DEBUG log level (#1347)

The above mentioned patches will be included in the upcoming `v0.16.0` release.

Let me know what you think, whether I am missing something, or whether you disagree with this process proposal. I am happy to take care of the `v0.16.0` release in the upcoming weeks.

Version bump will follow in an upcoming PR.

PR #1384, opened 2018-05-15
memberlist reconnect

stuartnelson3:
add reconnection support for dead peers

todo:

- [x] tests (only verified manually so far)
- [x] metrics
- [x] cleanup extra unused bits (most of this code is taken from https://github.com/hashicorp/serf/blob/80ab48778d/serf/serf.go)
- [x] decide on "smart" defaults for `DefaultReconnectInterval` and `DefaultReconnectTimeout`

edit:

I also included the `logWriter{}` wrapper I was using to expose memberlist logging. It's very verbose, and doesn't really conform to how we've been logging, so I'm not sure how best to expose it (or if I should just remove it from this PR).

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -153,6 +153,15 @@ func Join(
 		logger: l,
 		peers:  map[string]peer{},
 	}
+
+	if reconnectInterval != 0 {
+		go p.handleReconnect(reconnectInterval)
+	}
+	if reconnectTimeout != 0 {
+		go p.handleReconnectTimeout(5*time.Minute, reconnectTimeout)
+	}
+	p.register(reg)
+
 	p.delegate = newDelegate(l, reg, p)
 
 	cfg := memberlist.DefaultLANConfig()
@@ -189,14 +198,6 @@ func Join(
 	if n > 0 {
 		go p.warnIfAlone(l, 10*time.Second)
 	}
-
-	if reconnectInterval != 0 {
-		go p.handleReconnect(reconnectInterval)
-	}
-	if reconnectTimeout != 0 {
-		go p.handleReconnectTimeout(5*time.Minute, reconnectTimeout)
-	}
-	p.register(reg)
 
 	return p, nil
 }
@@ -329,6 +330,7 @@ func (p *Peer) peerJoin(n *memberlist.Node) {
 	}
 
 	p.peers[n.Name] = pr
+	p.peerJoinCounter.Inc()
 
 	if oldStatus == StatusFailed {
 		level.Debug(p.logger).Log("msg", "peer rejoined", "peer", pr.Node)
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -197,10 +197,6 @@ func Join(
 
 	p.setInitialFailed(resolvedPeers)
 
-	if n > 0 {
-		go p.warnIfAlone(l, 10*time.Second)
-	}
-
 	return p, nil
 }
 
@@ -407,22 +403,6 @@ func (p *Peer) peerUpdate(n *memberlist.Node) {
 
 	p.peerUpdateCounter.Inc()
 	level.Debug(p.logger).Log("msg", "peer updated", "peer", pr.Node)
-}
-
-func (p *Peer) warnIfAlone(logger log.Logger, d time.Duration) {
-	tick := time.NewTicker(d)
-	defer tick.Stop()
-
-	for {
-		select {
-		case <-p.stopc:
-			return
-		case <-tick.C:
-			if n := p.mlist.NumMembers(); n <= 1 {
-				level.Warn(logger).Log("NumMembers", n, "msg", "I appear to be alone in the cluster")
-			}
-		}
-	}
 }
 
 // AddState adds a new state that will be gossiped. It returns a channel to which
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -154,12 +154,6 @@ func Join(
 		peers:  map[string]peer{},
 	}
 
-	if reconnectInterval != 0 {
-		go p.handleReconnect(reconnectInterval)
-	}
-	if reconnectTimeout != 0 {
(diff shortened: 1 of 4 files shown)


cluster/cluster.go
```diff
@@ -32,15 +32,58 @@ type Peer struct {
...
-	DefaultTcpTimeout       = 10 * time.Second
-	DefaultProbeTimeout     = 500 * time.Millisecond
-	DefaultProbeInterval    = 1 * time.Second
+	DefaultPushPullInterval  = 60 * time.Second
+	DefaultGossipInterval    = 200 * time.Millisecond
+	DefaultTcpTimeout        = 10 * time.Second
+	DefaultProbeTimeout      = 500 * time.Millisecond
+	DefaultProbeInterval     = 1 * time.Second
+	DefaultReconnectInterval = 10 * time.Second
+	DefaultReconnectTimeout  = 6 * time.Hour
```
stuartnelson3: what should these values be?

stuartnelson3:
@simonpasquier

simonpasquier:
Thanks, I'll have a look!

stuartnelson3:
>when a peer restarts, the other peer still tries to reconnect even after the successful rejoin of the first one because the name (ULID) of the first peer has changed. One solution would be use the peer's address instead of its name as the key.

sounds good to me.

>with asymmetric configurations (eg peer A is started without --cluster.peer, peer B with --cluster.peer=<peer A>), peer B will never try to reconnect if A is down when B starts.

Ah, so checking the result of the initial `memberlist.Join` and adding any non-connected nodes to the `failedPeers` list. Makes sense.

stuartnelson3:
updated. feel free to suggest a better way to grab the nodes that we failed to connect to.

cluster/cluster.go
```diff
@@ -140,13 +195,84 @@ func Join(
...
+		return float64(len(p.failedPeers))
+	})
+	p.failedReconnectionsCounter = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "alertmanager_cluster_failed_reconnections_total",
+		Help: "A counter of the number of failed cluster peer reconnection attempts.",
+	})
+
+	p.reconnectionsCounter = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "alertmanager_cluster_reconnections_total",
+		Help: "A counter of the number of successful cluster peer reconnections.",
```
simonpasquier: `successful` to be removed.

cluster/cluster.go
```diff
@@ -140,13 +195,84 @@ func Join(
...
+		Name: "alertmanager_cluster_failed_peers",
+		Help: "Number indicating the current number of failed peers in the cluster.",
+	}, func() float64 {
+		p.peerLock.RLock()
+		defer p.peerLock.RUnlock()
+
+		return float64(len(p.failedPeers))
+	})
+	p.failedReconnectionsCounter = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "alertmanager_cluster_failed_reconnections_total",
```
simonpasquier: `alertmanager_cluster_reconnections_failed_total` might be more idiomatic?

cluster/cluster.go
```diff
@@ -106,7 +151,17 @@ func Join(
 		stopc:  make(chan struct{}),
 		readyc: make(chan struct{}),
 		logger: l,
+		peers:  map[string]peer{},
+	}
+
+	if reconnectInterval != 0 {
```
simonpasquier: It could be done at the very end of the function once it is certain that it will return without error.

cluster/cluster.go
```diff
@@ -140,13 +195,84 @@ func Join(
 		level.Debug(l).Log("msg", "joined cluster", "peers", n)
 	}
 
-	if n > 0 {
-		go p.warnIfAlone(l, 10*time.Second)
-	}
+	p.setInitialFailed(resolvedPeers)
```
simonpasquier: IMO it would be simpler to initialize `p.failedPeers` with all known peers before calling `ml.Join(...)`.

cluster/cluster.go
```diff
@@ -155,13 +281,130 @@ func (p *Peer) warnIfAlone(logger log.Logger, d time.Duration) {
...
+}
+
+func (p *Peer) peerUpdate(n *memberlist.Node) {
+	p.peerLock.Lock()
+	defer p.peerLock.Unlock()
+
+	pr, ok := p.peers[n.Address()]
+	if !ok {
+		// Why are we receiving an update from a node that never
+		// joined?
```
simonpasquier: Could be that the process has restarted and receives an out-of-bound notification?

cluster/cluster.go
brian-brazil: Reconnection should probably be indefinite, for as long as SD returns the AM.

stuartnelson3: Peers only exist from starting the binary (one of the `--cluster.peer` flags), or as an instance that connects to a running AM later. AM doesn't do any form of SD lookup to find its peers, so I think there needs to be some form of timeout since we have no way of knowing if a former peer is unreachable or has ceased to exist.

cluster/cluster.go
stuartnelson3: that's WAAAY better :)

stuartnelson3:
@simonpasquier I think I've addressed all your comments, let me know what you think

stuartnelson3:
Going to merge this so we can get another rc out

mxinden:
@stuartnelson3 Let me know if you want me to follow up with https://github.com/prometheus/alertmanager/pull/1363 and https://github.com/prometheus/alertmanager/pull/1364.

stuartnelson3:
as soon as I feel confident about https://github.com/prometheus/alertmanager/pull/1389 let's push this out. That's the only thing (in my mind) blocking 0.15.0

There were issues with message queueing (we were generating more messages than could be gossiped) that I need to resolve; I'll be working on it tomorrow.

PR #1386, opened 2018-05-17
cluster: advertise explicitly for empty addresses

simonpasquier:
While testing #1384 I found that the address advertised by memberlist isn't valid when the bind address is empty. cc @stuartnelson3


stuartnelson3:
sorry for the delay, re-running test

Issue #1387, opened 2018-05-17
Alertmanager memberlist not sending to all peers


stuartnelson3:
Investigating another duplicate (where the pipelines are firing at the correct time):

- peer0 sends correctly
- peer2 and peer3 receive the nflog notification
- peer1 does not receive the nflog notification. 15seconds pass, and it sends a duplicate.

After reading memberlist code, it seems that this is what's happening:

Every X sec, `memberlist.gossip()` is called https://github.com/hashicorp/memberlist/blob/master/state.go#L480

A random number of nodes is selected from the available members https://github.com/hashicorp/memberlist/blob/master/state.go#L485 (In our case, up to 3 nodes). From the documentation:

```
// kRandomNodes is used to select up to k random nodes, excluding any nodes where
// the filter function returns true. It is possible that less than k nodes are
// returned.
```

Queued messages are sent up to a certain limit (in our case, 3 times).

What I found in the logs is that peer0 gossiped its message to peer2 twice, and peer3 once. After sending its message 3 times, the message was removed from the broadcast queue. At this point, peer1 would only find out about this message as a result of a full state merge (which happens once every 60s).

The easiest solution is either to configure `config.GossipNodes` to an arbitrarily high number so that the probability of selecting all peers is essentially 1, or have alertmanagers propagate gossip messages that they haven't seen before, raising the probability (hopefully to 1) of all nodes seeing the message before `cluster.peer-timeout`.

I'm going to investigate both options.

stuartnelson3:
Configuring `config.GossipNodes` to an arbitrarily high value has no effect, as there is an upper bound on the number of nodes that will be gossiped to. Here's the peer selection code in question: https://play.golang.org/p/xOTwsHKyXkh

So, I'm adding message propagation from peers that haven't seen the message before. If a peer receives an nflog message and merges it into its internal state, then it will also gossip the message. This seems like a simple solution that will also halt propagation relatively quickly, but I might be missing something obvious. Am I missing something?

@brancz @fabxc @grobie @beorn7

simonpasquier:
Your explanation makes perfect sense. Have you tried increasing `TransmitLimitedQueue.RetransmitMult`? That being said, what you described seems like a valid solution.

stuartnelson3:
I changed `RetransmitMult` to scale based on the size of the list of `cluster.peer`s (with a minimum value of 3), and also set the nodes to gossip any messages they are receiving for the first time.

[...]

PR #1389, opened 2018-05-18
memberlist gossip

stuartnelson3:
meant to address https://github.com/prometheus/alertmanager/issues/1387

one current issue is that it seems `peer0` has a few notifications queued, but hasn't cleared them in over an hour. Some way of clearing old messages is probably needed.

The method for choosing `RetransmitMulti` is also something to be discussed. It's an important value, but I'm wary of exposing too many configuration options in alertmanager to alter memberlist. I think we need to come up with a "smart" value that also doesn't overwhelm the transmit queues .


cluster/cluster.go
```diff
@@ -107,7 +107,7 @@ func Join(
 		readyc: make(chan struct{}),
 		logger: l,
 	}
-	p.delegate = newDelegate(l, reg, p)
+	p.delegate = newDelegate(l, reg, p, len(knownPeers)/2)
```
simonpasquier: nit but I would do the computation here rather than add a new argument to `newDelegate()`.

stuartnelson3: you mean setting `RetransmitMulti` on p.delegate?

stuartnelson3:
>Do we want to re-gossip new/updated silences too?

Yeah, that would make sense

cluster/cluster.go
simonpasquier: Sorry for being inaccurate. I meant:

```go
retransmit := len(knownPeers)/2
if restransmit < 3 {
    retransmit = 3
}
p.delegate = newDelegate(l, reg, p, retransmit)
```

stuartnelson3:
Updated.

When running this branch at SC with only nflog entries being further gossiped, we were seeing messages queued up. Newer messages are sent preferentially by memberlist, but it's still an issue that we have an unbounded queue. Ideally, the queue would be able to empty itself, but that's not what I was seeing. I'll investigate this.

mxinden:
> The method for choosing RetransmitMulti is also something to be discussed. It's an important value, but I'm wary of exposing too many configuration options in alertmanager to alter memberlist. I think we need to come up with a "smart" value that also doesn't overwhelm the transmit queues .

I agree with keeping configuration options low (fow now).

What is your reasoning for:
```golang
retransmit := len(knownPeers) / 2
```

Thanks a lot for working on this!

stuartnelson3:
Note: The probability calculation I was doing was based on changing `GossipPeers`, not `RetransmitMult`. Both are important values, and I tried to factor them both in for this.

- `RetransmitMult` is how many times a message will be sent. Each send goes to one machine.
- `GossipPeers` is how many machines to gossip to in a single gossip interval.

Each gossip interval will send messages to all machines picked by `kRandomNodes`, but it will not necessarily exhaust all retransmits of a message; the message may remain in the queue to be sent at the next gossip interval. But, if `numNodes > retransmits`, then a message will only be sent to the first `N` nodes, where `N = retransmits`. I imagine this is why I was seeing messages being queued -- I was instructing memberlist to send a message many times, but based on the number of peers it was sending messages to, it wouldn't hit that retransmit limit and remove the message from its internal queue. Newer gossip messages would arrive, and the older ones could never be sent.

### Probability of a machine not being gossiped to during a gossip interval

**Caveat:** This might be wrong :)

Play along at:
https://gist.github.com/stuartnelson3/81ba1f750196e9fccaf2f7b23bdbf5ac

Tweaking the numbers available (probabilities are the chance of not receiving a message, and given as a percentage out of 100):

retransmits | GossipPeers | num AMs | p(NoGossip) | p(Gossip)
------------ | ------------- | -------------| -------------| -------------
1 | 3 | 3 |50.0030 | 25.0030
2 | 3 | 3 |2.6065 | 0.0679
3 | 3 | 3 |2.6065 | 0.0679
3 | 3 | 4 | 3.1591 | 0.0032
3 | 3 | 10 | 66.6667 | 2.6012
5 | 5 | 10 | 44.4444 | 0.0677

`p(Gossip)` is if each node is configured to gossip only newly received messages. Any message they've already seen will not be gossiped.

The testing was far from thorough given that I doubt anyone runs more than 10 instances, but it seemed like `len(knownPeers) / 2` was a good place to start. If more peers join, however, maybe this value should be updated.

Issue #331, opened 2016-04-28
Update the Architecture Diagram


mpchadwick:
The hand drawn one in the README doesn't exactly look great...

pracucci:
Is the hand-written architecture diagram still valid? Do you have any preference about a tool to "digitalize" it?

bigMacro:
I want to figure out the architecture, but the hand drawn is not exactly readable to me...

stuartnelson3:
The hand-written diagram looks to still be correct. Is there a preferred tool for making these diagrams? I'm all for a legible diagram of the internals.

bigMacro:
You can try OmniGraffle  (:

stuartnelson3:
This looks great!

>How to name the boxes inside the Dispatcher?

The dispatcher's main work is done within the loop inside `run()`. It is concerned with reading the incoming alerts (as you've indicated) and creating the aggregate groups (alerts that are grouped together as defined by `group_by` in the config). The aggregate group itself is a loop that executes a notification pipeline every `group_interval`. The first time an aggregate group is created, it waits `group_wait` before execution -- after that, it is always `group_interval`. So, I would probably say something like "group alerts" and try to indicate that each group executes a notification pipeline.

>How to name the boxes after the Router?

The Router is actually the Fanout stage, where the set of alerts is dispatched concurrently to a pipeline for every notification type for that receiver. There's the WaitStage, where a timer is set based on the server's position in the HA mesh (`--cluster.peer-timeout` * position), Dedup, Retry, and SetNotifies. SetNotifies sends a message to peers on the mesh saying what alerts it has sent out. So, an error from SetNotifies to the NotifyProvider, and the DedupStage queries the notify provider. It's hard to convey this in a single diagram since the DedupStage is querying the NotifyProvider for state that the NotifyProvider received from other alertmanagers :)

>Should the arrow from Silencer to Silence Storage be reversed?

The SilenceStage queries the SilenceProvider for data, but it doesn't actually change any data in the provider. I think the arrow should stay how it is.

There's one stage in the pipeline that's missing, the GossipSettleStage. It waits for the gossip's initial settling period, so that a rebooted alertmanager can receive recent notification messages from its peers before potentially firing duplicate alerts.

sgissi:
Thanks for the feedback @stuartnelson3. Below is the updated diagram, I switched the orientation to make it align better and have more space to grow the stages and add the clustering. Happy to adapt as needed, I followed the code but didn't twelve too deep.

[...]

PR #1394, opened 2018-05-24
Update Architecture diagram

### LINKED ISSUE #657: Architecture picture can have a better Font?
sorry, I can not read the lots of words in Architecture(https://github.com/prometheus/alertmanager/blob/master/doc/arch.jpg).
I think the font of prometheus will be better.(https://github.com/prometheus/prometheus/blob/master/documentation/images/architecture.svg)

Can we attach  a more explicit picture or something like that?

sgissi:
Update diagram from sketch to vector (fix #657 and #331)
Add draw.io XML source file.
Update README.md to display master doc/arch.jpg instead of specific branch.

Signed-off-by: Silvio Gissi <silvio@gissilabs.com>


README.md
```diff
@@ -335,7 +335,7 @@ Refer to [ui/app/CONTRIBUTING.md](ui/app/CONTRIBUTING.md).
 
 ## Architecture
 
-![](https://raw.githubusercontent.com/prometheus/alertmanager/4e6695682acd2580773a904e4aa2e3b927ee27b7/doc/arch.jpg)
+![](https://raw.githubusercontent.com/prometheus/alertmanager/master/doc/arch.jpg)
```
mxinden: Instead of having a fully qualified domain name here, why not just referencing the image relative to the repository?(`doc/arch.jpg`). You can see an example here: https://github.com/improbable-eng/thanos/blame/master/README.md#L1

Otherwise the `release-xxx` branch `README.md` would always point to the `master` branch image, which might diverge at some point.

sgissi:
Good point, updated the path. Prometheus use a SVG instead of JPG. Should we align to that? If yes, I can export the diagram in SVG and update the PR.

README.md
sgissi: Good call, PR updated with relative path in README.md.

mxinden:
I would prefer a svg.

sgissi:
@mxinden Thanks for the feedback, I noticed the font changed at draw.io for some reason, should be plain Arial. Updated the PR with SVG instead of JPG and pinned the font to be Arial.

sgissi:
Ok, now it should be ready to merge. Removed images from SVG and updated the draw.io XML file.

simonpasquier:
LGTM

mxinden:
Mhh, on Firefox 60.0.1 (64-bit) for Arch Linux the font seems to overflow the boxes. Anyone else seeing this?  Works fine on Chromium.

screenshot from 2018-05-31 13-55-43

simonpasquier:
The diagram renders fine for me on Firefox 60.0.1 for Fedora 28.

stuartnelson3:
`Version 60.0b12` on arch linux, the svg looks fine to me

PR #1396, opened 2018-05-28
Use Makefile.common from Prometheus

### LINKED ISSUE #1338: Use the main common makefile
https://github.com/prometheus/prometheus/blob/master/Makefile.common

simonpasquier:
Closes #1338

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -1,0 +1,100 @@
+# Copyright 2018 The Prometheus Authors
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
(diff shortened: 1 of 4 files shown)


simonpasquier:
@mxinden correct although `.PHONY` needs to include some targets that are only defined in `Makefile`.

Makefile.common
```diff
@@ -0,0 +1,100 @@
...
+promu:
+	GOOS= GOARCH= $(GO) get -u github.com/prometheus/promu
+
+$(FIRST_GOPATH)/bin/staticcheck:
+	GOOS= GOARCH= $(GO) get -u honnef.co/go/tools/cmd/staticcheck
+
+$(FIRST_GOPATH)/bin/govendor:
+	GOOS= GOARCH= $(GO) get -u github.com/kardianos/govendor
+
+.PHONY: all style check_license format build test vet assets tarball docker promu staticcheck $(FIRST_GOPATH)/bin/staticcheck govendor $(FIRST_GOPATH)/bin/govendor
```
mxinden: `govendor` is mentioned here, but I don't see a make target for it. Am I missing something?

simonpasquier: It's defined on L97.

simonpasquier: Pinging @krasi-georgiev but AFAICT it has been carried from the original Prometheus [Makefile](https://github.com/prometheus/prometheus/blob/f63e7db4cbdb616337ca877b306b9b96f7f4e381/Makefile#L106). I see no reason  to keep it but the change has to happen in prometheus/prometheus first.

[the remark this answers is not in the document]
simonpasquier: Yes this is my understanding.

Issue #1196, opened 2018-01-14
Feature request: restrict the identity of silence creator to one passed by some HTTP header


mitchellrj:
**What did you do?**
We configure a proxy in front of Alert Manager to authenticate, authorize and audit access, especially with regard to creating silences.

Currently a user can still enter any value they like for the "Creator" field of a silence through the web UI.

**What did you expect to see?**
It would be super if we could have an option to force the Creator value to one set via some HTTP header that is inserted into the request by the authenticating proxy. For example, if the proxy set the `X-Alertmanager-User` header, then the Alertmanager UI application would prepopulate the field and only permit values for "Creator" that matched that header value.

simonswine:
Ideally the header fields should be configurable as well. My identity proxy gives me: **`X-Auth-Request-Email`**

brian-brazil:
Any authorisation features would have to be implemented by a proxy in front of the alertmanager, as that's a rabbit hole we don't want to go down inside the project itself. I think defaulting the value of this field based on hardcoded header name would be okay.

Kellel:
I'm happy to spend some time working on this. I think it should be pretty simple to implement even with a configurable header.

Kellel:
Ok so I may have completely misunderstood the point of this feature request.

I have implemented a part of the api that overrides the createdBy field sent by the proxy, but I haven't endeavored to make the ui auto-populate the field yet. I guess that making the api return an error on mismatch might be better then just silently overriding the createdBy field.

So I guess I ask @mitchellrj @simonswine if I am on the right track? Or am I completely off in left field?

brian-brazil:
The original feature request was for forcing via a header, however as per https://github.com/prometheus/alertmanager/issues/1196#issuecomment-373429029 that's out of scope for the alertmanager. Returning a error if the user chooses to tweak the created by field is similar.

The best we can do is default the field value, and leave the rest to code outside the alertmanager.

mitchellrj:
Hi Kellen, first let me say thanks for stepping up and helping!

For some more context:

We run alert managerbl behind a proxy application that provides
authz/authn, and then passes authorised requests through to Alertmanager.
Along with the original request, the proxy application adds headers (e.g.
`X-Auth-User`). In order to have a sort of audit / non-repudiation function
around who has created / modified silences, we would like to force the
Creator field for end-user requests to always take the value of the header
passed through. That needn't be reflected in the UI, but would be much
clearer to the end user if it were. The threat scenario driving this desire
[...]

PR #1406, opened 2018-06-06
#1196 Implement ability to get createdBy from header

Kellel:
If a header is provided (most likely by a load balancer) that states who
the user is then we should use that value for the createdBy field for
proper accounting/auditability of silence creation.

The header used can be configured by setting the `auth_header` option in
the global context. The feature can be disabled by setting `auth_header:
""`.

The value set for `auth_header` must match the output from the go http
request parser

```
HTTP defines that header names are case-insensitive. The
request parser implements this by using CanonicalHeaderKey,
making the first character and any characters following a
hyphen uppercase and the rest lowercase.
```

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -523,6 +523,13 @@ func (api *API) setSilence(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// FEAT(#1196) Overwrite createdBy if value is passed in `AuthHeader` variable
+	// defaults to (x-auth-username)
+	creator, ok := r.Header[api.config.Global.AuthHeader]
+	if ok {
+		sil.CreatedBy = creator[0]
+	}
+
 	// This is an API only validation, it cannot be done internally
 	// because the expired silence is semantically important.
 	// But one should not be able to create expired silences, that
--- a/config/config.go
+++ b/config/config.go
@@ -342,6 +342,7 @@ var DefaultGlobalConfig = GlobalConfig{
 	ResolveTimeout: model.Duration(5 * time.Minute),
 	HTTPConfig:     &commoncfg.HTTPClientConfig{},
 
+	AuthHeader:      "X-Auth-Username",
 	SMTPHello:       "localhost",
 	SMTPRequireTLS:  true,
 	PagerdutyURL:    "https://events.pagerduty.com/v2/enqueue",
@@ -359,6 +360,8 @@ type GlobalConfig struct {
 	ResolveTimeout model.Duration `yaml:"resolve_timeout" json:"resolve_timeout"`
 
 	HTTPConfig *commoncfg.HTTPClientConfig `yaml:"http_config,omitempty" json:"http_config,omitempty"`
+
+	AuthHeader string `yaml:"auth_header,omitempty" json:"auth_header,omitempty"`
 
 	SMTPFrom         string `yaml:"smtp_from,omitempty" json:"smtp_from,omitempty"`
 	SMTPHello        string `yaml:"smtp_hello,omitempty" json:"smtp_hello,omitempty"`
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -320,6 +320,7 @@ func TestEmptyFieldsAndRegex(t *testing.T) {
 		Global: &GlobalConfig{
 			HTTPConfig:       &commoncfg.HTTPClientConfig{},
 			ResolveTimeout:   model.Duration(5 * time.Minute),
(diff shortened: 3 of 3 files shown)


brian-brazil:
This setting has nothing to do with authentication, so shouldn't have "auth" in the name - it's just a default. It also doesn't belong in the config file.

Though thinking on it, if you're setting this header couldn't you change the json request?

roidelapluie:
@brian-brazil setting a header is done by many reverse proxies implementations. Changing Json on the fly is less common.

brian-brazil:
The most this header should be doing is specifying a default value, and the name of the header should be hardcoded. If you want actual authorisation features, you'll need a proxy that understands the protocol.

Kellel:
I went with x-auth-username as default solely because I remember using it in the past for a similar usecase. 

I don't really care what the name of the header is, but this header _is_  expected to be passed from the system that does auth and is expected to be post authentication and authorization.

As far as exposing the field in the configfile I didn't originally plan to, but then was pleasantly surprised at how easy it was. So I went ahead and implemented it. I can remove it if desired.

brian-brazil:
> I went with x-auth-username as default solely because I remember using it in the past for a similar usecase.

I don't believe there's a standard header for this.

> I think we have a disconnect about what we think the point of this change is... It's my understanding that the header value should have a higher priority then the given json field.

I can see what you're trying to do, but that's not what was discussed in #1196. What you're adding here is a deployment-specific security feature, and those are out of scope for the Prometheus project as that starts us down the security rabbit hole. If you want to implement this you should do so via a reverse proxy that adjusts the silence creation requests, and which can also implement the rest of your authentication and authorisation requirements. To do effective authorisation with silence permissions you need to parse the request JSON anyway.

What you propose here is also not a great UX, as the user would still be required to provide a value for the creator by the UI - which would then be ignored. The feature discussed in #1196 is to provide a default value in the UI for this creator value (which is persisted in browser local storage), which removes that friction and is useful beyond your particular use case as mandatory fields are always a little annoying. If you propose a PR with that feature then we can review it.

Kellel:
Ok I see that I haven't implemented the ui features as requested, but I think what I have is pretty close to the intended purpose of the ticket.

Lets move this discussion back to the ticket to try and get the opinion of the original feature requester.

seanorama:
This appears to solve the auto-populating of "Creator". We also are craving that feature.

We are not concerned with limiting changes to the field after the "auto-populating" and that likely needs to be handle in separate code anyway.

**Could we proceed with work on this and discuss the "authorization" issues separately?**

@Kellel With the proposed code, is the header configurable? There is no standard header for this.

brian-brazil:
My previous comments on this still stand.

roidelapluie:
I think that now that we have the api v2 it should be straightforward for anyone to zrite their own UI on top of AM.

jorgelbg:
Considering that alertmanager doesn't deal with authentication and that the v2 of the API already supports `createdBy` from the payload perhaps what it is missing its support on the UI to populate the inputs with the provided value, which could be encoded in the URL in the same way that we pass `filter`.

I mean this is not intended to be a security/logging feature right? More a commodity for creating silences, the same way that remembering the creator in the browser local storage works.

brian-brazil:
@jorgelbg Pretty much. I'm not sure how we'd plumb the header through to the UI, I don't believe it's  directly accessible from javascript. I guess we could inject it into the html at http serving time.

jorgelbg:
@brian-brazil Do we really need it to be a header? I was thinking more in the line of passing it as a query parameter in the URL. It is already done for when we click the <kbd>Silence</kbd> button/link on the alerts list. The URL looks like: 
```
#/silences/new?
filter=%7Balertname%3D%22ExampleAlertAlwaysFiring%22%2C%20job%3D%22pushgateway%22%7D
```
IMHO this should suffice for setting `createdBy` in a programmatic way, for more complicated use cases the API could be used. We could even allow setting all the inputs of the "New silence" form from URL query parameters.

brian-brazil:
It'd be coming from a header, and I suspect the relevant proxies have more ability to work with headers than URL parameters. How we get that from a HTTP header into the UI is the question, as javascript has no header access and those buttons are created entirely in javascript.

> We could even allow setting all the inputs of the "New silence" form from URL query parameters.

I personally like the sound of that idea, but don't think it helps here.

SuperQ:

copolycube:
Hello @SuperQ , what is the process to get such MR reviewed and merged ?
I've been waiting for this feature for a long time.

PR #1407, opened 2018-06-07
fix set initial failed peers

stuartnelson3:
initially failed peers weren't being set correctly, which caused a panic (!) and prevented them from being removed.

@simonpasquier


simonpasquier:
@stuartnelson3 will have a look!

stuartnelson3:
updated

simonpasquier:
:+1:

PR #1416, opened 2018-06-15
cluster: don't track FQDN addresses as inital peers

simonpasquier:
When `resolvePeers()` can't resolve some of the peers' addresses, `setInitialFailed()` can get passed textual addresses. Those need to be filtered out.

cc @stuartnelson3


simonpasquier:
Thanks Stuart!

PR #1418, opened 2018-06-15
cluster: prune the queue if it contains too many items

### LINKED ISSUE #1409: Set max queue size for memberlist transmit queue
The queue of gossip messages is currently unbounded within alertmanager. An example implementation for dealing with this can be seen here: https://github.com/hashicorp/serf/commit/74bdd15c

During a recent network partition, an isolated alertmanager was queueing gossip messages without bound that could never be sent:

image

To not expose too many knobs, choose a single value to prune the queue to.

simonpasquier:
Closes #1409

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -95,7 +95,6 @@ const (
 	DefaultProbeInterval     = 1 * time.Second
 	DefaultReconnectInterval = 10 * time.Second
 	DefaultReconnectTimeout  = 6 * time.Hour
-	DefaultMaxQueueSize      = 100
 	maxGossipPacketSize      = 1400
 )
 
@@ -113,7 +112,6 @@ func Join(
 	probeInterval time.Duration,
 	reconnectInterval time.Duration,
 	reconnectTimeout time.Duration,
-	maxQueueSize int,
 ) (*Peer, error) {
 	bindHost, bindPortStr, err := net.SplitHostPort(bindAddr)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


mxinden:
> I'm also unsure about exposing the max queue size as a flag -- I think we're exposing a lot of flags, and if we can make a smart choice instead of having users wonder whether they need to tweak the setting or not, I would be in favor of that.

Agreed. I would prefer to only introduce the flag on request.

Do you see this as a blocker for the `v0.15.0-rc.3`?

With https://github.com/prometheus/alertmanager/pull/1415 merged, what are other reasons for messages being queued up?

stuartnelson3:
>With #1415 merged, what are other reasons for messages being queued up?

Messages will always be put on the internal queue to be sent out, but they should be getting regularly removed from the queue and sent.

I would prefer to have a large prune limit so that we don't accidentally delete a bunch of messages during memberlist's normal operation. The default in serf is 4096 if I remember correctly.

simonpasquier:
> I would prefer to have a large prune limit so that we don't accidentally delete a bunch of messages during memberlist's normal operation. The default in serf is 4096 if I remember correctly.

Definitely. Also I'll increase the prune interval to something bigger (15m is my thinking) because
1) it doesn't need to run often. As we said messages held in the queue are fairly small and it isn't a big deal to hold tens of thousands in memory for a couple of minutes (worst case).
2) gossip is a best effort: if you create/update lots of silences rapidly, it can take a while to replicate them. By default a push-pull operation happens every 1m but this interval can be changed by the user.

stuartnelson3:
Looks great, thanks for all the awesome work!

@mxinden I won't be able to do anything more with alertmanager for the next two weeks. When you're ready to release the rc/full 0.15.0, go for it!

PR #1422, opened 2018-06-19
Added functionality to alter Webhook message

victor-cr:
To extend Webhook functionality it is nice to have a possibility to change JSON via template or static text.

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -344,6 +344,9 @@ type WebhookConfig struct {
 
 	// URL to send POST request to.
 	URL string `yaml:"url" json:"url"`
+
+	// JSON message which will be sent instead of serialized data object
+    Message string `yaml:"message,omitempty" json:"message,omitempty"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -147,7 +147,11 @@ type WebhookMessage struct {
 
 // Notify implements the Notifier interface.
 func (w *Webhook) Notify(ctx context.Context, alerts ...*types.Alert) (bool, error) {
-	data := w.tmpl.Data(receiverName(ctx, w.logger), groupLabels(ctx, w.logger), alerts...)
+	var err error
+    var (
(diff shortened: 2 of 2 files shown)


brian-brazil:
Thanks for your PR, can you explain what you're trying to do here?

victor-cr:
I wanted to have control on the message body of Webhook. So I added an extra field into webhook_config to allow people modify current JSON to the format external system is expecting. With the change you are able to either put static text as message body or generate custom JSON by templates.

brian-brazil:
The intended way to do arbitrary integrations is to write something designed to take in the webhook, and then it will send out an appropriately encoded request to the system you wish to integrate with. It's not intended that the webhhook will directly integrate with arbitrary other protocols, as that would be intractable from a maintenance standpoint.

victor-cr:
I understand that it may be not the purpose of the alert manager. However, I think of several reasons why it might make sense:
1. Other configurations like Slack, Email, PagerDuty have an ability to change message content
2. Sometimes alert data contains sensitive information you would like to not to share for all webhooks
3. This change will help people to remove extra layers of integration software between alert system and the target system with already implemented RESTful API (like JIRA, Confluence, etc)
4. It is still optional. If configuration does not declare the field "message" it will work as you expect

brian-brazil:
> Other configurations like Slack, Email, PagerDuty have an ability to change message content

They do, but creating a generic integration that can integrate with anything is far more involved than what this PR would add. We have chosen that our integration point is the webhook as-is, and do not wish to have more than one generic integration as that'd kinda defeats the purpose of having a single generic integration. Unfortunately we can't support everything out of the box.

> Sometimes alert data contains sensitive information you would like to not to share for all webhooks

If the data is that sensitive, it would be unwise to send it to the Alertmanager at all as it's all acessible via the UI. Webhooks only get the alerts that are routed to them.

victor-cr:
Fine, I am not insisting on the change, it was quite helpful to me and I just wanted to share so it may be useful for others. If you consider that the change is not proper in terms of design view - it should be rejected.
Anyway, I am grateful for your time and patience.

Issue #1434, opened 2018-06-23
Needs RFC1918 Private Address or --cluster.advertise-address flag to start alertmanager


skorpy2009:
**What did you do?**
Try to start the alertmanager. I used `/opt/alertmanager-0.15.0.linux-amd64/alertmanager --config.file /etc/prometheus/alertmanager.yml`
**What did you expect to see?**
See the alertmanager start.
**What did you see instead? Under which circumstances?**
Alertmanager did not start. It started when I added `192.168.2.2/24` to any of my interfaces or when I added the `--cluster.advertise-address` flag.
It seems like it has something to do with: https://github.com/hashicorp/memberlist/blob/022f081/net_transport.go#L144

**Environment**

* System information:

	Linux 4.16.0-0.bpo.2-amd64 x86_64

* Alertmanager version:

```
alertmanager, version 0.15.0 (branch: HEAD, revision: 462c969d85cf1a473587754d55e4a3c4a2abc63c)
  build user:       root@bec9939eb862
  build date:       20180622-11:58:41
  go version:       go1.10.3
```

* Logs:
```
level=info ts=2018-06-23T06:35:35.484534947Z caller=main.go:174 msg="Starting Alertmanager" version="(version=0.15.0, branch=HEAD, revision=462c969d85cf1a473587754d55e4a3c4a2abc63c)"
level=info ts=2018-06-23T06:35:35.484603566Z caller=main.go:175 build_context="(go=go1.10.3, user=root@bec9939eb862, date=20180622-11:58:41)"
level=info ts=2018-06-23T06:35:35.494030639Z caller=cluster.go:155 component=cluster msg="setting advertise address explicitly" addr=<nil> port=9094
level=error ts=2018-06-23T06:35:35.494578525Z caller=main.go:201 msg="Unable to initialize gossip mesh" err="create memberlist: Failed to get final advertise address: Failed to parse advertise address \"<nil>\""
```

mxinden:
@skorpy2009 Thanks for reporting this issue.

Would you mind adding more details to "I added 192.168.2.2/24 to any of my interfaces"? It is surprising to me, that your system can not find its IP address or determine one.

roidelapluie:
I guess OP's server just has public IP addresses

simonpasquier:
@skorpy2009 can you share the IP addresses (eg `ip addr` output)? Looking at the code, I don't understand how it ended up with a `<nil>` address for the advertise address.

[...]

PR #1437, opened 2018-06-25
cluster: fail when no private address can be found

simonpasquier:
Relates to #1434.

### DIFF
--- a/cluster/advertise.go
+++ b/cluster/advertise.go
@@ -49,7 +49,7 @@ func calculateAdvertiseAddress(bindAddr, advertiseAddr string) (net.IP, error) {
 			return nil, errors.Wrap(err, "failed to get private IP")
 		}
 		if privateIP == "" {
-			return nil, errors.Errorf("no private IP found, explicit advertise addr not provided")
+			return nil, errors.New("no private IP found, explicit advertise addr not provided")
 		}
 		ip := net.ParseIP(privateIP)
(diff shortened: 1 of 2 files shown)


cluster/advertise.go
```diff
@@ -39,12 +44,12 @@ func calculateAdvertiseAddress(bindAddr, advertiseAddr string) (net.IP, error) {
 	}
 
 	if isAny(bindAddr) {
-		privateIP, err := sockaddr.GetPrivateIP()
+		privateIP, err := getPrivateAddress()
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to get private IP")
 		}
 		if privateIP == "" {
-			return nil, errors.Wrap(err, "no private IP found, explicit advertise addr not provided")
```
simonpasquier: `err` being nil, `errors.Wrap(err, ...)` was also nil.

cluster/advertise.go
```diff
@@ -39,12 +44,12 @@ func calculateAdvertiseAddress(bindAddr, advertiseAddr string) (net.IP, error) {
...
+			return nil, errors.Errorf("no private IP found, explicit advertise addr not provided")
```
mxinden: Why not use `errors.New` here?

simonpasquier: No particular reason. I'll update.

PR #1450, opened 2018-07-02
Allow router matching of annotations to the alert as well as labels

mb-m:
This allows for matching (and re-dispatching) an alert where annotations have
changed. Possible use case is an auto-remediation system which tags the alert
for human escalation (and re-injects with the same fingerprint (ie. labels))

@stuartnelson3 I've not discussed this because this seems trivial as per the guidelines, please let me know if it's not.


mb-m:
I'm not sure why the acceptance tests fail. On my local (go 1.10.3) the cluster test fails but the acceptance test passes.

mb-m:
cc @brian-brazil, @simonpasquier, @stuartnelson3 no thoughts on this?

brian-brazil:
Routing is only possible on labels, as that's the whole point of distinguishing labels from annotations. If you have an annotation that you want to match on, change it to be a label.

stuartnelson3:
Sorry, I didn't see this! Thanks for the contribution.

My initial reaction is what @brian-brazil has said. The separation between labels and annotations was (I believe) based on wanting to route on labels, and not annotations.

Routing based on labels+annotations in the situation you described, while it would still result in the same fingerprint (because the labels haven't changed), would result in some funny behavior as it's implemented here. `alert_group`s are defined by their route + fingerprint. Matching to a different route would result in a new `alert_group` being created, which you would then have to inhibit to prevent the original alert from still firing. This is based on labels, which haven't changed (because you've only changed the annotations). The result is that any user making use of this feature would be double-alerting themselves, which we want to avoid.

This is a roundabout way of saying that you can achieve your desired result by changing a label and then adding an inhibition to your config :)

stuartnelson3:
sounds interesting! I would be interested in hearing what you find.

PR #1452, opened 2018-07-02
Lookup parts in strings using regexp.MatchString

wbiller:
@stuartnelson3 For more complex templates it is sometimes required to check whether a string contains only specific parts. Therefore I added strings.Contains/strings.ContainsAny to the FuncMap

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -128,6 +128,8 @@ var DefaultFuncs = FuncMap{
 	"join": func(sep string, s []string) string {
 		return strings.Join(s, sep)
 	},
+	"containsAny": strings.ContainsAny,
+	"contains": strings.Contains,
 	"safeHtml": func(text string) tmplhtml.HTML {
 		return tmplhtml.HTML(text)
 	},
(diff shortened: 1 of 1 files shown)


brian-brazil:
To maintain consistency with Prometheus, this would be better done with a `match` function: https://prometheus.io/docs/prometheus/latest/configuration/template_reference/#strings

brian-brazil:
Thanks! Can you update the docs over in the docs repo?

szediktam:
Can I use regexp.MatchString to make it? I use alertmanager v0.12. But I need this function.
How can I make it. 

What are the functions of go template supported in alertmanger? How to use them?  @wbiller  @brian-brazil 
Thank you.

simonpasquier:
@szediktam the list of functions available is in the documentation: https://prometheus.io/docs/alerting/notifications/#functions
This change will only be available in the next version of AlertManager (not yet released).

simonpasquier:
@szediktam 

> Only the functions in https://prometheus.io/docs/alerting/notifications/#functions and
http://golang.org/pkg/text/template/#hdr-Functions are available in alertmanager template?

yes

> What shoule i dd if I want to use some other functions? Can I create a custom function ?
I want to judge a string is in or not in another string.

You have to build your own binary including this change or wait for the next (non patch) release.

szediktam:
OK. Thank you. @simonpasquier

Issue #1449, opened 2018-07-01
Clustered Alertmanager instances not joining via gossip


povilasv:
**What did you do?**

Started clustered alertmanager (v.0.15) in Kubernetes and restarted all nodes.

Ref https://github.com/prometheus/alertmanager/pull/1428

**What did you expect to see?**

Eventually alertmanager instances joining into gossip.

**What did you see instead? Under which circumstances?**

The problem is that when you restart all of the alertmanager instances in an environment like Kubernetes, DNS may contain old alertmanager instance IPs, but on startup (when `Join()` happens) none of the new instance IPs. As at the start DNS is not empty `resolvePeers waitIfEmpty=true`, will return and "islands" of 1 alertmanager instances will form.

* Alertmanager version:

alertmanager:v0.15.0-rc.1

* Alertmanager configuration file:
```
apiVersion: v1
kind: Service
metadata:
  labels:
    name: alertmanager-peers
  name: alertmanager-peers
  namespace: sys-mon
spec:
  clusterIP: None
  ports:
  - name: cluster
    protocol: TCP
    port: 8001
    targetPort: cluster
  selector:
    app: alertmanager
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: alertmanager
  namespace: sys-mon
spec:
  replicas: 3
  selector:
    matchLabels:
      app: alertmanager
  template:
    metadata:
      name: alertmanager
      labels:
        app: alertmanager
    spec:
      containers:
      - name: alertmanager
        image: prom/alertmanager:v0.15.0-rc.1
        args:
          - --config.file=/etc/alertmanager/config.yml
          - --web.listen-address=0.0.0.0:9093
          - --storage.path=/alertmanager
          - --web.external-url=https://alertmanager.dev.uw.systems
          - --cluster.listen-address=0.0.0.0:8001
          - --cluster.peer=alertmanager-peers.sys-mon:8001
```

* Logs:

logs of alertmanager1:

```
level=info ts=2018-06-21T14:35:44.824688253Z caller=main.go:141 build_context="(go=go1.10, user=root@f278953f13ef, date=20180323-13:05:10)"
[...]

PR #1436, opened 2018-06-25
notify: catch templating errors for Wechat

simonpasquier:
Closes #1425

PR #1421, opened 2018-06-18
fix: email template typo in alert-warning style

palmerabollo:
Fixed minor issue in email template.

PR #1455, opened 2018-07-04
config: fix regression with Pager Duty

### LINKED ISSUE #1451: 0.15.0 Overwriting defaults in pagerduty_configs details broken
Hi this is a problem I experienced when trying to update to 0.15.0

**What did you do?**
I ran `alertmanager-0.15.0.linux-amd64/amtool --alertmanager.url example.com check-config /usr/local/example.yml`

I expected to get the same output as with `alertmanager-0.14.0.linux-amd64/amtool` which was:
`Checking '/usr/local/example.yml'  SUCCESS`

Instead, the output was: 

```
Checking '/usr/local/example.yml'amtool: error: failed to validate 1 file(s)
  FAILED: yaml: unmarshal errors:
  line 22: key "firing" already set in map
```

**Environment**

* Alertmanager version:

	alertmanager, version 0.15.0 (branch: HEAD, revision: 462c969d85cf1a473587754d55e4a3c4a2abc63c)

```
global:
  smtp_smarthost: 'localhost:25'

route:
  receiver: 'pagerduty-receiver'
  group_by: ['alertname', 'cluster']
  group_wait: 30s
  group_interval: 5m  
  repeat_interval: 3h
  routes:
  - receiver: pagerduty-receiver
    match:
      severity: critical

receivers:
- name: 'pagerduty-receiver'
  pagerduty_configs:
  - service_key: <some_key>
    details: {
      body: 'bar',
      firing: 'foo'
    }
```

From what I can see, switching to using strict yaml unmarshalling (https://github.com/prometheus/alertmanager/pull/1318) caused this error because strict yaml unmarshalling throws an error when trying to overwrite an existing key in a map: https://github.com/go-yaml/yaml/blob/v2/decode.go#L645.

Another issue is that `details` seems to be treated as the same map in multiple `pagerduty_configs`. So if I was to add a second receiver at the end:
```
- name: 'pagerduty-receiver2'
  pagerduty_configs:
  - service_key: <some_key>
    details: {
      body: 'bar2',
      firing: 'foo2'
    }
```
The error code is then:
```
Checking '/usr/local/example.yml'  FAILED: yaml: unmarshal errors:
  line 22: key "firing" already set in map
  line 28: key "body" already set in map
  line 29: key "firing" already set in map
```
because it is attempting to overwrite `body` in the same map.

simonpasquier:
Closes #1451

cc @CatherineBowen @Allex1 if one of you wants to try it.

@stuartnelson3 since it is a regression, it might be a case for a v0.15.1 release. WDYT?

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -42,6 +42,14 @@ var (
 	// DefaultEmailSubject defines the default Subject header of an Email.
 	DefaultEmailSubject = `{{ template "email.default.subject" . }}`
 
+	// DefaultPagerdutyDetails defines the default values for PagerDuty details.
+	DefaultPagerdutyDetails = map[string]string{
+		"firing":       `{{ template "pagerduty.default.instances" .Alerts.Firing }}`,
+		"resolved":     `{{ template "pagerduty.default.instances" .Alerts.Resolved }}`,
+		"num_firing":   `{{ .Alerts.Firing | len }}`,
+		"num_resolved": `{{ .Alerts.Resolved | len }}`,
+	}
+
 	// DefaultPagerdutyConfig defines default values for PagerDuty configurations.
 	DefaultPagerdutyConfig = PagerdutyConfig{
 		NotifierConfig: NotifierConfig{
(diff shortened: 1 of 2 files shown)


Allex1:
@simonpasquier I've cloned your fork, built the alertmanager from this commit, tested my pd config and it's working fine. Thanks!

Waiting for a release...

stuartnelson3:
@simonpasquier Let's get https://github.com/prometheus/alertmanager/pull/1456 merged as well, and I want to have a look at https://github.com/prometheus/alertmanager/issues/1449 also. But yeah we should get 0.15.1 out soon for this

mxinden:
> @mxinden want to do 0.15.1?

@stuartnelson3 sure. Happy to do that.

Shall we include #1421, #1456,  #1436 and #1437 as well?

simonpasquier:
Maybe #1447 too if it is merged soon?

stuartnelson3:
Merged, let's go for it

PR #1456, opened 2018-07-04
cluster: make sure we don't miss the first pushPull

iksaif:
During the join, memberlist initiates a pushPull to get initial data.
Unfortunately, at this point the nflog and silence listener have not
been registered yet, so the first data arrives only after one pushPull
cycle (1min by default !).

### DIFF
--- a/cluster/delegate.go
+++ b/cluster/delegate.go
@@ -135,6 +135,7 @@ func (d *delegate) NotifyMsg(b []byte) {
 		level.Warn(d.logger).Log("msg", "decode broadcast", "err", err)
 		return
 	}
+	level.Debug(d.logger).Log("received", "NotifyMsg", "len", len(b), "key", p.Key)
 	s, ok := d.states[p.Key]
 	if !ok {
 		return
@@ -160,6 +161,7 @@ func (d *delegate) LocalState(_ bool) []byte {
 	all := &clusterpb.FullState{
 		Parts: make([]clusterpb.Part, 0, len(d.states)),
 	}
+
 	for key, s := range d.states {
 		b, err := s.MarshalBinary()
 		if err != nil {
@@ -189,10 +191,10 @@ func (d *delegate) MergeRemoteState(buf []byte, _ bool) {
 	}
 	d.mtx.RLock()
 	defer d.mtx.RUnlock()
-
 	for _, p := range fs.Parts {
 		s, ok := d.states[p.Key]
 		if !ok {
+			level.Debug(d.logger).Log("received", "unknown state key", "len", len(buf), "key", p.Key)
 			continue
 		}
 		if err := s.Merge(p.Data); err != nil {
--- a/cluster/delegate.go
+++ b/cluster/delegate.go
@@ -135,7 +135,7 @@ func (d *delegate) NotifyMsg(b []byte) {
 		level.Warn(d.logger).Log("msg", "decode broadcast", "err", err)
 		return
 	}
-	level.Debug(d.logger).Log("received", "NotifyMsg", "len", len(b), "key", p.Key)
+
 	s, ok := d.states[p.Key]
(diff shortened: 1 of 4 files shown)


clems4ever:
LGTM

cmd/alertmanager/main.go
```diff
@@ -206,7 +204,7 @@ func main() {
 			cancel()
 			peer.Leave(10 * time.Second)
 		}()
-		go peer.Settle(ctx, *gossipInterval*10)
+		go peer.Settle(ctx, *pushPullInterval)
```
mxinden: If I understand the `Settle()` function correctly, it initially waits for `interval` before starting to check whether the cluster is settled. By increasing this interval (`gossipInterval` -> `pushPullInterval`) marking a cluster as settled is delayed for every setup, even though it might already be settled.

Why not going back to a low `interval` and moving the `peer.Settle` below the `peer.Join` logic?

cmd/alertmanager/main.go
```diff
@@ -263,6 +261,18 @@ func main() {
...
 	}()
 
+	// Peer state listener have been registered, now we can join and get the initial state.
+	if peer != nil {
+		err = peer.Join(
+			*reconnectInterval,
+			*peerReconnectTimeout,
+		)
+		if err != nil {
+			level.Error(logger).Log("msg", "Unable to initialize gossip mesh", "err", err)
```
simonpasquier: s/initialize/join/

cmd/alertmanager/main.go
```diff
@@ -263,6 +261,18 @@ func main() {
 		wg.Wait()
 	}()
 
+	// Peer state listener have been registered, now we can join and get the initial state.
```
simonpasquier: s/listener/listeners/

cluster/delegate.go
```diff
@@ -135,6 +135,7 @@ func (d *delegate) NotifyMsg(b []byte) {
 		level.Warn(d.logger).Log("msg", "decode broadcast", "err", err)
 		return
 	}
+	level.Debug(d.logger).Log("received", "NotifyMsg", "len", len(b), "key", p.Key)
```
simonpasquier: I'm not sure the extra logging is required.

cluster/delegate.go
```diff
@@ -189,10 +191,10 @@ func (d *delegate) MergeRemoteState(buf []byte, _ bool) {
 	}
 	d.mtx.RLock()
 	defer d.mtx.RUnlock()
-
 	for _, p := range fs.Parts {
 		s, ok := d.states[p.Key]
 		if !ok {
+			level.Debug(d.logger).Log("received", "unknown state key", "len", len(buf), "key", p.Key)
```
simonpasquier: Ditto or it should be `Warn()`.

cmd/alertmanager/main.go
simonpasquier: I would also expect `peer.Settle()` to be called after `peer.Join()`.

cluster/delegate.go
iksaif: done

cluster/delegate.go
iksaif: Done

cmd/alertmanager/main.go
iksaif: done

cmd/alertmanager/main.go
iksaif: done

cmd/alertmanager/main.go
iksaif: moved, re-changed the interval value

mxinden:
Closes https://github.com/prometheus/alertmanager/issues/1457

mxinden:
@iksaif Thanks for your help!

PR #1461, opened 2018-07-10
Add support for adding alerts using amtool

### LINKED ISSUE #1417: amtool: support adding alerts
It would be nice if `amtool` supported adding alerts.

The syntax is difficult though, as we need two lists of key-value pairs: the alert's labels, and its annotations.

An potential example, trying to be consistent with `amtool silence add`:

```
amtool alert add SomeAlertName node=bar severity=critical \
  --annotation=summary='this is the summary' \
  --annotation=runbook='http://runbook.biz'
```
The first argument, if not in the form `key=value`, is assumed to be the alert's name, with each following key-value pair making up the labels. Annotations are added by explicitly adding the `--annotation=` flag.

This is really verbose. I think it would be useful to add alerts via `amtool`, but am wondering if the community has ideas on how it should look.

Additionally, there should be support for resolving the alert. Maybe just `--resolve` flag to resolve immediately, or do we want to support setting an arbitrary time to resolve the alert? `--resolve=<date>`?

bobmshannon:
Closes #1417

cc @stuartnelson3

### DIFF
--- a/cli/alert_add.go
+++ b/cli/alert_add.go
@@ -1,0 +1,117 @@
+// Copyright 2018 Prometheus Team
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


cli/alert_add.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+	}
+	if a.end != "" {
+		endsAt, err = time.Parse(time.RFC3339, a.end)
+		if err != nil {
+			return err
+		}
+	}
+
+	err = alertAPI.Push(context.Background(), client.Alert{
```
stuartnelson3: You could just `return alertAPI.Push()` instead of setting it to `err` and then returning that.

simonpasquier:
:+1: thanks for doing this!

PR #1463, opened 2018-07-11
enable templating of hipchat room_id

markvdw4:
I noticed that I was unable to template the room_id field of hipchat_config, upon reviewing the code I found that tmplText was never being called on room_id despite it being declared as a tmpl_string in the documentation at https://prometheus.io/docs/alerting/configuration/#%3Chipchat_config%3E

This PR simply calls tmplText on RoomID before passing it into the url Sprintf

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -759,7 +759,8 @@ func (n *Hipchat) Notify(ctx context.Context, as ...*types.Alert) (bool, error) 
 		data     = n.tmpl.Data(receiverName(ctx, n.logger), groupLabels(ctx, n.logger), as...)
 		tmplText = tmplText(n.tmpl, data, &err)
 		tmplHTML = tmplHTML(n.tmpl, data, &err)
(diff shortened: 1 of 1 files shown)


simonpasquier:
LGTM you would to sign the commit to comply with the DCO.

```
git commit -s --amend && git push
```

markvdw4:
Thanks for the quick response @simonpasquier. I think this should be good to go now.

markvdw4:
@stuartnelson3 addressing you as per CONTRIBUTING.md

mmiller1:
It seems true that given this change auth token is still not able to be templated, yeah?

brian-brazil:
Yes, we only template auth tokens if they are used for routing.

mmiller1:
Is it intentional that we don't allow it here, or would a PR be welcome to add it?

brian-brazil:
It's intentional, as auth tokens and other secrets should not be stored in labels.

PR #1465, opened 2018-07-11
cluster: Do not exit when failing to join cluster

mxinden:
Alertmanager is exiting with a non-zero exit code if the initial cluster
join fails. This behavior could be not wanted because:

- As Alertmanager is a critical component with an at-least-once
guarantee, failing on joining the cluster is unnecessary as
Alertmanager still functions by itself.

- In an environment like Kubernetes discovering peers via DNS, peers
might roll out one-by-one, leaving the DNS entries unpopulated for the
first peer of a set. Failing on initial join prevents a roll-out.

Instead of failing on the initial join this patch only logs the failure.
The cluster can be later joined via the `handleReconnect`.

This is a regression introduced in PR #1456 [1].

[1] https://github.com/prometheus/alertmanager/pull/1456

//CC @stuartnelson3 @simonpasquier @brancz

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -217,6 +217,9 @@ func (p *Peer) Join(
 	n, err := p.mlist.Join(p.resolvedPeers)
 	if err != nil {
 		level.Warn(p.logger).Log("msg", "failed to join cluster", "err", err)
+		if reconnectInterval != 0 {
+			level.Info(p.logger).Log("msg", fmt.Sprintf("will retry joining cluster every %v", reconnectInterval.String()))
+		}
 	} else {
 		level.Debug(p.logger).Log("msg", "joined cluster", "peers", n)
 	}
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -196,7 +196,7 @@ func main() {
 			*probeInterval,
 		)
 		if err != nil {
-			level.Error(logger).Log("msg", "Unable to initialize gossip mesh", "err", err)
+			level.Error(logger).Log("msg", "unable to initialize gossip mesh", "err", err)
 			os.Exit(1)
 		}
 	}
@@ -262,8 +262,7 @@ func main() {
 			*peerReconnectTimeout,
 		)
 		if err != nil {
-			level.Error(logger).Log("msg", "Unable to join gossip mesh", "err", err)
-			os.Exit(1)
+			level.Warn(logger).Log("msg", "unable to join gossip mesh", "err", err)
 		}
 		ctx, cancel := context.WithTimeout(context.Background(), *settleTimeout)
 		defer func() {


cluster/cluster.go
```diff
@@ -217,6 +217,9 @@ func (p *Peer) Join(
 	n, err := p.mlist.Join(p.resolvedPeers)
 	if err != nil {
 		level.Warn(p.logger).Log("msg", "failed to join cluster", "err", err)
+		if reconnectInterval != 0 {
```
squat: Nit: might be clearer to say “will retry joining cluster in %v” otherwise it sounds a little funny IMO

stuartnelson3: "will retry joining cluster in" makes it sound like a one-shot, while this is repeated :man_shrugging:

simonpasquier: How about returning the error only when reconnectInterval is zero? This way, `main()` can still treat the error as a hard fail.

```go
if err != nil {
    if reconnectInterval == 0 {
        return err
    }
    level.Info(p.logger).Log("msg", fmt.Sprintf("will retry joining cluster in %v", reconnectInterval.String()))
}
```

mxinden: I was thinking about that as well. Good point.

Still I think a function called `Join` should fail if the join action failed.

>  This way, main() can still treat the error as a hard fail

I would prefer Alertmanager not to `exit(1)` on any cluster related failures. A partitioned Alertmanager cluster is still better than no Alertmanager cluster. What do you think @simonpasquier?

mxinden: I am fine with either way, but I would like to include the fact that this is a periodic action. @squat how can we combine the two?

simonpasquier: > A partitioned Alertmanager cluster is still better than no Alertmanager cluster.

I can hear that. It's definitely a balance between being functional and correct. Provided that there's an alert firing when the cluster's size is less than expected, it's indeed better to start in case of a partition.

squat: “Retrying every” to me implies that this is a permanent state, not just on failure. In that case, the correct place for this is right after the connection failure. That way the line is logged after every failure, e.g.:
```console
failed to join cluster
retrying in %v
failed to join cluster
retrying in %v
```

squat: Again it’s a nit, but sounds odd to me as a native English speaker. Not a blocker by any means.

mxinden: > that this is a permanent state, not just on failure.

Discussed offline with @squat, as this  is indeed a permanent periodic process we can move on.

mxinden: Do you still want me to do any follow up changes in this regards?

simonpasquier: No I'm fine as it is now.

Issue #1464, opened 2018-07-11
Support timeouts for amtool


stuartnelson3:
`amtool` currently uses `context.Background()` for all its API interactions, but we should start supporting timeouts to make `amtool` better for usage in scripting.

Add support for setting the timeout in the config file, and as a global flag that overwrites the config setting.

Use `model.ParseDuration()` from `github.com/prometheus/common/model`

PR #1471, opened 2018-07-16
[amtool] Add timeout support to amtool commands

stuartnelson3:
implements https://github.com/prometheus/alertmanager/issues/1464

the whole global `timeout` and `execWithTimeout` are because I don't know how to parse a global variable and then access it from or pass it in to the different subcommand actions. I'm not too upset about having a global timeout variable since it will only affect one command per execution, but I would prefer to pass it in explicitly instead of doing this lame closure thing in `execWithTimeout`.

so, if you're more familiar with kingpin's api and know how to do this, I'm all ears :)

### DIFF
--- a/cli/alert_add.go
+++ b/cli/alert_add.go
@@ -17,10 +17,11 @@ import (
 	"context"
 
 	"fmt"
+	"time"
+
 	"github.com/prometheus/alertmanager/client"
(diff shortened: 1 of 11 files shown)


simonpasquier:
LGTM. I have no better solution than what you offer. The global timeout variable is not worse than for the AlertManager's URL.

PR #1475, opened 2018-07-23
Email is green if no firing alerts

bchess:
@stuartnelson3 

Emails that contain no firing alerts should not be red. This makes them green.

#859 was a past (never merged) attempt to do something similar, but this is simpler


stuartnelson3:
Thanks!

PR #1481, opened 2018-07-25
notify: log PagerDuty v1 response on BadRequest

adamdecaf:
We had a problem (our fault) and PagerDuty was sending back 400's, but it was hard to debug without the error response from PagerDuty.


notify/impl.go
```diff
@@ -605,14 +605,20 @@ func (n *PagerDuty) Notify(ctx context.Context, as ...*types.Alert) (bool, error
...
 
-func (n *PagerDuty) retryV1(statusCode int) (bool, error) {
+func (n *PagerDuty) retryV1(resp *http.Response) (bool, error) {
 	// Retrying can solve the issue on 403 (rate limiting) and 5xx response codes.
 	// 2xx response codes indicate a successful request.
 	// https://v2.developer.pagerduty.com/docs/trigger-events
+	statusCode := resp.StatusCode
 	if statusCode/100 != 2 {
 		return (statusCode == 403 || statusCode/100 == 5), fmt.Errorf("unexpected status code %v", statusCode)
+	} else if statusCode == 400 && resp.Body != nil {
```
simonpasquier: This won't work since a 400 `statusCode` would satisfy the first test at L612.

notify/impl.go
```diff
@@ -605,14 +605,23 @@ func (n *PagerDuty) Notify(ctx context.Context, as ...*types.Alert) (bool, error
...
 	// 2xx response codes indicate a successful request.
 	// https://v2.developer.pagerduty.com/docs/trigger-events
+	statusCode := resp.StatusCode
+
+	if statusCode == 400 && resp.Body != nil {
+		bs, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return false, fmt.Errorf("unexpected status code %v | problem reading response: %v", statusCode, err)
+		}
+		return false, fmt.Errorf("bad request: %v", string(bs))
```
mxinden: Maybe still worth printing the status code here?

adamdecaf:
@mxinden There only looks to be a body on 400. There's no 429 on the v1 endpoint and 403 is blank.

https://v2.developer.pagerduty.com/v2/docs/trigger-events 

It might be noisy, but right now there's no real way to see what's wrong from the PD side.

mxinden:
@adamdecaf sounds good to me. Printing the body is a good idea, and probably more signal than noise to most users.

@simonpasquier @stuartnelson3 any further thoughts, otherwise I will go ahead and merge.

PR #1483, opened 2018-07-27
fix: Update PagerDuty API V2 to send full details on resolve

### LINKED ISSUE #1440: using pagerduty and event rule filters. resolve events lost due to lack of context/payload
**What did you do?**
Tried to use alertmanager with pagerduty and their new event rule filters, filtering events on custom details and routing to a specific service.

triggers and resolve events getting routed to a specific pagerduty service

triggers sent to service successfully, resolve events not picked up by event rule filter and lost

**Environment**

* Alertmanager version:

	0.14.0

I can see that you only add payload to the event when its a trigger event, this is lost for resolve event and therefore can't use to filter those events to the same service.

https://github.com/prometheus/alertmanager/blob/387e684faa6a79261490f10faf309093f9c3bed1/notify/impl.go#L489-L507

Any reason not to provide a merge request removing this if statement?

This documentation seems to suggest that parts of the payload are required when sent to the events v2 endpoint: 

https://v2.developer.pagerduty.com/v2/docs/send-an-event-events-api-v2

mikebryant:
This adds compatiblity with PagerDuty's Event rules feature, allowing resolve events to be routed based on attributes

Fixes #1440

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -521,9 +521,13 @@ func (n *PagerDuty) notifyV2(
 		n.conf.Severity = "error"
 	}
 
-	var payload *pagerDutyPayload
-	if eventType == pagerDutyEventTrigger {
-		payload = &pagerDutyPayload{
+	msg := &pagerDutyMessage{
+		Client:      tmpl(n.conf.Client),
+		ClientURL:   tmpl(n.conf.ClientURL),
+		RoutingKey:  tmpl(string(n.conf.RoutingKey)),
+		EventAction: eventType,
+		DedupKey:    hashKey(key),
+		Payload: &pagerDutyPayload{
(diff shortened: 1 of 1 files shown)


simonpasquier:
LGTM. I see that @stuart-warren already +1s. Lets wait a bit for @scarvel8 to weigh in. In case of no response, we can merge it in a few days.

mikebryant:
I just got notification that PagerDuty have made a change to how their API resolves events:

> (PagerDuty Support)Aug 1, 05:58 PDT
> Hi Mike,Thanks for your patience while we worked on this. Our engineering team has made an update that should now allow resolve events that are sent with only a dedup_key to resolve alerts through global event rules.

So it looks like this PR is no longer necessary, but likely wouldn't hurt.

PR #1482, opened 2018-07-27
Avoid listener blocking

waterdrink:
close #1469

PR #1486, opened 2018-07-30
config/test: Count `<secret>` occurrences via golang strings

mxinden:
`honnef.co/go/tools/cmd/staticcheck` complains with
`config/config_test.go:260:32: regular expression does not contain any
meta characters (SA6004)`. Instead of using a RegEx this patch simply
switches to using Golangs `strings.Count` function.

This patch fixes our current CI failures, e.g. seen in https://github.com/prometheus/alertmanager/pull/1482.


brian-brazil:
:+1:

PR #1501, opened 2018-08-05
Fix a bunch of unhandled errors

juliusv:
...as discovered by "gosec" (many other ones reported, but not all make
a lot of sense to fix).

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -59,8 +59,8 @@ func init() {
 	numReceivedAlerts.WithLabelValues("firing")
 	numReceivedAlerts.WithLabelValues("resolved")
 
-	prometheus.Register(numReceivedAlerts)
-	prometheus.Register(numInvalidAlerts)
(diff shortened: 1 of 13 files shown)


mxinden:
@juliusv @stuartnelson3 What are your thoughts on including https://github.com/kisielk/errcheck in our CI runs?

stuartnelson3:
My initial reaction is that it's probably not necessary. For a project that doesn't have a ton of contributors, I'm guessing we would spend more time dealing with special cases than actual value from this. Is it in CI for prometheus?

juliusv:
Yeah, so gosec also does these checks and produced a *ton* of errors that I didn't choose to fix because it was either obviously not sensical (like checking the error return of every logging call) or I wasn't sure whether it would make sense. `errcheck` has a function exclusion list, which would get rid of this, if someone wants to do it and maintain that list.

PR #1502, opened 2018-08-05
pagerduty: allow custom url for v1 api

adamalex:
On the pagerduty v1 api codepath, the target url is always forced
to the default pagerduty url. This commit changes the logic so the
switch to the default v1 pagerduty url only happens if the url
hasn't already been overridden in configuration.

@stuartnelson3 please let me know if there are any adjustments you'd like me to make and I will do my best. We really enjoy using alertmanager at salesforce, and this change allows us to use our own internal endpoint that is compatible with pagerduty v1 api. Thanks for taking a look!

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -477,7 +477,7 @@ func (n *PagerDuty) notifyV1(
 		Details:     details,
 	}
 
-	if n.conf.URL.String() == "https://events.pagerduty.com/v2/enqueue" {
+	if n.conf.URL == config.DefaultGlobalConfig.PagerdutyURL {
 		apiURL, err := url.Parse("https://events.pagerduty.com/generic/2010-04-15/create_event.json")
 		if err != nil {
 			return false, err
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -477,11 +477,13 @@ func (n *PagerDuty) notifyV1(
 		Details:     details,
 	}
 
-	apiURL, err := url.Parse("https://events.pagerduty.com/generic/2010-04-15/create_event.json")
-	if err != nil {
-		return false, err
+	if n.conf.URL.String() == "https://events.pagerduty.com/v2/enqueue" {
+		apiURL, err := url.Parse("https://events.pagerduty.com/generic/2010-04-15/create_event.json")
+		if err != nil {
+			return false, err
+		}
+		n.conf.URL = &config.URL{apiURL}
 	}
-	n.conf.URL = &config.URL{apiURL}
 
 	if eventType == pagerDutyEventTrigger {
 		msg.Client = tmpl(n.conf.Client)
--- a/notify/impl_test.go
+++ b/notify/impl_test.go
@@ -17,6 +17,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/http/httptest"
 	"net/url"
 	"testing"
 	"time"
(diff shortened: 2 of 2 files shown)


notify/impl.go
```diff
@@ -477,11 +477,13 @@ func (n *PagerDuty) notifyV1(
 		Details:     details,
 	}
 
-	apiURL, err := url.Parse("https://events.pagerduty.com/generic/2010-04-15/create_event.json")
-	if err != nil {
-		return false, err
+	if n.conf.URL.String() == "https://events.pagerduty.com/v2/enqueue" {
```
mxinden: > .com/**v2**/enqueue

Why are we talking to the `v2` api in the `notifyV1` function?

adamalex:
Thanks for the quick response! We have an internal incident reporting system that was built to be compatible with the pagerduty v1 api. This allows us to easily target that system with off-the-shelf tooling like alertmanager, but since it is an internal system there are internal urls we need to use for it.

notify/impl.go
adamalex: The `https://events.pagerduty.com/v2/enqueue` value is the default value in alertmanager for the v2 api. The proposed logic checks whether that is the configured url (which would mean it hasn't been overridden) before switching to the default v1 api url. This allows the url to be overridden and used for the v1 api.

stuartnelson3:
Just so I understand:

You have an internal system that expects api calls that are the same as the pagerduty v1 api, and you want to configure this URL to target that system instead of our currently hardcoded v1 api url.

adamalex:
@stuartnelson3 yes, exactly 👍

simonpasquier:
I find it brittle to compare the URL value with the default global v2 URL which itself can be overridden.

I see several alternatives:
* Add a PagerDuty v1 URL to the global configuration (default to the doc's value) and pick either the v1 or v2 global if the URL is missing when unmarshaling receiver's configuration (depending on routing/service key presence). It clutters a bit more the global configuration.
* Same solution but without exposing the default v1 URL in the configuration.
* Add a hidden field to `PagerdutyConfig` indicating whether or not PagerDuty URL was explicitly defined. The flag would be set when unmarshaling the configuration and checked by `notifyV1()`.

In all cases, it could break existing setups that had (erroneously) set the PagerDuty URL in the receiver's configuration while using the v1 API. If we go this path, it should be clearly advertised.

adamalex:
Thanks for looking @simonpasquier I totally agree with your thoughts on this. I will hold off for a bit today to collect any other preferences/ideas from others on this thread. Then I will see what can be done!

adamalex:
Thinking about this a bit more, it could be made entirely backward-compatible if the new PagerDuty v1 URL parameter you mention in your first bullet point is set up as the _only_ thing that will override the default v1 url as coded today. If the parameter, say `pagerduty_v1_url` is present it will be used only in the v1 codepath, otherwise the existing default url would continue to always be used in the v1 codepath regardless of other url configuration.

simonpasquier:
@adamalex agreed. My concern is that we add yet another parameter to the global configuration. I'm not even sure why the global v2 URL was added in the first place.

adamalex:
@simonpasquier for now I added a test and made it a bit less brittle by comparing to the global default rather than duplicating that string literal. It is still not backward-compatible for use cases where someone was configuring a url but relying on it being overridden by the current v1 logic.

When I started implementing the new parameter for `pagerduty_v1_url` it looked a bit strange to have the existing `url` parameter available (and maybe populated) but not being used at all for that code path. I didn't want to go too far without some confidence you'd be happy with it.

I thought of another backward-compatible idea: adding a bool parameter to `<pagerduty_config>` such as:

```
[ custom_v1_url: <boolean> | default = false ]
```
If unset, the current behavior would be left intact. If set to true, it would use the url set in the `url` parameter, without overriding it with the default pagerduty v1 url.

If you could let me know your preference after thinking about this I will be happy to continue getting this PR in better shape.

adamalex:
Alternately, if you would like to avoid complicating the configuration and would prefer a simpler but non-backward-compatible option I am happy to work on documentation changes for that in addition to the logic.

adamalex:
@stuartnelson3 @simonpasquier I think I'm at a stopping point for the moment. When you get a chance could you let me know your preferences on the above? I am still happy to work on whatever solution you decide is best for the project.

stuartnelson3:
Sorry for the long delay, we were all off at a conference last week and are just starting to recover from it ..

I am personally not in favor of `custom_v1_url`.

If someone sets the pagerduty url to something that isn't the global default, I say we respect and use that url. Otherwise, we switch appropriately between v1 and v2 in their respective code paths. As to the specific implementation, I would prefer that a new apiURL be created and used rather than modifying the config's value.

adamalex:
Thanks all for taking a look at this. We ended up needing to switch to a different, non-pagerduty-compatible internal endpoint so we're now using webhook notification in alertmanager and forming our own api call. That approach is working just fine for us. The use case from this PR doesn't seem like a common one, so please feel free to close this if you like.

simonpasquier:
@adamalex thanks a lot for the heads-up!

> The use case from this PR doesn't seem like a common one.

I agree. It is best if we don't extend existing notifiers when the webhook notifier is sufficient.

stuartnelson3:
Thanks for your patience throughout the discussion!

PR #1503, opened 2018-08-06
provider/mem: Add `Put` `Subscribe` starvation test

mxinden:
TestAlertsSubscribePutStarvation tests starvation of `iterator.Close` and
`alerts.Put`. Both `Subscribe` and `Put` use the Alerts.mtx lock. `Subscribe`
needs it to subscribe and more importantly unsubscribe `Alerts.listeners`.
`Put` uses the lock to add additional alerts and iterate the `Alerts.listeners`
map.  If the channel of a listener is at its limit, `alerts.Lock` is blocked,
whereby a listener can not unsubscribe as the lock is hold by `alerts.Lock`.

This test will succeed once https://github.com/prometheus/alertmanager/pull/1482 is merged.

//CC @waterdrink @s-urbaniak @simonpasquier @stuartnelson3

### DIFF
--- a/provider/mem/mem_test.go
+++ b/provider/mem/mem_test.go
@@ -16,6 +16,7 @@ package mem
 import (
 	"fmt"
 	"reflect"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


provider/mem/mem_test.go
```diff
@@ -72,6 +73,63 @@ func init() {
...
+	go func() {
+		if err := alerts.Put(alertsToInsert...); err != nil {
+			putsErr <- err
+			return
+		}
+
+		putIsDone <- struct{}{}
+	}()
+
+	// Increase probability that `iterator.Close` is called before `alerts.Put`
```
simonpasquier: super nit but missing dot. Also isn't it "`iterator.Close` is called **after** `alerts.Put`"?

mxinden:
As promised, tests are green. Any further comments?

brancz:
nit: the commit message should be "Subscribe" not "Subscripe"

mxinden:
@brancz Thanks! :+1:

PR #1507, opened 2018-08-08
pull out shared code for storing alerts

stuartnelson3:
Both the inhibitor and provider/mem packages were doing pretty much the same thing for managing holding alerts in memory, so I decided to see how it would look if this were its own package. It ends up being pretty simple, just a few methods for interacting with the interface.

Is this necessary? I don't know, it only appeared ~~twice~~ thrice. I think it separates two different parts of the code and makes it a bit easier to reason about. Opinions?


provider/mem/mem.go
```diff
@@ -145,58 +121,41 @@ func (a *Alerts) GetPending() provider.AlertIterator {
...
 // Put adds the given alert to the set.
 func (a *Alerts) Put(alerts ...*types.Alert) error {
-	a.mtx.Lock()
-	defer a.mtx.Unlock()
 
 	for _, alert := range alerts {
 		fp := alert.Fingerprint()
 
-		if old, ok := a.alerts[fp]; ok {
+		if old, err := a.alerts.Get(fp); err == nil {
```
mxinden: What happens when `Get` returns an error?

store/store.go
```diff
@@ -0,0 +1,119 @@
...
+	SetGCCallback(func(*types.Alert))
+}
+
+var (
+	ErrNotFound = errors.New("alert not found")
+)
+
+type Alerts struct {
+	sync.Mutex
+	c  map[model.Fingerprint]*types.Alert
```
mxinden: What does `c` stand for? `cache`?

store/store.go
```diff
@@ -0,0 +1,119 @@
...
+	a.Lock()
+	defer a.Unlock()
+
+	delete(a.c, fp)
+	return nil
+}
+
+// List returns a buffered channel of all current Alerts. It should be entirely
+// consumed.
+func (a *Alerts) List() <-chan *types.Alert {
```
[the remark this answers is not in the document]
stuartnelson3: I've had success trying to model data via streams (like the channel) vs. a static slice. This really stems from my initial gut reaction to make this an interface, where returning a channel would (I think) result in more flexibility to implement that interface.

dispatch/dispatch.go
```diff
@@ -404,55 +393,52 @@ func (ag *aggrGroup) stop() {
...
 		ag.next.Reset(0)
 	}
 }
 
 func (ag *aggrGroup) empty() bool {
-	ag.mtx.RLock()
-	defer ag.mtx.RUnlock()
-
-	return len(ag.alerts) == 0
+	// TODO: Does this need to be drained?
```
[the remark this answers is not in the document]
stuartnelson3: We're returning `len(ag.alerts.List())`, which is a buffered channel -- I'm not sure if we need to drain all the alerts in the channel, or if we can just leave it and the garbage collector realizes that there are no references to the channel, so it can all be cleaned up.

provider/mem/mem.go
stuartnelson3: Right now, get only returns an error if the alert isn't found. This is me using knowledge of the concrete struct implementing this store interface -- if it were some sort of remote store then it could be a network error, for example, in which case we would .. probably ? .. want to continue as it is.

store/store.go
stuartnelson3: yep! happy to rename, just seemed like the Alerts struct has a small enough responsibility that it didn't really matter.

store/store.go
```diff
@@ -0,0 +1,119 @@
...
+	"context"
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/prometheus/alertmanager/types"
+	"github.com/prometheus/common/model"
+)
+
+type Store interface {
```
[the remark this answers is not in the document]
stuartnelson3: That's a good point. I originally created it because I thought I would want to use some other struct that implemented the interface for testing, but there's really no need to mock this store. This pattern is also done in `provider` -- we only have the memory store, but it still is done as an interface.

I'm a bit conflicted. I 100% agree that since this is only being implemented by one struct that it doesn't need to be an interface, but there's something in the back of my mind that's saying designing for the interface might be a better idea. Not sure, will ruminate upon this. Designing by interface has been super helpful for me in the past, which was really the motivation here, but that doesn't mean we need to stick with having an actual `interface`.

provider/mem/mem.go
mxinden: > This is me using knowledge of the concrete struct implementing this store interface

That seems a bit implicit to me. How about just catching the `err` anyways or adding a comment?

store/store.go
mxinden: Right, cool to keep structs that small. How about a small comment for _future-me_?

stuartnelson3:
@mxinden I think this is ready for another review

dispatch/dispatch.go
```diff
@@ -404,55 +393,51 @@ func (ag *aggrGroup) stop() {
...
 	if ag.empty() {
 		return
 	}
-	ag.mtx.Lock()
 
 	var (
-		alerts      = make(map[model.Fingerprint]*types.Alert, len(ag.alerts))
-		alertsSlice = make(types.AlertSlice, 0, len(ag.alerts))
+		alerts      = ag.alerts.List()
+		alertsSlice = make(types.AlertSlice, 0, len(alerts))
```
simonpasquier: Nit, this is knowing that `ag.alerts.List()` returns a buffered channel that is sized according to the number of alerts in the store. Would be more correct to use `ag.alerts.Count()` instead?

dispatch/dispatch.go
```diff
@@ -404,55 +393,51 @@ func (ag *aggrGroup) stop() {
...
-		for fp, a := range alerts {
+		for _, a := range alertsSlice {
 			// Only delete if the fingerprint has not been inserted
 			// again since we notified about it.
-			if a.Resolved() && ag.alerts[fp] == a {
-				delete(ag.alerts, fp)
+			fp := a.Fingerprint()
+			got, err := ag.alerts.Get(fp)
+			if err != nil {
+				// TODO: Log failure
```
simonpasquier: It could happen that the GC has removed the alert in between, right?

provider/mem/mem.go
```diff
@@ -41,40 +47,21 @@ type listeningAlerts struct {
...
-
-func (a *Alerts) runGC() {
-	for {
-		select {
-		case <-a.stopGC:
-			return
-		case <-time.After(a.intervalGC):
+	a.alerts.SetGCCallback(func(alerts []*types.Alert) {
+		for _, alert := range alerts {
+			m.Delete(alert.Fingerprint())
```
simonpasquier: Did you intentionally dropped the comment from L69-71? I think it is still relevant.

```go
// As we don't persist alerts, we no longer consider them after
// they are resolved. Alerts waiting for resolved notifications are
// held in memory in aggregation groups redundantly.
```

dispatch/dispatch.go
stuartnelson3: That is my understanding -- the flush lines up with the GC and removes the alert. I'll add a log line and a comment explaining why this could happen.

stuartnelson3:
ahh, I remember reading this bit:

```
so, as a rule, we should be defining them at callsites in consuming code, rather than in the packages that provide implementations.
```

No worries, will remove the interface.

inhibit/inhibit.go
```diff
@@ -201,26 +182,16 @@ func NewInhibitRule(cr *config.InhibitRule) *InhibitRule {
 		SourceMatchers: sourcem,
 		TargetMatchers: targetm,
 		Equal:          equal,
-		scache:         map[model.Fingerprint]*types.Alert{},
+		// Get the correct ctx in here.
+		scache: store.NewAlerts(context.Background(), 15*time.Minute),
```
stuartnelson3: I really want to get the correct `ctx` in here, so that when `Stop()` is called on `inhibitor`, it propagates down to the store running in each inhibitrule, but I'm having trouble figuring out how best to do it. The separation currently between creating the inhibitor and its inhibit rules, and running the inhibitor, makes sense. The alerts store does creation and starts running its GC in the same creation step, which my gut says is right, but is at odds with the separation in the inhibitor. 

Any ideas how to fix this? I don't want to pass the `ctx` into the inhibitor creation function and then add that as a struct member, i think that's against "best practices", but barring that, I'm not sure what to do. It's an equally bad idea to defer inhibit rule creation until `Run()` is called? I'm not sure.

stuartnelson3: @mxinden @simonpasquier

mxinden: I guess `store.Alerts.run` could be delayed to be triggered by `Run()` of the inhibitor. Why not call a `Stop` function on all alert stores inside `func (ih *Inhibitor) Stop() {`?

store/store.go
```diff
@@ -0,0 +1,127 @@
...
+	c := make(chan *types.Alert, len(a.c))
+	for _, alert := range a.c {
+		c <- alert
+	}
+	close(c)
+
+	return c
+}
+
+// Count implements Store.
```
mxinden: `Store` is no more.

store/store.go
```diff
@@ -0,0 +1,127 @@
...
+// Set implements Store. It unconditionally sets the alert in memory.
+func (a *Alerts) Set(alert *types.Alert) error {
+	a.Lock()
+	defer a.Unlock()
+
+	a.c[alert.Fingerprint()] = alert
+	return nil
+}
+
+// Delete implements Store.
```
mxinden: `Store` is no more.

store/store.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/prometheus/common/model"
+)
+
+var (
+	// ErrNotFound is returned if a Store cannot find the Alert.
+	ErrNotFound = errors.New("alert not found")
+)
+
+// Alerts implements Store using an in-memory map.
+type Alerts struct {
```
mxinden: AlertStore would be more descriptive to me. Alerts sounds like a list of alerts.

store/store.go
```diff
@@ -0,0 +1,127 @@
...
+
+// NewAlerts returns a new Alerts struct.
+func NewAlerts(ctx context.Context, gcInterval time.Duration) *Alerts {
+	a := &Alerts{
+		c:  make(map[model.Fingerprint]*types.Alert),
+		cb: func(_ []*types.Alert) {},
+	}
+
+	if gcInterval == 0 {
+		gcInterval = time.Minute
```
mxinden: How about making this default more explicit by referencing a variable at the top of the file:

```golang
var defaultGCInterval = time.Minute
```

This could even be exported and then used in `main.go` for `*alertGCInterval`, right?

store/store.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+var (
+	// ErrNotFound is returned if a Store cannot find the Alert.
+	ErrNotFound = errors.New("alert not found")
+)
+
+// Alerts implements Store using an in-memory map.
+type Alerts struct {
+	sync.Mutex
```
mxinden: How about a read and write mutex instead so we can have concurrent reads? (https://golang.org/pkg/sync/#RWMutex)

provider/mem/mem.go
```diff
@@ -148,58 +137,43 @@ func (a *Alerts) GetPending() provider.AlertIterator {
...
-	a.mtx.Lock()
-	defer a.mtx.Unlock()
 
 	for _, alert := range alerts {
 		fp := alert.Fingerprint()
 
-		if old, ok := a.alerts[fp]; ok {
+		// Check that there's an alert existing within the store before
+		// trying to merge.
+		if old, err := a.alerts.Get(fp); err == nil {
```
mxinden: > This is me using knowledge of the concrete struct implementing this store interface

That seems a bit implicit to me. How about just catching the `err` anyways or adding a comment?

store/store.go
```diff
@@ -0,0 +1,127 @@
...
+	c  map[model.Fingerprint]*types.Alert
```
mxinden: Right, cool to keep structs that small. How about a small comment for _future-me_ that this is a `cache`?

store/store.go
```diff
@@ -0,0 +1,127 @@
...
+	defer a.Unlock()
+
+	alert, prs := a.c[fp]
+	if !prs {
+		return nil, ErrNotFound
+	}
+	return alert, nil
+}
+
+// Set implements Store. It unconditionally sets the alert in memory.
```
mxinden: `Store` is no more.

store/store.go
```diff
@@ -0,0 +1,127 @@
...
+	for fp, alert := range a.c {
+		if alert.Resolved() {
+			delete(a.c, fp)
+			resolved = append(resolved, alert)
+		}
+	}
+	a.cb(resolved)
+}
+
+// Get implements Store.
```
mxinden: `Store` is no more.

store/store.go
```diff
@@ -0,0 +1,127 @@
...
+	if gcInterval == 0 {
+		gcInterval = time.Minute
+	}
+
+	go a.run(ctx, gcInterval)
+
+	return a
+}
+
+// SetGCCallback implements Store.
```
mxinden: `Store` is no more.

store/store.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/prometheus/alertmanager/types"
+	"github.com/prometheus/common/model"
+)
+
+var (
+	// ErrNotFound is returned if a Store cannot find the Alert.
+	ErrNotFound = errors.New("alert not found")
+)
+
+// Alerts implements Store using an in-memory map.
```
mxinden: `Store` is no more.

store/store.go
simonpasquier: Except that it will stutter: `store.AlertStore`

inhibit/inhibit.go
stuartnelson3: I guess we could do that, I just like how the store is relatively simple and the user doesn't have to deal with a "stop" or a "start". I'll add it to move this forward, I want this to be done.

stuartnelson3:
Updated -- thanks for the reviews! hopefully this passes and we can get it merged

simonpasquier:
@stuartnelson3 you have a conflict

dispatch/dispatch.go
```diff
@@ -204,7 +205,7 @@ type aggrGroup struct {
 	timeout func(time.Duration) time.Duration
 
 	mtx        sync.RWMutex
-	alerts     map[model.Fingerprint]*types.Alert
+	alerts     *store.Alerts
```
simonpasquier: if you feel it is necessary, you could move `alerts *store.Alerts` above as it isn't guarded by `mtx` anymore.

inhibit/inhibit.go
```diff
@@ -168,10 +153,10 @@ type InhibitRule struct {
 
 	mtx sync.RWMutex
```
simonpasquier: not needed anymore AFAICT

store/store.go
```diff
@@ -0,0 +1,129 @@
...
+// Set unconditionally sets the alert in memory.
+func (a *Alerts) Set(alert *types.Alert) error {
+	a.Lock()
+	defer a.Unlock()
+
+	a.c[alert.Fingerprint()] = alert
+	return nil
+}
+
+// Delete removes the Alert with th matching fingerprint from the store.
```
simonpasquier: s/th/the/

store/store.go
stuartnelson3: Yeah, that was the reason it's just `Alerts` :man_shrugging:

dispatch/dispatch.go
stuartnelson3: that's a good point, thanks for noticing that!

stuartnelson3:
Will let the tests run and then merge, thanks for your patience on the long-running PR!

stuartnelson3:
@mxinden are you cool with "squash and merge" for merging to master? creates the single commit in master, but maintains the history in the PR

mxinden:
@stuartnelson3 yes, sounds perfect. I am sorry for being a bit annoying in this regard. Just makes preparing the CHANGELOG.md a lot easier.

stuartnelson3:
No worries, I totally agree. I've been always selecting "squash and merge", instead of "merge" or "rebase and merge". It sounds like we both prefer that, so any especially long PRs (like this one) shouldn't muddy the main history.

PR #1511, opened 2018-08-12
feat: added routing tools to amtool

FUSAKLA:
Hi @stuartnelson3, we talked for a while at the PromCon about multi-tenancy of Alertmanager and increasing complexity of routing configuration.

So I thought some testing would be useful to avoid misconfiguration.

This PR adds `routing-test` command to the `amtool` which accepts label definition of an alert and returns all receivers which the alert resolves to. It allows to load Alertmanager config from file or remotely 

example usage is 
```
$ ./amtool routing-test --config.file=doc/examples/simple.yml  service=database
team-DB-pager
```

I also added flag `--verify.receivers` which also checks if resulting receivers match those specified and if not the return code is `1`. I'm not sure if it should be implemented in the amtool but it makes testing easier.

This should allow user to define tests for the routing rules. 
Also you can run it against the new config and actual config using the remote config loading and display changes in routing introduced by the new updates.

I'd be glad for any comments or suggestions on this.

### DIFF
--- a/cli/routing.go
+++ b/cli/routing.go
@@ -1,0 +1,120 @@
+// Copyright 2018 Prometheus Team
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cli
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+
+	"github.com/xlab/treeprint"
+
+	"github.com/prometheus/alertmanager/client"
+	"github.com/prometheus/alertmanager/dispatch"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+type routingShow struct {
+	configFile        string
+	labels            []string
+	expectedReceivers string
+	tree              treeprint.Tree
+	debugTree         bool
+}
+
+const (
+	routingHelp = `Prints alert routing tree
+
+Will print whole routing tree in form of ASCII tree view.
+
+Routing is loaded from a local configuration file or a running Alertmanager configuration.
+Specifying --config.file takes precedence over --alertmanager.url.
+
+Example:
+
+./amtool config routes [show] --config.file=doc/examples/simple.yml
+
+`
+	branchSlugSeparator = "  "
+)
+
+func configureRoutingCmd(app *kingpin.CmdClause) {
+	var (
+		c              = &routingShow{}
+		routingCmd     = app.Command("routes", routingHelp)
+		routingShowCmd = routingCmd.Command("show", routingHelp).Default()
+		configFlag     = routingCmd.Flag("config.file", "Config file to be tested.")
+	)
+	configFlag.ExistingFileVar(&c.configFile)
+	routingShowCmd.Action(execWithTimeout(c.routingShowAction))
+	configureRoutingTestCmd(routingCmd, c)
+}
+
+func (c *routingShow) routingShowAction(ctx context.Context, _ *kingpin.ParseContext) error {
+	// Load configuration form file or URL.
+	cfg, err := loadAlertmanagerConfig(ctx, alertmanagerURL, c.configFile)
+	if err != nil {
+		kingpin.Fatalf("%s", err)
+		return err
+	}
+	route := dispatch.NewRoute(cfg.Route, nil)
+	tree := treeprint.New()
+	convertRouteToTree(route, tree)
+	fmt.Println("Routing tree:")
+	fmt.Println(tree.String())
+	return nil
+}
+
+func getRouteTreeSlug(route *dispatch.Route, showContinue bool, showReceiver bool) string {
+	var branchSlug bytes.Buffer
+	if route.Matchers.Len() == 0 {
+		branchSlug.WriteString("default-route")
+	} else {
+		branchSlug.WriteString(route.Matchers.String())
+	}
+	if route.Continue && showContinue {
+		branchSlug.WriteString(branchSlugSeparator)
+		branchSlug.WriteString("continue: true")
+	}
+	if showReceiver {
+		branchSlug.WriteString(branchSlugSeparator)
+		branchSlug.WriteString("receiver: ")
+		branchSlug.WriteString(route.RouteOpts.Receiver)
+	}
+	return branchSlug.String()
+}
+
+func convertRouteToTree(route *dispatch.Route, tree treeprint.Tree) {
+	branch := tree.AddBranch(getRouteTreeSlug(route, true, true))
+	for _, r := range route.Routes {
+		convertRouteToTree(r, branch)
+	}
+}
+
+func getMatchingTree(route *dispatch.Route, tree treeprint.Tree, lset client.LabelSet) {
+	final := true
+	branch := tree.AddBranch(getRouteTreeSlug(route, false, false))
+	for _, r := range route.Routes {
+		if r.Matchers.Match(convertClientToCommonLabelSet(lset)) == true {
+			getMatchingTree(r, branch, lset)
+			final = false
+			if r.Continue != true {
+				break
+			}
+		}
+	}
+	if final == true {
+		branch.SetValue(getRouteTreeSlug(route, false, true))
+	}
+}
--- a/cli/test_routing.go
+++ b/cli/test_routing.go
@@ -1,0 +1,101 @@
+// Copyright 2018 Prometheus Team
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cli
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/prometheus/alertmanager/client"
+	"github.com/prometheus/alertmanager/dispatch"
+	"github.com/xlab/treeprint"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+const routingTestHelp = `Test alert routing
+
+Will return receiver names which the alert with given labels resolves to.
+If the labelset resolves to multiple receivers, they are printed out in order as defined in the routing tree.
+
+Routing is loaded from a local configuration file or a running Alertmanager configuration.
+Specifying --config.file takes precedence over --alertmanager.url.
+
+Example:
+
+./amtool config routes test --config.file=doc/examples/simple.yml --verify.receivers=team-DB-pager service=database
+
+`
+
+func configureRoutingTestCmd(cc *kingpin.CmdClause, c *routingShow) {
+	var routingTestCmd = cc.Command("test", routingTestHelp)
+
+	routingTestCmd.Flag("verify.receivers", "Checks if specified receivers matches resolved receivers. The command fails if the labelset does not route to the specified receivers.").StringVar(&c.expectedReceivers)
+	routingTestCmd.Flag("tree", "Prints out matching routes tree.").BoolVar(&c.debugTree)
+	routingTestCmd.Arg("labels", "List of labels to be tested against the configured routes.").StringsVar(&c.labels)
+	routingTestCmd.Action(execWithTimeout(c.routingTestAction))
+}
+
+// resolveAlertReceivers returns list of receiver names which given LabelSet resolves to.
+func resolveAlertReceivers(mainRoute *dispatch.Route, labels *client.LabelSet) ([]string, error) {
+	var (
+		finalRoutes []*dispatch.Route
+		receivers   []string
+	)
+	finalRoutes = mainRoute.Match(convertClientToCommonLabelSet(*labels))
+	for _, r := range finalRoutes {
+		receivers = append(receivers, r.RouteOpts.Receiver)
+	}
+	return receivers, nil
+}
+
+func printMatchingTree(mainRoute *dispatch.Route, ls client.LabelSet) {
+	tree := treeprint.New()
+	getMatchingTree(mainRoute, tree, ls)
+	fmt.Println("Matching routes:")
+	fmt.Println(tree.String())
+	fmt.Print("\n")
+}
+
+func (c *routingShow) routingTestAction(ctx context.Context, _ *kingpin.ParseContext) error {
+	cfg, err := loadAlertmanagerConfig(ctx, alertmanagerURL, c.configFile)
+	if err != nil {
+		kingpin.Fatalf("%v\n", err)
+		return err
+	}
+
+	mainRoute := dispatch.NewRoute(cfg.Route, nil)
+
+	// Parse lables to LabelSet.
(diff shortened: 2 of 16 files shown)


roidelapluie:
if we go that path, would a --debug option help too to see the routes taken by the labels?

FUSAKLA:
Hmm that sounds reasonable, but the issue is routes does not have any names so I'm not sure how should the output look like.

FUSAKLA:
also there is possibility to use something like https://github.com/xlab/treeprint
to also draw simple ASCII tree view. But that would be duplication of https://prometheus.io/webtools/alerting/routing-tree-editor/ functionality.

stuartnelson3:
Thanks for looking at this!

I think the treeprint suggestion would be pretty cool, personally. I think there's a huge benefit to having this, and being able to do it locally is something a lot of people would like. Although the tree routing visualizer on the website does everything in the browser and doesn't store or transmit any of the entered data, people are still cautious about copying and pasting their information into a website.

The functionality so far looks cool, but it would also be helpful to add something like `--tree` so that the route could be visualized as @roidelapluie said (or `--debug`, we can think about the naming)

stuartnelson3:
I'm also thinking where this should live .. maybe it should be under `amtool config routing-tree`

FUSAKLA:
Ok, no problem. I'll be happy to add the tree visualization option.

I like the `amtool config routing-tree` naming, but I'm not sure how should the rest look like.

`amtool config routing-tree test [--tree] lavel=value` (test the alert routing, `--tree` would print just tree for specified alert)

and

`amtool config routing-tree show`  (would print the tree or without the `show`) 

? Thanks for suggestions

stuartnelson3:
for a rough UI how about:
```
amtool config routes [show] # shows tree, default action with no subcommand.
amtool config routes test [--tree] label1=value1 label2=value2 labeln=valuen # see how alert creation implements this
```

We've been calling it a `routing-tree` for a while but I've been wondering about that name. Let's try just using `routes` as the subcommand and see how it feels for now, and see what others think about it.

FUSAKLA:
Ok I have a new version according to your suggestion

I added to vendor the https://github.com/xlab/treeprint library

The tree visualization could be better but the library is simple and works quite well.
But we can change it if you want.

Output looks like this:

```bash
$ ./amtool config routes --config.file test.yml 
Routing tree:
.
└── {}  continue: false  receiver: blackhole
    ├── {service="api"}  continue: true  receiver: slack
    │   ├── {severity="critical"}  continue: true  receiver: pagerduty
    │   └── {severity="low"}  continue: true  receiver: mattermost
    ├── {service="database"}  continue: true  receiver: pagerduty
    └── {service=~"^(?:.*)$"}  continue: true  receiver: email
        └── {service=~"^(?:.*)$"}  continue: true  receiver: email
            └── {service=~"^(?:.*)$"}  continue: false  receiver: email
```

The tree output in test mode omits `continue` and receivers of parent routes
```bash
$ ./amtool config routes test --tree --config.file test.yml service=api severity=critical 
Matching tree:
.
└── {}
    ├── {service="api"}
    │   └── {severity="critical"}  receiver: pagerduty
    └── {service=~"^(?:.*)$"}
        └── {service=~"^(?:.*)$"}
            └── {service=~"^(?:.*)$"}  receiver: email


pagerduty,email
```

cli/routing.go
```diff
@@ -0,0 +1,108 @@
...
+	tree := treeprint.New()
+	convertRouteToTree(route, tree)
+	fmt.Println("Routing tree:")
+	fmt.Println(tree.String())
+	return nil
+}
+
+func checkRoutingConfigInputFlags(alertmanagerURL *url.URL, configFile string) {
+	if alertmanagerURL != nil && configFile != "" {
+		kingpin.Fatalf("You can't use both --config.file and --alertmanager.url at the same time.")
```
stuartnelson3: I have a config file stored locally which points to an alertmanager URL, which prevents me from looking at local changes via config file.

```
# ./amtool config routes --config.file=examples/ha/alertmanager.yml 
amtool: error: You can't use both --config.file and --alertmanager.url at the same time.
```

I'm not sure how to check for it, but if both flags are provided then it makes sense to either warn or bail, otherwise I think the `--config.file` flag should take precedence, since it can only be passed via command line.

cli/routing.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	if alertmanagerURL == nil && configFile == "" {
+		kingpin.Fatalf("You have to specify one of --config.file or --alertmanager.url flags.")
+	}
+}
+
+func convertRouteToTree(route *dispatch.Route, tree treeprint.Tree) {
+	routeDecription := fmt.Sprintf("%s  continue: %t  receiver: %s", route.Matchers.String(), route.Continue, route.RouteOpts.Receiver)
```
stuartnelson3: When I run this against SoundCloud's config (which has many many routes), it ends up being fairly noisy. I'm willing to bet most people don't ever use `continue`, so how about we only mention continue if its value is set to true?

Also, I'm thinking about maybe only showing the receiver when a user uses `show`. When using `test`, the matches get written out at the bottom so maybe it's not necessary?

stuartnelson3: What was your motivation behind using two spaces between the struct members?

cli/test_routing.go
```diff
@@ -0,0 +1,104 @@
...
+	for _, r := range finalRoutes {
+		receivers = append(receivers, r.RouteOpts.Receiver)
+	}
+	return receivers, nil
+}
+
+func printMatchingTree(mainRoute *dispatch.Route, ls client.LabelSet) {
+	tree := treeprint.New()
+	getMatchingTree(mainRoute, tree, ls)
+	fmt.Println("Matching tree:")
```
stuartnelson3: Let's try `Matching routes` to stick with the subcommand name

cli/routing.go
stuartnelson3: >About the test and showing a receiver in the tree. I think it doesn't hurt and it's easier to see from which branch comes which receiver (even though it matches all leafs top-down to receivers left-to-right). But if you want I can remove it.

We can leave it in for now and think about it a bit more.

>Could you possibly share how the output looks like for such complex configuration? Just out of curiosity I was lazy to write more complex routing tree :)

It's a lot to sanitize, I'll figure out a way to get it to you. I'm on vacation for the next few days so it'll be early next week.

stuartnelson3: playing around with it a bit. it's a lot of information so I understand wanting to give it some more space.

cli/routing.go
FUSAKLA: I looked at it more and I thing it's not worth the complexity to check if the flag was set from the config file. I'd than need to check if the value of `alertmanager.url` changed from the default value and it would become hard to understand I think. 

For now I'll stick to your suggestion and just print warning that `--config.file` overrides the `alertmanager-url` flag.

cli/routing.go
FUSAKLA: Forget it if you'd need to sanitize it, I was just curious :)

cli/routing.go
stuartnelson3: :+1:

cli/routing.go
```diff
@@ -0,0 +1,120 @@
...
+	tree              treeprint.Tree
+	debugTree         bool
+}
+
+const (
+	routingHelp = `Prints alert routing tree
+
+Will print whole routing tree in form of ASCII tree view.
+
+Routing is loaded from configuration file or remote Alertmanager.
```
stuartnelson3:
```
Routing is loaded from a local configuration file or a running Alertmanager configuration.
```

cli/routing.go
```diff
@@ -0,0 +1,120 @@
...
+Use --config.file or --alertmanager.url flags to specify where to look for the configuration.
+
```
stuartnelson3:
```
Specifying --config.file takes precedence over --alertmanager.url
```

cli/test_routing.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/prometheus/alertmanager/client"
+	"github.com/prometheus/alertmanager/dispatch"
+	"github.com/xlab/treeprint"
+	"gopkg.in/alecthomas/kingpin.v2"
+)
+
+const routingTestHelp = `Test alert routing
+
+Will return receiver names which the alert with given labels resolves to.
+(If resolves to multiple receivers, they are printed out joined using ',' in order as defined in the routing tree)
```
stuartnelson3: The parens `()` can be removed, and change `If resolves` to `If the labelset resolves`. I think `joined using ','` can be removed, too.

cli/test_routing.go
```diff
@@ -0,0 +1,104 @@
...
+
+Routing is loaded from configuration file or remote Alertmanager.
+Use --config.file or --alertmanager.url flags to specify where to look for the configuration.
+Than specify labels defining the alert.
```
stuartnelson3: Use the same help text for this configuration description as in `routingHelp`

cli/test_routing.go
```diff
@@ -0,0 +1,104 @@
...
+Example:
+
+./amtool config routes test --config.file=doc/examples/simple.yml --verify.receivers=team-DB-pager service=database
+
+`
+
+func configureRoutingTestCmd(cc *kingpin.CmdClause, c *routingShow) {
+	var routingTestCmd = cc.Command("test", routingTestHelp)
+
+	routingTestCmd.Flag("verify.receivers", "Checks if specified receivers matches resolved receivers. (Returns exit status 1 if not)").StringVar(&c.expectedReceivers)
```
stuartnelson3: I think the text in the parens can be changed to something like `The command fails if the labelset does not route to the specified receivers.`. I think you don't need to have the parens, also.

cli/test_routing.go
```diff
@@ -0,0 +1,104 @@
...
+	routingTestCmd.Flag("tree", "Prints out matching routes tree").BoolVar(&c.debugTree)
+	routingTestCmd.Arg("labels", "Alert labels in name=value notation").StringsVar(&c.labels)
```
stuartnelson3: `List of labels to be tested against the configured routes`

cli/test_routing.go
```diff
@@ -0,0 +1,104 @@
...
+func printMatchingTree(mainRoute *dispatch.Route, ls client.LabelSet) {
+	tree := treeprint.New()
+	getMatchingTree(mainRoute, tree, ls)
+	fmt.Println("Matching routes:")
+	fmt.Println(tree.String())
+	fmt.Print("\n")
+}
+
+func (c *routingShow) routingTestAction(ctx context.Context, _ *kingpin.ParseContext) error {
+
```
stuartnelson3: This blank line can be deleted

cli/test_routing.go
```diff
@@ -0,0 +1,104 @@
...
+	// Load configuration form file or URL.
```
stuartnelson3: I don't think this comment is necessary, I think the next line is pretty self-documenting

cli/test_routing.go
```diff
@@ -0,0 +1,104 @@
...
+	cfg, err := loadAlertmanagerConfig(ctx, alertmanagerURL, c.configFile)
+	if err != nil {
+		kingpin.Fatalf("%s", err)
```
stuartnelson3: Change this to `kingpin.Fatalf("%v\n", err)`

cli/test_routing.go
```diff
@@ -0,0 +1,104 @@
...
+		return err
+	}
+
+	mainRoute := dispatch.NewRoute(cfg.Route, nil)
+
+	// Parse lables to LabelSet.
+	ls, err := parseLabels(c.labels)
+	if err != nil {
+		kingpin.Fatalf("Failed to parse labels: %s", err)
```
stuartnelson3: Change the `%s` to `%v\n`

cli/test_routing_test.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func TestRoutingTest(t *testing.T) {
+	var tests []*routingTestDefinition
+	tests = append(tests, &routingTestDefinition{configFile: "testdata/conf.routing.yml", alert: client.LabelSet{"test": "1"}, expectedReceivers: []string{"test1"}})
+	tests = append(tests, &routingTestDefinition{configFile: "testdata/conf.routing.yml", alert: client.LabelSet{"test": "2"}, expectedReceivers: []string{"test1", "test2"}})
+	tests = append(tests, &routingTestDefinition{configFile: "testdata/conf.routing-reverted.yml", alert: client.LabelSet{"test": "2"}, expectedReceivers: []string{"test2", "test1"}})
+	tests = append(tests, &routingTestDefinition{configFile: "testdata/conf.routing.yml", alert: client.LabelSet{"test": "volovina"}, expectedReceivers: []string{"default"}})
+
+	fmt.Printf("Running routing-test testsuit:\n")
```
stuartnelson3: This can be removed

cli/test_routing_test.go
```diff
@@ -0,0 +1,66 @@
...
+	}
+	return nil
+}
+
+func TestRoutingTest(t *testing.T) {
+	var tests []*routingTestDefinition
+	tests = append(tests, &routingTestDefinition{configFile: "testdata/conf.routing.yml", alert: client.LabelSet{"test": "1"}, expectedReceivers: []string{"test1"}})
+	tests = append(tests, &routingTestDefinition{configFile: "testdata/conf.routing.yml", alert: client.LabelSet{"test": "2"}, expectedReceivers: []string{"test1", "test2"}})
+	tests = append(tests, &routingTestDefinition{configFile: "testdata/conf.routing-reverted.yml", alert: client.LabelSet{"test": "2"}, expectedReceivers: []string{"test2", "test1"}})
+	tests = append(tests, &routingTestDefinition{configFile: "testdata/conf.routing.yml", alert: client.LabelSet{"test": "volovina"}, expectedReceivers: []string{"default"}})
```
stuartnelson3: Typically these would all be defined in the slice instead of appending them, so

```go
tests := []*routingTestDefinition{
        &routingTestDefinition{configFile: "testdata/conf.routing.yml", alert: client.LabelSet{"test": "1"}, expectedReceivers: []string{"test1"}},
        &routingTestDefinition{configFile: "testdata/conf.routing.yml", alert: client.LabelSet{"test": "2"}, expectedReceivers: []string{"test1", "test2"}},
}
```

cli/test_routing_test.go
```diff
@@ -0,0 +1,66 @@
...
+			t.Fatalf("failed to load test configuration: %v", err)
+		}
+		mainRoute := dispatch.NewRoute(cfg.Route, nil)
+		err = checkResolvedReceivers(mainRoute, test.alert, test.expectedReceivers)
+		if err != nil {
+			t.Fatalf("%v", err)
+		}
+		fmt.Println("  OK")
+	}
+	fmt.Println("")
```
stuartnelson3: I think this logging can be removed as well

cli/utils.go
```diff
@@ -70,6 +73,57 @@ func parseMatchers(inputMatchers []string) ([]labels.Matcher, error) {
 	return matchers, nil
 }
 
+// GetRemoteAlertmanagerConfigStatus returns status responsecontaining configuration from remote Alertmanager
+func GetRemoteAlertmanagerConfigStatus(ctx context.Context, alertmanagerURL *url.URL) (*client.ServerStatus, error) {
```
stuartnelson3: This should be lowercase, since it's not being used outside of this package. It can be exported later if the need arises.

cli/utils.go
```diff
@@ -70,6 +73,57 @@ func parseMatchers(inputMatchers []string) ([]labels.Matcher, error) {
...
+		if err != nil {
+			return nil, err
+		}
+		return status.ConfigJSON, nil
+	}
+	return nil, errors.New("Failed to get Alertmanager configuration.")
+}
+
+// ConvertCientToCommonLabelSet converts client.LabelSet to model.Labelset
+func ConvertCientToCommonLabelSet(cls client.LabelSet) model.LabelSet {
```
stuartnelson3: `convertClientToCommonLabelSet`. There's a typo and I don't think this needs to be exported.

FUSAKLA:
Thanks once more for all the comments. 
I didn't notice the vendor, I had bad global `.gitignore` :/

FUSAKLA:
Yay I'm glad you like this feature.
thanks for accepting it :)

PR #1514, opened 2018-08-13
provider/mem: cleanup closed listener in GC

s-urbaniak:
... rather than in the Subscribe method. Currently the cleanup for a
given Alert subscription is done in a blocking goroutine, started in
the Subscribe method.

This simplifies it by moving the cleanup to the GC.

Additionally it simplifies the subscribe method by setting up the
buffered channel big enough to fill it up with all pending alerts
preventing the necessity to start a goroutine in Subscribe at all.

/cc @mxinden @brancz

### DIFF
--- a/provider/mem/mem.go
+++ b/provider/mem/mem.go
@@ -75,6 +75,16 @@ func (a *Alerts) runGC() {
 			}
 		}
 
+		for i, l := range a.listeners {
+			select {
+			case <-l.done:
+				delete(a.listeners, i)
+				close(l.alerts)
+			default:
+				// listener is not closed yet, hence proceed.
+			}
+		}
+
 		a.mtx.Unlock()
 	}
 }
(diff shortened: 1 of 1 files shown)


s-urbaniak:
@stuartnelson3 thanks a lot for the review! :-) Ahh, I am still new in the code base and wasn't aware (yet), that there is some shared functionality, so #1507 makes sense :+1: I can help out factoring that cleanup as I after quickly scanning that PR [SetGCCallback](https://github.com/prometheus/alertmanager/pull/1507/files#diff-1f8872a5d54402aa220c723deb16293cR18) seems to be the right hook?!

pinging @mxinden for the decision point if we need another unit test.

stuartnelson3:
#1507 may or may not happen :) If @mxinden is happy with this I say merge it.

mxinden:
@s-urbaniak thanks for your first contribution! :tada:

PR #1516, opened 2018-08-13
Correctly encode query strings in notifiers

### LINKED ISSUE #1515: Notifier for HipChat broken in Alertmanager 0.15.1
**What did you do?**
running AM v0.15.1 to test 

The request URL towards HipChat server (on premise) as in 0.15.0:
POST /v2/room/0000/notification?auth_token=xxxxxxxxxXXXXXXXXXXxxxxxxxxxXXXXXXXXXXXX HTTP/1.1

Request in 0.15.1:
POST /v2/room/0000/notification%3Fauth_token=xxxxxxxxxXXXXXXXXXXxxxxxxxxxXXXXXXXXXXXX HTTP/1.1
-->> which results in a HTTP 404 response code from HipChat server

it seems that this behaviour is caused by changing the url typed as string to url.URL

changing as follows in _notify/impl.go_ did fix the issue 
```
-       apiURL.Path += fmt.Sprintf("v2/room/%s/notification?auth_token=%s", roomid, n.conf.AuthToken)
+       apiURL.Path += fmt.Sprintf("v2/room/%s/notification", roomid)
+       q := apiURL.Query()
+       q.Set("auth_token", fmt.Sprintf("%s", n.conf.AuthToken))
+       apiURL.RawQuery = q.Encode()

```

stuartnelson3:
In moving from a plain string to url.URL, we
incorrectly were setting the query string via the
path. The `?` signaling the start of the query
string would then be escaped when the URL was
turned into a string.

@mxinden @simonpasquier I would classify this as extremely important, we should get a release out as soon as this is merged IMO. I'm guessing hipchat, wechat, and opsgenie are all broken?

Fixes #1515

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -805,7 +805,10 @@ func (n *Hipchat) Notify(ctx context.Context, as ...*types.Alert) (bool, error) 
 		roomid   = tmplText(n.conf.RoomID)
 		apiURL   = n.conf.APIURL.Copy()
 	)
-	apiURL.Path += fmt.Sprintf("v2/room/%s/notification?auth_token=%s", roomid, n.conf.AuthToken)
+	apiURL.Path += fmt.Sprintf("v2/room/%s/notification", roomid)
+	q := apiURL.Query()
+	q.Set("auth_token", fmt.Sprintf("%s", n.conf.AuthToken))
+	apiURL.RawQuery = q.Encode()
 
 	if n.conf.MessageFormat == "html" {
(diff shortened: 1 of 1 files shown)


stuartnelson3:
I'll create an issue, but the first thing that comes to mind is recording the request body and making sure that stays the same

stuartnelson3:
@mxinden I'm technically on vacation .. do you have time to do a release today (or soon)?

PR #1521, opened 2018-08-14
*.go: Introduce errcheck enforcing error handling

mxinden:
Errcheck [1] enforces error handling accross all go files. Functions can
be excluded via `scripts/errcheck_excludes.txt`.

This patch adds errcheck to the `test` Make target.

[1] https://github.com/kisielk/errcheck

Follow up to https://github.com/prometheus/alertmanager/pull/1501#issuecomment-410598596. @stuartnelson3 @juliusv let me know what you think. Of course doing this at all is still up for discussion.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ include Makefile.common
 
 FRONTEND_DIR            = $(BIN_DIR)/ui/app
 DOCKER_IMAGE_NAME       ?= alertmanager
+ERRCHECK_BINARY         := $(FIRST_GOPATH)/bin/errcheck
 
 ifdef DEBUG
 	bindata_flags = -debug
@@ -59,3 +60,11 @@ clean:
 	rm template/internal/deftmpl/bindata.go
 	rm ui/bindata.go
 	cd $(FRONTEND_DIR) && $(MAKE) clean
+
+.PHONY: test
+test: common-test $(ERRCHECK_BINARY)
+	@echo ">> running errcheck with exclude file scripts/errcheck_excludes.txt"
+	$(ERRCHECK_BINARY) -verbose -exclude scripts/errcheck_excludes.txt -ignoretests ./...
+
+$(ERRCHECK_BINARY):
+	@go get github.com/kisielk/errcheck
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -411,7 +411,8 @@ func main() {
 		for {
 			select {
 			case <-hup:
-				reload()
+				// ignore error, already logged in `reload()`
+				_ = reload()
 			case errc := <-webReload:
 				errc <- reload()
 			}
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -280,7 +280,11 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 			return true, err
 		}
 	}
-	defer c.Quit()
+	defer func() {
+		if err := c.Quit(); err != nil {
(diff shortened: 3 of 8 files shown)


cmd/alertmanager/main.go
```diff
@@ -411,7 +411,9 @@ func main() {
 		for {
 			select {
 			case <-hup:
-				reload()
+				if err := reload(); err != nil {
+					level.Error(logger).Log("msg", fmt.Sprintf("failed to reload config: %v", err))
```
juliusv: This would introduce double error logging because the `reload()` function already logs this itself: https://github.com/prometheus/alertmanager/blob/0e361ccd6b0ae8704a541c5909f24e585400edb5/cmd/alertmanager/main.go#L325

notify/impl.go
```diff
@@ -280,7 +280,11 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 			return true, err
 		}
 	}
-	defer c.Quit()
+	defer func() {
+		if err := c.Quit(); err != nil {
+			level.Error(n.logger).Log("msg", "failed to close SMTP connection")
```
juliusv: Include the `err` in the log message?

Makefile
```diff
@@ -59,3 +60,10 @@ clean:
 	rm template/internal/deftmpl/bindata.go
 	rm ui/bindata.go
 	cd $(FRONTEND_DIR) && $(MAKE) clean
+
+.PHONY: test
+test: common-test
+	$(ERRCHECK_BINARY) -verbose -exclude scripts/errcheck_excludes.txt -ignoretests ./...
```
juliusv: If this fails, can you output a message that mentions where to find the exceptions file?

juliusv:
Generally no hard opinion, but seems like a good idea to have this, even if it means maintaining an exceptions file.

cmd/alertmanager/main.go
```diff
@@ -411,7 +411,9 @@ func main() {
 		for {
 			select {
 			case <-hup:
-				reload()
+				if err := reload(); err != nil {
+					// logged in reload()
```
stuartnelson3: Having an empty `if` block is currently viewed as an exception by our linter

scripts/errcheck_excludes.txt
```diff
@@ -0,0 +1,7 @@
+(github.com/prometheus/alertmanager/vendor/github.com/go-kit/kit/log.Logger).Log
+// Allowed log levels are enforced via kingpin
+(*github.com/prometheus/alertmanager/vendor/github.com/prometheus/common/promlog.AllowedLevel).Set
+
+(io.Closer).Close
+(*os.File).Close
+(io.Writer).Write
```
stuartnelson3: Does this catch the response writer above that you added the error logging to?

cmd/alertmanager/main.go
mxinden: I have opened up an issue upstream (errcheck) to ignore local functions, but so far no reply: https://github.com/kisielk/errcheck/issues/151

If nothing happens any time soon, I will pull out the logging code from `reload` as you suggested :+1:.

s-urbaniak: @mxinden `_ = reload` should be your friend (don't forget to add a comment why you are ignoring the returned error.

scripts/errcheck_excludes.txt
mxinden: Removed, wasn't needed.

mxinden:
@stuartnelson3 @juliusv would you mind taking another look?

PR #1522, opened 2018-08-15
Help tab

### LINKED ISSUE #1192: Add Help to TabList in the navigation
When looking at the Prometheus UI there's a "Help" button in the top navigation bar.
I propose to add the same for the Alertmanager and link to https://prometheus.io/docs/alerting/alertmanager/. This is also referenced in the README.md.

For me personally, it's always harder to find the Alertmanager docs compared to the Prometheus docs, thus I would make use of it quite frequently.

I quickly looked at the elm source code, but I don't have the time to start adding that button right now. Thoughts?

sipian:
Closes #1192 

Since [ui/bindata.go](https://github.com/prometheus/alertmanager/blob/master/ui/bindata.go) is auto-generated while building the elm source code, I have commited this file too.
If it is not necessary, I can `git rebase` to remove it.


sipian:
The CircleCI failure is not related to this. The build is passing successfully locally.

Issue #868, opened 2017-06-16
Display "natural" grouping (and make groups linkable)


beorn7:
If I understand correctly, the new UI only allows custom grouping by the labels given in the _Group_ tab. This implies that all alerts will be grouped by the same grouping label set.
The problem with that is that the alertmanager config allows different groupings for different alerts. That's the grouping that is relevant for sending out actual notifications. But this grouping is neither available in the UI, nor can it be re-created.
For a proper link-back from a notification to the alert group in AM, we need that grouping anyway.

If I'm missing any already existing UI element/feature, please let me know.

beorn7:
After talking to @w0rm , some conclusions:

The arbitrary grouping in the UI is neat if you are in a position where you want to review the totality of alerts. However, as the person on-call for something, you really want to see the one alert group that has just sent you a page. Or you want to see all the alert groups currently sending you alerts (which could be non-paging alerts). As described in #800 , filtering by receiver helps a lot, but it is really confusing if you cannot easily group the alerts in the same way as it is relevant for alerting. In any sufficiently large organization, you'll have very different groupings present, so even if you know the groupings of your alert, you couldn't even reproduce them on the same page right now because the UI only allows the same groupings for all displayed alert.

My suggestion:

In the grouping tab, offer an option _Configured notification grouping_. If you select that, the grouping by the given label will be grayed out, and the grouping displayed will be the one resulting from the configured routing tree. (Note that this will include duplication of alerts that are in multiple notification groups.) To implement that, each alert needs to also contain a field "groupIDs" with unique IDs of the notification groups it belongs to. This will allow the UI to group.

In addition, it would be good to be able to directly query alerts by groupID. In that way, we could finally link directly to the notification group that caused the notification in question. (This is something that never worked properly in Alertmanager, but was always highly sought after. Workflow: You get paged. You find a link in your page that leads you directly to the Alertmanager UI with that notification group open, which might have accumulated more alerts in the meantime.)

beorn7:
@stuartnelson3 FYI.

beorn7:
Discussed with @w0rm and @stuartnelson3 in person, once more.

[...]

PR #1525, opened 2018-08-21
*: remove v1/alerts/groups API endpoint

### LINKED ISSUE #1508: Remove alert_groups from alertmanager
I remember it being mentioned that we don't want to use this anymore, and as far as I can tell it is exposed via api but the neither the UI or the amtool use it. I also didn't see it mentioned in the proposal of the v2 api.

Is this something we can remove?

removal would be the opposite of http://petstore.swagger.io/?url=https://raw.githubusercontent.com/mxinden/alertmanager/apiv2/api/v2/openapi.yaml

@simonpasquier @mxinden

simonpasquier:
Closes #1508

_I had this change for a while in my backlog, just rebased on master to fix some conflicts._

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -71,6 +71,15 @@ var corsHeaders = map[string]string{
 	"Cache-Control":                 "no-cache, no-store, must-revalidate",
 }
 
+// Alert is the API representation of an alert, which is a regular alert
+// annotated with silencing and inhibition info.
+type Alert struct {
+	*model.Alert
+	Status      types.AlertStatus `json:"status"`
+	Receivers   []string          `json:"receivers"`
+	Fingerprint string            `json:"fingerprint"`
+}
+
 // Enables cross-site script calls.
 func setCORS(w http.ResponseWriter) {
 	for h, v := range corsHeaders {
@@ -89,20 +98,17 @@ type API struct {
 	peer           *cluster.Peer
 	logger         log.Logger
 
-	groups         groupsFn
 	getAlertStatus getAlertStatusFn
 
 	mtx sync.RWMutex
 }
 
-type groupsFn func([]*labels.Matcher) dispatch.AlertOverview
 type getAlertStatusFn func(model.Fingerprint) types.AlertStatus
 
 // New returns a new API.
 func New(
 	alerts provider.Alerts,
 	silences *silence.Silences,
-	gf groupsFn,
 	sf getAlertStatusFn,
 	peer *cluster.Peer,
 	l log.Logger,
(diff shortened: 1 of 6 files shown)


api/api.go
```diff
@@ -71,6 +71,15 @@ var corsHeaders = map[string]string{
 	"Cache-Control":                 "no-cache, no-store, must-revalidate",
 }
 
+// Alert is the API representation of an alert, which is a regular alert
+// annotated with silencing and inhibition info.
+type Alert struct {
```
mxinden: Great! Thanks for moving APIAlert into `api/api.go`.

simonpasquier:
@mxinden done!

mxinden:
In case one of you clicks merge first, would you mind squashing? Makes preparing the CHANGELOG.md easier for the next release.

stuartnelson3:
squash4lyfe

simonpasquier:
@mxinden done

mxinden:
Thanks @simonpasquier!

mpursley:
Hi @simonpasquier @mxinden, sounds like removing the "v1/alerts/groups" endpoint from the V1 API has broken compatibility with some external clients.. e.g. 
"Don't work with the latest version api (v2) alertmanager" - https://github.com/prymitive/karma/issues/115
"Incompatibility with alertmanager 0.16.x" - https://github.com/cloudflare/unsee/issues/273

I wonder if it make sense if we can just put api/v1/alerts/groups back in? (E.g. just re-add the code that was removed from this PR?)

As:
1. This is the only endpoint from the V1 API that was removed, and
2. It sounds like this endpoint was just removed for reasons like "this is a size-able chunk of code that probably hasn't been used for over a year".  Though, there are some remote clients (like unsee and karma) that are using this endpoint, so might be worth putting it back in to support these API clients?

roidelapluie:
Yes +1 for adding it back. We use karma.

simonpasquier:
@mpursley sorry for the pain, if I would have known that that endpoint was used, it wouldn't have been removed probably. The v1 API is deprecated and the plan is to implement something similar with the v2 API to accommodate karma (see https://github.com/prometheus/alertmanager/issues/868).

mpursley:
> @mpursley sorry for the pain, if I would have known that that endpoint was used, it wouldn't have been removed probably. The v1 API is deprecated and the plan is to implement something similar with the v2 API to accommodate karma (see #868).

Yes, understood. It looks like #868 is now updated to "P2", so hopefully that update/fix gets added soon(ish).  Just keep in mind that there will be users out there that are not able to update unsee/karma or alertmanager until this endpoint gets returned or replaced. 

Thx @simonpasquier

Issue #823, opened 2017-05-24
smtp use private CA certificate 


hiscal2015:
Hi, 

I want to enable tls for smtp in my local network, but seems there's no option to config the private smtp ca certificate, how can I do?

mxinden:
@hiscal2015 I am not quite sure what you mean here. Were you able to figure it out by yourself? If you need further documentation on how to configure Alertmanager, please see: https://prometheus.io/docs/alerting/configuration/ or give us more details.

miton18:
Same issue than @hiscal2015 
We need to put our CA certificate as smtp client config, this is needed by our smtp server.

on the net/smtp Dial call, you can give a custom TLS config which can take the CA certificate

https://golang.org/pkg/crypto/tls/#Config
Is it possible to add config key [here](https://github.com/prometheus/alertmanager/blob/f4430381498c704bdfd3575219abbe31a6296044/notify/impl.go#L267) ?

mxinden:
@miton18 are you suggesting something similar to https://github.com/prometheus/alertmanager/pull/1222 (_*: configure http client from config_) for e-mail giving you the option to configure [`TLSConfig`](https://github.com/prometheus/alertmanager/blob/master/vendor/github.com/prometheus/common/config/http_config.go#L76)?

miton18:
If it allow us to set a custom TLSConfig on tls.Dial() call => Yes

Another thing, what can I do if I want to use TLS config on another port ? https://github.com/prometheus/alertmanager/blob/f4430381498c704bdfd3575219abbe31a6296044/notify/impl.go#L266

miton18:
#1528

PR #1528, opened 2018-08-24
feat(email): support custom TLS certificates

miton18:


miton18:
#823

brian-brazil:
Please use the existing TLSConfig from common rather than doing things by hand.

miton18:
@brian-brazil from common ?

brian-brazil:
Yes, that's the one.

config/notifiers.go
```diff
@@ -163,6 +163,9 @@ type EmailConfig struct {
 	HTML         string            `yaml:"html,omitempty" json:"html,omitempty"`
 	Text         string            `yaml:"text,omitempty" json:"text,omitempty"`
 	RequireTLS   *bool             `yaml:"require_tls,omitempty" json:"require_tls,omitempty"`
+	TLSCA        string            `yaml:"tls_ca,omitempty" json:"tls_ca,omitempty"`
```
mxinden: Instead of injecting these into the `EmailConfig` struct, I think @brian-brazil was referring to use [TLSConfig](https://github.com/prometheus/alertmanager/blob/master/vendor/github.com/prometheus/common/config/http_config.go#L291) from the common package.

miton18: Just understand, you want I use the TLSConfig struct instead of 3 strings in EmailConfig

but server name will be redundant

https://github.com/prometheus/alertmanager/blob/master/vendor/github.com/prometheus/common/config/http_config.go#L300

brian-brazil: I don't see why server_name would be redundant, the DNS name you're using may not match the tls cert the server itself knows about.

miton18:
Better ?

notify/impl.go
```diff
@@ -264,7 +264,16 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 	}
 
 	if port == "465" {
-		conn, err := tls.Dial("tcp", n.conf.Smarthost, &tls.Config{ServerName: host})
+
+		if n.conf.TLSConfig.ServerName == "" {
+			n.conf.TLSConfig.ServerName = host
```
simonpasquier: This modifies the original configuration object and would be reflected in `/api/v1/status` (or the Status page). Better to do it after `commoncfg.NewTLSConfig(...)`.

notify/impl.go
```diff
@@ -264,7 +264,16 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 	}
 
 	if port == "465" {
-		conn, err := tls.Dial("tcp", n.conf.Smarthost, &tls.Config{ServerName: host})
+
```
simonpasquier: (nit) empty line not needed

miton18:
Is it OK for you ?

miton18:
Thx !

PR #1532, opened 2018-08-27
Migrate from go-bindata to vfsgen

### LINKED ISSUE #1487: Replace go-bindata
See https://github.com/prometheus/prometheus/issues/2411#issuecomment-404445255 and https://github.com/prometheus/prometheus/pull/4430

TL;DR `go-bindata` isn't maintained anymore and needs to be replaced by an equivalent project in better shape.

simonpasquier:
Closes #1487

Regarding whether to commit the generated files or not, the outcome from https://github.com/prometheus/prometheus/pull/4430#issuecomment-415009642 was to commit so `go get ...` would work. But this isn't enough for AlertManager since the Javascript code (eg `/ui/app/script.js`) is also generated but not committed.


simonpasquier:
> Regarding whether to commit the generated files or not

Ok so `/ui/app/script.js` being generated but not committed, Travis CI fails because the committed `asset/assets_vfsdata.go` doesn't match with the output of the CI.

simonpasquier:
To be exhaustive, the CI fails also because just like go-bindata, vfsgen doesn't preserve timestamps (see https://github.com/shurcooL/vfsgen/issues/26). I'm testing some ideas on how we could preserve timestamps since this is hitting us for the other Prometheus projects.

mxinden:
@simonpasquier let us know once this is ready for a review.

simonpasquier:
Now that the flakes are "fixed", this is ready for review though it deserves a bit of explanation.

To circumvent the problem that the file generation wasn't idempotent, I've implemented the [`idemvfs`](https://github.com/prometheus/alertmanager/pull/1532/files#diff-9815643b0f1ee0adbda94c82a8573928) package. This package combines a "bare" `http.FileSystem` object (as generated by `vfsgen`) with metadata information about the last known modification timestamp + size + SHA1 checksum of each file in the file system. The metadata is generated by [`asset/modtime_generate.go`](https://github.com/prometheus/alertmanager/pull/1532/files#diff-82ada776f1c551467cc97defc410a815) but most of it is handled in the [`idemvfsgen`](https://github.com/prometheus/alertmanager/pull/1532/files#diff-16aa97a1999561efaf60be00801e071d) package. When `vfsgen` walks the assets, the file system will return the stored metadata if the size & checksum of the "real" file match otherwise it will return the current information. When `make assets` is executed, the metadata file is regenerated first then a second command generates the assets. This is a bit convoluted but it works. I still need to write some tests for those packages.

As we have the same issue with idempotency for the other Prometheus projects, it would make sense to put this code in the common repository but maybe it should be tried here for a bit of time?

stuartnelson3:
>But that just might be my setup (CI runs green) with Go 1.11.

I've encountered this in a couple PRs -- gobindata is formatted differently between 1.10 and 1.11 :/

simonpasquier:
@mxinden @stuartnelson3 for simplicity, I've implemented something similar to what we had with `go-bindata` so all files/directories have a modification time of `time.Unix(1, 0)`. This is achieved by having a simple wrapper around `http.FileSystem` (see https://github.com/prometheus/alertmanager/pull/1532/files#diff-b08beeb7b47ec7a1eba03e3226528c33). As discussed directly with Max, this is good enough for now and provides exactly the same functionality as before.

Issue #1529, opened 2018-08-24
amtool requires the config file(even when not necessary)


sciffer:
**What did you do?**
Used the last compiled amtool for linux on a test machine and got an error, even for the following command:
"alertmanager/amtool --help"
Based on docs it is optional, reality shows otherwise in the last version(maybe from an earlier version as well). There is no justification for such config file when the defaults are fine, especially when you don't really need them - for example when testing configuration file validity/syntax.
**What did you expect to see?**
Expected to see the help output as the config file based on the documentation is optional.
**What did you see instead? Under which circumstances?**
The following error message:
amtool: error: could not load config file: open /home/XXX/.config/amtool/config.yml: not a directory
**Environment**
Ubuntu 14.04.3 LTS
* System information:

	Linux 3.13.0-83-generic x86_64

* Amtool version:

amtool, version 0.15.2 (branch: HEAD, revision: d19fae3bae451940b8470abb680cfdd59bfa7cfa)
  build user:       root@3101e5b68a55
  build date:       20180814-10:53:39
  go version:       go1.10.3

stuartnelson3:
I'm not able to reproduce this, on `v0.15.2` or latest master.

@simonpasquier, can you reproduce this?

simonpasquier:
I can trigger the error when `~/.config/amtool` is a file instead of a directory. @sciffer is it your case too?

sciffer:
I ran into that as well - which was my mistake.
Anyhow, I got an error when I had no amtool confog or folder at all - which
is why I opened the ticket.
I downloaded the latest amd64 linux binary package, I’ll retest it once
again and update but I’m certain of what I saw(just wanted to do the config
test and failed on missing config file).
I do remember in previous versions I did not have that issue at all.

FUSAKLA:
I also managed to reproduce this only the way @simonpasquier described.

Next possibility is having miss-configured `$HOME` variable but that's basically the same
```bash
HOME="./amtool" ./amtool --help
```

simonpasquier:
We could be a bit more permissive by only failing if the path is an actual file (eg using `os.Stat()`).

https://github.com/prometheus/alertmanager/blob/d03e0ba1168f36b69eea7f2a5adbc3877fb02aa0/cli/config/config.go#L37-L43

sciffer:
[...]

PR #1538, opened 2018-09-02
fix amtool: Fixed config path check in amtool

FUSAKLA:
Fixes https://github.com/prometheus/alertmanager/issues/1529

If `~/.config/amtool` path is file amtool fails on any command. This adds `os.Stat` check before loading the file as sugested by @simonpasquier .


FUSAKLA:
Yep sorry, I keep forgetting about it.

simonpasquier:
Thanks a lot for this!

Issue #270, opened 2016-03-09
OpsGenie: Empty description causes status code 400 and thus blocks notification even for later alert with non-empty description


m-o-e:
Whenever I restart alertmanager, my log is flooded with the following messages:

`time="2016-03-09T01:14:14+01:00" level=warning msg="Notify attempt 11 failed: unexpected status code 400" source="notify.go:193"`

Alertmanager quickly backs off, but continues to spam the log with these warnings.
This is without any active alerts. I think they are leftovers from past opsgenie alerts.

Proposed fix: Alertmanager should cancel notifications when (or shortly after) the alert that triggered them is closed.

grobie:
I'm not sure canceling notifications when the alert is gone is the right fix. Scenario: Alert starts, Alert is sent to Opsgenie, connection to opsgenie is unavailable, alert closes. We still want to let OpsGenie know that the alert has been closed once we can reach OpsGenie again.

I'd rather propose to find out why OpsGenie reponds with a 400 here. The status code indicates the request sent by alertmanager is invalid and we should fix that.

m-o-e:
I think the 400s may indeed happen when the alert is closed on the OpsGenie side,
and alertmanager ends up trying to 're-close' an already closed alert.

grobie:
In that case Alertmanager's duty would be done to close an alert and it should stop doing so. I propose to:

1. Add a debug log line to actually log the body of an unexpected response
2. Validate the theory
3. Parse the body of 400 responses and handle "already closed" as success

m-o-e:
Agree +1.

grobie:
@m-o-e Can you test #271? Let me know if I should build a binary for you.

m-o-e:
Now that's what I call quick response!

@grobie Yes a linux-amd64 binary would be nice (building is probably easy, but it's late here ;) )

grobie:
@m-o-e  [alertmanager.zip](https://github.com/prometheus/alertmanager/files/164345/alertmanager.zip)

m-o-e:
@grobie Thanks. I haven't gotten around to testing it yet. Will report back when I had a chance to test (hopefully this week).

amorken:
Incidentally, I think I've stumbled on the same problem, and my proposed fix is in PR #277. @m-o-e, if you feel like testing that branch I'd appreciate it. =)

m-o-e:
Sorry for the long latency, busy times here.

[...]

PR #1540, opened 2018-09-04
Log whole request and response on notify error

BrentDouglas:
When operating alertmanager the existing logs provide no context as to
why a request is failing. Logging the whole request and response makes
debugging these issues significantly simpler.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -25,6 +25,7 @@ import (
 	"mime/multipart"
 	"net"
 	"net/http"
+	"net/http/httputil"
 	"net/mail"
 	"net/smtp"
 	"net/textproto"
@@ -181,16 +182,16 @@ func (w *Webhook) Notify(ctx context.Context, alerts ...*types.Alert) (bool, err
 	if err != nil {
 		return true, err
 	}
-	resp.Body.Close()
+	defer resp.Body.Close()
 
-	return w.retry(resp.StatusCode)
+	return w.retry(req, resp)
 }
 
-func (w *Webhook) retry(statusCode int) (bool, error) {
+func (w *Webhook) retry(req *http.Request, resp *http.Response) (bool, error) {
 	// Webhooks are assumed to respond with 2xx response codes on a successful
 	// request and 5xx response codes are assumed to be recoverable.
-	if statusCode/100 != 2 {
-		return (statusCode/100 == 5), fmt.Errorf("unexpected status code %v from %s", statusCode, w.conf.URL)
+	if resp.StatusCode/100 != 2 {
+		return (resp.StatusCode/100 == 5), formatErr(req, resp)
 	}
 
 	return false, nil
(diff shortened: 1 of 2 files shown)


brian-brazil:
Many requests will contain secrets, as may some responses, so these cannot be logged.

BrentDouglas:
@stuartnelson3 I'm not sure if this is something that you guys would be interested but I was trying to work out why a bunch of alerts were failing the other day and ended up shipping this binary to the server to work it out. I figure this should be enabled in the config/log levels somehow as this is only really something that will be wanted while debugging but I thought I would open this to see if you would be interested.
I don't really work on go stuff so I'm not going to look into how you would go about enable/disabling this kind of logging unless you indicate that its something you might want.

BrentDouglas:
@brian-brazil Hi Brian, as I mentioned above I don't want to merge this in its current state. I think it would make debugging broken alerts much easier if an option were added to enable logging this info though. It looks like other people have wanted the same debug info in the past (#270).

I agree that this is not something that I would always want in the logs but I think the option of enabling it for debugging would be super useful.

brian-brazil:
> I agree that this is not something that I would always want in the logs but I think the option of enabling it for debugging would be super useful.

Our standard way of enabling debug logs would also violate our security policies, as we promise never to expose explicitly flagged secrets in logs: https://prometheus.io/docs/operating/security/#secrets

BrentDouglas:
Ok. No worries

mxinden:
https://github.com/prometheus/alertmanager/issues/1541 relates to this.

While I think the security promise to never log secrets is very valuable, it makes debugging notifiers very hard. Would enabling more detailed logging on log level *debug* be something we could change the security policy to in the long run?

Something like:

> Fields containing secrets in configuration files (marked explicitly as such in the documentation) will not be exposed in logs **(except debug level)** or via the HTTP API.

 (//CC @stuartnelson3 @brian-brazil)

simonpasquier:
I'd rather see this responsibility delegated to the integration itself. In the case of #1541, the [PagerDuty v2 API](https://v2.developer.pagerduty.com/v2/page/events-api-reference#!/Event/post_create_event_json) states that the response has an optional `message` field defined as "`A description of the problem, or "Event processed" if successful`". I would have no problem with the PagerDuty notifier adding this field (if present) to the error message.

brian-brazil:
> Would enabling more detailed logging on log level debug be something we could change the security policy to in the long run?

Users enable debug logging for all sorts of reasons, and may not expect secrets to be revealed by it if they're e.g. debugging alerts coming into the AM. I'd rather not have a security gotcha like that.

vitroth:
@simonpasquier The problem as I see it is that frequently the response `message` is not informative enough or actually doesn't match the problem.  While thats a usability issue with PagerDuty's API, we need more information exposed to us so we can identify why PD is returning an error on the alert.

PR #1543, opened 2018-09-07
README.md: update HA section

simonpasquier:
Closes #1434

PR #1544, opened 2018-09-07
test: don't run TestResolved() in parallel

simonpasquier:
Recently Travis CI started failing tests quite often (#1532 and #1543 for instance). In particular `TestResolved()` fails because the notifications aren't received in the expected timeframe (tolerance is 150ms).

### DIFF
--- a/test/with_api_v1/acceptance/send_test.go
+++ b/test/with_api_v1/acceptance/send_test.go
@@ -15,7 +15,6 @@ package test
 
 import (
 	"fmt"
-	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


stuartnelson3:
I've re-started the travis CI run 3 or 4 times and it keeps failing :/

simonpasquier:
Yep, I've tried too. Most of the time (but not always), `TestResolved()` fails because from the 10 concurrent iterations,  a few receive the expected notifications but not in the allowed time frame (which has a tolerance of 150ms).

mxinden:
I can't find any details on why this runs ten times. Test was introduced here: https://github.com/prometheus/alertmanager/commit/cec04341f76f830f33ca91e57d179f0addac3f14.

Let's go for less than 10 :+1:

test/with_api_v1/acceptance/send_test.go
```diff
@@ -272,10 +271,7 @@ receivers:
 func TestResolved(t *testing.T) {
 	t.Parallel()
 
-	var wg sync.WaitGroup
-	wg.Add(10)
-
-	for i := 0; i < 10; i++ {
+	for i := 0; i < 2; i++ {
 		go func() {
```
mxinden: Does go testing wait until all references to `t *testing.T` are gone? Otherwise this test will finish before all go routines are done, right?

simonpasquier: Oops, I forgot to remove the `go func() {}`.

PR #1559, opened 2018-09-20
Add support for images and links in the PagerDuty notification config

seanhoughton:
This pull request adds two additional fields (images and links) to the data sent to PagerDuty. Users can either provide a static value in the config or provide template strings which allows values to come from the alert definitions and labels themselves.

Note that this only works when using the v2 API. It could be implemented in the v1 api but it has been deprecated so it's probably not worth it.

I've tested it with both static and template values using my PagerDuty instance.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -554,50 +554,14 @@ func (n *PagerDuty) notifyV2(
 		n.conf.Severity = "error"
 	}
 
-	images := make([]pagerDutyImage, len(n.conf.Images))
-	for index, item := range n.conf.Images {
-		text, err := n.tmpl.ExecuteTextString(item.Text, data)
-		if err != nil {
-			return false, fmt.Errorf("failed to template %q: %v", item.Text, err)
-		}
-		images[index].Text = text
-
-		alt, err := n.tmpl.ExecuteTextString(item.Alt, data)
-		if err != nil {
-			return false, fmt.Errorf("failed to template %q: %v", item.Alt, err)
-		}
-		images[index].Alt = alt
-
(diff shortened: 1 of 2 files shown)


notify/impl.go
```diff
@@ -606,6 +623,42 @@ func (n *PagerDuty) Notify(ctx context.Context, as ...*types.Alert) (bool, error
 		details[k] = detail
 	}
 
+	images := make([]pagerDutyImage, len(n.conf.Images))
```
simonpasquier: As `images` and `links` are only used by `notifyV2`, it would make more sense to compute them there. Otherwise LGTM.

notify/impl.go
```diff
@@ -541,12 +554,50 @@ func (n *PagerDuty) notifyV2(
 		n.conf.Severity = "error"
 	}
 
+	images := make([]pagerDutyImage, len(n.conf.Images))
```
simonpasquier: You can simplify it further by leveraging the `tmpl` helper function.

```go
msg := &pagerDutyMessage{
   ...
  Images: make([]pagerDutyImage, len(n.conf.Images)),
  Links: make([]pagerDutyLink, len(n.conf.Links)),
   ...
}

for i, img := range n.conf.Images {
    msg.Images[i] = pagerDutyImage{
        Alt: tmpl(img.Alt),
        Src: tmpl(img.Src),
        Text: tmpl(img.Text),
    }
}

// Same for n.conf.Links
```
Templating errors are caught by the `tmplErr != nil` test after.

simonpasquier:
LGTM cc @stuartnelson3 @mxinden

PR #1553, opened 2018-09-18
*: Cut v0.16.0-beta.0

mxinden:
I would like to cut a first beta of Alertmanager v0.16.0. What do you think?

PR #1560, opened 2018-09-26
ui: Move UI to API v2

mxinden:
This patch makes the Alertmanager UI use the api/v2 endpoint. In
addition it adds logic to generate the elm side data model based on the
OpenAPI specification.

//CC @stuartnelson3 as discussed via e-mail.


ui/app/elm.json
```diff
@@ -15,7 +15,8 @@
             "elm/regex": "1.0.0",
             "elm/time": "1.0.0",
             "elm/url": "1.0.0",
-            "rtfeldman/elm-iso8601-date-strings": "1.1.2"
+            "rtfeldman/elm-iso8601-date-strings": "1.1.2",
+	    "NoRedInk/elm-json-decode-pipeline":    "1.0.0"
```
stuartnelson3: the alignment on this looks funny, this was added automatically?

ui/app/src/Data/Alert.elm
```diff
@@ -0,0 +1,63 @@
...
+import Data.AlertStatus exposing (AlertStatus, alertStatusDecoder, alertStatusEncoder)
+import Data.Receiver exposing (Receiver, receiverDecoder, receiverEncoder)
+import DateTime exposing (DateTime, dateTimeDecoder, dateTimeEncoder)
+import Json.Decode as Decode exposing (Decoder)
+import Json.Decode.Pipeline exposing (optional, required)
+import Json.Encode as Encode
+import Maybe exposing (map, withDefault)
+
+
+type alias Alert =
```
stuartnelson3: It would be great to update the spec to reflect required values so we can directly interact with the values we know the API will always return

ui/app/src/Data/InlineResponse200.elm
```diff
@@ -0,0 +1,36 @@
...
+
+   OpenAPI spec version: 0.0.1
+
+   NOTE: This file is auto generated by the openapi-generator.
+   https://github.com/openapitools/openapi-generator.git
+   Do not edit this file manually.
+-}
+
+
+module Data.InlineResponse200 exposing (InlineResponse200, inlineResponse200Decoder, inlineResponse200Encoder)
```
stuartnelson3: funny,  looks like a github syntax highlighting bug?

ui/app/src/Request/Alert.elm
```diff
@@ -0,0 +1,53 @@
...
+
+import Data.Alert exposing (Alert, alertDecoder, alertEncoder)
+import Dict
+import Http
+import Json.Decode as Decode
+
+
+basePath : String
+basePath =
+    "http://localhost"
```
stuartnelson3: How does this interact with using ports (elm-reactor will want to communicate on a different port than it's being served on, for example)?

mxinden:
> Only question I have right now is about how to use the generated methods for making HTTP requests to endpoints, since it looks like localhost is hardcoded in as the baseUrl.

I am still trying to figure this out. So far I have only used the generated data models and decoders due to that. But using the generated request logic would also be nice.

mxinden:
Relates to: https://github.com/OpenAPITools/openapi-generator/issues/1140

ui/app/src/Data/Alert.elm
mxinden: :+1: Adjusted accordingly across the entire spec.

mxinden:
@stuartnelson3 @w0rm I refactored all of the UI code to use the new API v2 with the generated model logic. This is still in *work-in-progress* but I would love your feedback. Especially everything marked with `TODO`. As my elm skills are a bit rusty feedback is very welcome.

mxinden:
From my side this is the last thing missing before we can cut the first *alpha* or *beta* in https://github.com/prometheus/alertmanager/pull/1553.

ui/app/src/Views/AlertList/AlertView.elm
```diff
@@ -135,30 +164,41 @@ addLabelMsg ( key, value ) =
...
+                    a
+                        [ class "btn btn-outline-info border-0"
+                        , href (newSilenceFromAlertLabels alert.labels)
+                        ]
+                        [ i [ class "fa fa-bell-slash-o mr-2" ] []
+                        , text "Silence"
+                        ]
 
         Nothing ->
+            -- TODO: What to return here? This case should never happen.
```
w0rm: If this should never happen,  then why is status a maybe type?

ui/app/src/Views/SilenceList/SilenceView.elm
```diff
@@ -26,15 +29,20 @@ view showConfirmationDialog silence =
...
+                        Pending ->
+                            dateView "Starts" silence.startsAt
 
-                Expired ->
-                    dateView "Expired" silence.endsAt
+                        Expired ->
+                            dateView "Expired" silence.endsAt
+
+                Nothing ->
+                    text ""
```
w0rm: Can we get rid of Maybe here or flatten this possibility into the status union type so that it can be Active, Pending, Expired or Undefined?

ui/app/src/Views/SilenceList/SilenceView.elm
```diff
@@ -100,53 +108,73 @@ editButton silence =
...
-                ]
 
-        _ ->
-            let
-                editUrl =
-                    String.join "/" [ "#/silences", silence.id, "edit" ]
-            in
+        editUrl =
+            -- TODO: silence.id should always be set. Can this be done nicer?
+            String.join "/" [ "#/silences", Maybe.withDefault "" silence.id, "edit" ]
```
w0rm: If the id is undefined, we could maybe generate a link to create a new silence?

ui/app/src/Views/SilenceList/SilenceView.elm
mxinden: The definition of `Silence` is generated based on the new OpenAPI specification of the API v2. I would prefer not to have an additional internal type which converts to the generated one.

Why does the specification not enforce the `status` property: E.g. when creating a silence that status property is not yet known. Thereby the API does not enforce the property.

Is that enough context? What are your thoughts?

ui/app/src/Views/AlertList/AlertView.elm
mxinden: The definition of `Silence` is generated based on the new OpenAPI specification of the API v2. I would prefer not to have an additional internal type which converts to the generated one.

Why does the specification not enforce the `status` property: E.g. when creating a silence that status property is not yet known. Thereby the API does not enforce the property.

Is that enough context? What are your thoughts?

w0rm:
My only concern is about introducing Maybes, why do we need them if the values are guaranteed to be there?

ui/app/src/Views/AlertList/AlertView.elm
w0rm: I see, shouldn't a silence to be created and a silence that we retrieve be two different types?

ui/app/src/Views/SilenceList/SilenceView.elm
mxinden: The id would never be undefined, as Alertmanager can only return Silences with ids.

mxinden:
> My only concern is about introducing Maybes, why do we need them if the values are guaranteed to be there?

That is a valid point. I guess we have two options here:

1. Current situation: Have one type `silence` which is both used to create silences (`id` can not be required) as well as retrieve silences. This means, that retrieved silences would have a `Maybe id` property.

2. Use two different types. One type `creatableSilence` (`id` is not required) to create a silence. One type `gettableSilence` (`id` is always defined) to retrieve silences.

While is see the benefit of option two for the UI, reducing the noise with less `Maybe` operations, I would also like to point out the extra complexity for external users that now need to differentiate between the two in their client. Thereby I am voting for the first approach, having a single `silence` type but also having more noise in the code via `Maybe` handle logic.

@stuartnelson3 @w0rm What are your thoughts?

@wing328 and @trenneman from the https://github.com/OpenAPITools/openapi-generator project: Are there any best practices around this?

trenneman:
Hi @mxinden, I do not know what the community's best practice is on this (maybe ask around on [Elm Discourse](https://discourse.elm-lang.org). I often keep it as a `Maybe` as my application typically doesn't care about the id. Later when saving the resource I can decide whether I need to PUT or POST it, depending on the state of the `Maybe`. But I do not know your application and I can imagine there are use cases where it is better to split them.

ui/app/src/Data/Alert.elm
```diff
@@ -0,0 +1,63 @@
...
+    { startsAt : Maybe DateTime
+    , updatedAt : Maybe DateTime
+    , endsAt : Maybe DateTime
+    , generatorURL : Maybe String
+    , labels : Dict String String
+    , annotations : Maybe (Dict String String)
+    , receivers : Maybe (List Receiver)
```
w0rm: Why does Alert have so many Maybe fields?

w0rm:
@mxinden as far as I see it is not only about an id, but also a status of a silence. Maybe we need one type for the data, that we PUT / POST to the api, and a type of a silence that we fetch to display. Anyway, the Silence type doesn't seem like a blocker to me. 

I wonder about the Alert type. It looks confusing, because almost every field is a Maybe. Is this really what the api may return?

stuartnelson3:
I suppose I'm not sure. The idea that I really like about OpenAPI is that we can have a single definition file, and then generate many different language's clients from that.

From that perspective, having the additional OpenAPI definition for querying the API (one without all the maybe/optional values) would end up being much simpler, since each client would be simpler as a result.

I agree with @mxinden that it's probably less confusing to have the single definition, but I think if we can properly indicate that one of the definitions is for ingesting, and the other for consuming, the resulting clients will be much simpler to work with.

I'm not convinced one way or the other, but I am leaning towards the (hopefully) onetime creation of the second definition and supporting documentation which then results in a simpler client to work with.

mxinden:
> I don't think types are any helpful when any field can be a Maybe.

@w0rm I think you have a very good point.

How about I split this PR up into two so that we can move forward with parts of this patch:

1. Move status and silences to API v2 (as is in this PR)

2. Split *Alert* type into *GettableAlert* and *PostableAlert* (better name suggestions welcome)

What do you think?

w0rm:
@mxinden sounds good to me! As for naming, I prefer a simple `Alert` for the gettable alert and something different for creating an alert, maybe `AlertCreate`.

stuartnelson3:
This compromise sounds good to me :+1:

mxinden:
Closing here in favour of https://github.com/prometheus/alertmanager/pull/1613.

PR #1562, opened 2018-09-28
test: Enable testing against cluster of Alertmanagers

mxinden:
Instead of only testing single instance Alertmanagers, this patch
enables individual tests to spin up Alertmanager clusters.

In addition it adds two tests:

1. A test firing alerts against a cluster, expecting to only receive a a
notification by one of the Alertmanager instances in the cluster.

2. A test firing alerts both against a single instance as well as a
cluster, making sure the output equals.

### DIFF
--- a/test/with_api_v2/acceptance.go
+++ b/test/with_api_v2/acceptance.go
@@ -30,6 +30,7 @@ import (
 
 	apiclient "github.com/prometheus/alertmanager/test/with_api_v2/api_v2_client/client"
 	"github.com/prometheus/alertmanager/test/with_api_v2/api_v2_client/client/alert"
+	"github.com/prometheus/alertmanager/test/with_api_v2/api_v2_client/client/general"
 	"github.com/prometheus/alertmanager/test/with_api_v2/api_v2_client/models"
 
 	httptransport "github.com/go-openapi/runtime/client"
@@ -44,7 +45,7 @@ type AcceptanceTest struct {
 
 	opts *AcceptanceOpts
 
-	ams        []*Alertmanager
+	amc        *AlertmanagerCluster
 	collectors []*Collector
 
 	actions map[float64][]func()
@@ -83,6 +84,9 @@ func NewAcceptanceTest(t *testing.T, opts *AcceptanceOpts) *AcceptanceTest {
 		opts:    opts,
 		actions: map[float64][]func(){},
 	}
+	// TODO: Should this really be set during creation time? Why not do this
+	// during Run() time, maybe there is something else long happening between
+	// creation and running.
 	opts.baseTime = time.Now()
 
 	return test
@@ -110,38 +114,42 @@ func (t *AcceptanceTest) Do(at float64, f func()) {
 	t.actions[at] = append(t.actions[at], f)
 }
 
-// Alertmanager returns a new structure that allows starting an instance
-// of Alertmanager on a random port.
-func (t *AcceptanceTest) Alertmanager(conf string) *Alertmanager {
-	am := &Alertmanager{
-		t:    t,
-		opts: t.opts,
+// AlertmanagerCluster returns a new AlertmanagerCluster that allows starting a
+// cluster of Alertmanager instances on random ports.
+func (t *AcceptanceTest) AlertmanagerCluster(conf string, size int) *AlertmanagerCluster {
+	amc := AlertmanagerCluster{}
+
+	for i := 0; i < size; i++ {
+		am := &Alertmanager{
+			t:    t,
+			opts: t.opts,
+		}
+
+		dir, err := ioutil.TempDir("", "am_test")
+		if err != nil {
+			t.Fatal(err)
+		}
+		am.dir = dir
+
+		cf, err := os.Create(filepath.Join(dir, "config.yml"))
+		if err != nil {
+			t.Fatal(err)
+		}
+		am.confFile = cf
+		am.UpdateConfig(conf)
+
+		am.apiAddr = freeAddress()
+		am.clusterAddr = freeAddress()
+
+		transport := httptransport.New(am.apiAddr, "/api/v2/", nil)
+		am.clientV2 = apiclient.New(transport, strfmt.Default)
+
+		amc.ams = append(amc.ams, am)
 	}
 
-	dir, err := ioutil.TempDir("", "am_test")
-	if err != nil {
-		t.Fatal(err)
-	}
-	am.dir = dir
+	t.amc = &amc
 
-	cf, err := os.Create(filepath.Join(dir, "config.yml"))
-	if err != nil {
-		t.Fatal(err)
-	}
-	am.confFile = cf
-	am.UpdateConfig(conf)
-
-	am.apiAddr = freeAddress()
-	am.clusterAddr = freeAddress()
-
-	t.Logf("AM on %s", am.apiAddr)
-
-	transport := httptransport.New(am.apiAddr, "/api/v2/", nil)
-	am.clientV2 = apiclient.New(transport, strfmt.Default)
-
-	t.ams = append(t.ams, am)
-
-	return am
+	return &amc
 }
 
 // Collector returns a new collector bound to the test instance.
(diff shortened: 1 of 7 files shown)


test/with_api_v2/acceptance.go
```diff
@@ -110,38 +114,43 @@ func (t *AcceptanceTest) Do(at float64, f func()) {
...
+			opts: t.opts,
+		}
 
-	cf, err := os.Create(filepath.Join(dir, "config.yml"))
-	if err != nil {
-		t.Fatal(err)
-	}
-	am.confFile = cf
-	am.UpdateConfig(conf)
+		// TODO: Maybe have all Alertmanagers share the same config?
```
stuartnelson3: that's probably fine, but since this is equivalent I don't see an issue

test/with_api_v2/acceptance.go
```diff
@@ -252,16 +256,45 @@ type Alertmanager struct {
...
+	}
+
+	for _, am := range amc.ams {
+		am.Start(peerFlags)
+	}
+
+	for _, am := range amc.ams {
+		err := am.WaitForCluster(len(amc.ams))
+		if err != nil {
+			// TODO: Properly handle this error
```
stuartnelson3: is this common? we probably should fail the test if for some reason the AM can't start

test/with_api_v2/acceptance.go
```diff
@@ -252,16 +256,45 @@ type Alertmanager struct {
...
 		"--log.level", "debug",
 		"--web.listen-address", am.apiAddr,
 		"--storage.path", am.dir,
 		"--cluster.listen-address", am.clusterAddr,
 		"--cluster.settle-timeout", "0s",
-	)
+	}
+	args = append(args, additionalArg...)
+
+	cmd := exec.Command("../../../alertmanager", args...)
```
stuartnelson3: It's been mentioned before, but it would be good to reorganize the main code so it can be executed in the test, instead of relying on the environment executing this test to have a binary, and that it's been built off the current code.

stuartnelson3: (just a note for the future)

test/with_api_v2/acceptance.go
```diff
@@ -299,6 +332,42 @@ func (am *Alertmanager) Start() {
...
+// WaitForCluster waits for the Alertmanager instance to join a cluster with the
+// given size.
+func (am *Alertmanager) WaitForCluster(size int) error {
+	params := general.NewGetStatusParams()
+	params.WithContext(context.Background())
+	var status general.GetStatusOK
+
+	// Poll for 2s
+	for i := 0; i < 20; i++ {
+		status, err := am.clientV2.General.GetStatus(params)
```
stuartnelson3: 👍 
another note: any plans on moving this into a top-level `client/v2` package?

test/with_api_v2/acceptance/cluster_test.go
```diff
@@ -0,0 +1,133 @@
+// Copyright 2015 Prometheus Team
```
stuartnelson3: `/s/2015/2018/`

test/with_api_v2/acceptance/cluster_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	co.Want(a.Between(2, 3), a.Alert("alertname", "test1").Active(1))
+
+	at.Run()
+
+	t.Log(co.Check())
+}
+
+// TestClusterVSInstance compares notifications send by Alertmanager cluster to
+// notifications send by single instance
```
stuartnelson3: `/s/sent/send/`

test/with_api_v2/acceptance/cluster_test.go
```diff
@@ -0,0 +1,133 @@
...
+	wg := sync.WaitGroup{}
+
+	for i, t := range tests {
+		collectors = append(collectors, t.Collector("webhook"))
+		webhook := a.NewWebhook(collectors[i])
+
+		amClusters = append(amClusters, t.AlertmanagerCluster(fmt.Sprintf(conf, webhook.Address()), clusterSizes[i]))
+
+		// Make sure acceptance test runs long enough.
+		// TODO: Refactor hack
```
stuartnelson3: Is this a note for this PR?

test/with_api_v2/acceptance/cluster_test.go
```diff
@@ -0,0 +1,133 @@
...
+		amClusters[i].Push(a.At(1), a.Alert("alertname", "test1"))
+		collectors[i].Want(a.Between(2, 20), a.Alert("alertname", "test1").Active(1))
+
+		wg.Add(1)
+	}
+
+	for time := 0; time < 10; time++ {
+		// random bool
+		if rand.Intn(2) == 0 {
```
stuartnelson3: Why are you wanting to push alerts randomly to the clusters?

test/with_api_v2/acceptance.go
```diff
@@ -341,11 +431,19 @@ func (am *Alertmanager) Push(at float64, alerts ...*TestAlert) {
 
 		_, err := am.clientV2.Alert.PostAlerts(&params)
 		if err != nil {
+			fmt.Println("====== err")
```
simonpasquier: debugging leftover?

test/with_api_v2/acceptance.go
mxinden: Yes, that would also speed up the tests by a lot. Will look into that soonish.

test/with_api_v2/acceptance.go
mxinden: So that other projects would be able to vendor it? Great idea! I think this should take place in a follow up PR then though.

test/with_api_v2/acceptance/cluster_test.go
mxinden: You mean `/s/send/sent/`?

test/with_api_v2/acceptance/cluster_test.go
mxinden: You are right. Deterministic tests are nicer.

mxinden:
@simonpasquier The goal of the clustering feature of Alertmanager is to ensure high availability while still (in the happy path) act as if it would be a single instance. `TestClusterVSInstance` tests this behaviour by starting a single instance Alertmanager and a cluster (3) Alertmanagers, firing the same alerts against both and comparing the output.

What are your thoughts on this? In my opinion this covers a lot of logic with relatively little effort.

mxinden:
@stuartnelson3 @simonpasquier would you mind taking another look?

test/with_api_v2/acceptance/cluster_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+		amClusters = append(amClusters, t.AlertmanagerCluster(fmt.Sprintf(conf, webhook.Address()), clusterSizes[i]))
+
+		wg.Add(1)
+	}
+
+	for _, time := range []float64{0, 2, 4, 6, 8} {
+		for i, amc := range amClusters {
+			alert := a.Alert("alertname", fmt.Sprintf("test1-%v", time))
+			amc.Push(a.At(float64(time)), alert)
```
simonpasquier: `float64` conversion isn't needed.

test/with_api_v2/acceptance/cluster_test.go
```diff
@@ -0,0 +1,125 @@
...
+			collectors[i].Want(a.Between(time, time+float64(5)), alert.Active(time))
```
simonpasquier: ditto

mxinden:
> here we don't care about whether notifications are received in the expected timeframe, instead we check that both receivers receive the same notifications at roughly the same time. Is that correct?

Yes, that is correct @simonpasquier. I am open for alternative suggestions.

> Ok for me.

Is that a *looks good to me*?

@stuartnelson3 any further comments?

simonpasquier:
LGTM :smirk: 

> I am open for alternative suggestions.

Nope, I'm ok with this approach. Just that it didn't clicked immediately for me...

mxinden:
> There were a couple things on my wishlist that I mentioned, like running each alertmanager instance from a function rather than executing a local binary, but those were all for other PRs.

Yes, definitely on my TODO list.

PR #1563, opened 2018-10-01
Make a note about the firewall in HA mode

giganteous:
It took me quite some digging, I hope this helps others (e.g. #1550)

as side note: Without turning on debug logs, it doesn't really show the clustering is not working. Shall I report that as an issue?

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,8 @@ be configured to communicate with each other. This is configured using the
 The chosen port in the `cluster.listen-address` flag is the port that needs to be
 specified in the `cluster.peer` flag of the other peers.
 
+> Important: Both UDP and TCP are needed in alertmanager 0.15 and higher for the cluster to work.
+
 The `cluster.advertise-address` flag is required if the instance doesn't have
 an IP address that is part of [RFC 6980](https://tools.ietf.org/html/rfc6890)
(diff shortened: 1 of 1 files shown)


stuartnelson3:
>as side note: Without turning on debug logs, it doesn't really show the clustering is not working. Shall I report that as an issue?

Yeah, report it as an issue and we'll figure out how best to surface this. Maybe drawing attention to it in the UI or something.

mxinden:
According to the docs memberlist falls back to TCP if it can not connect via UDP. Is that not the case?

stuartnelson3:
Hm, I've never tested that behavior (I also wasn't aware of it). The standard behavior is to do small delta changes via UDP and the full-sync via TCP, so maybe we need to reword this, but I still think we should mention that both protocols are expected for normal operation.

Hashfyre:
Somehow the website still doesn't show the update in this PR:
https://prometheus.io/docs/alerting/alertmanager/#high-availability

simonpasquier:
@Hashfyre it is because this PR modified the README file of the repository but the documentation you pointed is at https://github.com/prometheus/docs/blob/master/content/docs/alerting/alertmanager.md. Do you want to open a PR there?

PR #1568, opened 2018-10-02
[FIX] Fix ineffectual assignment for default GC interval

jannickfahlbusch:
This fixes the ineffectual assignment for the default GC interval when specifiying a duration of 0

Signed-off-by: Jannick Fahlbusch <git@jf-projects.de>


mxinden:
Thanks @jannickfahlbusch for the catch and fix!

stuartnelson3:
Thanks!

PR #1575, opened 2018-10-07
[amtool] support assuming first label is alertname in alert query with matchers

stuartnelson3:
The documentation stated that the `alertname=`
part of a matcher could be dropped and it would be
assumed that the first value was the alertname.
This behavior was only honored if there was a
single matcher, but failed if there were multiple.

Any time we have one or more matchers, check to
see if the first matcher fails parsing. If so,
assume it's intended to be used as the alertname,
and append that value to the matcher.


stuartnelson3:
@bitfehler

Issue #1571, opened 2018-10-04
alertmanager_notifications_total's description isn't consistent with its definition


simonpasquier:
Source: https://github.com/prometheus/alertmanager/issues/1569#issuecomment-427014052

The description of the `alertmanager_notifications_total` metric is not aligned with what happens in the code. The description says "attempted notifications" while the counter is only increased when the notification has been successfully sent. It is definitely better to have a count of attempted notifications as currently you can't alert if all notifications fail.

Interestingly we also have `alertmanager_notification_latency_seconds_count` which is the count of attempted notifications.

https://github.com/prometheus/alertmanager/blob/30aae38c95af7267f15b5cefeddb4de40cb0f5cd/notify/notify.go#L40-L44

https://github.com/prometheus/alertmanager/blob/30aae38c95af7267f15b5cefeddb4de40cb0f5cd/notify/notify.go#L663-L679

PR #1578, opened 2018-10-08
Changes alertmanager_notifications_total description.

ChrsMark:
alertmanager_notifications_total is increased only on
successful notifications, and not on any attempted
notification as the current description points

closes https://github.com/prometheus/alertmanager/issues/1571


brian-brazil:
We should change the metric to match the description, as total+failed is easier to work with than success+failed.

PR #1586, opened 2018-10-16
run as nobody

### LINKED ISSUE #1585: Running as non-root
Does alertmanager need to run as root by default?

It looks like it was changed a while back to use the user `nobody` like [prometheus](https://github.com/prometheus/prometheus/blob/master/Dockerfile#L14) in this [pr](https://github.com/prometheus/alertmanager/pull/41) after it was raised in this [issue](https://github.com/prometheus/alertmanager/issues/39).

Looking back at the history is was removed after the Dockerfile was [deleted](https://github.com/prometheus/alertmanager/commit/84c7af965276fa31f8683b913090aa9cc085db89#diff-3254677a7917c6c01f55212f86c57fbf) and [readded](https://github.com/prometheus/alertmanager/commit/d80fd2690298710d866afe6f1cb6455ebe48b177#diff-3254677a7917c6c01f55212f86c57fbf).

InsidersByte:
closes #1585

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,10 @@ COPY amtool                       /bin/amtool
 COPY alertmanager                 /bin/alertmanager
 COPY examples/ha/alertmanager.yml /etc/alertmanager/alertmanager.yml
 
+RUN mkdir -p /alertmanager && \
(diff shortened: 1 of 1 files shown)


mxinden:
Cool, thanks @InsidersByte.

Issue #1587, opened 2018-10-17
Alert grouping in PagerDuty is not very useful


marcan:
When triggering multiple alerts that get grouped together, the result in PagerDuty winds up being like this:

![image](https://user-images.githubusercontent.com/98387/47069933-1acefe80-d22b-11e8-95db-9f8ec32179bb.png)

Only the alert summary gets updated, and subsequent messages get logged. The alert details are not updated (it still reflects the original message and summary). You have to click through the detailed log to see each individual API message sent from Alertmanager.

Ideally I'd want grouped alerts to show up individually in PagerDuty, and then be associated with the same incident, but AIUI there isn't an API for this.

The main issue here is that as further alerts are grouped together, new notifications are *not* triggered. This is dangerous, because it means that alerts can go unnoticed quite silently (until the operator manually polls Prometheus and/or Alertmanager). Basically, prometheus is "editing" an existing alert instead of creating new ones. The only workaround is to disable grouping altogether, but then this invalidates the grouping of *simultaneous* alerts, which means alert storms when several things break due to one root cause.

I'm honestly not sure what the right solution here is, but maybe something like generate new Alerts per group with a new `dedup_key` whenever additional alerts are grouped together, and try to do something smart about resolution events, like track when sent events have each of their root alerts resolved and send the resolution event then? With the current behavior, the only clean setup I can see is to just disable alert grouping altogether.

* Alertmanager version:

`alertmanager, version 0.15.2 (branch: HEAD, revision: d19fae3bae451940b8470abb680cfdd59bfa7cfa)`

marcan:
Alternatively, and this I think is useful as a global option, maybe a new grouping behavior can be introduced where alerts are *only* grouped inside the initial `group_wait` period. After notifications are sent for a group, it is effectively frozen, and new alerts matching the group key would create a new group instead. Does that make sense? I think that would work better in the context of PagerDuty, and perhaps other alert mechanisms.

marcan:
And apparently Alertmanager doesn't actually support disabling grouping, which makes this even worse. I supposed I can just throw in there a bunch of labels I don't expect to ever collide (like alertname/instance/yada), but this is kind of silly.

stuartnelson3:
>The main issue here is that as further alerts are grouped together, new notifications are not triggered.

This is the entire concept behind alertmanager's grouping. After an initial notification, comprising whatever alerts are received within `group_wait`, is sent, new notifications are only sent if:

[...]

PR #1588, opened 2018-10-17
add group_by_all support

kirillsablin:
This PR adds support for group_by_all parameter.

If it is true, all alert's labels will be used for grouping. 
If each alert has unique set of labels, this setting will effectively disable aggregation of different alerts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,10 @@ route:
   # multiple alerts coming in for cluster=A and alertname=LatencyHigh would
   # be batched into a single group.
   group_by: ['alertname', 'cluster']
-
+  
+  # if true all labels of alert will be used for grouping. If true then group_by should be empty.
+  group_by_all: false
+  
   # When a new group of alerts is created by an incoming alert, wait at
   # least 'group_wait' to send the initial notification.
   # This way ensures that you get multiple alerts for the same group that start
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ route:
   # a very low alert volume or your upstream notification system performs 
   # its own grouping. Example: group_by: [...]
   group_by: ['alertname', 'cluster']
-  
+
   # When a new group of alerts is created by an incoming alert, wait at
   # least 'group_wait' to send the initial notification.
   # This way ensures that you get multiple alerts for the same group that start
--- a/README.md
+++ b/README.md
@@ -71,10 +71,11 @@ route:
   # multiple alerts coming in for cluster=A and alertname=LatencyHigh would
   # be batched into a single group.
   #
-  # To aggregate by all labels use '...' instead of label names. 
-  # The wildcard label should be used only alone.
-  # For example: group_by: [...]
-  # WARNING! This will create a separate alert for each unique label set, so use it wisely
+  # To aggregate by all possible labels use '...' as the sole label name. 
+  # This effectively disables aggregation entirely, passing through all 
+  # alerts as-is. This is unlikely to be what you want, unless you have 
+  # a very low alert volume or your upstream notification system performs 
+  # its own grouping. Example: group_by: [...]
   group_by: ['alertname', 'cluster']
   
   # When a new group of alerts is created by an incoming alert, wait at
--- a/README.md
+++ b/README.md
@@ -70,10 +70,10 @@ route:
   # The labels by which incoming alerts are grouped together. For example,
   # multiple alerts coming in for cluster=A and alertname=LatencyHigh would
   # be batched into a single group.
+  # To aggregate by all labels use '...' instead of label names. 
+  # The wildcard label should be used only alone.
+  # For example: group_by: [...]
   group_by: ['alertname', 'cluster']
-  
-  # if true all labels of alert will be used for grouping. If true then group_by should be empty.
-  group_by_all: false
   
   # When a new group of alerts is created by an incoming alert, wait at
   # least 'group_wait' to send the initial notification.
--- a/README.md
+++ b/README.md
@@ -70,9 +70,11 @@ route:
   # The labels by which incoming alerts are grouped together. For example,
   # multiple alerts coming in for cluster=A and alertname=LatencyHigh would
   # be batched into a single group.
+  #
   # To aggregate by all labels use '...' instead of label names. 
   # The wildcard label should be used only alone.
   # For example: group_by: [...]
+  # WARNING! This will create a separate alert for each unique label set, so use it wisely
   group_by: ['alertname', 'cluster']
   
   # When a new group of alerts is created by an incoming alert, wait at
--- a/config/config.go
+++ b/config/config.go
@@ -494,7 +494,7 @@ func (c *GlobalConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 // A Route is a node that contains definitions of how to handle alerts.
 type Route struct {
-	Receiver   string `yaml:"receiver,omitempty" json:"receiver,omitempty"`
+	Receiver string `yaml:"receiver,omitempty" json:"receiver,omitempty"`
 
 	GroupByStr []string `yaml:"group_by,omitempty" json:"group_by,omitempty"`
 	GroupBy    []model.LabelName
--- a/config/config.go
+++ b/config/config.go
@@ -541,7 +541,7 @@ func (r *Route) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	}
 
 	if len(r.GroupBy) > 0 && r.GroupByAll {
-		return fmt.Errorf("cannot have wildcard group_by and other other labels at same time")
+		return fmt.Errorf("cannot have wildcard group_by and other other labels at the same time")
 	}
 
 	groupBy := map[model.LabelName]struct{}{}
--- a/config/config.go
+++ b/config/config.go
@@ -494,9 +494,10 @@ func (c *GlobalConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 // A Route is a node that contains definitions of how to handle alerts.
 type Route struct {
-	Receiver   string            `yaml:"receiver,omitempty" json:"receiver,omitempty"`
-	GroupBy    []model.LabelName `yaml:"group_by,omitempty" json:"group_by,omitempty"`
-	GroupByAll bool              `yaml:"group_by_all,omitempty" json:"group_by_all,omitempty"`
+	Receiver   string `yaml:"receiver,omitempty" json:"receiver,omitempty"`
+	GroupBy    []model.LabelName
+	GroupByStr []string `yaml:"group_by,omitempty" json:"group_by,omitempty"`
+	GroupByAll bool
 
 	Match    map[string]string `yaml:"match,omitempty" json:"match,omitempty"`
 	MatchRE  map[string]Regexp `yaml:"match_re,omitempty" json:"match_re,omitempty"`
@@ -526,9 +527,20 @@ func (r *Route) UnmarshalYAML(unmarshal func(interface{}) error) error {
 			return fmt.Errorf("invalid label name %q", k)
 		}
 	}
+	for _, l := range r.GroupByStr {
+		if l == "..." {
+			r.GroupByAll = true
+		} else {
+			labelName := model.LabelName(l)
+			if !labelName.IsValid() {
+				return fmt.Errorf("invalid label name %q in group_by list", l)
+			}
+			r.GroupBy = append(r.GroupBy, labelName)
+		}
+	}
 
 	if len(r.GroupBy) > 0 && r.GroupByAll {
-		return fmt.Errorf("cannot have group_by_all is true and non-empty group_by")
+		return fmt.Errorf("cannot have wildcard group_by and other other labels at same time")
 	}
 
 	groupBy := map[model.LabelName]struct{}{}
--- a/config/config.go
+++ b/config/config.go
@@ -541,7 +541,7 @@ func (r *Route) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	}
 
 	if len(r.GroupBy) > 0 && r.GroupByAll {
-		return fmt.Errorf("cannot have wildcard group_by and other other labels at the same time")
+		return fmt.Errorf("cannot have wildcard group_by (`...`) and other other labels at the same time")
 	}
 
 	groupBy := map[model.LabelName]struct{}{}
--- a/config/config.go
+++ b/config/config.go
@@ -496,6 +496,7 @@ func (c *GlobalConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 type Route struct {
 	Receiver string            `yaml:"receiver,omitempty" json:"receiver,omitempty"`
 	GroupBy  []model.LabelName `yaml:"group_by,omitempty" json:"group_by,omitempty"`
+	GroupByAll bool `yaml:"group_by_all,omitempty" json:"group_by_all,omitempty"`
 
 	Match    map[string]string `yaml:"match,omitempty" json:"match,omitempty"`
 	MatchRE  map[string]Regexp `yaml:"match_re,omitempty" json:"match_re,omitempty"`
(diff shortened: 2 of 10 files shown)


kirillsablin:
@stuartnelson3

kirillsablin:
Well, use cases are different, and sometimes such functionality is needed

kirillsablin:
also it will help in situations like #1587

roidelapluie:
Actually @brian-brazil this is a useful feature.

At the beginning I was also thinking that group_by: [] would do this. But it does the opposite.

marcan:
Yeah, Alertmanager is a useful bridge between Prometheus and a bunch of other things. "Just talk to Prometheus directly" is throwing away the usefulness of the glue that is Alertmanager. Sure, this may not fit certain ideas of best practices, but not all environments are the same and not everyone has the same requirements with alert grouping. This seems like a simple enough feature to add, why not do so if it solves some peoples' problems?

simonpasquier:
From my POV, one big reason for not adding it is that it makes the AlertManager configuration even more difficult to get right and understand. Also I'm not convinced that it helps for #1587.

As for not grouping some alerts, I would rather add a special label to my alerting rules that would be based of all the labels necessary to uniquely identify an alert instance.

```yaml
  - alert: MyAlert
    expr: my_metric == 0
    for: 1m
    labels:
      unique_key: "{{ $labels.foo }}/${{ $labels.bar }}/and/many/more"
```

And then use `group_by [alertname,job,instance,unique_key]` would be enough to have all my alerts ungrouped without having AlertManager to know about all possible label names.

marcan:
It helps for #1587 because it guarantees that all alerts page individually, by never grouping anything. This is a plausible config for smaller shops, where even if everything is down the page storm is manageable.

Honestly, I don't see how mucking up all alert rules with an extra custom label is better than this one-shot config... at that point you might as well add all possible unique labels to group_by, it's less work than messing with each alert rule individually.

VR6Pete:
This is absolutely required where you want to feed through individual alerts to PD / OpsGenie etc... - its sort of useless at the moment when grouped together from a reporting perspective and actually knowing what to respond too. 

But I still would want to retain grouping for informal slack notifications etc....

simonpasquier:
As @beorn7 already [commented](https://github.com/prometheus/alertmanager/issues/1587#issuecomment-430964745), features shouldn't be added to alleviate the lack of PD or OpsGenie. I can find at least one past [example](https://github.com/prometheus/alertmanager/pull/444#issuecomment-424685490) where a feature had been asked for similar reasons and eventually OpsGenie fixed the limitation on their end. You should try to get those people involved here.

VR6Pete:
That example is a different situation and its not clear what the lack of support actually from an opsgenie / pager duty perspective when its alert manager firing the alerts?

roidelapluie:
Okay, this is not a limitation of anything, it is more a limitation of HA in alertmanager.

Here is my usecase:

We have a webhook receiver that can only manage one alert at the time. That's it.


We tried to implement this in the webhook receiver (threating each alert individually)... but then.. it is too slow and the second alertmanager sends the alerts too!! Also, when alert 5 fails, we need to redo all the 4 first alerts, we can not fail atomically.

This can be avoided by this pull request. And it is more a way to deal with alertmanager HA.

roidelapluie:
We use queues for alerts too but in another use case which is out of scope here.

The idea here is:

I expect group_by: [] to +not group alerts+. Instead, it creates a massive group.

We have a receiver that, for exemple, posts to twitter. We are limited in lenght so we must send the alerts 1 by 1.

If we do that in the receiver, it is possible that out of 10 alerts, 1 fails, but we need to reply 500 to the alertmanager, so that the complete group would be re-sent. Or it will take a lot of time and the 2nd alert manager will also fire the alerts.

For those two reasons this feature makes sense.

brian-brazil:
The Alertmanager is fundamentally an at-least-once delivery system. I don't think we should be adding (contentious) features to support webhooks that aren't prepared to deal with that.

simonpasquier:
> I expect group_by: [] to +not group alerts+. Instead, it creates a massive group.

From my POV, it does the right thing: grouping on an empty set of label names means that you want to put all alerts in the same bucket.

If you really want it, you could generate a label that identifies every alert uniquely:

```yaml
  - alert: SomeAlert
    expr: my_metric == 0
    for: 1m
    labels:
      key: "{{ range $k, $v := $labels }}{{ $k }}={{ $v }}{{end}}"
```

marcan:
I still fail to see how a simple toggle to *disable* existing behavior is so controversial, and why everyone keeps proposing horrible workarounds that replace a single line in the Alertmanager config with changes to every single alerting rule in Prometheus.

I get how this may not be something fitting certain ideas of how things should work or how external services should behave, but we've covered several use cases already where this simple feature would be useful to people. Clearly there is value here. Why the aversion to 1:1 alert routing? Alertmanager isn't just a grouper, it does more things, like silencing and converting alerting protocols. Why tie those things to mandatory grouping? This isn't some horrible subversion of what Alertmanager does or how it should behave in a stack. It's basically just trivial syntactic sugar for a behavior that people can already achieve in an ugly, non-ergonomic way.

brian-brazil:
> that replace a single line in the Alertmanager config with changes to every single alerting rule in Prometheus.

We generally do not add features that can already be handled via configuration management.

> Alertmanager isn't just a grouper, it does more things, like silencing and converting alerting protocols.  Why tie those things to mandatory grouping? This isn't some horrible subversion of what Alertmanager does or how it should behave in a stack. 

The purpose of the Alertmanager is to increase the signal to noise ratio of notifications, grouping and throttling are innate parts of that. Allowing this would be a subversion of the goals of the alertmanager, as it is one of several features that users request for the purposes of spamming themselves due to being used to less powerful systems (without the notion of labels), rather than following https://docs.google.com/document/d/199PqyG3UsyXlwieHaqbGiWVa8eMWi8zzAn0YfcApr8Q/preview.

supermch:
I don't see how asking all of our customers to add a certain label to every single alert config which looks horrible or alternatively having to pre-process their configs is a better idea than a simple configuration.

kirillsablin:
>The purpose of the Alertmanager is to increase the signal to noise ratio of notifications, grouping and throttling are innate parts of that.
But not only these. There are also http_config, pagerduty_config, webhooks_config which means that they are also responsibilities of Alertmanager.

marcan:
> The purpose of the Alertmanager is to increase the signal to noise ratio of notifications, grouping and throttling are innate parts of that. Allowing this would be a subversion of the goals of the alertmanager, as it is one of several features that users request for the purposes of spamming themselves due to being used to less powerful systems (without the notion of labels), rather than following

The goal of software should not be to enforce a particular philosophy on its users. The goal of software should be to empower users to choose the configuration that suits them best. It's fine to nudge users towards not hurting themselves with documentation, but it's not productive to actively alienate those users with legitimate use cases for such features. While there is obviously no requirement that software implement every possible feature that every possible user might use, refusing to implement a simple "turn off this behavior I don't need or want" toggle on ideological grounds is, in my opinion, just poor form. You're basically saying "I know better than everyone else, so you can either do things my way or write your own code".

You know full well I was also an SRE at Google, and I agree with most of Rob's essay. But not everyone is Google. I've seen SRE teams at Google in monitoring hell with 10 pages per day. I've seen teams at Google that never got paged. There are many different scenarios that are very different from Google scale. In the extreme, small single-homed systems are often assumed to be able to stay up for years on end without paging. There is no point in optimizing alerting/monitoring rules for cases like that, because you don't even have the data to know what you need to monitor! E.g. symptom-based alerting is fine and dandy when you have a large distributed system with many frontends where you can afford to lose some with zero impact, but for a little single-homed service, damn right I care about finding out that it's running out of RAM *before* it actually goes down. There is no one-size-fits-all monitoring philosophy.

roidelapluie:
The pushgateway is an official exporter. This is a recognition that Prometheus someone's needs to do exceptions in its principles. In would see this as an exception needed for some well defined use cases.

discordianfish:
FWIW, fully agree with @marcan here.

beorn7:
Has anybody ever run into a situation where the (current) `group_by: []` (with an empty list, i.e. put every alert into one group) would have been useful? In my own AM experience, I have never used an empty `group_by`, and I had to think what it would do when I saw it discussed here.

Even if there is controversy if "no grouping" is useful, perhaps we can all agree that "no grouping" is still saner than "all alerts in one group"? In that case, we could change the behavior of `group_by: []`. The followers of pure Promethean philosophy would not notice as they would never do that. Those that want to switch off grouping could do it now easily (i.e. without creating an artificial label with a unique label), but at least there is not an explicit option for it so we wouldn't create the impression it would be a recommended practice. And those who argue doing no grouping is shooting into your own foot could be satisfied by the fact that grouping all alerts into one group is even worse foot-shooting.

Finally, in case somebody _really_ wants to put all alerts into the same group, you could still say `group_by: [some_random_label_name_that_will_never_be_used]` to accomplish that.

beorn7:
> Yes, a daily email with all currently firing alerts.

Fair point.

There goes my idea to reconcile all demands…

stuartnelson3:
My initial reaction after reading some of the feedback is against including this. For a small alerting setup/individual route where it is reasonable to send a notification for each alert, it doesn't seem like it would be too hard to come up with a unique group_by labelset, and for a large setup, this feature would end up hurting users with spam.

However, it seems that this is a feature that some users very much want. If @mxinden is also fine with it, I think we can review and trial the feature.

marcan:
@stuartnelson3 The problem with coming up with a unique group_by labelset is that it's fragile. It't what I'm doing now, sure, but what if tomorrow I add some alerts that make that labelset no longer unique? It's a ticking time bomb; having the Alertmanager config contain a labelset that has to stay in sync with the potential behavior of all Prometheus alerts (which ultimately depends on what targets you're scraping and what metrics they expose, which means it could even change behind your back with a software update) is definitely not what I'd consider hygienic or good engineering.

Or: just because it's at a scale small enough that I want to get notified for every alert doesn't mean I want my config to be fragile and to have to keep track of all potential labelsets.

Also, notifying for every alert is entirely reasonable for large systems if alerts are being grouped by a downstream receiver, as some support features like this themselves (and thus overlap with Alertmanager in functionality, so users should be able to choose whichever side they prefer to do the grouping).

mxinden:
@stuartnelson3 [comment](https://github.com/prometheus/alertmanager/pull/1588#issuecomment-434293283) is very much inline with my own thoughts on this.

Given the high community involvement (which is very much appreciated) I am also voting for the "review and trial" of this feature. This is not an approval of the particular implementation of this pull request, but rather of the idea of a one-notification-per-alert feature.

beorn7:
Implementation-wise, I'd prefer to not have a separate config option for this. Perhaps a magic label name (that is an illegal label name otherwise) like `*` or `...` could be used to mean "group by all".

stuartnelson3:
>Implementation-wise, I'd prefer to not have a separate config option for this. Perhaps a magic label name (that is an illegal label name otherwise) like * or ... could be used to mean "group by all".

This sounds good to me. I lean towards `group_by: [*]`, I read that as an expansion to everything.

@beorn7 is this how you imagined it? @mxinden thoughts?

beorn7:
@stuartnelson3 Yes, that's the best I could come up with so far. (Is `*` something you can leave unquoted in Yaml?) I don't like "magic" strings (we already have too many of them in the Prometheus universe), but this one looks intuitive: `group_by: []` vs `group_by [*]` vs `group_by [alertname, zone]`.

simonpasquier:
> I lean towards group_by: [*]

+1 if we need to pick something

> Is * something you can leave unquoted in Yaml?

Yes

kirillsablin:
I'll reimplement with '*'

kirillsablin:
Unfortunately, 
according to 
http://yaml.org/spec/current.html#id2535812 
and 
http://yaml.org/spec/current.html#alias/syntax, 
alias is starting with *, so it have to be quoted. 

Do you agree with 
`group_by: ['*']` ?

beorn7:
Too bad…

However, `group_by: [...]` would be valid YAML. Should we go with the latter?

beorn7:
`group_by: [...]` would also avoid any wrong assumption that we support globbing. (Seeing `group_by: ['*']`, people might try `group_by: ['some_prefix_*']` to group by all labels starting with `some_prefix_`.)

stuartnelson3:
@beorn7 raises a good point. `[...]` sounds good, too.

discordianfish:
Good point, +1 on `[...]`

README.md
```diff
@@ -70,8 +70,11 @@ route:
   # The labels by which incoming alerts are grouped together. For example,
   # multiple alerts coming in for cluster=A and alertname=LatencyHigh would
   # be batched into a single group.
+  # To aggregate by all labels use '...' instead of label names.
```
beorn7: There should be an explicit warning that this will create a separate alert for each unique label set, which is only useful in very specific scenarios.

marcan:
Suggestion for the text:

> To aggregate by all possible labels use '...' as the sole label name. This effectively disables aggregation entirely, passing through all alerts as-is. This is unlikely to be what you want, unless you have a very low alert volume or your upstream notification system performs its own grouping.

README.md
```diff
@@ -70,8 +70,14 @@ route:
...
   # be batched into a single group.
+  #
+  # To aggregate by all possible labels use '...' as the sole label name. 
+  # This effectively disables aggregation entirely, passing through all 
+  # alerts as-is. This is unlikely to be what you want, unless you have 
+  # a very low alert volume or your upstream notification system performs 
+  # its own grouping. Example: group_by: [...]
   group_by: ['alertname', 'cluster']
-
+
```
stuartnelson3: nit, can you remove this whitespace?

config/config.go
```diff
@@ -525,6 +528,21 @@ func (r *Route) UnmarshalYAML(unmarshal func(interface{}) error) error {
 			return fmt.Errorf("invalid label name %q", k)
 		}
 	}
+	for _, l := range r.GroupByStr {
+		if l == "..." {
+			r.GroupByAll = true
+		} else {
+			labelName := model.LabelName(l)
+			if !labelName.IsValid() {
```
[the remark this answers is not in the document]
kirillsablin: It is here to preserve original behaviour.
 Since '...' is not valid label, I had to remove requirement of validation from it, but we still need validation for all other cases.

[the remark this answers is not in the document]
simonpasquier: ack!

config/config.go
```diff
@@ -525,6 +528,21 @@ func (r *Route) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+				return fmt.Errorf("invalid label name %q in group_by list", l)
+			}
+			r.GroupBy = append(r.GroupBy, labelName)
+		}
+	}
+
+	if len(r.GroupBy) > 0 && r.GroupByAll {
+		return fmt.Errorf("cannot have wildcard group_by and other other labels at same time")
```
simonpasquier: (nit) `at the same time`

config/config.go
```diff
@@ -525,6 +528,21 @@ func (r *Route) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+			labelName := model.LabelName(l)
+			if !labelName.IsValid() {
+				return fmt.Errorf("invalid label name %q in group_by list", l)
+			}
+			r.GroupBy = append(r.GroupBy, labelName)
+		}
+	}
+
+	if len(r.GroupBy) > 0 && r.GroupByAll {
+		return fmt.Errorf("cannot have wildcard group_by and other other labels at the same time")
```
mxinden:
```suggestion
		return fmt.Errorf("cannot have wildcard group_by (`...`) and other labels at the same time")
```

config/config_test.go
```diff
@@ -144,6 +144,47 @@ receivers:
...
+	in := `
+route:
+  group_by: ['alertname', 'cluster', '...']
+  receiver: team-X-mails
+receivers:
+- name: 'team-X-mails'
+`
+	_, err := Load(in)
+
+	expected := "cannot have wildcard group_by and other other labels at the same time"
```
mxinden: See comment above.

dispatch/route.go
```diff
@@ -158,6 +162,9 @@ type RouteOpts struct {
 	// What labels to group alerts by for notifications.
 	GroupBy map[model.LabelName]struct{}
 
+	// Use all alert's labels to group
```
mxinden:
```suggestion
	// Use all alert labels to group.
```

Not sure we need the `'s` here.

Issue #980, opened 2017-09-08
[Bug] smtp.icoremail.net:465 fail to send mail alert due to 'does not advertise the STARTTLS extension'


regardfs:
Here is my config:

```
global:
  resolve_timeout: 15s

route:
  receiver: 'slack-notifications'
  group_wait: 10s
  group_interval: 1m
  repeat_interval: 4h
  group_by: ["alertname", "instance"]
  routes:
    - receiver: 'email'
      group_by: ['alertname', 'cluster', 'instance']

receivers:
  - name: slack-notifications
    slack_configs:
    - send_resolved: true
      api_url: https://hooks.slack.com/services/adadadad123123xxxxxxxx
      channel: '#alert'
      text: '{{ template "slack.text" . }}'
  - name: 'email'
    email_configs:
    - to: '123@it.com'
      send_resolved: true
      require_tls: true
      from: 123@it.com
      smarthost: smtp.icoremail.net:465
      auth_username: 123@it.com
      auth_password: xxxx
      html: '{{ template "email.test.html" . }}'
      headers: { Subject: "[Alert] Prometheus Alert Email" }

templates:
  - '/etc/alertmanager/templates/alertText.tmpl'
  - '/etc/alertmanager/templates/alertEmail.tmpl'
```
Then start service It could not send alert mail with following error:

> time="2017-09-08T08:46:47Z" level=error msg="Error on notify: require_tls: true (default), but "smtp.icoremail.net:465" does not advertise the STARTTLS extension" source="notify.go:283" 
time="2017-09-08T08:46:47Z" level=error msg="Notify for 1 alerts failed: require_tls: true (default), but "smtp.icoremail.net:465" does not advertise the STARTTLS extension" source="dispatch.go:262" 

So, i modify **require_tls** to **false**, so I receive message error like

> time="2017-09-08T08:29:52Z" level=error msg="Error on notify: *smtp.plainAuth failed: wrong host name" source="notify.go:283" 
time="2017-09-08T08:29:52Z" level=error msg="Notify for 1 alerts failed: *smtp.plainAuth failed: wrong host name" source="dispatch.go:262" 

When changed to gmail , no error , no exception happened. so , I wander does it not supporting icoremail ?

stuartnelson3:
Using port 465 enters a hardcoded path that attempts to use TLS:

https://github.com/prometheus/alertmanager/blob/master/notify/impl.go#L275-L285

[...]

PR #1591, opened 2018-10-21
FIX "PLAIN" auth during notification via smtp-over-tls on port 465

ckressibucher:
When creating a `NewClient`, pass only the hostname
as value for the `host` parameter, instead of `n.conf.Smarthost`,
which is hostname port.

This is the same as `smtp.Dial` does, which is called in
the else-branch.

The value passed as `host` argument to `NewClient` is
later compared against the `plainAuth.host` value:
`net/smtp.(*Client).Auth` and `net/smtp.(*plainAuth).Start`
(plainAuth is created created here: <https://github.com/prometheus/alertmanager/blob/master/notify/impl.go#L245>)

In my tests, this fixed the problems described in #980 

Note: after this change, the strings which are compared in
https://golang.org/src/net/smtp/auth.go at (`if server.Name != a.host`)
are derived from the same config value. I think that is correct, but as it
might affect security, it should be reviewed carefully.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -276,7 +276,7 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 		if err != nil {
 			return true, err
 		}
-		c, err = smtp.NewClient(conn, n.conf.Smarthost)
+		c, err = smtp.NewClient(conn, host)
 		if err != nil {
 			return true, err
 		}


simonpasquier:
Thanks for the PR! A quick look makes me believe that it is correct. Especially when looking at `smtp.Dial()` code and seeing the same pattern. Also the fact that the second parameter of `smtp.NewClient` is named `host` and not `addr`. That being said and as you noted, it would require careful reviewing and testing. FWIW this part of the code hasn't changed since #704 that introduced the support for port 465.

```go
func Dial(addr string) (*Client, error) {
	conn, err := net.Dial("tcp", addr)
	if err != nil {
		return nil, err
	}
	host, _, _ := net.SplitHostPort(addr)
	return NewClient(conn, host)
}
```

stuartnelson3:
I don't feel well versed enough in email to understand the implications of this code change. If anyone feels confident about their knowledge and ability to test this, by all means raise a hand.

dirkmb:
:+1: for this PR.
Just debugged the 'wrong host error' myself and created the same fix... 
should have checked the open PRs first :(

dirkmb:
If there would be no ServerName set, the DialWithDialer function in crypto/tls/tls.go:132 sets ServerName to the hostname (which does not include the port number)
Also the Dial method in net/smtp/smtp.go:51 does parse the hostname to NewClient.

So what's the reason this PR is not merged yet? ;)

chanjarster:
@simonpasquier Thanks, that works. I have tested gmail(port 587) and another mail provider (port 465 with ssl), everything goes fine.

simonpasquier:
@mxinden @stuartnelson3 this PR looks reasonable to me and seems to fix the original issue. Any problem from your side merging it?

PR #1594, opened 2018-10-22
Make OpsGenie API Key a templated string

twz123:
This is analogous of what is done for PagerDuty secrets.

Probably needs a doc update as well.

/cc @stuartnelson3

### DIFF
--- a/notify/impl_test.go
+++ b/notify/impl_test.go
@@ -235,7 +235,7 @@ func TestOpsGenie(t *testing.T) {
 		Tags:        `{{ .CommonLabels.Tags }}`,
 		Note:        `{{ .CommonLabels.Note }}`,
 		Priority:    `{{ .CommonLabels.Priority }}`,
-		APIKey:      `s3cr3t`,
+		APIKey:      `{{ .ExternalURL }}`,
 		APIURL:      &config.URL{u},
 	}
 	notifier := NewOpsGenie(conf, tmpl, logger)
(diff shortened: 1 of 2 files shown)


brian-brazil:
My understanding is that this is not used for routing, has this changed?

twz123:
OpsGenie has so-called Team Integrations, which are essentially team-specific versions of global (account-wide) Integrations. They allow teams to configure their incoming alert rules separately from each other. Such Team Integrations get their own API keys. So this is somewhat a special version of an Alert->Team routing.

notify/impl_test.go
```diff
@@ -218,12 +220,28 @@ func readBody(t *testing.T, r *http.Request) string {
 }
 
 func TestOpsGenie(t *testing.T) {
+	templateDir, err := ioutil.TempDir("", "opsgenie-")
```
brian-brazil: Why are you doing all this setup? You can test this like the others.

twz123: Ah, you mean to create a file in testdata? Didn't occur to me :see_no_evil: Will do!

brian-brazil: You should't need to create any files. The use case here is that the apikey is coming from an alert label, and any random template expression is sufficient to verify that templating is being applied.

brian-brazil:
Thanks! I'd suggest a read of https://prometheus.io/docs/operating/security/#alertmanager

PR #1608, opened 2018-11-05
Throw error if no auth mechanism is match

### LINKED ISSUE #1555: Throw error when no auth mechanism is matched
https://github.com/prometheus/alertmanager/blob/d5b8e1028695866d52abd36fcf1973b1fdc9c807/notify/impl.go#L254

In case we are asked to auth and no auth mechanism matched and returned from function, we should return error saying something like: "No mechanism matched server offer: 250-AUTH NTLM"

glefloch:
This pull request aims to close #1555 by returning an error if no auth mechanism is found.

Signed-off-by: glefloch <glfloch@gmail.com>

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -251,7 +251,7 @@ func (n *Email) auth(mechs string) (smtp.Auth, error) {
 			return LoginAuth(username, password), nil
 		}
 	}
-	return nil, nil
+	return nil, errors.New("Unknown auth mechanism: " + mechs)
 }
 
 // Notify implements the Notifier interface.
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -226,7 +226,7 @@ func (n *Email) auth(mechs string) (smtp.Auth, error) {
 		case "CRAM-MD5":
 			secret := string(n.conf.AuthSecret)
 			if secret == "" {
-				err.Add(errors.New("Missing secret for CRAM-MD5 auth mechanism"))
+				err.Add(errors.New("missing secret for CRAM-MD5 auth mechanism"))
 				continue
 			}
 			return smtp.CRAMMD5Auth(username, secret), nil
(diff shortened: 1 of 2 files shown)


notify/impl.go
```diff
@@ -251,7 +251,7 @@ func (n *Email) auth(mechs string) (smtp.Auth, error) {
 			return LoginAuth(username, password), nil
 		}
 	}
-	return nil, nil
+	return nil, errors.New("Unknown auth mechanism: " + mechs)
```
simonpasquier: There are several error paths:

* the remote server offer none of CRAM-MD5, AUTH or LOGIN.
* the remote server offer supported mechanisms but the AlertManager configuration doesn't match (for instance, missing auth password with PLAIN).

I think the error message should be clear on the reason. Maybe [`type.MultiError`](https://github.com/prometheus/alertmanager/blob/b63b5600741b07795e50cedf8818a853bacc5f76/types/types.go#L211) can be useful here.

glefloch: Well, as both errors cannot append at the same time, can we consider to return another error in that case?

simonpasquier: Sure. Though you can still have PLAIN and CRAM-MD5 available but neither AuthSecret nor AuthPassword set.

glefloch:
@simonpasquier I have used the `types.MultiError` object. I am going to have a look to failed tests.

simonpasquier:
I've restarted the Travis tests as it is probably a flake. In any case, it might be good to add a few unit tests.

glefloch:
@simonpasquier, I have added unit tests. Could you tell me if this is what you expected?

simonpasquier:
Thanks for this! :+1: for me.

notify/impl.go
```diff
@@ -220,19 +220,21 @@ func NewEmail(c *config.EmailConfig, t *template.Template, l log.Logger) *Email
...
 func (n *Email) auth(mechs string) (smtp.Auth, error) {
 	username := n.conf.AuthUsername
-
+	err := &types.MultiError{}
 	for _, mech := range strings.Split(mechs, " ") {
 		switch mech {
 		case "CRAM-MD5":
 			secret := string(n.conf.AuthSecret)
 			if secret == "" {
+				err.Add(errors.New("Missing secret for CRAM-MD5 auth mechanism"))
```
mxinden: I think errors should start with a lower case letter: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

glefloch: You are right, I updated errors and unit tests.

PR #1609, opened 2018-11-07
Use quoted-printable transfer encoding for the email notifier

free:
Use [`quoted-printable`](https://en.wikipedia.org/wiki/Quoted-printable) transfer encoding for the text sent out by the email notifier.

Turns out MIME has a maximum line length of 998 bytes ([RFC 2822](https://www.ietf.org/rfc/rfc2822.txt), section 2.1.1) and some SMTP servers will enforce that by inserting a `<CR><LF>` after 998 characters, whether that's in the middle of a URL or HTML tag, potentially breaking the email's contents.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -23,6 +23,7 @@ import (
 	"io/ioutil"
 	"mime"
 	"mime/multipart"
+	"mime/quotedprintable"
 	"net"
 	"net/http"
 	"net/mail"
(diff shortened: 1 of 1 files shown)


simonpasquier:
LGTM. I suppose that you've tested it with a real SMTP server?

free:
> I suppose that you've tested it with a real SMTP server?

Yes I did, sorry for failing to mention that. I've been running an alertmanager with this change for a couple of days now. And I have seen a few alerts with lines longer than 998 characters displayed without corruption, so it works as expected.

simonpasquier:
Thanks a lot!

Issue #1581, opened 2018-10-10
when failure recovery,i can not receive message


eason0420:
the config file:

```yaml
receivers:
  - name: 'yunwei'
    wechat_configs:
    - send_resolved: true
      to_party: '@all'
      to_user: "@all"
      agent_id: '1000002'
      corp_id: 'wxxxxxxxxxb'
      api_secret: '9xxxxxxxxxxxxxxxxx0'
      message: "{{ .CommonLabels.alertname }} \n{{ .CommonLabels.instance }} \n{{ .CommonLabels.severity }} \n{{ .CommonAnnotations.description }}\n{{ .CommonAnnotations.value }}"
```

i can receive the alert, but when failure recovery,i can not receive message

simonpasquier:
Can you run with `--log.level=debug` and share the logs?

eason0420:
@simonpasquier  the logs

level=debug ts=2018-10-10T12:24:40.988643786Z caller=dispatch.go:201 component=dispatcher msg="Received alert" alert="OS[Disk Used][8296a78][active]"
level=debug ts=2018-10-10T12:24:50.988995266Z caller=dispatch.go:445 component=dispatcher aggrGroup="{}/{severity=\"warning\"}:{instance=\"center_db_slave\"}" msg=Flushing alerts="[OS[Disk Used][8296a78][active]]"
level=debug ts=2018-10-10T12:24:50.989310427Z caller=impl.go:907 msg="Notifying Wechat" incident="{}/{severity=\"warning\"}:{instance=\"center_db_slave\"}"
level=debug ts=2018-10-10T12:24:51.168254816Z caller=impl.go:1002 msg="response: {\"errcode\":0,\"errmsg\":\"ok\",\"invaliduser\":\"\"}" incident="{}/{severity=\"warning\"}:{instance=\"center_db_slave\"}"
level=debug ts=2018-10-10T12:25:00.989203273Z caller=dispatch.go:445 component=dispatcher aggrGroup="{}/{severity=\"warning\"}:{instance=\"center_db_slave\"}" msg=Flushing alerts="[OS[Disk Used][8296a78][active]]"
level=debug ts=2018-10-10T12:25:10.98936351Z caller=dispatch.go:445 component=dispatcher aggrGroup="{}/{severity=\"warning\"}:{instance=\"center_db_slave\"}" msg=Flushing alerts="[OS[Disk Used][8296a78][active]]"
[...]

PR #1611, opened 2018-11-07
types: fix alert merging

simonpasquier:
Alert merging assumed that EndsAt would always be empty for firing alerts. This is no longer true starting with Prometheus v2.4.0: EndsAt is set to a multiple of the evaluation interval or resend interval (whichever is the largest). See https://github.com/prometheus/prometheus/pull/4550

The issue might be in https://github.com/prometheus/alertmanager/issues/1581 and it has been raised on the [prometheus users mailing list](https://groups.google.com/d/msg/prometheus-users/87uL3bno_8c/jthEiSRaBAAJ) too.

I've added more tests because even though the code change is small, it is quite tricky to get it right (and I'm not sure I did!).

cc @brian-brazil @stuartnelson3 @mxinden

### DIFF
--- a/types/types_test.go
+++ b/types/types_test.go
@@ -16,6 +16,7 @@ package types
 import (
 	"reflect"
 	"sort"
+	"strconv"
 	"testing"
 	"time"
 
@@ -26,13 +27,17 @@ import (
 func TestAlertMerge(t *testing.T) {
 	now := time.Now()
 
+	// By convention, alert A is always older than alert B.
 	pairs := []struct {
 		A, B, Res *Alert
 	}{
 		{
+			// Both alerts have the Timeout flag set.
+			// StartsAt is defined by Alert A.
+			// EndsAt is defined by Alert B.
 			A: &Alert{
 				Alert: model.Alert{
-					StartsAt: now.Add(-time.Minute),
+					StartsAt: now.Add(-2 * time.Minute),
 					EndsAt:   now.Add(2 * time.Minute),
 				},
 				UpdatedAt: now,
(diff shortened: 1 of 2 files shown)


mxinden:
Just a high level question. I need to give this further thoughts.

> Alert merging assumed that EndsAt would always be empty for firing alerts.

How can the EndsAt ever be empty with:
https://github.com/prometheus/alertmanager/blob/625604df90b0f2e080f7d32fea4aa891675276d6/api/v1/api.go#L448-L451

Wouldn't it rather be:

> Alert merging assumed that Alert.Timeout would always be true for firing alerts.

mxinden:
I don't see https://github.com/prometheus/alertmanager/pull/1553 (v0.16.0) happening in the short term. Do you think we should cut a patch release for the previous v0.15.x series?

types/types_test.go
```diff
@@ -55,12 +115,148 @@ func TestAlertMerge(t *testing.T) {
...
+				Alert: model.Alert{
+					StartsAt: now.Add(-2 * time.Minute),
+					EndsAt:   now,
+				},
+				UpdatedAt: now.Add(time.Minute),
+			},
+			Res: &Alert{
+				Alert: model.Alert{
+					StartsAt: now.Add(-2 * time.Minute),
+					EndsAt:   now,
```
mxinden: I am confused by this. If I have two alerts, where one is firing and one is not, I expect the merged alert to still be firing, right? Especially as none of them have the Timeout flag set.

stuartnelson3: Because B is received after A, it's EndsAt value "wins" because A isn't resolved (is what I'm reading from the code and comments).

simonpasquier: > Because B is received after A

Exactly, an alert that is effectively resolved wins over an older alert that was firing previously.

mxinden: Right got it. Thanks for clarifying @stuartnelson3 & @simonpasquier!

mxinden:
Should we merge this first into `release-0.15` branch and then merge `release-0.15` into `master`. That way we don't rewrite history via a cherry-pick. What do you think @stuartnelson3 @simonpasquier.

stuartnelson3:
>Should we merge this first into release-0.15 branch and then merge release-0.15 into master. That way we don't rewrite history via a cherry-pick. What do you think @stuartnelson3 @simonpasquier.

Sounds fine to me

mxinden:
@simonpasquier would you mind rebasing your one commit on top of release-0.15 branch?

simonpasquier:
PR rebased on top of `release-0.15`. I had an additional acceptance test in `test/with_api_v2/acceptance/send_test.go` that wasn't relevant anymore on the release branch since the V2 API doesn't exist. I'll add it back after this PR is merged back into `master`.

mxinden:
Will follow up with a CHANGELOG pull request in a bit.

mxinden:
>  I had an additional acceptance test in test/with_api_v2/acceptance/send_test.go that wasn't relevant anymore on the release branch since the V2 API doesn't exist. I'll add it back after this PR is merged back into master.

@simonpasquier would you mind following up with a PR for this test?

PR #1613, opened 2018-11-09
ui: Move /status & /silences to API v2

mxinden:
This patch makes the Alertmanager UI (/status & /silences) use the
api/v2 endpoint. In addition it adds logic to generate the elm side data
model based on the OpenAPI specification.

In addition requesting alerts from api v1 is done in a separate commit to be
able to revert it once alerts also come from api v2.


ui/app/src/Views/SilenceList/Views.elm
```diff
@@ -97,6 +100,23 @@ states =
...
+
+
+filterSilenceByState : State -> Silence -> Bool
+filterSilenceByState state silence =
+    -- TODO: Can this be done cleaner?
+    Maybe.withDefault False <| Maybe.map (.state >> (==) state) <| .status silence
+
+
+
+{-
```
stuartnelson3: This can be removed?

stuartnelson3:
Thanks for doing all this. We haven't touched the elm code in a while, but it's really nice coming back to this language and feeling confident in the changes we make thanks to the type system :)

ui/app/src/Data/Alert.elm
```diff
@@ -0,0 +1,63 @@
...
+import Data.Receiver as Receiver exposing (Receiver)
+import DateTime exposing (DateTime)
+import Dict exposing (Dict)
+import Json.Decode as Decode exposing (Decoder)
+import Json.Decode.Pipeline exposing (optional, required)
+import Json.Encode as Encode
+
+
+type alias Alert =
+    { startsAt : Maybe DateTime
```
mxinden: These will change once I move `/alerts` to the new API as well.

stuartnelson3:
Looks good to me!

mxinden:
Thanks @w0rm and @stuartnelson3. I will follow up with moving the `/alerts` page as well to API v2.

PR #1622, opened 2018-11-14
cmd/alertmanager: use buffered channel for signal

simonpasquier:
Similar to https://github.com/prometheus/prometheus/pull/4865


brian-brazil:
:+1:

PR #1623, opened 2018-11-15
test: add acceptance test for firing alerts with EndsAt

simonpasquier:
Follow up of https://github.com/prometheus/alertmanager/pull/1611#issuecomment-438953319

### DIFF
--- a/test/with_api_v2/acceptance/send_test.go
+++ b/test/with_api_v2/acceptance/send_test.go
@@ -26,8 +26,15 @@ import (
 // notification is eventually sent at least once and ideally exactly
 // once.
 
-func TestMergeAlerts(t *testing.T) {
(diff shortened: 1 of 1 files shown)


test/with_api_v2/acceptance/send_test.go
```diff
@@ -104,6 +104,72 @@ receivers:
 	t.Log(co.Check())
 }
 
+// This test is similar to TestMergeAlerts except that the firing alerts have
```
mxinden: As far as I can tell this only differs in 3 lines (`am.Push`). How about deduplicating this logic and using sub-tests (via `r.Run`)?

simonpasquier: Sure, I've refactored the code to reduce duplication.

PR #1631, opened 2018-11-20
[gossip] Don't merge expired gossip messages

### LINKED ISSUE #1624: Alertmanager not GC'ing silences
**What did you do?**

Checked the "expired" tab on the silences page, and saw silences going back 2 months in time. The alertmanager cluster is using the default retention time, `120h`.

Expired silences for the last `120h`.

Expired silences for the last 2 months.

**Environment**

* System information:

	`Linux 4.4.10+soundcloud x86_64`

* Alertmanager version:

	`0.15.3`

* Logs:
Unfortunately, we don't have much logging or metrics around this.

stuartnelson3:
If they're expired, they should be cleaned up on
the next GC cycle, but merging them in means that
they'll probably be gossip'd continually between
the cluster members.

fixes #1624 

The issue was in the `merge` function: https://github.com/prometheus/alertmanager/blob/master/silence/silence.go#L752-L756

Expired silences would get deleted every 15 minutes (the hard-coded maintenance interval), but on each full-state sync an alertmanager would get a bunch of expired silences. Since they didn't exist in its state cache, they would be merged back in. Since the merge happens every minute, they were seemingly guaranteed to stick around. If all the alertmanagers were started at the same time they might GC all the silences before a full state sync, but if they're offset in their maintenance timing, they would keep sharing the same expired silences with each other. This explains why I saw the silences "GC'd", because I had looked within this magical window of "maintenance has run, but a full state sync hasn't yet", but then they came back.

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -403,7 +403,7 @@ func (s *Silences) setSilence(sil *pb.Silence) error {
 		return err
 	}
 
-	s.st.merge(msil)
+	s.st.merge(msil, s.now())
 	s.broadcast(b)
 
 	return nil
@@ -717,8 +717,10 @@ func (s *Silences) Merge(b []byte) error {
 	s.mtx.Lock()
 	defer s.mtx.Unlock()
 
+	now := s.now()
+
 	for _, e := range st {
-		if merged := s.st.merge(e); merged && !cluster.OversizedMessage(b) {
+		if merged := s.st.merge(e, now); merged && !cluster.OversizedMessage(b) {
 			// If this is the first we've seen the message and it's
 			// not oversized, gossip it to other nodes. We don't
 			// propagate oversized messages because they're sent to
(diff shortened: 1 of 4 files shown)


silence/silence.go
```diff
@@ -717,7 +717,12 @@ func (s *Silences) Merge(b []byte) error {
 	s.mtx.Lock()
 	defer s.mtx.Unlock()
 
+	now := s.now()
+
 	for _, e := range st {
+		if e.ExpiresAt.After(now) {
```
mxinden: I am confused, so if the silence has not yet expired, continue? Shouldn't this be:

```golang
 if e.ExpiresAt.Before(now) { 
```

simonpasquier: I had the same remark :-)

simonpasquier: Also are notification logs affected too?

stuartnelson3: I went for using `time.Before()`. It makes it easier to read in my mind.

stuartnelson3:
Oh dear, my force-pushings are being reported!

stuartnelson3:
>Also are notification logs affected too?

I expect so, we've just never noticed because they don't appear in the UI.

stuartnelson3:
The code between these two is so similar, the difference is just a generic away ..

PR #1633, opened 2018-11-21
Fix route prefix for the API v2

### LINKED ISSUE #1627: API v2 doesn't honor the web.route-prefix argument
**What did you do?**

Start AlertManager with `--web.route-prefix=/foo`

`curl localhost:9093/foo/api/v2/silences` returns 200 OK.

404 not found.

`curl localhost:9093/foo/api/`**v1**`/silences` returns 200 OK.

simonpasquier:
Closes #1627 

While I was looking at the code, it occurred to me that we use [`os.Exit`](https://godoc.org/os#Exit) improperly as it doesn't run deferred functions. I refactored that part in this PR but it can go in another PR if someone prefers.

Finally, there's still a problem with the UI: the Status tab doesn't render properly but I'm not familiar enough with the front-end code to know how to fix it. @stuartnelson3 and @mxinden, you probably know better.

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -129,6 +129,10 @@ func newMarkerMetrics(marker types.Marker) {
 const defaultClusterAddr = "0.0.0.0:9094"
 
 func main() {
+	os.Exit(run())
+}
+
+func run() int {
 	if os.Getenv("DEBUG") != "" {
 		runtime.SetBlockProfileRate(20)
 		runtime.SetMutexProfileFraction(20)
@@ -177,7 +181,7 @@ func main() {
 	err := os.MkdirAll(*dataDir, 0777)
 	if err != nil {
 		level.Error(logger).Log("msg", "Unable to create data directory", "err", err)
-		os.Exit(1)
+		return 1
 	}
 
 	var peer *cluster.Peer
@@ -197,7 +201,7 @@ func main() {
 		)
 		if err != nil {
 			level.Error(logger).Log("msg", "unable to initialize gossip mesh", "err", err)
-			os.Exit(1)
+			return 1
 		}
 	}
 
@@ -216,7 +220,7 @@ func main() {
 	notificationLog, err := nflog.New(notificationLogOpts...)
 	if err != nil {
 		level.Error(logger).Log("err", err)
-		os.Exit(1)
+		return 1
 	}
 	if peer != nil {
 		c := peer.AddState("nfl", notificationLog, prometheus.DefaultRegisterer)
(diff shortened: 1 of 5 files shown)


cmd/alertmanager/main.go
```diff
@@ -478,17 +509,6 @@ func extURL(listen, external string) (*url.URL, error) {
 	return u, nil
 }
 
-func listen(listen string, apiV1Handler *route.Router, apiV2Handler http.Handler, logger log.Logger) {
```
mxinden: :+1: Thanks for cleaning this up!

test/with_api_v2/acceptance.go
```diff
@@ -290,6 +291,7 @@ func (amc *AlertmanagerCluster) Start() error {
 
 // Start the alertmanager and wait until it is ready to receive.
 func (am *Alertmanager) Start(additionalArg []string) error {
+	am.t.Helper()
```
mxinden: Fancy. I wasn't aware of this.

test/with_api_v2/acceptance/web_test.go
```diff
@@ -0,0 +1,40 @@
...
+  group_by: []
+  group_wait:      1s
+  group_interval:  1s
+  repeat_interval: 1h
+
+receivers:
+- name: "default"
+`
+
+	at := a.NewAcceptanceTest(t, &a.AcceptanceOpts{RoutePrefix: "/foo"})
```
mxinden: Would you mind adding a comment, why we don't need to test whether the API is working with the prefix. Something like:

`The test framework polls the API with the given prefix during Alertmanager startup and thereby ensures proper configuration.

simonpasquier: Done. BTW I had forgotten to add `test/with_api_v1/acceptance/web_test.go`, this is fixed with the latest version.

stuartnelson3:
>Finally, there's still a problem with the UI: the Status tab doesn't render properly but I'm not familiar enough with the front-end code to know how to fix it. @stuartnelson3 and @mxinden, you probably know better.

@mxinden did you have a chance to look at the status issue?

cmd/alertmanager/main.go
```diff
@@ -129,6 +129,10 @@ func newMarkerMetrics(marker types.Marker) {
 const defaultClusterAddr = "0.0.0.0:9094"
 
 func main() {
+	os.Exit(run())
+}
+
+func run() int {
```
stuartnelson3: this is so close to fixing another issue, the integration tests relying on a locally built binary.. if you have time, could you expand this to take all the command line flags as arguments (or options/config struct), and then use this function in the integration tests? a separate PR is cool

mxinden:
> > Finally, there's still a problem with the UI: the Status tab doesn't render properly but I'm not familiar enough with the front-end code to know how to fix it. @stuartnelson3 and @mxinden, you probably know better.
> 
> @mxinden did you have a chance to look at the status issue?

Not yet. Given that this bug is not introduced through this pull request I would like to delay the fix to a follow up pull request.

cmd/alertmanager/main.go
simonpasquier: Yep

PR #1637, opened 2018-11-26
ui: Pass parameterized api url down to status logic

mxinden:
Previously the api url was hardcoded in the elm logic fetching the
status of an Alertmanager. Instead it is now being passed down like we
already did for /alerts, /silences and /receivers.

Addresses https://github.com/prometheus/alertmanager/pull/1633#issuecomment-441578324. This was introduced by me in https://github.com/mxinden/alertmanager/commit/e4e053b18eacd777c5aab9d5bf489f9f99b328a1.

@stuartnelson3 @w0rm @simonpasquier would you mind giving this a review?


mxinden:
Thanks @w0rm and @stuartnelson3!

Issue #1638, opened 2018-11-26
api: Deduplicate alerts definition via `allOf` parameter


mxinden:
Now that the fix for https://github.com/OpenAPITools/openapi-generator/issues/1465 was merged, we can deduplicate the `alert` definition in the `openapi.yaml` spec like we did for silences.

More details:  https://github.com/prometheus/alertmanager/blob/091a8a83b16e2f04f25d832237cf42846ef939b4/api/v2/openapi.yaml#L355-L359

PR #1640, opened 2018-11-28
api/v2: Extract shared properties of gettable and postable alert

mxinden:
With issue 1465 on openapi-generator [1] being fixed, we can not extract
shared properties of the gettable and postable alert definition into a
shared object (`alert`) like we do for silence, gettable silence and
postable silence.

In addition this patch does the following changes to the UI:

- Use `List GettableAlert` instead of plural type definition like
`GettableAlerts` because the plural definitions are not generated
anymore.

- Fix openapi-generator-cli docker image to specific hash.

Fixes https://github.com/prometheus/alertmanager/issues/1638 and addresses current master branch failures.

[1] https://github.com/OpenAPITools/openapi-generator/issues/1465


w0rm:
Looks better, the plural type aliases seemed a bit strange to me!

simonpasquier:
LGTM

PR #1651, opened 2018-12-05
WIP: Extend functions avaible in go template

sylr:
By adding sprig's FuncMap


stuartnelson3:
Is there a specific function you want?

roidelapluie:
Actually it would make sense to also include that in prometheus

roidelapluie:
date functions, map function, list functions would all be super neat.

brian-brazil:
> Actually it would make sense to also include that in prometheus

Many of those functions are already there, and many more don't make sense - plus we'd have to document them all.

roidelapluie:
Well I wanted to manipulate the a label map in console templates and that was not possible. But maybe I am abusing console_templates :)

sylr:
Many of those  ? The DefaultFuncs contains 7 functions, sprig add dozens.

FYI [Helm use this library](https://github.com/helm/helm/blob/master/docs/chart_template_guide/functions_and_pipelines.md#template-functions-and-pipelines)

stuartnelson3:
This looks a lot like giving users a gun to shoot themselves with. Are there specific things you want to do right now in notification templates that aren't supported by the current templating?

sylr:
I want to split a string into a slice to be able to remove unwanted labels from `.CommonLabels`

sylr:
> This looks a lot like giving users a gun to shoot themselves with.

Enabling users of your software beyond what you can think of is not a bad thing.

stuartnelson3:
What are these unwanted labels you're trying to remove?

stuartnelson3:
The default subject can definitely be improved, especially given that it only displays the label values and not the keys they go with. You can filter out unwanted labels using `.Remove`, like what's being done here:

https://github.com/prometheus/alertmanager/blob/master/template/default.tmpl#L4
https://prometheus.io/docs/alerting/notifications/#kv-methods

stuartnelson3:
https://www.youtube.com/watch?v=5h-8V5qY2j0 the first lightning talk might be helpful, too

brian-brazil:
> The default subject can definitely be improved, especially given that it only displays the label values and not the keys they go with

That's on purpose, some mobile phone providers have limits on message size and if it gets too long it becomes an MMS which is far less reliable. What each label value means should be reasonably obvious to the recipient within their context.

sylr:
Yes and that's precisely why I need sprig's functions, [to be able to make a list](https://github.com/Masterminds/sprig/blob/master/docs/lists.md#lists-and-list-functions) and pass is to `.Remove`

brian-brazil:
Usually I'd list each of the labels I want for this sort of thing and render them directly, rather than trying to blacklist the ones I don't.

sylr:
That's a good idea. Would you consider a `.Select` method on label lists ?

brian-brazil:
There's no need, you can do `{.labelname1}{.labelname2}` already.

sylr:
OK bear with me here what I had in mind :

```go
{{ define "__subject" -}}
{{- labelSelection = split .Annotations["alert-labels"] "," -}}
[{{ .Status | toUpper }}{{ if eq .Status "firing" }}:{{ .Alerts.Firing | len }}{{ end }}] 
{{- .GroupLabels.SortedPairs.Values | join " " -}} 
{{- if gt (len .CommonLabels) (len .GroupLabels) }}(
{{- with .CommonLabels.Select $labelSelection -}}
{{ .Values | join " " }}
{{- end -}}
){{- end -}}
{{- end }}
```

That way I don't need to hardcode the labels and duplicate the receivers and I can define the labels that interest me in the alert definition.

It's just the idea, might not be syntactically valid

simonpasquier:
In general I would rather add the strictly needed functionality rather than pulling a whole library like `sprig`. FWIW you could already achieve something similar using `match`:

```
{{ with .CommonLabels }}
{{ $allowed_labels := .alert_labels }}
{{ range $k, $v := . }}
{{ $regexp := printf "(^|,)%s(,|$)" $k }}
{{ if match $regexp $allowed_labels }}{{ $v }} {{ end }}
{{ end }}
{{ end }}
```

sylr:
Thanks for the snippet, I'll try it.

But like @roidelapluie said there are a bunch of functions which could be useful for everyone (date functions, map function, list functions).

And as I said it's the lib used by Helm for the templating so it's not a jump in the dark.

sylr:
Hi,

I need to know if you are going to accept this so that I give some time into improving the documentation to mention Sprig functions.

Thank you.

stuartnelson3:
The consensus appears to be that we're not currently interested in adding the additional functionality. If I've misunderstood the comments from the other maintainers, feel free to write in your support for this.

Issue #1610, opened 2018-11-07
Document how to disable clustering feature


marcan:
Alertmanager randomly failed to start on host reboot with:

```
level=warn ts=2018-11-07T12:01:44.001177603Z caller=cluster.go:148 component=cluster err="couldn't deduce an advertise address: no private IP found, explicit advertise addr not provided"
level=error ts=2018-11-07T12:01:44.015077812Z caller=main.go:199 msg="unable to initialize gossip mesh" err="create memberlist: Failed to get final advertise address: No private IP address found, and explicit IP not provided"
```

I do not use the cluster feature. Unfortunately the documentation does not mention how to turn this off, and it seems to be turned on by default. I guessed that `--cluster.listen-address=` might work, and that seems to have done the trick, but that isn't documented.

Personally, I find this being turned on by default an unwelcome surprise. I do not expect random services to enable networked HA features and open up random ports without explicit configuration (what are the security implications here?). But if it's going to be turned on by default, at least the docs should make it clear so that people know to turn it off if they don't need it.

```
alertmanager, version 0.15.2 (branch: non-git, revision: d19fae3)
  build user:       portage@basestar
  build date:       20180822-11:28:16
  go version:       go1.10.3
```

simonpasquier:
It is turned on by default so a node knowing about no peer at startup can still be contacted by other nodes to form the cluster. Feel free to send a PR to update the documentation.

mxinden:
With https://github.com/prometheus/alertmanager/pull/1654 merged I am closing here. Thanks @geekodour.

@marcan thanks for reporting.

salfredogonzalez:
So how can i disable the option?

stuartnelson3:
As stated in the issue description, `--cluster.listen-address=` will turn off clustering

PR #1654, opened 2018-12-06
added documentation

geekodour:
ref. #1610


geekodour:
@stuartnelson3 that seems better, I'll update it.

PR #1656, opened 2018-12-10
fixes #1649 : Changed DefaultGlobalConfig to a function and added comments

### LINKED ISSUE #1649: Reload doesn't detect change in proxy_url settings
**What did you do?**
Removed the proxy_url setting and reloaded via the API

proxy_url settings to no longer be there.

No change in proxy_url existence.

Changing the config and hitting the /-/reload endpoint; it will detect a change to the details (IP or port) but if I comment the proxy_url line out - it will not detect that. Nor will it detect adding a proxy_url if there wasn't one there before.

A full restart changes the settings appropriately.

**Environment**

* System information:

Darwin 17.7.0 x86_64
Linux 3.10.0-862.3.3.el7.x86_64 x86_64

* Alertmanager version:
```
  alertmanager, version 0.15.2 (branch: HEAD, revision:d19fae3bae451940b8470abb680cfdd59bfa7cfa)
  build user:       root@3101e5b68a55
  build date:       20180814-10:55:04
  go version:       go1.10.3
```
* Prometheus version:

	Not relevant

```
---
global:
  http_config:
    proxy_url: http://127.0.0.1:5432
```

```
Not relevent
```

* Logs:
When reloading there are no errors in the logs:
```
level=info ts=2018-12-04T00:04:29.178120962Z caller=main.go:322 msg="Loading configuration file" file=alertmanager.yml
```

geekodour:
closes  #1649
the variable `DefaultGlobalConfig` was being used to initialize values, but it stored previous information due to which some things were persisting in the newer initialization.

In this PR, `DefaultGlobalConfig` is changed to a function so that it returns a fresh `GlobalConfig`  for initialization.

also added some comments.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -32,7 +32,7 @@ import (
 // Secret is a string that must not be revealed on marshaling.
 type Secret string
 
-// MarshalYAML implements the yaml.Marshaler interface.
+// MarshalYAML implements the yaml.Marshaler interface for Secret.
 func (s Secret) MarshalYAML() (interface{}, error) {
 	if s != "" {
 		return "<secret>", nil
@@ -46,7 +46,7 @@ func (s *Secret) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return unmarshal((*plain)(s))
 }
 
-// MarshalJSON implements the json.Marshaler interface.
+// MarshalJSON implements the json.Marshaler interface for Secret.
 func (s Secret) MarshalJSON() ([]byte, error) {
 	return json.Marshal("<secret>")
 }
@@ -205,7 +205,7 @@ func (c Config) String() string {
 	return string(b)
 }
 
-// UnmarshalYAML implements the yaml.Unmarshaler interface.
+// UnmarshalYAML implements the yaml.Unmarshaler interface for Config.
 func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	// We want to set c to the defaults and then overwrite it with the input.
 	// To make unmarshal fill the plain data struct rather than calling UnmarshalYAML
@@ -219,7 +219,7 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	// We have to restore it here.
 	if c.Global == nil {
 		c.Global = &GlobalConfig{}
-		*c.Global = DefaultGlobalConfig
+		*c.Global = DefaultGlobalConfig()
 	}
 
 	names := map[string]struct{}{}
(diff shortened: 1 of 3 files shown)


geekodour:
@mxinden @simonpasquier  yes i'll add the tests today. :smile:

mxinden:
No rush. Thanks for looking into it.

Would you mind targeting the release-0.16 branch instead of master? As this is a bug fix it would be nice to include it in the v0.16.x series.

geekodour:
@simonpasquier I don't quite understand what to check for in [`TestEmptyFieldsAndRegex()`](https://github.com/prometheus/alertmanager/blob/master/config/config_test.go#L460)

if i add the following: (which is non empty `http_config` key)
```yml
global:
    http_config:
        proxy_url:
```
to
https://github.com/prometheus/alertmanager/blob/f1d33fbcc2bdb6e2661333f954cc540b704633ab/config/testdata/conf.good.yml#L1-L15

but then modify  [`TestEmptyFieldsAndRegex()`](https://github.com/prometheus/alertmanager/blob/master/config/config_test.go#L460) which seems to be testing empty fields and regex as its name suggest and is using the file `conf.empty-fields.yml`

I do not understand why would  [`TestEmptyFieldsAndRegex()`](https://github.com/prometheus/alertmanager/blob/master/config/config_test.go#L460) check for non-empty value of `http_config`

moreover  [`TestEmptyFieldsAndRegex()`](https://github.com/prometheus/alertmanager/blob/master/config/config_test.go#L460) runs `reflect.DeepEqual` so have to write `expectedConf` for `conf.good.yml` ?

---

also is there any way to run a single test? instead of running `make test`?
I tried running `go test config_test.go` which just finds a [bunch of dependencies instead](http://dpaste.com/23KC3V2.txt).

simonpasquier:
My idea was to modify `TestEmptyFieldsAndRegex()` to load first a non-empty configuration: 

```go
func TestEmptyFieldsAndRegex(t *testing.T) {
        ...
        // Load a non-empty configuration to ensure that all fields are overwritten.
        // See https://github.com/prometheus/alertmanager/issues/1649.
	_, _, err := LoadFile("testdata/conf.good.yml")
	if err != nil {
		t.Errorf("Error parsing %s: %s", "testdata/conf.good.yml", err)
	}

	config, _, err := LoadFile("testdata/conf.empty-fields.yml")
	if err != nil {
		t.Errorf("Error parsing %s: %s", "testdata/conf.empty-fields.yml", err)
	}
```

For the test to be relevant, you need to add an `http_config` to [`conf.good.yml`](https://github.com/prometheus/alertmanager/blob/f1d33fbcc2bdb6e2661333f954cc540b704633ab/config/testdata/conf.good.yml#L1).

> also is there any way to run a single test? instead of running make test?

You can do `go test -mod=vendor ./config/...` to use the vendored dependencies instead of the Go modules cache.

geekodour:
@simonpasquier understood, thanks for explaining.
I made the required changes and ran the tests on both master and this branch,
they behave as expected.

brian-brazil:
I think there might be other places where this is happening.

brian-brazil:
I was investigating a bug which turned out to be something else, though I did notice this pattern in a few places. That we don't use this pattern anywhere in Prometheus or the AM currently is a bit suspicious.

mxinden:
Alright. If there are no further objections I will go ahead and merge.

mxinden:
Thanks @geekodour for the fix!

Issue #1657, opened 2018-12-11
Support for --log.format=json


kfdm:
Related to #1128 and prometheus/prometheus#4876 it should be reasonably? easy to update the prometheus/common package to support the same `--log.format=json`

I imagine this is already too late for 0.16.0 but perhaps this can go in the next version.

I can try to give this PR a try myself, but my go is weak so if someone gets to it before I do, that would also work :)

simonpasquier:
Closed by #1658

PR #1658, opened 2018-12-11
Update prometheus/common and add support for --log.format

kfdm:
I am not sure if I did this correctly, but I'm willing to accept guidance on how to format things properly. :)

This should handle --log.format for #1657

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -38,6 +38,7 @@ import (
 	"github.com/prometheus/alertmanager/config"
 	"github.com/prometheus/alertmanager/dispatch"
 	"github.com/prometheus/alertmanager/inhibit"
+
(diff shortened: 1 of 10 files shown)


kfdm:
Ok, I'll take care of that tomorrow. :)

kfdm:
Assuming that Prometheus is considered an external package from alertmanager's perspective, I believe I have grouped the packages as requested.

simonpasquier:
Agreed, the problem exists in the current code base and isn't specific to your change. This group should be split into 2:

https://github.com/prometheus/alertmanager/blob/f1d33fbcc2bdb6e2661333f954cc540b704633ab/cmd/alertmanager/main.go#L33-L53

simonpasquier:
Ah sorry, I missed the latest update! All good then :+1:

PR #1664, opened 2018-12-12
Use an initial regex to shorten search list

stuartnelson3:
With so many receivers, a lot can be dropped with
an initial regex search. This greatly speeds up
the ranking done by the jaro-winkler search.

@grobie


stuartnelson3:
@w0rm when you have a chance, could you help me apply this same regex match to https://github.com/prometheus/alertmanager/blob/e768371a3917bf0f5dbe8dd326cf3f35a6bc26b0/ui/app/src/Views/GroupBar/Updates.elm#L112 ? It seems like doing this initial rough filtering makes the search a lot faster.

ui/app/src/Utils/Match.elm
```diff
@@ -1,9 +1,32 @@
...
 
+filter : String -> List { a | name : String } -> List { a | name : String }
+filter search list =
+    let
+        re =
+            "(.+)?"
+                ++ (String.split "" search |> String.join "(.+)?")
+                ++ "(.+)?"
+                |> Regex.fromString
+                |> Maybe.withDefault Regex.never
```
stuartnelson3: Any recommendation on how to make this less messy is greatly appreciated :)

stuartnelson3:
@simonpasquier do you have an alertmanager with many receivers to test this against? It seems like  a much better experience at SC with these changes.

ui/app/src/Views/ReceiverBar/Views.elm
```diff
@@ -1,5 +1,6 @@
 module Views.ReceiverBar.Views exposing (view)
 
+import Debouncer.Messages as Debouncer
```
stuartnelson3: todo: remove

ui/app/src/Views/AlertList/Views.elm
```diff
@@ -62,7 +63,7 @@ view { alerts, groupBar, filterBar, receiverBar, tab, activeId } filter =
             ]
         , case alerts of
             Success alerts_ ->
-                alertGroups activeId filter groupBar alerts_
+                Lazy.lazy4 alertGroups activeId filter groupBar alerts_
```
w0rm: 👍

ui/app/src/Utils/Match.elm
w0rm: Maybe this would read nicer:

```
("(.+)?" ++ String.join "(.+)?" (String.split "" search) ++ "(.+)?")
```

ui/app/src/Utils/Match.elm
```diff
@@ -1,9 +1,32 @@
...
+    in
+    list
+        |> List.filter
+            (\x ->
```
w0rm: This could be `(Regex.find re x.name |> List.head) /= Nothing`

ui/app/src/Utils/Match.elm
```diff
@@ -1,9 +1,32 @@
...
+import Regex
 import Utils.List exposing (zip)
 
 
+filter : String -> List { a | name : String } -> List { a | name : String }
+filter search list =
+    let
+        re =
+            "(.+)?"
+                ++ (String.split "" search |> String.join "(.+)?")
```
w0rm: What if search contains meaningful regex characters, e.g. `.`?

stuartnelson3: That's a good point. If someone wants to use the receiver bar as a regex, for example `mysql-.*`, then this will create an unexpected regex.

SuperQ:

PR #1666, opened 2018-12-13
Backport Lookup parts in strings using regexp.MatchString into release-0.15

sylr:
cherry picked from commit 4e8a910b9d78d3c1b5f011e9e59efb4c5f2b765e

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,7 @@ build-all: assets build
 assets: go-bindata ui/bindata.go template/internal/deftmpl/bindata.go
 
 go-bindata:
-	-@$(GO) get -u github.com/jteeuwen/go-bindata/...
+	-@$(GO) get -u github.com/jteeuwen/go-bindata/go-bindata
 
 template/internal/deftmpl/bindata.go: template/default.tmpl
 	@go-bindata $(bindata_flags) -mode 420 -modtime 1 -pkg deftmpl -o template/internal/deftmpl/bindata.go template/default.tmpl
--- a/template/template.go
+++ b/template/template.go
@@ -128,6 +128,7 @@ var DefaultFuncs = FuncMap{
 	"join": func(sep string, s []string) string {
 		return strings.Join(s, sep)
 	},
+	"match": regexp.MatchString,
 	"safeHtml": func(text string) tmplhtml.HTML {
 		return tmplhtml.HTML(text)
 	},


sylr:
I would need this into release-0.15 to improve my notification templates because I can't move yet to master or 0.16.0-alpha as `Unsee` & `Karmaa` do not work with it due to the removal of the `/api/v1/alerts/groups endpoint`.

mxinden:
Mhh, the DCO bot is still complaining. Could you maybe try changing the order of the `Signed-off` lines?

Makefile
```diff
@@ -31,7 +31,7 @@ build-all: assets build
 assets: go-bindata ui/bindata.go template/internal/deftmpl/bindata.go
 
 go-bindata:
-	-@$(GO) get -u github.com/jteeuwen/go-bindata/...
+	-@$(GO) get -u github.com/jteeuwen/go-bindata/go-bindata
```
mxinden: What was the issue with `...`?

sylr: It just does not work anymore: https://github.com/golang/go/issues/29241

brian-brazil:
We generally don't maintain older versions, and even then it'd only be considered for critical bug fixes - which this is not.

simonpasquier:
:+1: for me. Security-wise, it would also be nice to have the current release built with the latest Go version.

sylr:
@mxinden that would be brilliant :)

stuartnelson3:
As @brian-brazil said, we don't typically maintain older versions, and cutting a release to include a templating function for a single user feels a bit weird to me, but if you want to then 👍 I would prefer not to make this a habit, though.

sylr:
I knew it was too good to be true ...

A breaking change has been introduced in a minor version (v0.16.0-alpha.0: removal /api/v1/alerts/groups) making me currently unable to use it.

Meanwhile I need to move forward and improve the alertmanager's alerts because my colleagues which are part of the on call rotation have qualified the stacking of labels in the title to be messy and unhelpful (not really their words but let's stay polite) and urged me to do something about it.

I had something better for the templating but I needed the `match` function which did not make the 0.15.3 release.

But anyway, I took the matter in my own hands and compiled my own branch.

Thanks @mxinden to have proposed to make a release, It's nice to see there are persons in this project which care a bit about the actual users.

PR #1667, opened 2018-12-14
api/v2: Add CORS support

simonpasquier:
We do something similar for the v1 API.

https://github.com/prometheus/alertmanager/blob/cd4a524848ca037f7daa1a27a238f731c0e4c0c7/api/v1/api.go#L132-L139

It will also help Prometheus detecting if the AlertManager supports the v2 API or not.

### DIFF
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -21,6 +21,16 @@ import (
 	"sync"
 	"time"
 
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"github.com/go-openapi/loads"
+	"github.com/go-openapi/runtime/middleware"
+	"github.com/go-openapi/strfmt"
+	prometheus_model "github.com/prometheus/common/model"
+	"github.com/prometheus/common/version"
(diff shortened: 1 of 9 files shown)


api/v2/api.go
```diff
@@ -103,7 +102,8 @@ func NewAPI(alerts provider.Alerts, sf getAlertStatusFn, silences *silence.Silen
 
 	openAPI.Logger = func(s string, i ...interface{}) { level.Error(api.logger).Log(i...) }
 
-	api.Handler = openAPI.Serve(nil)
+	handleCORS := cors.New(cors.Options{}).Handler
```
mxinden: I guess one could also do `handler := cors.Default().Handler`, not sure if `cors.Options{}` always does the default?

simonpasquier:
I'm not that familiar with CORS but the library claims to support the [W3C CORS recommandation](https://www.w3.org/TR/cors/) which is probably more than adding a set of hard-coded headers to the response.

mxinden:
Thanks @simonpasquier!

PR #1674, opened 2018-12-19
Enrich templating functions

sylr:
This PR does two things.

First it adds two new functions usable in templates:

 - `split(separator, string) []string`: convert a string to an array of strings
 - `urlEncode(string) string`: convert a string to another usable in URL query parameters

Then it adds a `Select([]string) KV` method on the KV struct.

The goal of all of this is to improve the templates of pagerduty, i.e., selecting only useful labels in the alert description:

```
        pagerduty_configs:
        - routing_key: xxxxxxxxxxxxxxxxxxxxxxxxxx
          details:
            firing: |
              {{ range .Alerts.Firing }}Labels:
              {{ range .Labels.SortedPairs -}}
              - {{ .Name }} = {{ .Value }}
              {{ end }}
              Annotations:
              {{ range .Annotations.SortedPairs -}}
              - {{ .Name }} = {{ .Value }}
              {{ end }}
              Source: {{ .GeneratorURL }}

              {{ end }}
          description: |
            [{{ .Status | toUpper }}{{ if eq .Status "firing" }}:{{ .Alerts.Firing | len }}{{ end }}] {{ .GroupLabels.SortedPairs.Values | join " " }}{{ " " }}
            {{- if gt (len .CommonLabels) (len .GroupLabels) -}}
              (
              {{- if .CommonAnnotations.alert_title_labels -}}
                {{- $allowed_labels := split "," .CommonAnnotations.alert_title_labels -}}
                {{- with .CommonLabels.Remove .GroupLabels.Names -}}
                  {{- with .Select $allowed_labels -}}
                    {{- .Values | join " " -}}
                  {{- end -}}
                {{- end -}}
              {{- else -}}
                {{- with .CommonLabels.Remove .GroupLabels.Names -}}
                  {{- .Values | join " " -}}
                {{- end -}}
              {{- end -}}
              )
            {{- end -}}
```

If you decide to surprise me for once and merge this I'll make a PR for the documentation.

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -134,6 +134,9 @@ var DefaultFuncs = FuncMap{
 	"join": func(sep string, s []string) string {
 		return strings.Join(s, sep)
 	},
+	"split": func(sep string, s string) []string {
+		return strings.Split(s, sep)
+	},
 	"match": regexp.MatchString,
 	"safeHtml": func(text string) tmplhtml.HTML {
 		return tmplhtml.HTML(text)
--- a/template/template.go
+++ b/template/template.go
@@ -141,6 +141,9 @@ var DefaultFuncs = FuncMap{
 	"safeHtml": func(text string) tmplhtml.HTML {
 		return tmplhtml.HTML(text)
 	},
+	"urlEncode": func(text string) string {
+		return url.QueryEscape(text)
+	},
 	"reReplaceAll": func(pattern, repl, text string) string {
 		re := regexp.MustCompile(pattern)
 		return re.ReplaceAllString(text, repl)
--- a/template/template.go
+++ b/template/template.go
@@ -212,6 +212,22 @@ func (kv KV) Remove(keys []string) KV {
 	return res
 }
 
+// Select returns a copy of the key/value set with the given keys.
+func (kv KV) Select(keys []string) KV {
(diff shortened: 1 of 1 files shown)


sylr:
> This is not necessary

It's certainly isn't but it is a reflection of the frustration resulting from trying to improve your projects (prometheus, alertmanager, client_golanfg) based on real corporate use cases and being almost constantly denied.

stuartnelson3:
>It's certainly isn't but it is a reflection of the frustration resulting from trying to improve your projects (prometheus, alertmanager, client_golang) based on real corporate use cases and being almost constantly denied.

A pull request is not the correct forum for this discussion. If you would like to discuss your frustrations in a constructive manner, I recommend writing to the mailing list.

brian-brazil:
> split(separator, string) []string

The example you give here is a layering violation, alert rules shouldn't decide how they are formatted and what label are important - that's for the alertmanager.

> urlEncode(string) string: convert a string to another usable in URL query parameters

Go templates should be doing this automatically. Is there a case where they aren't?

sylr:
> The example you give here is a layering violation, alert rules shouldn't decide how they are formatted and what label are important - that's for the alertmanager.

So your solution is to make a receiver per alert to have the correct templating ?

Issue #1682, opened 2018-12-23
Allow negative (re) matching in silences


juliusv:
Similar to https://github.com/prometheus/alertmanager/issues/1023, but for silences.

Currently you can only have `=` and `=~` matchers in silences. I recently wanted to silence all jobs from a given Prometheus server *except* for some specific ones, and being able to do a negative regex match would have been useful there, especially since RE2 doesn't support negative lookaheads.

geekodour:
Hi @juliusv I'd like to work on this, if it's not a high priority issue.
Will implementing `!~` is what we want here, so that the command could work?
```
$ amtool silence add job !~".+exporter" instance="someinstance"
```

juliusv:
@geekodour Awesome!

I was actually talking about the web UI, but then of course `amtool` should support it as well. I haven't talked to any of the Alertmanager maintainers yet whether there is a fundamental reason why we don't support this yet (can't think of any), but this would probably be good to check first.

roidelapluie:
I think getting it implemented at the UI level will be difficult and plenty of discussions, whereas getting it done at amtool level should be straightforward.

stuartnelson3:
>I haven't talked to any of the Alertmanager maintainers yet whether there is a fundamental reason why we don't support this yet (can't think of any), but this would probably be good to check first.

I think this is most likely an artifact of the original implementation, which only supported direct or regex matches.

Adding this to `amtool` should be straightforward, as we already have the `!~` and `!=` concept from prometheus. How to handle this in the ui would (to me) be more a question of what makes most sense for selecting it. Maybe a simple dropdown is enough?

juliusv:
@stuartnelson3 A `==` / `!=` / `=~` / `!~` dropdown between the label name and label value would make sense to me as a first step. Then you could get rid of the regex checkbox to the right.

geekodour:
@juliusv working on this PR
It seems the following documented syntax does not work too:
https://github.com/prometheus/alertmanager/blob/cfc0d9c55882dd501ec881ebe4606cf8d78bf8b6/cli/silence_add.go#L61-L64

will it be fine if add this into the same PR?

stuartnelson3:
bah, it appears that got dropped at some point. I added it back to `amtool alert query` in https://github.com/prometheus/alertmanager/pull/1575.

It should (hopefully) a quick change, I guess just make sure it's in a separate commit in the PR.

geekodour:
@stuartnelson3 

[...]

PR #1693, opened 2019-01-05
support for assuming first label is alertname in silence add and query

geekodour:
ref #1575
This PR adds support for assuming the first label is `alertname` in silence add and query. It was already existing on silence query, did some cleanup.

Will create another PR for #1682


stuartnelson3:
Can you try rebasing on latest master and pushing? it seems to be failing on something that was fixed only yesterday.

PR #1699, opened 2019-01-09
feat(api): add /api/version handler for getting version

prymitive:
With API v2 there are now 2 API endpoints a client could use, but there is no dicovery mechanism to tell which API version are avaiable, so a client needs to make assumptions or try all possible versions.
Part of the problem is that there is no non-versioned API endpoint that exposes version information, it's only avaiable via api/v*/status.
This adds api/version endpoint that's only returing alertmanager version, this information can be used by the client to decide which API version to use. It doesn't fully resolve discoverability problem since it requires a client to keep information about all Alertmanager releases, but it helps while being useful on it's own.

Signed-off-by: Łukasz Mierzwa <l.mierzwa@gmail.com>


api/version.go
```diff
@@ -0,0 +1,45 @@
...
+// VersionHandler returns Alertmanager version
+func VersionHandler(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(http.StatusOK)
+
+	b, err := json.Marshal(&versionResponse{
+		Version: version.Version,
+	})
+	if err != nil {
+		log.Printf("Error marshalling JSON: %s", err)
```
prymitive: Note that this is using stdlib logger package, not go-kit.
To use go-kit it looks like I would need to pass extra information from main.go into this handler OR a logger instance. Felt like it wouldn't be a pure handler in that case and I wasn't sure what's the best pattern for it, api subpackages are biggish so they create a logger for the instance of api object, while this is just a view handler.

prymitive:
Not sure if this is the best way, but having a version agnostic endpoint for getting version seems like a good thing to have IMHO.

brian-brazil:
You can get this information from the build info on /metrics.

prymitive:
Using '/metrics' is definitely possible and would work for me, but it's a different format, wouldn't it be better to have this as JSON for an API client?
Either way I don't have strong feelings about it, but having `api/version` seems like a low hanging fruits, assuming that people agree it's beneficial, if not I'm happy to use `/metrics` in my code.

brian-brazil:
One thing is, what's the format of this version endpoint? Is it following v1, v2, or a future v3?

prymitive:
The intention for `api/version` is to only contain version and to not be a part of any versioned api, so effectively it would have it's own format that wouldn't need to change.

simonpasquier:
Late to the party but I concur with @mxinden. Thanks @prymitive for your understanding!

Issue #1700, opened 2019-01-09
PagerDuty 400 BadRequest for Alerts with len(summary)>1024


sbueringer:
**What did you do?**
Alert with too long description was pushed to Alertmanager.

**What did you expect to see?**
Alert should have been pushed to PagerDuty.

**What did you see instead? Under which circumstances?**
We got 400 BadRequest (log: "unexpected status code 400").

Hi,
we had an issue with our Alerts which are routed from Prometheus to Alertmanager to Pagerduty. We used a template for the summary field, which in some cases lead to summaries with more than 1024 characters, which isn't supported by PagerDuty (https://v2.developer.pagerduty.com/docs/send-an-event-events-api-v2). Unfortunatley, this wasn't directly visible to us and can only be detected by monitoring the Alertmanager log file. The result was that we just missed the Alerts.

I think it would be a good idea to trim the summary to 1024 characters (e.g. like this https://github.com/c445/alertmanager/blob/2808c2851995a0db7e57d9b8f5bc5fbd986514f9/notify/impl.go#L526-L539) before sending the Alert to Pagerduty. Also I think it would be nice to have a few debug log statements in this area, because there was no way without changing the source code & recompiling to find out the response body from Pagerduty.

If it's ok for you, I can open a PR to address this issue.

stuartnelson3:
Opening a PR sounds great, looking forward to it!

gouthamve:
We're still seeing this on `0.17.0`:
```
level=error ts=2019-06-10T14:52:29.294545153Z caller=notify.go:339 component=dispatcher msg="Error on notify" err="cancelling notify retry for \"pagerduty\" due to unrecoverable error: Event object is invalid: Description is too long (maximum is 1024 characters)"
```

Not sure what's wrong though. Can we reopen?

simonpasquier:
@gouthamve got non-ASCII characters in the description?

stuartnelson3:
The docs only say "characters", so I'm not sure if they actually mean "bytes". I'm guessing they mean "bytes", if we're running into this issue still. Guess we should include tests for all these trim functions we're adding.

simonpasquier:
Hmm, I tried to reproduce (it happens that you can post to the PagerDuty API with any random routing key and get a 200 if the payload is valid!) but couldn't. When I modify the code to not truncate the description field at 1024 characters, the Pager Duty API returns the same error as you pasted.

simonpasquier:
[...]

PR #1701, opened 2019-01-09
trim PagerDuty message summary to 1024 chars, add PagerDuty debug log

sbueringer:
Fixes https://github.com/prometheus/alertmanager/issues/1700.

PagerDuty Alerts are rejected (a 400 BadRequest is sent back from PagerDuty)
when the summary field is longer than 1024 characters
(https://v2.developer.pagerduty.com/docs/send-an-event-events-api-v2).

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -634,7 +634,7 @@ func (n *PagerDuty) notifyV2(
 		if err != nil {
 			return false, fmt.Errorf("failed to read error response from PagerDuty(status: %d): %v", resp.StatusCode, err)
 		}
-		level.Debug(n.logger).Log("msg", fmt.Sprintf("Received error response from PagerDuty(status: %d): %s", resp.StatusCode, string(body)), "incident", key)
+		level.Debug(n.logger).Log("msg", "Received error response from PagerDuty", "incident", key, "code", resp.StatusCode, "body", string(body))
 	}
 
 	return n.retryV2(resp.StatusCode)
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -632,7 +632,7 @@ func (n *PagerDuty) notifyV2(
 	if resp.StatusCode == http.StatusBadRequest {
 		body, err := ioutil.ReadAll(resp.Body)
 		if err != nil {
-			return false, fmt.Errorf("failed to read error response from PagerDuty(status: %d): %v", resp.StatusCode, err)
+			return false, fmt.Errorf("failed to read error response from PagerDuty (status: %d): %v", resp.StatusCode, err)
 		}
 		level.Debug(n.logger).Log("msg", "Received error response from PagerDuty", "incident", key, "code", resp.StatusCode, "body", string(body))
 	}
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -577,6 +577,12 @@ func (n *PagerDuty) notifyV2(
 		n.conf.Severity = "error"
 	}
 
+	summary := tmpl(n.conf.Description)
+	summaryRunes := []rune(summary)
+	if len(summaryRunes) > 1024 {
+		summary = string(summaryRunes[:1018]) + " [...]"
+	}
+
 	msg := &pagerDutyMessage{
 		Client:      tmpl(n.conf.Client),
 		ClientURL:   tmpl(n.conf.ClientURL),
@@ -586,7 +592,7 @@ func (n *PagerDuty) notifyV2(
 		Images:      make([]pagerDutyImage, len(n.conf.Images)),
 		Links:       make([]pagerDutyLink, len(n.conf.Links)),
 		Payload: &pagerDutyPayload{
-			Summary:       tmpl(n.conf.Description),
+			Summary:       summary,
 			Source:        tmpl(n.conf.Client),
 			Severity:      tmpl(n.conf.Severity),
 			CustomDetails: details,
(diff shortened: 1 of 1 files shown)


notify/impl.go
```diff
@@ -577,6 +577,10 @@ func (n *PagerDuty) notifyV2(
 		n.conf.Severity = "error"
 	}
 
+	summary := tmpl(n.conf.Description)
+	if len(summary) > 1024 {
```
simonpasquier: This won't be always correct if the string contains UTF-8 characters larger than a byte.

https://play.golang.org/p/LSIA-RjSXHU

notify/impl.go
```diff
@@ -613,15 +617,21 @@ func (n *PagerDuty) notifyV2(
...
-		return true, err
+		return true, fmt.Errorf("failed to post message to PagerDuty: %v", err)
 	}
 	defer resp.Body.Close()
 
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return true, fmt.Errorf("failed to parse response from PagerDuty: %v", err)
+	}
+	level.Debug(n.logger).Log("msg", fmt.Sprintf("Received response from PagerDuty(status: %d): %s", resp.StatusCode, string(body)), "incident", key)
```
simonpasquier: Maybe log only when PagerDuty replies with 400? Looking at https://v2.developer.pagerduty.com/docs/events-api-v2#api-response-codes--retry-logic the retry logic implemented by the AlertManager isn't correct as we retry on all responses other than `2XX`.

notify/impl.go
sbueringer: Fixed via: https://play.golang.org/p/SbFfy6LLUiZ

notify/impl.go
sbueringer: Fixed.
Are you sure the retry logic is wrong? https://github.com/prometheus/alertmanager/blob/ec7b4686fdfab2d530407ce28250e29a1e7e8178/notify/impl.go#L712-L714
Looks like we're not retrying on 400.

notify/impl.go
```diff
@@ -613,15 +619,24 @@ func (n *PagerDuty) notifyV2(
...
-		return true, err
+		return true, fmt.Errorf("failed to post message to PagerDuty: %v", err)
 	}
 	defer resp.Body.Close()
 
+	// See: https://v2.developer.pagerduty.com/docs/events-api-v2#api-response-codes--retry-logic
+	if resp.StatusCode == http.StatusBadRequest {
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return true, fmt.Errorf("failed to parse response from PagerDuty: %v", err)
```
stuartnelson3: The link you've provided says to not retry on 400BadRequest; shouldn't this be `false`?

Adding some more context to the error would also probably be helpful, like letting the user know that the failure was in parsing an error response.

notify/impl.go
simonpasquier: Sorry I somehow misread the code...

notify/impl.go
sbueringer: Oh sorry of course your right, I changed it to false. I also added some more context and replaced "failed to parse" with "failed to read" because the body is not actually parsed here.

notify/impl.go
```diff
@@ -577,6 +577,12 @@ func (n *PagerDuty) notifyV2(
 		n.conf.Severity = "error"
 	}
 
+	summary := tmpl(n.conf.Description)
+	summaryRunes := []rune(summary)
+	if len(summaryRunes) > 1024 {
+		summary = string(summaryRunes[:1024])
```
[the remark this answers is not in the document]
sbueringer: Yep, why not. I adjusted the PR.

sbueringer:
PTAL @simonpasquier

notify/impl.go
```diff
@@ -613,15 +619,24 @@ func (n *PagerDuty) notifyV2(
...
 	}
 	defer resp.Body.Close()
 
+	// See: https://v2.developer.pagerduty.com/docs/events-api-v2#api-response-codes--retry-logic
+	if resp.StatusCode == http.StatusBadRequest {
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return false, fmt.Errorf("failed to read error response from PagerDuty(status: %d): %v", resp.StatusCode, err)
+		}
+		level.Debug(n.logger).Log("msg", fmt.Sprintf("Received error response from PagerDuty(status: %d): %s", resp.StatusCode, string(body)), "incident", key)
```
simonpasquier: I would use additional keys to report the status code and response:

```go
level.Debug(n.logger).Log("msg", "Received error response from PagerDuty", "incident", key, "code", resp.StatusCode, "body", string(body))
```

notify/impl.go
```diff
@@ -613,15 +619,24 @@ func (n *PagerDuty) notifyV2(
...
-		return true, err
+		return true, fmt.Errorf("failed to post message to PagerDuty: %v", err)
 	}
 	defer resp.Body.Close()
 
+	// See: https://v2.developer.pagerduty.com/docs/events-api-v2#api-response-codes--retry-logic
+	if resp.StatusCode == http.StatusBadRequest {
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return false, fmt.Errorf("failed to read error response from PagerDuty(status: %d): %v", resp.StatusCode, err)
```
simonpasquier: Add a space between `PagerDuty` and `(`.

sbueringer: Done

notify/impl.go
sbueringer: Fixed like you suggested.

notify/impl.go
simonpasquier: Hmm, it is still there...

sbueringer: Sorry I somehow missed that I had this twice, in the log statement and in the error message... . Now it should really be fixed :)

Issue #1702, opened 2019-01-10
Inconsistency in generation of .proto files


geekodour:
I tried generating `.proto` files using `make proto`

It generated files with a lot of differences as discussed in #1682 and the helpful top comment also got removed. (gist attached)

As alertmanager uses the gogo/protobuf implementation, 

from the gogo/protobuf readme
```
protoc-gen-gogofast (same as gofast, but imports gogoprotobuf)
protoc-gen-gogofaster (same as gogofast, without XXX_unrecognized, less pointer fields)
protoc-gen-gogoslick (same as gogofaster, but with generated string, gostring and equal methods)
```
and `genproto.sh` uses `gogofast`
https://github.com/prometheus/alertmanager/blob/dba283edd0d05c6a295b6b5c63bc1bb692f7683d/scripts/genproto.sh#L30

so it is supposed to generate the `XXX_unrecognized` when using `gogofast` but the files in `master` do not show any `XXX_unrecognized` 

**What did you do?**
```
- Install protoc`v3.5.1`
- go install github.com/gogo/protobuf/protoc-gen-gogo
- go install github.com/gogo/protobuf/protoc-gen-gogofast
- go get github.com/grpc-ecosystem/grpc-gateway/protoc-gen-grpc-gateway
- go get golang.org/x/tools/cmd/goimports
```

I even tried installing downto `v0.4` of gogo/protobuf for `protoc-gen-gogo` 

**What did you expect to see?**
What @juliusv generated,
diff: https://gist.github.com/juliusv/033a489a93c118be9184457c44b9cc63

**What did you see instead? Under which circumstances?**
`make proto`
diff: https://gist.github.com/geekodour/52ae6d8283996b88c064526f97aafdcf

**Environment**
* System information:
Linux 4.13.9-300.fc27.x86_64 x86_64
* Alertmanager version:
master

cc: @fabxc @juliusv

simonpasquier:
This is because `scripts/genproto.sh` doesn't lock down the versions of its dependencies. Eventually we need to implement the same mechanism as https://github.com/prometheus/prometheus/pull/5075.

PR #1707, opened 2019-01-14
Added version locking for protobuf extensions in genproto.sh

geekodour:
Fixes: #1702 

Signed-off-by: Hrishikesh Barman <hrishikeshbman@gmail.com>


mxinden:
@simonpasquier would you mind looking into this with your experience through https://github.com/prometheus/prometheus/pull/5075?

geekodour:
@simonpasquier I am having some issues understanding the rationale behind using `tools.go`. then installing it using `go install` I'll look into this again tomorrow.

the I do:
(the commit id is the exact version from the current script in this PR)
```
$ go get github.com/gogo/protobuf/protoc-gen-gogo@971cbfd2e72b513a28c74af7462aee0800248d69 
```
then the following is added into `go.mod`
```
github.com/gogo/protobuf v0.0.0-20171123125729-971cbfd2e72b
```

can you look into it?

simonpasquier:
The idea behind `tools.go` is that it will the module recorded in `go.mod`. Otherwise running `make unused` will remove it. You can also notice that `go install` is used with the `-mod=vendor` flag which uses the files vendored with the project.

simonpasquier:
@geekodour yes any change to the `vendor/` directory should be committed as well. The idea is that AlertManager's builds are reproducible even when the external dependencies are unavailable or tampered (though this last case shouldn't happen with Go modules).

geekodour:
checking the failed test.

simonpasquier:
Last thing: please add a step to Circle CI that would run `make proto` to ensure that the generated code always matches with the protobuf definition. See https://github.com/prometheus/prometheus/blob/24f19f03dbcb8791837b5d4e95827c180c319497/.circleci/config.yml#L19-L26

.circleci/config.yml
```diff
@@ -19,6 +19,14 @@ jobs:
     - setup_remote_docker
     - run: make promu
     - run: make
+    - run:
+       command: |
+         curl -s -L https://github.com/protocolbuffers/protobuf/releases/download/v3.5.1/protoc-3.5.1-linux-x86_64.zip > /tmp/protoc.zip
```
mxinden: Great, thanks for also adding this!

PR #1711, opened 2019-01-16
api/v2: Disable serving swagger spec and redoc UI

mxinden:
By default go-swagger serves the swagger spec and the redoc UI. This
patch disables both.

This does not need to be permanent, but for now I would prefer to reduce the attach-surface. What are your thoughts?

//CC @metalmatze (Thanks for the catch!)


simonpasquier:
:+1:

PR #1713, opened 2019-01-16
WIP: Check that AuthUsername is configured before trying SMTP authentication methods

metazool:
This small change fixes the problem described in this mailing list thread: https://groups.google.com/forum/#!topic/prometheus-users/bM8vFlJ0WBY 

Where an SMTP server advertises the AUTH extension but does not require it, and alertmanager is intentionally configured without any credentials.

Don't know if it is robust enough for you @mxinden @stuartnelson3 ... I built and tested this change using the Dockerfile at https://hub.docker.com/r/prom/alertmanager/dockerfile but I had to update the base image to golang:1.11


simonpasquier:
I'm on the fence regarding this change as it might make it less easy to troubleshoot when you forget to provide the SMTP username and your server requires authentication. Maybe a configuration field explicitly disabling authentication (eg `disable_authentication: true`) would be a better approach. But again it looks like a very specific case which is better resolved on the server side.

metazool:
Thank you for having a proper consideration of this. I have no idea how common it is for SMTP servers to support the AUTH extension but not require it. I tried poking at some of the Gmail SMTP relays listed in this excellent SO answer but couldn't get a session going. https://stackoverflow.com/a/29828911/4319767 

Making common configuration problems harder to debug is definitely a mistake for this PR. However it would be good to know if others come across this problem. It might be reasonably common within internal corporate networks where alertmanager typically resides? 

I will keep trying to get this sorted out on the server side and in the meantime this is a workaround. Thanks again for your feedback @simonpasquier

simonpasquier:
Ok, this is a concrete issue for many users (#1721). While relaxing the constraints, we need to account for systems that require authentication and the AlertManager configuration lacking credentials...

simonpasquier:
As [suggested](https://github.com/prometheus/alertmanager/issues/1721#issuecomment-457530668) by @hoffie, it would make sense to return early from `auth()` if `n.conf.AuthUsername` is empty and log a debug message there.

notify/impl.go
```diff
@@ -221,6 +221,13 @@ func NewEmail(c *config.EmailConfig, t *template.Template, l log.Logger) *Email
 // auth resolves a string of authentication mechanisms.
 func (n *Email) auth(mechs string) (smtp.Auth, error) {
 	username := n.conf.AuthUsername
+
+	// If no username is set, keep going without authentication
```
simonpasquier: missing dot at the end of the comment.

notify/impl.go
```diff
@@ -327,6 +334,8 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 		}
 	}
 
+	// If SMTP authentication is configured, check it
+
```
simonpasquier: those 2 comment lines can be removed.

notify/impl.go
```diff
@@ -1551,6 +1560,21 @@ func (a *loginAuth) Start(server *smtp.ServerInfo) (string, []byte, error) {
...
 }
 
+type noAuth struct {}
+
+func NoAuth() smtp.Auth {
+	return &noAuth{}
+}
+
+func (a *noAuth) Start(server *smtp.ServerInfo) (string, []byte, error) {
+	// return proto == "" to indicate that the authentication should be skipped."
```
simonpasquier: Leading `"` to be removed.

notify/impl.go
metazool: i'm very new to Go. i will sort the formatting issues after lunch. i felt awkward about having to adapt the surrounding tests to fit this approach but its better than adding to the code smell in the Notify function. wish i had time and scope to get stuck right into it ;)

notify/impl.go
```diff
@@ -221,6 +221,13 @@ func NewEmail(c *config.EmailConfig, t *template.Template, l log.Logger) *Email
 // auth resolves a string of authentication mechanisms.
 func (n *Email) auth(mechs string) (smtp.Auth, error) {
 	username := n.conf.AuthUsername
+
+	// If no username is set, keep going without authentication.
+	if n.conf.AuthUsername == "" {
+		level.Debug(n.logger).Log("msg", "smtp_auth_username is not configured. Attempting to send email without authenticating")
```
simonpasquier: suggestion: `attempting to send email without authentication because smtp_auth_username is empty`

notify/impl.go
```diff
@@ -1551,6 +1558,21 @@ func (a *loginAuth) Start(server *smtp.ServerInfo) (string, []byte, error) {
 	return "LOGIN", []byte{}, nil
 }
 
+type noAuth struct{}
```
simonpasquier: Please move this after `func (a *loginAuth) Next(fromServer []byte, more bool) ([]byte, error) {...}`

metazool: Hah, and there i was wondering why `loginAuth` didnt have a `Next` interface

metazool:
As you can see i've got in a massive git pickle now trying to retrospectively sign commits, thanks for all your comments, no time left for this today!

metazool:
Ok I have replaced this PR with #1739 now and will withdraw this one, thanks for all the feedback

PR #1724, opened 2019-01-25
replace labels to const to reduce hardcode

yeya24:
In cluster/delegate.go, I change the label name full_state and update to const value.
Signed-off-by: yeya24 <ben.ye@daocloud.io>

### DIFF
--- a/cluster/delegate.go
+++ b/cluster/delegate.go
@@ -24,8 +24,12 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-// Maximum number of messages to be held in the queue.
-const maxQueueSize = 4096
+const (
+	// Maximum number of messages to be held in the queue.
+	maxQueueSize = 4096
+	fullState    = "full_state"
+	update       = "update"
+)
 
 // delegate implements memberlist.Delegate and memberlist.EventDelegate
 // and broadcasts its peer's state in the cluster.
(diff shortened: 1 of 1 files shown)


cluster/delegate.go
```diff
@@ -25,8 +25,12 @@ import (
 )
 
 // Maximum number of messages to be held in the queue.
```
mxinden: This comment would need to be moved below `const (`.

yeya24: done

cluster/delegate.go
```diff
@@ -24,9 +24,13 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-// Maximum number of messages to be held in the queue.
-const maxQueueSize = 4096
+const (
+	// Maximum number of messages to be held in the queue.
+	maxQueueSize = 4096
```
mxinden: Can you run gofmt on this?

```diff
diff -u ./cluster/delegate.go.orig ./cluster/delegate.go
--- ./cluster/delegate.go.orig	2019-01-25 18:08:08.853426105 +0000
+++ ./cluster/delegate.go	2019-01-25 18:08:08.853426105 +0000
@@ -27,10 +27,10 @@
 const (
 	// Maximum number of messages to be held in the queue.
 	maxQueueSize = 4096
-	fullState = "full_state"
-	update = "update"
-
+	fullState    = "full_state"
+	update       = "update"
 )
+
 // delegate implements memberlist.Delegate and memberlist.EventDelegate
 // and broadcasts its peer's state in the cluster.
 type delegate struct {
```

That should fix the CI failures.

yeya24: done

Issue #1725, opened 2019-01-25
Panic on access :9093/api/v2/status after upgrade to 0.16.0


vbatychko-modeln:
**What did you do?**
curl :9093/api/v2/status

**What did you expect to see?**
Valid REST reply

**What did you see instead? Under which circumstances?**
curl: (52) Empty reply from server

```
2019/01/25 14:53:31 http: panic serving 127.0.0.1:34314: runtime error: invalid memory address or nil pointer dereference
alertmanager: goroutine 325160 [running]:
alertmanager: net/http.(*conn).serve.func1(0xc00066ea00)
alertmanager: /usr/local/go/src/net/http/server.go:1746 +0xd0
alertmanager: panic(0xc36040, 0x149c340)
alertmanager: /usr/local/go/src/runtime/panic.go:513 +0x1b9
alertmanager: github.com/prometheus/alertmanager/cluster.(*Peer).Name(0x0, 0xc000784a20, 0xc00002f7e0)
alertmanager: /app/cluster/cluster.go:554 +0x22
alertmanager: github.com/prometheus/alertmanager/api/v2.(*API).getStatusHandler(0xc00002f7a0, 0xc00065c000, 0x0, 0x0)
alertmanager: /app/api/v2/api.go:132 +0xb7
alertmanager: github.com/prometheus/alertmanager/api/v2.(*API).getStatusHandler-fm(0xc00065c000, 0xb1cb01, 0xc000960728)
alertmanager: /app/api/v2/api.go:102 +0x34
alertmanager: github.com/prometheus/alertmanager/api/v2/restapi/operations/general.GetStatusHandlerFunc.Handle(0xc0004d0db0, 0xc00065c000, 0xc000163800, 0xecbf40)
alertmanager: /app/api/v2/restapi/operations/general/get_status.go:19 +0x30
alertmanager: github.com/prometheus/alertmanager/api/v2/restapi/operations/general.(*GetStatus).ServeHTTP(0xc0004eed80, 0xed4260, 0xc000c7b7a0, 0xc00065c000)
alertmanager: /app/api/v2/restapi/operations/general/get_status.go:54 +0x164
alertmanager: github.com/go-openapi/runtime/middleware.NewOperationExecutor.func1(0xed4260, 0xc000c7b7a0, 0xc00065c000)
alertmanager: /app/vendor/github.com/go-openapi/runtime/middleware/operation.go:28 +0x75
[...]

PR #1726, opened 2019-01-25
api/v2: Check if cluster peer is nil

mxinden:
When users start Alertmanager with `--cluster.listen-address=`, the
cluster will not be initialized, hence api.peer will be `nil`.

With this patch, api v2 skips populating the peers array in case
`api.peer` is nil.

Fixes https://github.com/prometheus/alertmanager/issues/1725


mxinden:
@yeya24 right, thanks for the hint, sorry for the low quality patch. Added a test and the fix.

api/v2/api.go
```diff
@@ -146,19 +144,27 @@ func (api *API) getStatusHandler(params general_ops.GetStatusParams) middleware.
...
-		},
 	}
 
-	for _, n := range api.peer.Peers() {
-		address := n.Address()
-		resp.Cluster.Peers = append(resp.Cluster.Peers, &open_api_models.PeerStatus{
-			Name:    &n.Name,
-			Address: &address,
-		})
+	// If api.peers == nil, Alertmanager cluster feature is disabled.
```
stuartnelson3: I would reverse this comment. "If alertmanager cluster feature is disabled, then api.peers == nil".

mxinden:
[Looks like](https://github.com/prometheus/alertmanager/issues/1725#issuecomment-457822503) we will need to remove `cluster` in `alertmanagerStatus` from the required fields. I will follow up with that.

api/v2/api.go
```diff
@@ -146,19 +144,27 @@ func (api *API) getStatusHandler(params general_ops.GetStatusParams) middleware.
...
+		peers := []*open_api_models.PeerStatus{}
+		for _, n := range api.peer.Peers() {
+			address := n.Address()
+			peers = append(peers, &open_api_models.PeerStatus{
+				Name:    &n.Name,
+				Address: &address,
+			})
+		}
+
+		resp.Cluster = &open_api_models.ClusterStatus{
```
stuartnelson3: The openapi spec (judging by the generated elm models) don't allow for `resp.Cluster` to be nil. Instead of modifying the spec, it might be more user-friendly to indicate that clustering is disabled.

stuartnelson3: Could we just say
```go
if api.peers != nil {
    // stuff as usual
} else {
    cluster.status = "disabled"
    cluster.peers = []peers{}
    cluster.name = ...
}
```

Not sure what to do about `cluster.name`. Could either send an empty string and just hide the field, or also set it to "disabled".

I like acknowledging that there's no clustering in the ui, but I'm happy to discuss if you want to do what we previously did and hide the cluster status output entirely.

mxinden: Introducing a "disabled" cluster status sounds like a great idea to me.

What are your thoughts in regards to introducing a breaking change to the spec? In case that is fine for a bug fix, I would suggest making `cluster.peers` and `cluster.name` optional fields. If a breaking change is not an option I would be fine with an empty array and an empty ( or "disabled") string.

stuartnelson3: I think we've been pretty clear that the openapi spec can (and probably will) change. I'm fine with a breaking change, we just have to remember to call it out and do a minor version bump :)

mxinden: > remember to call it out

Sure. I will do that in the changelog.

> do a minor version bump

That would be v0.17.0 then. Given that this is a bug fix and that we should get this out rather sooner than later I would vote for v0.16.1.

stuartnelson3: >Given that this is a bug fix and that we should get this out rather sooner than later I would vote for v0.16.1.

If it's a breaking change then it shouldn't be patch level

stuartnelson3:
Looks good to me

simonpasquier:
:+1:

knweiss:
@mxinden I just ran into this issue when I was testing HEAD of the alertmanager master branch. Is there a reason why you didn't merge #1726 into the master branch as well?

mxinden:
Instead of merging this patch into the _master_ branch, we decided to merge a breaking fix into _master_ via https://github.com/prometheus/alertmanager/pull/1728. I have just checked out current _master_, build Alertmanager and ran it with `--cluster.listen-address=` which returns the following json on _api/v2/status_:

```json
  "cluster": {
    "peers": null,
    "status": "disabled"
  },
```

It does not trigger a nil pointer exception in the Go http handler.

@knweiss if this issue still persists for you, would you mind creating a new Github issue? Thanks for the help!

knweiss:
@mxinden Sorry, I'm a little bit in a hurry and only have time for a quick reply: It no longer triggers the nil pointer exception but when I click "Status" in the web UI I get the following error:
```
Unexpected response from api: Problem with the value at json.cluster.peers: null Expecting a LIST
```
It comes from this line:
```
ui/app/src/Utils/Api.elm:            "Unexpected response from api: " ++ err_ 
```
(My  `api/v2/status` is the same as you mentioned above.)

mxinden:
@knweiss we changed `peers` to be optional along with the patch above: 

```elm
type alias ClusterStatus =
    { name : Maybe String
    , status : Status
    , peers : Maybe (List PeerStatus)
}
```
https://github.com/prometheus/alertmanager/blob/f40ecee77326d169dec5092d28510c60a9593f0b/ui/app/src/Data/ClusterStatus.elm#L25

Can you make sure your browser is not using a cached version of the front-end?

knweiss:
@mxinden Branch `release-0.16` (tag `v0.16.1` ) worked fine but master branch (commit f40ecee77 as of today) did not. I still got the error line and could reproduce the problem by switching between both executables. 

However, reloading without Firefox cache did the trick. Now the master branch executable renders the Status page, too! Thanks!

(I could be wrong but I also think I'm seeing the red "disabled" label for the first time in the v0.16.1 version. I'll keep an eye on caching effects from now on.)

mxinden:
@knweiss I think we should disable caching of our web assets to prevent confusions like this. I prepared https://github.com/prometheus/alertmanager/pull/1817. Let me know what you think.

Issue #1723, opened 2019-01-24
Limit number of concurrent GET API requests


beorn7:
Expensive API calls could easily take out Alertmanager. 

Scenario: Everything is on fire, millions of alerts, and hundreds of engineers in the company want to look at the AM UI. The default landing page returns all the alerts (which is kind of an issue on its own), leading to an extreme load to server the UI.

The most important task of AM is to deliver alerts. Thus, it makes sense to limit the number of concurrently processed API request. It should have a reasonable default, and it should be configurable. We have similar things in place for the Prometheus server. Studying those can server as a source of inspiration. (Although I doubt we want to queue API requests (like we queue PromQL queries). Instead, just return a 503 with a meaningful errar message.)

This becomes even more pressing if we update the silenced/inhibited state upon API calls as suggested in #979 .

beorn7:
Yes, this is meant to be the inquiry part, and not the "production path". I made this explicit in the title.

jdelvecchio:
>  It should have a reasonable default, and it should be configurable
What's it default and how is it configurable ? I don't see any reference to this new parameter on the documentation here https://prometheus.io/docs/alerting/configuration/

Currently experiencing issues with the limit and would like to be able to configure it

beorn7:
It's a command line flag, try `alertmanager --help`.

It makes more sense to ask questions like this on the [prometheus-users mailing list](https://groups.google.com/forum/#!forum/prometheus-users) rather than in a closed(!) GitHub issue. On the mailing list, more people are available to potentially respond to your question, and the whole community can benefit from the answers provided.

PR #1733, opened 2019-02-01
Stn/correctly mark api silences

### LINKED ISSUE #979: Querying alert state (silence/inhibition) lags behind (by up to `group_interval`)
If a silence is created, or _if alerts come in that belong to an already existing (silenced) alert group_, it takes some time until the affected alerts are actually displayed as silenced in the UI, presumably because alerts are only marked as silenced if they _would_ trigger a notification but then that notification is not sent because of a silence. As a result, alerts are silenced correctly as in “a silenced alerts never creates a spurious notification” but alerts displayed as firing in the UI might very well be silenced.

This is a known issue for quite some time. IIRC the rationale was that marking alerts as silenced outside of the usual notification cycle is (in theory) computationally too expensive. However, we now have a very nice preview feature (while creating a silence, you can display affected alerts), and if you display a silence, you will even see the affected alerts. This makes me think we should be able to implement something that marks alerts as silenced sooner, at least on a best-effort basis. It's kind of unnerving right now: I see a firing alert. I think I have silenced that a couple of hours ago. I search for the silence to check if I did something wrong. The silence displays me all the alerts silenced by this silence. I see the alert in question. But it is displayed as not-silenced on the alerts tab.

The issue is particularly relevant if there is a long `group_interval`, which we regularly use for sub-critical severity levels. Essentially, if there is a silenced alert already, displayed as such, and then another alert comes in that belongs to the same group and is also silenced, it takes up to the length of the `group_interval` (in our case 12h!) until the alert is actually displayed as silenced.

stuartnelson3:
fixes #979 

to note: 
- @beorn7 is working on https://github.com/prometheus/alertmanager/issues/1723 to prevent multiple users requesting the same information in parallel
- open to ideas on how to streamline the test writing
- i'm interested in adding a "staleness" time on the alerts so that a status doesn't have to be updated every time, but only if the alert hasn't been checked in e.g. 15sec, but this could be added in a separate PR


beorn7:
On my review queue.

Note about the plans to make this more efficient: My idea would be to not just add some kind of freshness (as this would still allow out-of-date results, i.e. a naive user will still be confused if they create a silence, immediately after that reload the alerts page, and still see the silenced alerts as not-silenced). My idea is around establishing a versioning of the silence state,  to be increased each time a new silence is created or an old silence is expired. Then record for each alert the highest version against which its silenced status was checked. Then, we only need to re-check if the current silence state version is higher than the recorded version for the alert. This will also help during the normal sending of notifications (where we currently have to check against all silences each time a notification is due for sending). A similar technique can be applied for inhibitions.

api/v2/api.go
```diff
@@ -53,6 +53,8 @@ type API struct {
 	peer           *cluster.Peer
 	silences       *silence.Silences
 	alerts         provider.Alerts
+	muter          types.Muter
```
mxinden: If `muter` is only about inhibition, let's also call it that way. In addition, this could also just be a function. What do you think?
```suggestion
	isInhibited          func(model.LabelSet) bool
```

api/v2/api.go
```diff
@@ -238,6 +249,28 @@ func (api *API) getAlertsHandler(params alert_ops.GetAlertsParams) middleware.Re
...
+		// be refreshed?
+		// Will check if the alert is inhibited, and update its status.
+		api.muter.Mutes(a.Labels)
+		// Will check if the alert is silenced, and update its status.
+		sils, err := api.silences.Query(
+			silence.QState(types.SilenceStateActive),
+			silence.QMatches(a.Labels),
+		)
+		if err != nil {
+			level.Error(api.logger).Log("msg", "Querying silences failed", "err", err)
```
mxinden:
```suggestion
			level.Error(api.logger).Log("msg", "querying silences failed", "err", err)
```

api/v2/api.go
```diff
@@ -238,6 +249,28 @@ func (api *API) getAlertsHandler(params alert_ops.GetAlertsParams) middleware.Re
 			continue
 		}
 
+		// TODO: Add a timestamp now()+X, after which the Mutes() will
+		// be refreshed?
+		// Will check if the alert is inhibited, and update its status.
+		api.muter.Mutes(a.Labels)
+		// Will check if the alert is silenced, and update its status.
+		sils, err := api.silences.Query(
```
mxinden: If I understand correctly, this logic determines whether a given alert is silenced or not. If so, that domain logic should rather live in `github.com/prometheus/alertmanager/silence`. What do you think?

api/v2/api.go
beorn7: There is already `type MuteFunc`. A function `f` of the suggested signature can be converted into a `Muter` by `MuteFunc(f)`. I'd say accepting the interface here (which could be satisfied by the function you suggest) is preferred.

But I'm with you WRT naming the member variable with something that suggests this is only about inhibitions.

api/v2/api.go
beorn7: Yup. this is now mostly (but not exactly) replicated from https://github.com/prometheus/alertmanager/blob/master/notify/notify.go#L407-L425
Perhaps this could be unified in some way?

api/v2/api.go
```diff
@@ -238,6 +249,28 @@ func (api *API) getAlertsHandler(params alert_ops.GetAlertsParams) middleware.Re
 			continue
 		}
 
+		// TODO: Add a timestamp now()+X, after which the Mutes() will
+		// be refreshed?
```
beorn7: I think we should use a different strategy, see my earlier comment. But let's not worry about the performance right now.

api/v2/api.go
stuartnelson3: The `Muter` interface's description of its method, and its single implementater `Inhibitor`, is a bit confusing. Our implementation relies on a side effect of the particular impl from `Inhibitor`, and isn't actually checking if the alert is inhibited -- instead, if it is, internal to the method, the alert's status is updated.

This is a long way of saying I'm thinking of naming the struct member something along the lines of `setInhibitedStatus`, and separately (an issue or something) discuss how we should either drop or update this interface.

Because we're relying on a particular implementation detail, I think the "correct" thing to do is NOT use the interface, but the actual concrete implementation.

Comments?

api/v2/api.go
stuartnelson3: The `silences` package exposes this `Query` api so as to be externally consumed. It seems to me that this is being used correctly. If we want to change/simplify the `silence` package's `Query` api, such as adding a `QueryActiveAlerts` method or something, maybe that should be a separate PR?

api/v2/api.go
```diff
@@ -72,9 +74,17 @@ type API struct {
 type getAlertStatusFn func(prometheus_model.Fingerprint) types.AlertStatus
 
 // NewAPI returns a new Alertmanager API v2
-func NewAPI(alerts provider.Alerts, sf getAlertStatusFn, silences *silence.Silences, peer *cluster.Peer, l log.Logger) (*API, error) {
+func NewAPI(
+	alerts provider.Alerts,
+	marker types.Marker,
+	sf getAlertStatusFn,
```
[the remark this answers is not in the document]
stuartnelson3: marker is no longer being passed in. check the new impl, it seems easier to understand to me that we pass everything in as a single `setAlertStatus` fn, instead of passing in one part (the `marker`) at creation time, and then passing in the updated-at-reload part separately in `Update`.

api/v2/api.go
beorn7: Quite a bit has changed now. It looks good to me as it is now, but I leave the final call to @mxinden as he raised this point in the first place.

api/v2/api.go
beorn7: I'm fine with solving this later, but perhaps add a TODO then in the `setAlertStatus` function to somehow dedupe the code with https://github.com/prometheus/alertmanager/blob/master/notify/notify.go#L407-L425 eventually?

@mxinden is currently working on un-munging the many pieces of logic munged together in `main.go`, and this puts this quite significant (and partially duplicated) piece of logic somewhere in `main.go`. I can believe that doing the un-munging in a separate and coordinated effort might even make more sense.

beorn7:
Thanks, @stuartnelson3 , for doing this in your one day in between vacations.

@mxinden what do you think? I'd go for merging this (more or less) as is, and then get it straightened in the planned refactoring.

beorn7:
I merged master into this to now check if things work as intended with the recent changes to master.

beorn7:
I added a commit for comment nits, but also found an issue. Will comment in the code.

cmd/alertmanager/main.go
```diff
@@ -476,3 +478,26 @@ func md5HashAsMetricValue(data []byte) float64 {
...
+		inhibitor.Mutes(labels)
+		// TODO(beorn7): The following code is almost exactly replicated in notify/notify.go.
+		sils, err := silences.Query(
+			silence.QState(types.SilenceStateActive),
+			silence.QMatches(labels),
+		)
+		if err != nil {
+			return fmt.Errorf("failed to query silences: %v", err)
+		}
+
```
beorn7: Similar to the code in `notify.go`, don't we have to unset the silence marker of no silences match the alert? I.e.
`n.marker.SetSilenced(labels.Fingerprint())`. (And if that's true, it would prove the point that we should not replicate the code but have it shared somewhere.)

stuartnelson3: This is true, good catch. Updating the code.

beorn7: Thanks. (Would prefer to actually de-replicate the code, but we can do that later.)

cmd/alertmanager/main.go
```diff
@@ -476,3 +478,28 @@ func md5HashAsMetricValue(data []byte) float64 {
 	copy(bytes, smallSum)
 	return float64(binary.LittleEndian.Uint64(bytes))
 }
+
+func setAlertStatus(inhibitor *inhibit.Inhibitor, marker types.Marker, silences *silence.Silences) func(model.LabelSet) error {
```
mxinden: Does this need to be inside `main.go`? I feel like it would be better placed close to `types.Marker`, given that `marker.Status` represents the `getAlertStatus` call. What do you think? Might be included in the TODO below.

> ` 		// TODO(beorn7): The following code is almost exactly replicated in notify/notify.go.`

stuartnelson3: The main reason I put this here is because it seemed like a bundling of imperative things specific to this use case, and then passing them on down to the api. I sort of view `marker.Status` being `getAlertStatus` as a coincidence, and that we created the `getAlertStatus` type as a way to abstract away which specific impl we're using.

Since this particular function is tying together three different packages (`silences`, `inhibit`, and `types`), it felt more natural to me to happen in main. My personal (and debatable, of course) philosophy is that it's easier to define interfaces (easier to unit test), and then assemble all of them together at the top level (for a full acceptance test).

This was original inspired by a [destroy all software screencast](https://www.destroyallsoftware.com/screencasts/catalog/functional-core-imperative-shell), but I've probably diverged from the author's original intent. Happy to talk more, but maybe not in a PR :)

The short answer is: I'm happy to move it if you feel strongly about it. Binding several different packages together didn't leave me with a clear signal where to put it, so `main` felt fine for now. When we figure out how to deduplicate this code this should be easy to remove/replace.

stuartnelson3:
>Thanks for investing into the thorough end-to-end test!

Definitely! I'm hoping that we can start adding more now that there's sort of a "model" for how it works, and turn some of the verbose testing into helper methods as we see more patterns. The actual testing is fairly simple, it's just the struct creation + api call generates a lot of visual noise.

PR #1736, opened 2019-02-04
api: Combine v1 and v2 into generic api

mxinden:
Instead of cmd/alertmanager/main.go instantiating and starting both api
v1 and v2, delegate that work to a generic api combining the two.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -1,0 +1,96 @@
+// Copyright 2019 Prometheus Team
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


beorn7:
Hmmm, this looks like a bit of less clear layering that we pass in the router to the new constructor and get the mux back. (It's also cutting right through my WIP code change for #1723. Which should of course not keep you from doing it...) Just FTR. I'll try to address in my upcoming PR...

beorn7:
I guess the source of the problem is that we are mixing the router from the common package (for v1 and all non-API endpoints) with directly registering an http.Handler with the mux (for v2). Not sure if there is a good way to solve it.

PR #1739, opened 2019-02-05
Allow sending of unauthenticated SMTP requests when smtp_auth_username is not supplied

### LINKED ISSUE #1721: Unable to send email notification without SMTP authentication
Alertmanager won't send an email notification if no SMTP authentication options are configured. These messages appear repeatedly in the log file:

```
level=error ts=2019-01-23T15:49:04.898510015Z caller=notify.go:332 component=dispatcher msg="Error on notify" err="missing password for PLAIN auth mechanism; missing password for LOGIN auth mechanism"
level=error ts=2019-01-23T15:49:04.898681067Z caller=dispatch.go:177 component=dispatcher msg="Notify for alerts failed" num_alerts=1 err="missing password for PLAIN auth mechanism; missing password for LOGIN auth mechanism"
```

Config:

```yaml
global:
  smtp_from: <my_email>
route:
  receiver: root-route-receiver
  group_by:
  - alertname
  group_wait: 2m
  group_interval: 1m
  repeat_interval: 24h
receivers:
- name: root-route-receiver
  email_configs:
  - to: <my_email>
    smarthost: <my_smtp_host>:25
    require_tls: false
```

Noticed this issue after updating to v0.16.0. My same configuration worked on 0.15.1.

metazool:
This PR is intended to replace #1713 which I got into a mess trying to retrospectively sign commits using git rebase.

It adds a noAuth SMTP authentication mechanism which is returned when no `smtp_auth_username` is supplied, allowing this option where a server advertises AUTH but does not require it

CC @simonpasquier - sorry for the extra effort here. 

Closes #1721

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -225,7 +225,7 @@ func (n *Email) auth(mechs string) (smtp.Auth, error) {
 	// If no username is set, keep going without authentication.
 	if n.conf.AuthUsername == "" {
 		level.Debug(n.logger).Log("msg", "smtp_auth_username is not configured. Attempting to send email without authenticating")
-		return NoAuth(), nil
+		return nil, nil
 	}
 
 	err := &types.MultiError{}
@@ -1570,21 +1570,6 @@ func (a *loginAuth) Next(fromServer []byte, more bool) ([]byte, error) {
 			return nil, errors.New("unexpected server challenge")
 		}
 	}
-	return nil, nil
-}
-
-type noAuth struct{}
-
-func NoAuth() smtp.Auth {
-	return &noAuth{}
-}
-
-func (a *noAuth) Start(server *smtp.ServerInfo) (string, []byte, error) {
-	// return proto == "" to indicate that the authentication should be skipped.
-	return "", []byte{}, nil
-}
-
-func (a *noAuth) Next(fromServer []byte, more bool) ([]byte, error) {
 	return nil, nil
 }
 
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -221,6 +221,13 @@ func NewEmail(c *config.EmailConfig, t *template.Template, l log.Logger) *Email 
 // auth resolves a string of authentication mechanisms.
 func (n *Email) auth(mechs string) (smtp.Auth, error) {
 	username := n.conf.AuthUsername
+
+	// If no username is set, keep going without authentication
+	if n.conf.AuthUsername == "" {
+        level.Debug(n.logger).Log("msg", "smtp_auth_username is not configured. Attempting to send email without authenticating")
+		return NoAuth(), nil
+    }
+
 	err := &types.MultiError{}
 	for _, mech := range strings.Split(mechs, " ") {
 		switch mech {
@@ -326,6 +333,8 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 			return true, fmt.Errorf("starttls failed: %s", err)
 		}
 	}
+
+	// If SMTP authentication is configured, check it
 
 	if ok, mech := c.Extension("AUTH"); ok {
 		auth, err := n.auth(mech)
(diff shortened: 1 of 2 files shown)


notify/impl.go
```diff
@@ -1551,6 +1558,21 @@ func (a *loginAuth) Start(server *smtp.ServerInfo) (string, []byte, error) {
 	return "LOGIN", []byte{}, nil
 }
 
+type noAuth struct{}
```
simonpasquier: Please move the `noAuth` code either before or after the `loginAuth` code.

simonpasquier:
:+1: for me. @mxinden can you have a look?

thiagoarrais:
First of all, thank you very much for your contribution, @metazool. Sorry for the unsolicited interruption, but I believe there is a bug here.

I've built this branch and my version generates a dangling AUTH command. Like so:

```smtp
250-AUTH LOGIN
250 Ok
AUTH
501 Syntax: AUTH mechanism [initial-response]
```

I'm not sufficiently fluent in Go to be more helpful, but I've been able to eliminate the dangling AUTH by commenting out the call to `c.Auth`. Maybe this should also be parametrized by the auth mechanism?

simonpasquier:
Ha, it seems that `net/smtp` doesn't do what says in the [documentation](https://golang.org/pkg/net/smtp/#Auth):

> It can return proto == "" to indicate that the authentication should be skipped.

Looking at the [code](https://golang.org/src/net/smtp/smtp.go?s=5600:5635#L189), it doesn't really check if proto is empty and sends the `AUTH` command altogether.

metazool:
Nice catch @thiagoarrais - i never rebuilt my container image based on this code, i was blithely assuming that any failures would get picked up by end-to-end tests in CI but tests would not have been looking for this case.

The fix for the issue that I have applied to our production instance of alertmanager at BGS is just the following https://github.com/metazool/alertmanager/commit/6be5210ece5c8fbb8efd286e3c27e2d3a15b51f9 never calling the auth method at all if the username isn't set. @simonpasquier reckoned that would make other authentication related issues harder to track down. And I wasn't a fan of my own approach there for code style reasons, there's already quite a lot of muggy conditional code which could do with unpacking in `func (n *Email) Notify`... that's a bigger job than i can take on at the moment :/

notify/impl.go
```diff
@@ -221,6 +221,13 @@ func NewEmail(c *config.EmailConfig, t *template.Template, l log.Logger) *Email
 // auth resolves a string of authentication mechanisms.
 func (n *Email) auth(mechs string) (smtp.Auth, error) {
 	username := n.conf.AuthUsername
+
+	// If no username is set, keep going without authentication.
+	if n.conf.AuthUsername == "" {
+		level.Debug(n.logger).Log("msg", "smtp_auth_username is not configured. Attempting to send email without authenticating")
+		return NoAuth(), nil
```
simonpasquier: Returning `nil, nil` should be fine then.

simonpasquier:
@metazool I think that the changes required aren't too drastic. That being said, I'd understand if you don't have time to follow up, just say and I'll carry on.

simonpasquier:
@metazool friendly ping. If you don't have the bandwidth, I can finalize your PR.

simonpasquier:
I've taken the liberty to add a commit to your pull request addressing my last comment. Hope you won't mind!

thiagoarrais:
This fixes my issue, @simonpasquier .

stuartnelson3:
@simonpasquier this is good to go whenever you think it's ready

simonpasquier:
Since @thiagoarrais reported that it is working for him, I'm going to merge it. Many thanks to @metazool for doing most of the work!

AdamVonHorn:
When will this be in a release? I found I cannot update alertmanager because this is something a few people rely on for notifications.

simonpasquier:
@AdamVonHorn there's no official schedule (yet) for AlertManager releases but IIUC 0.17.0 shouldn't be too long.

HeyItsFrank:
@simonpasquier oh I see you're worried about the stigma?

isantospardo:
@simonpasquier would it be possible to get it released in 0.17?

isantospardo:
Thanks @simonpasquier

PR #1743, opened 2019-02-06
Introduce concurrency limit for GET requests and a general timeout for HTTP

### LINKED ISSUE #1723: Limit number of concurrent GET API requests

beorn7:
Fixes #1723 .

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -14,7 +14,9 @@
 package api
 
 import (
+	"fmt"
 	"net/http"
+	"runtime"
 	"time"
 
 	apiv1 "github.com/prometheus/alertmanager/api/v1"
@@ -70,18 +72,39 @@ func New(
 	}, nil
 }
 
-// Register all APIs with the given router and return a mux.
-func (api *API) Register(r *route.Router, routePrefix string) *http.ServeMux {
+// Register all APIs. It registers APIv1 with the provided router directly. As
+// APIv2 works on the http.Handler level, this method also creates a new
+// http.ServeMux and then uses it to register both the provided router (to
+// handle "/") and APIv2 (to handle "<routePrefix>/api/v2"). The method returns
+// the newly created http.ServeMux.
+//
+// If the provided value for timeout is positive, it is enforced for all HTTP
+// requests. (Negative or zero results in no timeout at all.)
+//
+// If the provided value for concurrency is positive, it limits the number of
+// concurrently processed GET requests. Otherwise, the number of concurrently
+// processed GET requests is limited to GOMAXPROCS or 8, whatever is
+// larger. Status code 503 is served for GET requests that would exceed the
+// concurrency limit.
+func (api *API) Register(
+	r *route.Router, routePrefix string,
+	timeout time.Duration, concurrency int,
+) *http.ServeMux {
+	limiter := makeLimiter(timeout, concurrency)
+
 	api.v1.Register(r.WithPrefix("/api/v1"))
 
 	mux := http.NewServeMux()
-	mux.Handle("/", r)
+	mux.Handle("/", limiter(r))
 
 	apiPrefix := ""
 	if routePrefix != "/" {
 		apiPrefix = routePrefix
 	}
-	mux.Handle(apiPrefix+"/api/v2/", http.StripPrefix(apiPrefix+"/api/v2", api.v2.Handler))
+	mux.Handle(
+		apiPrefix+"/api/v2/",
+		limiter(http.StripPrefix(apiPrefix+"/api/v2", api.v2.Handler)),
+	)
 
 	return mux
 }
@@ -93,4 +116,47 @@ func (api *API) Update(cfg *config.Config, resolveTimeout time.Duration) error {
 	}
 
 	return api.v2.Update(cfg, resolveTimeout)
+}
+
+// makeLimiter returns an HTTP middleware that sets a timeout for HTTP requests
+// and also limits the number of concurrently processed GET requests to the
+// given number.
+//
+// If timeout is < 1, no timeout is enforced.
+//
+// If concurrency is < 1, GOMAXPROCS is used as the concurrency limit but at least 8.
+//
+// The returned middleware serves http.StatusServiceUnavailable (503) for requests that
+// would exceed the number.
+func makeLimiter(timeout time.Duration, concurrency int) func(http.Handler) http.Handler {
+	if concurrency < 1 {
+		concurrency = runtime.GOMAXPROCS(0)
+		if concurrency < 8 {
+			concurrency = 8
+		}
+	}
+	inFlightSem := make(chan struct{}, concurrency)
+
+	return func(h http.Handler) http.Handler {
+		concLimiter := http.HandlerFunc(func(rsp http.ResponseWriter, req *http.Request) {
+			if req.Method == http.MethodGet { // Only limit concurrency of GETs.
+				select {
+				case inFlightSem <- struct{}{}: // All good, carry on.
+					defer func() { <-inFlightSem }()
+				default:
+					http.Error(rsp, fmt.Sprintf(
+						"Limit of concurrent GET requests reached (%d), try again later.\n", concurrency,
+					), http.StatusServiceUnavailable)
+					return
+				}
+			}
+			h.ServeHTTP(rsp, req)
+		})
+		if timeout <= 0 {
+			return concLimiter
+		}
+		return http.TimeoutHandler(concLimiter, timeout, fmt.Sprintf(
+			"Exceeded configured timeout of %v.\n", timeout,
+		))
+	}
 }
--- a/api/v1/api.go
+++ b/api/v1/api.go
@@ -253,6 +253,7 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 		// are no alerts present
 		res      = []*Alert{}
 		matchers = []*labels.Matcher{}
+		ctx      = r.Context()
 
 		showActive, showInhibited     bool
 		showSilenced, showUnprocessed bool
@@ -326,9 +327,11 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 	defer alerts.Close()
 
 	api.mtx.RLock()
-	// TODO(fabxc): enforce a sensible timeout.
 	for a := range alerts.Next() {
 		if err = alerts.Err(); err != nil {
+			break
+		}
+		if err = ctx.Err(); err != nil {
 			break
 		}
 
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -146,9 +146,11 @@ func run() int {
 		retention       = kingpin.Flag("data.retention", "How long to keep data for.").Default("120h").Duration()
 		alertGCInterval = kingpin.Flag("alerts.gc-interval", "Interval between alert GC.").Default("30m").Duration()
 
-		externalURL   = kingpin.Flag("web.external-url", "The URL under which Alertmanager is externally reachable (for example, if Alertmanager is served via a reverse proxy). Used for generating relative and absolute links back to Alertmanager itself. If the URL has a path portion, it will be used to prefix all HTTP endpoints served by Alertmanager. If omitted, relevant URL components will be derived automatically.").String()
-		routePrefix   = kingpin.Flag("web.route-prefix", "Prefix for the internal routes of web endpoints. Defaults to path of --web.external-url.").String()
-		listenAddress = kingpin.Flag("web.listen-address", "Address to listen on for the web interface and API.").Default(":9093").String()
+		externalURL    = kingpin.Flag("web.external-url", "The URL under which Alertmanager is externally reachable (for example, if Alertmanager is served via a reverse proxy). Used for generating relative and absolute links back to Alertmanager itself. If the URL has a path portion, it will be used to prefix all HTTP endpoints served by Alertmanager. If omitted, relevant URL components will be derived automatically.").String()
+		routePrefix    = kingpin.Flag("web.route-prefix", "Prefix for the internal routes of web endpoints. Defaults to path of --web.external-url.").String()
+		listenAddress  = kingpin.Flag("web.listen-address", "Address to listen on for the web interface and API.").Default(":9093").String()
+		getConcurrency = kingpin.Flag("web.get-concurrency", "Maximum number of GET requests processed concurrently. If negative or zero, the limit is GOMAXPROC or 8, whatever is larger.").Default("0").Int()
+		getTimeout     = kingpin.Flag("web.timeout", "Timeout for HTTP requests. If negative or zero, no timeout is set.").Default("0").Duration()
 
 		clusterBindAddr = kingpin.Flag("cluster.listen-address", "Listen address for cluster.").
 				Default(defaultClusterAddr).String()
@@ -397,7 +399,7 @@ func run() int {
 
 	ui.Register(router, webReload, logger)
 
-	mux := api.Register(router, *routePrefix)
+	mux := api.Register(router, *routePrefix, *getTimeout, *getConcurrency)
 
 	srv := http.Server{Addr: *listenAddress, Handler: mux}
 	srvc := make(chan struct{})
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -195,6 +195,7 @@ func (api *API) getAlertsHandler(params alert_ops.GetAlertsParams) middleware.Re
 		// are no alerts present
 		res      = open_api_models.GettableAlerts{}
 		matchers = []*labels.Matcher{}
+		ctx      = params.HTTPRequest.Context()
 	)
 
 	if params.Filter != nil {
@@ -224,9 +225,11 @@ func (api *API) getAlertsHandler(params alert_ops.GetAlertsParams) middleware.Re
 	defer alerts.Close()
 
 	api.mtx.RLock()
-	// TODO(fabxc): enforce a sensible timeout.
 	for a := range alerts.Next() {
 		if err = alerts.Err(); err != nil {
+			break
+		}
+		if err = ctx.Err(); err != nil {
 			break
 		}
 


beorn7:
@mxinden I'm banging my head against a wall when investigating the Travis errors: `cannot find package "github.com/prometheus/alertmanager/ui/app"` Hmm, yes, that package does not exist. But why does it even want it? It's not referred anywhere I can see. Also, it works locally.

beorn7:
I have a lead: the `go generate` command does not use Go Modules in my local run because apparently the env variables don't make it through to it. However, in Travis, presumably GOPATH is not set, so Go Modules are activated automatically.

beorn7:
That's definitely the reason. Now trying to find a solution. (BTW: This would fail in master in the same way.)

beorn7:
Still banging my head against a wall. The way to reproduce: set GOPATH to something else, and then run `make assets`. I tried to make the `assets` package a go module on its own, but then we get double-imports. @simonpasquier as our local go module expert: Do you know a way out of this?

api/api.go
```diff
@@ -70,18 +72,39 @@ func New(
...
 
-// Register all APIs with the given router and return a mux.
-func (api *API) Register(r *route.Router, routePrefix string) *http.ServeMux {
+// Register all APIs. It registers APIv1 with the provided router directly. As
+// APIv2 works on the http.Handler level, this method also creates a new
+// http.ServeMux and then uses it to register both the provided router (to
+// handle "/") and APIv2 (to handle "<routePrefix>/api/v2"). The method returns
+// the newly created http.ServeMux.
+//
+// If the provided value for timeout is positive, it is enforced for all HTTP
```
mxinden: Instead of using a value smaller equal-zero as the null value, I would prefer using a pointer.  What do you think?

api/api.go
```diff
@@ -70,18 +72,39 @@ func New(
...
+// requests. (Negative or zero results in no timeout at all.)
+//
+// If the provided value for concurrency is positive, it limits the number of
+// concurrently processed GET requests. Otherwise, the number of concurrently
```
mxinden: See _null_ comment above.

api/api.go
```diff
@@ -94,3 +117,46 @@ func (api *API) Update(cfg *config.Config, resolveTimeout time.Duration) error {
...
+	inFlightSem := make(chan struct{}, concurrency)
+
+	return func(h http.Handler) http.Handler {
+		concLimiter := http.HandlerFunc(func(rsp http.ResponseWriter, req *http.Request) {
+			if req.Method == http.MethodGet { // Only limit concurrency of GETs.
+				select {
+				case inFlightSem <- struct{}{}: // All good, carry on.
+					defer func() { <-inFlightSem }()
+				default:
+					http.Error(rsp, fmt.Sprintf(
```
mxinden: I guess logging this would defeat the purpose of a rate limiter trying to prevent overload, but how about exposing a counter for this?

beorn7:
Hmm, I just pushed a whitespace change to a branch freshly branched off master, and there the `go generate` works, but then errcheck finds a lot of problems all of a sudden.

mxinden:
Interesting. Ignoring `err` on `Log` should be excluded. Will look further into that.

Travis build link for reference: https://travis-ci.org/prometheus/alertmanager/builds/489679923

api/api.go
beorn7: I have just propagated the flag value (where, I think, we have no choice of giving non-positive values that meaning). With a pointer, we needed a separate translation from non-positive flag values to a nil pointer. And then let this function panic for non-positive values.

In different news: Isn't considering a non-positive timeout value as "no timeout" relatively common / intuitive?

beorn7:
I think a bunch of things in our build are pulled in with current versions. That might explain changing behavior (and is at odds with the go-modules approach of reproducible builds).

beorn7:
I'll get to the bottom of this tomorrow with a clear head… Very mysterious. :thinking:

api/api.go
beorn7: Oh sure. There is even a middleware in client_golang just for that.

beorn7: I have added a "requests in flight" gauge.

In different news: The api/v2 calls are currently not instrumented at all because they act on the Handler level while the current instrumentation acts on the Router level only. I guess that's an oversight. Fixing it is a bit more involved. I suggest to do it after the refactoring you have already started. I left a TODO note.

mxinden: Thanks for adding the middleware. Yes, API v2 is not yet instrumented. On my agenda.

beorn7: Slight update: For some reason (which I cannot remember), Stuart and I decided to have the in-flight middleware not partitioned by method. Since the limit only applies to GET requests, I have now only instrumented the GET requests, not even using the middleware (which in turn now includes the api/v2 calls into this particular instrumentation).

I'm not sure if we should instrument the in-flight requests in general, including method and handler labels. We can still estimate the overlapping request load from the latency and rate. Thoughts welcome. If it is in general useful to instrument in-flight requests partitioned by method, we should change the middleware in client_golang.

simonpasquier:
Regarding the `errcheck` failures, I suspect that Travis CI messes up with `GOPATH`. I'm trying https://github.com/prometheus/alertmanager/pull/1748to investigate/fix it.

beorn7:
Rebased on top of master with @simonpasquier 's and my fixes. Let's see what's happening now.

api/api.go
beorn7: Whatever we decide above, we should do the same here, obviously…

api/api.go
beorn7: I _think_ the reason for not partitioning the `InstrumentHandlerInFlight` middleware is that you cannot partition by status code as that one is only determined once the request is not in flight anymore. We could, of course, still partition by method. However, the middleware is already in popular use. If we really want to break all those users (it's an easy code change to adjust to a GaugeVec without labels, but still), it should be worth it.

I'd say for now, another v0.10 thing. Let's go with the GET only instrumentation here for now... (not actually using the middleware, in fact).

api/api.go
```diff
@@ -24,12 +26,24 @@ import (
...
 
+var (
+	requestsInFlight = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "alertmanager_http_get_requests_in_flight",
+		Help: "Current number of HTTP GET requests being processed.",
+	})
+)
+
+func init() {
+	prometheus.MustRegister(requestsInFlight)
```
mxinden: I am not a big fan of the global metric registry. It might be less verbose but also implicit and more difficult when it comes to vendoring. I am trying the steer this project away from it. What do you think in this regard? Would passing a `Registerer` to `API.New()` also be an option?

api/api.go
mxinden: I guess there is a lot of cleanup in terms of our instrumentation, which we should probably do in a follow up patch.

Never the less we could also do "alertmanager_http_requests_in_flight { method="GET" }`, what do you think?

api/api.go
```diff
@@ -94,3 +133,50 @@ func (api *API) Update(cfg *config.Config, resolveTimeout time.Duration) error {
...
+					return
+				}
+			}
+			h.ServeHTTP(rsp, req)
+		})
+		if timeout <= 0 {
+			return concLimiter
+		}
+		return http.TimeoutHandler(concLimiter, timeout, fmt.Sprintf(
+			"Exceeded configured timeout of %v.\n", timeout,
```
mxinden:
```suggestion
			"exceeded configured timeout of %v.\n", timeout,
```

api/v1/api.go
```diff
@@ -326,11 +327,13 @@ func (api *API) listAlerts(w http.ResponseWriter, r *http.Request) {
 	defer alerts.Close()
 
 	api.mtx.RLock()
-	// TODO(fabxc): enforce a sensible timeout.
```
mxinden: :tada:

api/api.go
mxinden: Given that we will probably have more optional configurations to come, how about adopting the `Option` pattern here, which we already use in `silence.go` and `nflog.go`. That enables us to do the validation at API creation time, giving users early feedback of invalid inputs at startup time:

```diff
diff --git a/api/api.go b/api/api.go
index 8241f730..8ee1fd01 100644
--- a/api/api.go
+++ b/api/api.go
@@ -48,6 +48,36 @@ func init() {
 type API struct {
 	v1 *apiv1.API
 	v2 *apiv2.API
+
+	// Timeout for HTTP GET requests.
+	getTimeout *time.Duration
+	// Maximum number of GET requests processed concurrently.
+	getConcurrency *int
+}
+
+// Option configures a new API implementation.
+type Option func(*API) error
+
+// WithGetTimeout sets the timeout for HTTP GET requests.
+func WithGetTimeout(t *time.Duration) Option {
+	return func(api *API) error {
+		api.getTimeout = t
+		return nil
+	}
+}
+
+// WithGetConcurrency sets the get concurrency for HTTP GET requests.
+func WithGetConcurrency(c *int) Option {
+	return func(api *API) error {
+		if c != nil && *c < 1 {
+			return fmt.Errorf(
+				"expected GetConcurrency API parameter to be greater or equal to one but got: '%v'",
+				c,
+			)
+		}
+		api.getConcurrency = c
+		return nil
+	}
 }
 
 // New creates a new API object combining all API versions.
@@ -57,6 +87,7 @@ func New(
 	sf func(model.Fingerprint) types.AlertStatus,
 	peer *cluster.Peer,
 	l log.Logger,
+	opts ...Option,
 ) (*API, error) {
 	v1 := apiv1.New(
 		alerts,
@@ -78,10 +109,18 @@ func New(
 		return nil, err
 	}
 
-	return &API{
+	api := &API{
 		v1: v1,
 		v2: v2,
-	}, nil
+	}
+
+	for _, o := range opts {
+		if err := o(api); err != nil {
+			return nil, err
+		}
+	}
+
+	return api, nil
 }
 
 // Register all APIs. It registers APIv1 with the provided router directly. As
@@ -100,9 +139,8 @@ func New(
 // concurrency limit.
 func (api *API) Register(
 	r *route.Router, routePrefix string,
-	timeout time.Duration, concurrency int,
 ) *http.ServeMux {
-	limiter := makeLimiter(timeout, concurrency)
+	limiter := makeLimiter(api.getTimeout, api.getConcurrency)
 
 	api.v1.Register(r.WithPrefix("/api/v1"))
 
@@ -144,14 +182,14 @@ func (api *API) Update(cfg *config.Config, resolveTimeout time.Duration) error {
 //
 // The returned middleware serves http.StatusServiceUnavailable (503) for requests that
 // would exceed the number.
-func makeLimiter(timeout time.Duration, concurrency int) func(http.Handler) http.Handler {
-	if concurrency < 1 {
-		concurrency = runtime.GOMAXPROCS(0)
-		if concurrency < 8 {
-			concurrency = 8
+func makeLimiter(timeout *time.Duration, concurrency *int) func(http.Handler) http.Handler {
+	if concurrency == nil {
+		*concurrency = runtime.GOMAXPROCS(0)
+		if *concurrency < 8 {
+			*concurrency = 8
 		}
 	}
-	inFlightSem := make(chan struct{}, concurrency)
+	inFlightSem := make(chan struct{}, *concurrency)
 
 	return func(h http.Handler) http.Handler {
 		concLimiter := http.HandlerFunc(func(rsp http.ResponseWriter, req *http.Request) {
@@ -172,10 +210,10 @@ func makeLimiter(timeout time.Duration, concurrency int) func(http.Handler) http
 			}
 			h.ServeHTTP(rsp, req)
 		})
-		if timeout <= 0 {
+		if timeout == nil {
 			return concLimiter
 		}
-		return http.TimeoutHandler(concLimiter, timeout, fmt.Sprintf(
+		return http.TimeoutHandler(concLimiter, *timeout, fmt.Sprintf(
 			"Exceeded configured timeout of %v.\n", timeout,
 		))
 	}
diff --git a/cmd/alertmanager/main.go b/cmd/alertmanager/main.go
index 66a78934..dfffb4cf 100644
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -150,8 +150,8 @@ func run() int {
 		externalURL    = kingpin.Flag("web.external-url", "The URL under which Alertmanager is externally reachable (for example, if Alertmanager is served via a reverse proxy). Used for generating relative and absolute links back to Alertmanager itself. If the URL has a path portion, it will be used to prefix all HTTP endpoints served by Alertmanager. If omitted, relevant URL components will be derived automatically.").String()
 		routePrefix    = kingpin.Flag("web.route-prefix", "Prefix for the internal routes of web endpoints. Defaults to path of --web.external-url.").String()
 		listenAddress  = kingpin.Flag("web.listen-address", "Address to listen on for the web interface and API.").Default(":9093").String()
-		getConcurrency = kingpin.Flag("web.get-concurrency", "Maximum number of GET requests processed concurrently. If negative or zero, the limit is GOMAXPROC or 8, whatever is larger.").Default("0").Int()
-		getTimeout     = kingpin.Flag("web.timeout", "Timeout for HTTP requests. If negative or zero, no timeout is set.").Default("0").Duration()
+		getConcurrency = kingpin.Flag("web.get-concurrency", "Maximum number of GET requests processed concurrently. If negative or zero, the limit is GOMAXPROC or 8, whatever is larger.").Int()
+		getTimeout     = kingpin.Flag("web.timeout", "Timeout for HTTP requests. If negative or zero, no timeout is set.").Duration()
 
 		clusterBindAddr = kingpin.Flag("cluster.listen-address", "Listen address for cluster.").
 				Default(defaultClusterAddr).String()
@@ -300,6 +300,7 @@ func run() int {
 		marker.Status,
 		peer,
 		log.With(logger, "component", "api"),
+		getTimeout, getConcurrency,
 	)
 
 	if err != nil {
@@ -400,7 +401,7 @@ func run() int {
 
 	ui.Register(router, webReload, logger)
 
-	mux := api.Register(router, *routePrefix, *getTimeout, *getConcurrency)
+	mux := api.Register(router, *routePrefix)
 
 	srv := http.Server{Addr: *listenAddress, Handler: mux}
 	srvc := make(chan struct{})
```

mxinden: Given that there seems to be no possibility in kingpin to tell, whether a user specified a flag or not, except comparing it to the default value :flushed:  , I guess this discussion is obsolete anyways and we can keep it as it is right now. I am very happy if you can proof this to be wrong @beorn7.

simonpasquier:
Nothing much to add to @mxinden's comments from my side.

api/api.go
beorn7: Will look into Kingpin behavior.

I am not opposed to the options pattern, with the rationale you stated, that we will probably add more options later. However, my impression is that the "options struct" pattern has won over "options functions". And interestingly, AM uses the former for silences and the latter for nflog. Thus, neither of the two would be more consistent with the existing AM code. I would thus gravitate towards an options struct.

About the pointer vs. value: I'm not sold on the idea that the additional complication is worth it to use `nil` as a sentinel value for "default behavior". To prove my point, look at `*concurrency = runtime.GOMAXPROCS(0)`: This code line will result in a runtime error if `concurrency` is nil, but in your suggested code, you execute that code line precisely under the condition that `concurrency` is nil. Surely just an easy oversight, but it proves that the handling of a pointer is a bit trickier than a plain value. Plus, we need to come up with a meaning for a non-positive timeout and concurrency limits anyway. Alternatively, we could error out for non-positive timeout or concurrency limit, but that's additional validation. Finally, I'd like to have a way to specify the `--web.timeout` flag in a way that says "no timeout" other than _not_ specifying the flag at all. That's useful if you have a template in your configuration management that fills in a certain value for that flag. If you want to set no timeout, it's handy to specify "0" as a value rather than changing the whole template to not provide the flag at all.

In any case, I'll study Kingpin behavior, as said, and then come up with a suggestion.

api/api.go
beorn7: I think the `method` label only makes sense if we have plan to instrument the in-flight gauge for other methods, too. It kind of suggests that by being a label and not part of the metric name.
If you think we should eventually instrument all methods that way, we can certainly go down the label approach.

api/api.go
beorn7: I'm not a fan of the global registry, either. I just wanted to stay consistent with the way it is and then have a separate refactoring to make it local.

But now I have seen what you did in #1741, which is already merged. Thus, a incremental migration to local Registerers is already ongoing. Happy to follow that path here.

api/api.go
beorn7: That's actually not a Go error message (which, by convention, is lower case) but the content of the HTTP response. As such, it is shown in isolation and should follow normal English spelling rules. (The default error message of the TimeoutHandler is also capitalized. It is even using HTML, but I don't think we have to go that far.)

beorn7:
Responded to the comments, but haven't yet pushed the commit to actually address them.
Give me some time...

cmd/alertmanager/main.go
```diff
@@ -146,9 +147,11 @@ func run() int {
...
-		listenAddress = kingpin.Flag("web.listen-address", "Address to listen on for the web interface and API.").Default(":9093").String()
+		externalURL    = kingpin.Flag("web.external-url", "The URL under which Alertmanager is externally reachable (for example, if Alertmanager is served via a reverse proxy). Used for generating relative and absolute links back to Alertmanager itself. If the URL has a path portion, it will be used to prefix all HTTP endpoints served by Alertmanager. If omitted, relevant URL components will be derived automatically.").String()
+		routePrefix    = kingpin.Flag("web.route-prefix", "Prefix for the internal routes of web endpoints. Defaults to path of --web.external-url.").String()
+		listenAddress  = kingpin.Flag("web.listen-address", "Address to listen on for the web interface and API.").Default(":9093").String()
+		getConcurrency = kingpin.Flag("web.get-concurrency", "Maximum number of GET requests processed concurrently. If negative or zero, the limit is GOMAXPROC or 8, whatever is larger.").Default("0").Int()
```
stuartnelson3: my gut says "whichever" instead of "whatever", but not something that's super high on my list.

beorn7: Your gut is a native en-us speaker. I believe what it says.

api/api.go
beorn7: So yeah, kingpin flag values default to the zero value. Even the pointers just point to a variable with the zero value. Thus, I see no gain in introducing `nil` as a marker value for "default", even if I were of the opinion that that would be a good idea in general.

beorn7: I think it's best to give the `api.New` method an options struct as an argument. Working on it...

api/api.go
beorn7: I will add the label (lower-case `get` for now, as this is (sadly) the norm in client_golang, cf. https://github.com/prometheus/client_golang/issues/387 ) plus a TODO to instrument other method's, too.

I'll also add a Registerer as one of the options.

beorn7:
Sorry for the delay. Got sidetracked all the time. Finally pushed that new commit. @stuartnelson3 is gone for vacation again. @mxinden please have another look.

beorn7:
@mxinden Do you think you could have a look at this any time soon?

api/api.go
```diff
@@ -32,56 +36,149 @@ import (
...
-	v2 *apiv2.API
+	v1                       *apiv1.API
+	v2                       *apiv2.API
+	requestsInFlight         prometheus.Gauge
+	concurrencyLimitExceeded prometheus.Counter
+	timeout                  time.Duration
+	inFlightSem              chan struct{}
+}
+
+// Options for the creation of an API object. Alerts, Silences, and StatusFunc
```
mxinden: Thanks for adding detailed comments here.

api/api.go
```diff
@@ -32,56 +36,149 @@ import (
...
 // API represents all APIs of Alertmanager.
 type API struct {
-	v1 *apiv1.API
-	v2 *apiv2.API
+	v1                       *apiv1.API
+	v2                       *apiv2.API
+	requestsInFlight         prometheus.Gauge
+	concurrencyLimitExceeded prometheus.Counter
+	timeout                  time.Duration
+	inFlightSem              chan struct{}
```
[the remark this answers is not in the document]
beorn7: The problem is that we call `limitHandler` twice, once to handle the router handler, once to handle the apiv2 handler. If we moved the `inFlightSem` into `limitHandler`, we would limit the GET requests separately for each one. But we want one limit for all of the HTTP handling. (That was the reason why I had a function that returns a function (the middleware) that returns a function (the HandlerFunc) in the previous version. I could avoid that confusion by having inFlightSem as a member of `API` and limitHandler a method of `API`.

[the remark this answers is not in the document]
beorn7: @mxinden did my comment clarify the issue? Or do you think this should be changed in another way?

beorn7:
Thanks @simonpasquier . @mxinden I assume things are clarified now. We can still rework this if needed. I just would to like to get this merged as #1733 needs to address some merge conflicts with it and I have yet _another_ change on top of all that in my pipeline. Thus, I'm merging this now to get all those ends tied up. It also unblocks your work on #1744.

PR #1744, opened 2019-02-06
*: Introduce config coordinator bundling config specific logic

mxinden:
Instead of handling all config specific logic inside
Alertmangaer.main(), this patch introduces the config coordinator
component.

Tasks of the config coordinator:
- Load and parse configuration
- Notify subscribers on configuration changes
- Register and manage configuration specific metrics

This patch still has a couple of `TODO`s. I am opening a pull request anyways, to get early feedback whether a refactoring like this is wanted.

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -15,8 +15,6 @@ package main
 
 import (
 	"context"
-	"crypto/md5"
-	"encoding/binary"
 	"fmt"
 	"net"
 	"net/http"
@@ -56,18 +54,6 @@ import (
 )
 
 var (
-	configHash = prometheus.NewGauge(prometheus.GaugeOpts{
-		Name: "alertmanager_config_hash",
-		Help: "Hash of the currently loaded alertmanager configuration.",
-	})
-	configSuccess = prometheus.NewGauge(prometheus.GaugeOpts{
-		Name: "alertmanager_config_last_reload_successful",
-		Help: "Whether the last configuration reload attempt was successful.",
-	})
-	configSuccessTime = prometheus.NewGauge(prometheus.GaugeOpts{
-		Name: "alertmanager_config_last_reload_success_timestamp_seconds",
-		Help: "Timestamp of the last successful configuration reload.",
-	})
 	requestDuration = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Name:    "alertmanager_http_request_duration_seconds",
@@ -88,9 +74,6 @@ var (
 )
 
 func init() {
-	prometheus.MustRegister(configSuccess)
-	prometheus.MustRegister(configSuccessTime)
-	prometheus.MustRegister(configHash)
 	prometheus.MustRegister(requestDuration)
 	prometheus.MustRegister(responseSize)
 	prometheus.MustRegister(version.NewCollector("alertmanager"))
@@ -263,14 +246,6 @@ func run() int {
 	}
 	defer alerts.Close()
 
-	var (
-		inhibitor *inhibit.Inhibitor
-		tmpl      *template.Template
-		pipeline  notify.Stage
-		disp      *dispatch.Dispatcher
-	)
-	defer disp.Stop()
-
 	api, err := api.New(api.Options{
 		Alerts:      alerts,
 		Silences:    silences,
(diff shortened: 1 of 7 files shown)


beorn7:
This is acting on precisely the same code @stuartnelson3 and I are currently working on.
Can we perhaps postpone the refactoring until we are done with that?

beorn7:
See #1743 and #1733 .

mxinden:
> This is acting on precisely the same code @stuartnelson3 and I are currently working on.
Can we perhaps postpone the refactoring until we are done with that?

@beorn7 absolutely. The two other patches have precedence.

beorn7:
In case I sounded too negative: I like the refactoring in general. :o)

beorn7:
I think it's now “safe” to continue with this refactoring. Perhaps you should sync with @stuartnelson3 about basic ideas what should be in main and what should be pushed into packages, cf. his [comment](https://github.com/prometheus/alertmanager/pull/1733#discussion_r257962598).

I'll now work on implementing my ideas about making the silence and inhibition checks less expensive. But that should not touch the configuration logic too much.

config/coordinator.go
```diff
@@ -0,0 +1,140 @@
...
+
+	configHashMetric        prometheus.Gauge
+	configSuccessMetric     prometheus.Gauge
+	configSuccessTimeMetric prometheus.Gauge
+
+	subscribers []func(*Config)
+}
+
+// NewCoordinator returns a new coordinator with the given configuration file
+// path.
```
beorn7: Perhaps mention that the config file is not loaded yet.

config/coordinator.go
```diff
@@ -0,0 +1,140 @@
...
+
+	"github.com/prometheus/client_golang/prometheus"
+
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+)
+
+// Coordinator coordinates Alertmanager configurations beyond the lifetime of a
+// single configuration.
+// TODO: Make Coordinator thread safe
```
beorn7: Is that necessary? Isn't it the advantage of the coordinator with the subscription model that you don't need that anymore?

beorn7:
The failed CircleCI check succeeded when I reran it. Flaky test...

config/coordinator.go
```diff
@@ -0,0 +1,140 @@
...
+
+package config
+
+import (
+	"crypto/md5"
+	"encoding/binary"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+
```
simonpasquier: (nit) can be removed.

config/coordinator.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+// loadFromFile triggers a configuration load, discarding the old configuration.
+func (c *Coordinator) loadFromFile() error {
+	level.Info(c.logger).Log(
+		"msg", "Loading configuration file",
+		"file", c.cfgFilePath,
+	)
+
+	conf, plainCFG, err := LoadFile(c.cfgFilePath)
```
simonpasquier: (nit) s/plainCFG/plainCfg/

cmd/alertmanager/main.go
```diff
@@ -304,30 +279,24 @@ func run() int {
...
+	configCoordinator := config.NewCoordinator(
+		*configFile,
+		prometheus.DefaultRegisterer,
+		log.With(logger, "component", "configuration"),
+	)
+	configCoordinator.Subscribe(func(conf *config.Config) {
 		tmpl, err = template.FromGlobs(conf.Templates...)
 		if err != nil {
-			return err
+			level.Error(logger).Log("err", fmt.Errorf("failed to parse templates: %v", err.Error()))
```
simonpasquier: If we enter this branch, the error wouldn't bubble up anymore. Is that intended? Wouldn't it make sense for `Coordinator.Subscribe` to accept `func(*Config) error`?

mxinden: Right, that got me thinking as well. If `Coordinator` accepts `func(*Config) error`, it would need to handle errors returned by other components logic. What should it do with such error? It can't know what kind of errors are possible, hence it doesn't know how to handle them, right? How would you handle the errors in the `Coordinator` code?

config/coordinator.go
mxinden: Say there are two reloads triggered via the API, this would require write access to `config *Config` in a concurrent fassion. That should be synchronized, right?

config/coordinator.go
mxinden: Good point. Added.

config/coordinator.go
mxinden: In general, Go would use `plainCFG` see: https://github.com/golang/go/wiki/CodeReviewComments#initialisms.

I decided to replace all `cfg` with `config` to stay consistent with `config.go`, let me know is you think differently.

cmd/alertmanager/main.go
simonpasquier: In the Prometheus code, we apply all "reloaders" (equivalent to subscribers), log an error message  for any that fails and return a generic error message tot the caller in case one reloader failed.

https://github.com/prometheus/prometheus/blob/e4a741cb7dfa24978097971c2c8d5c8cc12e6cef/cmd/prometheus/main.go#L714-L744

config/coordinator.go
simonpasquier: Hmm, I wouldn't consider `CFG` as an initialism or acronym but I'm not a linguistic expert :-)

config/coordinator.go
beorn7: Yes. "Make Coordinator thread safe" made me think you want to somehow enable concurrent reloads.

I think we either can protect the `Reload` method by a mutex, or we can document that `Reload` must only be called from a single goroutine. (I could imagine the latter is naturally the case, but if not, a mutex might be the most convenient solution.) If that's what you meant anyway, I'd suggest to move this TODO to the `Reload` method, reworded as "protect this method with a mutex" and add to the doc comment (for now) that `Reload` is not concurrency-safe.

cmd/alertmanager/main.go
mxinden: Ok, good point. Did the relevant changes. Instead of returning a generic error message, I am returning and logging a specific one. Let me know if that is ok for you. Should be better for debugging.

config/coordinator.go
mxinden: Given that configuration reloading is not a super time sensitive thing and does not need to be parallelized I added `Mutex` to protect `subscribers` and `config`. Let me know what you think.

config/coordinator.go
```diff
@@ -0,0 +1,159 @@
...
+		c.logger.Log(
+			"msg", "loading configuration file failed",
+			"file", c.configFilePath,
+			"err", err,
+		)
+		return err
+	}
+
+	err = c.notifySubscribers()
+	if err != nil {
```
stuartnelson3: any reason for not having the `if err := foo.Thing(); err != nil` pattern?

mxinden: Oh, right. That looks better. Adjusted above as well.

mxinden:
Given that you both approved I will merge. Let me know if you want me to do any follow ups.

PR #1746, opened 2019-02-07
[WIP] Added negative (re) matching in silences

geekodour:
See #1682 

I am not yet sure if this is a working PR but it does add `!=` and `!~` into the amtool and the alertmanager UI.

Think these methods will require changes too. But not changing does not break anything so far.

https://github.com/prometheus/alertmanager/blob/5d1bd285ca0c5c15d639d744ff8a75abc87b6599/types/match.go#L36-L45

https://github.com/prometheus/alertmanager/blob/5d1bd285ca0c5c15d639d744ff8a75abc87b6599/types/match.go#L71-L81

Todo:
- [ ] Write tests
- [ ] Add dropdown in silence create form

cc @mxinden @juliusv


mnaser:
@geekodour are you currently pursuing this?  this is super useful for us because watchdog alerts will start firing if we silence a specific env :(

geekodour:
@mnaser It's been almost a year since I worked on this issue, I might pick it up if it is still a good idea.

@juliusv please let me know if it'll be a good idea to work on this?

also what about the react implementation. iirc, I was facing some issues with the drop down in elm and it was working as expected with amtool manually but I was unsure whether it was working or not because I did not write any tests at that time.

artburkart:
It would be valuable if `amtool` had this functionality, even if the UI didn't.

PR #1747, opened 2019-02-07
Fix build problems

beorn7:
This will fix the errcheck problems, and it might have an impact on the `go generate` part.

It indeed looks like Circle somehow keeps `GOPATH` in the environment for external commands (via `go generate` or the `errcheck` command), while Travis somehow removes it or builds outside of it or whatever. More investigation needed, but this removes a few parts of the equation at least.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ PRECHECK_OPTIONS_bzr = version
 build-all: assets apiv2 build
 
 assets: ui/app/script.js ui/app/index.html ui/app/lib template/default.tmpl
-	cd $(PREFIX)/asset && $(GO) generate
+	GO111MODULE=$(GO111MODULE) $(GO) generate ./asset
 	@$(GOFMT) -w ./asset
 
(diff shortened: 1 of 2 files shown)


simonpasquier:
Same conclusion for me: one way or another, Travis CI changed something to their setup and it acts as if the project was outside of `GOPATH`. I noticed that it automatically exports `GO111MODULE=auto` too recently. I wish there was a changelog somewhere...

beorn7:
That's a lead... Still I don't quite get why the `go generate` part is affected in that way. If they still build within `GOPATH`, the `GO111MODULE=auto` should re-create the original go behavior. But somehow, it wants modules, but that doesn't seem to work in the sub dir... or something... Anyway, let's get this merged (on green) so that we have one less thing to distract us from the `go generate` problem.

beorn7:
@simonpasquier fine to merge this?

beorn7:
Since you closed #1748 in lieu if this, I guess yes. Just want to get master green again. Merging now...

simonpasquier:
Thanks for this @beorn7! I'll dig further to understand what has changed and broke...

PR #1764, opened 2019-02-22
Modify the self-inhibition prevention semantics

beorn7:
This has been discussed in #666 (issue of hell...).

As concluded there, the cleanest semantics is most likely the
following: "An alert that matches both target and source side cannot
inhibit alerts for which the same is true." The two open questions
were:
1. How difficult is the implementation?
2. Is it needed?

This relatively simple commit proves that the answer to (1) is: Not
very difficult. (This also includes a performance-improving
simplification, which would have been possible without a change of
semantics.)

The answer to (2) is twofold:

For one, the original use case in #666 wasn't solved by our interim
solution. What we solved is the case where the self-inhibition is
triggered by a wide target match, i.e. I have a specific alert that
should inhibit a whole group of target alerts without inhibiting
itself. What we did _not_ solve is the inverted case: Self-inhibition
by a wide source match, i.e. an alert that should only fire if none of
a whole group of source alert fires. I mean, we "fixed" it as in, the
target alert will _never_ be inhibited, but @lmb in #666 wanted the
alert to be inhibited _sometimes_ (just not _always_).

The other part is that I think that the asymmetry in our interim
solution will at some point haunt us. Thus, I really would like to get
this change in before we do a 1.0 release.

In practice, I expect this to be only relevant in very rare cases. But
those cases will be most difficult to reason with, and I claim that
the solution in this commit is matching what humans intuitively
expect.

### DIFF
--- a/inhibit/inhibit_test.go
+++ b/inhibit/inhibit_test.go
@@ -131,7 +131,7 @@ func TestInhibitRuleHasEqual(t *testing.T) {
 			r.scache.Set(v)
 		}
 
-		if _, have := r.hasEqual(c.input); have != c.result {
+		if _, have := r.hasEqual(c.input, false); have != c.result {
 			t.Errorf("Unexpected result %t, expected %t", have, c.result)
 		}
 	}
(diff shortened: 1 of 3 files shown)


beorn7:
Rebased on top of master.

inhibit/inhibit_test.go
```diff
@@ -140,55 +140,87 @@ func TestInhibitRuleHasEqual(t *testing.T) {
...
+			// Matches both source and target filters of rule1,
+			// inhibited because sourceAlert1 matches only the
+			// source filter of rule1.
+			target:   model.LabelSet{"s1": "1", "t1": "1", "e": "1"},
+			expected: true,
+		},
+		{
+			// Matches both source and target filters of rule2,
+			// inhibited because sourceAlert2 matches also both the
+			// source and target filterof rule1.
```
stuartnelson3: `filter of`

stuartnelson3: Should this say `not inhibited`? `expected=false` is in the test, and it seems to be fitting your definition of when not to self-inhibit.

beorn7: Fixed.

beorn7:
Note: Please don't squash when you merge this as I have already more work on top of the commits in this PR.

mxinden:
I will take a look at this tomorrow morning.

inhibit/inhibit_test.go
```diff
@@ -140,55 +140,87 @@ func TestInhibitRuleHasEqual(t *testing.T) {
...
+			// Matches both source and target filters of rule1,
+			// inhibited because sourceAlert1 matches only the
+			// source filter of rule1.
+			target:   model.LabelSet{"s1": "1", "t1": "1", "e": "1"},
+			expected: true,
+		},
+		{
+			// Matches both source and target filters of rule2,
+			// not inhibited because sourceAlert2 matches also both the
+			// source and target filter of rule1.
```
stuartnelson3: should this be `rule2`? It's not inhibited because this alert and the source alert both match target + source rules in the inhibition rule2.

beorn7:
> Given that releasing Alertmanager usually takes a bit and the last minor release was a month ago, I would look into the next one (v0.17.0) soonish. I would suggest to get this patch into v0.17.0.
> 
> What do you all think?

Yes, that would be great.

inhibit/inhibit_test.go
beorn7: Yes, you are right. Thanks for catching yet another blunder in the doc comment.

beorn7:
Doc comment issue fixed. I'll merge this on green, and I'll send a PR for the documentation update.

PR #1770, opened 2019-02-26
config: update coordinator's logs

simonpasquier:
Remove duplicate logs and add a log when the reloading is complete.


beorn7:
Promu seems to have a problem downloading. But code looks good to me.

PR #1771, opened 2019-02-26
Create a `Muter` implementation for silences

beorn7:
This encapsulates the logic of querying and marking silenced
alerts. It removes the code duplication flagged earlier.

I removed the error returned by the setAlertStatus function as we were
only logging it, and that's already done anyway when the error is
received from the `silence.Query` call (now in the `Mutes` method).

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -29,6 +29,7 @@ import (
 
 	"github.com/prometheus/alertmanager/cluster"
 	"github.com/prometheus/alertmanager/config"
+	"github.com/prometheus/alertmanager/inhibit"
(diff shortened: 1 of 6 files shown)


notify/notify.go
```diff
@@ -359,72 +359,27 @@ func (n *GossipSettleStage) Exec(ctx context.Context, l log.Logger, alerts ...*t
 	return ctx, alerts, nil
 }
 
-// InhibitStage filters alerts through an inhibition muter.
-type InhibitStage struct {
+// MuteStage filters alerts through a Muter.
+type MuteStage struct {
```
stuartnelson3: :heart_eyes:

PR #1774, opened 2019-02-27
Performance improvement checking the silenced state

beorn7:
I have also found a bug with zero retention time while doing so. See individual commits.

The first commit just adds a benchmark, which showed me that we need this optimization for high-alert scenarios.

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -23,6 +23,7 @@ import (
 	"os"
 	"reflect"
 	"regexp"
+	"sort"
 	"sync"
 	"time"
 
@@ -111,22 +112,68 @@ func NewSilencer(s *Silences, m types.Marker, l log.Logger) *Silencer {
 
 // Mutes implements the Muter interface.
 func (s *Silencer) Mutes(lset model.LabelSet) bool {
-	sils, err := s.silences.Query(
-		QState(types.SilenceStateActive),
-		QMatches(lset),
+	fp := lset.Fingerprint()
+	ids, markerVersion, _ := s.marker.Silenced(fp)
+
+	var (
+		err        error
+		sils       []*pb.Silence
+		newVersion = markerVersion
 	)
+	if markerVersion == s.silences.Version() {
+		// No new silences added, just need to check which of the old
+		// silences are still revelant.
+		if len(ids) == 0 {
+			// Super fast path: No silences ever applied to this
+			// alert, none have been added. We are done.
+			return false
+		}
+		// This is still a quite fast path: No silences have been added,
+		// we only need to check which of the applicable silences are
+		// currently active. Note that newVersion is left at
+		// markerVersion because the Query call might already return a
+		// newer version, which is not the version our old list of
+		// applicable silences is based on.
+		sils, _, err = s.silences.Query(
+			QIDs(ids...),
+			QState(types.SilenceStateActive),
+		)
+	} else {
+		// New silences have been added, do a full query.
+		sils, newVersion, err = s.silences.Query(
+			QState(types.SilenceStateActive),
+			QMatches(lset),
+		)
+	}
 	if err != nil {
 		level.Error(s.logger).Log("msg", "Querying silences failed, alerts might not get silenced correctly", "err", err)
 	}
 	if len(sils) == 0 {
-		s.marker.SetSilenced(lset.Fingerprint())
+		s.marker.SetSilenced(fp, newVersion)
 		return false
 	}
-	ids := make([]string, len(sils))
-	for i, s := range sils {
-		ids[i] = s.Id
+	idsChanged := len(sils) != len(ids)
+	if !idsChanged {
+		// Length is the same, but is the content the same?
+		for i, s := range sils {
+			if ids[i] != s.Id {
+				idsChanged = true
+				break
+			}
+		}
 	}
-	s.marker.SetSilenced(lset.Fingerprint(), ids...)
+	if idsChanged {
+		// Need to recreate ids.
+		ids = make([]string, len(sils))
+		for i, s := range sils {
+			ids[i] = s.Id
+		}
+		sort.Strings(ids) // For comparability.
+	}
+	if idsChanged || newVersion != markerVersion {
+		// Update marker only if something changed.
+		s.marker.SetSilenced(fp, newVersion, ids...)
+	}
 	return true
 }
 
@@ -139,6 +186,7 @@ type Silences struct {
 
 	mtx       sync.RWMutex
 	st        state
+	version   int // Increments whenever silences are added.
 	broadcast func([]byte)
 	mc        matcherCache
 }
@@ -441,7 +489,9 @@ func (s *Silences) setSilence(sil *pb.Silence) error {
 		return err
 	}
 
-	s.st.merge(msil, s.now())
+	if s.st.merge(msil, s.now()) {
+		s.version++
+	}
 	s.broadcast(b)
 
 	return nil
@@ -615,7 +665,7 @@ func QState(states ...types.SilenceState) QueryParam {
 // QueryOne queries with the given parameters and returns the first result.
 // Returns ErrNotFound if the query result is empty.
 func (s *Silences) QueryOne(params ...QueryParam) (*pb.Silence, error) {
-	res, err := s.Query(params...)
+	res, _, err := s.Query(params...)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


silence/silence.go
```diff
@@ -780,7 +827,9 @@ func (s *Silences) SetBroadcast(f func([]byte)) {
 type state map[string]*pb.MeshSilence
 
 func (s state) merge(e *pb.MeshSilence, now time.Time) bool {
+	id := e.Silence.Id
 	if e.ExpiresAt.Before(now) {
+		delete(s, id) // Just in case it is still lingering.
```
mxinden: I am not sure this preserves the silence deletion logic. Wouldn't it delete all copies of a silence during cluster replication once a user expires it? Even though the silence is logically useless once it is expired, thereby safe to delete, the UI differentiates between expired and non-existing silences.

mxinden:
In regards to the long term vision above: I am not opposed to introducing the state version with this patch.

beorn7:
> What do you think of leveraging a database in the long run? Could be included in the binary.

I can't really say. I'm not an expert in embedded databases. My gut feeling is that it will be hard to cater for exactly our use case without a lot of customization. The expensive part here is the quadratic number of regexp matches to be performed (number of alerts being checked * number of silences * average number of regexp matchers per silence). With the change here, we are essentially tracking the regexp matches already calculated, but not precisely (i.e. if _anything_ gets added, we re-match everything). This could be improved even more, but probably isn't worth the additional complication at this point. An embedded database that would do this kind of optimization “for free” would be great.

silence/silence.go
```diff
@@ -111,22 +112,54 @@ func NewSilencer(s *Silences, m types.Marker, l log.Logger) *Silencer {
...
+	fp := lset.Fingerprint()
+	ids, markerVersion, _ := s.marker.Silenced(fp)
+
+	var (
+		err        error
+		sils       []*pb.Silence
+		newVersion = markerVersion
 	)
+	if markerVersion == s.silences.Version() {
+		// No new silences added, just need to change which of the
```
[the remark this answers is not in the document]
beorn7: FIxed.

brancz:
Sounds like we could get a lot of performance improvements by applying a similar strategy as the inverted index used in tsdb. We've been wanting to try reusing it for a while now, it was first brought up at the dev summit 2017 if I recall correctly.

silence/silence.go
```diff
@@ -625,16 +661,17 @@ func (s *Silences) QueryOne(params ...QueryParam) (*pb.Silence, error) {
...
-// Query for silences based on the given query parameters.
-func (s *Silences) Query(params ...QueryParam) ([]*pb.Silence, error) {
+// Query for silences based on the given query parameters. It returns the
+// resulting silences and the state version the result is based on.
+func (s *Silences) Query(params ...QueryParam) ([]*pb.Silence, int, error) {
 	start := time.Now()
 	s.metrics.queriesTotal.Inc()
 
-	sils, err := func() ([]*pb.Silence, error) {
+	sils, version, err := func() ([]*pb.Silence, int, error) {
```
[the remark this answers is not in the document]
beorn7: I was wondering myself. My guess is a hack to use `return` in line 674 as "throwing an exception". As this is causing confusion, I'll think about refactoring it.

silence/silence.go
beorn7: I'm not an expert of this codebase, so I am not sure if I got this right. But I think you got confused here by a somewhat unfortunate naming. (Naming is hard, as we know…) There are essentially two kinds of expirations for silences:
1. The silence itself expires, i.e. it's `EndsAt` date has been reached, or it “was expired” by setting its `endsAt` date to “now”.
2. The silence expires from the storage, i.e. we don't want to handle it in the storage anymore, which is the case if its `ExpiresAt` date has been reached (which is saved in the wrapping `pb.MeshSilence` and not in the silence itself).

This line is about the second case, i.e. we are really not interested in this silence anymore at all. The reason why this `merge` method returns `false` in that case without doing anything is that we won't bother merging the silence into the storage if we are not interested in it anymore.

The problem this creates happens if this collides with the way we explicitly expire silences (the first case) in combination with a very short or zero retention time (something that should not happen in practice, but could be set): If we expire a silence (the first case), we set its `EndsAt` to “now”. This happens on a copy of the silence (for concurrency safety or something), and then we `merge` the modified, copied silence (which still has the same ID) back into the storage. But what we also do (and must do) is to set the `ExpiresAt` to “now” + retention time. In case that's a finite time, everything is fine, but if the retention time is zero, now `ExpiresAt` is also “now”. At the time we call `merge`, the “now” is already increased a bit, and thus, `merge` sees the silence as not only expired (the first case) but also as expired from the storage (the second case) and thus it returns false and doesn't do anything, which means that _the original silence we copied for the expiration is still in the storage in unchanged form_. And that's a bug.

beorn7: And I think I have found a better fix, thanks to the rubberducking above.
I think a cleaner solution is to use the same “now” for merging the silence as the “now” we used to set the silence as expired (the first case).
I'll modify the respective commit. (It requires the internal `setSilence` to take a “now” parameter similar to the `merge`.)

beorn7:
There were also two TODOs I still wanted to resolve. Next push imminent…

beorn7:
@brancz Note that the approach so far was to lazily evaluate muted state when we need it, i.e. an the time an alert is due to create a notification. That was a bit too lazy for the UI, thus we are working on making whatever the API calls report the current muting state. That's still lazy as we are not updating everything when a new silence is created, and we are not updating everything everytime an alert comes in. With an indexing approach, all those indexes (call them inverted or not) needed to be kept up to date, which might be a lot of effort giving the volatile nature of alerts, inhibitions, and silences coming and going all the time. (Not saying it wouldn't work, just noting that it might be more complicated than it looks. This PR is essentially an attempt to do something relatively simple to harvest low-hanging fruit.)

beorn7:
Now the comment is fixed, the two TODOs I forgot about are done, and the fix for the expiration bug is much nicer. I'm still working on a commit to avoid that closure.

beorn7:
And now also removed that closure that confused us. (The `prometheus.Timer` helped here.)

beorn7:
Tests seem flakier than usual. Works locally. Apparently, there was a temporary problem downloading errcheck. But then tests sometimes fail for other reasons.

beorn7:
I have removed the `-u` from the `go get` for `errcheck` because I think it makes more sense without it.
However, that doesn't really explain why the tests all fail when `go get`ing `errcheck`. Just a byproduct of my troubleshooting.

beorn7:
And BTW: Despite force-pushing (or because of it), I tried to create a meaningful commit structure. So please merge without squashing.

beorn7:
Still, my Makefile change made tests less flaky, as it appears…

beorn7:
Running this now on our canary cluster.

mxinden:
> Running this now on our canary cluster.

@beorn7 Would you mind sharing some stats here to estimate the impact of the change?

beorn7:
Wo "only" have 283 active alerts on our canary cluster and 45 silences, at which scale everything is blazing fast. Will have to create artificial alerts to load test properly.

beorn7:
I have now 5000 alerts and 50 silences (that silence effectively everything).
If I request all alerts (without showing me silenced ones), the query should go through all alerts, but not send anything to the browser (because otherwise the load would probably be dominated by shoveling data via the network).

It takes about 90ms to query the relevant alert API endpoint. After creating a new silence (and thus incrementing the version of the silence state), the API call takes about 170ms.

beorn7:
And now 5000 alerts, 500 silences, full query after updated state: 1000ms, full query with same state: 100ms.

I think that's pretty decent. SoundCloud has usually only 50 silences in place, and we are already a pretty heavy Prometheus user. And you would usually not query all alerts in the first place, in particular not in larger shops. (I talked to @stuartnelson3 about future steps to never have the API return an unlimited amount of alerts. Currently, the landing page by default displays everything without any pagination. That's not good.)

stuartnelson3:
If a user has so many alerts that this causes problems, rendering all of them (as is currently done) probably already makes it nearly unusable.

>I talked to @stuartnelson3 about future steps to never have the API return an unlimited amount of alerts. Currently, the landing page by default displays everything without any pagination. That's not good.

We've made plans to reduce the amount of alerts sent to the main page, and also to reduce the amount rendered to the page. But, that's for another PR.

silence/silence.go
mxinden: I understand the reasoning for deleting the storage silence given that the to-be-merged silence is expired. Thanks for the detailed explanation.

>  The silence expires from the storage, i.e. we don't want to handle it in the storage anymore, which is the case if its `ExpiresAt` date has been reached (which is saved in the wrapping `pb.MeshSilence` and not in the silence itself).

@beorn7 I am not sure this is true. A silence that is expired should still be in the storage at least for some reasonable amount of time to give the user feedback that the silence is indeed expired.

silence/silence.go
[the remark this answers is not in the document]
mxinden: Thanks for cleaning this up @beorn7!

stuartnelson3:
@mxinden any more thoughts? I'm happy with this

silence/silence.go
beorn7: Yes, it "should". But if the user has set zero retention time, then that's what they get: zero retention. That's the current state of the code, and that's by design. We could enforce a minimum retention time, but that has nothing to do with this PR.

Please also note that I have removed the explicit deletion and just use a consistent "now" so that the bug that expired silences stay unchanged (provided a zero retention time has been configured) is fixed with this PR in a clean way.

beorn7:
@mxinden is there anything left preventing merging this?

beorn7:
Ooooh, race condition in the review process. :o)

silence/silence.go
mxinden: Right, understood. Thanks for that!

> Please also note that I have removed the explicit deletion and just use a consistent "now" so that the bug that expired silences stay unchanged (provided a zero retention time has been configured) is fixed with this PR in a clean way.

Yes, saw that. Looks good to me.

mxinden:
@beorn7 merged without squash.

beorn7:
> merged without squash.

:bowing_man:

PR #1776, opened 2019-02-28
Make the silence preview show also muted alerts

beorn7:
Even if an alert is already silenced and/or muted, I still want to
know that my newly created silence will affect it.


beorn7:
Tests fail because #1774 is not part of this, which contains a fix for errcheck.

beorn7:
I'll wait for #1774 so that we can see green tests here.

beorn7:
Rebased to see if CI is satisfied now.

beorn7:
Needed to re-run one (flaky acceptance test), but at least it's all green now.

PR #1779, opened 2019-03-01
update Makefile.common and license headers

pgier:
Sync Makefile.common to latest which updates promu version
and adds license check to default target.
Add missing license headers.

Signed-off-by: Paul Gier <pgier@redhat.com>


pgier:
@simonpasquier I added the license information to `api/v2/restapi/configure_alertmanager.go` in order to pass the license check, however this causes the test to fail because the file no longer matches the generated file.
I could add a rule to the license check to ignore this file, similar to the other generated file, or is there a way to customize the generated file(s) to include the license?

simonpasquier:
`make clean` removes everything in `api/v2/restapi` (eg including `api/v2/restapi/configure_alertmanager.go`) though this file doesn't have to be regenerated (see [comment](https://github.com/prometheus/alertmanager/blob/master/api/v2/restapi/configure_alertmanager.go#L1)). The `clean` target should probably be tweaked to deal with this case:

```diff
diff --git a/Makefile b/Makefile
index aa8ebeb..a9c3477 100644
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,7 @@ test/with_api_v2/api_v2_client/models test/with_api_v2/api_v2_client/client: api
 .PHONY: clean
 clean:
        - rm -f asset/assets_vfsdata.go
-       - rm -r api/v2/models api/v2/restapi test/with_api_v2/api_v2_client/models test/with_api_v2/api_v2_client/client
+       - for f in $$(find api/v2 test/with_api_v2 -name *.go); do if head -q -n 1 $$f | grep -q "DO NOT EDIT"; then rm -f $$f; fi; done
        - cd $(FRONTEND_DIR) && $(MAKE) clean
 
 .PHONY: test
```

mxinden:
Given that we don't do any configuration in `api/v2/restapi/configure_alertmanager.go` I decided to remove it on each generation run to keep complexity low. As long as we don't need it to be persisted I would suggest continuing to remove it. Not a strong opinion though.

Both excluding it from the license check or using the [`--copyright-file=      #   copyright file used to add copyright header`](https://goswagger.io/generate/server.html) flag sounds good to me.

simonpasquier:
`--copyright-file` seems promising!

pgier:
I went ahead with the `--copyright-file` option since I ran into some problems with the other ideas.  For example, if we try to remove most generated files but leave the `configure_alertmanager.go` then make sees the existing directories and thinks it doesn't need to regenerate them.

I'm also not crazy about the way that `make clean` removes a bunch of files that are in git.  It seems like after a new git clone/checkout, a `make clean` should basically be a no-op.  Should these generated files not be checked into git at all?  Or should we rename the clean target to something like regenerate-api and regenerate-assets?

pgier:
Also, I didn't include the year in the copyright header because I was worried about the travis build breaking in 2020 and later.

mxinden:
The only reason why we check-in the files into version control is to not require developers to have the openapi tooling installed on their machine. Given that this causes so much trouble I am also fine with removing it. I guess we could even build all the go-based-tooling from the vendor directory like we do with vfsdata.go, right @simonpasquier?

pgier:
Updated to resolve merge conflict.  @simonpasquier @mxinden Any other changes needed before this can be merged?

PR #1789, opened 2019-03-08
apiv2: Allow empty matcher

### LINKED ISSUE #1731: creating an alert with labelname and no matcher creates a 400 in the interface
br

I often create alerts with random label name and no value, that silences everything in the AM.

But it generates a 400 http error in the interface.

roidelapluie:
Fixes #1731

### DIFF
--- a/pkg/parse/parse_test.go
+++ b/pkg/parse/parse_test.go
@@ -116,6 +116,15 @@ func TestMatchers(t *testing.T) {
 			}(),
 		},
 		{
+			// not correct as prometheus syntax but the new api v2 will provide this
+			input: `{foo=}`,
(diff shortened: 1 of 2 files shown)


pkg/parse/parse_test.go
```diff
@@ -115,6 +115,15 @@ func TestMatchers(t *testing.T) {
 				return append(ms, m, m2)
 			}(),
 		},
+		{
+			// not correct as prometheus syntax but the new api v2 will provide this
```
stuartnelson3: which api v2 are you referring to?

simonpasquier:
Fixed by #1869. Feel free to re-open @roidelapluie if you don't agree ;-)

PR #1791, opened 2019-03-11
re-add alert groups endpoint

stuartnelson3:
requested in #868 

this contains a lot that is open to discussion, including:

- what the definition of the returned struct/struct members for the alert groups endpoint should be
- the implementation of finding the groups, particularly the weird bit with the giving the `EnrichedAlert` a reference to a slice of its receivers. `staticcheck` isn't happy about how I went about doing this, either.
- all the naming (like `EnhancedAlert`), and the path name `/alerts/groups`

I decided to wait on writing any tests until we can figure out what it is we want to have in this pr


stuartnelson3:
PR that removed alert groups originally: https://github.com/prometheus/alertmanager/pull/1525/files

api/v2/openapi.yaml
```diff
@@ -189,6 +189,50 @@ paths:
           $ref: '#/responses/InternalServerError'
         '400':
           $ref: '#/responses/BadRequest'
+  /alerts/groups:
+    get:
+      tags:
+        - alertgroups
```
mxinden:
```suggestion
        - alertgroup
```

Given that the other endpoints have singular tags, I would suggest the same here.

api/v2/openapi.yaml
```diff
@@ -410,6 +454,21 @@ definitions:
...
+    type: array
+    items:
+      $ref: '#/definitions/alertGroup'
+  alertGroup:
+    type: object
+    properties:
+      labels:
+        $ref: '#/definitions/labelSet'
+      receiver:
+        type: string
```
mxinden: What do you think of reusing the `receiver` definition below?

Why is `receiver` an object? I thought we might want to expose extra info about a single receiver in the future.

api/v2/api.go
```diff
@@ -389,6 +395,89 @@ func (api *API) postAlertsHandler(params alert_ops.PostAlertsParams) middleware.
...
+	res := make(open_api_models.AlertGroups, 0, len(alertGroups))
+
+	for _, alertGroup := range alertGroups {
+		ag := &open_api_models.AlertGroup{
+			Receiver: alertGroup.Receiver,
+			Labels:   modelLabelSetToAPILabelSet(alertGroup.Labels),
+			Alerts:   make([]*open_api_models.GettableAlert, 0, len(alertGroup.Alerts)),
+		}
+
+		for _, ea := range alertGroup.Alerts {
```
mxinden: What does `ea` stand for?

api/v2/api.go
```diff
@@ -389,6 +395,89 @@ func (api *API) postAlertsHandler(params alert_ops.PostAlertsParams) middleware.
...
+			state := string(ea.Status.State)
```
mxinden: This seems to duplicate a lot of logic in `getAlertsHandler`, what do you think of consolidating some of it?

api/v2/api.go
stuartnelson3: enriched alert. happy to change it, it's fairly cryptic.

api/v2/api.go
stuartnelson3: i'll see what i can do

api/v2/openapi.yaml
stuartnelson3: Hm, any idea what we might expose? A receiver is mostly connection information that we don't want to expose.

stuartnelson3:
Aw, if I accept your suggestions they fail because they have no DCO

dispatch/dispatch.go
```diff
@@ -133,6 +138,134 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
...
+// AlertGroup represents how alerts alerts exist within an aggrGroup
+type AlertGroup struct {
+	Alerts   []*EnrichedAlert `json:"alerts"`
+	Labels   model.LabelSet   `json:"labels,omitempty"`
+	Receiver string           `json:"receiver,omitempty"`
+}
+
+// EnrichedAlert is the representation of an alert, with status info and
+// fingerprint.
+type EnrichedAlert struct {
```
stuartnelson3: Since this is also being used in the apiv2 package, maybe it should be moved to `types`?

api/v2/openapi.yaml
mxinden: Not aware of anything specific. I am fine with either way, I would just prefer to be consistent.

api/v2/api.go
mxinden: That would be helpful, thanks.

dispatch/dispatch.go
```diff
@@ -133,6 +138,134 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
...
+}
+
+type AlertGroups []*AlertGroup
+
+func (ag AlertGroups) Swap(i, j int)      { ag[i], ag[j] = ag[j], ag[i] }
+func (ag AlertGroups) Less(i, j int) bool { return ag[i].Labels.Before(ag[j].Labels) }
+func (ag AlertGroups) Len() int           { return len(ag) }
+
+// Groups returns a slice of AlertGroups from the dispatcher's internal state.
+func (d *Dispatcher) Groups(matchers []*labels.Matcher, receiverFilter *regexp.Regexp, silenced, inhibited, active bool) AlertGroups {
```
simonpasquier: Passing all those filtering parameters, doing the filter in this function and having to pass the `setAlertStatus` function to the dispatcher feels a bit awkward. Same for the fact that `EnrichedAlert` contains information that doesn't really concern the `dispatch` package.

What about passing callback functions to `dispatch.Groups` instead?

```go
// EnrichedAlert is an alert with its associated receivers.
type EnrichedAlert struct {
        *types.Alert
        Receivers *[]string
}


func (d *Dispatcher) Groups(routeFilter func(*Route) bool, alertFilter func(*types.Alert) bool) AlertGroups {
        ...
        for route, ags := range d.aggrGroups {
                if !routeFilter(route) {
                        continue
                }
                ...
                for _, ag := range ags {
                        ...
                        for a := range alerts {
                                if !alertFilter(a) {
                                        continue
                                }
```

dispatch/dispatch.go
```diff
@@ -133,6 +138,134 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
...
+					// ea stores a pointer to this slice,
+					// so any changes to this slice will be
+					// available from ea.
+					r = append(r, receiver)
+				} else {
+					// First time we've seen this alert.
+					// Initialize a new receivers slice.
+					r := []string{receiver}
+					receivers[fp] = r
+					ea.Receivers = &r
```
simonpasquier: To get rid of the pointer to slice trick, this function could also return the `map[model.Fingerprint][]string` variable and let the caller look up the receiver list using the alert's fingerprint.

api/v2/openapi.yaml
stuartnelson3: Unfortunately, I'm fine either way too :) if you want to add the extra info let me know, otherwise i'll leave it as-is

stuartnelson3:
Made a lot of changes based on @simonpasquier 's suggestions, when you have a chance take a look. Did some rough manual testing and it seems to be working.

api/v2/api.go
```diff
@@ -53,6 +54,7 @@ type API struct {
 	peer           *cluster.Peer
 	silences       *silence.Silences
 	alerts         provider.Alerts
+	groups         groupsFn
```
mxinden: For me this is ambiguous, does it return groups of alerts, silences, receivers, apples, ...? How about `alertGroups`?

api/v2/api.go
```diff
@@ -389,6 +341,143 @@ func (api *API) postAlertsHandler(params alert_ops.PostAlertsParams) middleware.
...
+			return alertgroup_ops.NewGetAlertGroupsBadRequest().WithPayload(err.Error())
+		}
+
+		matchers = append(matchers, matcher)
+	}
+
+	if params.Receiver != nil {
+		receiverFilter, err = regexp.Compile("^(?:" + *params.Receiver + ")$")
+		if err != nil {
+			return alertgroup_ops.
```
mxinden: In the above `for` loop we log the error, here we don't. Can we be consistent with either way?

api/v2/openapi.yaml
mxinden: Then I would prefer reusing the receiver definition below. (We might want to expose the receiver type in the future.)

api/v2/api.go
stuartnelson3: Apples would be nice. Will update :)

api/v2/api.go
```diff
@@ -389,6 +341,144 @@ func (api *API) postAlertsHandler(params alert_ops.PostAlertsParams) middleware.
...
+
+		if !alertMatchesFilterLabels(&a.Alert, matchers) {
+			return false
+		}
+
+		return true
+	}
+}
+
+func routeFilter(receiverFilter *regexp.Regexp) func(r *dispatch.Route) bool {
```
simonpasquier: (nit) unless you wanted to test it in isolation, I would inline this function in `getAlertGroupsHandler()` since it is only used there.

stuartnelson3:
ptal when you two have a chance

dispatch/dispatch.go
```diff
@@ -133,6 +133,86 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
 	}
 }
 
+// AlertGroup represents how alerts exist within an aggrGroup.
+type AlertGroup struct {
+	Alerts   []*types.Alert `json:"alerts"`
```
simonpasquier: (nit) don't need the json tags anymore.

prymitive:
> Logic in `dispatch` looks very clean! Just small comments, nothing major.
> 
> Given that this could fix the compatibility issues with karma, @prymitive do you have thoughts on this as well?

In general this seem to make it easier for me, for example karma already has a concept of alert group, I don't see anything that could be problematic for me in any obvious ways. That being said I'm currently traveling so I'm not gonna pretend that I've put a lot of effort into reviewing this or trying to use this code.

stuartnelson3:
Added a test

prymitive:
FYI I'll be back home next week and I'll be able to do some integration tests agains this PR for karma, if that's helpful

stuartnelson3:
Definitely! would love to hear if this is useful for you, or if there's anything you would like to add/change

mohsen0:
Can this PR be progressed?
Thanks

stuartnelson3:
>Can this PR be progressed?

Waiting for some feedback from @prymitive, and if there are any more questions/comments from @mxinden and @simonpasquier

prymitive:
I hope to get some work done today evening

prymitive:
I was able to get it working easily with those changes and it seems to be working perfectly 🎉
I used the alert generator I have on http://karma-demo.herokuapp.com and I got exact same set of alerts via the v2 API 🎉
It also looks like I could kill some of the ugly code I have in the backend if I would drop support for non openapi alertmanager versions.
There's really not much feedback I can provide other than "it works" since all karma is doing is pulling all silences and alerts, so it's a very basic use of the API provided.
Thanks! Great work!

stuartnelson3:
great to hear! once I get a final :+1: from @mxinden and @simonpasquier this will be merged

divishkumar:
@stuartnelson3 You got a chance to look at Max's comment ? Would be great to have this merged in soon. Thanks for your work on this :)

stuartnelson3:
>@stuartnelson3 You got a chance to look at Max's comment ? Would be great to have this merged in soon. Thanks for your work on this :)

thanks for the ping, i missed his message!

>Other than #1791 (comment) this looks good to me. Thanks @stuartnelson3!

sorry for missing that log line, I added it to the alert groups endpoint but not the normal alerts endpoint. hopefully now this PR is good to go

mohsen0:
@mxinden Are you happy with the latest changes?

xocasdashdash:
Once this PR is merged, when do you think the next alertmanager version is going to be released?

stuartnelson3:
I plan on releasing 0.17.0 next week, which will include this PR.

PR #1792, opened 2019-03-12
api/v2: move generated client code from test to api/v2

pgier:
- Move the generated api/v2 client code out of the test directory
and into the api/v2 directory with models and restapi.
- Remove duplicate models directory
- Update tests to use api/v2 package for models and client


pgier:
@mxinden please take a look

PR #1794, opened 2019-03-12
Fix bad syntax for struct tag value

knweiss:
This fixes the following error found by go vet:

config/notifiers.go:546:2: struct field tag
`yaml:"url_title,omitempty" json:"url_title,omitempty` not compatible with
reflect.StructTag.Get: bad syntax for struct tag value (govet)
	URLTitle string   `yaml:"url_title,omitempty" json:"url_title,omitempty`
	^

Signed-off-by: Karsten Weiss <knweiss@gmail.com>

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -543,7 +543,7 @@ type PushoverConfig struct {
 	Title    string   `yaml:"title,omitempty" json:"title,omitempty"`
 	Message  string   `yaml:"message,omitempty" json:"message,omitempty"`
 	URL      string   `yaml:"url,omitempty" json:"url,omitempty"`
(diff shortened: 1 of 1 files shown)


knweiss:
@mxinden You may also want to consider running [golangci-lint](https://github.com/golangci/golangci-lint) instead. It's the inofficial successor of `gometalinter` and integrates many checkers/linters (`go vet` is one of them) and has a practical default config. Also, its architecture is IMHO really nice. (The `go-swagger` project uses it, too.)

simonpasquier:
golangci-lint looks interesting, thanks for the pointer @knweiss! It would also unify our use of `staticcheck` and `errcheck` with a single tool.

PR #1798, opened 2019-03-14
cli: update amtool to use apiv2

pgier:
Update the amtool cli to use the v2 API
Includes godoc improvements

### DIFF
--- a/cli/alert_query.go
+++ b/cli/alert_query.go
@@ -74,7 +74,6 @@ func configureQueryAlertsCmd(cc *kingpin.CmdClause) {
 }
 
 func (a *alertQueryCmd) queryAlerts(ctx context.Context, _ *kingpin.ParseContext) error {
-	filter := []string{}
 	if len(a.matcherGroups) > 0 {
 		// Attempt to parse the first argument. If the parser fails
 		// then we likely don't have a (=|=~|!=|!~) so lets assume that
@@ -85,7 +84,6 @@ func (a *alertQueryCmd) queryAlerts(ctx context.Context, _ *kingpin.ParseContext
 		if err != nil {
 			a.matcherGroups[0] = fmt.Sprintf("alertname=%s", m)
 		}
-		filter = append(filter, a.matcherGroups[0])
 	}
 
 	// If no selector was passed, default to showing active alerts.
@@ -93,12 +91,12 @@ func (a *alertQueryCmd) queryAlerts(ctx context.Context, _ *kingpin.ParseContext
 		a.active = true
 	}
 
-	alertParams := alert.NewGetAlertsParams().WithContext(ctx)
-	alertParams.Active = &a.active
-	alertParams.Inhibited = &a.inhibited
-	alertParams.Silenced = &a.silenced
-	alertParams.Unprocessed = &a.unprocessed
-	alertParams.Filter = filter
+	alertParams := alert.NewGetAlertsParams().WithContext(ctx).
+		WithActive(&a.active).
+		WithInhibited(&a.inhibited).
+		WithSilenced(&a.silenced).
+		WithUnprocessed(&a.unprocessed).
+		WithFilter(a.matcherGroups)
 
 	amclient := NewAlertmanagerClient(alertmanagerURL)
 
--- a/cli/alert_query.go
+++ b/cli/alert_query.go
@@ -17,13 +17,11 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 
-	"github.com/prometheus/client_golang/api"
-	"gopkg.in/alecthomas/kingpin.v2"
+	kingpin "gopkg.in/alecthomas/kingpin.v2"
 
+	"github.com/prometheus/alertmanager/api/v2/client/alert"
 	"github.com/prometheus/alertmanager/cli/format"
-	"github.com/prometheus/alertmanager/client"
 	"github.com/prometheus/alertmanager/pkg/parse"
 )
 
@@ -76,7 +74,7 @@ func configureQueryAlertsCmd(cc *kingpin.CmdClause) {
 }
 
 func (a *alertQueryCmd) queryAlerts(ctx context.Context, _ *kingpin.ParseContext) error {
-	var filterString = ""
+	filter := []string{}
 	if len(a.matcherGroups) > 0 {
 		// Attempt to parse the first argument. If the parser fails
 		// then we likely don't have a (=|=~|!=|!~) so lets assume that
@@ -87,19 +85,25 @@ func (a *alertQueryCmd) queryAlerts(ctx context.Context, _ *kingpin.ParseContext
 		if err != nil {
 			a.matcherGroups[0] = fmt.Sprintf("alertname=%s", m)
 		}
-		filterString = fmt.Sprintf("{%s}", strings.Join(a.matcherGroups, ","))
+		filter = append(filter, a.matcherGroups[0])
 	}
 
-	c, err := api.NewClient(api.Config{Address: alertmanagerURL.String()})
-	if err != nil {
-		return err
-	}
-	alertAPI := client.NewAlertAPI(c)
 	// If no selector was passed, default to showing active alerts.
 	if !a.silenced && !a.inhibited && !a.active && !a.unprocessed {
 		a.active = true
 	}
-	fetchedAlerts, err := alertAPI.List(ctx, filterString, a.receiver, a.silenced, a.inhibited, a.active, a.unprocessed)
+
+	alertParams := alert.NewGetAlertsParams().WithContext(ctx)
+	alertParams.Active = &a.active
+	alertParams.Inhibited = &a.inhibited
+	alertParams.Silenced = &a.silenced
+	alertParams.Unprocessed = &a.unprocessed
+	alertParams.Filter = filter
+
+	amclient := NewAlertmanagerClient(alertmanagerURL)
+
+	getOk, err := amclient.Alert.GetAlerts(alertParams)
+
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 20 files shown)


cli/utils.go
```diff
@@ -125,51 +127,58 @@ func convertClientToCommonLabelSet(cls client.LabelSet) model.LabelSet {
...
+		labelSet[name] = value
 	}
 
 	return labelSet, nil
 }
 
-// Only valid for when you are going to add a silence
-func TypeMatchers(matchers []labels.Matcher) (types.Matchers, error) {
-	typeMatchers := types.Matchers{}
+// TypeMatchers Only valid for when you are going to add a silence
```
mxinden:
```suggestion
// TypeMatchers only valid for when you are going to add a silence
```

cli/utils.go
```diff
@@ -125,51 +127,58 @@ func convertClientToCommonLabelSet(cls client.LabelSet) model.LabelSet {
...
-			return types.Matchers{}, err
+			return models.Matchers{}, err
 		}
 		typeMatchers = append(typeMatchers, &typeMatcher)
 	}
 	return typeMatchers, nil
 }
 
-// Only valid for when you are going to add a silence
+// TypeMatcher Only valid for when you are going to add a silence
```
mxinden:
```suggestion
// TypeMatcher only valid for when you are going to add a silence
```

cli/silence_import.go
```diff
@@ -59,20 +58,21 @@ func configureSilenceImportCmd(cc *kingpin.CmdClause) {
...
 }
 
-func addSilenceWorker(ctx context.Context, sclient client.SilenceAPI, silencec <-chan *types.Silence, errc chan<- error) {
+func addSilenceWorker(ctx context.Context, sclient *silence.Client, silencec <-chan *models.PostableSilence, errc chan<- error) {
 	for s := range silencec {
-		silenceID, err := sclient.Set(ctx, *s)
-		sid := s.ID
+		params := silence.NewPostSilencesParams().WithContext(ctx)
+		params.Silence = s
+		postOk, err := sclient.PostSilences(params)
```
simonpasquier: This part should probably be adapted as `POST /api/v2/silences` can only return 200 or 400 and it should update (recreate) an existing silence.

https://github.com/prometheus/alertmanager/blob/f40ecee77326d169dec5092d28510c60a9593f0b/api/v2/openapi.yaml#L77-L86

cli/silence_import.go
```diff
@@ -123,7 +119,7 @@ func (c *silenceImportCmd) bulkImport(ctx context.Context, _ *kingpin.ParseConte
 
 	count := 0
 	for dec.More() {
-		var s types.Silence
+		var s models.PostableSilence
```
simonpasquier: It needs to be mentioned somewhere that silence backups created prior to this PR won't be supported after.

cli/silence_query.go
```diff
@@ -101,35 +99,35 @@ func (c *silenceQueryCmd) query(ctx context.Context, _ *kingpin.ParseContext) er
 		if err != nil {
 			c.matchers[0] = fmt.Sprintf("alertname=%s", c.matchers[0])
 		}
-		filterString = fmt.Sprintf("{%s}", strings.Join(c.matchers, ","))
+		filter = append(filter, c.matchers[0])
```
simonpasquier: Same remark as for alerts.

cli/utils.go
pgier: Done

cli/utils.go
pgier: Done

cli/alert_query.go
```diff
@@ -87,19 +85,25 @@ func (a *alertQueryCmd) queryAlerts(ctx context.Context, _ *kingpin.ParseContext
 		if err != nil {
 			a.matcherGroups[0] = fmt.Sprintf("alertname=%s", m)
 		}
-		filterString = fmt.Sprintf("{%s}", strings.Join(a.matcherGroups, ","))
+		filter = append(filter, a.matcherGroups[0])
```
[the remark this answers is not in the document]
pgier: Done

cli/format/format.go
```diff
@@ -36,14 +36,18 @@ func InitFormatFlags(app *kingpin.Application) {
...
-	FormatConfig(*client.ServerStatus) error
+	FormatSilences([]models.GettableSilence) error
+	FormatAlerts([]*models.GettableAlert) error
+	FormatConfig(*models.AlertmanagerStatus) error
 }
 
 // Formatters is a map of cli argument names to formatter interface object.
 var Formatters = map[string]Formatter{}
 
+func FormatDateTime(input strfmt.DateTime) string {
```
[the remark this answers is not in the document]
pgier: Done

cli/silence_add.go
```diff
@@ -149,24 +149,26 @@ func (c *silenceAddCmd) add(ctx context.Context, _ *kingpin.ParseContext) error
...
-	silenceID, err := silenceAPI.Set(ctx, silence)
+	amclient := NewAlertmanagerClient(alertmanagerURL)
+
+	postOk, err := amclient.Silence.PostSilences(silenceParams)
 	if err != nil {
 		return err
 	}
-
-	_, err = fmt.Println(silenceID)
+	fmt.Println(*postOk.Payload)
```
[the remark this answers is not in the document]
pgier: Done

cli/alert_query.go
```diff
@@ -87,19 +85,25 @@ func (a *alertQueryCmd) queryAlerts(ctx context.Context, _ *kingpin.ParseContext
...
-	}
-	alertAPI := client.NewAlertAPI(c)
 	// If no selector was passed, default to showing active alerts.
 	if !a.silenced && !a.inhibited && !a.active && !a.unprocessed {
 		a.active = true
 	}
-	fetchedAlerts, err := alertAPI.List(ctx, filterString, a.receiver, a.silenced, a.inhibited, a.active, a.unprocessed)
+
+	alertParams := alert.NewGetAlertsParams().WithContext(ctx)
+	alertParams.Active = &a.active
```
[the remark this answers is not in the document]
pgier: Done

cli/silence_query.go
pgier: Done

cli/silence_import.go
pgier: This seems to work similarly to v1 AFAICT.  In both cases if the silence ID does not exist, the API returns a 400 with "not found" in the error message.

simonpasquier: Digging more into the code I think we can refine the API here. The handler is returning any error from the silences as-is but it could intercept the `ErrNotFound` error and turn it into a 404 response. It would be cleaner for the client as it would only have to check the status code and not rely on the fact that the returned body contains the `not found` string. @mxinden @stuartnelson3 what do you think?

https://github.com/prometheus/alertmanager/blob/a94088406ab7212acadaa01f4e86467ee1eac8f6/api/v2/api.go#L660-L664

https://github.com/prometheus/alertmanager/blob/a94088406ab7212acadaa01f4e86467ee1eac8f6/silence/silence.go#L502-L511

cli/silence_import.go
pgier: I did a bit more testing around this and basic backwards compatibility seems to work ok.  I exported silences using the v1 amtool, and then imported the file using the v2 amtool.  I'll also add a note to the changelog about possible incompatibility.

cli/silence_import.go
stuartnelson3: It would be nice to avoid the string comparison

mxinden: @simonpasquier sounds good to me.

pgier: Created #1816 to add the 404 response to the v2 rest API.

pgier: Rebased to get the changes from #1816 and updated the cli code to check the error type instead of doing the string compare.

cli/utils.go
```diff
@@ -106,17 +105,21 @@ func loadAlertmanagerConfig(ctx context.Context, alertmanagerURL *url.URL, confi
 		return cfg, nil
 	}
 	if alertmanagerURL != nil {
-		status, err := getRemoteAlertmanagerConfigStatus(ctx, alertmanagerURL)
+		configStatus, err := getRemoteAlertmanagerConfigStatus(ctx, alertmanagerURL)
 		if err != nil {
 			return nil, err
 		}
-		return status.ConfigJSON, nil
+		conf, err := config.Load(*configStatus.Config.Original)
```
simonpasquier: it could be `return config.Load(*configStatus.Config.Original)` here.

cli/utils.go
```diff
@@ -106,17 +105,21 @@ func loadAlertmanagerConfig(ctx context.Context, alertmanagerURL *url.URL, confi
 		return cfg, nil
 	}
 	if alertmanagerURL != nil {
```
simonpasquier: not directly related to this PR but if you feel it, this would be a more natural flow:

```go
if alertmanagerURL == nil {
  return nil, errors.New("failed to get Alertmanager configuration")
}
...
```

cli/utils.go
pgier: Done!

cli/utils.go
```diff
@@ -21,13 +21,13 @@ import (
 	"os"
 	"path"
 
-	"github.com/prometheus/alertmanager/client"
+	"github.com/prometheus/alertmanager/api/v2/client/general"
+	"github.com/prometheus/alertmanager/api/v2/models"
+	"github.com/prometheus/alertmanager/config"
```
simonpasquier: `github.com/prometheus/alertmanager/config` is imported twice (see line below)...

pgier: Fixed!

PR #1799, opened 2019-03-15
fix golint error

yasongxu:
No functional change, fix golint error in 3 files

1. struct field generatorUrl should be generatorURL
2. if block ends with a return statement, so drop this else and outdent its block


simonpasquier:
You need to sign your commit to comply with the DCO check.

PR #1817, opened 2019-03-29
ui/web: Set HTTP headers to prevent asset caching

mxinden:
This patch adds HTTP headers to each web asset response, disabling
caching for the payload. This prevents users running an old version of
the UI against a new version of Alertmanager once they updated. On the
down-side it increases page load size on subsequent page loads.

What do you all think?

### DIFF
--- a/ui/web.go
+++ b/ui/web.go
@@ -31,24 +31,32 @@ func Register(r *route.Router, reloadCh chan<- chan error, logger log.Logger) {
 	r.Get("/metrics", promhttp.Handler().ServeHTTP)
 
 	r.Get("/", func(w http.ResponseWriter, req *http.Request) {
+		disableCaching(w)
+
 		req.URL.Path = "/static/"
 		fs := http.FileServer(asset.Assets)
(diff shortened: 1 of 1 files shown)


brian-brazil:
Are etags enabled so that we don't refetch the whole scripe if we don't have to?

stuartnelson3:
or could we generate a unique string at compile time, and version `script.js` with that?

mxinden:
While etags or adding a unique opaque string to `script.js` is a valid option, I would prefer going with the simplest solution for now. `script.js` is 90kb large, thereby I don't think reloading is a big deal. Once someone complains about high load times we can further iterate on it.

What do you think @brian-brazil @stuartnelson3?

stuartnelson3:
I'm fine with that. I actually have caching disabled and didn't notice. Total cache-able payload seems to be ~350kb, I'm guessing folks can live with that.

mxinden:
If there are no objections by @brian-brazil I will merge this patch tomorrow.

brian-brazil:
No objections from my side.

PR #1822, opened 2019-04-03
notify: redact Slack webhook URL from logs

simonpasquier:
I discovered that the Slack URL is logged in clear-text when an error occurs. This PR makes sure that the information isn't leaked anymore.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -585,9 +585,10 @@ func (n *Slack) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 		return false, err
 	}
 
-	resp, err := post(ctx, c, n.conf.APIURL.String(), contentTypeJSON, &buf)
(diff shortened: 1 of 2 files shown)


brian-brazil:
This classifies as a security bug per https://prometheus.io/docs/operating/security/#secrets.

mxinden:
I can cut `v1.6.2` tomorrow. Thoughts by others?

brian-brazil:
Are there any other notifiers that could have the same issue?

simonpasquier:
> Are there any other notifiers that could have the same issue?

The Slack notifier is the only one making use of a [`config.SecretURL`](https://github.com/prometheus/alertmanager/blob/master/config/config.go#L122) but I missed that the Wechat notifier is only passing the token in the URL... I'll send another PR to fix it.

PR #1825, opened 2019-04-04
notify: redact more secret data from logs

simonpasquier:
Follow-up of https://github.com/prometheus/alertmanager/pull/1822

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -194,11 +194,16 @@ type PagerDuty struct {
 	conf   *config.PagerdutyConfig
 	tmpl   *template.Template
 	logger log.Logger
+	apiV1  string // for tests.
 }
 
 // NewPagerDuty returns a new PagerDuty notifier.
 func NewPagerDuty(c *config.PagerdutyConfig, t *template.Template, l log.Logger) *PagerDuty {
-	return &PagerDuty{conf: c, tmpl: t, logger: l}
+	n := &PagerDuty{conf: c, tmpl: t, logger: l}
+	if c.ServiceKey != "" {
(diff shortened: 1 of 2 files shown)


notify/impl.go
```diff
@@ -1210,8 +1210,7 @@ func (n *Pushover) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 		return false, err
 	}
 
-	apiURL := "https://api.pushover.net/1/messages.json"
-	u, err := url.Parse(apiURL)
+	u, err := url.Parse(n.apiURL)
```
brian-brazil: We've using this on line 1218 too

simonpasquier: Thanks!

notify/impl_test.go
```diff
@@ -74,6 +74,73 @@ func TestPagerDutyRetryV2(t *testing.T) {
...
+		&config.PagerdutyConfig{
+			ServiceKey: config.Secret(key),
+			HTTPConfig: &commoncfg.HTTPClientConfig{},
+		},
+		createTmpl(t),
+		log.NewNopLogger(),
+	)
+	notifier.apiV1 = srv.URL
+
+	ok, err := notifier.Notify(ctx, []*types.Alert{}...)
```
mxinden: This seems to be the same for each added test. Could this be a table-driven test with [subtests](https://blog.golang.org/subtests) instead? 

Let me know if you want to delay it, as it is not necessary for the security patch release.

simonpasquier: Rather than table-driven tests (to keep each test focused on one receiver), I'll create a helper function.

mxinden: Sounds good, thanks!

simonpasquier: Done!

PR #1830, opened 2019-04-10
test/cli: add basic amtool cli tests

pgier:
Working on some acceptance tests for amtool

### DIFF
--- a/test/cli/acceptance.go
+++ b/test/cli/acceptance.go
@@ -1,0 +1,665 @@
+// Copyright 2019 Prometheus Team
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


simonpasquier:
On the first look, it seems ok. I'm pondering whether it would make sense to have these tests live with the other acceptance tests in `test/with_api_v2` as some parts of the code are already common.

pgier:
Originally, I was thinking that these tests would verify backwards compatibility in amtool while making changes like the recent update from v1 to v2.  So for future updates, e.g. if there was ever an api v3, ideally the amtool tests would stay the same and would provide an automated compatibility check.  Although I'm not sure how practical this would be.

stuartnelson3:
any more thoughts on this?

mxinden:
Ping @simonpasquier @pgier.

pgier:
Rebased onto latest master, fixed a merge conflict, and made some minor simplifications.
@simonpasquier can you take a look?

PR #1837, opened 2019-04-15
Ability to set dedup_key manually via config.

1992kk:
dedup_key can be driven from configs as well.

### DIFF
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -317,7 +317,7 @@ func (n *PagerDuty) notifyV2(
 		ClientURL:   tmpl(n.conf.ClientURL),
 		RoutingKey:  tmpl(string(n.conf.RoutingKey)),
 		EventAction: eventType,
-		DedupKey:    hashKey(key),
+		DedupKey:    tmpl(n.conf.DedupKey),
 		Images:      make([]pagerDutyImage, len(n.conf.Images)),
 		Links:       make([]pagerDutyLink, len(n.conf.Links)),
 		Payload: &pagerDutyPayload{
--- a/notify/impl.go
+++ b/notify/impl.go
@@ -317,7 +317,7 @@ func (n *PagerDuty) notifyV2(
 		ClientURL:   tmpl(n.conf.ClientURL),
 		RoutingKey:  tmpl(string(n.conf.RoutingKey)),
 		EventAction: eventType,
-		DedupKey:    tmpl(n.conf.DedupKey),
+		DedupKey:    hashkey(tmpl(string(n.conf.DedupKey))),
 		Images:      make([]pagerDutyImage, len(n.conf.Images)),
 		Links:       make([]pagerDutyLink, len(n.conf.Links)),
 		Payload: &pagerDutyPayload{


brian-brazil:
There's no need for this to be configurable, we already set it to the correct value based on the group.

1992kk:
This is related to https://github.com/prometheus/alertmanager/issues/1587. 

> When triggering multiple alerts via the API and those alerts get grouped, only the alert summary gets updated, and subsequent messages get logged. The alert details are not updated (it still reflects the original message and summary). At this time, you have to click through the detailed log to see each individual API message sent from AlertManager. 

To workaround this problem, we can add the timestamp of one of the incidents in the description for that batch of alerts which will be creating an incident on Pagerduty. If any new alerts come up, they will be grouped under the same bunch in alertmanager but their notification will use a different dedup_key to create a new incident on pagerduty.

> global:
>   resolve_timeout: 60m
> 
> route:
>   group_by: ['alertname']
>   group_wait: 60s
>   group_interval: 60s
>   repeat_interval: 1h
>   receiver: 'Beta.Pager'
> 
> receivers:
> - name: 'Beta.Pager'
>   pagerduty_configs:
>   - service_key: 'xxxxxxxxxxxxxxx'
>     description: "[{{ .Status | toUpper }}{{ if eq .Status \"firing\" }}:{{ .Alerts.Firing | len }}{{ end }}] {{ .GroupLabels.SortedPairs.Values | join \" \" }} {{ if gt (len .CommonLabels) (len .GroupLabels) }}({{ with .CommonLabels.Remove .GroupLabels.Names }}{{ .Values | join \" \" }}{{ end }}){{ end }} : {{- if .CommonAnnotations.time -}}{{- .CommonAnnotations.time -}}{{- else -}}{{- with index .Alerts 0 -}}{{- .Annotations.time -}}{{- end -}}{{- end -}}"
>     dedup_key: "[{{ .Status | toUpper }}{{ if eq .Status \"firing\" }}:{{ .Alerts.Firing | len }}{{ end }}] {{ .GroupLabels.SortedPairs.Values | join \" \" }} {{ if gt (len .CommonLabels) (len .GroupLabels) }}({{ with .CommonLabels.Remove .GroupLabels.Names }}{{ .Values | join \" \" }}{{ end }}){{ end }} : {{- if .CommonAnnotations.time -}}{{- .CommonAnnotations.time -}}{{- else -}}{{- with index .Alerts 0 -}}{{- .Annotations.time -}}{{- end -}}{{- end -}}"

Idea is to achieve grouping and batching of notifications on Pager. 

Please suggest if there is any other idea to achieve this.

simonpasquier:
Someone from PagerDuty [commented](https://github.com/prometheus/alertmanager/issues/1587#issuecomment-446304666) on this issue that they were looking at improvements on their side. I suggest you check with them what's the progress...

1992kk:
> Someone from PagerDuty [commented](https://github.com/prometheus/alertmanager/issues/1587#issuecomment-446304666) on this issue that they were looking at improvements on their side. I suggest you check with them what's the progress...

Is this a big change at Alertmanager end?

simonpasquier:
My understanding is that Pager Duty should accommodate with how AlertManager works today. But I'd love to get an update from them on the GitHub issue...

1992kk:
Thanks @simonpasquier .  

Considering that this is a new feature request for Pagerduty, I'm assuming it'll not be a quick process for getting this implemented in Pagerduty. 

I was hoping if we can have control over the dedup_key via alertmanager, so we can solve/workaround this problem.

crockk:
When triggering an event with the [Events V2 API](https://developer.pagerduty.com/docs/events-api-v2/trigger-events/), you can specify `dedup_key` in the payload. I believe this should be configurable in the Alertmanager Pagerduty config. This is applicable for my use case, as we are trying to open alerts in PD using Alertmanager, and resolve them using Ansible playbooks. If the Ansible playbook fails to resolve it, we know we must take action.

SuperQ:

PR #1846, opened 2019-04-22
Use github.com/oklog/run not archived oklog/oklog

metalmatze:
https://github.com/oklog/oklog was archived and the run group package was extracted as https://github.com/oklog/run. We should use this instead for possible future updates.

Similar to https://github.com/prometheus/prometheus/pull/5491.


mxinden:
Thanks @metalmatze!

PR #1861, opened 2019-04-26
*: enable default linters

simonpasquier:


notify/impl.go
```diff
@@ -1273,6 +1269,7 @@ func tmplHTML(tmpl *template.Template, data *template.Data, err *error) func(str
 // maximum length requirements on deduplication keys.
 func hashKey(s string) string {
 	h := sha256.New()
+	//nolint: errcheck
```
mxinden: Can you add a comment that `Hash.Write` never returns an error with a link to https://golang.org/pkg/hash/#Hash ? Or is this common knowledge for everyone else?

simonpasquier: Sure, done!

mxinden:
Thanks @simonpasquier.

Issue #1818, opened 2019-03-29
opsgenie: Move from not supported  `team` field  to `responders` for alert routing.


bwplotka:
* Alertmanager version:

	`v0.16.1`

Essentially we used to configure Opsgenie receiver with a `Team` string: https://github.com/prometheus/alertmanager/blob/master/config/notifiers.go#L451

And this is recommended (and only) way to specify the exact Opsgenie team in Alertmananger's receiver: [example](https://www.robustperception.io/using-opsgenie-with-the-alertmanager)

However, teams are not identified by `team`  names in Opsgenie (maybe they used to?)

They are identified by `responder` field: https://docs.opsgenie.com/docs/alert-api#section-create-alert

Which can contain `name` or `id` of the team: https://docs.opsgenie.com/docs/team-api#section-get-team

If you look closely there is literally NO `team` field anymore. So this field is most likely not supported (is used to be?) Is it ignored? It's definitely not documented. I think it works? kind of? looks like it's meant to be deprecated.

We recently had an incident where team renames caused Opsgenie routing to ignore alerts, so we assume it works it's just really bad as `team string` is now display name only. 

IMO this defines Acceptance Criteria: (edited)

* AM user should be able to use official way of referencing exact team for Opsgenie Create alert API: Rename sent field to`responders` from `teams`.
* Allow to reference teams by IDs not only names.

Would you accept PR to fix this?

ivan-valkov:
It looks like we allow for the proper responders format - https://github.com/prometheus/alertmanager/blob/89b0ff05743dbedfdfa8d2433c58ba4d5ddf7202/notify/impl.go#L964

But the fields is still called `teams` not `responders`:
https://github.com/prometheus/alertmanager/blob/89b0ff05743dbedfdfa8d2433c58ba4d5ddf7202/notify/impl.go#L880

simonpasquier:
It seems this was overlooked when switching from the OpsGenie v1 API to v2 (#1061) and it somehow works by accident:
https://docs.opsgenie.com/docs/migration-guide-for-alert-rest-api#section-4-responders-field

simonpasquier:
Also related to #1342

bwplotka:
I am happy to fix this if you are ok with this @simonpasquier

simonpasquier:
@bwplotka sure go ahead. You're more than qualified, especially as an OpsGenie user :smile:

arnarg:
Is there any progress on this?

[...]

Issue #1342, opened 2018-04-25
Specify OpsGenie responders


zemek:
It looks like right now we can only set `teams:` in the OpsGenie config. The OpsGenie API says that the `responders` field supports "Teams, users, escalations and schedules" https://docs.opsgenie.com/docs/alert-api

It would be useful to be able to send alerts to a specific schedule within a team, rather than just team itself

pasali:
+1 for this feature.

simonpasquier:
Comparing the OpsGenie documentation and the AlertManager code, it seems that the OpsGenie API has changed a bit (again?).

The [documentation](https://docs.opsgenie.com/docs/alert-api#section-create-alert) states that the JSON object for creating an alert can contain a `responders` field (and the `teams` field is undocumented):

```json
{
    ...
    "responders":[
        {
            "name":"NOC",
            "type":"team"
        },
        {
            "username":"trinity@opsgenie.com",
            "type":"user"
        },
        {
            "name":"Nightwatch Escalation",
            "type":"escalation"
        },
        {
            "name":"First Responders Schedule",
            "type":"schedule"
        }
    ]
}
```

The AlertManager code uses the `teams` field though:

https://github.com/prometheus/alertmanager/blob/ec2cc57d28b66590b2429b03fdba068eae32b17a/notify/impl.go#L985-L995

@iksaif @Tom-Fawcett or anybody with OpsGenie access wanting to have a look?

zemek:
we might submit a PR for this at the end of the month or so if no one has started working on this

vitaliiko:
We need this feature as well

bwplotka:
+1 happy to make a PR

bwplotka:
unless @zemek you have something

zemek:
@bwplotka been too busy so haven't actually done anything on this :( feel free to make a PR!

PR #1863, opened 2019-04-26
opsgenie: Moved from deprecated, non documented teams to responders field.

bwplotka:
Fixes: https://github.com/prometheus/alertmanager/issues/1818
Fixes: https://github.com/prometheus/alertmanager/issues/1342

Some decisions I made:
* "Teams" config option will fail on unmarshalling the config. I marked it as deprecated. AM is 0.x version it should be fine (?)
* I allow the use case for dynamic set of responders from alert's label using templating. Seems like it fits into how others fields are treated.
* I allow for full flexibility, so you can use any type you want. There is no validation. I don't think it is worth as Opsgenie changes those things (?). Let me know if it's needed. I think if the type is wrong and create alert will fail it's fine as "validation"

Signed-off-by: Bartek Plotka <bwplotka@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@
 !.golangci.yml
 !/cli/testdata/*.yml
 !/cli/config/testdata/*.yml
+!/config/testdata/*.yml
 !/doc/examples/simple.yml
 !/circle.yml
 !/.travis.yml
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -15,8 +15,11 @@ package config
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 	"time"
+
+	"github.com/pkg/errors"
 
 	commoncfg "github.com/prometheus/common/config"
 )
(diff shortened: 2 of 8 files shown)


bwplotka:
Will fix tests tomorrow.

bwplotka:
Tests should be fixed.

.gitignore
```diff
@@ -9,7 +9,7 @@
 
 !.golangci.yml
 !/cli/testdata/*.yml
-!/cli/config/testdata/*.yml
+!/config/testdata/*.yml
```
simonpasquier: `!/cli/config/testdata/*.yml` should be kept AFAICT.

config/notifiers.go
```diff
@@ -459,17 +460,38 @@ type OpsGenieConfig struct {
...
+	if c.Teams != "" {
+		// Should we allow this for backward compatibility? Should we fill responders from this?
+		return errors.New("teams field is deprecated in Opsgenie config. Use responders instead.")
+	}
+
+	return nil
+}
+
+type OpsGenieConfigResponder struct {
+	// One of those 3 should be filled.
```
simonpasquier: Can we add a check for this when unmarshaling the struct and return an error if necessary?

config/notifiers.go
```diff
@@ -459,17 +460,38 @@ type OpsGenieConfig struct {
...
 func (c *OpsGenieConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	*c = DefaultOpsGenieConfig
 	type plain OpsGenieConfig
-	return unmarshal((*plain)(c))
+	if err := unmarshal((*plain)(c)); err != nil {
+		return err
+	}
+
+	if c.Teams != "" {
+		// Should we allow this for backward compatibility? Should we fill responders from this?
```
simonpasquier: Given that we can't propagate the deprecation to the user in the logs, I think that a hard failure is preferred.

config/notifiers.go
```diff
@@ -459,17 +460,38 @@ type OpsGenieConfig struct {
 	Description string            `yaml:"description,omitempty" json:"description,omitempty"`
 	Source      string            `yaml:"source,omitempty" json:"source,omitempty"`
 	Details     map[string]string `yaml:"details,omitempty" json:"details,omitempty"`
-	Teams       string            `yaml:"teams,omitempty" json:"teams,omitempty"`
-	Tags        string            `yaml:"tags,omitempty" json:"tags,omitempty"`
-	Note        string            `yaml:"note,omitempty" json:"note,omitempty"`
-	Priority    string            `yaml:"priority,omitempty" json:"priority,omitempty"`
+	// Teams field is deprecated.
```
simonpasquier: No strong feeling about this one but I'd rather remove the field as part of the change.

.gitignore
bwplotka: Ok (:

config/notifiers.go
bwplotka: Up to you guys (: The problem is that we will not fail the config unmarshalling if people will be still on `Teams` (and they will be). Thoughts?

simonpasquier: The unmarshaling should still fail as we use [`yaml.UnmarshalStrict()`](https://godoc.org/gopkg.in/yaml.v2#UnmarshalStrict).

bwplotka:
Thanks for initial review @simonpasquier , PTAL

simonpasquier:
@bwplotka you would need to resolve the conflicts...

bwplotka:
done (:

bwplotka:
Is this build flake? After changelog rebase suddnely tests are red

simonpasquier:
Yes, it looks like a flake, I've re-run the workflow.

hrancic:
A question, if you would be so kind. Will the yml configuration file still work with alertmanager 0.18.0  if teams are set up according to https://prometheus.io/docs/alerting/configuration/#opsgenie_config ?

```
# Comma separated list of team responsible for notifications.
[ teams: <tmpl_string> ]
```

or should it be changed to something like:

```
responders:
     name: <tmpl_string>
     type: team
```

?

simonpasquier:
@hrancic the documentation should be updated (and fixed) in a few minutes.

hrancic:
Another question. Our current configurations look something like this:

```teams: <tmpl_string>```

these templates (which have some conditional logic) are creating a comma separated lists of teams. We would like to upgrade our alertmanager version. How do you propose we migrate to the new version?

**Comment by bwplotka** [comment]
@hrancic do you mean how to specify multiple teams?
```
responders:
- name: "teamA"
  type: team
- name: "teamB"
  type: team
- name: "teamC"
  type: team
(...)
```

**Comment by hrancic** [comment]
yes, but writing them one by one is out of the question. Will `responders: <tmpl_string>` work? Where this template creates:

```  
- name: "teamA" 
  type: team
- name: "teamB"
  type: team
- name: "teamC"
  type: team
(...)
 ```

bwplotka:
> yes, but writing them one by one is out of the question

Why?

> Will responders: <tmpl_string> work?

It will not work in current version. (:

hrancic:
We don't know who responders are before the alert is created. We create an alert and then decide its destination while sending it. `teams: <tmpl_string>` was a model which enabled us to determine the teams after the alert is fired but before sending it to opsgenie. We would like to have something like this:

``` 
responders:
 - names : <tmpl_string>
   type: teams
```

bwplotka:
From my side that is a fair request. Do you mind adding an Github issue with those details on AM? We can improve this indeed. (:

anandsinghkunwar:
@hrancic Was an issue filed?

PR #1864, opened 2019-04-27
Default grouping in ui

### LINKED ISSUE #868: Display "natural" grouping (and make groups linkable)

w0rm:
custom-grouping

Show alert groups from the alertmanager config by default and hide the custom grouping behind the checkbox.
 
* If `customGrouping` is present in the url, then fetch alerts from the `/alerts` endpoint and group on the frontend
* Otherwise (this is now the default) show alert groups from the `/alerts/groups` endpoint

closes #868


w0rm:
@stuartnelson3 I'm not sure why the circle ci tests are failing.

stuartnelson3:
Not sure, will take a look at it soon (probably Monday)

thanks for doing this!

w0rm:
@beorn7 does it make you happy? :D

api/v2/openapi.yaml
```diff
@@ -475,6 +475,10 @@ definitions:
         type: array
         items:
           $ref: '#/definitions/gettableAlert'
+    required:
+      - labels
+      - receiver
+      - alerts
```
stuartnelson3: thanks for fixing the openapi definition!

ui/app/src/Utils/Filter.elm
```diff
@@ -22,6 +22,7 @@ import Url exposing (percentEncode)
 type alias Filter =
     { text : Maybe String
     , group : Maybe String
+    , customGrouping : Bool
```
stuartnelson3: Whenever I look at boolean flags/query strings, I wonder if it would be better to have a concrete setting, like `grouping=default` or `grouping=custom`. It helps keep things from ballooning to ever larger sizes. For example, I'm not super thrilled with how I introduced `silenced=true` and `inhibited=true`, but I wasn't sure if it would be better or worse to have `alert_state=silenced|inhibited`.

Any thoughts on this? Stay with the boolean flag, or move use a `groups` query string? @mxinden @simonpasquier @w0rm

w0rm: @stuartnelson3 I think the missing parameter should be the default. Because this is how you open it anyways. So `grouping=default` is not really needed.

beorn7:
UI-wise, this looks good to me. Thank you very much. I haven't looked at the code.

One feature request (which might or might not already work) is to be able to deep-link one group (with the idea to link back to that group from a notification rather than to all alerts of the receiver, as we do it now). If that's not already possible, it could also be added in a later PR (perhaps backend changes are necessary). Just saying it so that we keep it in mind.

stuartnelson3:
>One feature request (which might or might not already work) is to be able to deep-link one group (with the idea to link back to that group from a notification rather than to all alerts of the receiver, as we do it now). If that's not already possible, it could also be added in a later PR (perhaps backend changes are necessary). Just saying it so that we keep it in mind.

Yeah, we should be able to add the alert groups fingerprint to the response and "link" to it that way. There would be a few backend changes, nothing major, but enough to warrant a new pr. thanks for the reminder!

PR #1865, opened 2019-04-29
Added documentation on docker usage

DrPyser:
This is an invitation to add more documentation related to docker usage.

A new `Docker` section is added to the readme, with subsections for describing exposed ports, volumes, environment variables and commands.

I only added the information I could gather from reading the dockerfile(https://github.com/prometheus/alertmanager/blob/master/Dockerfile). I invite contributors, members and people with better knowledge and experience on this software to help add relevant information relative to usage of the docker image.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,32 @@ You can also build just one of the binaries in this repo by passing a name to th
 $ make build BINARIES=amtool
 ```
 
+## Docker
+
+<!--- example of complete docker run command to run a container --->
+<!--- example of a docker-compose.yml service snippet --->
+
+### Ports
+
+* `9093` : main api port
+
+### Volumes
+
+* `/alertmanager`: storage directory
+
+These paths can also be mapped to volumes:
+* `/etc/alertmanager/alertmanager.yml`: default configuration file path
+
+### Environment variables
+
+<!--- Add documentation here --->
+
+### Command and entrypoint
+The container entrypoint is the [`alertmanager` command](https://prometheus.io/docs/alerting/configuration/)<!---More complete documentation of the command would be useful here--->. 
+
+The default docker command passed to the entrypoint is `--config.file=/etc/alertmanager/alertmanager.yml --storage.path=/alertmanager`. Specifying the docker command(using `docker run quay.io/prometheus/alertmanager $COMMAND`) thus override those default arguments.
+
+
 ## Example
 
 This is an example configuration that should cover most relevant aspects of the new YAML configuration format. The full documentation of the configuration can be found [here](https://prometheus.io/docs/alerting/configuration/).


README.md
```diff
@@ -50,6 +50,32 @@ You can also build just one of the binaries in this repo by passing a name to th
 $ make build BINARIES=amtool
 ```
 
+## Docker
+
+<!--- example of complete docker run command to run a container --->
```
mxinden: How about linking to the `Procfile` of this repository, as it gives a user a full setup on their local machine.

mxinden: https://github.com/prometheus/alertmanager/blob/master/Procfile

README.md
```diff
@@ -50,6 +50,32 @@ You can also build just one of the binaries in this repo by passing a name to th
...
+<!--- example of a docker-compose.yml service snippet --->
+
+### Ports
+
+* `9093` : main api port
```
mxinden:
```suggestion
* `9093`: main api port
* `9094`: cluster port 
```

README.md
```diff
@@ -50,6 +50,32 @@ You can also build just one of the binaries in this repo by passing a name to th
...
+
+### Volumes
+
+* `/alertmanager`: storage directory
+
+These paths can also be mapped to volumes:
```
mxinden: I am a bit confused by this comment as `/alertmanager` can also be mapped to volumes.

README.md
```diff
@@ -50,6 +50,32 @@ You can also build just one of the binaries in this repo by passing a name to th
...
+* `/etc/alertmanager/alertmanager.yml`: default configuration file path
+
+### Environment variables
```
mxinden: I am not aware of any. Configuration is limited to the config file and the command line flags.

README.md
DrPyser: I'm talking about an actual invokation of `docker run ...`, using volumes(`-v`) and port(`-p`) mapping as appropriate, then the command arguments. The `Procfile` contains example invokations of the `alertmanager` binary, but it will look different when using `docker run`, so it can be confusing to people less familiar with docker and unix.

README.md
DrPyser: `/alertmanager` is declared as a volume in the Dockerfile, so will always be a volume. But any file or directory can be mounted as a volume. I thought a good idea to mention files and directories that people likely also want to mount as volumes. 

I guess the wording can be changed to something like 
```suggestion
You might also want to use volumes for these paths:
```

stuartnelson3:
other than the DCO, is there anything preventing this PR from being merged?

README.md
```diff
@@ -50,6 +50,33 @@ You can also build just one of the binaries in this repo by passing a name to th
 $ make build BINARIES=amtool
 ```
 
+## Docker
+
+<!--- example of complete docker run command to run a container --->
+<!--- example of a docker-compose.yml service snippet --->
+
+### Ports
```
simonpasquier: Exposed ports?

README.md
simonpasquier: Same for me. I'd remove this section.

README.md
```diff
@@ -50,6 +50,33 @@ You can also build just one of the binaries in this repo by passing a name to th
...
+* `/alertmanager`: storage directory
+
+These paths can also be mapped to volumes:
+* `/etc/alertmanager/alertmanager.yml`: default configuration file path
+
+### Environment variables
+
+<!--- Add documentation here --->
+
+### Command and entrypoint
```
simonpasquier: "Entrypoint" alone?

README.md
```diff
@@ -50,6 +50,33 @@ You can also build just one of the binaries in this repo by passing a name to th
...
+The container entrypoint is the [`alertmanager` command](https://prometheus.io/docs/alerting/configuration/)<!---More complete documentation of the command would be useful here--->. 
+
+The default docker command passed to the entrypoint is `--config.file=/etc/alertmanager/alertmanager.yml --storage.path=/alertmanager`. Specifying the docker command(using `docker run quay.io/prometheus/alertmanager $COMMAND`) thus override those default arguments.
```
simonpasquier: Regarding this paragraph, I'm not sure we want to get into this level of details as 1) the arguments may change over time and 2) it is more a matter of the container runtime.

bploetz:
A lost afternoon and some Googling lead me here. I had been trying to mount the configuration for alertmanager via a volume, and the Dockerfile shown on Docker Hub (https://hub.docker.com/r/prom/alertmanager/dockerfile) shows the config file location used by alertmanager as `/etc/alertmanager/config.yml`. While trying to figure out why my config wasn't being honored I stumbled on this ticket to discover the config file location is actually `/etc/alertmanager/alertmanager.yml`.

Will the changes noted in this ticket also update the Dockerfile shown on Docker Hub?

README.md
DrPyser: Fair enough, we could link to a documentation page for the binary.

simonpasquier:
@bploetz the inaccuracy on Docker Hub has already been reported but we need someone from Docker Hub to clean up things (see https://github.com/docker/hub-feedback/issues/1847).

fredericrous:
the issue on docker hub does not prevent this project to get a proper readme on github documenting the docker image.
Having not found a documentation on docker, people turn to github.

SuperQ:

Issue #1868, opened 2019-04-29
Filter query string not correctly sent or parsed


stuartnelson3:
i just noticed while working on a pr that my filtering wasn't working. after some looking, I realized that I was only getting the first item in the filter when parsing.

Previously, a filter would be split on commas like this:

filter: {alertname="DiskRunningFull"
filter: dev="sdb2"
filter: instance="example2"}
when parsed, results in:
matchers: [alertname="DiskRunningFull" dev="sdb2" instance="example2"]

After a recent change to the API spec, now we only have a single filter item (it's not split on commas):

filter: {alertname="DiskRunningFull", dev="sdb2", instance="example2"}
when parsed, results in:
matchers: [alertname="DiskRunningFull"]

relevant change:
https://github.com/prometheus/alertmanager/commit/bc373f562fcd4a765d59c6b44e8dfca4bcd2fe4c

Silence preview is also broken. The silence preview feature sends a DIFFERENT filter format. We need to send the same type of query to this endpoint, and we also need to send the right type of query.

filter: alertname=DiskRunningFull,dev=sdb2,instance=example2
matchers: [alertname="DiskRunningFull,dev"]

This is something we need to fix ASAP.

@simonpasquier @mxinden I'll look at this today

PR #1869, opened 2019-04-29
Correctly construct filter query string for api

stuartnelson3:
Updating the api in
https://github.com/prometheus/alertmanager/commit/bc373f562fcd4a765d59c6b44e8dfca4bcd2fe4c
resulted in the UI sending query strings for the
filter that no longer match what the backend
expects. This brings that in to line.

fixes https://github.com/prometheus/alertmanager/pull/1868

Question:
The `filter` query string that is visible to users does not match the filter string format expected by the server now that https://github.com/prometheus/alertmanager/commit/bc373f562fcd4a765d59c6b44e8dfca4bcd2fe4c has been merged. I've opted to keep the user-facing format the same in the event of users having bookmarked certain URLs with filters in them. I would like to do this for now and we can discuss what we want to do about that in the future, but this is up for discussion. While it's good to keep the user facing URL the same, in my opinion, it does lead to a more or less duplicated function for generating the query string. Maybe that's the lesser of two evils.

This also solves https://github.com/prometheus/alertmanager/issues/1731
image


simonpasquier:
LGTM after testing though I've got no clue whether the code is ok or not...

stuartnelson3:
going to go ahead and merge.

@mxinden have you had a chance to look at generating the elm client from our openapi spec? I remember at one time there were issues with it hardcoding the URL or something

mxinden:
> @mxinden have you had a chance to look at generating the elm client from our openapi spec? I remember at one time there were issues with it hardcoding the URL or something

@stuartnelson3 I tried back when we moved the UI to API v2. I remember not being able to configure the backend without changing _openapi.yaml_. As of today, the UI only uses the generated model, which in my eyes is the most important part.

PR #1871, opened 2019-04-30
test/with_api_v1: delete duplicate close of http body.

johncming:
delete duplicate close of http body.


johncming:
> looks good. can you sign the DCO? `git commit --amend -s` and force push to your branch

ok. done.

Issue #911, opened 2017-07-19
Improvement of the UI of the main alerts page for alertmanager


kpachhai:
At the moment, if we have lots of alerting rules and accordingly, if there are lots of alerts being fired off, the alerts page just shows a long list of all the alerts. It would be nice to improve the UI for this page so it's easy on the eyes. I was thinking more along the lines of if we could improve the UI so that we summarize different kinds of alerts on the first page, and then maybe it'll show how many servers/nodes were affected with that particular alert and any other summary that might be helpful. This would be in a list categorized by different kinds of alerts. Then, each row would have the link that would take you to another page that then shows all the details for that specific alert.

Basically, an overhaul of the main alertmanager page

stuartnelson3:
Thank you for your comment. Could you provide some more information on your use case, such as why you would like to see all alerts? Are you having, e.g., the same alert firing across hundreds of nodes?

The UI currently has a filtering and grouping feature that we hope would allow most users to "create" an alerts page that is relevant to them and be of a manageable size.

An idea that immediately comes to mind is collapsing the groupings and showing the metadata you're interested in, and then having a "click to expand" feature. This already exists, but on the individual alert level. We also need to be mindful of the needs of all users -- it sounds like you might have many more alerts than the average user, and one of our challenges is providing a single product for teams of very different sizes. The current setup might be working well for 90% of use cases, and introducing an extra step to seeing alerts could be annoying to them.

kpachhai:
I think my use case scenario would be the following:

I want to be able to have a page where on first glance, I can see only the alertname that is currently being fired(maybe one alertname per row). And in addition, where i can also see the total number of servers affected by this alerting rule. This comes in handy when we're watching hundreds of servers and somehow, majority of the servers are affected by this one alerting rule. At the moment, if this were to happen, it would be a long page of alerts with each server/node in each row. This becomes clunky when we also have some inhibition rules in place and it's very hard to navigate around.

Let me know if I was clear. If not, I'll try to post a graphic representation of what I'm talking about.

stuartnelson3:
Hm, I'm thinking of two different things:

1. Collapse the underlying alert information and just show the grouping (be it alertname, or whatever custom grouping a user creates), with information showing number of alerts affected by it.
[...]

PR #1876, opened 2019-05-03
Collapse alert groups when initially viewing ui

stuartnelson3:
image

helps with frontend rendering, and (in my opinion) makes it more usable. clicking the `+` expands the group, much like showing/hiding annotations for individual alerts.

requested in #911


w0rm:
As far as I understand this would only allow to expand one group. Can we allow expanding multiple groups? Can we also have a button to expand all groups?

stuartnelson3:
>As far as I understand this would only allow to expand one group. Can we allow expanding multiple groups?

We definitely can. I thought about this but did the simple thing first. Also, since we only allow a single alert's annotation information to be open at a time, I thought I could do that as well.

I'm guessing, for implementation, we would store a List of `Labels` and compare against that.

>Can we also have a button to expand all groups?

We definitely could. I'm wondering how often it would be used. But, if there are users/alertmanager installations running that have relatively few alerts/alert groups, maybe it would make sense to provide this.

Maybe you could give some suggestions on how to show this, and where to place it (I have very little UX skill..)

>If we add a button to expand all groups, can we keep this preference in the localStorage like the username?

+1

>If there is a single group, can we always expand it?

Makes sense

w0rm:
@stuartnelson3 let’s chat about this on Monday 

Some ideas:
Would it be cool to collapse groups with too many alerts and auto-expand groups with a few? 
Shall we show the number of alerts in a group?

beorn7:
Thanks for the PR and the discussion. Dynamically autoexpanding if a group has very few alerts and/or there are very few groups altogether would be awesome (cf. how GitHub renders notifications https://github.com/notifications , they even have partial expansion there, perhaps that would be interesting here as well, to see an example alert").

Would be great to see a number next to unexpanded groups ("44 alerts") or partially expanded groups ("41 more alerts, see more...")

stuartnelson3:
My plans for this PR:
- for a single group, expand it always
- support expanding multiple groups
- add a button to toggle expand/collapse all (user setting stored in localStorage)
- add alert count behind group labels

beorn7:
And of course, this PR doesn't have to solve everything. It's already a great start.

stuartnelson3:
@w0rm could you help a bit with the styling? in particular, the alert count seems to be on the "baseline" and should probably be bumped up higher, and the "expand all" button doesn't quite seem to be lined up correctly.

image

ui/app/index.html
```diff
@@ -19,12 +19,16 @@
             var app = Elm.Main.init({
                 flags: {
                     production: true,
-                    defaultCreator: localStorage.getItem('defaultCreator')
+                    defaultCreator: localStorage.getItem('defaultCreator'),
+                    groupExpandAll: JSON.parse(localStorage.getItem('groupExpandAll'))
```
stuartnelson3: After a lot of anger and confusion, I did some reading and figured out that booleans cannot be stored into local storage. Is this an ok way to get the boolean back out? Not sure if there's a convention around it.

ui/app/src/Views/AlertList/Updates.elm
```diff
@@ -26,20 +29,44 @@ update msg ({ groupBar, alerts, filterBar, receiverBar, alertGroups } as model)
...
+                            ( Success groups, newGroupBar )
+
+                        Initial ->
+                            ( Initial, groupBar )
+
+                        Loading ->
+                            ( Loading, groupBar )
+
+                        Failure e ->
+                            ( Failure e, groupBar )
```
stuartnelson3: This is sorting the alerts into alert groups immediately when they're returned from the API. I wasn't sure how to have a pattern match `Initial`, `Loading`, and `Failure e`, because they're technically of type `ApiData GettableAlerts`. If there's a way to compact this, that would be awesome.

w0rm: I think this is fine.

w0rm:
@stuartnelson3 I adjusted the styling:

Screenshot 2019-05-11 at 20 31 55

Issue #1882, opened 2019-05-09
PagerDuty notifier does not honor custom details section


vj396:
**What did you do?**
I tried to update the pd integration with custom details section.

**What did you expect to see?**
i wanted it to show me custom details section.

**What did you see instead? Under which circumstances?**
the default details section
**Environment**

* System information:
Linux 4.15.0-48-generic x86_64
* Alertmanager version:
```./alertmanager --version
alertmanager, version 0.17.0 (branch: master, revision: 6749f9faa9b87f4ca153481e8ab4332c89f0360b)
  build user:       user@somesystem
  build date:       20190509-21:17:53
  go version:       go1.12.5
```
* Prometheus version:

	insert output of `prometheus -version` here (if relevant to the issue)

* Alertmanager configuration file:
```
global:
  smtp_smarthost: '<redacted>'
  smtp_from: '<redacted>'
route:
  group_by: [<some list of tags>]
  group_wait: 30s
  group_interval: 1m
  repeat_interval: 1h
  receiver: test-receiver
  routes:
  - receiver: test-receiver
    continue: false
    match:
      param: 'value'
      severity: 'critical'
receivers:
- name: 'test-receiver'
  pagerduty_configs:
  - routing_key: '<redacted>'
    send_resolved: true
    client_url: '<redacted>'
    details:
      resolved: '{{ template "pagerduty.default.instances" .Alerts.Resolved }}'
      client_url: '{{ template "pagerduty.default.instances" .Alerts.Resovled }}'
```

* Prometheus configuration file:
```
insert configuration here (if relevant to the issue)
```

* Logs:
```
insert Prometheus and Alertmanager logs relevant to the issue here
```

vj396:
looks like i was doing it wrong.

gopiextrahop:
What was your fix? Could you share it?

PR #1883, opened 2019-05-09
PagerDuty custom details section fix

vj396:
This is a very rudimentary fix for the issue [https://github.com/prometheus/alertmanager/issues/1882](https://github.com/prometheus/alertmanager/issues/1882)


vj396:
looks like i was doing it wrong.

PR #1892, opened 2019-05-16
Use scheme and auth info if present from URL

### LINKED ISSUE #1891: [Regression] Scheme, basic auth in `amtool --alertmanager.url` no longer respected
**What did you do?**

```bash
amtool --alertmanager.url "https://user:password@alertmanager.example.com/" silence add ...
```

Exit code 0 indicating success.

```
amtool: error: Post http://alertmanager.example.com/api/v2/silences: dial tcp 1.2.3.4:80: connect: connection refused
```

Note the incorrect scheme (HTTP, not HTTPS). In addition there is no basic auth added, although that isn't obvious from the output.

**Environment**

Latest `amtool`. This behavior used to work as expected. It used to use all information present in the parsed URL including basic auth, if present, scheme and host.

**More information**

It seems as though https://github.com/prometheus/alertmanager/commit/74fa2236f75c41f849205d5d44c0f30548b7e479#diff-6ce4ff495858e55c738d815316b48cb4R65 adds the `NewAlertmanagerClient()` function which only copies across the `Host` field from the URL, and ignores the `User` (we use basic auth to protect access to our alert manager) as well as ignoring the scheme (we use HTTPS).

aeijdenberg:
Fixes #1891

### DIFF
--- a/cli/root.go
+++ b/cli/root.go
@@ -25,6 +25,8 @@ import (
 	"github.com/prometheus/alertmanager/api/v2/client"
 	"github.com/prometheus/alertmanager/cli/config"
 	"github.com/prometheus/alertmanager/cli/format"
+
+	clientruntime "github.com/go-openapi/runtime/client"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


aeijdenberg:
(Tests seem to be a somewhat flaky when running locally, so trying a force-push with no changes to re-trigger build)

PR #1893, opened 2019-05-18
Textarea with auto resize in UI

m-masataka:
I often use comment field in Silence Form Page with multiple lines of text.
I believe auto expand field is convenient for us.


stuartnelson3:
hey, sorry for the long delay in responding!

I like how the text area resizes in this PR, but the questions I'm thinking about currently:

- the text doesn't resize while I'm typing, only when focusing on the textarea. resizing while typing would probably also be nice
- if we make the change above, is this then something that we want? GitHub, for example (since I'm using their UI right now), auto-expands this comment box as I type, and it's a nice feature that I didn't actually even notice until looking at this PR. So maybe we should add it ..

do you have the time to implement resize-on-type? I think then we could look at the code and see if it makes sense to add this.

m-masataka:
@stuartnelson3 Thanks for your comment.
Now I implemented textarea resize on keypress. Does this match your comment of it?

ui/app/src/Utils/FormValidation.elm
```diff
@@ -28,19 +36,55 @@ fromResult result =
...
+
+updateTextAreaHeight : Int -> ValidatedField -> ValidatedField
+updateTextAreaHeight scrollHeight field =
+    let
+        rows =
+            ((toFloat scrollHeight - config.padding) / config.lineHeight)
+                |> ceiling
+                |> clamp config.minRows config.maxRows
+    in
+    { field | rows = rows }
```
w0rm: I see that there are two ways to update `rows`, the first way is using the number of lines in the value and the second way is using the `scrollHeight`. If the first way is working, we don't really need the second way. We don't even need to extend `ValidatedField` to store `rows`, we only need to introduce a view for the comment field that calculates the rows based on the number of lines in the value.

m-masataka: Thanks for your review.
That’s true.  Using scrollHeight is my old implementation of it. 
revised it

w0rm: @m-masataka sorry for not making this clear. Because `rows` can be calculated from the value, it doesn't need to be stored in model and can be done in the view function. The proposed change adds `rows` to every input, even though this is only used in the comment textarea.

What do you think of introducing a separate view function that would render the comment textarea and calculate `rows`?

m-masataka: @w0rm Sorry, I misunderstood.
> Because rows can be calculated from the value, it doesn't need to be stored in the model and can be done in the view function.

Exactly.

> The proposed change adds rows to every input, even though this is only used in the comment textarea.
> What do you think of introducing a separate view function that would render the comment textarea and calculate the rows attribute?

I see. I think textarea must be separated. 
Do you mean that create new function similar to ``validatedField`` ? 
If so, I'll revise that soon.

w0rm: @m-masataka yeah, exactly! Just a separate function similar to `validatedField` for the comment textarea. The `rows` attribute only makes sense for `textarea` we don't need it for all regular input fields.

ui/app/src/Utils/Views.elm
```diff
@@ -131,6 +134,56 @@ validatedField htmlField labelText classes inputMsg blurMsg field =
                 ]
 
 
+validatedTextareaField : (List (Attribute msg) -> List (Html msg) -> Html msg) -> String -> String -> (String -> msg) -> msg -> ValidatedField -> Html msg
+validatedTextareaField htmlField labelText classes inputMsg blurMsg field =
+    let
+        lineCount =
+            lines field.value
+                |> length
```
w0rm: `List` is always imported in Elm, so you can write `List.length` here and remove `import List exposing (length)`

ui/app/src/Utils/Views.elm
```diff
@@ -131,6 +134,56 @@ validatedField htmlField labelText classes inputMsg blurMsg field =
...
+                |> clamp 3 15
+    in
+    case field.validationState of
+        Valid ->
+            div [ class <| "d-flex flex-column form-group has-success " ++ classes ]
+                [ label [] [ strong [] [ text labelText ] ]
+                , htmlField
```
w0rm: There is no need in htmlField argument, because it is always textarea.

ui/app/src/Utils/Views.elm
```diff
@@ -131,6 +134,56 @@ validatedField htmlField labelText classes inputMsg blurMsg field =
                 ]
 
 
+validatedTextareaField : (List (Attribute msg) -> List (Html msg) -> Html msg) -> String -> String -> (String -> msg) -> msg -> ValidatedField -> Html msg
+validatedTextareaField htmlField labelText classes inputMsg blurMsg field =
+    let
+        lineCount =
+            lines field.value
```
w0rm: `String` is always imported, so you can write `String.lines` here and remove `import String exposing (lines)` from the top

m-masataka: Sorry, lack of elm knowledge.
Thank's for Review.

w0rm: No worries, thanks for working on this!

PR #1905, opened 2019-05-29
Use Circle CI's org context

simonpasquier:
Circle CI's [contexts](https://circleci.com/docs/2.0/configuration-reference/#context) allow to declare environment level variables at the org level instead of per project.


mxinden:
@simonpasquier would you mind adding some more details to this pull request?

simonpasquier:
Sure, I updated the description. The goal is to avoid configuring secrets per project but rather at the org level.

PR #1908, opened 2019-06-02
Support that send wechat textcard message to user or group

pulord:
- Use Environment Variable First if APISecret/CorpID not set
- Support new message type "textcard"
- Support than send message to Group

https://work.weixin.qq.com/api/doc#90000/90135/90243


config/config.go
```diff
@@ -376,17 +377,31 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 			}
 
 			if wcc.APISecret == "" {
-				if c.Global.WeChatAPISecret == "" {
-					return fmt.Errorf("no global Wechat ApiSecret set")
+				// Use Environment Variable First
+				apiSecret := Secret(os.Getenv("APISecret"))
```
simonpasquier: We don't support configuration by environment variables. It's an organization-wide policy (see https://github.com/prometheus/prometheus/issues/2357) so not specific to AlertManager.

notify/impl.go
```diff
@@ -789,17 +798,38 @@ func (n *Wechat) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
...
-		ToParty: tmpl(n.conf.ToParty),
-		Totag:   tmpl(n.conf.ToTag),
-		AgentID: tmpl(n.conf.AgentID),
-		Type:    "text",
-		Safe:    "0",
+		}
 	}
+
+	s, _ := json.Marshal(msg)
+	level.Debug(n.logger).Log("msg", "wechat Request", "body", s)
```
simonpasquier: We don't log JSON payload usually as it can be very verbose. And it duplicates the encoding already done at L838.

brian-brazil: May this also contain secrets?

config/config.go
pulord: I want to protect the apisecret, so use environment variables to store secrets。
I will fix it.

notify/impl.go
pulord: > May this also contain secrets?

Json payload only have post data, no secrets, the secrets used to get accesstoken from wechat.

config/config.go
pulord: > IIUC the user can't configure `ToAppChat` and any of `ToUser`, `ToParty` or `Totag` at the same time. If correct, it should be checked when loading the configuration. The same goes for `MessageType` which can only have 2 values from what I understand.
> 
> Do we want default values for the new attributes?
> 
> Is there an English version of the API documentation?

1.ToAppChat ToUser, ToParty or Totag can config at the same time, 
2.if ToAppChat is not empty, and ToUser/ToParty/Totag make no sense.
3. MessageType have two value "text" and "textcard", it Represents the representation of the message content, I will add default value "text". thanks

notify/impl.go
pulord: > We don't log JSON payload usually as it can be very verbose. And it duplicates the encoding already done at L838.

OK I delete it

config/config.go
simonpasquier: > if ToAppChat is not empty, and ToUser/ToParty/Totag make no sense.

In general, we prefer to prevent such cases as it could confuse users. I suggest that you expand the `(c *WechatConfig) UnmarshalYAML` method so it fails when someone configures ToAppChat with ` ToUser`, `ToParty` or `Totag`.

> MessageType have two value "text" and "textcard", it Represents the representation of the message content, I will add default value "text". thanks

Same here, if `MessageType` is something else than the expected values then it should fail.

template/default.tmpl
```diff
@@ -65,8 +65,8 @@ AlertmanagerUrl:
 {{ define "wechat.default.to_party" }}{{ end }}
 {{ define "wechat.default.to_tag" }}{{ end }}
 {{ define "wechat.default.agent_id" }}{{ end }}
-
-
+{{ define "wechat.default.agent_id" }}text{{ end }}
```
simonpasquier: this line is a duplicate of the previous one and you're missing a few others fields.

simonpasquier:
@pulord you'd need to run `make assets` and commit the changes to `asset/assets_vfsdata.go`.

pulord:
> @pulord you'd need to run `make assets` and commit the changes to `asset/assets_vfsdata.go`.

make assets error

`cd /Users/putaohong/Documents/work/git-code/go/src/github.com/prometheus/alertmanager/ui/app && /Library/Developer/CommandLineTools/usr/bin/make script.js
>> building elm-env docker image
The command '/bin/sh -c npm install -g   elm@0.19.0   elm-format@0.8.0   elm-test@0.19.0-beta5   uglify-js@3.4.7' returned a non-zero code: 1
make[1]: *** [elm-env] Error 1
make: *** [ui/app/script.js] Error 2`

simonpasquier:
Hmm, it works for me but I'm not on MacOS... Nevermind I'll push the `make assets` change on top of your PR once we're ok with it.

pulord:
> Hmm, it works for me but I'm not on MacOS... Nevermind I'll push the `make assets` change on top of your PR once we're ok with it.

I have fixed ! pls help fix the CI Error, thanks

template/default.tmpl
pulord: fixed

config/notifiers.go
```diff
@@ -109,9 +109,6 @@ var (
 			VSendResolved: false,
 		},
 		Message: `{{ template "wechat.default.message" . }}`,
-		ToUser:  `{{ template "wechat.default.to_user" . }}`,
```
simonpasquier: no need to remove those lines

config/notifiers.go
```diff
@@ -432,21 +429,40 @@ type WechatConfig struct {
...
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
 func (c *WechatConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	*c = DefaultWechatConfig
 	type plain WechatConfig
-	return unmarshal((*plain)(c))
+
+	if err := unmarshal((*plain)(c)); err != nil {
+		return nil
```
simonpasquier: return err

config/notifiers.go
```diff
@@ -432,21 +429,40 @@ type WechatConfig struct {
...
+	}
+	if c.ToAppChat != "" && (c.ToUser+c.ToParty+c.ToTag != "") {
```
simonpasquier: Please add unit tests for all cases.

config/notifiers.go
```diff
@@ -109,9 +109,6 @@ var (
 			VSendResolved: false,
 		},
 		Message: `{{ template "wechat.default.message" . }}`,
-		ToUser:  `{{ template "wechat.default.to_user" . }}`,
-		ToParty: `{{ template "wechat.default.to_party" . }}`,
-		ToTag:   `{{ template "wechat.default.to_tag" . }}`,
 		AgentID: `{{ template "wechat.default.agent_id" . }}`,
```
simonpasquier: Please add default values for the new fields (`MessageType`, `ToAppChat`, ...).

template/default.tmpl
```diff
@@ -65,8 +65,7 @@ AlertmanagerUrl:
 {{ define "wechat.default.to_party" }}{{ end }}
 {{ define "wechat.default.to_tag" }}{{ end }}
 {{ define "wechat.default.agent_id" }}{{ end }}
-
-
+{{ define "wechat.default.chat_id" }}text{{ end }}
```
simonpasquier: You can remove the `text` part. You should also add something for the other new fields.

config/notifiers.go
pulord: > no need to remove those lines

if don't remove default value。I can’t check the ToUser/ToTag/ToParty if them is empty!
unless  I use tmplText in UnmarshalYAML to get real value，it‘s not good。

on the other side,  I think there is no need to configure the template inside, it's better that all configurations come from the same file，What is your opinion?

config/notifiers.go
pulord: my mistake, sorry for that

config/notifiers.go
pulord: OK

config/notifiers.go
simonpasquier: Right, sorry for confusing you. Lets just check that `MessageType` matches the allowed values.

config/notifiers.go
pulord: done

pulord:
@simonpasquier check again, thanks for you help

pulord:
> @simonpasquier check again, thanks for your help

@simonpasquier @brian-brazil Hi, pls check this PR, thanks

config/notifiers.go
```diff
@@ -432,21 +429,40 @@ type WechatConfig struct {
...
-	AgentID   string `yaml:"agent_id,omitempty" json:"agent_id,omitempty"`
+	APISecret   Secret `yaml:"api_secret,omitempty" json:"api_secret,omitempty"`
+	CorpID      string `yaml:"corp_id,omitempty" json:"corp_id,omitempty"`
+	Message     string `yaml:"message,omitempty" json:"message,omitempty"`
+	APIURL      *URL   `yaml:"api_url,omitempty" json:"api_url,omitempty"`
+	ToUser      string `yaml:"to_user,omitempty" json:"to_user,omitempty"`
+	ToParty     string `yaml:"to_party,omitempty" json:"to_party,omitempty"`
+	ToTag       string `yaml:"to_tag,omitempty" json:"to_tag,omitempty"`
+	AgentID     string `yaml:"agent_id,omitempty" json:"agent_id,omitempty"`
+	MessageType string `yaml:"message_type" json:"message_type"`
```
simonpasquier: The default value should be "`text`".

template/default.tmpl
```diff
@@ -67,7 +67,6 @@ AlertmanagerUrl:
 {{ define "wechat.default.agent_id" }}{{ end }}
```
simonpasquier: We still need default values for the new fields: `Title`, `Description`, `URL`, ...

simonpasquier:
You would need to rebase your PR because of https://github.com/prometheus/alertmanager/pull/1929. Changes should go to `notify/wechat/` now.

SuperQ:

Issue #1850, opened 2019-04-23
Support creating silences based on alert filter


stuartnelson3:
Currently, the alert filter can be used to narrow down the list of alerts being shown, but a "pre-filled silence form in one click" is only supported for a firing alert. Being able to construct an alert filter that matches several firing alerts and then having a button to navigate to the silence form with those values pre-filled would be nice.

Anecdotally, I find myself filtering down to a handful of alerts, and then having to click on a single concrete alert whose pre-filled form I have to modify to include the other alerts from the previous page.

stuartnelson3:
closed by #1911

PR #1911, opened 2019-06-04
Create Silence from filter Bar

m-masataka:
In our operation, we often create silence after filtering alert with filter bar. 
So we need the button that create from filter.
This PR is related to #1850.


ui/app/src/Utils/Filter.elm
```diff
@@ -231,6 +232,31 @@ stringifyMatcher { key, op, value } =
...
 
+convertFilterMatchers : List Matcher -> List Data.Matcher.Matcher
+convertFilterMatchers matchers_ =
+    case matchers_ of
+        [] ->
+            []
+
+        list ->
+            list
+                |> List.map convertFilterMatcher
```
w0rm: I don't think we need the `convertFilterMatchers` function, because it can be replaced with `List.map convertFilterMatcher matchers`. 

There is no need to check for the empty list, because `List.map convertFilterMatcher [] == []`

ui/app/src/Utils/Filter.elm
```diff
@@ -231,6 +232,31 @@ stringifyMatcher { key, op, value } =
...
+
+
+convertFilterMatcher : Matcher -> Data.Matcher.Matcher
+convertFilterMatcher { key, op, value } =
+    { name = key
+    , value = value
+    , isRegex =
+        case op of
```
w0rm: `isRegex = op == RegexMatch` is shorter

ui/app/src/Views/FilterBar/Views.elm
```diff
@@ -122,6 +126,13 @@ view { matchers, matcherText, backspacePressed } =
                         , span
                             [ class "input-group-btn" ]
                             [ button [ class "btn btn-primary", disabled isDisabled, onClickAttr ] [ text "+" ] ]
+                        , a
+                            [ class "btn btn-outline-info border-0"
+                            , href (newSilenceFromMatchers dataMatchers)
+                            ]
+                            [ i [ class "fa fa-bell-slash-o mr-2" ] []
+                            , text "Silence"
+                            ]
```
w0rm: Could you please attach a screenshot to your PR? It would make it easier to review!

ui/app/src/Utils/Filter.elm
m-masataka: thanks for your review. I will remove it.

ui/app/src/Views/FilterBar/Views.elm
m-masataka: OK I'll attach the screenshot.

m-masataka:
output

w0rm:
@m-masataka thank you!

PR #1921, opened 2019-06-11
Add objectives to Summaries

beorn7:
With the next release of client_golang, Summaries will not have
objectives by default. To not lose the objectives we have right now,
explicitly state the current default objectives.

### DIFF
--- a/nflog/nflog.go
+++ b/nflog/nflog.go
@@ -104,12 +104,14 @@ func newMetrics(r prometheus.Registerer) *metrics {
 	m := &metrics{}
 
 	m.gcDuration = prometheus.NewSummary(prometheus.SummaryOpts{
-		Name: "alertmanager_nflog_gc_duration_seconds",
-		Help: "Duration of the last notification log garbage collection cycle.",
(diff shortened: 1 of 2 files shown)


brian-brazil:
Peeking at the code, both have fixed 15m cycles so the quantiles will be of no use and thus should be removed.

beorn7:
If the authors wanted meaningful quantiles, we should not remove them but increase `MaxAge` to an appropriate time. But perhaps they never really wanted quantiles. I'll leave that to the maintainers to decide.

stuartnelson3:
While I didn't add this code, I personally am not interested in quantiles for the internal nflog/silence GC. @simonpasquier any thoughts?

simonpasquier:
:+1: for me on removing quantiles, I don't think they're particularly useful.

beorn7:
Thanks for the feedback. Quantiles are explicitly set to empty now, which has the same behavior with the current and the future version of client_golang.

I also adjusted the commit text.

beorn7:
All tests green. Please merge or give me :+1: to do so.

PR #1924, opened 2019-06-14
Display response body when bad status.

m-masataka:
We occasionally get “Bad Request” without any message of this reason when Create Silence via webUI. (e.g. Input past date to “End”)
We need to get the reason of this message. 


stuartnelson3:
Input past end date is currently prevented in the UI:

image

are there other issues you're encountering when trying to create silences? Or are you somehow getting around this check?

m-masataka:
@stuartnelson3 Thanks for reply.
I know that validation is work.
But, that validation compares "Start" time with "End" time.
The "Bad Request" occurs when both of  start time and end time are past time.
It occurs in case of  "The gap of server time and client time" or  "Edit start time manually" etc...
I know it's very rare case, but I think it's important to show why did it fail.

PR #1926, opened 2019-06-17
Update prometheus/client_golang to v1.0.0

beorn7:


beorn7:
Oops. OK, alertmanager is actually using the API client, which needs updates to its usage.

Will work on it.

beorn7:
Let's see if that does the trick. Also paging @krasi-georgiev to see if that's the right way of adjusting to the client API changes.

krasi-georgiev:
Client API changes look good.

Issue #1566, opened 2018-10-02
Recreating silence should auto-fill comment with previous comment


stuartnelson3:
If a silence expires but needs to be extended (e.g. maintenance is taking longer than expected, priorities shift and an issue causing an alert has not been resolved), the current `recreate` button blanks out the comment when going to the silence page. It would be nice to copy over the previous message.

mikkelbu:
Looks like this should be closed by #1927

simonpasquier:
Yes, thanks!

PR #1927, opened 2019-06-17
Recreate silence with previous comment.

m-masataka:
When recreate silence by expired one, we want to copy over the previous comment and matchers. ( not only matchers)

This PR is related to #1566.


ui/app/src/Views/SilenceList/SilenceView.elm
```diff
@@ -114,7 +117,7 @@ editButton silence =
         Expired ->
             a
                 [ class "btn btn-outline-info border-0"
-                , href (newSilenceFromMatchers silence.matchers)
+                , href recreateUrl
```
w0rm: Hey! What do you think about reusing the new silence page and extending it to support passing an optional comment field rather than introducing the whole new route?

m-masataka: @w0rm Thank you for making a comment.
The reason why I introduce whole new route is to deal with possibilities change of SilenceView's field. This means to Avoid implement ``newSilenceFromMacherAndCommentAndXXXAndXXX..etc`` , when anyone need to add field in SilenceForm.

If low probability of extending new field in the future, implement ``newSilenceFromMatchersAndComment`` is simple and best way.
So, I can't make up my mind which is better.
Which do you think is better?

w0rm: Hey! Sorry for taking so long to respond. `newSilenceFromMatchersAndComment` sounds like a simpler and better option to me.

m-masataka: Okey.
I’ll change this PR.
Wait a moment please.

w0rm:
@m-masataka thank you!

PR #1928, opened 2019-06-18
template: add tests

simonpasquier:
One small improvement and more tests to make sure that it is valid.

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -15,16 +15,15 @@ package template
 
 import (
 	"bytes"
+	tmplhtml "html/template"
 	"io/ioutil"
 	"net/url"
 	"path/filepath"
 	"regexp"
 	"sort"
 	"strings"
-	"time"
-
-	tmplhtml "html/template"
 	tmpltext "text/template"
+	"time"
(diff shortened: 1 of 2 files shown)


template/template.go
```diff
@@ -15,16 +15,15 @@ package template
 
 import (
 	"bytes"
+	tmplhtml "html/template"
```
stuartnelson3: this was formatted like this by `goimports`? weird

simonpasquier: It was most probably a manual renaming.

template/template.go
```diff
@@ -325,6 +324,9 @@ func (t *Template) Data(recv string, groupLabels model.LabelSet, alerts ...*type
 					delete(commonAnnotations, an)
 				}
 			}
+			if len(commonLabels) == 0 && len(commonAnnotations) == 0 {
```
mxinden: Why not move this below `for _, a := range alerts[1:] {`? Am I missing something?

simonpasquier: Yes it works too.

Issue #1517, opened 2018-08-13
Create tests to prevent regressions in notifier requests


stuartnelson3:
Issue is being created as a result of https://github.com/prometheus/alertmanager/pull/1516

My initial idea is to store the request, verify it's working / visually inspect it, and then have test requests be compared to that stored request. It's not perfect, but we'll be notified if we have changed the request being sent and can decide if we need to update the stored request or it's a regression.

Taken from https://www.youtube.com/watch?v=yszygk1cpEc&t=14m27s

Thoughts @simonpasquier @mxinden ?

simonpasquier:
Another improvement would be to have  `notifier/XXX.go` and `notifier/XXX_test.go` files for every notifier instead of the monolithic `notifier/impl.go`.

Issue #1400, opened 2018-05-31
Decouple notifiers into a modular design


calston:
Notifiers should be decoupled from the main application logic more to make it easier to test and integrate support for new endpoints.

Currently it's far too intimidating to contribute new notification destinations quickly as it's not at all clear where these should be added in various places without having to know the entire source tree.

stuartnelson3:
Our recommendation currently is to integrate via webhook, as we are not adding new notification types right now.

calston:
Oh okay, so you're going to stick with a terrible architecture and close my suggestion without really reading the point.

Maybe I should stick to Elasticsearch ecosystem.

stuartnelson3:
Snide remarks rarely achieve anything positive. Pointing out to me in a friendly way that I mis-read your issue, however, would have been constructive.

Alertmanager is supported by free labor. If you're unhappy with the architecture, please submit a suggestion on how to fix it, and a pull request implementing it, instead of simply insulting what is currently there.

calston:
Yeah I was first opening an issue as a point of tracking a feature branch. 

This is not the first time someone on the Prometheus project has just immediately closed my issue without giving it any thought, and it reflects very poorly on the project.

juliusv:
Hey people. As far as I read the above, @stuartnelson3 mis-read the initial issue and maybe closed it a bit too hastily. That's not ideal, but responding with insults ("terrible architecture") is not helping either. Please try to remain respectful and assume the best in others.

I think the initial proposal was a good one at face value, except if you factor in the current policy of not adding new notification mechanisms, it limits its value a bit (because if no new mechanisms are added in AM directly, it will only help people who want to add a new mechanism in a fork). Independently of that, it could still make sense to modularize this if it leads to cleaner code in general.

To go a bit into why (as far from what I have been following) new notification mechanisms are currently on hold: they are not easily testable unless you have access to the necessary receiver infrastructure, and people who have contributed them have not always put in the effort to maintain them over time, even when they promised to do so. Thus currently the recommendation is to integrate new ones via the webhook, for better or worse.

ant31:
@calston could you explain why using the webhooks isn't satisfying enough for your use case?

calston:
@ant31 Webhooks has nothing to do with this case. But if you insist on the generic, not everything is a webservice, and not all webservices have the same API. Having a proxy web service to take the alertmanager and turn it into something that isn't one is just ridiculous.

[...]

PR #1929, opened 2019-06-18
Split the notify package into sub packages

simonpasquier:
Instead of keeping all notifiers in the notify package, it splits them into individual sub-packages. This improves readability and maintainability of the code.

Relates to https://github.com/prometheus/alertmanager/issues/1400 and https://github.com/prometheus/alertmanager/issues/1517.


cmd/alertmanager/main.go
```diff
@@ -92,6 +101,54 @@ func instrumentHandler(handlerName string, handler http.HandlerFunc) http.Handle
 
 const defaultClusterAddr = "0.0.0.0:9094"
 
+// buildReceiverIntegrations builds a list of integration notifiers off of a
+// receiver config.
+func buildReceiverIntegrations(nc *config.Receiver, tmpl *template.Template, logger log.Logger) ([]notify.Integration, error) {
```
mxinden: Shouldn't this be inside the `github.com/prometheus/alertmanager/notify` package?

stuartnelson3: Depending on @simonpasquier's reasoning, I could understand it being moved here. It seems like this is a bunch of imperative code using the notify package components, so it's kind of the glue tying together a library? :man_shrugging:, will wait to see what he has to say.

simonpasquier: I saw no concrete reason to keep it in `github.com/prometheus/alertmanager/notify` as it is the glue between configuration and notifiers.

mxinden: :+1:

.circleci/config.yml
```diff
@@ -43,7 +43,11 @@ jobs:
           username: user
           password: pass
           EOF
-    - run: make
+    - run:
+        command: make
+        environment:
+          # Run garbage collection more aggresively to avoid getting OOMed.
+          GOGC: "20"
```
mxinden: Am I right that CircleCI gives us an 8gb machine? https://circleci.com/docs/2.0/executor-types/

I am surprised that we are being OOMed in that case.

simonpasquier: No we got 4GB because we use the Docker executor with the free plan. We could use a machine executor but it takes more time to start. IMO we need to ask the CNCF for a paid account that would allow to use bigger [resources](https://circleci.com/docs/2.0/configuration-reference/#resource_class). I can bring it up during our next sync meeting.

stuartnelson3: I'm surprised we even hit 4gb, to be honest.

PR #1931, opened 2019-06-19
notify/pagerduty: fix images

### LINKED ISSUE #1784: PagerDuty support with images is broken
The PagerDuty receiver sends invalid data when the `images` property is set.

https://github.com/prometheus/alertmanager/blob/0a2d108bb9b5cd1efd7d3d868cf8632ecf4cf6f8/notify/impl.go#L507-L511

The [API](https://v2.developer.pagerduty.com/docs/send-an-event-events-api-v2#section-the-images-property) doesn't expect `text`, it should probably be `href`.

Reported by https://groups.google.com/d/msg/prometheus-users/CsyY9lEAUyA/yEPEOGt7CAAJ

simonpasquier:
Closes #1784

For clarity, I've chosen to rename the field `text` to `href` in the configuration though it could break people (but I suspect that it isn't widely used).

Unfortunately I don't have a way to validate it. Maybe @seanhoughton, @sbueringer, @mikebryant or @adamdecaf could help (I've looked at the Git history for people that contributed to the PagerDuty code).

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -225,7 +225,7 @@ type PagerdutyLink struct {
 type PagerdutyImage struct {
 	Src  string `yaml:"src,omitempty" json:"src,omitempty"`
 	Alt  string `yaml:"alt,omitempty" json:"alt,omitempty"`
-	Text string `yaml:"text,omitempty" json:"text,omitempty"`
+	Href string `yaml:"href,omitempty" json:"href,omitempty"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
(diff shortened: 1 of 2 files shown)


mikebryant:
I've verified this change using https://v2.developer.pagerduty.com/docs/send-an-event-events-api-v2, and successfully created an image link in a real alert using:
```
    "images": [
        {
            "src": "https://vignette.wikia.nocookie.net/reddwarf/images/5/54/Flibble.jpg/revision/latest?cb=20070516034130",
            "href": "https://google.com"
        }
    ],
```

simonpasquier:
Thanks a lot @mikebryant

PR #1932, opened 2019-06-20
Limit the number of tests executed in parallel

simonpasquier:
By default Go uses GOMAXPROCS but a Circle CI executor has many cores (> 30) while the CPU and RAM resources are throttled (a medium instance has 2 vCPUs and 4GB of RAM). If we don't limit this to the number of allocated cores, the job is likely to get OOMed and killed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,8 +46,11 @@ jobs:
     - run:
         command: make
         environment:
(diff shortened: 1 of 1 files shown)


simonpasquier:
I retried the job 2 times and it didn't fail. Not an absolute guarantee but a good sign nevertheless.

PR #1936, opened 2019-06-23
fix memory visibility error

bigMacro:
One day, alertmanager could not send alerts in my production. So i made a dump file.  The dump file shows that `Dispather` blocks on cleaning up empty aggrGroups while aggrGroup blocks on `ag.done` in `aggrGroup.stop()`.  After analysing the code, `aggrGroup.stop()` should terminate the `aggrGroup.run()` loop, and then `aggrGroup.stop()` exits normally. But in fact, it does not work as we hope. 

I think that there is a memory visibility error. `aggrGroup.run()` runs as a goroutine which initializes `ag.done`, and `aggrGroup.stop()` runs in another goroutine which reads `ag.done` channel itself and waiting value from this channel. This is a data race! If those two goroutine runs on different cpu cores, `aggrGroup.stop()` may be read from a nil channel even though `aggrGroup.run()` has initialized the channel. So i fix this error by initializing `ag.done` in `newAggrGroup()`.

There is may be another error. If `aggrGroup.stop()` runs before `aggrGroup.run()`, `aggrGroup.stop()` may be also waiting from a nil channel because `ag.done` not be initialized in `aggrGroup.run()`.  Of course, it seems that `aggrGroup.stop()` won't run before `aggrGroup.run()`. But i don't think that initializing  shared variables in one goroutine is a good practice.


bigMacro:
It's very hard to reproduce this erorr, so i sorry that i cann't make a unit test. All of above is my analysis and this is the only reason that i provide.

stuartnelson3:
Thanks for the in-depth analysis! I'll give this a read and look through it, and see if I can come up with a unit test.

simonpasquier:
The code looks ok. I see no reason why the `done` channel wasn't initialized in the `newAggrGroup()` function. I don't see an easy to write a unit test for this either.
There might also be a race condition if `ag.stop()` is [called](https://github.com/prometheus/alertmanager/blob/9d5a7a271a6c0e50ee6b9da068204c9d25d6e19d/dispatch/dispatch.go#L122) before the first alert is [added](https://github.com/prometheus/alertmanager/blob/9d5a7a271a6c0e50ee6b9da068204c9d25d6e19d/dispatch/dispatch.go#L270) to the aggregation group?

simonpasquier:
Thanks!

PR #1946, opened 2019-06-28
Add Microsoft Teams notifier

arneschulz1984:
Some weeks ago, my company decides to use Microsoft Teams.
So we have to migrate from Slack to Microsoft Teams.

After looking around, my preferred solution will be a native Prometheus Alertmanager Notifier.
I do not like Grafana-Alerts so much and an extra container for forwarding the Webhook i also do not like so much.

So pleas have a look at my native implementation of the Microsoft Teams Notifier implementation.


cmd/alertmanager/main.go
```diff
@@ -16,6 +16,7 @@ package main
 import (
 	"context"
 	"fmt"
+	"github.com/prometheus/alertmanager/notify/msteams"
```
AndreasBergmeier6176: Maybe set up your editor to run `go fmt` on save?

simonpasquier:
Thanks for suggesting this change. Unfortunately, we are currently not accepting new integrations as noted in the [documentation](https://prometheus.io/docs/alerting/configuration/#receiver).

We can only provide the stability and performance we want to provide if we can properly maintain the codebase. This includes, amongst others, test integrations in an automated and scalable fashion. For this reason, we are suggesting people integrate with the help of our generic interfaces. We have an integrations page on which integrations using our generic interfaces are listed.

Even if existing integrations can not be tested in an automated fashion, we will not remove them for reasons of compatibility. This also means that any additions we take on, or any changes to existing integrations we make or accept, will mean maintaining and testing these until at least the next major version, realistically even beyond that.

Feel free to question this answer on our [developer mailing list](https://groups.google.com/forum/#!forum/prometheus-developers), but be aware it's unlikely that you will get a different answer.

simonpasquier:
Closing it given the current moratorium on receivers. We're aware that it isn't a satisfying answer but the reasons listed above still stand for now.

mmethner:
Too bad that the PR was rejected, thanks anyway for the work.

PR #1948, opened 2019-07-02
api/v2: fix alerts with multiple receivers

### LINKED ISSUE #1942: v0.17.0 - Receivers in the v2 alerts endpoint are wrong when using `continue: true` in routes
**What did you do?**
Use a `continue: true` statement in your routing configuration.
Then send an alert that match the `continue` receiver and the next receiver.

While going to /api/v1/alerts or /api/v2/alerts, listed receivers for an alert that match a `continue` receiver + another receiver should be the same in both API.

The /api/v2/alerts is listing the same receiver twice, the one that don't have the `continue` statement. 
```
"receivers": [
  {
    "name": "r1"
  },
  {
    "name": "r1"
  }
],
```
The /api/v1/alerts works as expected.
```
"receivers": [
  "urgent",
  "r1"
],
```

**Environment**
The bug is present in v0.17.0
Using version v0.16.2 works as expected.

Relevant configuration:
```
route:
  receiver: r1
  group_by:
  - clusterName
  - projectName
  - alertname
  - severity
  - namespace
  routes:
  - receiver: blackhole
    match:
      alertname: DeadMansSwitch
  - receiver: urgent
    match_re:
      severity: (high|critical)
    continue: true
  - receiver: r1
    match:
      owner: name1
  - receiver: r2
    match:
      owner: name2
  group_wait: 30s
  group_interval: 5m
  repeat_interval: 1h
```

Thanks guys.

simonpasquier:
Closes #1942

### DIFF
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -432,8 +432,8 @@ func alertToOpenAPIAlert(alert *types.Alert, status types.AlertStatus, receivers
 	endsAt := strfmt.DateTime(alert.EndsAt)
 
 	apiReceivers := make([]*open_api_models.Receiver, 0, len(receivers))
-	for _, name := range receivers {
-		apiReceivers = append(apiReceivers, &open_api_models.Receiver{Name: &name})
+	for i := range receivers {
+		apiReceivers = append(apiReceivers, &open_api_models.Receiver{Name: &receivers[i]})
 	}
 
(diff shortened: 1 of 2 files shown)


simonpasquier:
@mxinden @stuartnelson3 friendly ping. I think it would be worth having it in v0.18.0.

Issue #1944, opened 2019-06-26
v0.17.0 stuck in a loop when trying to send email if invalid mail template is used


zellerfagyi:
**What did you do?**
- Start up AlertManager with a misconfiguration: in receivers.email_configs.html a 'email.html' file was referred as a template, however it was not a template but rather a HTML file
(alertmanager configuration and email.html attached)
- POST the following alert to /api/v1/alerts:

`+ curl -X POST http://alertmanagerhost:9593/api/v1/alerts -d '[{
                "status": "firing",
                "labels": {
                        "alertname": "Grafana healthcheck won't start up",
                        "service": "grafana_hc_watch",
                        "severity":"critical",
                        "instance": "instance.na.me"
                },
                "annotations": {
                        "summary": "<arbitrary summary of the issue>"
                }}]'
Response: {"status":"success"}`

**What did you expect to see?**
- Expected to see that AlertManager sends a mail to the configured mail address

**What did you see instead? Under which circumstances?**
- AlertManager sent a mail as expected, but the mail body was entirely empty. Mail subject was filled out correctly with the expected content (based on alertname, instance, etc.)
- But then after the initial mail sent out, it looked like it got stuck in a loop and sent the same (empty) mail over and over in 10 second intervals until the AlertManager process was manually terminated.
- Would expect the mail not to be sent out at all or fail more gracefully instead of being stuck in a loop and sending mail every 10 seconds

**Environment**

* System information:

	Linux 3.10.0-957.1.3.el7.x86_64 x86_64

* Alertmanager version:

	`$ /usr/local/bin/alertmanager --version
alertmanager, version 0.17.0 (branch: HEAD, revision: c7551cd75c414dc81df027f691e2eb21d4fd85b2)
  build user:       root@932a86a52b76
  build date:       20190503-09:10:07
  go version:       go1.12.4`

* Prometheus version:

	don't think it's relevant

* Alertmanager configuration file:
```
global:
  resolve_timeout: 5m

templates:
- '/export/alertmanager/*.tmpl'

route:
  group_by: ['alertname']
  group_wait: 10s
  group_interval: 10s
  repeat_interval: 1h
  receiver: 'mail-receiver'
receivers:
- name: 'slack-receiver'
  slack_configs:
  - api_url: <slack url>
    channel: '#grafana-alerts'
    title: "PROD {{ range .Alerts }}{{ .Annotations.summary }}\n{{ end }}"
[...]

PR #1953, opened 2019-07-09
notify/email: wrap all errors for easier debugging

simonpasquier:
In addition, this commit passes the current context to the TCP dialer and it doesn't log any QUIT errors if the email delivery wasn't successful.

Related #1944 
Related #980

### DIFF
--- a/notify/email/email.go
+++ b/notify/email/email.go
@@ -17,7 +17,6 @@ import (
 	"bytes"
 	"context"
 	"crypto/tls"
-	"errors"
 	"fmt"
 	"mime"
 	"mime/multipart"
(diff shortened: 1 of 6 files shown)


notify/email/email.go
```diff
@@ -112,77 +112,87 @@ func (n *Email) auth(mechs string) (smtp.Auth, error) {
 
 // Notify implements the Notifier interface.
 func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
+	// TODO: move the check to the config pacakge.
```
mxinden:
```suggestion
	// TODO: move the check to the config package.
```

Is this supposed to happen in this pull request or a follow up?

simonpasquier: I planned for a follow up PR.

PR #1960, opened 2019-07-11
cmd/alertmanager: reject invalid external URLs

### LINKED ISSUE #1912: Template email not replacing external and generator URL
**What did you do?**
In the compose I set up:
command:
- '--web.external-url=localhost:9093'
But still in the email I am getting docker ID and the port, which I guess is default behavior.

I want my email to have correct Alert Manager and source addresses.

* Alertmanager version:

	Latest from docker

* Prometheus version:

	Latest from docker

simonpasquier:
Closes #1912

Using `--web.external-url=foo:123` instead of `--web.external-url=http://foo:123` was accepted before but it's (almost) never what you want as `html/template` will escape the value to `#ZgotmplZ` because it only accepts `http://...`, `https://...` or `mailto://...` links.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -14,12 +14,12 @@
 package api
 
 import (
+	"errors"
 	"fmt"
 	"net/http"
 	"runtime"
 	"time"
 
-	"github.com/pkg/errors"
 	apiv1 "github.com/prometheus/alertmanager/api/v1"
 	apiv2 "github.com/prometheus/alertmanager/api/v2"
 	"github.com/prometheus/alertmanager/cluster"
@@ -96,7 +96,7 @@ func (o Options) validate() error {
 // call is also needed to get the APIs into an operational state.
 func New(opts Options) (*API, error) {
 	if err := opts.validate(); err != nil {
-		return nil, errors.Wrap(err, "invalid API options")
+		return nil, fmt.Errorf("invalid API options: %s", err)
 	}
 	l := opts.Logger
 	if l == nil {
@@ -128,6 +128,7 @@ func New(opts Options) (*API, error) {
 		log.With(l, "version", "v2"),
 		opts.Registry,
 	)
+
 	if err != nil {
 		return nil, err
 	}
--- a/api/api.go
+++ b/api/api.go
@@ -14,12 +14,12 @@
 package api
 
 import (
-	"errors"
 	"fmt"
 	"net/http"
 	"runtime"
 	"time"
 
+	"github.com/pkg/errors"
 	apiv1 "github.com/prometheus/alertmanager/api/v1"
 	apiv2 "github.com/prometheus/alertmanager/api/v2"
 	"github.com/prometheus/alertmanager/cluster"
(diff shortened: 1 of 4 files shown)


cmd/alertmanager/main.go
```diff
@@ -481,16 +482,25 @@ func clusterWait(p *cluster.Peer, timeout time.Duration) func() time.Duration {
...
+type hostnameFunc func() (string, error)
+
+func extURL(logger log.Logger, listen, external string) (*url.URL, error) {
+	return hostnameFunc(os.Hostname).resolveExtURL(logger, listen, external)
+}
+
+func (h hostnameFunc) resolveExtURL(logger log.Logger, listen, external string) (*url.URL, error) {
 	if external == "" {
-		hostname, err := os.Hostname()
+		hostname, err := h()
```
brian-brazil: Where's this defined?

api/api.go
```diff
@@ -96,7 +96,7 @@ func (o Options) validate() error {
 // call is also needed to get the APIs into an operational state.
 func New(opts Options) (*API, error) {
 	if err := opts.validate(); err != nil {
-		return nil, fmt.Errorf("invalid API options: %s", err)
+		return nil, errors.Wrap(err, "invalid API options")
```
brian-brazil: There seems to be unrelated changes in here.

cmd/alertmanager/main.go
```diff
@@ -499,7 +509,11 @@ func extURL(listen, external string) (*url.URL, error) {
 	if err != nil {
 		return nil, err
 	}
+	if u.Scheme != "http" && u.Scheme != "https" {
+		return nil, errors.Errorf("%q: invalid %q scheme, only 'http' and 'https' are supported", u.String(), u.Scheme)
+	}
 
+	fmt.Println("path:", u.Path, "raw", u.RawPath)
```
brian-brazil: Debug code?

api/api.go
simonpasquier: Yup, will do it in another PR.

cmd/alertmanager/main.go
simonpasquier: Totally :)

cmd/alertmanager/main.go
simonpasquier: It's the method receiver.

brian-brazil:
:+1:

cmd/alertmanager/main.go
```diff
@@ -481,16 +482,25 @@ func clusterWait(p *cluster.Peer, timeout time.Duration) func() time.Duration {
 	}
 }
 
-func extURL(listen, external string) (*url.URL, error) {
+type hostnameFunc func() (string, error)
```
mxinden: What is this indirection through `extURL -> hostnameFunc -> resolveExtURL` for? I don't understand why this is not just a single function.

simonpasquier: To resolve the external URL we need to get the hostname but I didn't want to call `os.Hostname` in the tests so I went for this convoluted way. I'll add a comment unless you can think of something clearer.

mxinden: How about passing `os.Hostname` as an argument to `resolveExtURL`? Would remove the complexity of instantiating a new type.

PR #1963, opened 2019-07-15
dispatch: use strings.Builder instead of []byte

simonpasquier:


simonpasquier:
@mxinden totally correct. I was working on some other change and thought it would be a bit cleaner this way.

PR #1965, opened 2019-07-19
notify/opsgenie: log error from OpsGenie API

simonpasquier:
Ref: https://groups.google.com/d/msg/prometheus-users/3sAQ9kxEgPI/rE-ADUYiBAAJ

Tested against OpsGenie API, the error message looks like this:

```
cancelling notify retry for \"opsgenie\" due to unrecoverable error: unexpected status code 422: {\"message\":\"Request body is not processable. Please check the errors.\",\"errors\":{\"priority\":\"should be one of [ P1, P2, P3, P4, P5 ]\"},\"took\":0.005,\"requestId\":\"610e618e-7138-4898-9c70-c4eef2d68bf8\"}
```

### DIFF
--- a/notify/opsgenie/opsgenie.go
+++ b/notify/opsgenie/opsgenie.go
@@ -18,11 +18,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"net/http"
 	"strings"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
 	commoncfg "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
 
@@ -85,7 +88,7 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 	}
 	defer notify.Drain(resp)
 
-	return n.retry(resp.StatusCode)
+	return n.retry(resp.StatusCode, resp.Body)
 }
 
 // Like Split but filter out empty strings.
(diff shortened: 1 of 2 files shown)


notify/opsgenie/opsgenie.go
```diff
@@ -189,14 +192,19 @@ func (n *Notifier) createRequest(ctx context.Context, as ...*types.Alert) (*http
...
-		return true, fmt.Errorf("unexpected status code %v", statusCode)
-	} else if statusCode/100 != 2 {
-		return false, fmt.Errorf("unexpected status code %v", statusCode)
+func (n *Notifier) retry(statusCode int, body io.Reader) (bool, error) {
+	if statusCode/100 == 2 {
+		return false, nil
 	}
 
-	return false, nil
+	err := errors.Errorf("unexpected status code %v", statusCode)
```
bwplotka: I think we put too many things into this function to just name it `retry` (: 

Can we have `isRetriable` with what we have previously and then have this:
```
err := errors.Errorf("unexpected status code %v", statusCode)
if body != nil {
		if bs, errRead := ioutil.ReadAll(body); errRead == nil {
			err = errors.Errorf("%s: %s", err, string(bs))
		}
	}
```
in `Notify` if the `isRetriable(..)` is false?

simonpasquier: Hmm, it wouldn't work as-is because we can have errors that shouldn't be retried and for which logging the response body is still useful IMO.

FWIW the current form is consistent with what we have in other places (eg [Slack](https://github.com/prometheus/alertmanager/blob/4706dc7e6aebb486aeeac8c385905acc8c32fcd6/notify/slack/slack.go#L183-L198)). If anything we should factor out what is common across notifiers.

simonpasquier:
I'll have a look for the next step.

PR #1974, opened 2019-07-25
notify: refactor code to retry requests

simonpasquier:
Follow up of #1965

### DIFF
--- a/notify/util.go
+++ b/notify/util.go
@@ -24,6 +24,7 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
 
 	"github.com/prometheus/alertmanager/template"
 	"github.com/prometheus/alertmanager/types"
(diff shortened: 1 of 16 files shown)


notify/util.go
```diff
@@ -139,3 +140,57 @@ func GetTemplateData(ctx context.Context, tmpl *template.Template, alerts []*typ
...
+	// Function to return additional information in the error message.
+	CustomDetailsFunc func(code int, body io.Reader) string
+	// Additional HTTP status codes that should be retried.
+	RetryCodes []int
+}
+
+// Process returns a boolean indicating whether the request should be retried
+// and an optional error if the request has failed. If body is not nil, it will
+// be included in the error message.
+func (r *Retrier) Process(statusCode int, body io.Reader) (bool, error) {
```
stuartnelson3: I don't want to bikeshed this, but I feel like `Check` would be a better method name. I guess I don't know. Neither is much butter than the other I suppose.

simonpasquier: I like `Check` better too.

PR #1977, opened 2019-07-29
notify: don't use the global metrics registry

simonpasquier:
This was the last package that was still using the global registry.

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -246,9 +246,7 @@ func newMetrics(r prometheus.Registerer) *metrics {
 		m.numFailedNotifications.WithLabelValues(integration)
 		m.notificationLatencySeconds.WithLabelValues(integration)
 	}
-	if r != nil {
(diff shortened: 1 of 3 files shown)


notify/notify.go
```diff
@@ -261,8 +206,64 @@ type NotificationLog interface {
...
+		"slack",
+		"opsgenie",
+		"webhook",
+		"victorops",
+	} {
+		m.numNotifications.WithLabelValues(integration)
+		m.numFailedNotifications.WithLabelValues(integration)
+		m.notificationLatencySeconds.WithLabelValues(integration)
+	}
+	if r != nil {
```
mxinden: This check should only ever evaluate to false in tests, right? We could prevent passing in `nil` in production code.  Creating a `Registerer` in tests should be cheap. I would suggest panicking if `r` is `nil` and instantiating a new registry in the unit tests.

simonpasquier: > This check should only ever evaluate to false in tests, right?

correct I'm fine with your suggestion.

PR #1979, opened 2019-07-31
Fix dedup repeat time check misses peer wait consideration

garrettlish:
If we don't consider peer wait, dedup repeat interval passed or not check is not accurate. Especially sometimes the repeat time is very small and smaller than the max peer wait time will cause duplicated alerts from alertmanager peer.


garrettlish:
got some problems, close it

PR #1980, opened 2019-07-31
Fix dedup repeat time check misses cluster wait

garrettlish:
If we do not check cluster wait, dedup repeat interval passed check is inaccurate.
Especially sometimes the repeat time is too small and smaller than the cluster peer wait time will cause duplicated alerts from alertmanager peer.

Signed-off-by: xiancli <xiancli@ebay.com>


garrettlish:
the timestamp adjustment might not be correct if the nflog keeps broadcasts across the cluster

Issue #1660, opened 2018-12-12
Resolution notifications dependency on group_interval


kapodes:
**What did you do?**
Sent an alert and waited for it to hit `resolve_timeout` and be resolved

**What did you expect to see?**
Notification about alert being resolved right away after it was resolved according to Alertmanager

**What did you see instead? Under which circumstances?**
Nothing until `group_interval` timer hit.

```
alertmanager, version 0.15.3 (branch: HEAD, revision: d4a7697cc90f8bce62efe7c44b63b542578ec0a1)
  build user:       root@4ecc17c53d26
  build date:       20181109-15:40:48
  go version:       go1.11.2
```

* Alertmanager configuration file:
```
global:
  resolve_timeout: 1m
route:
  group_wait: 5s
  group_interval: 5m
  receiver: slack-notifications

receivers:
  - name: "slack-notifications"
    slack_configs:
      - channel: "#alerts"
        send_resolved: true
        ..........
```

* Logs:
```
level=debug ts=2018-12-12T15:36:34.910107275Z caller=dispatch.go:445 component=dispatcher aggrGroup={}:{} msg=Flushing alerts=[TestAlert[c508c62][active]]
level=debug ts=2018-12-12T15:41:34.910626655Z caller=dispatch.go:445 component=dispatcher aggrGroup={}:{} msg=Flushing alerts=[TestAlert[c508c62][resolved]]
```

If I decrease `group_interval` to something like `5s` I will see notification almost instantly after alert is resolved.

I think `send_resolved` logic shouldn't be dependent on `group_interval`. 
It's logical not to send new alert notifications from same group right away but resolution messages are not new alerts...

kapodes:
Consider updating docs to mention resolution notifications are also affected. 
It took me some time to understand what is happening.

```
# When the first notification was sent, wait 'group_interval' to send a batch
# of new alerts that started firing for that group.
group_interval: 5m
```

mxinden:
@kapodes would you like to prepare a patch to improve the docs?

TheMeier:
There has been no activity on this issue for a long time. In the meantime the documentation also was improved. If you still think further documentation is needed, please feel free to provide a PR to implement it.

PR #1981, opened 2019-07-31
Improve documentation for group_interval

lauriku:
After struggling for a while to understand that alert resolution notifications follow `group_interval` as well, and reading through this issue: https://github.com/prometheus/alertmanager/issues/1660 , here's a suggestion on how to improve the documentation a bit.

I'm open to different approaches on formatting or language.

Signed-off-by: Lauri Kuittinen <lauriku@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,8 @@ route:
   group_wait: 30s
 
   # When the first notification was sent, wait 'group_interval' to send a batch
-  # of new alerts that started firing for that group.
+  # of new alerts that started firing for that group. Alert resolution notifications
+  # follow this interval as well.
   group_interval: 5m
 
   # If an alert has successfully been sent, wait 'repeat_interval' to
--- a/doc/examples/simple.yml
+++ b/doc/examples/simple.yml
@@ -34,7 +34,8 @@ route:
   group_wait: 30s
 
   # When the first notification was sent, wait 'group_interval' to send a batch
-  # of new alerts that started firing for that group.
+  # of new alerts that started firing for that group. Alert resolution notifications
+  # follow this interval as well.
   group_interval: 5m
 
   # If an alert has successfully been sent, wait 'repeat_interval' to


README.md
```diff
@@ -86,7 +86,8 @@ route:
   group_wait: 30s
 
   # When the first notification was sent, wait 'group_interval' to send a batch
-  # of new alerts that started firing for that group.
+  # of new alerts that started firing for that group. Alert resolution notifications
```
brian-brazil: This doesn't read cleanly. I'd suggest something like "to send the next batch of alerts for that group" that implicitly covers both.

Frodox: > to send the next batch of alerts for that group

not so clean also.
Maybe

`to send the next batch of notifications for that group` ? 
because *alert* means something like *firing*, not *resolved* :thinking:

brian-brazil: That doesn't work. There's no such thing as a a batch of notifications, only a group of alerts.

Frodox: ugh. It is totally confusing. Any ideas about improving the docs? I really want to clarify, that `group_interval' affects both firing and resolved notifications

Frodox: As you wrote here https://www.robustperception.io/whats-the-difference-between-group_interval-group_wait-and-repeat_interval

> Instead we wait for the group_interval since the last notification was sent to the group, and then send all alerts firing (and any resolved alerts) to the receiver.

It is good enough. But it is not official docs :D

AllenzhLi: A related issue: the repeat_interval is also confusing.

In some case, I received notification more than once in a repeat interval. This is surprising.

AllenzhLi:
A related issue: the repeat_interval is also confusing.

In some case, I received notification more than once in a repeat interval. This is surprising.

tonyhuangzerohash (approved):

tonyhuangzerohash (approved):

SuperQ:

PR #1982, opened 2019-08-01
Check whether firing alert is resolved from peer during dedup

garrettlish:
Right now we only check the firing alerts are in notify log alerting list. However, the alert which notified already and resolved by another peer which not shown in alerting list but only shown in resolved list. We need to exclude these kinds of resolved alerts, otherwise, the user will get duplicated alerts from different alertmanager nodes.

Signed-off-by: xiancli <xiancli@ebay.com>


simonpasquier:
Hmm, it doesn't not seem quite right as it could potentially drop legit notifications. Say that a group has alert A and B firing which trigger a notification. After `group_interval`, alert A gets resolved and another notification is sent. After another `group_interval`, alert A is firing again but the notification would be dropped with this change.

garrettlish:
Thanks @simonpasquier for your comments. 
From https://github.com/prometheus/alertmanager/blob/master/notify/notify.go#L554-L571 and https://github.com/prometheus/alertmanager/blob/master/notify/notify.go#L708-L713, you can see resolved alerts are from AlertIterator (endsAt > now) rather than nflog query.
So (a) assume that a group has alert A and B firing which trigger a notification. (b) after `group_interval`, alert A gets resolved and another notification is sent. (c) After another `group_interval`, *alert A is firing again but the notification would be sent again with this change. It is because (b) will update nflog resolved alerts list to remove alert A since it is no longer in AlertIterator*. Do I answer your question? Any thoughts, please advise.

simonpasquier:
Sorry I don't understand what issue you're trying to solve. Having a reproducer would be good.

garrettlish:
@simonpasquier 
Assume:
1. We have 2 alert manager nodes in the cluster: alertManagerA, alertManagerB.
2. The group interval is 5 seconds
3. The cluster wait is 10 seconds 
Here is the reproduce steps:
1. alertA generates at 09:00:01. alertManagerA picks up alertA at 09:00:05 and send it out (broadcast nflog which includes alertA in firing list to alertManagerB)
2. alertA resolved at 09:00:06, alertManagerA picks up alertA at 09:00:10 and resolve it (broadcast nflog which includes alertA in resolved list to alertManagerB)
3. alertManagerB which starts 09:00:05 and wait 10 seconds (cluster wait), it picks alertA which timestamp is 09:00:01 and check nflog, since alertA is NOT in firing list and we don't check nflog resolved list, alertA sent by alertManagerB again (duplicated alert)

garrettlish:
retest this please

simonpasquier:
Yes, notifications tend to duplicate when `group_interval` is less than the maximum "cluster wait" time. This is why the default value for `group_interval` (eg 5m) is larger than what it takes for AlertManager instances to process the group.

garrettlish:
@simonpasquier 
Do you think we need to handle `group_interval` less than the maximum "cluster wait" time use case? From my understanding, I think it is common since usually we might have 3 or more alert manager nodes in the cluster which means that the max cluster wait time might be > 15 * 2 = 30 seconds, and `group_interval` < 30 seconds should be common as well (such as db metrics related alerting).

brian-brazil:
> group_interval < 30 seconds should be common as well (such as db metrics related alerting).

Humans can't respond that quickly, and this is only subsequent grouped alerts. For that tight a control loop, I'd be looking to have automation that didn't involve Prometheus and Alertmanager.

garrettlish:
okay, thanks @brian-brazil @simonpasquier

PR #1999, opened 2019-08-14
config: use the origin field instead of local var

johncming:
The origin field of config is not currently in use. It can be used to replace local variables

### DIFF
--- a/config/coordinator.go
+++ b/config/coordinator.go
@@ -95,7 +95,7 @@ func (c *Coordinator) notifySubscribers() error {
 
 // loadFromFile triggers a configuration load, discarding the old configuration.
 func (c *Coordinator) loadFromFile() error {
-	conf, plainConfig, err := LoadFile(c.configFilePath)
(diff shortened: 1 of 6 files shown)


config/coordinator.go
```diff
@@ -104,7 +104,7 @@ func (c *Coordinator) loadFromFile() error {
 	c.config = conf
 	c.configSuccessMetric.Set(1)
 	c.configSuccessTimeMetric.Set(float64(time.Now().Unix()))
-	hash := md5HashAsMetricValue(plainConfig)
+	hash := c.config.MD5()
```
simonpasquier: Since it's in the `config` package, `Coordinator` can access `conf.original` so this line doesn't have to change much (eg `hash := md5HashAsMetricValue(conf.original)`) and we don't need to introduce `Config.MD5()`.

johncming: ok, I will change it.

simonpasquier:
Thanks!

PR #2002, opened 2019-08-16
Pagerduty: implement append_common_annotations and append_common_labels

simonjefford:
We have found a need to dynamically add KV pairs to the [`details` field in `pagerduty_config`](https://prometheus.io/docs/alerting/configuration/#pagerduty_config). I have found similar requests in the Prometheus users Google group - one for [pagerduty itself](https://groups.google.com/forum/#!searchin/prometheus-users/pagerduty$20details%7Csort:date/prometheus-users/TD00BnzbG2M/8ClEg7q3BQAJ) and one for [opsgenie](https://groups.google.com/forum/#!searchin/prometheus-users/pagerduty$20details%7Csort:date/prometheus-users/TD00BnzbG2M/8ClEg7q3BQAJ) and I can't find any follow-up or consensus on whether or not this is something desirable by the project. Apologies if I've missed some discussion somewhere.

This PR adds 2 boolean options to `pagerduty_config`, `append_common_annotations` and `append_common_labels` which simply appends the relevant KV pairs to `details`.

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -213,6 +213,9 @@ type PagerdutyConfig struct {
 	Class       string            `yaml:"class,omitempty" json:"class,omitempty"`
 	Component   string            `yaml:"component,omitempty" json:"component,omitempty"`
 	Group       string            `yaml:"group,omitempty" json:"group,omitempty"`
+
+	AppendCommonAnnotations bool `yaml:"append_common_annotations,omitempty" json:"append_common_annotations,omitempty"`
+	AppendCommonLabels      bool `yaml:"append_common_labels,omitempty" json:"append_common_labels,omitempty"`
 }
 
 // PagerdutyLink is a link
--- a/notify/pagerduty/pagerduty.go
+++ b/notify/pagerduty/pagerduty.go
@@ -250,6 +250,19 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 		details[k] = detail
 	}
 
+	if n.conf.AppendCommonAnnotations {
+		for k, v := range data.CommonAnnotations {
+			level.Info(n.logger).Log(k, v)
+			details[k] = v
+		}
+	}
+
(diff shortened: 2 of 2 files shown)


brian-brazil:
Discussion at the Barcelona dev summit was that we'd have Alertmanager handle the basic use cases, and anything more complicated could go via the webhook.

This particular PR for example is rather niche. What if you wanted to modify the labels/annotations somehow?

simonjefford:
Generalising config by having the ability to add KV pairs to details that you don't know about at config time doesn't feel particularly niche or complex. Maintaining a webhook to get that functionality does.

brian-brazil:
It's niche as it does it in exactly one way, in the simplest case what if I wanted annotations to override labels? Or only wanted some annotations?

simonjefford:
Well, if I knew in advance which annotations or labels I wanted and which ones I didn't then I would just explicitly specify those as templates in my config. The PR should probably change though so that those get precedence.

As for CommonAnnotations overriding CommonLabels or vice-versa I admit, I don't really have an answer for that, other than more config, which I'm guessing wouldn't be an acceptable approach?

simonjefford:
Or, failing that, dropping a warning into the logs that a value has been overridden?

0robustus1:
I've found myself wanting a very similar feature.
@brian-brazil would it be acceptable to implement it as suggested including the precedence for manually set fields (e.g. in order to update the values of some label or annotation)?
If that is the case i could prepare an updated PR.

charlseo:
@simonjefford we have configured this in our own alertmanager image and it indeed presents alerts in more readable format and no problem with other functionality. Is there any reason this PR has not been approved? I think this should be a good feature to have

SoloJacobs:
Hi there! Since we haven't heard back in a while, I’m going to close this PR for now to keep our queue organized.
Feel to submit a new PR whenever you're ready to pick this back up.

Kind regards

PR #2008, opened 2019-08-22
Vendor prometheus 2.12.0

codesome:

PR #2009, opened 2019-08-26
*: remove dependency on github.com/prometheus/prometheus

simonpasquier:
We depend on the Prometheus repository only for the `github.com/prometheus/prometheus/pkg/labels` package and from this package, we use only the [`Matcher`](https://godoc.org/github.com/prometheus/prometheus/pkg/labels#Matcher) struct.

Knowing that this code hasn't evolved for years now, it seems safe and simpler to duplicate it in AlertManager rather than the heavy dependency we have right now (see #2008). In addition Prometheus has an indirect dependency on `github.com/hashicorp/memberlist` and many other Hashicorp libs which can cause version changes whenever we need to bump github.com/prometheus/prometheus.


brian-brazil:
Once we unify the two labels packages, we might want to kick this to common.

simonpasquier:
> Once we unify the two labels packages, we might want to kick this to common.

I thought we were going away from common.

PR #2011, opened 2019-08-26
Allow slashes in receivers

0robustus1:
It seems that alertmanager was splitting the `Receiver` in templates by the first slash and only exposing the value prior to this first slash.
I couldn't find out whether there was a particular reason for this as `Receiver`s with slashes seem to be working well within every other part of the alertmanager.

Changes:
* Do not split receiver by slash in templates instead expose the complete receiver
* Percent-encode the `Receiver` within the default alertmanager URL.


simonpasquier:
Thanks! It looks like the code has been like that forever but I can't think of a justification for it. Maybe @stuartnelson3 has some thoughts?

stuartnelson3:
Can't think of anything. Seems fine to me

simonpasquier:
Thanks!

Issue #1875, opened 2019-05-03
duplicate alert groups returned by api


stuartnelson3:
**What did you do?**

view the alert list page via the ui

**What did you expect to see?**

non-duplicated alert groups

**What did you see instead? Under which circumstances?**

some alert groups are duplicated. i'm not sure why yet.

**Environment**

* System information:

	insert output of `uname -srm` here

* Alertmanager version:

	0.17.0

Of the 110 currently active alert groups at soundcloud, only 2 of the groups are duplicated (for a total of 4 groups being returned from the api, when only 2 of them should be). The groups appear to be perfectly duplicated.

stuartnelson3:
This might be an artifact of that particular group's configuration. It only affects a single team. I'm thinking there might be some race condition in creating the groups internal to alertmanager when the routes are configured in a certain way. This would also explain why they have been seeing duplicate messages.

taisph:
We currently see a duplicated alert in the UI with 0.17.0 but not with amtool alert query. Clicking "+ Info" on either expands both alerts in the UI. Not sure if it is the same issue.

psalaberria002:
I also see duplicate groups. Version 0.18.

![Screen Shot 2019-07-10 at 01 04 42](https://user-images.githubusercontent.com/2410564/60928743-ca169e80-a2ae-11e9-962e-2a4e69523e7b.png)

simonpasquier:
I can reproduce as well. It happens when I configure routes with the same `group_by` and `continue: true`. I'll dig into it...

simonpasquier:
https://github.com/prometheus/alertmanager/pull/1964 has fixed the problem on the API v2 side but there's still a glitch in the UI: when several groups have the same group labels, clicking on the +/- button will expand/collapse all of them.

![screencast](https://user-images.githubusercontent.com/2587585/63594843-d1241280-c5b7-11e9-822a-c77b31fddec1.gif)

simonpasquier:
Closed by #2012 and #1964

PR #2012, opened 2019-08-27
Grouping label's expand button with grouping id

m-masataka:
In alerts screen, Expand/Collapse button cooperated with grouping labels.
So when several groups have the same grouping labels, this button will expand all of them. Related to #1875 .
To fix this issue, I implemented expand button with grouping id. (this id is only used UI)


m-masataka:
alertmanager

ui/app/src/Views/AlertList/Views.elm
```diff
@@ -98,23 +98,23 @@ defaultAlertGroups activeId activeGroups expandAll groups =
...
         _ ->
             div [ class "pl-5" ]
-                (List.map
-                    (\{ labels, alerts } ->
-                        alertGroup activeId activeGroups (Dict.toList labels) alerts expandAll
-                    )
-                    groups
+                (List.indexedMap Tuple.pair groups
+                    |> List.map
+                        (\( index, group ) ->
```
w0rm: There is no need to first do `indexedMap` to build tuples and then map over tuples. You can call `List.indexedMap` with `\index group ->`.

w0rm:
A small improvement, everything else looks good! I think that after this change this should also perform better, because the key type of a Set is simpler. 

Thank you!

simonpasquier:
Thanks a lot @m-masataka! You need to rebase your PR after #2013 is merged.

@stuartnelson3 @mxinden once this gets merged, I suggest that we consider a new release of AlertManager as #1875 is quite annoying and has been reported in different occasions. I can sign up for preparing the new release.

ui/app/src/Views/AlertList/Views.elm
m-masataka: Thanks!

PR #2026, opened 2019-09-09
Add progress bar to Silence List

m-masataka:
I implement the progress bar with Silence. 
Because we want to get to know how long will the Silence be Active intuitively.  
スクリーンショット 2019-09-09 9 08 47


w0rm:
UX-wise, I have a few concerns about this PR:

1. This introduces two ways to display the time in the same line, on the left we have the time stamp, and on the right we are going to have the remaining time. Maybe we can decide which one is more useful, and only show another one in a tooltip?
2. The choice of a progress bar makes it look like we have real time updates, but this doesn't happen to the list of silences.

m-masataka:
Thank you for lead me this issue @w0rm .
> The choice of a progress bar makes it look like we have real time updates, but this doesn't happen to the list of silences.

That is just as you indicated. 
I don’t think progress bar is the best way.
I just want to how lang does the silence expired.
Current view is not suitable for me, because we are in different timezone, etc ...

PR #2027, opened 2019-09-09
Alertmanager: Add Microsoft Teams Webhook notification.

Myslik:
This adds notifications to Microsoft Teams. It is using the Webhook connector. Let me know what you think.

It should be simple as this
```
receivers:
- name: teams-receiver
  teams_configs:
    - webhook_url: https://outlook.office.com/webhook/[REDACTED]
```

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -53,6 +53,7 @@ import (
 	"github.com/prometheus/alertmanager/notify/pagerduty"
 	"github.com/prometheus/alertmanager/notify/pushover"
 	"github.com/prometheus/alertmanager/notify/slack"
+	"github.com/prometheus/alertmanager/notify/teams"
 	"github.com/prometheus/alertmanager/notify/victorops"
 	"github.com/prometheus/alertmanager/notify/webhook"
 	"github.com/prometheus/alertmanager/notify/wechat"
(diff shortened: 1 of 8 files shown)


mxinden:
We're not adding any more receivers to AlertManager (see [doc](https://prometheus.io/docs/alerting/configuration/#receiver)). You could implement a custom binary via the Alertmanager [webhook](https://prometheus.io/docs/alerting/configuration/#webhook_config) receiver. Once that project exists we are happy to list it on the [integrations](https://prometheus.io/docs/operating/integrations/#alertmanager-webhook-receiver) page.

I am sorry that we can't accept your work here. To prevent this from happening in the future, please open a Github issue prior to working on a feature so that we can discuss it beforehand.

Myslik:
@mxinden Are direct integrations to be phased out in the future? Is webhook receiver preferred approach?

simonpasquier:
Yes the webhook receiver is the recommended way to integrate 3rd-party systems with AlertManager. In case you aren't aware of it, https://github.com/idealista/prom2teams already does this.

Myslik:
Thanks for info

mxinden:
There is no decision in regards to what should happen with the existing integrations. We are not accepting new ones due to the maintenance burden this putts on the maintainers.

Yes, the webhook receiver is the preferred approach.

PR #2028, opened 2019-09-10
Ensure nil registry not registered

jtlisi:
The types package does not follow the common pattern of avoiding metrics registration when a nil registry parameter is passed. This copies the pattern used in the silences/nflog/etc package.

Signed-off-by: Jacob Lisi <jacob.t.lisi@gmail.com>

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -122,8 +122,12 @@ func (m *memMarker) registerMetrics(r prometheus.Registerer) {
 	alertsActive := newAlertMetricByState(AlertStateActive)
 	alertsSuppressed := newAlertMetricByState(AlertStateSuppressed)
 
-	r.MustRegister(alertsActive)
-	r.MustRegister(alertsSuppressed)
+	if r != nil {
+		r.MustRegister(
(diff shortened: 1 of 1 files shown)


simonpasquier:
I'm not sure we have a consistent rule regarding this: there's at least another [place](https://github.com/prometheus/alertmanager/blob/d5f6f8892824a485663301b9a542c6b6dff91c27/notify/notify.go#L249) where a nil registerer isn't allowed. So I think we need first to decide which way we want to promote. I would lean toward not allowing nil registerers to make sure that we don't forget to register metrics. And I think @mxinden has a [similar opinion](https://github.com/prometheus/alertmanager/pull/1977#discussion_r317534981).

stuartnelson3:
I would also prefer not allowing nil registerers

simonpasquier:
@jtlisi would you mind amending the PR to remove all checks for nil registerers?

jtlisi:
@simonpasquier There is not much of a point to this PR if it does not check for nil registerers.

PR #2042, opened 2019-09-20
 Clustered Etcd Alert Store

dcwangmit01:
Please refer to the Design doc [design.md](https://github.com/prometheus/alertmanager/pull/2042/files#diff-f6c2927c69493eec41cdb26c41bc0d21) which provides background on this PR.  Here's the most relevant snippet.

## Status Quo

The current Alertmanager clustering implementation will allow multiple
instances of Alertmanager to sync recently-updated notification log entries and
alert silences via the gossip protocol.  The syncing of alert state is not yet
implemented.

Currently, Alertmanager alert state is entirely stored in process memory.  It
is lost upon restart of Alertmanager.  This, in conjunction with the fact that
Alertmanager alert state is not synced with clustered peers presents several
problems:

* Alert emitters must send alerts to every single Alertmanager instance in the
  cluster.  If some Alertmanager instances receive alerts that other instances
  do not, their state will drift apart.

  * Typically, the alert emitter is Prometheus which uses service discovery to
    locate all Alertmanager instances, and then sends alerts to each individual
    target.  Custom alert emitters may not have that service discovery
    capability and must be explicity configured.

* Because the alert must be sent to every Alertmanager instance, this precludes
  a typical HA deployment pattern of having clients (i.e. alert emitters) send
  alerts through a Load Balancer that fronts several Alertmanager instances.
  This pattern is typical in Kubernetes.

* Restarting Alertmanager instances or increasing the replicaCount for the
  Alertmanager cluster will result in the new members having very different
  alert states than current members.

* If all of the Alertmanagers reboot at the same time, all active alerts are
  lost.  New alerts will be created in each instance memory as alerts come in
  through the API.  However, these alerts will have different startsAt times
  than the prior alerts in memory.  The effect is that "status=resolved"
  notifications will never be sent for the original set of alerts in memory.
  Because some notification receivers consider unique instances of alerts to be
  a combination of alert fingerpint and startsAt, these notification reciverse
  will never receive "status=resolved" notifications.

The issues above may be addressed by implementing a share persistent data store
for alert state.

## Design

How the Etcd Alerts Storage Provider Works

Initialization (Synchronous)

* Upon startup, Alertmanager tries to connect to Etcd.  If the connection
  attempt is unsuccessful after a timeout, then Alertmanager will fail hard
  because it is probably a configuration error.  A supervisor such as
  Kubernetes should restart it.

* Once Alertmanager successfully connects to Etcd, it proceeds with "best
  effort" to write alert updates to Etcd.  Alert updates are the current
  in-memory "running count" state of an alert, and not the log of the entire
  alert stream.

* Alertmanager also initiates an Etcd watch client to receive alert updates
  from Etcd.  As other Alertmanagers in the cluster write to Etcd, all of the
  Alertmanagers receive and process those updates.

* If Etcd disappears for any reason (e.g. network partition, quorum loss,
  etc.), Alertmanager will continue running independently to maintain
  availability.  Once connectivity to Etcd is restored, Alertmanager will
  resume writing alerts to Etcd, and receiving alert updates from Etcd.

* As the last part of initialization, Etcd will read and load all alert objects
  from Etcd.

Ongoing Operation (Async)

* Alertmanager can receive alerts from both API requests AND the Etcd watch
  subscription.  Alerts from both of these sources are sent to the etcd alert
  provider via Put().  To prevent circular loops and promote state convergence,
  the alert is put into Etcd only when there is a difference between the alert
  in memory and the alert in Etcd.

* The Etcd clientv3 library will automatically reconnect if the Etcd server
  goes down or up, or upon resuming network connectivity if a partition had
  occurred.

* The long running watch subscription will reconnect if the watch channel is
  closed.


dcwangmit01:
Prior 2 forced pushes hopefully address build failures.  Changed my dev compiler to go 1.13 to match CI.

doc/design/clustered-etcd-alert-store/design.md
```diff
@@ -0,0 +1,150 @@
...
+    capability and must be explicity configured.
+
+* Because the alert must be sent to every Alertmanager instance, this precludes
+  a typical HA deployment pattern of having clients (i.e. alert emitters) send
+  alerts through a Load Balancer that fronts several Alertmanager instances.
+  This pattern is typical in Kubernetes.
+
+* Restarting Alertmanager instances or increasing the replicaCount for the
+  Alertmanager cluster will result in the new members having very different
+  alert states than current members.
```
mxinden: Can you add more details in which scenario scaling the alertmanager cluster is not a irregular process?

mxinden: > Restarting Alertmanager instances

When each instance has persistence they should have a *good-enough* snapshot of the current notification and silcence state on disk. Thereby end-users should not have duplicate notifications.

doc/design/clustered-etcd-alert-store/design.md
```diff
@@ -0,0 +1,150 @@
...
+Date: 2019-09-20
+
+Author: David C Wang <dcwangmit01@gmail.com>
+
+## Status Quo
+
+The current Alertmanager clustering implementation will allow multiple
+instances of Alertmanager to sync recently-updated notification log entries and
+alert silences via the gossip protocol.  The syncing of alert state is not yet
+implemented.
```
mxinden: > The syncing of alert state is not **yet** implemented.

So far I was not aware of anyone asking for this to be implemented.

doc/design/clustered-etcd-alert-store/design.md
```diff
@@ -0,0 +1,150 @@
...
+problems:
+
+* Alert emitters must send alerts to every single Alertmanager instance in the
+  cluster.  If some Alertmanager instances receive alerts that other instances
+  do not, their state will drift apart.
+
+  * Typically, the alert emitter is Prometheus which uses service discovery to
+    locate all Alertmanager instances, and then sends alerts to each individual
+    target.  Custom alert emitters may not have that service discovery
+    capability and must be explicity configured.
```
mxinden: > Custom alert emitters may not have that service discovery capability and must be explicity configured.

I don't understand how discovering / configuring e.g. 3 instead of 1 alertmanager instance is more than a linear additional overhead.

mxinden: Is this under the assumption that the *alert emitter* is not under the control of the administrator and thus can not be easily changed to emit alerts not to one but to all Alertmanagers?

doc/design/clustered-etcd-alert-store/design.md
```diff
@@ -0,0 +1,150 @@
...
+* If all of the Alertmanagers reboot at the same time, all active alerts are
+  lost.  New alerts will be created in each instance memory as alerts come in
+  through the API.  However, these alerts will have different startsAt times
+  than the prior alerts in memory.  The effect is that "status=resolved"
+  notifications will never be sent for the original set of alerts in memory.
+  Because some notification receivers consider unique instances of alerts to be
+  a combination of alert fingerpint and startsAt, these notification reciverse
+  will never receive "status=resolved" notifications.
+
+The issues above may be addressed by implementing a share persistent data store
```
mxinden:
```suggestion
The issues above may be addressed by implementing a shared persistent data store
```

doc/design/clustered-etcd-alert-store/design.md
brian-brazil: Noone has, there have been one or two people who looked for a forwarder of some kind.

doc/design/clustered-etcd-alert-store/design.md
```diff
@@ -0,0 +1,150 @@
...
+* Alert emitters must send alerts to every single Alertmanager instance in the
+  cluster.  If some Alertmanager instances receive alerts that other instances
+  do not, their state will drift apart.
+
+  * Typically, the alert emitter is Prometheus which uses service discovery to
+    locate all Alertmanager instances, and then sends alerts to each individual
+    target.  Custom alert emitters may not have that service discovery
+    capability and must be explicity configured.
+
+* Because the alert must be sent to every Alertmanager instance, this precludes
```
brian-brazil: This design doesn't change that, for reliability you still need to send it to multiple Alertmanagers. Comparing with stateless services isn't really useful.

doc/design/clustered-etcd-alert-store/design.md
```diff
@@ -0,0 +1,150 @@
...
+  alerts through a Load Balancer that fronts several Alertmanager instances.
+  This pattern is typical in Kubernetes.
+
+* Restarting Alertmanager instances or increasing the replicaCount for the
+  Alertmanager cluster will result in the new members having very different
+  alert states than current members.
+
+* If all of the Alertmanagers reboot at the same time, all active alerts are
+  lost.  New alerts will be created in each instance memory as alerts come in
+  through the API.  However, these alerts will have different startsAt times
```
brian-brazil: This isn't considered a problem. startsAt is an implementation detail, and resolved notifications are best effort.

doc/design/clustered-etcd-alert-store/design.md
```diff
@@ -0,0 +1,150 @@
...
+  than the prior alerts in memory.  The effect is that "status=resolved"
+  notifications will never be sent for the original set of alerts in memory.
+  Because some notification receivers consider unique instances of alerts to be
+  a combination of alert fingerpint and startsAt, these notification reciverse
```
brian-brazil: This is a bad assumption on the part of those receivers. An alerts labels are all that identify it, and we have never indicated otherwise.

doc/design/clustered-etcd-alert-store/design.md
```diff
@@ -0,0 +1,150 @@
...
+alert silences via the gossip protocol.  The syncing of alert state is not yet
+implemented.
+
+Currently, Alertmanager alert state is entirely stored in process memory.  It
+is lost upon restart of Alertmanager.  This, in conjunction with the fact that
+Alertmanager alert state is not synced with clustered peers presents several
+problems:
+
+* Alert emitters must send alerts to every single Alertmanager instance in the
+  cluster.  If some Alertmanager instances receive alerts that other instances
```
brian-brazil: This will cause duplicate notifications, but no problems beyond that. This is considered acceptable.

brian-brazil:
I agree with @mxinden, this doesn't seem to be adding anything while introducing a lot of complexity and a CP system which we've purposefully avoided. I'd also wonder how well something like etcd could deal with a full alert volume worth of updated.

dcwangmit01:
@mxinden, @brian-brazil, @simonpasquier.  Thanks for the reviews, questions, and the consideration.

Here's some more information to help clarify the themes of the review comments you've made.


**MORE ON BACKGROUND**

My team is building a telco mobile phone network composed of many subsystems from many vendors. A lot of our work involves impedance matching the monitoring of these systems into native Prometheus.  We run Prometheus in the middle, without control of the systems being monitored nor the systems receiving the notifications.

In the startups I've worked for, we only cared about active alerts.  If an alert happened, we took care of it immediately and it was off the books as soon as it disappeared from the Alertmanager dashboard.  In the giant company I work for now, Alertmanager must trigger tickets to be opened so that issues can be triaged, prioritized, and routed to the right vendor's team.

The requirements of working with telcos (and financial companies) are pretty rigid.  Every *firing* alert notification (webhook) of a unique fingerprint must eventually receive a *resolved* alert notification once the situation is fixed. The notification receiver directly translates alert notifications into ticket openings and closings in their work tracking system.

The Alertmanager code, out of the box, is fully capable of supporting the idea that every *firing* alert notification eventually receives a *resolved* alert notification. That must have been the intention, because there exists a `resolve_timeout` configuration that will automatically close alerts and trigger *resolved* notifications.  This coupled with the fact that Alertmanager maintains the same *startsAt* for subsequent notifications of the same fingerprint (until closure) means that Alertmanager already associates multiple notifications can be associated into a single alert instance (e.g. ticket).  For example, the alert notifications that opened and closed a ticket two months ago is a different alert instance than the notifications with the same fingerprint that we received today which resulted in a new ticket being created.  The Alertmanager design, whether intentional or not, showed great foresight to be capable of supporting these use cases.

Alertmanager supports what my team needs, so why is this PR necessary?  The problem is corner cases. There are situations where Alertmanager will lose track of an alert and never send a *resolved* notification, which downstream operational systems need in order to close tickets. This PR makes sure that every opening of an alert notification eventually receives close notification (and subsequently the tickets are opened and closed as well). These corner cases stem from the fact that Alertmanager has an in-memory alert store and starts with a fresh empty state upon startup.

Here's a simplified real-world example, considering an Alertmanager cluster of replicaCount=1.

* Alertmanager starts up and with a global `resolve_timeout=5m` and a route's `group_wait=30s` and runs for a while.

* At some time in the future, a *firing* alert with a new fingerprint arrives at the API. After 30s of `group_wait`, Alertmanager triggers a *firing* notification to the operational system which opens a ticket.

* 3 minutes later Alertmanager gets killed by Kubernetes and moved to a new Kubernetes node because an operator triggered a rolling update.

* Alertmanager starts up with fresh empty state on a new Kubernetes node.  During this startup duration, the alert is resolved.

* Alertmanager either:
  * Misses the *resolved* alert.  In this case no *resolved* notification is sent to the downstream operational system to close the corresponding ticket.
  * OR receives a *resolved* alert at its API for a fingerprint that does not exist in its memory store.  Because Alertmanager has never seen this fingerprint before, no *resolved* notification is sent to the downstream operational system to close a ticket.


The chances that the corner case above happens for any particular alert may be small, but when there are thousands of alerts flowing through the system the issue happens quite often.  The effect is that the operational system ends up with quite a few orphaned tickets.

This PR addresses the root of the problem by enabling Alertmanager to save state in a network data store, and then reload state when it starts up again.  This is disabled by default.  It has a nice side effect that all Alertmanager in a cluster can sync alert state.


**REGARDING CODE COMPLEXITY**

Here's the analysis

* Effectively, there are an additional ~350 lines of code changes, almost all additions.

* There are about ~1100 lines of new code if you exclude the go mod files, vendor directory, supporting documentation, and docker-compose demo files.

* The *mem* provider was forked to an additional *etcd* provider to keep the code paths separate.  The tests are the same except for the addition of etcd specific tests.  The code could be DRY'd up if the *etcd* provider could be enhanced with persistent store callbacks.  That would save 174 lines of code.

* 641 of the ~1100 lines of code are tests
```
 provider/etcd/etcd_client_test.go  | 272 +++++++++++++++
 provider/etcd/etcd_test.go         | 369 +++++++++++++++++++++
```

* 525 of the ~1100 lines of code have an effect on the Alertmanager Binary
```
 Makefile.common                    |   1 +
 api/v1/api_test.go                 |   1 +
 cmd/alertmanager/main.go           |  31 +-
 inhibit/inhibit_test.go            |   1 +
 provider/etcd/etcd.go              | (+20 lines top of 174 from mem.go) +++++++++++
 provider/etcd/etcd_client.go       | 295 ++++++++++++++++
 provider/provider.go               |   2 +
```

* There are less than ~350 lines of changes to support etcd, when excluding the mem.go fork.

doc/design/clustered-etcd-alert-store/design.md
dcwangmit01: Scaling the cluster shouldn't be often.  The point I was trying to make is that there are multiple ways of getting the Alertmanagers within the same cluster to have different internal alert states, because alert state is not persisted.  Scaling up or down the cluster is one.  Doing a kubernetes rolling update of the Alertmanager Statefulset is another.  Chaoskube kills pods randomly.  Cluster-autoscaler moves pods off of nodes before downscaling the number of nodes, etc.

dcwangmit01: It is not the state of notifications or silences that is the concern of this PR.  This PR mostly involves the internal state of the alerts which are captured in the *mem* provider.  In our situation receiving duplicate notifications is okay.  What we have trouble with is Alertmanager losing its internal alert state and never ever sending notifications to close alerts that are no longer active.

doc/design/clustered-etcd-alert-store/design.md
dcwangmit01: Acknowledged.  I hope my teams aren't the only ones with this need.  Perhaps we are on the cutting edge of integrating with legacy systems at telcos.

doc/design/clustered-etcd-alert-store/design.md
dcwangmit01: Kubernetes may move Alertmanager pods around a will as nodes disappear, chaoskube kills them, rolling updates occur, the cluster autoscaler kills pods to pack them on other nodes before downscaling, etc.  Each of these occurrences can move an Alertmanager instance to new Pod with a new IP.  Unless the number of Alertmanager replicas never changes AND the alert emitters know how to query an updated target list from a Kubernetes headless service, the configuration will have to be manually managed.  We're hoping to avoid administrator intervention.

dcwangmit01: Yes.  That that is a very real deployment model.  If a vendor does not support metrics and only supports network notifications, we may ask them to format and send their events as *alerts* that are compatible with the Alertmanager API. 

Since these vendors would be outside of our Kubernetes cluster and would not have access to Alertmanagers in our private kubernetes overlay network, the only way they can send *alerts* to our Alermanagers is to traverse via Kubernetes Loadbalancer integration, which would hit exactly one target.

My team has already written a few alert-exporter type software components that impedance match between fault-based (not metrics-based) legacy systems and Alertmanager.  Some of these tools can read custom API faults tables and translate them directly into alerts sent to Alertmanager.  We even convert SNMP Traps to Alertmanager alerts.  Painful.

doc/design/clustered-etcd-alert-store/design.md
dcwangmit01: Acknowledged.  If this PR has any chance of merging I'll fix it.

doc/design/clustered-etcd-alert-store/design.md
dcwangmit01: The effect of this PR is that the alert does go to each Alertmanager in the cluster, even if it is sent to a single Alertmanager instance.  If an alert gets to any individual Alertmanager instance, the alert state, if it is a new fingerprint or has been modified, is written to Etcd.  Then, the other Alertmanagers in the cluster will receive the updated alert state through the *etcd watch* clients that they have initiated.

doc/design/clustered-etcd-alert-store/design.md
dcwangmit01: It may be an issue for downstream notification receivers which open and close tickets based on notifications.  I've detailed it below in the thread.

doc/design/clustered-etcd-alert-store/design.md
dcwangmit01: True.  It is an assumption made by reading the current implementation of the Alertmanager code.  You are correct that it has never been advertised as such.

The trouble is how can we make Alertmanager integration with ticketing systems better. Let's say a memory fault alert triggered a notification and a ticket on memory module replacement 2 months ago, and was fixed.  If that same alert happened again today, Alertmanager can consider it the same alert but the downstream ticketing system must track them as two separate issues.  Thus, the result is the assumption and dependence on startsAt.  Do you have a recommendation for distinguishing between the two instances of alerts?

doc/design/clustered-etcd-alert-store/design.md
dcwangmit01: Agreed, there is nothing wrong with duplicate notifications.  The concern is that some there are corner cases where *resolved* notifications are never sent for alerts that are no longer active.

doc/design/clustered-etcd-alert-store/design.md
brian-brazil: That's going to be a heuristic, after some amount of time you'd open a new ticket rather than re-opening the old one.

doc/design/clustered-etcd-alert-store/design.md
brian-brazil: If it's entirely inside a k8 cluster, then SD can handle it. If it's going across clusters then you probably have well-known DNS names which will be used.

doc/design/clustered-etcd-alert-store/design.md
brian-brazil: That sounds like a case for a custom alert forwarder, which doesn't need to live in the AM itself.

doc/design/clustered-etcd-alert-store/design.md
brian-brazil: That's also considered acceptable. There's cases where we drop firing alerts too, but at the end of the day the goal is to provide sufficiently reliable notifications to a human in a reasonable time frame when there's an issue.

brian-brazil:
> The requirements of working with telcos (and financial companies) are pretty rigid. Every firing alert notification (webhook) of a unique fingerprint must eventually receive a resolved alert notification once the situation is fixed.

That's something best tracked outside the AM then, like Prometheus itself it's not designed for this sort of 100% precise behaviour.

This also sounds like a case of https://www.robustperception.io/running-into-burning-buildings-because-the-fire-alarm-stopped

> The Alertmanager code, out of the box, is fully capable of supporting the idea that every firing alert notification eventually receives a resolved alert notification. That must have been the intention, because there exists a resolve_timeout configuration that will automatically close alerts and trigger resolved notifications.

That's not right, that's only a default in case the alert source doesn't provide one. At some point if a Prometheus goes away we have to consider the alert to be no longer firing. That doesn't mean the underlying issue is resolved though, nor that the alert has stopped firing.

simonpasquier:
@dcwangmit01 thanks for all the details you've put into this.

My concern about maintenance isn't strictly about the code, it's also with all the "peripheral" activities like helping users to debug issues, writing documentation, explaining when you would need to use etcd vs. memory store.

In practice, you will still have to deal with "unresolved" notifications even if at a lower rate. So I's agree with Brian that it would be best handled by an external service.

dcwangmit01:
Thanks guys for the feedback!

gherlein:
So reading this thread carefully iI think this is the summary:

1.  the maintainers feel that the performance of Alert Manager is aimed at informing humans, not automation, and the "best effort" tracking of alert state is acceptable for the human use case
2.  the maintainers feel that they would have to take on some degree of additional work to support this enhanced capability, and for their use of Alert Manager the new feature is not worth the work

Did I get that correctly?

simonpasquier:
Like for any opensource project, we estimate the addition of a new feature from several angles:
* Does it fit with the project's philosophy and architecture? Alertmanager is designed to work without depending on external services. Requiring etcd (even as an option) would set a precedent.
* Does it solve a pain point for most users? Currently we have no evidence that there's a significant demand.
* Can we add this to our maintenance workload?

_"no is temporary, yes is forever"_ (courtesy of [Solomon Hykes](https://twitter.com/solomonstre/status/715277134978113536?s=20))

mpeters:
Having best effort persistence across restarts is something I and many other people would appreciate. It seems there is some thought in designing the providers to be modular even if there is only a single one currently. Would there be a way to allow external providers? This way AM doesn't have to take on the maintenance or worry about design philosophy, but this pain can be solved for users who need it.

doc/design/clustered-etcd-alert-store/design.md
Bpoe: I was very surprised that the instances did not sync the alerts. We were looking at using alertmanager outside the usual Prometheus and Kubernetes usage, but this prevented it.

PR #2043, opened 2019-09-23
types: remove unused equal method and add test case.

johncming:
matcher miss a method for equal.  The current way is that compare two structures directly.  matcher needs a special method for compare equal. 

add a test case for matchers compares.

### DIFF
--- a/types/match.go
+++ b/types/match.go
@@ -14,11 +14,10 @@
 package types
 
 import (
+	"bytes"
 	"fmt"
 	"regexp"
 	"sort"
-
-	"bytes"
 
 	"github.com/prometheus/common/model"
 )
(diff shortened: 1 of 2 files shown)


types/match.go
```diff
@@ -80,6 +81,11 @@ func (m *Matcher) Match(lset model.LabelSet) bool {
 	return string(v) == m.Value
 }
 
+// Equal checks whether two matchers are equal.
+func (m *Matcher) Equal(o *Matcher) bool {
```
simonpasquier: IIUC `Matchers.Equal()` isn't used anywhere in the Alertmanager repository so I would remove that method instead.

go.mod
```diff
@@ -19,6 +19,8 @@ require (
 	github.com/jessevdk/go-flags v1.4.0
 	github.com/kylelemons/godebug v0.0.0-20160406211939-eadb3ce320cb
 	github.com/matttproud/golang_protobuf_extensions v1.0.1
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
```
simonpasquier: You need to use Go 1.13 (or not commit this change).

johncming: ok.

types/match.go
johncming: how about change to private method? compare two arrays sometimes is useful.

johncming: ok

johncming:
updated.

PR #2045, opened 2019-09-25
nflog: use errors.New instead of fmt.Errorf for no custom error msg.

johncming:
use errors.New instead of fmt.Errorf for no custom error msg.


johncming:
> Could you also update this line? Thanks!
> 
> https://github.com/prometheus/alertmanager/blob/dfcaf26658c3aad5342d1f65bd00a4966d8bb12d/nflog/nflog.go#L196

ok

johncming:
updated.

simonpasquier:
:+1:

PR #2052, opened 2019-09-26
README.md: add table of contents

simonpasquier:


mxinden:
How do we ensure that the *table of contents* stays in sync? I thought about this in the past, but couldn't find an easy-to-use tool for it.

simonpasquier:
I've used https://github.com/ekalinin/github-markdown-toc for this. There's still a manual copy/paste operation required.

PR #2056, opened 2019-09-27
dispatch: convert to a function.

johncming:
`cleanup` is an important function that involves the close of alerts, convert to a function is better.


simonpasquier:
I don't find personally that the change makes the code clearer. But happy to hear a different opinion from @mxinden and @stuartnelson3.

stuartnelson3:
yeah, I don't really see any change here 🤷‍♂

PR #2057, opened 2019-10-02
Add email notify Message-Id Header

n33pm:
## Problem
Rspamd - `MISSING_MID (2.5)`

Reference: [jordan-wright/email](https://github.com/jordan-wright/email/blob/041e0cec78b06dd35f6431f4fa92f2b29fa30d67/email.go#L612)

### DIFF
--- a/notify/email/email.go
+++ b/notify/email/email.go
@@ -16,11 +16,9 @@ package email
 import (
 	"bytes"
 	"context"
-	"crypto/rand"
 	"crypto/tls"
 	"fmt"
-	"math"
-	"math/big"
+	"math/rand"
 	"mime"
 	"mime/multipart"
 	"mime/quotedprintable"
@@ -29,7 +27,6 @@ import (
 	"net/smtp"
 	"net/textproto"
 	"os"
-	"regexp"
 	"strings"
 	"time"
 
@@ -46,12 +43,11 @@ import (
 
 // Email implements a Notifier for email notifications.
 type Email struct {
-	conf   *config.EmailConfig
-	tmpl   *template.Template
-	logger log.Logger
+	conf     *config.EmailConfig
+	tmpl     *template.Template
+	logger   log.Logger
+	hostname string
 }
-
-var reHeaderMessageId = regexp.MustCompile(`(?im)^message-id:`)
 
 // New returns a new Email notifier.
 func New(c *config.EmailConfig, t *template.Template, l log.Logger) *Email {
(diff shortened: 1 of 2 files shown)


simonpasquier:
Hmm, we can't copy code from https://github.com/jordan-wright without keeping the copyright and the permission notice. I'd prefer that we implement our own id generator and `<timestamp.random_number@hostname>`  is probably enough.

notify/email/email.go
```diff
@@ -239,6 +243,11 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 	multipartBuffer := &bytes.Buffer{}
 	multipartWriter := multipart.NewWriter(multipartBuffer)
 
+	id, err := generateMessageID()
```
simonpasquier: For sanity, you need to make sure that the `Message-Id` header doesn't already exist.

n33pm: done

notify/email/email.go
```diff
@@ -236,6 +243,13 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 		fmt.Fprintf(buffer, "%s: %s\r\n", header, mime.QEncoding.Encode("utf-8", value))
 	}
 
+	if reHeaderMessageId.Find(buffer.Bytes()) == nil {
```
simonpasquier: I'd rather check that `n.conf.Headers` doesn't contain the `message-id` key.

notify/email/email.go
```diff
@@ -16,15 +16,20 @@ package email
 import (
 	"bytes"
 	"context"
+	"crypto/rand"
```
simonpasquier: I think that `math/rand` is enough for what we need here. I would add a `generator *rand.Rand` field to the `Email` struct as well as a `hostname string` field (so `os.Hostname()` is called only once). Then `generateMessageID()` would become a method of `*Email`.

notify/email/email.go
n33pm: done

notify/email/email.go
n33pm: - `generateMessageID()` is not needed anymore
- add `os.Hostname()` to `func New`

simonpasquier:
LGTM. I've added a small commit on top to cover the change in the unit tests.

n33pm:
> LGTM. I've added a small commit on top to cover the change in the unit tests.

thx..

PR #2059, opened 2019-10-02
docs: comments for template examples in simple.yml

base698:


doc/examples/simple.yml
```diff
@@ -10,6 +10,8 @@ global:
   hipchat_api_url: 'https://hipchat.foobar.org/'
 
 # The directory from which notification templates are read.
+#   See: https://prometheus.io/docs/alerting/notification_examples/#defining-reusable-templates
```
brian-brazil: This is fairly advanced stuff, I'm not sure we should be pointing users to here from a very basic example.

simonpasquier: I agree.

base698: This references a template but has no corresponding example.  It was confusing when I was trying to find a way to update the text in a notification.  Maybe take it out or provide a complete example?

brian-brazil: Removing it is probably best.

Issue #2010, opened 2019-08-26
Silence creation is broken when using the grammarly extension


tivvit:
This is probably an issue with Grammarly, but I thought it is better to let others know.

Grammarly for Chrome
Version
14.926.5955

alertmanager: 0.18.0

What happened?
When I try to create silence preview button does not work. The whole web GUI is broken (i.e. silence creation block is present on the alerts tab).

After disabling Grammarly extension everything works as expected

simonpasquier:
It would probably better be directed to Grammarly. We understand that you want to make AlertManager's users aware of it (appreciated!) but it looks like there is little we can do.

stephenreddek:
@simonpasquier I think this can be mitigated by disabling grammarly on each textarea per https://github.com/elm/html/issues/44#issuecomment-534665947 ? Would you accept a PR that does that?

simonpasquier:
I'll defer to @stuartnelson3 and @w0rm whether this is ok or not.

w0rm:
@stephenreddek solution looks good to me!

simonpasquier:
Given that we had the issue reported twice already (#2010 and https://github.com/prometheus/alertmanager/issues/2022), I think it is worth having this change.

w0rm:
Fixed in #2061

PR #2061, opened 2019-10-03
Disable the grammarly plugin on all textareas - Fix 2010

stephenreddek:
Provides a fix for #2010 . This adds an attribute which the grammarly plugin respects to all textareas. This effectively keeps the plugin from inserting any nodes and causing the virtual dom from crashing.

@w0rm


brian-brazil:
I'm not sure it makes sense to try to work around each problematic 3rd party tool that breaks our pages one by one.

w0rm:
There is an [ongoing investigation](https://discourse.elm-lang.org/t/runtime-errors-caused-by-chrome-extensions/4381) about different browser extensions that can mess with the page. Hopefully the majority would be addressed by changes in Elm but this usually takes time.

In the meantime it doesn’t really hurt to address this one with just a few lines of code and make someone’s life happier :)

simonpasquier:
Given that we had the issue reported twice already (#2010 and https://github.com/prometheus/alertmanager/issues/2022), I think it is worth having this change.

PR #2074, opened 2019-10-18
Optimize etcd reads writes

dcwangmit01:


dcwangmit01:
Ah, darn.  Cancel.  I think I was faster than the github UI changing my branch.

PR #2081, opened 2019-10-23
At least one non-empty silence matcher

### LINKED ISSUE #1897: Add check that at least one silence matcher matches non-empty
To avoid accidentally silencing everything, I propose that similar to PromQL selectors that we check that one matcher matches non-empty. Those who explicitly want to silence everything can still match on `alertname="~.+"`

qoops-1:
Closes #1897

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -434,6 +434,18 @@ func validateMatcher(m *pb.Matcher) error {
 	return nil
 }
 
+func matchesEmpty(m *pb.Matcher) bool {
+	switch m.Type {
+	case pb.Matcher_EQUAL:
+		return m.Pattern == ""
+	case pb.Matcher_REGEXP:
+		matched, _ := regexp.MatchString(m.Pattern, "")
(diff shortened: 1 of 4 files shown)


silence/silence.go
```diff
@@ -434,17 +434,34 @@ func validateMatcher(m *pb.Matcher) error {
...
 	}
+	allMatchEmpty := true
 	for i, m := range s.Matchers {
 		if err := validateMatcher(m); err != nil {
 			return fmt.Errorf("invalid label matcher %d: %s", i, err)
 		}
+		allMatchEmpty = allMatchEmpty && matchesEmpty(m)
+	}
+	if allMatchEmpty {
+		return errors.New("at least one matcher must not match empty string")
```
simonpasquier: (nit) "at least one matcher must not match the empty string"

qoops-1: Fixed

simonpasquier:
Thanks!

PR #2087, opened 2019-11-01
cluster: use atoi instead.

johncming:


stuartnelson3:
we're specifically parsing using a uint for a reason, since we don't want to accidentally allow a negative port value

Issue #2084, opened 2019-10-25
UI: filtering on receiver fails when the receiver contains special characters like '+' 


simonpasquier:
While investigating an issue, I've discovered that the UI doesn't work as expected when the receiver name contains meta-characters such as `+`. This is because the API v2 specification says that the `receiver` parameter is interpreted as a regular expression.

![Screenshot_2019-10-25 Alertmanager](https://user-images.githubusercontent.com/2587585/67564635-abd09380-f723-11e9-918f-f08970db22a9.png)

![Screenshot_2019-10-25 Alertmanager(1)](https://user-images.githubusercontent.com/2587585/67564634-abd09380-f723-11e9-9dc2-216d6741cb28.png)

![Screenshot_2019-10-25 Alertmanager(2)](https://user-images.githubusercontent.com/2587585/67564632-abd09380-f723-11e9-8149-b92cfb0006f7.png)

simonpasquier:
Closed by #2090

PR #2090, opened 2019-11-05
fix receiver regex

m-masataka:
I think "/ ... /g" disturb correct matching. 
Remove unnecessary characters.

Related to #2084 .


simonpasquier:
Thanks!

PR #2093, opened 2019-11-06
notifier: add notifier that logs alerts to stdout or file

jackhammer2k:
Adds a notifier that logs alerts to a file or standard out. Very useful if uploaded to a log service like Elastic stack or similar to get a history about the alerts. Although we have time series about firing alerts, we loose information encoded in text (e.g. summary). Especially alerts not coming from Prometheus may contain relevant information as annotation, which is lost otherwise.

Features like log rotation should be done externally (e.g. logrotate).


brian-brazil:
You can use the existing webhook for this.

simonpasquier:
Thanks for suggesting this change. Unfortunately, we are currently not accepting new integrations. As Brian noted, this can be achieved using the existing webhook integration (BTW there's already [alertmanager2es](https://github.com/cloudflare/alertmanager2es) if you want to send notifications to Elasticsearch).

We can only provide the stability and performance we want to provide if we can properly maintain the codebase. This includes, amongst others, test integrations in an automated and scalable fashion. For this reason, we are suggesting people integrate with the help of our generic interfaces. We have an integrations page on which integrations using our generic interfaces are listed.

Even if existing integrations can not be tested in an automated fashion, we will not remove them for reasons of compatibility. This also means that any additions we take on, or any changes to existing integrations we make or accept, will mean maintaining and testing these until at least the next major version, realistically even beyond that.

Feel free to question this answer on our [developer mailing list](https://groups.google.com/forum/#!forum/prometheus-developers), but be aware it's unlikely that you will get a different answer.

jackhammer2k:
Thanks for the detailed explanation. I fully agree that you have to keep the amount of maintained code manageble and do provide generic interfaces to externalize custom integrations. However I have to say that this one might be an exception you should think about to still integrate. 

- Its just as generic as the webhook notifier. The webhook notifier uses common HTTP REST communication to allow external integrations. In the same manner the log notifier uses common logging, which Alertmanager already provides anyway, to allow external integrations if necessary at all. Thus custom logging products adaptions and log(-file) handling are externalized as well.
- It adds just a few lines of code to maintain.
- Its an easy win to be able to say that the Alertmanager provides a history of alerts.

jackhammer2k:
Meanwhile we implemented an external service receiving and logging webhook messages:
https://github.com/tomtom-international/alertmanager-webhook-logger

Maybe that one can be mentioned as "integration" in docs.

simonpasquier:
> Maybe that one can be mentioned as "integration" in docs.

Sure feel free to open a PR (see https://github.com/prometheus/docs/pull/1525 for an example).

PR #2094, opened 2019-11-06
template: add the method RemoveKeys to KV

vrischmann:
KV.Remove is not callable inside a template because it needs a slice of string which can't be created in a template.
Introduce the KV.RemoveKeys method which takes a vararg of strings and then calls KV.Remove.

I came across this problem because I was trying to do this in my Slack template:
```
{{ define "__subject" }}{{ if .GroupLabels.type }}{{ .GroupLabels.type }} {{ end }} {{with .GroupLabels}}{{.Remove "type"}}{{ .SortedPairs.Values | join " / " }}{{end}}{{end}}
```

except this doesn't work because of this error:
```
template: default.tmpl:6:223: executing \"__subject\" at <\"type\">: can't handle \"type\" for arg of type []string
```
As far as I can tell there's no way to create a slice/array in a Go template so right now `Remove` is completely unusable inside a template.

This PR adds a method `RemoveKeys` which takes a vararg string and simply calls Remove itself.

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -195,6 +195,11 @@ func (kv KV) SortedPairs() Pairs {
 	return pairs
 }
 
+// RemoveKeys returns a copy of the key/value set without the given keys.
+func (kv KV) RemoveKeys(key ...string) KV {
+	return kv.Remove(key)
+}
+
 // Remove returns a copy of the key/value set without the given keys.
 func (kv KV) Remove(keys []string) KV {
 	keySet := make(map[string]struct{}, len(keys))
--- a/template/template_test.go
+++ b/template/template_test.go
@@ -77,6 +77,20 @@ func TestKVSortedPairs(t *testing.T) {
 	}
 }
 
+func TestKVRemoveKeys(t *testing.T) {
+	kv := KV{
+		"key1": "val1",
+		"key2": "val2",
+		"key3": "val3",
+		"key4": "val4",
+	}
+
+	kv = kv.RemoveKeys("key2", "key4")
+
+	expected := []string{"key1", "key3"}
+	require.EqualValues(t, expected, kv.Names())
+}
+
 func TestKVRemove(t *testing.T) {
 	kv := KV{
 		"key1": "val1",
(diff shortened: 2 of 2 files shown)


brian-brazil:
It looks to me like you could handle this with a loop and an if to remove
the label.

vrischmann:
For my real use case I need to pipe `join " / "` to the result, how would I do that using a loop ?

Also, from some limited testing using a loop/if to filter keys is really cumbersome and unclear, I'd much rather have access to the Remove method.

brian-brazil:
> how would I do that using a loop ?

Only output the / on 2nd and subsequent iterations.

vrischmann:
True but that makes it even worse than the simple loop/if which is already a mess compared to simply having the ability to call `Remove`.

Am I right in saying `Remove` was/is intended to be used in the template ?

brian-brazil:
> Am I right in saying Remove was/is intended to be used in the template ?

Yes, but that it's in a template isn't your issue here. You need to pass a slice of strings, and there's no way to construct that.

vrischmann:
I don't really understand. If `Remove` is intended to be used in a template but there's just no way to use it, to me it sounds like it's broken, which is why I suggested to add another method making it usable.

brian-brazil:
It is usable in a template, if you use one of the places that there's already a slice of strings. See https://github.com/prometheus/alertmanager/blob/master/template/default.tmpl#L4

vrischmann:
ok, my mistake, I missed that.

So you were right, the problem is that I can't create a slice of strings in a template. Is there any reason to not add this `RemoveKeys` method which would work around this problem ?

brian-brazil:
I'd prefer to handle it more generally by allowing creation of slices of strings, rather than having to add functions for two different types everywhere.

vrischmann:
Understood. Would this be a function in the [DefaultFuncs](https://github.com/prometheus/alertmanager/blob/master/template/template.go#L127-L144) ? 

if that's all that's needed I can make a PR.

brian-brazil:
Yes, it would be.

PR #2101, opened 2019-11-12
template: add the default template function stringSlice

vrischmann:
Since it's impossible to create a string slice in a Go template by
default, add a function to work around this problem.

The use case is to make it easy to call KV.Remove with an arbitrary
slice inside a template.

This is a follow up to https://github.com/prometheus/alertmanager/pull/2094


brian-brazil:
Thanks!

The docs over in the docs repo will also need updating.

PR #2105, opened 2019-11-18
nflog: write as func and fix locker usage.

johncming:
1. Closures are better written as functions.
2. Defer breaks the use of lock-unlock


stuartnelson3:
@simonpasquier thoughts?

simonpasquier:
I don't feel strongly about closures vs. methods.

Regarding locking/unlocking, I don't see a problem right now. But the code you've submitted could have a deadlock:

```
	l.mtx.RLock()
	if le, ok := l.st[stateKey(q.groupKey, q.recv)]; ok {
                // No call to RUnlock().
		return []*pb.Entry{le.Entry}, nil
	}
	l.mtx.RUnlock()
```

stuartnelson3:
I guess I don't really see the benefit of this

johncming:
> I don't feel strongly about closures vs. methods.
> 
> Regarding locking/unlocking, I don't see a problem right now. But the code you've submitted could have a deadlock:
> 
> ```
> 	l.mtx.RLock()
> 	if le, ok := l.st[stateKey(q.groupKey, q.recv)]; ok {
>                 // No call to RUnlock().
> 		return []*pb.Entry{le.Entry}, nil
> 	}
> 	l.mtx.RUnlock()
> ```

1. yes, deadlock. I will fix it.
2. RUnlock undoes a single RLock call; it does not affect other simultaneous readers. It is a run-time error if rw is not locked for reading on entry to RUnlock.

Changed to a function, because this is not a typical scenario where closures are used, only `params`. See if this change is necessary, and if not, keep only the changes to lock.

PR #2110, opened 2019-11-20
Fix TestClusterJoinAndReconnect on macos

sylr:
Still trying to fix `github.com/prometheus/alertmanager/test/with_api_v2/acceptance`


simonpasquier:
Thanks!

PR #2125, opened 2019-12-02
pointer makes more sense for omitempty marshaler

00arthur00:
Signed-off-by: arthur yang <yang_yapo@126.com>


simonpasquier:
Thanks for the PR. The [YAML library](https://godoc.org/gopkg.in/yaml.v2#Marshal) is able to omit zero value structs. The JSON marshaling is only supported because we need it for the API v1 (eg once we remove the API v1, it can go away).
Overall I don't think it's worth the change.

00arthur00:
@simonpasquier  Do we have a deadline about V2?

00arthur00:
Thanks for your explaination.
Will close it.

Issue #1997, opened 2019-08-13
Feature Suggestion - Webhook integration should indicate timeout


vsliouniaev:
The webhook configuration timeout is dependent on configuration in alertmanager, which isn't visible to the endpoint receiving the request. It would be quite useful to have this information available in the receiving service, particularly where it acts as a proxy for integrating with something down-stream.

## Scenario
I have an in-house routing service for alerts that is gathering events from multiple systems. In order to integrate with Alertmanager, I have a service that transforms Alertmanager-format webhooks and sends these on to my in-house alerting service. I would like to set my transforming service timeouts to be lower than Alertmanager is using, omit any retry logic and rely on Alertmanager to perform retries as it normally would. I can do this already, if I know what the alertmanager settings are, but if I have multiple teams that have their own configurations, this becomes more complex to deal with.

This approach would allow me to just use the alertmanager-indicated information to set my timeout.

## Possible solution
Use a query parameter, e.g. `?timeout=30s` in order to indicate this information from Alertmanager, in the same way certain Kubernetes requests behave.

simonpasquier:
Prometheus sends a `X-Prometheus-Scrape-Timeout-Seconds` header when scraping targets. I suppose we could take the same approach.

johncming:
> Prometheus sends a `X-Prometheus-Scrape-Timeout-Seconds` header when scraping targets. I suppose we could take the same approach.

add timeout below, is ok?
https://github.com/prometheus/common/blob/b5fe7d854c42dc7842e48d1ca58f60feae09d77b/config/http_config.go#L75-L86

PR #2132, opened 2019-12-06
notify/webhook: add timeout config for hints. #1997

johncming:
Only the timeout option is added for webhook.

If modify 'HTTPClientConfig', some notifier may not support `X-Prometheus-Scrape-Timeout-Seconds`  in http headers.

https://github.com/prometheus/common/blob/b5fe7d854c42dc7842e48d1ca58f60feae09d77b/config/http_config.go#L75-L86

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -20,8 +20,8 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-
 	commoncfg "github.com/prometheus/common/config"
+	"github.com/prometheus/common/model"
 )
 
 var (
@@ -409,6 +409,9 @@ type WebhookConfig struct {
 
 	// URL to send POST request to.
 	URL *URL `yaml:"url" json:"url"`
+
+	// Timeout to tell server cancel response after timeout
+	Timeout model.Duration `yaml:"timeout" json:"timeout"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
--- a/notify/webhook/webhook.go
+++ b/notify/webhook/webhook.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"time"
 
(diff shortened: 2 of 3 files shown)


simonpasquier:
Thanks for tackling this. IMO, we don't need a `Timeout` setting in the webhook configuration. The timeout is already set when Alertmanager processes the alerts group, it is carried in the `context.Context` and you can compute it from `ctx.Deadline()`.

Also we want a header name specific to Alertmanager rather than reusing `X-Prometheus-Scrape-Timeout-Seconds`. Maybe `X-Alertmanager-Timeout-Seconds`?

cc @vsliouniaev

brian-brazil:
Hmm, what would a client do with the knowledge that the Alertmanager has timed out? Cancelling an attempt to send a notification that was just about to work seems counter-productive.

vsliouniaev:
I believe this is addressing this issue: https://github.com/prometheus/alertmanager/issues/1997, though the name of the header here isn't intuitive

johncming:
> X-Alertmanager-Timeout-Seconds

You mentioned two parts:
1) the name of the header, how about `X-Alertmanager-Notify-Timeout-Seconds` like `X-Prometheus-Scrape-Timeout-Seconds` with extra verb.

2) Whether need to set a timeout in config or use the remaining timeout of the server.

johncming:
> Hmm, what would a client do with the knowledge that the Alertmanager has timed out? Cancelling an attempt to send a notification that was just about to work seems counter-productive.

`webhook` feature is very common, and the potential extensibility is very good. Can it  consider optimizing them to use them differently from other methods like email? 

The timeout here is more of a hint.

brian-brazil:
My question is is there actually a good use case for this, or will users end up blindly following typical timeout patterns and end up making their notifications less reliable. Keep in mind that we have users who misunderstand what a 1s group intervals means, and presumably wouldn't also get its impact on timeouts.

PR #2136, opened 2019-12-11
move walkRoute to dispatch package.

johncming:
move walkRoute to dispatch package for better encapsulation and add test case.

### DIFF
--- a/dispatch/route_test.go
+++ b/dispatch/route_test.go
@@ -267,3 +267,88 @@ routes:
 		}
 	}
 }
+
+func TestRouteWalk(t *testing.T) {
+	in := `
+receiver: 'notify-def'
+
+routes:
+- match:
+    owner: 'team-A'
+
+  receiver: 'notify-A'
+
+  routes:
+  - match:
+      env: 'testing'
+
+    receiver: 'notify-testing'
+    group_by: [...]
+
+  - match:
+      env: "production"
+
+    receiver: 'notify-productionA'
+    group_wait: 1m
+
+    continue: true
+
+  - match_re:
+      env: "produ.*"
+      job: ".*"
+
+    receiver: 'notify-productionB'
+    group_wait: 30s
+    group_interval: 5m
+    repeat_interval: 1h
+    group_by: ['job']
+
+
+- match_re:
+    owner: 'team-(B|C)'
+
+  group_by: ['foo', 'bar']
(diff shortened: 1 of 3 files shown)


dispatch/route_test.go
```diff
@@ -267,3 +267,90 @@ routes:
...
+
+	var ctree config.Route
+	if err := yaml.UnmarshalStrict([]byte(in), &ctree); err != nil {
+		t.Fatal(err)
+	}
+	tree := NewRoute(&ctree, nil)
+
+	var got, expected []string
+
+	// DFS
```
simonpasquier: This can be removed

dispatch/route_test.go
```diff
@@ -267,3 +267,90 @@ routes:
...
+      group_wait: 2m
+`
+
+	var ctree config.Route
+	if err := yaml.UnmarshalStrict([]byte(in), &ctree); err != nil {
+		t.Fatal(err)
+	}
+	tree := NewRoute(&ctree, nil)
+
+	var got, expected []string
```
simonpasquier: lets move `var got |]string` right before calling `tree.Walk()`.

dispatch/route_test.go
```diff
@@ -267,3 +267,90 @@ routes:
...
+
+	// DFS
+	expected = []string{
```
simonpasquier: `expected = []string{` -> `expected := []string{`

dispatch/route_test.go
johncming: ok.

dispatch/route_test.go
johncming: ok.

dispatch/route_test.go
johncming: ok

simonpasquier:
The DCO is failing.

dispatch/route_test.go
simonpasquier: still to be removed :smirk:

johncming:
> The DCO is failing.

@simonpasquier fixed

simonpasquier:
Please remove this [comment](https://github.com/prometheus/alertmanager/pull/2136/files#diff-8d3bcd8a072f2a0eb4586f92a18fa168R334) and it will be good to go. Thanks!

simonpasquier:
Thanks!

Issue #2121, opened 2019-11-30
Webhook Notifier 429 Handling


justinwatkinson:
**What did you do?**

Instrumented a webhook receiver to receive Alertmanager payloads to forward to a 3rd party system.

**What did you expect to see?**

When the destination webhook returns a 429 error, the retry logic with back-off should be applied.

**What did you see instead? Under which circumstances?**

429 is treated the same as a 404 or other client error, which does longer retries (10 sec according to my local tests).

**Environment**

* System information:

Linux 4.15.0-54-generic x86_64

* Alertmanager version:

Master (89db9090177c13e1ee5e72ae81e1713a6288a637) & 0.19.0

* Prometheus version:

N/A

* Alertmanager configuration file:
```
global:
  resolve_timeout: 5m

route:
  group_by: ['alertname']
  group_wait: 10s
  group_interval: 10s
  repeat_interval: 1h
  receiver: 'web.hook'
receivers:
- name: 'web.hook'
  webhook_configs:
  - url: 'http://127.0.0.1:9001/'

```

* Prometheus configuration file:
N/A

* Logs:
```
level=error ts=2019-11-30T03:20:43.832Z caller=notify.go:372 component=dispatcher msg="Error on notify" err="unexpected status code 429: http://127.0.0.1:9001/" context_err="context deadline exceeded"
level=error ts=2019-11-30T03:20:43.832Z caller=dispatch.go:301 component=dispatcher msg="Notify for alerts failed" num_alerts=1 err="unexpected status code 429: http://127.0.0.1:9001/"
level=error ts=2019-11-30T03:20:53.833Z caller=notify.go:372 component=dispatcher msg="Error on notify" err="Post http://127.0.0.1:9001/: dial tcp 127.0.0.1:9001: connect: connection refused" context_err="context deadline exceeded"
level=error ts=2019-11-30T03:20:53.833Z caller=dispatch.go:301 component=dispatcher msg="Notify for alerts failed" num_alerts=1 err="Post http://127.0.0.1:9001/: dial tcp 127.0.0.1:9001: connect: connection refused"
```

**Sample 429 Code**
```
package main

import (
	"log"
	"net/http"
)

func handler(rw http.ResponseWriter, r *http.Request) {
	log.Println("Received call")
	rw.WriteHeader(http.StatusTooManyRequests)
}

func main() {
	http.HandleFunc("/", handler)
	http.ListenAndServe(":9001", nil)
}
```

Sample Logs from http server
```
[...]

PR #2145, opened 2019-12-19
Allow retry for 429 on webhook receiver

justinwatkinson:
To address #2121 - proposing patch which adds 429 response to the retry codes for webhooks only.

### DIFF
--- a/notify/webhook/webhook.go
+++ b/notify/webhook/webhook.go
@@ -60,6 +60,7 @@ func New(conf *config.WebhookConfig, t *template.Template, l log.Logger) (*Notif
 			CustomDetailsFunc: func(int, io.Reader) string {
 				return conf.URL.String()
 			},
+			RetryCodes: []int{http.StatusTooManyRequests},
 		},
 	}, nil
 }
--- a/notify/webhook/webhook_test.go
+++ b/notify/webhook/webhook_test.go
@@ -15,6 +15,7 @@ package webhook
 
 import (
 	"fmt"
+	"net/http"
 	"net/url"
 	"testing"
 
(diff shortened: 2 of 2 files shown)


roidelapluie:
This behavior should be optional and opt-in.

brian-brazil:
This shouldn't be configurable at all, we should have one consistent protocol for webhook. If you want the Alertmanager to retry, a 5xx will do it.

justinwatkinson:
Indeed, 5xx is the state of the world today for some of the things I work on.  It became a discussion on the semantics of a 429 in other contexts including outside of Prometheus.

When I first encountered this, it took a while to dig through the Google group to get to the 5xx answer.  Perhaps a simple doc improvement to clarify this protocol improvement is in order instead?

brian-brazil:
Do we really need to document the basic meaning of HTTP status codes?

justinwatkinson:
What is the desired behavior if Retry-After is included in the 429 response?  I believe as of today its ignored since it is not mandatory, and perhaps that's the most correct way for Alertmanager as the client of the webhook to react.

roidelapluie:
That code does not take care about the Retry-After anyway, so in a way @brian-brazil is right. On the other way, if you want to send your alerts to systems that are not alertmanager specific they could reply with a 429. In that case, should it be the responsibility of the admin to ensure no 429 can come from those?

roidelapluie:
I think that in this case it might be worth it to make it as an option - it will probably not cost much to maintain and might help some users. Now the option could be retry_429 or retriable_codes (default: [5xx]).

justinwatkinson:
Right... I borrowed most of it from the Slack receiver before it was reverted, which also didn't handle Retry-After IIRC.  Feels like it's a good time to kill this PR and continue discussion in the issue anyways.

roidelapluie:
Yes that is why it got reverted ... because it did not respect the retry after :)

brian-brazil:
Keep in mind that webhook receivers are intended to be designed specifically to work with what the Alertmanager sends them, so by that logic the Alertmanager shouldn't need any configuration to decide how to deal with the responses either.

simonpasquier:
I'm also in favor of keeping the webhook notifier simple and stick with retrying only on 5xx.
Alertmanager could handle in a better way notification systems such as Slack that indicate when it's safe to retry.

PR #2155, opened 2020-01-06
email: generate message ID headers

ddevault:
This GenerateMessageId is taken from go-message:

https://github.com/emersion/go-message/blob/master/mail/mail.go

I am the original author of this file; I agree to distribute it in
alertmanager regardless of the original license.

Signed-off-by: Drew DeVault <sir@cmpwn.com>

### DIFF
--- a/notify/email/email.go
+++ b/notify/email/email.go
@@ -16,7 +16,9 @@ package email
 import (
 	"bytes"
 	"context"
+	"crypto/rand"
 	"crypto/tls"
+	"encoding/binary"
 	"fmt"
 	"math/rand"
 	"mime"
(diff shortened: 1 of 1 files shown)


ddevault:
I'm not familiar with dep. How do I update this to account for the new dependency?

simonpasquier:
Thanks for the PR!

For vendoring, you would need to run `make unused` (or `GO111MODULE=on go mod vendor`) and commit any change in the `vendor/` directory.

The Message-ID header is already set before the code you've added so your change would need to go there:

https://github.com/prometheus/alertmanager/blob/b4ac213809a96fe6d834303d61a60bc71a9d356a/notify/email/email.go#L248-L250

And you can notice that the hostname is retrieved only once when the struct is initialized (eg no need to call `os.Hostname()` for every notification).

ddevault:
Hm, I wonder why the existing message ID code wasn't working as expected. I should have caught that. I'll investigate further.

simonpasquier:
this was added quite recently (in the latest release in fact) so maybe you were running an older version of Alertmanager

ddevault:
That'll be it, I'm on 0.19.Cheers!

PR #2162, opened 2020-01-15
Add a method that returns a sorted list of a KV unique values

machine424:
Signed-off-by: ayoub mrini <amrini@wiremind.fr>

I don't know if one can do that using a simple combination of already-existing functions.

Made it easier to get unique values of a KV in a template, in cases where we group alerts by `job` for example and we need to get, in a template, the unique firing alerts' names, I think this question: https://stackoverflow.com/questions/59202685/alertmanager-templates-remove-duplicate-alertnames-for-grouped-notification gives more detail.

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -224,6 +224,21 @@ func (kv KV) Values() []string {
 	return kv.SortedPairs().Values()
 }
 
+// UniqValues returns a sorted list of unique values in the LabelSet.
+func (kv KV) UniqValues() []string {
+	vs := kv.Values()
+	vsCopy := make([]string, len(vs))
+	copy(vsCopy, vs)
+	sort.Strings(vsCopy)
+	uniqVs := vsCopy[:0]
+	for _, v := range vsCopy {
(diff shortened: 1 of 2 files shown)


brian-brazil:
This function wouldn't help you do that, a KV would be a single alert.

If you've so many alerts that this matters I'd suggest looking at either changing your group_by, or reducing your alerting volume down to something more reasonable.

machine424:
You could say it; this method makes no sense for a `KV` given the fact that alerts' names go under the same key `alertname` ...
I can implement it for an array of `KV` or for `Alerts` but it'll not be of a big utility.
Adding the `alertname` in `group_by` means more emails and I cannot reduce the alert volume.
Thanks for your answer. I'll close this.
a last question: I don't know how Go templates evaluation works, can we range over `Alerts` get all `alertname`, pass them to `stringSlice` DefaultFunc to get an array, in this case we will only need to filter out duplicates ?

brian-brazil:
Go templates are not very advanced, any heavy processing is meant to be done before the data gets to them. Unfortunately they were the only real option back when templating was first added. It's rare you can't do something that's good enough though.

PR #2173, opened 2020-01-27
[Minor][one line change] Fix an error message about start and end time validation. EOM

celian-garcia:

PR #2179, opened 2020-02-06
api/v2: Fix silence creation error message

simonpasquier:
@celian-garcia I need to check that the CI pipeline is still working after a configuration change I did hence I'm submitting the follow-up for #2173


celian-garcia:
IMHO the best message and less confusing message would be 

``failed to create silence: start time must be before end time``

celian-garcia:
Haha you just set the message to ``failed to create silence: end time must be before start time`` !!!

The right message is ``failed to create silence: start time must be before end time`` 

Start must be **before** end XD

simonpasquier:
Thanks! I must be tired :sleepy:

PR #2188, opened 2020-02-18
notify: add retry field to debug log

simonpasquier:


notify/notify.go
```diff
@@ -665,7 +665,7 @@ func (r RetryStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.Ale
 			r.metrics.numNotifications.WithLabelValues(r.integration.Name()).Inc()
 			if err != nil {
 				r.metrics.numFailedNotifications.WithLabelValues(r.integration.Name()).Inc()
-				level.Debug(l).Log("msg", "Notify attempt failed", "attempt", i, "integration", r.integration.Name(), "receiver", r.groupName, "err", err)
+				level.Debug(l).Log("msg", "notify attempt failed", "attempt", i, "integration", r.integration.Name(), "receiver", r.groupName, "retry", retry, "err", err)
```
brian-brazil: We can use capital letters, this isn't an `error` which might get prefixed.

PR #2200, opened 2020-03-06
Fixed dispatcher metrics registration

pracucci:
The new dispatcher metrics are not registered on the provided registerer, but on the default one. This PR fixes it, as well as handle the case the input registerer is `nil`.


roidelapluie:
Thanks!

simonpasquier:
Not supporting `nil` registries was on purpose. It allows to detect when you accidentally forgot to pass a registry.

See https://github.com/prometheus/alertmanager/pull/2202

PR #2202, opened 2020-03-10
dispatch: don't allow nil registry for metrics

simonpasquier:


roidelapluie:
This is a patter we have in prometheus too. Why do you want to get rid of it?

beorn7:
Indeed. It helps for testing and such.

I don't quite buy the [argument of accidentally passing `nil`](https://github.com/prometheus/alertmanager/pull/2200#issuecomment-597014447). Writing `NewDispatcherMetrics(nil)` is quite explicit and shouldn't happen accidentally (or you could say that even if `nil` is disallowed, you could still "accidentally" pass in a no-op registry).

simonpasquier:
Oki doki. I'll close it then...

PR #2203, opened 2020-03-10
api/v2: return empty array of peers when disabled

simonpasquier:
Before this change, the API returns a `null` value for `peers`:

```
$ curl -s localhost:9093/api/v2/status | jq '.cluster'
{           
  "peers": null,           
  "status": "disabled"  
}  
```

After

```
{           
  "peers": [],           
  "status": "disabled"  
}
```

### DIFF
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -166,6 +166,7 @@ func (api *API) getStatusHandler(params general_ops.GetStatusParams) middleware.
 		},
 		Cluster: &open_api_models.ClusterStatus{
 			Status: &status,
+			Peers:  []*open_api_models.PeerStatus{},
 		},
 	}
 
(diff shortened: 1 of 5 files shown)


brian-brazil:
Why are you touching v2? Noone should be using that, and it's not maintained.

roidelapluie:
Just to make sure @simonpasquier , did you check that the UI still works as expected with this?

thanks

simonpasquier:
@mxinden 

> I guess given that the implementation diverges from the specification this is a bug fix and not a breaking change?

Yes. AFAICT there's no way we can tell the code generated from the OpenAPI spec to return an empty array instead of `nil`.

> did you check that the UI still works as expected with this?

@roidelapluie yes

PR #2206, opened 2020-03-10
add shebang for test script

zgfh:
add shebang for test script  to fix goland run script error
image

### DIFF
--- a/examples/ha/send_alerts.sh
+++ b/examples/ha/send_alerts.sh
@@ -1,3 +1,4 @@
+#!/usr/bin/env bash
 alerts1='[
   {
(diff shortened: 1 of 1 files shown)


examples/ha/send_alerts.sh
```diff
@@ -1,3 +1,4 @@
+#!/bin/bash
```
simonpasquier: can you replace this with `#!/usr/bin/env bash`?

zgfh: updated

simonpasquier:
Thanks!

Issue #2182, opened 2020-02-12
Race condition in dispatcher code


jtlisi:
I noticed a mild race condition in the dispatch package.

I wrote a small test in the `dispatch` package to demonstrate a race condition in the dispatch code:

Test Added to `dispatch/dispatch_test.go:

```go
func TestDispatcherRace(t *testing.T) {
	logger := log.NewNopLogger()
	marker := types.NewMarker(prometheus.NewRegistry())
	alerts, err := mem.NewAlerts(context.Background(), marker, time.Hour, logger)
	if err != nil {
		t.Fatal(err)
	}
	defer alerts.Close()

	timeout := func(d time.Duration) time.Duration { return time.Duration(0) }
	dispatcher := NewDispatcher(alerts, nil, nil, marker, timeout, logger, NewDispatcherMetrics(prometheus.NewRegistry()))
	go dispatcher.Run()
	dispatcher.Stop()
}
```

Test Output:
```bash
$ go test ./dispatch -run TestDispatcherRace -race
==================
WARNING: DATA RACE
Write at 0x00c0002e0580 by goroutine 26:
  github.com/prometheus/alertmanager/dispatch.(*Dispatcher).Run()
      /Users/jtlisi/tmp/alertmanager/dispatch/dispatch.go:112 +0x1d7

Previous read at 0x00c0002e0580 by goroutine 20:
  github.com/prometheus/alertmanager/dispatch.TestDispatcherRace()
      /Users/jtlisi/tmp/alertmanager/dispatch/dispatch.go:254 +0x677
  testing.tRunner()
      /usr/local/Cellar/go/1.13.6/libexec/src/testing/testing.go:909 +0x199

Goroutine 26 (running) created at:
  github.com/prometheus/alertmanager/dispatch.TestDispatcherRace()
      /Users/jtlisi/tmp/alertmanager/dispatch/dispatch_test.go:537 +0x58e
  testing.tRunner()
      /usr/local/Cellar/go/1.13.6/libexec/src/testing/testing.go:909 +0x199

Goroutine 20 (finished) created at:
  testing.(*T).Run()
      /usr/local/Cellar/go/1.13.6/libexec/src/testing/testing.go:960 +0x651
  testing.runTests.func1()
      /usr/local/Cellar/go/1.13.6/libexec/src/testing/testing.go:1202 +0xa6
  testing.tRunner()
      /usr/local/Cellar/go/1.13.6/libexec/src/testing/testing.go:909 +0x199
[...]

PR #2208, opened 2020-03-12
Fix race condition in dispatcher

jtlisi:
Fixes: #2182 

By ensuring operations on the `cancel` and `ctx` field are wrapped with a mutex this removes the race condition from the dispatcher.

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -258,7 +258,7 @@ func (d *Dispatcher) Stop() {
 		return
 	}
 	d.mtx.Lock()
-	if d == nil || d.cancel == nil {
+	if d.cancel == nil {
 		return
 	}
 	d.cancel()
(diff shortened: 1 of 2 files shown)


jtlisi:
@simonpasquier I added the test from the original issue. Still not passing the CI test cases but the failures look unrelated to this PR.

simonpasquier:
Thanks! The CI is a bit flaky sometimes. I've restarted it and will merge on green.

dispatch/dispatch.go
```diff
@@ -255,11 +254,13 @@ func (d *Dispatcher) Groups(routeFilter func(*Route) bool, alertFilter func(*typ
 
 // Stop the dispatcher.
 func (d *Dispatcher) Stop() {
+	d.mtx.Lock()
```
simonpasquier: Ah the failure is legit! Alertmanager crashes on startup because the `Stop()` function is called with a `nil` receiver. We need this:

```go
       if d == nil {
           return
       }
	d.mtx.Lock()
	if d.cancel == nil {
            d.mtx.Unlock()
	    return nil
	}
        ...
```

simonpasquier:
Thanks!

Issue #2201, opened 2020-03-09
Feature Request: add an optional not_equal field to inhibit_rules


fredhaug:
The alerts in AlertManager may have severity labels to route alerts to the right receivers. It should also be possible to inhibit an alert with a lower severity level if an alert with a higher severity level and otherwise equal labels is present, e.g.:

```
inhibit_rules:
  - source_match:
      severity: 'critical'
    target_match:
      severity: 'warning'
    not_equal: ['severity']
```

It does not work to simply put `equal: ['alertname']` because the alert may consist of other labels that should be considered. Currently, it is possible to specify the labels that should be equal, but it should also be possible to specify a list of labels that should not be equal, so that the other labels can be compared for equality without necessarily knowing the label names in advance.

beorn7:
As a data point: At SoundCloud, we traditionally had an inhibit rule in place for exactly that purpose, and we just listed a bunch of typical labels in the `equal` list. @fredhaug is right that this wasn't perfect. An alert could be the same for all those labels but still alert for something different (e.g. a disk-full alert on the same node might fire as critical for one partition and as warning for another partition, and the corresponding `partition` label might not have been anticipated when creating the list). For that use case, it would indeed be better to have an option available similar to what @fredhaug suggests. I'm just not happy with the `not_equal` wording. It would also be mutually exclusive with the `equal` field. (It wouldn't make sense to provide both.) How about not creating a new field but let the existing `equal` field understand something like `equal: ['...']` (which we already have for `group_by`). In this case, it would mean “all labels must be equal except those mentioned in both `source_match` and `target_match`”.

brian-brazil:
> except those mentioned in both source_match and target_match

This feels off, why would we exclude these labels? We don't consider them when deciding what to include.

beorn7:
They are different (almost) by definition. Example above with my suggested syntax:
```
inhibit_rules:
  - source_match:
      severity: 'critical'
    target_match:
      severity: 'warning'
    not_equal: ['...']
```

If `severity` was included, the inhibition rule would never inhibit anything because `severity` will never be equal between source and target alert.

PR #2214, opened 2020-03-20
docs: Add warning that inhibition occurs on missing `equal`

Dominik-K:
I would say mentioning this wherever possible would help that others get to know this as early as possible. The [`simple.yml` is also referenced from the documentation](https://github.com/prometheus/docs/blob/master/content/docs/alerting/configuration.md#L52-L53). Means it's part of the documentation itself, isn't it?

We've learned it the harder way: All services monitored with the `ServiceMonitor` from the [`Prometheus Operator`](https://github.com/coreos/prometheus-operator) get a `service` label. Therefore we've set `equal: ['service']`. Except alerts based on the [`kube-state-metrics` service](https://github.com/kubernetes/kube-state-metrics) which needed an extra `equal: ['container']` instead. We didn't know this while configuring this (relates to https://github.com/prometheus/alertmanager/issues/2201).
 
[Other users experienced the "default inhibition"](https://groups.google.com/forum/#!msg/prometheus-users/8EVrpGReoMY/F0lXEn4mBAAJ) with other examples, too. 

BTW: You already considered to link to the example config from the `Readme.md` instead of doubling it there? It's out of sync anyway (looking at the `equal` values).

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,6 +148,10 @@ inhibit_rules:
   target_match:
     severity: 'warning'
   # Apply inhibition if the alertname is the same.
+  # CAUTION: 
+  #   If all label names listed in `equal` are missing 
+  #   from both the source and target alerts,
+  #   the inhibition rule will apply!
   equal: ['alertname']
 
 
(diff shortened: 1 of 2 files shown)


simonpasquier:
Thanks!

PR #1819, opened 2019-03-29
[WIP] cluster: Secure cluster traffic via mutual TLS

mxinden:
Use github.com/mxinden/memberlist-tls-transport to optionally secure the
Alertmanager cluster traffic via mutual tls.

This is an early work-in-progress version, but one can already test it out locally. I added everything to get up and running quickly:

1. `go get -u github.com/cloudflare/cfssl/cmd/...`
2. `cd examples/ha/tls/ &&  make start`

Issue #1322, opened 2018-04-12
Securing the gossip protocol?


wrouesnel:
Is there anyway at the moment to add TLS authentication to the gossip protocol (or is it secure as-is?)

I need to comply with a policy that all traffic is encrypted by default within our infrastructure, so ideally there'd be support for specifying TLS server and client certificates on the command line to secure connections. Is this in the roadmap somewhere?

brian-brazil:
It might happen sometime in the future, but I wouldn't hold your breath.

hoffie:
I am also interested in this. Would it have to be TLS? It seems like the newly introduced `memberlist` implementation also supports AES-GCM encryption using the [SecretKey](https://godoc.org/github.com/hashicorp/memberlist#Config) Config option. This could be added somewhere in the memberlist initialization code in [cluster.go](https://github.com/prometheus/alertmanager/blob/master/cluster/cluster.go#L114).

Would this be an improvement which would be accepted by the project? If so, I might try to integrate and test this.

brian-brazil:
I'd prefer not to get into ad-hoc security mechanisms that some of our dependencies happen to support, that could bring with it the same maintenance challenges we're currently trying to avoid with auth elsewhere.

wrouesnel:
The trouble is that dependency does make securing it non-trivial since it likes to mix up UDP and TCP, so most of the usual tricks like putting stunnel in front of it don't work. It's kind of glaring because Prometheus everywhere else can be secured with TLS, but this can't.

Would you be open to a patch which (at least currently) traded some performance of the gossip protocol for allowing a set of regular TLS parameters? - i.e. a flag which enables encryption and limits the gossip library to using TCP only so regular TLS can be used? I wouldn't think that would constrain the future too much, since pretty much everything uses TLS in some form, and if you stick with gossip then nothing stops expanding the choice later to use whatever mechanism that allows.

brian-brazil:
I think options for our standard `tls_config` client options would be okay. That'd have to go in the config file somewhere.

mxinden:
@wrouesnel https://github.com/prometheus/alertmanager/pull/1763/ adds a design document as well as a prove of concept implementation to secure the gossip traffic. Please have a look. I would be interested in your thoughts.

lilic:
[...]

PR #2216, opened 2020-03-21
doc: Propose using Memberlist Keyring to protect a cluster

zecke:
Create a document to propose an easy (implementation and operation) way
to protect the production cluster from accidental and unwanted members.

Provide a reference implementation in addition to the design document.

TODO(zecke): Figure out how to test this feature properly.

Signed-off-by: Holger Hans Peter Freyther <automatic+am@freyther.de>


brian-brazil:
Thanks for your PR. We're looking at adding TLS generally, so don't want to add other auth systems.

zecke:
Thank you for your reply and sorry for being late to the party. I have seen the design document and wanted to propose a more simple design for a narrower problem. If we focus on integrity and authentication (e.g. something provided by an HMAC) and leave out confidentiality (e.g. ignore known plaintext in gossiped message) we end up with a solution orders of magnitude easier to implement and operate.

Going all in on TCP + X509 + TLS is nice but has certain consequences for operating an AM:

* Certificates will expire "unexpectedly". I listed three major companies not able to renew certificates in time. It's bound to happen for many users/orgs. The failure for AM will be less dramatic as it fails open but is a failure mode never the less.

* TLS is difficult to implement. Connections must be broken when certificates expire, are revoked... Time needs to be roughly synchronized (not sure what requirement on time we have today).

* TCP for everything brings wanted and unwanted side-effects. The exposure to head of line blocking is one of them.

mxinden:
Thanks for putting work into this and writing a design document.

Just for documentation purposes I am linking the initial issue https://github.com/prometheus/alertmanager/issues/1322 the [design doc for Membership over TLS](https://github.com/mxinden/alertmanager/blob/master/doc/design/secure-cluster-traffic.md) and the corresponding work-in-progress pull request https://github.com/prometheus/alertmanager/pull/1819 here.

> TCP for everything brings wanted and unwanted side-effects. The exposure to head of line blocking is one of them.

TCP head of line blocking is happening per connection. Given the low bandwidth usage of the gossip protocol I doubt this would be an issue. Please correct me if I am missing something.

Having a simple solution for the problem of distinct clusters merging would be great. On the contrary I do see the maintenance overhead of eventually maintaining two solutions.

TheMeier:
The TLS based securing of the cluster has been implemented in https://github.com/prometheus/alertmanager/issues/1322 in the meantime (still marked EXPERIMENTAL though).

Regarding certificate handling. A lot has changed here, automatics issuing (think ACME, vault, cert-manager) is much more common today.

@zecke do you still want to pursue this?

SoloJacobs (commented):
```diff
-Hi @zecke ,
-
-It seems that the TLS approach has since been implemented, thus this approach is obsolete. Do you feel like something is missing aka. would updating this PR provide any value?

+Seems like I missed @TheMeier comment, sorry!
```
Kind regards, Solomon Jacobs

Issue #2178, opened 2020-02-04
Enforce 512KB event size limit for PagerDuty events


shamilpd:
PagerDuty will soon start enforcing a 512 KB event size limit on both the V1 and V2 api (https://v2.developer.pagerduty.com/docs/events-api-v2). This limit has been mentioned in the documentation for a while but was not enforced.

We regularly receive events that breach this limit. After investigating we found that these large events are formed by giant text blobs from the `firing` and `resolving` fields (https://github.com/prometheus/alertmanager/blob/master/config/notifiers.go#L49-L50) in `details`. It appears that these fields can have the descriptions of an unbounded list of alerts. These lists can be viewed in the PagerDuty UI but it's not clear how valuable they are to the users when the list becomes very large (over a hundred).

We have a few proposals for fixing this issue:

1. The simplest option is to truncate the `firing` and `resolving` fields to 250KB each which will leave 12KB for the rest of the message.

2. Truncate `firing` to 500KB. If it needs less than 500KB, the rest is used for `resolving` (or reverse the order if `resolving` is deemed more important than `firing`). Basically we ensure that the space taken up by `firing` plus `resolving` doesn't exceed 500KB but without a hard boundary on each field.

3. The above two options have the possibility that the truncation can happen inside an alert description. If it's important to make sure that alert descriptions are always displayed in full, we can apply a limit on the number of alerts used to render `firing` and `resolving`. However, unless this limit is quite small, it won't do a good job of ensuring that the full message size is below 512KB because the size of alert descriptions can vary.

We prefer options one or two. Regardless of what option is chosen, there should also be a size check after the message is encoded into JSON. If other text fields are added to the message in the future, this check will make sure that overall message size is under 512KB.

simonpasquier:
We already do this kind of limit checks at other places but AFAICT it's never been the case for the posted payload itself. We can't rely on the `firing` and `resolving` keys being present in the custom details as the user is able to change this from the configuration.
I'd rather check the size of the message to be posted. If it is above the 512KB limit, we replace the `Custom Details` field (which is likely the offender) with something like `{"warning": "custom details not included because otherwise the message would exceed the maximum size of 512KB."}`.

shamilpd:
That sounds good to me! 
I will get started on a PR for this soon. 
Do you know when the next alertmanager release is planned to come out?

roidelapluie:
There is no fixed planning at this stage as the repo is not hyper active.

simonpasquier:
[...]

PR #2225, opened 2020-04-02
Enforce 512KB event size limit for Pagerduty events

shamilpd:
PagerDuty will soon start enforcing a size limit of 512KB on all events. To prevent sending events that are larger than that, this PR implements the proposal that was decided in this issue https://github.com/prometheus/alertmanager/issues/2178

The size of events are checked after they are encoded into json. If the size is above 512KB, the custom details (or details for v1 events) are replaced with an error message.

This was my first time writing Go so the code may not be the best :)

### DIFF
--- a/notify/pagerduty/pagerduty.go
+++ b/notify/pagerduty/pagerduty.go
@@ -37,6 +37,8 @@ import (
 
 const MaxEventSize int = 512000
 
+var truncatedCustomDetailsMsg = fmt.Sprintf("Custom details have been removed because the original event exceeds the maximum size of %s", units.MetricBytes(MaxEventSize).String())
+
 // Notifier implements a Notifier for PagerDuty notifications.
 type Notifier struct {
 	conf    *config.PagerdutyConfig
@@ -117,23 +119,21 @@ func (n *Notifier) encodeMessage(msg *pagerDutyMessage) (bytes.Buffer, error) {
 	}
 
 	if buf.Len() >= MaxEventSize {
-		truncatedMsg := fmt.Sprintf("Custom details have been removed because the original event exceeds the maximum size of %s", units.MetricBytes(MaxEventSize).String())
-
 		if n.apiV1 != "" {
 			errorMsg, ok := msg.Details["error"]
-			if ok && errorMsg == truncatedMsg {
+			if ok && errorMsg == truncatedCustomDetailsMsg {
 				return buf, errors.New("PagerDuty event is too big")
 			}
 
-			msg.Details = map[string]string{"error": truncatedMsg}
+			msg.Details = map[string]string{"error": truncatedCustomDetailsMsg}
 			return n.encodeMessage(msg)
 		} else {
 			errorMsg, ok := msg.Payload.CustomDetails["error"]
-			if ok && errorMsg == truncatedMsg {
+			if ok && errorMsg == truncatedCustomDetailsMsg {
 				return buf, errors.New("PagerDuty event is too big")
 			}
 
-			msg.Payload.CustomDetails = map[string]string{"error": truncatedMsg}
+			msg.Payload.CustomDetails = map[string]string{"error": truncatedCustomDetailsMsg}
 			return n.encodeMessage(msg)
 		}
 	}
--- a/notify/pagerduty/pagerduty.go
+++ b/notify/pagerduty/pagerduty.go
@@ -114,7 +114,7 @@ func (n *Notifier) encodeMessage(msg *pagerDutyMessage) (bytes.Buffer, error) {
 	}
 
 	if buf.Len() >= n.conf.MaxEventSize {
-		truncated_msg := "Custom details have been removed because the original event was too big"
+		truncated_msg := fmt.Sprintf("Custom details have been removed because the original event exceeds the maximum size of %dKB", n.conf.MaxEventSize/1000)
 
 		if n.apiV1 != "" {
 			error_msg, ok := msg.Details["error"]
--- a/notify/pagerduty/pagerduty.go
+++ b/notify/pagerduty/pagerduty.go
@@ -37,8 +37,6 @@ import (
 
 const MaxEventSize int = 512000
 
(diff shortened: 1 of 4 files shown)


shamilpd:
Hi @simonpasquier, do you mind taking a look when you have time? Thanks.

chhetripradeep:
I think pagerduty has enforced this payload limit. Any chance of reviewing and merging this PR. Thank you @shamilpd for the work.

simonpasquier:
Thanks for the PR and sorry for the late review!

shamilpd:
@simonpasquier  Thank you for the review! I have made the changes you asked for.

shamilpd:
Hello @simonpasquier @roidelapluie Can we move forward and merge this PR?
We (PagerDuty) are planning to start enforcing this limit sometime in the summer.

notify/pagerduty/pagerduty.go
```diff
@@ -107,6 +110,37 @@ type pagerDutyPayload struct {
...
 }
 
+func (n *Notifier) encodeMessage(msg *pagerDutyMessage) (bytes.Buffer, error) {
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(msg); err != nil {
+		return buf, errors.Wrap(err, "failed to encode PagerDuty message")
+	}
+
+	if buf.Len() >= MaxEventSize {
+		truncatedMsg := fmt.Sprintf("Custom details have been removed because the original event exceeds the maximum size of %s", units.MetricBytes(MaxEventSize).String())
```
roidelapluie, line 120: Can this be a package-level variable?

roidelapluie:
I am happy to merge this, I have posted one last question/comment.

roidelapluie:
Also, should we really fail at the client side? Can't we still send the request anyway and see what happens?

notify/pagerduty/pagerduty.go
shamilpd: @roidelapluie  If I add something like
```
const TruncatedCustomDetailsMsg string = fmt.Sprintf("Custom details have been removed because the original event exceeds the maximum size of %s", units.MetricBytes(MaxEventSize).String())
```
I get a compile error because the expression is not a constant. 
I can have something like
```
const TruncatedCustomDetailsMsg string = "Custom details have been removed because the original event exceeds the maximum size of %s"
```
or have the size limit hardcoded in the message
```
const TruncatedCustomDetailsMsg string = "Custom details have been removed because the original event exceeds the maximum size of 512KB"
```
Which do you prefer?

roidelapluie: var TruncatedCustomDetailsMsg string = fmt.Sprintf("Custom details have been removed because the original event exceeds the maximum size of %s", units.MetricBytes(MaxEventSize).String())

shamilpd:
@roidelapluie Well the client-side failing is only happening if the message is still too large after truncation (replacing custom details). If a request was still sent to PagerDuty, it would return a 400 and the client would fail anyway. The difference here is ease of visibility. By failing in the client, it should be a lot easier to see why it failed. 

As for the case of whether to do truncation at all, we need to bias towards not dropping as many events as possible since a dropped event means someone potentially doesn't get paged. This would upset our customers.

notify/pagerduty/pagerduty.go
roidelapluie: My point is that it is maybe not needed to compute it upon each request

shamilpd: Ah! I thought package level variables can only be `const`
Sorry, I am new to golang :)
I will make the change.

roidelapluie: Also it can be unexported

notify/pagerduty/pagerduty.go
```diff
@@ -107,6 +112,35 @@ type pagerDutyPayload struct {
...
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(msg); err != nil {
+		return buf, errors.Wrap(err, "failed to encode PagerDuty message")
+	}
+
+	if buf.Len() >= MaxEventSize {
+		if n.apiV1 != "" {
+			errorMsg, ok := msg.Details["error"]
+			if ok && errorMsg == truncatedCustomDetailsMsg {
+				return buf, errors.New("PagerDuty event is too big")
```
roidelapluie, line 125: Should we set the limit here, like, expected at most X bytes, got Y bytes ?

roidelapluie: to help operators

notify/pagerduty/pagerduty.go
roidelapluie: It is important to have it out of the function because it is a recursively function. We need to be sure it is the same between iterations. While it was true I'm your implementation, future code changes might have created an infinite loop.

shamilpd:
Thank you @roidelapluie @simonpasquier for reviewing this again!
I have made the requested changes: removed the recursion and added a warning log.

shamilpd:
@roidelapluie  Is there anything I can do to move forward with this or are we ready to merge?

notify/pagerduty/pagerduty.go
```diff
@@ -34,6 +35,8 @@ import (
 	"github.com/prometheus/alertmanager/types"
 )
 
+const MaxEventSize int = 512000
```
roidelapluie, line 38: this can be unexported

notify/pagerduty/pagerduty.go
```diff
@@ -107,6 +110,33 @@ type pagerDutyPayload struct {
 	CustomDetails map[string]string `json:"custom_details,omitempty"`
 }
 
+func (n *Notifier) encodeMessage(msg *pagerDutyMessage) (bytes.Buffer, error) {
+	var buf bytes.Buffer
+	if err := json.NewEncoder(&buf).Encode(msg); err != nil {
+		return buf, errors.Wrap(err, "failed to encode PagerDuty message")
+	}
+
+	if buf.Len() >= MaxEventSize {
```
roidelapluie, line 119:
```suggestion
	if buf.Len() > MaxEventSize {
```

simonpasquier:
Thanks for your work!

PR #2235, opened 2020-04-17
Add redirect on / to routePrefix

LucasBoisserie:
When `routePrefix` is set alertmanager doesn't redirect traffic from `/` to target `routePrefix`
Prometheus implemented it with an http redirect to targeted `routePrefix`,  [here](https://github.com/prometheus/prometheus/blob/cd12f0873c3eb2031f7ba9b2e169449aa1012e3f/web/web.go#L328-L330)

Add the same code, to have the same comportement


simonpasquier:
Thanks!

Issue #1445, opened 2018-06-29
Tests fail in some environments


TheTincho:
Hi, I have just uploaded 0.15.0 to Debian, and immediately got test failures from 4 arches. I was ready to attribute this to some arch problem, but looking at the code, and at #1434, I think this is a problem with how memberlist deals with IP addresses (which seems pretty hackish, btw, and does not seem to take into account ipv6 properly).

These are the errors:
```
=== RUN   TestJoinLeave
--- FAIL: TestJoinLeave (0.01s)
	require.go:794: 
			Error Trace:	cluster_test.go:44
			Error:      	Received unexpected error:
			            	Failed to get final advertise address: Failed to parse advertise address "<nil>"
			            	create memberlist
			            	github.com/prometheus/alertmanager/cluster.Join
			            		/home/tincho/prometheus-alertmanager-0.15.0+ds/build/src/github.com/prometheus/alertmanager/cluster/cluster.go:207
			            	github.com/prometheus/alertmanager/cluster.TestJoinLeave
			            		/home/tincho/prometheus-alertmanager-0.15.0+ds/build/src/github.com/prometheus/alertmanager/cluster/cluster_test.go:29
			            	testing.tRunner
			            		/usr/lib/go-1.10/src/testing/testing.go:777
			            	runtime.goexit
			            		/usr/lib/go-1.10/src/runtime/asm_s390x.s:986
			Test:       	TestJoinLeave
=== RUN   TestReconnect
--- FAIL: TestReconnect (0.03s)
	require.go:794: 
			Error Trace:	cluster_test.go:97
			Error:      	Received unexpected error:
			            	Failed to get final advertise address: Failed to parse advertise address "<nil>"
			            	create memberlist
			            	github.com/prometheus/alertmanager/cluster.Join
			            		/home/tincho/prometheus-alertmanager-0.15.0+ds/build/src/github.com/prometheus/alertmanager/cluster/cluster.go:207
			            	github.com/prometheus/alertmanager/cluster.TestReconnect
			            		/home/tincho/prometheus-alertmanager-0.15.0+ds/build/src/github.com/prometheus/alertmanager/cluster/cluster_test.go:82
			            	testing.tRunner
			            		/usr/lib/go-1.10/src/testing/testing.go:777
			            	runtime.goexit
			            		/usr/lib/go-1.10/src/runtime/asm_s390x.s:986
			Test:       	TestReconnect
=== RUN   TestRemoveFailedPeers
[...]

PR #2237, opened 2020-04-20
Secure cluster traffic via mutual TLS

### LINKED ISSUE #1322: Securing the gossip protocol?

hooten:
Co-authored-by: Sharad Gaur <sharadgaur@gmail.com>
Signed-off-by: Dustin Hooten <dustinhooten@gmail.com>

This pull request makes it possible to use mutual TLS for cluster communications.

By adding the path to a TLS configuration file using the command line flag `--cluster.tls-config`, TLS will be used for inter-peer gossip.

Mutual TLS is achieved by configuring Memberlist to use a TLS implementation of the `memberlist.Transport` interface. This approach has been submitted in a [design doc](https://github.com/prometheus/alertmanager/blob/master/doc/design/secure-cluster-traffic.md) and implemented as a proof-of-concept in https://github.com/prometheus/alertmanager/pull/1819. This pull request continues that work.

Closes #1322

**Updates December 30, 2020**: TLS server config uses common code from `prometheus/exporter-toolkit`.
**Updates January 2021**: TLS config also allows client config using code from `prometheus/common`.


brian-brazil:
This PR should not be considered for merging until the tls code is in common, it is not planned to ever have more than 1 copy of that code.

hooten:
Happy to have the rest of it reviewed. We can clean up the TLS code before merging.

cluster/tls_transport.go
```diff
@@ -0,0 +1,335 @@
...
+	err := t.listener.Close()
+	<-t.done
+	return err
+}
+
+// WriteTo is a packet-oriented interface that creates a connection,
+// writes to it, and closes it. It also returns a time stamp of when
+// the packet was written.
+func (t *TLSTransport) WriteTo(b []byte, addr string) (time.Time, error) {
+	dialer := &net.Dialer{Timeout: DefaultTcpTimeout}
```
mxinden, line 181: Do I understand correctly, that this is opening up a new TCP connection for each packet to send? What do you think of something along the lines of a [connection pool](https://github.com/mxinden/memberlist-tls-transport/blob/master/tls_transport.go#L178)?

sharadgaur: It adds additional complexity as we need to manage the lifecycle of the connection and the Memberlist does not provide any good way to handle it. I would not recommend using a connection pool in this case. 
In our testing, we did not see any issues yet. If you like we can run additional load tests to identify if we need a pool or not.

cluster/tls_connection.go
```diff
@@ -0,0 +1,152 @@
...
+	case 'p':
+		return packet, nil
+	default:
+		return none, errors.New("invalid byte")
+	}
+}
+
+// writePacket writes all the bytes in one operation so no concurrent write happens in between.
+// It prefixes the connection type, the from address and the message length.
+func writePacket(conn net.Conn, fromAddr string, b []byte) error {
```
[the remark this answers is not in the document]
hooten: Great idea!

csmarchbanks:
> In my proof-of-concept I extracted the memberlist specific logic to a separate project for other memberlist users to reuse. What do you think about doing the same here? Is the additional complexity worth the effort?

We already moved tsdb back inside of Prometheus to avoid dealing with version upgrades and such across projects. It is up to the maintainers of Alertmanager, but I would keep the lessons of tsdb in mind.

cluster/tls_transport.go
brian-brazil: What sort of network latency did the testing have?

sharadgaur: An individual request is taking less than 3 mill seconds. Here are the benchmark results
goos: darwin
goarch: amd64
pkg: github.com/prometheus/alertmanager/cluster
BenchmarkWriteTo-12    	1000000000	         0.00304 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/prometheus/alertmanager/cluster	0.111s
Success: Benchmarks passed.

brian-brazil: The question isn't about how long things take over localhost, it's how many round trips may be required over a 200ms+ connection.

sharadgaur: Oh got it. I will run a test against remotely deployed Alertmanger next week. If we see any issue we will implement a connection pool.
Thank you.

cluster/tls_connection.go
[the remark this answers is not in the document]
hooten: We made a change to use protobuf for this. Any thoughts? Are there ways we could improve it? Thanks!

cluster/tls_transport.go
hooten: We've added a connection pool :)

hooten:
I've made the requested updates. I'd appreciate another review. Thanks!

brian-brazil:
I've only taken a quick peek, and noticed that there's no config for the client-side parts of this. You should use https://github.com/prometheus/common/blob/20c99e7aa07352b599bd0517cd5ca945f4af0407/config/http_config.go#L325 for that.

hooten:
Thanks! I will update it to use that one.

cluster/testdata/tls_config_node1.yml
```diff
@@ -1,5 +1,9 @@
 tls_server_config:
-  cert_file: "server.crt"
-  key_file: "server.key"
-  client_ca_file: "tls-ca-chain.pem"
+  cert_file: "certs/node1.pem"
+  key_file: "certs/node1-key.pem"
+  client_ca_file: "certs/ca.pem"
   client_auth_type: "VerifyClientCertIfGiven"
+tls_client_config:
```
hooten, line 6: @brian-brazil is this the sort of config that [you suggested I add](https://github.com/prometheus/alertmanager/pull/2237#issuecomment-758888894)? Or was it something else? Thanks!

cluster/tls_config.go
```diff
@@ -0,0 +1,52 @@
...
+// limitations under the License.
+
+package cluster
+
+import (
+	"crypto/tls"
+	"io/ioutil"
+	"path/filepath"
+
+	common "github.com/prometheus/common/config"
```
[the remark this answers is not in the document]
hooten: removed the rename

cluster/tls_config.go
```diff
@@ -0,0 +1,52 @@
...
+func GetTLSTransportConfig(configPath string) (*TLSTransportConfig, error) {
+	if configPath == "" {
+		return nil, nil
+	}
+	bytes, err := ioutil.ReadFile(configPath)
+	if err != nil {
+		return nil, err
+	}
+	config := &TLSTransportConfig{
+		TLSServerConfig: &web.TLSStruct{
```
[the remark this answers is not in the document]
hooten: removed this configuration. Now just unmarshaling

hooten:
@simonpasquier I'd appreciate a review whenever you get a chance.

robertjsullivan:
Hey y'all, we would love to see this merged!  Any updates on when that will happen?

beorn7:
@gotjosh would this be something for you to look at, to at the same time get more familiar with the codebase?

gotjosh:
I'm happy to take a look at it eventually but my current backlog seems to not finish. This is quite a big change though, and I have 3 other things here (1 PR open, 1 to review and another to implement 😄 ).

I'll add it to my list but I don't expect this to be soonish.

beorn7:
@gotjosh what's your current state? Do you think you'll get to this any time soon?

@hooten thanks for the recent commits. I tried to find out why the linter is failing in the test. Perhaps the problem can be solved by merging in (or rebasing on top of) master.

gotjosh:
@beorn7 I'll set aside some time to review this soon. @hooten can you please rebase with master so that I can do some testing please?

beorn7:
BTW, the test failures seem real (expired certificate) rather than just flaky.

hooten:
Yes, I noticed that. Busy week at work. I will fix it soon

hooten:
The expired certs issue has been fixed.

hooten:
I would appreciate it if someone could review this.

beorn7:
@gotjosh @simonpasquier what is your state on this? If you cannot review this any time soon, can you delegate this to someone else at Red Hat or Grafana who is invested here?

cluster/tls_connection.go
```diff
@@ -0,0 +1,170 @@
...
+
+	"github.com/gogo/protobuf/proto"
+	"github.com/hashicorp/memberlist"
+	"github.com/pkg/errors"
+
+	"github.com/prometheus/alertmanager/cluster/clusterpb"
+)
+
+const (
+	version      = "v1.0.0"
```
csmarchbanks, line 33: Should this be a stable version to begin with? Also, should we check versions somewhere and gracefully error if a version is not compatible?

cluster/tls_connection.go
```diff
@@ -0,0 +1,170 @@
...
+	msg, err := proto.Marshal(&clusterpb.MemberlistMessage{
+		Version:  version,
+		Kind:     clusterpb.MemberlistMessage_PACKET,
+		FromAddr: fromAddr,
+		Msg:      b,
+	})
+	if err != nil {
+		return errors.Wrap(err, "unable to marshal memeberlist packet message")
+	}
+	buf := make([]byte, uint32length)
```
csmarchbanks, line 96: Probably not a huge deal, but by making this buffer only 4 bytes an extra allocation is happening for each write. Instead this buffer should probably be len(msg) + 4 bytes long, then you can PutUint32 and copy msg in without an extra allocation.

Same with writeStream()

csmarchbanks (commented):
I left a few minor comments to discuss/address, but nothing blocking. Since this is an additive change and all TLS support is already marked as experimental I move to merge this soon, and allow more people to test out the functionality. We can always provide fixes in additional PRs as they come up.

cluster/tls_connection.go
```diff
@@ -0,0 +1,170 @@
...
+
+const (
+	version      = "v1.0.0"
+	uint32length = 4
+)
+
+// tlsConn wraps net.Conn with connection pooling data.
+type tlsConn struct {
+	connection net.Conn
+	timeout    time.Duration
```
gotjosh, line 40: I don't think this `timeout` is used anywhere else but the dialer, do we need an attribute for it?

cluster/tls_connection.go
```diff
@@ -0,0 +1,170 @@
...
+	lock       sync.Mutex
```
gotjosh, line 41: nit: throughout a couple of files here. We tend to use the name of `mtx`  for the locks and place it above the variables it's protecting followed by a space. I don't think it's consistent throughout _all_ of the codebase but is never a bad time to start.

```suggestion
	mtx       sync.Mutex
```

gotjosh (commented):
I apologise as I've been sitting on half of my review for a while, but overall the changes look good - my comments are mostly nits.

My take here is given we've cut a new version a couple of weeks ago and have had several patches to it accordingly, we can let this bake in master for a while before we cut a new version to ensure we have time to catch anything (if it comes!).

beorn7:
'm very sorry that it took so long to rally some reviewers, but @gotjosh and @csmarchbanks finally got to it a month ago. Both seem to just have a few minor nits. @hooten do you think you'll be able to address the comments? Then we could finally merge this and make it available for a somewhat wider public to try out.

hooten:
Yep, Thanks for the reviews! I will get to it this week

cluster/tls_connection.go
hooten: Removed, thanks!

cmd/alertmanager/main.go
```diff
@@ -227,6 +228,11 @@ func run() int {
 		return 1
 	}
 
+	tlsTransportConfig, err := cluster.GetTLSTransportConfig(*tlsConfigFile)
+	if err != nil {
+		level.Error(logger).Log("msg", "bad tls transport configuration", "err", err)
```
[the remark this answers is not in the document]
hooten: Fixed

docs/https.md
```diff
@@ -79,3 +80,63 @@ basic_auth_users:
...
+  [ prefer_server_cipher_suites: <bool> | default = true ]
+
+  # Elliptic curves that will be used in an ECDHE handshake, in preference
+  # order. Available curves are documented in the go documentation:
+  # https://golang.org/pkg/crypto/tls/#CurveID
+  [ curve_preferences:
+    [ - <string> ] ]
+
+tls_client_config:
+  # CA certificate to validate the server certificate with.
```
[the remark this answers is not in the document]
hooten: Fixed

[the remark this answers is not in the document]
hooten: Fixed

docs/https.md
```diff
@@ -79,3 +80,63 @@ basic_auth_users:
...
+    [ ca_file: <filepath> ]
+    
+    # Certificate and key files for client cert authentication to the server.
+    [ cert_file: <filepath> ]
+    [ key_file: <filepath> ]
+    
+    # ServerName extension to indicate the name of the server.
```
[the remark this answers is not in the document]
hooten: Fixed

cluster/tls_connection.go
hooten: Thanks, Chris. Does adding a capacity argument solve this issue? Like so https://github.com/prometheus/alertmanager/pull/2237/commits/1018b286c8dd998ff74d6ce7c9b843a6f651b8e0 ?

cluster/tls_connection.go
hooten: Good questions. Thoughts on this approach? https://github.com/prometheus/alertmanager/pull/2237/commits/2fdbea685bca8dee14c6802fc73ff3de07144936

cluster/connection_pool.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+	var err error
+	key := fmt.Sprintf("%s/%d", addr, int64(timeout))
+	conn, ok := pool.connections[key]
+	if !ok || !conn.alive() {
+		conn, err = dialTLSConn(addr, timeout, pool.tlsConfig)
+		if err != nil {
+			return nil, err
+		}
+		pool.connections[key] = conn
```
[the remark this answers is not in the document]
hooten: I suspect that the number of connections in the pool will remain small unless peer addresses change regularly/frequently. Maybe it would be worth declaring bankruptcy on all the connections in the pool once it reaches a certain size (e.g., 1000). Is there a better way?

hooten:
@gotjosh, @csmarchbanks. Thanks again for the reviews. I pushed some fixes and left a couple comments/questions in response a few days ago. Let me know what you think. Thanks!

hooten:
@gotjosh, @csmarchbanks. Thanks again for the reviews. I pushed some fixes and left a couple comments/questions in response a few days ago. Let me know what you think. Thanks!

csmarchbanks (commented):
One ask to try on the memory leak, I am happy with your solutions to the other two comments I brought up and resolved them, thanks!

If it is too much of a pain to use an LRU or something else in the connection pool I would be ok doing that later/if it becomes an issue as long as @gotjosh is ok there being a known potential memory leak.

hooten:
Thanks @csmarchbanks, I will give that a try sometime in the next couple of days

cluster/connection_pool.go
```diff
@@ -0,0 +1,67 @@
...
+	"errors"
+	"fmt"
+	"sync"
+	"time"
+)
+
+type connectionPool struct {
+	connections map[string]*tlsConn
+	tlsConfig   *tls.Config
+	lock        sync.Mutex
```
gotjosh, line 26:
```suggestion
	mtx        sync.Mutex
```
last nit given you're about to touch this file as part of Chris's comments.

cluster/tls_connection.go
```diff
@@ -0,0 +1,175 @@
...
+	buf := make([]byte, uint32length, uint32length+len(msg))
+	binary.LittleEndian.PutUint32(buf, uint32(len(msg)))
+	_, err = conn.Write(append(buf, msg...))
+	return err
+}
+
+// read returns a packet for packet connections or an error if there is one.
+// It returns nothing if the connection is meant to be streamed.
+func (conn *tlsConn) read() (*memberlist.Packet, error) {
+	if conn.connection == nil {
```
gotjosh, line 122: my understanding was that `conn.mtx` is protecting connection, any particular reason why we're not using the lock to read the connection here?

cluster/tls_transport.go
```diff
@@ -0,0 +1,341 @@
...
+func (t *TLSTransport) DialTimeout(addr string, timeout time.Duration) (net.Conn, error) {
+	conn, err := dialTLSConn(addr, timeout, t.tlsClientCfg)
+	if err != nil {
+		t.writeErrs.WithLabelValues("stream").Inc()
+		return nil, errors.Wrap(err, "failed to dial")
+	}
+	err = conn.writeStream()
+	if err != nil {
+		t.writeErrs.WithLabelValues("stream").Inc()
+		return conn.connection, errors.Wrap(err, "failed to create stream connection")
```
gotjosh, line 219: Similar thoughts here - perhaps, we're assuming we want to sync writing but not reading - is this correct?

gotjosh (approved):
Apologies for the delay, but overall this LGTM. I don't want to delay this any further so I've left one nit and two comments. 

Thanks @hooten for addressing my feedback.

cluster/tls_connection.go
hooten: Good catch. I looked at it more and I think the lock was incidentally unnecessary because `read` is only called after `rcvTLSConn`, which creates a new connection object.

However, I'm adding a lock around the reading of the connection to make this safer to use.

cluster/tls_transport.go
hooten: I think this is a similar situation to the above. The connection is never reused so it is incidentally working. I'm adding a new function and some safety measures.

cluster/connection_pool.go
hooten: I took away the mutex here because the lru cache provides its own locking mechanism, which I think suffices.

cluster/connection_pool.go
```diff
@@ -0,0 +1,78 @@
...
+		conn, ok := value.(*tlsConn)
+		if ok && conn.alive() {
+			return conn, nil
+		}
+	}
+	conn, err := dialTLSConn(addr, timeout, pool.tlsConfig)
+	if err != nil {
+		return nil, err
+	}
+	pool.cache.Add(key, conn)
```
hooten, line 68: @csmarchbanks I updated this to use an lru cache. I removed the mutex from this file because the cache provides locking. However,  I'm wondering if I still need it between the `pool.cache.Get` and the `pool.cache.Add`. Thoughts?

csmarchbanks: I think yes as two concurrent operations could each dial a TLS connection and then add the connection for the same key?

csmarchbanks: You might be able to use `PeekOrAdd` to avoid a mutex, but it would also have a bit of extra complexity. Basically:
1. Do the get
2. If not found, dial a TLS connection
3. Use PeekOrAdd
4. If found, use the "old" connection returned by PeekOrAdd, if added use the created connection.

Whichever way looks cleaner is fine by me.

hooten: I tried both ways. In the end, the mutex ended looking cleaner. 

The `PeekOrAdd` would have been nice if I didn't always need to check whether the connection is alive when I get it back from the cache.

csmarchbanks (approved):
Re-ran the tests and they pass now, looks like a flaky acceptance test. :+1: from me, I will wait for Monday to see if any comments/concerns come through and will then plan to merge this!

hooten:
🎉 Thanks @csmarchbanks @gotjosh @beorn7 @mxinden @brian-brazil @sharadgaur !

markmsmith:
+100. this is awesome, thank you!  
Now that this has landed, is there a release planned soon, or are there any other blocking issues?

aauVadim:
Hello! Thanks a lot for adding this feature, seems like a very great thing! 

I have a few questions; 
1. What format would TLS Config file be? I tried to follow up with few formats, and it doesn't seem that those are the right ones. 
2. Considering question #1, when I pass `--cluster.tls-config` flag down to AM, it fails to start. Error is: `level=error ts=2022-03-04T21:21:25.340Z caller=coordinator.go:118 component=configuration msg="Loading configuration file failed" file=alertmanager.yml err="open alertmanager.yml: no such file or directory"` Would that be because config file possibly wrong? 
3. Is this feature enabled in by default in Alertmanager, or I need to download additional libraries? Not entirely clear. 


I would highly appreciate advice

Scrin:
1. An example TLS config file can be found [here](https://github.com/prometheus/alertmanager/blob/main/docs/https.md#gossip-traffic)
2. The main config file (typically `alertmanager.yml`, provided via `--config.file`) is different form the TLS config file, make sure you are not mixing them up or missing the main config file. Something like:
```/bin/alertmanager --config.file=/etc/alertmanager/alertmanager.yml --cluster.tls-config=/etc/alertmanager/tls.yml``` (and any other cli flags) should work
3. I believe it should be enabled by default in AM if you provide a valid configuration, at least on the docker image the config is all you need

**Comment by aauVadim** [comment]
@Scrin Thanks a lot for reply! 
Yeah, seems that I totally did not see the document there. Though was googling and searching for the tls config file example. Exactly what I needed!   

I am not using a ready made docker image, I'm building my own. At this point I get an error `alertmanager: error: unknown long flag '--cluster.tls-config', try --help` where `--help` does not list `--cluster.tls-config` flag at all. Alertmanager version 0.23.0 

Maybe I've missed a step in installation process?

**Comment by Scrin** [comment]
0.23.0 was released before this was merged, so it's not available in that version yet, and the next version doesn't seem to be released yet. If you build AM from the main branch (instead of the v0.23.0 tag) you should get this feature (Docker users can use `prom/alertmanager:main` docker image for that)

File: cluster/tls_transport_test.go [hunk 18]
```diff
@@ -0,0 +1,213 @@
...
+		bindPort      int
+		inputIp       string
+		inputPort     int
+		expectedIp    string
+		expectedPort  int
+		expectedError string
+	}{
+		{bindAddr: localhost, bindPort: 9094, inputIp: "10.0.0.5", inputPort: 54231, expectedIp: "10.0.0.5", expectedPort: 54231},
+		{bindAddr: localhost, bindPort: 9093, inputIp: "invalid", inputPort: 54231, expectedError: "failed to parse advertise address \"invalid\""},
+		{bindAddr: "0.0.0.0", bindPort: 0, inputIp: "", inputPort: 0, expectedIp: "random"},
```
dswarbrick, line 77: Such a test is not very packager-friendly. Most distros will run unit tests in clean room environment when packaging, which can often mean a host with only a loopback interface (with 127.0.0.1 and _usually_ [::1] configured), and no default route.

On Debian buildbots, and presumably other distros with similar clean room package building policy, this fails with:
```
=== RUN   TestFinalAdvertiseAddr
    tls_transport_test.go:89: 
        	Error Trace:	/<<PKGBUILDDIR>>/build/src/github.com/prometheus/alertmanager/cluster/tls_transport_test.go:89
        	Error:      	Expected nil, but got: &errors.errorString{s:"no private IP address found, and explicit IP not provided"}
        	Test:       	TestFinalAdvertiseAddr
--- FAIL: TestFinalAdvertiseAddr (0.00s)
```
This is due to the `bindAddr` being "0.0.0.0" and the `inputIp` being empty.

I see that TestClusterJoinAndReconnect in cluster/cluster_test.go at least checks first whether a private IP exists, and skips the test if none is found. Perhaps it makes sense to omit this particular test case in such a scenario also.

simonpasquier: > I see that TestClusterJoinAndReconnect in cluster/cluster_test.go at least checks first whether a private IP exists, and skips the test if none is found. Perhaps it makes sense to omit this particular test case in such a scenario also.

I agree and the skip in cluster_test.go exists exactly for the same constraint (https://github.com/prometheus/alertmanager/issues/1445).

PR #2257, opened 2020-05-13
Update dependencies

roidelapluie:
Update membership

Update common (support HTTP/2 client)


roidelapluie:
I have not tested that; this is still WIP

simonpasquier:
Thanks!

> Update Prometheus to 2.18.1

Alertmanager doesn't have to depend on Prometheus.

Can you switch from `github.com/cenkalti/backoff` to `github.com/cenkalti/backoff/v4`? the only use is in [notify/notify.go](https://github.com/prometheus/alertmanager/blob/8d050daf513ed96767dbfc678447f09885be5ddf/notify/notify.go#L23). As far as I can tell and given we're only using a small part of the interface, it shouldn't be a problem to upgrade.

roidelapluie:
Updated

client/client.go
```diff
@@ -142,7 +142,7 @@ func (h *httpStatusAPI) Get(ctx context.Context) (*ServerStatus, error) {
 		return nil, fmt.Errorf("error creating request: %v", err)
 	}
 
-	_, body, _, err := h.client.Do(ctx, req) // Ignoring warnings.
+	_, body, err := h.client.Do(ctx, req) // Ignoring warnings.
```
simonpasquier, line 145: the comment can be removed (same below)

roidelapluie:
We need to check if memberlist is compatible with what we have now

roidelapluie:
I have tested and the membership upgrade seems to be compatible with alertmanager 0.20.0

PR #2260, opened 2020-05-18
cluster: log error on reconnect failures

simonpasquier:
We forgot to log the error message from the `Join()` function which makes it harder to understand what's happening...

cc @liliC


s-urbaniak:
lgtm :+1:

lilic:
Nice, thanks!

PR #2261, opened 2020-05-18
api/v2: add path and method to API v2 logs

simonpasquier:
When an API v2 handler logged a message, the log wouldn't include the
path and method. Since different handlers perform the same
validations (e.g. matchers for alerts and silences), it isn't easy to
know which handler was invoked (though the logged filename + line number provides a hint).

### DIFF
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -224,14 +224,14 @@ func (api *API) getAlertsHandler(params alert_ops.GetAlertsParams) middleware.Re
 
 	matchers, err := parseFilter(params.Filter)
 	if err != nil {
-		level.Error(logger).Log("msg", "failed to parse matchers", "err", err)
+		level.Error(logger).Log("msg", "Failed to parse matchers", "err", err)
 		return alertgroup_ops.NewGetAlertGroupsBadRequest().WithPayload(err.Error())
 	}
 
 	if params.Receiver != nil {
 		receiverFilter, err = regexp.Compile("^(?:" + *params.Receiver + ")$")
 		if err != nil {
-			level.Error(logger).Log("msg", "failed to compile receiver regex", "err", err)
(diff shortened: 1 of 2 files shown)


api/v2/api.go
```diff
@@ -124,8 +124,6 @@ func NewAPI(
 	openAPI.SilenceGetSilencesHandler = silence_ops.GetSilencesHandlerFunc(api.getSilencesHandler)
 	openAPI.SilencePostSilencesHandler = silence_ops.PostSilencesHandlerFunc(api.postSilencesHandler)
 
-	openAPI.Logger = func(s string, i ...interface{}) { level.Error(api.logger).Log(i...) }
```
simonpasquier, line 127: After reading through the openapi code, it seems that it isn't used anywhere because we're not using `*github.com/prometheus/alertmanager/api/v2/restapi.Server` directly.
In any case the code here didn't work as expected, it should have been `level.Error(api.logger).Log("msg", fmt.Sprintf(s, i...))`.

api/v2/api.go
```diff
@@ -216,18 +214,20 @@ func (api *API) getAlertsHandler(params alert_ops.GetAlertsParams) middleware.Re
 		// are no alerts present
 		res = open_api_models.GettableAlerts{}
 		ctx = params.HTTPRequest.Context()
+
+		logger = log.With(api.logger, "handler", "alerts", "methods", "GET")
```
roidelapluie, line 218: do we have to hardcode those info?

mxinden: You could probably extract some of them from `params.HTTPRequest`.

simonpasquier: Thanks for the feedback! I've updated the code.

roidelapluie:
In prometheus/prometheus log `msg` fields are capitalized. Consistency would be good

Issue #2226, opened 2020-04-04
date/time selector


pnnlrck8sstatus:
It would be nice if the new silence dialog could support a date/time selector such as https://www.jqueryscript.net/images/Lightweight-jQuery-Date-Input-Picker.jpg

This would make it easier to select now() or next monday.

w0rm:
Closed via https://github.com/prometheus/alertmanager/pull/2262

PR #2262, opened 2020-05-18
Add date picker to silence form views

m-masataka:
Added date picker to silence form views to select date easier.
see #2226 
datepicker-demo


simonpasquier:
Thanks for tackling this one! It seems that the CI isn't happy with your changes yet :)

simonpasquier:
This looks really good. I guess it would be good to support time too but probably it isn't that easy (at least the component supports only dates)?

roidelapluie:
should we pick up a component that support both? can we reuse Prometheus' component?

w0rm:
After a brief research I found this component: https://package.elm-lang.org/packages/mercurymedia/elm-datetime-picker/latest/

It supports both date and time, it also supports selecting the duration, which is what we need. 

I didn't look in depth and what the integration would look like. @m-masataka I wonder if you have checked this one?

m-masataka:
@w0rm Please give me some advice.
I tried [Panaguitus/elm-datepicker](https://package.elm-lang.org/packages/PanagiotisGeorgiadis/elm-datepicker/latest/) because [mercurymedia/elm-datetime-picker](https://package.elm-lang.org/packages/mercurymedia/elm-datetime-picker/latest/)'s time selector always reset 00:00 when open datetime picker( I think it is not useful for anyone ) . 
However [Panaguitus/elm-datepicker](https://package.elm-lang.org/packages/PanagiotisGeorgiadis/elm-datepicker/latest/) uses a component that has duplicated name with ui/app/src/DateTime.elm.
How can I avoid this component name duplication?  ( or I have to use other component?)

here is the demo↓
datetimepicker

w0rm:
@m-masataka feel free to just move this module into Utils!

m-masataka:
I checked some datetime picker components, but I didn't find the perfect one.
Therefore I implement original datetime picker that based on [mercurymedia/elm-datetime-picker](https://package.elm-lang.org/packages/mercurymedia/elm-datetime-picker/latest/). 
Please check it out @w0rm.
demo-datetimepicker

w0rm:
@m-masataka why did you have to inline the third party lib, can you please summarise what was not working in the original? 

I also wonder whether there might be a copyright issue by inlining the code like this.

m-masataka:
@w0rm the original one mostly work fine,  it’s just a matter of preference.
I want to use date picker(not datetime picker) when there is no need to edit time. But  the original time picker always reset 0:00 when date is picked.
In addtion, I don't like `select` field as time-picker much.

> I also wonder whether there might be a copyright issue by inlining the code like this.

That is true for sure. I'll implement full scratch code as needed.
But in this PR, is it better if I only add [mercurymedia/elm-datetime-picker](https://package.elm-lang.org/packages/mercurymedia/elm-datetime-picker/latest/)?

w0rm:
@m-masataka Oh, I see. Sorry, you actually wrote about this before!

`Panaguitus/elm-datepicker` didn't work either? I think renaming or getting rid of our internal types is totally fine if that is needed.

m-masataka:
`Panaguitus/elm-datepicker` is perfect for me expect unnecessary view component such as today button and date limit.
`Panaguitus/elm-datepicker`  strongly depends on other datetime component such as [elm-datetime](https://package.elm-lang.org/packages/PanagiotisGeorgiadis/elm-datetime/1.3.0/). I'm concerned about that datetime component will conflict with other packages.

m-masataka:
> I think renaming or getting rid of our internal types is totally fine if that is needed.

Ah, we can avoid any conflict. 
Which do you think is better we should use, single picker or  duration picker? @w0rm

w0rm:
@m-masataka I think picking a duration is better, this way a user won't have to open the dialog twice for each date time.

Screenshot 2020-06-14 at 20 36 51

m-masataka:
Thank you @w0rm.
Unfortunately, I have one bad news about `Panaguitus/elm-datepicker`.
This duration picker can't pick same date as start and end.  ([here](https://github.com/PanagiotisGeorgiadis/Elm-DatePicker/blob/e991ff88e75e4f02f5c24cf2160cbc24a11f92f2/src/DateRangePicker.elm#L255) is the code.)
I think this feature is not user-friendly. What's your opinion on this?

We have some choice
1. Use `Panaguitus/elm-datepicker`'s duration picker
2. Use `Panaguitus/elm-datepicker`'s datetime picker ( not use duration picker)
3. Use `mercurymedia/elm-datetime-picker`'s duration picker 
4. Implement datepicker inside alertmanager with full scratch code

I think 3 (or 4) is better than others.

w0rm:
@m-masataka I fully agree with you that 3 or 4 is better.

Are you willing to contribute a custom date picker (option 4)? Long term I would prefer this option, because it is easier to migrate our own code if a new version of Elm comes up.

m-masataka:
Hi @w0rm 
I Implemented DateTimePicker inside alertmanager (and this is not third party libs).
Please review this.

datetime-picker-demo

roidelapluie:
cc @juliusv

juliusv:
Looks pretty IMO! I can't review Elm stuff though, I have never touched it :( @w0rm or @simonpasquier?

ui/app/src/Utils/DateTimePicker/Types.elm
```diff
@@ -0,0 +1,77 @@
...
+    , mouseOverDay : Maybe Posix
+    , startDate : Maybe Posix
+    , endDate : Maybe Posix
+    , startTime : Maybe Posix
+    , endTime : Maybe Posix
+    }
+
+
+type alias PickerConfig msg =
+    { zone : Zone
```
w0rm, line 26: I think it is safe to assume that this is always going to be in `utc`. You can simplify a lot of code and just use `utc` everywhere.

ui/app/src/Views/SilenceForm/Views.elm
```diff
@@ -44,9 +45,52 @@ view maybeId { matchers, comment } defaultCreator { form, silenceId, alerts, act
...
+                        [ div [ class "modal-content" ]
+                            [ div [ class "modal-header" ]
+                                [ button
+                                    [ class "close ml-auto"
+                                    , onClick (CloseDateTimePicker |> UpdateField)
+                                    ]
+                                    [ text "x" ]
+                                ]
+                            , div [ class "modal-body" ]
+                                [ viewDateTimePicker dateTimePickerConfig form.dateTimePicker ]
```
w0rm, line 68: If we use `utc`, then `viewDateTimePicker` could take `UpdateDateTimePicker` and the `dateTimePickerConfig` would not be needed.

ui/app/src/Views/SilenceForm/Types.elm
```diff
@@ -124,25 +136,50 @@ toSilence { id, comment, matchers, createdBy, startsAt, endsAt } =
 
 fromSilence : GettableSilence -> SilenceForm
 fromSilence { id, createdBy, comment, startsAt, endsAt, matchers } =
+    let
+        startsPosix =
+            case Utils.Date.timeFromString (DateTime.toString startsAt) of
```
w0rm, line 141: You can use [`Result.toMaybe`](https://package.elm-lang.org/packages/elm/core/latest/Result#toMaybe)

ui/app/src/Views/SilenceForm/Types.elm
```diff
@@ -124,25 +136,50 @@ toSilence { id, comment, matchers, createdBy, startsAt, endsAt } =
...
+                Ok starts ->
+                    Just starts
+
+                _ ->
+                    Nothing
+
+        endsPosix =
+            case Utils.Date.timeFromString (DateTime.toString endsAt) of
```
w0rm, line 149: Use [`Result.toMaybe`](https://package.elm-lang.org/packages/elm/core/latest/Result#toMaybe)

ui/app/src/Views/SilenceForm/Updates.elm
```diff
@@ -172,6 +174,55 @@ updateForm msg form =
...
+                startsAtTime =
+                    case timeFromString form.startsAt.value of
+                        Ok time ->
+                            Just time
+
+                        _ ->
+                            form.pickedStart
+
+                endsAtTime =
+                    case timeFromString form.endsAt.value of
```
w0rm, line 209: Pipe to [`Result.toMaybe`](https://package.elm-lang.org/packages/elm/core/latest/Result#toMaybe)

w0rm:
@m-masataka looks great, thank you for implementing this non trivial feature! 

I left a few comments that may simplify things a bit. But I am fine if this goes as is.

ui/app/src/Views/SilenceForm/Views.elm
m-masataka: Thank's. I removed ...Config.

simonpasquier:
Thanks a lot @m-masataka for the hard work! I'm not really qualified to review the Elm code but I'll trust @w0rm :)

PR #2267, opened 2020-05-25
support equal_all field to simplify inhibit rules

alburthoffman:


brian-brazil:
Can you explain what you're trying to do?

alburthoffman:
yeah. let me use this thread to explain.

alburthoffman:
What we want to do is to inhibit lower severity alerts with higher severity alerts. like the traffic of given region might be up and down, but we don't want to get fatal/critical/warning alerts from time to time.

In order to do that, we have to define inhibit rules explicitly like
```yaml
inhibit_rules:
- equal:
  - alertname
  - _rulename_
  - _namespace_
  - resource_id
  source_match:
    severity: FATAL
  target_match:
    severity: CRITICAL
```

The problem is we have lots of rules, more than 10K, and different alerts have different labels. then we have to define the equal field for each of them....
That would make things complex. It would be easy to provide one field to compare all if true, which will simplify all inhibit rules to few.

brian-brazil:
Inhibits are intended to be pretty coarse gained, e.g. an entire datacenter falling over. If you're getting so many notifications that this is a problem, I'd wonder if alerting levels need tuning. This PR would is also over-specialised, by tying grouping into selection

roidelapluie:
Also you will end up in a situation where the labels in equal will be empty and then it will inhibit things it should not.

alburthoffman:
@roidelapluie the behavior can be changed to if no common labels and equalAll is true, then return false.
@brian-brazil "an entire datacenter falling over" is a valid use case. It's also good to use inhibit rules to deal with flapping alerts. It's not easy to tune alerts. because different alerts have different ways. some are using static thresholds, some are using dynamic thresholds, some are using ML models.  sometimes it's super hard to tune in ML models. The other thing is that we might need to record down the flapping alerts, and analyze. That might help us in some cases.

brian-brazil:
Flapping alerts are orthogonal, nothing in this PR will change how often they happen as that's a Prometheus thing rather than an Alertmanager thing.

alburthoffman:
what do you mean by Prometheus thing instead of Alert manager thing? Could u provide more details? Thx

beorn7:
I don't think it's a good idea to generally inhibit all alerts that are firing in an otherwise equally labeled version with a higher severity. _Sometimes_ alerts are that way, there is a warning threshold x where it will file a ticket, and there is a critical a threshold where it will page you. We are all very used to that pattern because it was baked into traditional alerting tools like Nagios. _However_ that's not how alerts look in general. Once you have freed your mind from the notion that every alert _has_ to look that way, you'll see that most cases follow another pattern. It's also a nice mental exercise to force yourself to formulate the alerts _without_ that traditional warn/critical pattern. Often, a better alerting comes out of that. (And if not, you are still free to use the old pattern after all, but then as an informed decision.)

A quite neat example is the SLO-based alerting described in https://landing.google.com/sre/workbook/chapters/alerting-on-slos/ . Here, the page could go off without the ticket (error budget wasn't burned for a while but then a sudden spike happens), a ticket could go off without a page (slow burn of error budget), or both go off (steadily degrading error budget burn), or none (all good). In other word, it is important information which alert level fires.

Therefore, I'd strongly recommend to make "auto-inhibition" an opt-in feature of a well defined subset of alerts.

SuperQ:

Issue #2147, opened 2019-12-19
Alert on alertmanager_notifications_failed_total 


freeseacher:
**What did you do?**
I have and alert for alertmanager_notifications_failed_total 
```
  - alert: AlertmanagerNotificationsFailing
    expr: sum (increase(alertmanager_notifications_failed_total[1m])) by (integration) !=0
    labels:
      priority: P1
      service: alertmanager
    annotations:
      summary: 'Alertmanager is failing sending notications to {{$labels.integration}}'
```

It look like it should work fine but it is not. It is okay to get failure several times. 
What am i actually looking for is some metric showing that alert cannot be sent to upstream and alertmanager give up doing it

Any ideas ?
May be along with route label it will be much helpfull.

simonpasquier:
Not sure what you mean by

> It look like it should work fine but it is not. It is okay to get failure several times.

The metric should be sufficient to detect notification issues.

```
sum by (integration) (rate(alertmanager_notifications_failed_total[5m]))
 /
sum by (integration) (rate(alertmanager_notifications_total[5m])) > 0
```

>  May be along with route label it will be much helpfull.

There's an unbounded number of routes so it can't be a label.

I'm closing it for now. If you have further questions, please use our [user mailing list](https://groups.google.com/forum/#!forum/prometheus-users), which you can also search.

freeseacher:
@simonpasquier 
> There's an unbounded number of routes so it can't be a label.

that statement still can't out of my mind. and i really tried to live my life with it. 
for example `prometheus_rule_group_interval_seconds` is much more unbound, but still exposes filename and even rule group. 

I also tried to make an alert with 
```
sum by (integration) (rate(alertmanager_notifications_failed_total[5m]))
 /
sum by (integration) (rate(alertmanager_notifications_total[5m])) > 0
```
and i have 36 different receivers with that kind of integration. for at least one per team. 
And only one direct integration with slack, that will be triggered in case of opsgenie is not available. 

So getting that alert is at least misleading and forces me to use debug level of logging to even understand whats going on.

roidelapluie:
> @simonpasquier
> 
> > There's an unbounded number of routes so it can't be a label.
> 
> that statement still can't out of my mind. and i really tried to live my life with it.
> for example `prometheus_rule_group_interval_seconds` is much more unbound, but still exposes filename and even rule group.


[...]

PR #2273, opened 2020-06-02
notify: improve logs on notification errors

simonpasquier:
Alertmanager can experience occasional failures when sending
notifications to an external service. If the operation succeeds after
some retry, the 'alertmanager_notifications_failed_total' metric
increases but nothing is logged (unless running with log.level=debug).
Hence an operator might receive an alert about notification failures but
wouldn't know which integration was failing.

With this change, notification failures are logged at the warning level.
To avoid log flooding, similar failures on retries aren't logged.
Additional information on the failing integration has also been added.

cc @roidelapluie for review. Based on user's feedback from https://github.com/prometheus/alertmanager/issues/2147#issuecomment-626221046.

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -24,6 +24,7 @@ import (
 	"github.com/cespare/xxhash"
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
+	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
 
@@ -88,6 +89,11 @@ func (i *Integration) Name() string {
 // Index returns the index of the integration.
 func (i *Integration) Index() int {
 	return i.idx
+}
+
+// String implements the Stringer interface.
+func (i *Integration) String() string {
+	return fmt.Sprintf("%s[%d]", i.name, i.idx)
 }
 
 // notifyKey defines a custom type with which a context is populated to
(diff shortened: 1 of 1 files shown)


notify/notify.go
```diff
@@ -665,23 +664,27 @@ func (r RetryStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.Ale
 			r.metrics.numNotifications.WithLabelValues(r.integration.Name()).Inc()
 			if err != nil {
 				r.metrics.numFailedNotifications.WithLabelValues(r.integration.Name()).Inc()
-				level.Debug(l).Log("msg", "Notify attempt failed", "attempt", i, "integration", r.integration.Name(), "receiver", r.groupName, "retry", retry, "err", err)
 				if !retry {
-					return ctx, alerts, fmt.Errorf("cancelling notify retry for %q due to unrecoverable error: %s", r.integration.Name(), err)
+					return ctx, alerts, errors.Wrapf(err, "%s/%s: notify retry canceled due to unrecoverable error after %d attempts", r.groupName, r.integration.String(), i)
+				}
+				if ctx.Err() == nil && (iErr == nil || err.Error() != iErr.Error()) {
+					// Log the error if the context isn't done and the error isn't the same as before.
```
roidelapluie, line 671: should we log to debug all the retries?

notify/notify.go
```diff
@@ -665,23 +664,27 @@ func (r RetryStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.Ale
...
+					level.Warn(l).Log("msg", "Notify attempt failed, will retry later", "attempts", i, "err", err)
 				}
 
 				// Save this error to be able to return the last seen error by an
 				// integration upon context timeout.
 				iErr = err
 			} else {
+				level.Debug(l).Log("msg", "Notify success", "attempts", i)
```
roidelapluie, line 679: Should we log to info if i > 1 ?

roidelapluie:
LGTM, added a few suggestions that you are free to pick up

notify/notify.go
simonpasquier: Good idea!

notify/notify.go
simonpasquier: I'm on the fence as it could get highly verbose. Since we log an info message on success including the number of attempts, I feel that it's ok not logging every retry.

roidelapluie: Ok, if I am right logged "attempts" field will always be 1 then, is that correct?

roidelapluie: Oh no it won't be 1 if the error is different :)

PR #2274, opened 2020-06-02
Allow limiting maximum number of alerts in webhook

juliusv:
The webhook notifier is the only notifier that does not allow templating
on the Alertmanager side. Users who encounter occasional alert storms
(10ks of alerts going off at once for the same group) have reported
webhook receiver systems not being able to cope with the load caused by
the resulting large webhook notifier messages (the alerting rules also
contained large annotations that can't be stripped away due to lack of
templating). Reducing group size also wasn't an option, but this change
proposes to allow truncating the list of alerts sent in the webhook body
to a provided maximum length. This assumes that e.g. if a group receives
20k alerts, you really are fine only receiving 10k because you wouldn't
be able to check them all anyway.

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -409,6 +409,10 @@ type WebhookConfig struct {
 
 	// URL to send POST request to.
 	URL *URL `yaml:"url" json:"url"`
+	// MaxAlerts is the maximum number of alerts to be sent per webhook message.
+	// Alerts exceeding this threshold will be truncated. Setting this to 0
+	// allows an unlimited number of alerts.
+	MaxAlerts int `yaml:"max_alerts" json:"max_alerts"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -412,7 +412,7 @@ type WebhookConfig struct {
 	// MaxAlerts is the maximum number of alerts to be sent per webhook message.
 	// Alerts exceeding this threshold will be truncated. Setting this to 0
 	// allows an unlimited number of alerts.
-	MaxAlerts uint32 `yaml:"max_alerts" json:"max_alerts"`
+	MaxAlerts uint64 `yaml:"max_alerts" json:"max_alerts"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -412,7 +412,7 @@ type WebhookConfig struct {
 	// MaxAlerts is the maximum number of alerts to be sent per webhook message.
 	// Alerts exceeding this threshold will be truncated. Setting this to 0
 	// allows an unlimited number of alerts.
-	MaxAlerts int `yaml:"max_alerts" json:"max_alerts"`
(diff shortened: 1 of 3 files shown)


config/notifiers.go
```diff
@@ -409,6 +409,10 @@ type WebhookConfig struct {
 
 	// URL to send POST request to.
 	URL *URL `yaml:"url" json:"url"`
+	// MaxAlerts is the maximum number of alerts to be sent per webhook message.
+	// Alerts exceeding this threshold will be truncated. Setting this to 0
+	// allows an unlimited number of alerts.
+	MaxAlerts int `yaml:"max_alerts" json:"max_alerts"`
```
roidelapluie, line 415:
```suggestion
	MaxAlerts uint32 `yaml:"max_alerts" json:"max_alerts"`
```

simonpasquier:
LGTM once you apply Julien's suggestion :)

config/notifiers.go
juliusv: Done, had to adjust other lines as well.

notify/webhook/webhook.go
```diff
@@ -73,9 +73,17 @@ type Message struct {
 	GroupKey string `json:"groupKey"`
 }
 
+func truncateAlerts(maxAlerts uint32, alerts []*types.Alert) []*types.Alert {
+	if maxAlerts != 0 && uint32(len(alerts)) > maxAlerts {
```
roidelapluie, line 77:
```suggestion
	if maxAlerts != 0 &&len(alerts) > int(maxAlerts) {
```

roidelapluie: I think this is safer as we know that maxalert will always fit in an int, but len() cound not fit in uint32.

Also, maybe we should use uint() instead of uint32() for simplicity.

roidelapluie:
Another question:

Should we have a field in the webhook with the number of suppressed alerts? (and should that require moving to version: 5)

brian-brazil:
Adding a new field does not require a version bump, only if we break something.

notify/webhook/webhook.go
juliusv: How about I just use a uint64... then anything will fit in there, and it's not like we need to save space for this one :)

roidelapluie: uint sounds nicer to me and just shows what we want: a positive int. but it's up to you.

roidelapluie:
> Adding a new field does not require a version bump, only if we break something.

OK, that would still break the receivers who UnmarshalStrict; so it is somehow breaking

brian-brazil:
I'd not worry about someone choosing to use UnmarshalStrict on a server-generated message, strictness is for configs.

roidelapluie:
Ack

roidelapluie:
Then I'd suggest we add that field. If I set a <SMS> limit at 10 and I have 10 SMS, somehow I'd like to know if that was 100 alerts or just 10.

notify/webhook/webhook.go
simonpasquier: FWIW I'm fine with a `uint64`

juliusv:
I added a `truncatedAlerts` field (happy to name it differently if people have opinions), and made things be a uint64.

simonpasquier:
LGTM

PR #2276, opened 2020-06-04
Propagate labels to Opsgenie details

### LINKED ISSUE #682: opsgenie: propagate label k/v to opsgenie details
See discussion in: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/prometheus-users/Jmpxr8O5fXI/JTdlyfjBDAAJ

By default map k/v labels to the details in opsgenie alert api: https://www.opsgenie.com/docs/web-api/alert-api#createAlertRequest

It makes sense to use this part of the api as the k/v style of labels maps nicely over to their API, as well as leaving more room for data (8k, vs tags which is max 20 tags, each 50 char = 20x50=1000).

The more/better data is structured to opsgenie messages, the easier it is to route, and the richer the  semantics of the messages will be.

ricoberger:
Introduce two new config options for the Opsgenie receiver:

- `all_labels_as_details`: If `true`, all labels are propagated to the Opsgenie details (default is `false`).
- `labels_as_details`: An optional list of labels, which should be propagated to the Opsgenie details.

If the `all_labels_as_details` and `labels_as_details` value isn't set, the existing `details` option is used.

Closes #682.

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -22,7 +22,6 @@ import (
 	"github.com/pkg/errors"
 
 	commoncfg "github.com/prometheus/common/config"
-	"github.com/prometheus/common/model"
 )
 
 var (
@@ -457,18 +456,17 @@ type OpsGenieConfig struct {
 
 	HTTPConfig *commoncfg.HTTPClientConfig `yaml:"http_config,omitempty" json:"http_config,omitempty"`
 
-	APIKey           Secret                    `yaml:"api_key,omitempty" json:"api_key,omitempty"`
-	APIURL           *URL                      `yaml:"api_url,omitempty" json:"api_url,omitempty"`
-	Message          string                    `yaml:"message,omitempty" json:"message,omitempty"`
-	Description      string                    `yaml:"description,omitempty" json:"description,omitempty"`
-	Source           string                    `yaml:"source,omitempty" json:"source,omitempty"`
-	Details          map[string]string         `yaml:"details,omitempty" json:"details,omitempty"`
-	DetailsLabels    []string                  `yaml:"details_labels,omitempty" json:"details_labels,omitempty"`
-	DetailsLabelsAll bool                      `yaml:"-" json:"-"`
-	Responders       []OpsGenieConfigResponder `yaml:"responders,omitempty" json:"responders,omitempty"`
-	Tags             string                    `yaml:"tags,omitempty" json:"tags,omitempty"`
-	Note             string                    `yaml:"note,omitempty" json:"note,omitempty"`
-	Priority         string                    `yaml:"priority,omitempty" json:"priority,omitempty"`
+	APIKey                 Secret                    `yaml:"api_key,omitempty" json:"api_key,omitempty"`
+	APIURL                 *URL                      `yaml:"api_url,omitempty" json:"api_url,omitempty"`
+	Message                string                    `yaml:"message,omitempty" json:"message,omitempty"`
+	Description            string                    `yaml:"description,omitempty" json:"description,omitempty"`
+	Source                 string                    `yaml:"source,omitempty" json:"source,omitempty"`
+	Details                map[string]string         `yaml:"details,omitempty" json:"details,omitempty"`
+	DetailsUseCommonLabels bool                      `yaml:"details_use_common_labels,omitempty" json:"details_use_common_labels,omitempty"`
+	Responders             []OpsGenieConfigResponder `yaml:"responders,omitempty" json:"responders,omitempty"`
+	Tags                   string                    `yaml:"tags,omitempty" json:"tags,omitempty"`
+	Note                   string                    `yaml:"note,omitempty" json:"note,omitempty"`
+	Priority               string                    `yaml:"priority,omitempty" json:"priority,omitempty"`
 }
 
 const opsgenieValidTypesRe = `^(team|user|escalation|schedule)$`
@@ -494,19 +492,8 @@ func (c *OpsGenieConfig) UnmarshalYAML(unmarshal func(interface{}) error) error 
 		}
 	}
 
-	if c.DetailsLabels != nil && c.Details != nil {
-		return errors.Errorf("OpsGenieConfig can only contain details or details_labels, but both fields were provided")
-	}
-
-	for _, l := range c.DetailsLabels {
-		if l == "..." {
-			c.DetailsLabelsAll = true
-		} else {
-			labelName := model.LabelName(l)
-			if !labelName.IsValid() {
-				return errors.Errorf("invalid label name %q in details_labels list", l)
-			}
-		}
+	if c.Details != nil && c.DetailsUseCommonLabels == true {
+		return errors.Errorf("OpsGenieConfig can only contain details or details_use_common_labels, but both fields were provided")
 	}
 
 	return nil
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -22,6 +22,7 @@ import (
 	"github.com/pkg/errors"
 
 	commoncfg "github.com/prometheus/common/config"
+	"github.com/prometheus/common/model"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


simonpasquier:
Thanks for looking into this.

Instead of having 2 more parameters, maybe we can interpret `labels_as_details: ['...']` as "include all common labels in the details" (this would be similar to what has been done for `group_by`).

As it stands, `labels_as_details` and `details` are exclusive so it would be good IMO to fail the configuration parsing if the user configures both at the same time. Or maybe the resulting map should be a merge of both with `details` taking precedence over `labels_as_details`?

brian-brazil:
> this would be similar to what has been done for group_by

Grouping labels are not the same as common labels. The naming of these proposed config fields is confusing.

ricoberger:
Hi @simonpasquier and @brian-brazil, I made the suggested changes, so that there is only a `details_labels` field. The handling is similar to the `group_by` handling.

If the `details_labels` and `details` field is provided the loading of the configuration will fail now.

config/notifiers.go
```diff
@@ -22,6 +22,7 @@ import (
 	"github.com/pkg/errors"
 
 	commoncfg "github.com/prometheus/common/config"
+	"github.com/prometheus/common/model"
```
brian-brazil, line 25: Please don't add any dependencies on model.

config/notifiers.go
```diff
@@ -456,16 +457,18 @@ type OpsGenieConfig struct {
...
-	Tags        string                    `yaml:"tags,omitempty" json:"tags,omitempty"`
-	Note        string                    `yaml:"note,omitempty" json:"note,omitempty"`
-	Priority    string                    `yaml:"priority,omitempty" json:"priority,omitempty"`
+	APIKey           Secret                    `yaml:"api_key,omitempty" json:"api_key,omitempty"`
+	APIURL           *URL                      `yaml:"api_url,omitempty" json:"api_url,omitempty"`
+	Message          string                    `yaml:"message,omitempty" json:"message,omitempty"`
+	Description      string                    `yaml:"description,omitempty" json:"description,omitempty"`
+	Source           string                    `yaml:"source,omitempty" json:"source,omitempty"`
+	Details          map[string]string         `yaml:"details,omitempty" json:"details,omitempty"`
+	DetailsLabels    []string                  `yaml:"details_labels,omitempty" json:"details_labels,omitempty"`
```
brian-brazil, line 466: This seems redundant, you can already do this with the existing config.

ricoberger: Yes but as mentioned in https://github.com/prometheus/alertmanager/issues/682#issuecomment-476138382 this would be a shorter form of it. Also I think the overhead isn't that much and it is more flexible than introducing a `detailsLabelsAll` parameter.

ricoberger:
> This is still misleading. Common is not all, in fact all is not possible.

So you want to change the option to `details_common_labels` or sth. similar?

notify/opsgenie/opsgenie.go
```diff
@@ -120,9 +120,29 @@ func (n *Notifier) createRequest(ctx context.Context, as ...*types.Alert) (*http
...
+		}
+	} else if len(n.conf.DetailsLabels) > 0 {
+		details = make(map[string]string, len(n.conf.DetailsLabels))
+		for _, k := range n.conf.DetailsLabels {
+			label := string(k)
+
+			if v, ok := data.CommonLabels[label]; ok {
+				details[label] = v
+			} else {
+				details[label] = "N/A"
```
brian-brazil, line 138: This is business logic which may not apply to everyone.

config/notifiers.go
ricoberger: I removed the dependency.

config/notifiers.go
ricoberger: I changed this, so that there is only a `details_use_common_labels` option.

config/notifiers.go
```diff
@@ -491,6 +492,10 @@ func (c *OpsGenieConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
 		}
 	}
 
+	if c.Details != nil && c.DetailsUseCommonLabels == true {
```
bwplotka, line 495: :hugs: 

```suggestion
	if c.Details != nil && c.DetailsUseCommonLabels {
```

config/notifiers.go
```diff
@@ -491,6 +492,10 @@ func (c *OpsGenieConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
...
+		return errors.Errorf("OpsGenieConfig can only contain details or details_use_common_labels, but both fields were provided")
```
bwplotka, line 496: I have a tiny proposal: What about having this option `CommonLabelsAsDetails` and if true it will populate details from common labels, and then anything in `Details` will be added **additionally**? It will be bit more flexible for user and also quite simpler to implement and ensure. (:

notify/opsgenie/opsgenie.go
```diff
@@ -120,9 +120,18 @@ func (n *Notifier) createRequest(ctx context.Context, as ...*types.Alert) (*http
 
 	tmpl := notify.TmplText(n.tmpl, data, &err)
 
-	details := make(map[string]string, len(n.conf.Details))
-	for k, v := range n.conf.Details {
-		details[k] = tmpl(v)
+	var details map[string]string
```
bwplotka, line 123: Let's make map in both cases. It's not strictly needed to know size upfront -> it's only for micro optimization which is not needed here. (:

config/notifiers.go
```diff
@@ -456,16 +456,17 @@ type OpsGenieConfig struct {
...
-	Tags        string                    `yaml:"tags,omitempty" json:"tags,omitempty"`
-	Note        string                    `yaml:"note,omitempty" json:"note,omitempty"`
-	Priority    string                    `yaml:"priority,omitempty" json:"priority,omitempty"`
+	APIKey                 Secret                    `yaml:"api_key,omitempty" json:"api_key,omitempty"`
+	APIURL                 *URL                      `yaml:"api_url,omitempty" json:"api_url,omitempty"`
+	Message                string                    `yaml:"message,omitempty" json:"message,omitempty"`
+	Description            string                    `yaml:"description,omitempty" json:"description,omitempty"`
+	Source                 string                    `yaml:"source,omitempty" json:"source,omitempty"`
+	Details                map[string]string         `yaml:"details,omitempty" json:"details,omitempty"`
+	DetailsUseCommonLabels bool                      `yaml:"details_use_common_labels,omitempty" json:"details_use_common_labels,omitempty"`
```
bwplotka, line 465: I would propose:

```suggestion
	CommonLabelsAsDetails bool                      `yaml:"common_labels_as_details,omitempty" json:"common_labels_as_details,omitempty"`
```
as per my other comment.

ricoberger:
Hi @bwplotka, good point. I adjusted the code accordingly.

notify/opsgenie/opsgenie.go
```diff
@@ -120,7 +120,14 @@ func (n *Notifier) createRequest(ctx context.Context, as ...*types.Alert) (*http
 
 	tmpl := notify.TmplText(n.tmpl, data, &err)
 
-	details := make(map[string]string, len(n.conf.Details))
+	details := make(map[string]string)
+
```
bwplotka, line 124: to be very strict this new line could be omitted. It's best to only put new line to separate logically different sections, whereas the map is related to below if. Not a blocker (:

bwplotka:
Also do you think we can add a quick unit test @ricoberger ? (:

ricoberger:
> Also do you think we can add a quick unit test @ricoberger ? (:

Sure. I added some test cases to the existing Opsgenie test.

brian-brazil:
> then anything in Details will be added additionally?

I had had a similar thought. Would there be harm in always doing this, not needing a config option?

Also, if there is harm then is common labels safe as it's unpredictable what labels will actually be present?

bwplotka:
Yea, good question. I think this is breaking compatibility, but with very tiny bad effect, just additional details. It's up to maintainers to decide, cc @simonpasquier 

Since we are 0.x I think it's ok to just remove bool option and do it ;p 

Sorry @ricoberger for so many direction changes. Open source is a collaborative work!

brian-brazil:
> I think this is breaking compatibility but with very tiny bad effect, just additional details. It's up to maintainers to decide

Are you using OpsGenie so you can say exactly what might break? If that is the case then common labels is incorrect, and group labels should be used instead - though if it's group labels then you can do it with existing templating as you know the labels a-priori.

ricoberger:
> Sorry @ricoberger for so many direction changes. Open source is a collaborative work!

No problem 🙂

> > I think this is breaking compatibility but with very tiny bad effect, just additional details. It's up to maintainers to decide
> 
> Are you using OpsGenie so you can say exactly what might break? If that is the case then common labels is incorrect, and group labels should be used instead - though if it's group labels then you can do it with existing templating as you know the labels a-priori.

I think this shouldn't break anything in Opsgenie. Existing details would be the same, so that created routing rules shouldn't be effected.

Should I adjust the PR accordingly?

brian-brazil:
> I think this shouldn't break anything in Opsgenie. 

In that case making this hardcoded would seem the best approach.

ricoberger:
I removed the config option and adjusted the tests.

brian-brazil:
:+1: 

The docs over in the docs repo will also need an update.

osterman:
This is wondeful! Any chance a new release will be cut @bwplotka?

simonpasquier:
@osterman last release was mid-June. I'd expect we can have another one beginning of September.

Links2004:
Any timeline for the next release, this feature would help us a lot.

srueg:
Really looking forward to this, thanks! 🎉 
Any news on a release?

kedare:
Looks like the option is now gone or has never been released ?
I don't see it in the current documentation :
https://prometheus.io/docs/alerting/latest/configuration/#opsgenie_config

ricoberger:
I think this was never released, because last release is from last summer.

FlxPeters:
That should be included in 0.22.0, right?

roidelapluie:
Thank you, I have missed it!

ties:
For any other reader: In https://github.com/prometheus/alertmanager/pull/2276/commits/4b59db0adcb14f3e6b0b75efdbab3ffb7be737f6 it was changed to default to always passing all labels to OpsGenie.

darkl0rd:
Is there a way to override the behavior? We also raise alerts for our support department, some of the labels we add should ideally not  be visible to them? I’ve tried setting details explicitly, but it seems all labels are always added, no matter what I define?

ricoberger:
Hi @darkl0rd, no the behavior can not be changed, but it should be possible to overwrite the values for the labels which shouldn't be visible. When you set a label in the details section with an empty value, the original value for the label value shouldn't be visible in Opsgenie: https://github.com/prometheus/alertmanager/blob/44f8adc06af5101ad64bd8b9c8b18273f2922051/notify/opsgenie/opsgenie.go#L129

darkl0rd:
Thanks for the quick and detailed response. This should do the trick.

darkl0rd:
> Thanks for the quick and detailed response. This should do the trick.

Overriding with an empty value does indeed clear out the label. But the key remains visible in OpsGenie. Would it be an idea to adjust the code to only publish those keys which actually have a value set?

darkl0rd:
Honestly, after playing around with it some more - I'd really like to have a way to get rid of the ExtraProperties all together. As labels differ per alert type, it's a hell of a job to determine which ones I need to blank out where. I'd much rather just explicitly define which ones I do want to show or even just disable the entire "details" section when the alert is sent to our first line support department.

PR #2277, opened 2020-06-04
.circleci/config.yml: bump Prometheus orb version

simonpasquier:
Version `0..6.0` of the orb was buggy and led to container images not being pushed to the registries.

cc @juliusv


juliusv:
:+1: Thanks!

Issue #2286, opened 2020-06-07
Webhook notify retry too many times


bigHave:
**What did you do?**

In our prod environment, we use alertmanger to manage alerts fired from prometheus, and deliver alerts to another notify service by webhook receiver, as we use our own notify service to manage alert status and record alert history and so on.  Once there our notify service is down for about 10 minutes when alertmanager happen to deliver a firing alert to it. 
**What did you expect to see?**

alertmanager try to deliver that alert a few times and then abort.
**What did you see instead? Under which circumstances?**
Then I find in our record system that alertmanager has tried to deliver the very same alert so many times, like 200 times in 10 minutes.

So I did a experiment to see how webhook retry works.

**Environment**
I build alertmanager from souce (master branch) and run with following configuration(no cluster peer, just single instane), log level set to debug.

* Alertmanager configuration file:
```
global:
  resolve_timeout: 5m

route:
  group_by: ['alertname']
  group_wait: 10s
  group_interval: 10s
  repeat_interval: 1h
  receiver: 'web.hook'
receivers:
- name: 'web.hook'
  webhook_configs:
  - url: 'http://127.0.0.1:10086/alertmanager_test'
inhibit_rules:
  - source_match:
      severity: 'critical'
    target_match:
      severity: 'warning'
    equal: ['alertname', 'dev', 'instance']

```
where 127.0.0.1:10086/alertmanager_test was a simple http server that always response http status 500.

some of logs went as follows
* Logs:
```
level=debug ts=2020-06-01T12:20:12.936Z caller=dispatch.go:138 component=dispatcher msg="Received alert" alert=server down![1d7ec64][active]
level=debug ts=2020-06-01T12:20:12.936Z caller=dispatch.go:473 component=dispatcher aggrGroup="{}:{alertname=\"server down!\"}" msg=flushing alerts=[server down![1d7ec64][active]]
level=debug ts=2020-06-01T12:20:12.937Z caller=notify.go:668 component=dispatcher msg="Notify attempt failed" attempt=1 integration=webhook receiver=web.hook retry=true err="unexpected status code 500: http://127.0.0.1:10086/alertmanager_test"
level=debug ts=2020-06-01T12:20:13.417Z caller=notify.go:668 component=dispatcher msg="Notify attempt failed" attempt=2 integration=webhook receiver=web.hook retry=true err="unexpected status code 500: http://127.0.0.1:10086/alertmanager_test"
[...]

PR #2287, opened 2020-06-07
webhook retry timeout

bigHave:
For webhook retry, 
1. set retry timeout related to group_interval to avoid too many retries, like 1/5 of group_interval(in this case, retry timeout is max elapsed time of backoff ticker). Only if when group_interval is too large, set timeout to default.
2. if tick.C closed, reading from channel  ``` <-tick.C``` will get endless zero value, which will cause endless retry request until ticker closed. this looks like a bug to me, whether change timeout related to group_interval or not.

this PR is created to solve my problem discussed in  #2286


bigHave:
sry, seems something went wrong

PR #2289, opened 2020-06-09
Add image build for s390x architecture

vitt-bagal:
Added changes for s390x image build


vitt-bagal:
@simonpasquier Is it possible to retrigger failed `test check` ? As I could run `test` locally without any error.

simonpasquier:
sure, i've also rebased the PR against `release-0.21` so it gets included in the next release.

PR #2291, opened 2020-06-09
notify retry optimize

bigHave:
1.  always retry during a flush
2. return if timer stopped(by accident or what)

as discussed in issue #2286


simonpasquier:
Superseded by #2290. Thanks for the work and sorry for the duplication.

PR #2305, opened 2020-06-17
Add note about details handling for Opsgenie

roidelapluie:
See https://github.com/prometheus/docs/pull/1663


brian-brazil:
:+1:

PR #2316, opened 2020-07-02
config: add MarshalSecrets flag

### LINKED ISSUE #1985: Feature request: override secret censoring during marshaling to YAML
**What did you do?**

I am looking for a way to programmatically generate Alertmanager configurations. I imported `github.com/prometheus/alertmanager/config` into my project and used it to generate an Alertmanager config, which I then marshaled into a YAML string and wrote out to a Kubernetes ConfigMap.

**What did you expect to see? What did you see instead?**

I expected a working Alertmanager config, but all the API keys for my receivers had been marshaled into the string "<secret>". It would be great if there was a way to override this custom marshaling and get a working Alertmanager config with API keys and other secrets uncensored. They would still need to be censored by default, for places like the Alertmanager UI.

**Other approaches**

If secrets could be passed into the Alertmanager config using file locations, rather than directly as strings, then there would be no need to override the secret marhsaling behaviour. e.g.

```
slack_configs:
- api_url: /path/to/api_url
```

bobheadxi:
closes #1985

> we're trying to update Alertmanager configuration programatically (rather than by hand). The config package structs make this a lot easier, but unfortunately it is impossible to write updated configuration back to disk for Alertmanager to reload and use since secrets get replaced. Adding something like config.MarshalSecrets would allow anyone using the config package directly to write secrets when needed

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -34,6 +34,9 @@ const secretToken = "<secret>"
 
 var secretTokenJSON string
 
+// MarshalSecrets toggles whether or not secrets are marshalled
+var MarshalSecrets bool
+
 func init() {
 	b, err := json.Marshal(secretToken)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


francisminu:
Does that mean, at some point we might be able to pass a config to the POST reload method rather than being loaded from what is there on the disk?
-Thanks!

PR #2317, opened 2020-07-06
do not omit "empty" booleans in yaml output

0robustus1:
The "empty" boolean is the false value. This means that values of false will not
be present in the YAML output. As such they won't be displayed in the config
section on the Status overview page (/#/status), which can be especially
confusing for the SMTPRequireTLS (smtp_require_tls) field of the configuration
as this one will default to true.


simonpasquier:
Thanks!

PR #2325, opened 2020-07-16
Change HA exemple defaults to be more reasonable

roidelapluie:
10s+10s does not really change.

Give 30s for group wait and 5m for group interval seems better.


brian-brazil:
:+1:

simonpasquier:
Thanks!

PR #2330, opened 2020-07-24
WIP: Silence based on values outside the alert labels

zecke:
Sometimes it's desirable to silence the alerts that come from
a specific source, e.g. a single Prometheus instance that does
behave in a funny way and turning it off or using netfilter is
not an option.

Expand the the Muter interface to consume the alert labels and
an extra set of labels. Provide the generatorURL as an extra
label and prefix it with __ because it is a shared namespace.

Related: #344


types/types.go
```diff
@@ -390,18 +390,22 @@ func (a *Alert) Merge(o *Alert) *Alert {
 	return &res
 }
 
-// A Muter determines whether a given label set is muted. Implementers that
-// maintain an underlying Marker are expected to update it during a call of
-// Mutes.
+func (a Alert) GetExtraLabels() model.LabelSet {
+	return model.LabelSet{"__generatorURL": model.LabelValue(a.GeneratorURL)}
```
brian-brazil, line 394: __generatorURL is a valid (though odd) label that a Prometheus could send. This is best done via a separate mechanism, rather than trying to make it part of silencing via labels. This should also have no impact on inhibits - the generator URL should never affect alert identity.

SuperQ:

PR #2332, opened 2020-07-28
Add UUID To Alert

linclaus:
* add UUID attribute to each alert, so that we can locate them

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ jobs:
         command: sudo rm -rf /usr/local/go
     # Whenever the Go version is updated here, .promu.yml should also be updated.
     - go/install:
-        version: "1.13"
+        version: "1.14"
     - run:
         name: Remove generated code
         command: make clean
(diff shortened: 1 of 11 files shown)


roidelapluie:
That would not work in a HA setup.

linclaus:
> 
> 
> That would not work in a HA setup.

Can you outline the reasons?

simonpasquier:
Because when you have mulitple Alertmanager instances, each of them will generate a different UUID value while the alerts are the same. I'm not sure what you're trying to accomplish here.

roidelapluie:
And I think that alerts also offer a fingerprint now.

brian-brazil:
Alerts are already uniquely identified by their labels.

linclaus:
> And I think that alerts also offer a fingerprint now.

Fingerprint can only distinguish between alerts with different labels, but what if I want to distinguish each alert that has been sent?

linclaus:
> Because when you have mulitple Alertmanager instances, each of them will generate a different UUID value while the alerts are the same. I'm not sure what you're trying to accomplish here.

So we will receive multiple alerts from one Alertmanager route when there have multiple Alertmanager instances? From the recipient's point of view, I want to record and trace all the alerts I received.

SuperQ:

PR #2349, opened 2020-08-12
Remove resolve timeout from the config shipped by default

roidelapluie:
It is not needed with Prometheus and can confuse users.


simonpasquier (approved):
Thanks!

PR #2357, opened 2020-08-20
change snapshot Rlock() to Lock()

betNevS:
Snapshot() in `nflog.go` is a write operation， Lock() should be used to instead of Rlock().


simonpasquier:
Thanks for the patch! AFAICT the `mtx` field protects the [`st`](https://github.com/prometheus/alertmanager/blob/a8ed0d58509d0251adcb159552b9ab098e206f1f/nflog/nflog.go#L88-L89) field and the `Snapshot()` method only reads its value so `RLock()` is valid here.

PR #2363, opened 2020-09-02
Display Silence and Alert dates in ISO8601 format

nicholascapo:
* This changes the Alert view the SilenceList and SilenceView


simonpasquier:
I don't have strong preferences on date formatting for the UI but I'm not sure that switching to ISO8601 format is more useful than the current format.

nicholascapo:
My use case is actually comparing the dates in AM to the dates a lot of other tools (Grafana, PagerDuty, etc) that use the ISO format. If they all had the same structure, doing quick diffs in my head would be easier (e.g "How long before this page did the Silence expire?")

The change would also make these dates consistent with the format on the New Silence Page.

Personally I like the ISO format, but internal and external consistency is the goal here.

simonpasquier:
Thanks for the input Nicholas. I agree that reducing cognitive overhead is important. I'll let @roidelapluie and @w0rm chime in too...

roidelapluie:
We should keep that consistent with Prometheus.

nicholascapo:
See also: https://github.com/prometheus/prometheus/pull/7884

w0rm:
> We should keep that consistent with Prometheus.

I agree that we should keep it consistent! 👍 if we are going change this in Prometheus too.

nicholascapo:
Any word on this change? Can I get an approval?

w0rm (approved):

w0rm:
@nicholascapo sure, to make it more consistent with Prometheus 👍

PR #2373, opened 2020-09-15
[#2372] Move config reload metrics to Coordinator.Reload()

### LINKED ISSUE #2372: Coordinator subscriber errors are excluded in alertmanager_config_last_reload_successful
**What did you do?**
Update a template and triggered a config reload. This template was invalid.
The metric alertmanager_config_last_reload_successful being set to false.

The metric alertmanager_config_last_reload_successful was set to true, while the Coordinator was reporting an invalid template in the logs.

**Extras**
This triggered an incident for us. What happened is AM got scheduled on non-preemptible nodes and therefore was running for a while. This template was updated a month ago and the reload metric remained true, even after a bunch of reloads during the template update and now, hence we never noticed the reload to fail. After draining these nodes, all AM's got rescheduled and could not boot up because of invalid template syntax which we were not aware of.

In other words, this template failure is silent on reload and catastrophic on reboot. Especially in combination with Thanos ruler. As ruler will not boot up if it can not resolve an IP for AM. Wich in k8s will happen if there are no ready pods behind a service.

This is due to the following code: 
https://github.com/prometheus/alertmanager/blob/1fdff6b3f9394805b3b9a58fd26e9f055d85b282/config/coordinator.go#L119-L130

https://github.com/prometheus/alertmanager/blob/1fdff6b3f9394805b3b9a58fd26e9f055d85b282/config/coordinator.go#L96-L111

After setting the metrics to success, AM triggers the subscribers to do their work: 
https://github.com/prometheus/alertmanager/blob/1fdff6b3f9394805b3b9a58fd26e9f055d85b282/config/coordinator.go#L136-L143

The template subscriber is created here: https://github.com/prometheus/alertmanager/blob/1fdff6b3f9394805b3b9a58fd26e9f055d85b282/cmd/alertmanager/main.go#L381-L454

and does not expose any type of metrics if this succeeded or not. 

The following code makes the reboot catastrophic:
https://github.com/prometheus/alertmanager/blob/1fdff6b3f9394805b3b9a58fd26e9f055d85b282/cmd/alertmanager/main.go#L456-L458

**Environment**

* System information:

* Alertmanager version:
	v0.21.0

```
insert configuration here
```

* Logs:
```
ts=2020-09-11T12:20:27.391Z caller=coordinator.go:137 component=configuration msg="one or more config change subscribers failed to apply new config" file=/etc/config/alertmanager.yml err="failed to parse templates: template: messa
gebird.tmpl:21: invalid syntax"
```

OGKevin:
Closes #2372

Signed-off-by: Kevin Hellemun <17928966+OGKevin@users.noreply.github.com>

### DIFF
--- a/config/coordinator.go
+++ b/config/coordinator.go
@@ -16,11 +16,11 @@ package config
 import (
 	"crypto/md5"
 	"encoding/binary"
+	"sync"
+
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
-	"sync"
-	"time"
 )
 
 // Coordinator coordinates Alertmanager configurations beyond the lifetime of a
@@ -139,7 +139,7 @@ func (c *Coordinator) Reload() error {
 	}
 
 	c.configSuccessMetric.Set(1)
-	c.configSuccessTimeMetric.Set(float64(time.Now().Unix()))
+	c.configSuccessTimeMetric.SetToCurrentTime()
 	hash := md5HashAsMetricValue([]byte(c.config.original))
 	c.configHashMetric.Set(hash)
 
--- a/config/coordinator.go
+++ b/config/coordinator.go
@@ -16,12 +16,11 @@ package config
 import (
 	"crypto/md5"
 	"encoding/binary"
-	"sync"
-	"time"
-
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
+	"sync"
+	"time"
 )
 
 // Coordinator coordinates Alertmanager configurations beyond the lifetime of a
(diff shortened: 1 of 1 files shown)


config/coordinator.go
```diff
@@ -16,12 +16,11 @@ package config
...
 	"crypto/md5"
 	"encoding/binary"
-	"sync"
-	"time"
-
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
 	"github.com/prometheus/client_golang/prometheus"
+	"sync"
+	"time"
```
simonpasquier, line 23: could you revert that part?

config/coordinator.go
```diff
@@ -139,9 +134,15 @@ func (c *Coordinator) Reload() error {
 			"file", c.configFilePath,
 			"err", err,
 		)
+		c.configSuccessMetric.Set(0)
 		return err
 	}
 
+	c.configSuccessMetric.Set(1)
+	c.configSuccessTimeMetric.Set(float64(time.Now().Unix()))
```
simonpasquier, line 142:
```suggestion
	c.configSuccessTimeMetric.SetToCurrentTime()
```

simonpasquier (commented):
Very nice! Just a few nits and we're good.

config/coordinator.go
OGKevin: ah my bad., auto format.

simonpasquier:
LGTM, will merge on green. Thanks!

PR #2378, opened 2020-09-24
Webhook supports custom message, Make webhook notification more generic. (e.g. dingtalk/wechat ...)

virnet:
Signed-off-by: 孙林耀 <sunlinyao@vip.qq.com>

### DIFF
--- a/notify/webhook/webhook.go
+++ b/notify/webhook/webhook.go
@@ -100,8 +100,30 @@ func (n *Notifier) Notify(ctx context.Context, alerts ...*types.Alert) (bool, er
 	}
 
 	var buf bytes.Buffer
-	if err := json.NewEncoder(&buf).Encode(msg); err != nil {
-		return false, err
+	if n.conf.Json != "" {
+		body, err := n.tmpl.ExecuteTextString(n.conf.Json, data)
+		if err != nil {
+			return false, err
+		}
+		if !json.Valid([]byte(body)) {
+			return false, fmt.Errorf("Json格式异常: %s ", body)
+		}
+		if err := json.Indent(&buf, []byte(body), "", "  "); err != nil {
+			return false, fmt.Errorf("Json格式异常: %s ", body)
+		}
+	} else if n.conf.Text != "" {
+		body, err := n.tmpl.ExecuteTextString(n.conf.Json, data)
+		if err != nil {
+			return false, err
+		}
+		err = json.NewEncoder(&buf).Encode(body)
+		if err != nil {
+			return false, err
+		}
+	} else {
+		if err := json.NewEncoder(&buf).Encode(msg); err != nil {
+			return false, err
+		}
 	}
 
 	req, err := http.NewRequest("POST", n.conf.URL.String(), &buf)
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@
 !/.travis.yml
 !/.promu.yml
 !/api/v2/openapi.yaml
+!examples/dingtalk/alertmanager.yml
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -372,6 +372,10 @@ type WebhookConfig struct {
 	// Alerts exceeding this threshold will be truncated. Setting this to 0
 	// allows an unlimited number of alerts.
 	MaxAlerts uint64 `yaml:"max_alerts" json:"max_alerts"`
+
+	// Webhook content format
+	Json string `yaml:"json,omitempty" json:"json,omitempty"`
+	Text string `yaml:"text,omitempty" json:"text,omitempty"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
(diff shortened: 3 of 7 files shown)


simonpasquier:
Thanks for your contribution! It's already been discussed before (see https://github.com/prometheus/alertmanager/issues/701) but we're not sure yet how it can be solved properly.
In any case, generating JSON data from Go templates isn't the right approach IMHO as it's going to be error prone and brittle.

notify/webhook/webhook.go
```diff
@@ -100,8 +100,30 @@ func (n *Notifier) Notify(ctx context.Context, alerts ...*types.Alert) (bool, er
...
+			return false, err
+		}
+		if !json.Valid([]byte(body)) {
+			return false, fmt.Errorf("Json格式异常: %s ", body)
+		}
+		if err := json.Indent(&buf, []byte(body), "", "  "); err != nil {
+			return false, fmt.Errorf("Json格式异常: %s ", body)
+		}
+	} else if n.conf.Text != "" {
+		body, err := n.tmpl.ExecuteTextString(n.conf.Json, data)
```
yo000, line 115: s/n.conf.Json/n.conf.Text/

notify/webhook/webhook.go
```diff
@@ -100,8 +100,30 @@ func (n *Notifier) Notify(ctx context.Context, alerts ...*types.Alert) (bool, er
...
+		if err != nil {
+			return false, err
+		}
+		err = json.NewEncoder(&buf).Encode(body)
+		if err != nil {
+			return false, err
+		}
```
yo000, line 122: No need to json encode if this is Text template imho, "buf.WriteString(body)" should be sufficient

yo000:
I use virnet patch with minor mods (included), so we can notify to webhooks with formencoded content POST API.

in alertmanager.yml :
```templates:
- '/usr/local/etc/alertmanager/template/form.tmpl'

receivers:
- name: 'astreinte_sms'
  webhook_configs:
    - url: 'https://sms.service.com/api/sms/send'
      content_type: 'application/x-www-form-urlencoded'
      text: '{{ template "sms_service.text" . }}'
```

In template/form.tmpl :
```
cat << EOF >> template/form.tpl
{{ define "sms_service.text" }}key=__insert_token__&message={{ .GroupLabels.status }}:{{ .GroupLabels.alertname }}:{{ .CommonAnnotations.description }}&destinataires="+0123456789"&expediteur=sender{{ end }}
EOF
```

[patch_ctype_after_PR2378.diff.txt](https://github.com/prometheus/alertmanager/files/6869035/patch_ctype_after_PR2378.diff.txt)

thequailman:
What's stopping this from being merged?  This would be really nice to have.

notify/webhook/webhook.go
virnet: thanks

roidelapluie:
See Simon's comment.

I agree that it is awesome to be able to customize the message, but go templates are far from ideal for this.

yuri-tceretian:
@roidelapluie @simonpasquier what do you think about [jsonnet ](https://jsonnet.org/) as an alternative?

simonpasquier:
@yuri-tceretian I think that @jan--f did an experiment but it wasn't really successful (the processing wasn't really efficient).
As much as I like jsonnet (or at least not dislike it), I'm wary that there would be a usability gap as not so many people are familiar with it.

jan--f:
Yes the experiment can still be found at https://github.com/jan--f/alertmanager/tree/webhook-text-templates.
This uses github.com/google/go-jsonnet 0.17, so maybe something has changed since then.
Simply run `go test ./notify/webhook/ -bench . -memprofile mem.pprof -cpuprofile cpu.pprof -count 10` on one of the last three commits on that branch (they implement different approaches).

This is a very naive implementation no doubt but I think I saw 40x - 80x slowdown even for a simple oneline template.

notify/webhook/webhook.go
MicroOps-cn: > No need to json encode if this is Text template imho, "buf.WriteString(body)" should be sufficient

If neither the JSON template nor the text template is configured, then the message should be converted into JSON according to the original logic.

MicroOps-cn:
> Thanks for your contribution! It's already been discussed before (see #701) but we're not sure yet how it can be solved properly. In any case, generating JSON data from Go templates isn't the right approach IMHO as it's going to be error prone and brittle.

I also think it's not very wise to generate JSON through go templates, but I haven't found a better solution.

When you say 'it's going to be error prone and brick', do you mean that there are special symbols that can cause JSON formatting exceptions when generating JSON?

I have come up with a way to use the custom function toJson/safeJson to escape any special characters that may occur, thus avoiding this issue.

simonpasquier:
My main worry is that it will very hard for people to use and debug. Go templates aren't made to generate structured format like JSON.

TheMeier:
IMHO that overlaps with https://github.com/prometheus/alertmanager/pull/2184

123-soleil:
I applied this PR to release 0.29 before building because I needed this feature, and it works perfectly:
No build errors
No errors during operation

SoloJacobs:
I am closing this PR in favor of https://github.com/prometheus/alertmanager/pull/5011 since it has not had any activity.

Kind regards

PR #2379, opened 2020-09-24
notify/pagerduty: Filter out empty images and links

BenoitKnecht:
PagerDuty Event API v2 [1] requires images to have an `src` property, and links
to have an `href` property.

This commit filters out images and links that don't satisfy those conditions,
to avoid getting an HTTP 400 error in response.

This also adds flexibilty when using templates to configure images and links,
as it's now possible to omit images or links by letting the template return an
empty string for the `src` or `href` property, respectively.

[1]: https://developer.pagerduty.com/docs/events-api-v2/trigger-events/#context-properties

Signed-off-by: Benoît Knecht <bknecht@protonmail.ch>


simonpasquier (approved):
Thanks!

fredr:
Hey! Just ran into this one. Any chance of getting this into a release? /cc @simonpasquier

PR #2380, opened 2020-09-27
Add some missing PagerDuty fields to docs

grdryn:
These fields are all available on the PagerdutyConfig type:
https://github.com/prometheus/alertmanager/blob/59a96579cc9d117ae72cfb6ad45033ee746c6ba8/config/notifiers.go#L201..L203

I took the descriptions from the PagerDuty events-api-v2 docs:
https://developer.pagerduty.com/docs/events-api-v2/overview/

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -406,6 +406,15 @@ images:
 links:
   [ <link_config> ... ]
 
+# The part or component of the affected system that is broken.
+[ component: <tmpl_string> ]
+
+# A cluster or grouping of sources.
+[ group: <tmpl_string> ]
+
+# The class/type of the event.
+[ class: <tmpl_string> ]
+
 # The HTTP client's configuration.
 [ http_config: <http_config> | default = global.http_config ]
 ```


grdryn:
@simonpasquier this finally follows up on https://github.com/prometheus/docs/pull/1627

docs/configuration.md
```diff
@@ -406,6 +406,15 @@ images:
 links:
   [ <link_config> ... ]
 
+# The part or component of the affected system that is broken.
+  [ component: <tmpl_string> ]
```
simonpasquier, line 410: can you remove the trailing spaces? I believe they are unneeded.

docs/configuration.md
```diff
@@ -406,6 +406,15 @@ images:
...
+
+# A cluster or grouping of sources.
+  [ group: <tmpl_string> ]
```
simonpasquier, line 413: ditto

docs/configuration.md
```diff
@@ -406,6 +406,15 @@ images:
...
+
+# The class/type of the event.
+  [ class: <tmpl_string> ]
```
simonpasquier, line 416: ditto

docs/configuration.md
grdryn: Done (I presume you meant leading spaces rather than trailing spaces). :+1:

simonpasquier: Of course! Thanks for deciphering :)

Issue #2361, opened 2020-08-27
Separate metrics for temporary and permanently failed notifications


SuperQ:
With the current `alertmanager_notifications_failed_total` metric, it's difficult to tell failures that work on a retry, and failures that have been permanently dropped.

mneverov:
hi, i'd like to work on this

mapshen:
@simonpasquier are we making a release sometime this month? Would like to start using these metrics.

PR #2383, opened 2020-10-04
Add metric for permanently failed notifications

mneverov:
Fixes: #2361 

Signed-off-by: Max Neverov <neverov.max@gmail.com>

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -213,9 +213,11 @@ type NotificationLog interface {
 }
 
 type metrics struct {
-	numNotifications           *prometheus.CounterVec
-	numFailedNotifications     *prometheus.CounterVec
-	notificationLatencySeconds *prometheus.HistogramVec
+	numNotifications                   *prometheus.CounterVec
+	numTotalFailedNotifications        *prometheus.CounterVec
+	numNotificationRequestsTotal       *prometheus.CounterVec
+	numNotificationRequestsFailedTotal *prometheus.CounterVec
+	notificationLatencySeconds         *prometheus.HistogramVec
 }
 
 func newMetrics(r prometheus.Registerer) *metrics {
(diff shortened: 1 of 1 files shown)


mneverov:
@SuperQ could you please take a look?

notify/notify.go
```diff
@@ -665,8 +672,9 @@ func (r RetryStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.Ale
 			r.metrics.notificationLatencySeconds.WithLabelValues(r.integration.Name()).Observe(time.Since(now).Seconds())
 			r.metrics.numNotifications.WithLabelValues(r.integration.Name()).Inc()
 			if err != nil {
-				r.metrics.numFailedNotifications.WithLabelValues(r.integration.Name()).Inc()
+				r.metrics.numTotalFailedNotifications.WithLabelValues(r.integration.Name()).Inc()
 				if !retry {
+					r.metrics.numPermanentlyFailedNotifications.WithLabelValues(r.integration.Name()).Inc()
```
simonpasquier, line 677: the metric needs to be incremented before L701 too (when the context is done).

mneverov: you're right. Fixed, thanks.

notify/notify.go
```diff
@@ -225,11 +226,16 @@ func newMetrics(r prometheus.Registerer) *metrics {
...
 		}, []string{"integration"}),
-		numFailedNotifications: prometheus.NewCounterVec(prometheus.CounterOpts{
+		numTotalFailedNotifications: prometheus.NewCounterVec(prometheus.CounterOpts{
 			Namespace: "alertmanager",
 			Name:      "notifications_failed_total",
 			Help:      "The total number of failed notifications.",
 		}, []string{"integration"}),
+		numPermanentlyFailedNotifications: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: "alertmanager",
+			Name:      "notifications_permanently_failed_total",
```
simonpasquier, line 236: Not totally convinced by the metric name. We should also count the number of tries.
I'm wondering if we shouldn't use the existing `alertmanager_notifications_failed_total` to count notifications that couldn't be delivered permanently and introduce `alertmanager_notification_requests_total`/`alertmanager_notification_requests_failed_total` metrics to count notification request attempts and failures.

mneverov: @simonpasquier thanks for the update. I tried to be non-intrusive and keep the existing meaning for the `notifications_failed_total`. I agree that your suggested metrics have much clearer intent given these names. 
Should I introduce them and delete the old metric or just add the new ones?

mneverov:
@simonpasquier could you please have another look?

simonpasquier (commented):
:+1: thanks!

Issue #2323, opened 2020-07-16
favicon doesn't resolve with path prefix


daluca:
**What did you do?**
Launched a docker version of alertmanager using the below config to add a path prefix and reverse proxy to access alertmanager

The browser is looking for the favicon at the root level without the path prefix
```
https://#######.#####.###/favicon.ico
```
``` json
{"GET":{"scheme":"https","host":"#######.#####.###","filename":"/favicon.ico","remote":{"Address":"###.###.###.###:443"}}}
```

The rest of the application runs fine and is adding the path prefix

**What did you expect to see?**
The favicon appear in the browser tab

**What did you see instead? Under which circumstances?**
The default icon for my browser

**Environment**

* System information:

        Linux 4.14.170 x86_64

* Alertmanager version:

        alertmanager, version 0.21.0 (branch: HEAD, revision: 4c6c03ebfe21009c546e4d1e9b92c371d67c021d)
          build user:       root@dee35927357f
          build date:       20200617-08:54:02
          go version:       go1.14.4

* Alertmanager configuration file:
```
global:
  resolve_timeout: 5m

route:
  group_by: ['alertname']
  group_wait: 10s
  group_interval: 10s
  repeat_interval: 1h
  receiver: 'web.hook'
receivers:
- name: 'web.hook'
  webhook_configs:
  - url: 'http://127.0.0.1:5001/'
inhibit_rules:
- source_match:
    severity: 'critical'
  target_match:
    severity: 'warning'
  equal: ['alertname', 'dev', 'instance']

```

* docker-compose.yml
```
  alertmanager:                                      
    image: prom/alertmanager                         
    container_name: alertmanager
    restart: unless-stopped
    command:
      - --config.file=/etc/alertmanager/alertmanager.yml
      - --storage.path=/alertmanager
      - --data.retention=120h
      - --web.external-url=https://#######.#####.###/alertmanager
      - --cluster.listen-address=:9094 
      - --cluster.peer=#######.#####.###:9094
      - --log.format=json
    ports:
      - 9093:9093/tcp
      - 9094:9094/tcp
    labels:
      - traefik.enable=true
      - traefik.http.routers.alertmanager.rule=Host(`#######.#####.###`) && (PathPrefix(`
/alertmanager`))
      - traefik.http.services.alertmanager.loadbalancer.server.port=9093
`) && (PathPrefix(`/alertmanager`))
      - traefik.http.routers.alertmanager.tls=true
[...]

PR #2390, opened 2020-10-10
registering /favicon.ico prior to setting a prefix in the router

JBerny:
The favico.ico is always requested at /favico.ico by the browser. You may need to disable browser caching to see this working.


JBerny:
Fixes #2323

roidelapluie:
I think the fix would be to correctly set <link rel="icon" href="favicon.ico" />

PR #2391, opened 2020-10-11
Add support for client certificates in amtool

Hsn723:
I'd like to propose the ability to provide a client certificate in `amtool` to support use cases where an instance of `alertmanager` is behind and Ingress that expects a client certificate.


Hsn723:
@simonpasquier 
Thank you for the feedback. I have implemented your suggestions.

It also seems I'm failing some CI tests unrelated to `amtool`. Should I rebase?

roidelapluie:
I suggested that we base the effort on Prometheus' protool and maybe provide a config file so we can support basic auth passwords too.

roidelapluie:
cc @brian-brazil

roidelapluie:
In both ways, file or not, we should add the necessary helpers in common to support this use case

TheMeier:
Sounds like it could be useful. Maybe it is time to revive that PR

TheMeier:
Hm although there is 
  http.config.file
  	HTTP client configuration file for amtool to connect to Alertmanager.
  	The format is https://prometheus.io/docs/alerting/latest/configuration/#http_config.

which allows to configure TLS, so is that actually  needed?

TheMeier:
I guess this is obsoleted by https://github.com/prometheus/alertmanager/pull/2764

SoloJacobs:
Hi @Hsn723 ,

it looks like this PR is out of date and has some overlap with other changes, that have been merged in the mean time. So, I'll close this. We appreciate the time you put into this. If feel like opening an up-to-date PR, that I'll make sure to review it in a timely manor.

Kind regards, Solomon Jacobs

Issue #876, opened 2017-06-20
Time of day based alert routing/notification


brian-brazil:
We've had numerous requests for routing alerts based on the time of day/week. This issue is to track those.

tyrken:
So, @brian-brazil (@fabxc ?), could you provide any design guidance on how to implement said feature, as I think I'd like to suppress all 'severity: warning' alerts overnight rather than putting loads of time-based rule duplication into my Prometheus alert rules.  I really don't want to do that as warnings are still valid & worth warning about if I go looking for current Alerts active - I just don't want to be woken up for them.  They also shouldn't seem to resolve every evening & start again in the morning.

After being burnt wasting effort on #709 I want some suggestion up front of what might be accepted from the maintainers.

ainola:
Indeed, it would be nice to see whether it is within the scope of AM. It'd be great to have this feature.

mpsmith:
I would like to be able to have time of day, or day of week influence which receiver an alert is sent to.  i.e. - during daytime/business hours, alerts might go to a slack channel, vs during night/weekends, same alerts might go to pagerduty, or on e-mail for the current on-call person.

From the Alertmanager perspective, it could be nice to use existing label matching to control routing to different receivers based on datetime.   

i.e. - 
```
- match_re:
      alertname: MyAlert
      day: '(Monday|Tuesday|Wednesday|Thursday|Friday)'
   receiver: slack_appteam
- match_re:
      alertname: MyAlert
      day: '(Saturday|Sunday)'
   receiver: pagerduty_appteam
```

Something similar for time of day?   It's a little tricker, and in routing, it would be nice keep things simple...  i.e. - match_re on a label like 'time_window: business_hours', but I'm not sure how to get that meaningful label in there from the alert manager perspective without some sort of relabeling within alertmanager itself, and prometheus passing along an alert date/time.   I'm a Prometheus/Alertmanager newbie, so apologies in advance if I'm missing something obvious here.

An approach that generates meaningful date/time labels on the alerts, means those labels could also be used in inhibition as per @tyrken 's request to inhibit warnings for some or all alerts overnight.

One of my main drivers for this is to not introduce time based rule duplication in all my prometheus alerts, as that feels cumbersome.

etiennechabert:
+1 : Just here to says it would be a lovely feature to be able to sleep well during the night and have some alerts only during business hours / days

stuartnelson3:
[...]

Issue #226, opened 2016-01-10
Send resolved notification for silenced alerts


fabxc:
Alerts should be marked as resolved as they are silenced so a respective resolved notification is sent. This removes the need to manually resolve these in PagerDuty and friends.

We should probably provide information that tells whether it was a an actual resolve or a resolve-via-silence.

This puts another dimension to the problem of pre-computing silences (i.e. not silence at notification time) in a sane way.

raypettersen:
This would be very useful for teams. Hope to see this feature soon.

lswith:
you could implement this with sending an acknowledge instead of a resolved notification?

jkemp101:
I think I have a slightly different use case.  We often have 3 levels of alerts info, warning, critical.  At critical Pager Duty gets notified. All alert levels send notifications to hipchat. I often silence an alert when I see its at the warning level so if it goes critical it doesn't notify pager duty.  But then I miss the resolved hipchat message so I have to keep checking the status of the alert manually.  I don't really want to silence resolves.

I wouldn't mind a different message going HC letting everyone know I silenced it, that would be useful info.  But I wouldn't want a message going to HC stating the alert has be resolved when I silence it, that would be misleading.

homelessnessbo:
+1 we use our own status-board to notify external users if any problems in infrastructure occurs. We use webhooks to create events there, so, when problem occurs - we want external users to see it on status-board but sometimes we don't want to get bunch of emails (slack, etc) notifies as we already know about the problem, so we silence an alert. But if there is no resolve web-hook since alert is silenced - we see that problem is still present on our status-board when it's actually resolved. That brings some confusion and it would be nice if silenced alerts would report resolved state.

kamalmarhubi:
This would be incredibly useful.

@fabxc
> This puts another dimension to the problem of pre-computing silences (i.e. not silence at notification time) in a sane way.

Is there another issue or something with an explanation of what you mean here?

kamalmarhubi:
@fabxc could you elaborate a bit? I'm willing to put half a day or so into this, as it'd be a great improvement to the Prometheus + Pagerduty combo. The implementation looks fairly straighforward except for uncertainty from your note above.

kamalmarhubi:
@fabxc ping on the above question: how do you see this interact with precomputed silences? I would like to work on this, but would like to avoid doing something that has to be thrown away in the near future.

kamalmarhubi:
[...]

PR #2393, opened 2020-10-13
Add time-based muting to routing tree

benridley:
Addresses #876 

This adds the ability to define time intervals as per [this design document](https://docs.google.com/document/d/1pf-rPDQUGJUHazyr5vanTO6ft3loNZO9UoVpvhShFtA/edit?usp=sharing). 

Time intervals are defined in their own configuration section like so:
```yaml
mute_time_intervals:
  - name: business_hours
    time_intervals:
      - weekdays: ['monday:friday']
        times: 
        - start_time: "09:00"
          end_time: "17:00"
```
Then they can be referenced in the routing tree like so:
```yaml
  # The root route on which each incoming alert enters.
route:
  group_by: ['alertname', 'cluster', 'service']
  group_wait: 1s
  group_interval: 30s

  # A default receiver
  receiver: team-X-mail
  routes:
    - match:
        alertname: NodeIsDown
      mute_times:
        - business_hours
      receiver: team-X-mail
```
Time zones are not supported at the moment because it would break the ability to run Alertmanager on Windows in a reliable way. Once [this issue](https://github.com/golang/go/issues/38453) is addressed we can add support fairly easily.


benridley:
Just added some additional tests for the config

benridley:
@simonpasquier Can you please let me know if there's anything else I can do for this one? Thank you!

Neylion:
This would help our team greatly. Hope you get time to look at this sooner rather than later! @simonpasquier

juliusv:
@simonpasquier There seems to be good interest in this one... in case you are held up by other stuff, maybe someone else wants to take a look? @roidelapluie @beorn7 

I am unfortunately totally swamped at the moment.

config/config.go
```diff
@@ -408,9 +428,20 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if len(c.Route.Match) > 0 || len(c.Route.MatchRE) > 0 {
 		return fmt.Errorf("root route must not have any matchers")
 	}
+	if len(c.Route.MuteTimes) > 0 {
+		return fmt.Errorf("root route cannot have any mute times")
```
roidelapluie, line 432: Why this limitation?

docs/configuration.md
```diff
@@ -202,6 +212,65 @@ route:
...
+
+A `mute_time_interval` specifies a named interval of time that may be referenced
+in the routing tree to mute particular routes for particular times of the day.
+
+```yaml
+name: <string>
+time_intervals:
+  [ - <time_interval> ... ]
+
+```
```
roidelapluie, line 225:
```suggestion
```

```

File: timeinterval/timeinterval.go [hunk 3]
```diff
@@ -0,0 +1,473 @@
...
+// within the interval.
+type TimeInterval struct {
+	Times       []TimeRange       `yaml:"times,omitempty"`
+	Weekdays    []WeekdayRange    `yaml:"weekdays,flow,omitempty"`
+	DaysOfMonth []DayOfMonthRange `yaml:"days_of_month,flow,omitempty"`
+	Months      []MonthRange      `yaml:"months,flow,omitempty"`
+	Years       []YearRange       `yaml:"years,flow,omitempty"`
+}
+
+/* TimeRange represents a range of minutes within a 1440 minute day, exclusive of the End minute. A day consists of 1440 minutes.
```
roidelapluie, line 35:
```suggestion
// TimeRange represents a range of minutes within a 1440 minute day, exclusive of the End minute. A day consists of 1440 minutes.
```

timeinterval/timeinterval.go
```diff
@@ -0,0 +1,473 @@
...
+	1: "monday",
+	2: "tuesday",
+	3: "wednesday",
+	4: "thursday",
+	5: "friday",
+	6: "saturday",
+}
+
+var months = map[string]int{
+	"january":   1,
```
roidelapluie, line 136: Months in English start with a capital letter if I'm right

notify/notify.go
```diff
@@ -729,3 +745,44 @@ func (n SetNotifiesStage) Exec(ctx context.Context, l log.Logger, alerts ...*typ
...
+
+	muted := false
+	for _, mtName := range muteTimeNames {
+		mt, ok := mts.muteTimes[mtName]
+		if !ok {
+			return ctx, alerts, errors.Errorf("mute time %s doesn't exist in config", mtName)
+		}
+		for _, ti := range mt {
+			if ti.ContainsTime(now) {
+				muted = true
```
roidelapluie, line 777:
```suggestion
				muted = true
				break
```
You can break the outer loop here

timeinterval/timeinterval.go
benridley: They're just stored as lowercase here for convenience and then downcased during the lookup, so users can specify it with or without capitals.

notify/notify.go
```diff
@@ -729,3 +745,44 @@ func (n SetNotifiesStage) Exec(ctx context.Context, l log.Logger, alerts ...*typ
...
+			}
+		}
+	}
+	// If the current time is inside a mute time, all alerts are removed from the pipeline
+	if muted {
+		lvl := level.Debug(l)
+		lvl.Log("msg", "Notifications not sent, route is within mute time")
```
roidelapluie, line 784: We can do those 2 lines into 1

notify/notify.go
```diff
@@ -729,3 +745,44 @@ func (n SetNotifiesStage) Exec(ctx context.Context, l log.Logger, alerts ...*typ
...
+	muteTimes map[string][]timeinterval.TimeInterval
+}
+
+func NewTimeMuteStage(mt map[string][]timeinterval.TimeInterval) *TimeMuteStage {
+	return &TimeMuteStage{mt}
+}
+
+// Exec implements the stage interface for TimeMuteStage
+// TimeMuteStage is responsible for muting alerts whose route is not in an active time
+func (mts TimeMuteStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.Alert) (context.Context, []*types.Alert, error) {
```
roidelapluie, line 759:
```suggestion
func (tms TimeMuteStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.Alert) (context.Context, []*types.Alert, error) {
```

notify/notify.go
```diff
@@ -279,10 +294,11 @@ func (pb *PipelineBuilder) New(
 	ms := NewGossipSettleStage(peer)
 	is := NewMuteStage(inhibitor)
 	ss := NewMuteStage(silencer)
+	mts := NewTimeMuteStage(muteTimes)
```
roidelapluie, line 297:
```suggestion
	tms := NewTimeMuteStage(muteTimes)
```

notify/notify.go
benridley: Thanks, good pickup. I'll fix this.

timeinterval/timeinterval.go
roidelapluie: Ok, then this is not tested

timeinterval/timeinterval.go
```diff
@@ -0,0 +1,473 @@
...
+		return min
+	}
+	if n >= max {
+		return max
+	}
+	return n
+}
+
+// ContainsTime returns true if the TimeInterval contains the given time, otherwise returns false
+func (tp TimeInterval) ContainsTime(t time.Time) bool {
```
roidelapluie, line 338: Do we check somewhere for empty time intervals? WIth an empty time interval, this would always return true

timeinterval/timeinterval.go
```diff
@@ -0,0 +1,473 @@
...
+	}
+	if tp.Months != nil {
+		in := false
+		for _, validMonths := range tp.Months {
+			if t.Month() >= time.Month(validMonths.Begin) && t.Month() <= time.Month(validMonths.End) {
+				in = true
+				break
+			}
+		}
+		if !in {
```
roidelapluie, line 390: We do not need this thanks to break, we do not need `in` at all

timeinterval/timeinterval.go
```diff
@@ -0,0 +1,473 @@
...
+			}
+			// Clamp to the boundaries of the month to prevent crossing into other months
+			begin = clamp(begin, -1*daysInMonth, daysInMonth)
+			end = clamp(end, -1*daysInMonth, daysInMonth)
+			if t.Day() >= begin && t.Day() <= end {
+				in = true
+				break
+			}
+		}
+		if !in {
```
roidelapluie, line 378: We do not need this thanks to break, we do not need in at all

timeinterval/timeinterval.go
```diff
@@ -0,0 +1,473 @@
...
+	}
+	if tp.Years != nil {
+		in := false
+		for _, validYears := range tp.Years {
+			if t.Year() >= validYears.Begin && t.Year() <= validYears.End {
+				in = true
+				break
+			}
+		}
+		if !in {
```
roidelapluie, line 414: We do not need this thanks to break, we do not need `in` at all

timeinterval/timeinterval.go
```diff
@@ -0,0 +1,473 @@
...
+			return false
+		}
+	}
+	return true
+}
+
+// Converts a string of the form "HH:MM" into the number of minutes elapsed in the day
```
roidelapluie, line 421: full stop

timeinterval/timeinterval.go
```diff
@@ -0,0 +1,473 @@
...
+		return 0, err
+	}
+	timeStampMinutes, err := strconv.Atoi(timestampComponents[1])
+	if err != nil {
+		return 0, err
+	}
+	if timeStampHours < 0 || timeStampHours > 24 || timeStampMinutes < 0 || timeStampMinutes > 60 {
+		return 0, fmt.Errorf("timestamp %s out of range", in)
+	}
+	// Timestamps are stored as minutes elapsed in the day, so multiply hours by 60
```
roidelapluie, line 441: full stop

timeinterval/timeinterval.go
```diff
@@ -0,0 +1,473 @@
...
+	mins = timeStampHours*60 + timeStampMinutes
+	return mins, nil
+}
+
+// Converts a range that can be represented as strings (e.g. monday:wednesday) into an equivalent integer-represented range
```
roidelapluie, line 446: full stop

timeinterval/timeinterval.go
```diff
@@ -0,0 +1,473 @@
...
+		}
+		r.setBegin(start)
+		r.setEnd(End)
+		return nil
+	}
+	val, err := r.memberFromString(in)
+	if err != nil {
+		return err
+	}
+	r.setBegin(val)
```
roidelapluie, line 470: Do we set them individualy somewhere? Can we do with one setter function ? Do we need those extra setters, can't we set EndDate directly?

timeinterval/timeinterval_test.go
```diff
@@ -0,0 +1,576 @@
...
+}
+
+var dayOfWeekStringTestCases = []struct {
+	dowString   string
+	ranges      []WeekdayRange
+	expectError bool
+}{
+	{
+		dowString:   "['monday:friday', 'saturday']",
+		ranges:      []WeekdayRange{{InclusiveRange{Begin: 1, End: 5}}, {InclusiveRange{Begin: 6, End: 6}}},
```
roidelapluie, line 173: There are no negative tests here:

`[5:1]`

`[-1]`

timeinterval/timeinterval.go
benridley: While this makes sense due to the semantics of the time interval (where not specifying a restriction means all possible times are included), I think an empty time interval is probably a symptom of an error. I'll add a check for this, thanks.

cmd/alertmanager/main.go
```diff
@@ -410,6 +411,12 @@ func run() int {
 			integrationsNum += len(integrations)
 		}
 
+		// Build the map of time interval names to mute time definitions
+		muteTimes := make(map[string][]timeinterval.TimeInterval)
+		for _, ti := range conf.MuteTimeIntervals {
+			muteTimes[ti.Name] = ti.TimeIntervals
```
roidelapluie, line 417: Should we check for duplicate names?

cmd/alertmanager/main.go
```diff
@@ -410,6 +411,12 @@ func run() int {
 			integrationsNum += len(integrations)
 		}
 
+		// Build the map of time interval names to mute time definitions
+		muteTimes := make(map[string][]timeinterval.TimeInterval)
```
roidelapluie, line 415: Can we pre-allocate the size?

cmd/alertmanager/main.go
```diff
@@ -410,6 +411,12 @@ func run() int {
 			integrationsNum += len(integrations)
 		}
 
+		// Build the map of time interval names to mute time definitions
```
roidelapluie, line 414: full stop

config/config.go
```diff
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/prometheus/alertmanager/timeinterval"
```
roidelapluie, line 28: Can we split this internal import from the others?

config/config.go
```diff
@@ -216,13 +217,32 @@ func resolveFilepaths(baseDir string, cfg *Config) {
 	}
 }
 
+// A MuteTimeInterval represents a named set of time intervals for which a route should be muted.
```
roidelapluie, line 220:
```suggestion
// MuteTimeInterval represents a named set of time intervals for which a route should be muted.
```

config/config.go
```diff
@@ -216,13 +217,32 @@ func resolveFilepaths(baseDir string, cfg *Config) {
...
+type MuteTimeInterval struct {
+	Name          string                      `yaml:"name" json:"name"`
+	TimeIntervals []timeinterval.TimeInterval `yaml:"time_intervals"`
+}
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface for MuteTimeInterval
```
roidelapluie, line 226: full stop

timeinterval/timeinterval.go
benridley: How do we avoid this? We need to loop over all the days/months/years etc to check if at least one of them matches, and return false if it doesn't.

config/config.go
roidelapluie, line 432: we should be consistent and pick a unique name

mute times
mute times interval ?

docs/configuration.md
```diff
@@ -113,6 +113,10 @@ receivers:
 # A list of inhibition rules.
 inhibit_rules:
   [ - <inhibit_rule> ... ]
+
+# A list of mute time intervals for muting routes
```
roidelapluie, line 117: full stop

docs/configuration.md
```diff
@@ -202,6 +212,65 @@ route:
...
+## `<mute_time_interval>`
+
+A `mute_time_interval` specifies a named interval of time that may be referenced
+in the routing tree to mute particular routes for particular times of the day.
+
+```yaml
+name: <string>
+time_intervals:
+  [ - <time_interval> ... ]
+
```
roidelapluie, line 224: extra white line not needed in the exemple

docs/configuration.md
```diff
@@ -202,6 +212,65 @@ route:
...
+  months:
+  [ - <month_range> ...]
+  years:
+  [ - <year_range> ...]
+```
+
+All these fields are optional and if left unspecified allow any value to match the interval.
+Some fields support ranges and negative indices, and are detailed below:
+
+`times`: A list of time-ranges. They are inclusive of the starting time and exclusive
```
roidelapluie, line 246:
```suggestion
`times_range`: Ranges inclusive of the starting time and exclusive
```

timeinterval/timeinterval.go
roidelapluie: We need the return false

But if we break we will not need the `if`

roidelapluie: sorry I did speak too fast :)

timeinterval/timeinterval.go
benridley: Thanks, I'll add tests for this.

timeinterval/timeinterval.go
benridley: No problem, I felt this was clunky as well initially but I couldn't see a better way :) Cheers

timeinterval/timeinterval.go
benridley: I've thought more about this, I think we should leave this as is. This makes sense with the semantics of a time interval and modifying it would complicate the function by possibly returning only a single error case.

config/config.go
```diff
@@ -216,13 +217,32 @@ func resolveFilepaths(baseDir string, cfg *Config) {
 	}
 }
 
+// MuteTimeInterval represents a named set of time intervals for which a route should be muted.
+type MuteTimeInterval struct {
+	Name          string                      `yaml:"name" json:"name"`
```
roidelapluie, line 222: I don't think we need json here

config/config.go
```diff
@@ -216,13 +217,32 @@ func resolveFilepaths(baseDir string, cfg *Config) {
...
+}
+
 // Config is the top-level configuration for Alertmanager's config files.
 type Config struct {
-	Global       *GlobalConfig  `yaml:"global,omitempty" json:"global,omitempty"`
-	Route        *Route         `yaml:"route,omitempty" json:"route,omitempty"`
-	InhibitRules []*InhibitRule `yaml:"inhibit_rules,omitempty" json:"inhibit_rules,omitempty"`
-	Receivers    []*Receiver    `yaml:"receivers,omitempty" json:"receivers,omitempty"`
-	Templates    []string       `yaml:"templates" json:"templates"`
+	Global            *GlobalConfig      `yaml:"global,omitempty" json:"global,omitempty"`
```
roidelapluie, line 240: @simonpasquier any idea why we have json here?

config/config.go
roidelapluie: This is not addressed. Also we use 'must not' above

config/config.go
```diff
@@ -430,12 +464,28 @@ func checkReceiver(r *Route, receivers map[string]struct{}) error {
...
+	}
+	return nil
+}
+
 // DefaultGlobalConfig returns GlobalConfig with default values.
 func DefaultGlobalConfig() GlobalConfig {
 	return GlobalConfig{
-		ResolveTimeout: model.Duration(5 * time.Minute),
-		HTTPConfig:     &commoncfg.HTTPClientConfig{},
-
```
roidelapluie, line 438: why removing this?

benridley:
@roidelapluie Thanks for the comments. I removed the json tags (except the one where you've asked for clarity with @simonpasquier). Also some of the changes (the GlobalConfig part) appeared to be an inadvertent formatting change, so I've reverted it.

roidelapluie:
> @roidelapluie Thanks for the comments. I removed the json tags (except the one where you've asked for clarity with @simonpasquier). Also some of the changes (the GlobalConfig part) appeared to be an inadvertent formatting change, so I've reverted it.

Now that I think of it the json tags might be required for the API.

benridley:
Yes I think you're right, but there's currently no mechanism that I can
find to manipulate or examine this config via the API which makes me think
it's not necessary for these structs

roidelapluie:
Hello @simonpasquier , can you have a look at this?

config/config.go
beorn7: Good question. I was wondering myself. I'm not even sure that all types in the config would nicely marshal to JSON.

beorn7:
@simonpasquier are you able to review this? If not, I'm happy to help out here with a second review, but it would be good to know if you are fine with this delegation.

beorn7:
Quick update: @simonpasquier is unlikely to find time for this any time soon. I'll try to find out about the necessity for the json tags, and give this a review, too.

beorn7:
So far, I can see that we use the JSON marshalling of config fields for the `status` endpoint of the v1 API.

While v1 is deprecated (AFAIK), it would be better to not break it, of course.

config/config.go
beorn7: Now they should. See #2463

beorn7:
So yeah, I'm afraid, we need `MarshalJSON` methods so that the whole config can be marshaled to JSON. I haven't found a reason why we need `UnmarshalJSON`, but I guess we should just keep parity here.

The marshaling is used in the `api/v1/status` endpoint. And currently, it looks like the following:
```json
      "mute_time_intervals": [
        {
          "Name": "foobar",
          "TimeIntervals": [
            {
              "Times": null,
              "Weekdays": [
                {
                  "Begin": 1,
                  "End": 3
                }
              ],
              "DaysOfMonth": null,
              "Months": null,
              "Years": null
            }
          ]
        }
      ]
```

The YAML looks much nicer:
```YAML
mute_time_intervals:
- name: foobar
  time_intervals:
  - weekdays: ['monday:wednesday']
```

@benridley do you think you can add the JSON (un)marshaling methods? #2463 is a PR where I did the same for the Matchers. (Same story here: we didn't understand why we need JSON marshaling and only added YAML marshaling initially.)

beorn7:
Also, it might be good to first resolve the merge conflicts…

benridley:
Cheers @beorn7, good spot with the v1 API. 

I've added support for JSON marshalling. In my travels I discovered the encoding.TextMarshaler interface, which actually seemed more appropriate for these structs. I moved most of the logic to that interface, and now JSON and YAML marshalling simply consume that. 

Similarly for JSON unmarshalling, I just delegate to the YAML unmarshaler as that's where I implemented all the validation logic and the YAML parser can parse JSON anyway.

Final result looks much better when hitting the `/api/v1/status` endpoint:
```yaml
      "mute_time_intervals": [
        {
          "Name": "test_interval",
          "TimeIntervals": [
            {
              "times": [
                {
                  "start_time": "09:00",
                  "end_time": "17:00"
                }
              ],
              "weekdays": [
                "monday:friday",
                "sunday"
              ],
              "days_of_month": [
                "-7:-1"
              ],
              "months": [
                "1:3"
              ],
              "years": [
                "2020:2025",
                "2030:2035"
              ]
            }
          ]
        }
      ]
```

beorn7:
Cool. I'll have a closer look ASAP (just very busy at the moment). @roidelapluie if you want to take a look in the meantime, that would be wonderful.

beorn7:
I'm still stuck with other things. Will only have time for a review next week. In the meantime, I welcome any and all other qualified reviewers (@roidelapluie and @simonpasquier are the obvious candidates, but this might also be a good exercise for @aSquare14  or @gotjosh :wave: :smile: ).

notify/notify_test.go
```diff
@@ -719,3 +721,86 @@ func TestMuteStageWithSilences(t *testing.T) {
 		t.Fatalf("Unmuting failed, expected: %v\ngot %v", in, got)
 	}
 }
+
+func TestTimeMuteStage(t *testing.T) {
+	// Route mutes alerts outside business hours
```
beorn7, line 726: Nit: End on period.

beorn7 (changes requested):
A few comments, more on the nitpicky level.

The code looks OK to me, but I don't know enough about the AM routing stages to give a final verdict. @roidelapluie you seem to have more context. Are you fine with code correctness? And since this is quite invasive, it would be also good to know if @simonpasquier agrees with this in principle.

docs/configuration.md
```diff
@@ -168,6 +172,12 @@ match_re:
 # been sent successfully for an alert. (Usually ~3h or more).
 [ repeat_interval: <duration> | default = 4h ]
 
+# Times when the route should be muted. These must match the name of a
+# mute time interval defined in the mute_time_intervals section. 
+# Additionally, the root node cannot have any mute times.
```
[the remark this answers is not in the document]
beorn7: I'd say the alert still matches the route, but doesn't fire, but my understanding of the code is limited.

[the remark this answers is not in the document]
beorn7: Whatever is the truth, it should be clearly explained here.

gotjosh:
I completely missed the original notification from you @beorn7 - I'll try and take a look over this week.

benridley:
@roidelapluie @beorn7  @simonpasquier 
Apologies for the delay on the latest changes requested. I've been thinking more about the architecture and how it might be improved and would be keen to get your thoughts. Currently the `group_interval` determines how frequently aggregation groups will wake up and send notifications (flush). This has an effect on mute times, as it determines how 'responsive' the alerts are when transitioning between mute time intervals. If a group becomes unmuted between flushes, nothing will happen until the next flush.

I'm thinking rather than overload `group_interval` with this usage, perhaps we move the mute time stuff a bit higher up in the pipeline and have groups inside mute times wake up more frequently to check for changes. This would make things more responsive and avoid complicating the various intervals and wait settings that already exist.

beorn7:
@benridley sounds all very reasonable. I'm not an export on the notification pipeline myself. I would think the time-based muting should follow the same semantics as silencing and inhibition (which, I presume, will create a notification immediately if it was previously completely muted and then the silence expires or a inhibition condition doesn't apply anymore).

roidelapluie:
I do not think that this is correct. The current logic should cover this. I will double check that.

benridley:
> I do not think that this is correct. The current logic should cover this. I will double check that.

Thanks Julien. I tested with the mute times, and it appears to be delayed by the group_interval as mentioned but perhaps there's a mechanism silences and inhibitions use I'm not leveraging.

roidelapluie:
> > I do not think that this is correct. The current logic should cover this. I will double check that.
> 
> Thanks @roidelapluie. I tested with the mute times, and it appears to be delayed by the group_interval as mentioned but perhaps there's a mechanism silences and inhibitions use I'm not leveraging.

My experience with inhibition, at least, shows that as soon as the initial alert is gone, it is firing.

beorn7:
> My experience with inhibition, at least, shows that as soon as the initial alert is gone, it is firing.

Yes, and as soon as a silence expires, the still firing alerts will immediately create new notifications.

However, what @benridley is saying (IIUC) is that with the implementation in this PR, if a muted time interval is over, the muted alerts might only get active again by up to one full group_interval.

I believe we all agree that the semantics for time-based muting should be the same as muting from inhibition or silences. We just need to figure out how to accomplish that.

benridley:
@beorn7  @roidelapluie I did a test with inhibition, and it works the same way. The alert will always be 'firing', but if an inhibition expires between `group_interval`s, notifications won't be sent until the next flush. 

The fact that all of us were somewhat unaware of this has convinced me it's probably not as much of a problem as I thought, so I think the way forward is to just make this clear in the docs. What do you think?

beorn7:
Makes sense, @benridley . Still, the fact that an alert that becomes un-inhibited might need up to a full group_interval to notify again is a bit worrying. It's not uncommon to have long gorup_intervals. @simonpasquier what's your context here? Would that be worth filing an issue? Once I find time, I'll try and investigate this a bit more.

beorn7:
I wasn't wowed, so this is a weak hire, but at least confidently so.

Julien was fairly fluid doing the napkin math. This was also the first interview in which the candidate put a price tag on the network traffic, which I guess is a pretty nice trait, especially for the Cloud Platform team. He readily realized the "shape" of the data and used the metric ID as some kind of key, realizing that this shaves off 85% of the data size in the storage.  It was also easy to trigger him making the tradeoff when compressing data with gzip etc. Another highlight was that he instantly realized that deleting the data after a given retention time is not trivial.

To accomplish the above, he suggested something like a linked list (to easily "front-truncate"). He needed quite a bit of guidance to realize that the management overhead would be larger than the data size. And he never realized that a linked list is more a structure to be used in RAM rather than in data as laid out on disk or such. At the time we discussed this topic, we were almost out of time, so I couldn't guide him to other ideas and essentially had to tell him to use something like hourly blocks to proceed. He apparently had no prior knowledge about varint or delta encodings, but he could come up with a few things on the spot, like recording the offset to the starting value rather than the whole value, even realized that there is a problem with the number of bytes to use for those as they are not guaranteed to never exceed a certain value (but he didn't make it to "inventing" varint or similar). With a moderate amount of guidance, he also figured out that the timestamp can go away completely if they are completely regular and we know the beginning timestamp of a block (connecting to the coding exercise).

Concerns: He assumed the data stream on the wire would be continuous but never discussed staggered sending of metrics. He seems to have a slight tendency to pull out existing services and consider them as magic black boxes, e.g. he proposed to enqueue the incoming samples first in a queueing system to increase the resilience in case our ingestion server goes down (ignoring that this just shifts the ingestion problem to the queueing system, which is probably much worse at it than our dedicated ingestion server).

Overall, I still think it's an acceptable performance. In particular, Julien's profile of stronger and weaker points might be a good match to the cloud platform team.

beorn7:
(In case you got a weird email notification here, that was a copy&paste mishap. The unintended comment is deleted now.)

beorn7:
In different news, if you rebase this on top of master (or merge master into your branch), you should get rid of the failing test_frontend.

I believe the normal test failure is just due to flakiness. I'm currently re-running the test.

beorn7:
I'm still planning to play with this and explore the muting behavior in more detail as soon as I find time.

If in the meantime someone that actually knows about those things could chime in, ever better…

benridley:
Thanks @beorn7. I'll also do some more investigation... And rebase the branch soon!

beorn7:
I played with silencing and inhibition timings, and (sadly) I have to confirm that whatever combination I tried, an alert becoming unsilenced or uninhibited will only ever trigger a notification after the `group_interval` has passed normally for the notification group it is in. There is no difference between inhibitions and silencing.

I understand that for a group that actually had already sent out a notification, e.g. following timing:
- t=0: alert A fires and notifies
- t=1: silence is put into place silencing A
- t=2: alert B fires, which is in the same group as A and also silenced by the silence
- t=3: silence expires
- t=0+group_interval: alert notifies again
In this situation, the group _has_ notified, and the fact that it was silenced in between is arguably not a reason to notify before the group interval has passed.

However, if the alert comes in _after_ the silence has put into place, I'd argue the behavior is rather problematic:
- t=0: silence is put into place
- t=1: alert A fires but does not notify because it is silenced by the silence
- t=2: silence expires
- t=1+group_interval: alert notifies for the first time
In other words, with a slightly unlucky timing of silences or inhibitions, it's easily possible that you have to wait by up to group_interval to get a notification for a certain group _for the first time!_

This is not that dramatic with the default group_interval of 5m, but for sub-critical alerts, a much longer group_interval is quite common, and this might surprise people quite a bit. Intuitively, I'd say, muted alerts (via silencing or inhibition or via the new time-based muting in this PR) should be treated as non-firing alerts. And with this thought, this totally blends into the ancient issue #226 about sending resolved notifications for silenced alerts. As you can see by following the discussion over there, this is definitely not-as-easy-as-it-looks. I guess we need to discuss this separately and ideally with those that have the historical context why things work as they do in AM.

Coming back to this PR: Obviously, we won't change the behavior here in this PR, and the time-based muting should follow the same semantics as any other muting. So I'd say we are good to go with this PR, let's just wait for @benridley to rebase to get the final blessing.

benridley:
Thanks for validating that @beorn7. I've rebased and the tests are looking much happier. 

Do we want to add some additional words around group_interval in the docs to indicate this behavior? The current documentation is:
```
# How long to wait before sending a notification about new alerts that
# are added to a group of alerts for which an initial notification has
# already been sent. (Usually ~5m or more.)
```
which doesn't give much indication of the issue we've discovered. Adding some words here might stop people from being surprised by delays in notification delivery.

beorn7:
@benridley I totally agree that the current documentation is incomplete. However, let's first get clarity if the current behavior is actually the intended behavior or a bug (in which case we would document it very differently). I'll file an issue about this topic.

In the meantime, we can merge this PR.

beorn7 (approved):

docs/configuration.md
```diff
@@ -202,6 +215,67 @@ route:
...
+        times:
+        - start_time: HH:MM
+          end_time: HH:MM
+
+`weeekday_range`: A list of days of the week, where the week begins on Sunday and ends on Saturday.
+Days should be specified by name (e.g. ‘Sunday’). For convenience, ranges are also accepted
+of the form <start_day>:<end_day> and are inclusive on both ends. For example:
+`[‘monday:wednesday','saturday', 'sunday']`
+
+`days_of_month_ramge`: A list of numerical days in the month. Days begin at 1.
```
machine424, line 265: `days_of_month_range`?

machine424: Do you want me to fix it?

machine424: here https://github.com/prometheus/alertmanager/pull/2497

lopesit0:
Can someone explain to me how to integrate this functionality?

benridley:
Hey @lopesit0, no problem!
This has been merged into master now, but isn't part of an official release yet, so if you download Alertmanager using traditional methods (e.g. downloading them from https://prometheus.io/download/ or DockerHub), this functionality won't be there. If you want to try it out, you can build the master branch yourself which is very easy thanks to the Golang toolchain. Simply download Go and then clone this repository from GitHub. Once you've done that, on the command line simply run
```bash
go build -o alertmanager cmd/alertmanager/main.go
```
which will give you an alertmanager binary with this included. 

You can find the raw documentation also in the [master branch](https://github.com/prometheus/alertmanager/blob/master/docs/configuration.md) which explains how to configure your time intervals. 

Just be warned this is new and potentially unstable. If you do decide to run it, please feel free to report back here or open issues against the repository so we can fix then ASAP!

lopesit0:
> Hey @lopesit0, no problem!
> This has been merged into master now, but isn't part of an official release yet, so if you download Alertmanager using traditional methods (e.g. downloading them from https://prometheus.io/download/), this functionality won't be there. If you want to try it out, you can build the master branch yourself which is very easy thanks to the Golang toolchain. Simply download Go and then clone this repository from GitHub. Once you've done that, on the command line simply run
> 
> ```shell
> go build -o alertmanager cmd/alertmanager/main.go
> ```
> 
> which will give you an alertmanager binary with this included.
> 
> You can find the raw documentation also in the [master branch](https://github.com/prometheus/alertmanager/blob/master/docs/configuration.md) which explains how to configure your time intervals.
> 
> Just be warned this is new and potentially unstable. If you do decide to run it, please feel free to report back here or open issues against the repository so we can fix then ASAP!
> 
> EDIT: Alternatively, I've realized there's a Docker image built from master. So the easiest way to try this out might be using the official docker image built from master.
> `docker pull prom/alertmanager:master`

Thanks for your answer :)

ijungmann:
Is there any way to make use of this in the helm chart at the moment?  It seems like the operator does not handle this yet.

beorn7:
We need a release of Alertmanager. I think @roidelapluie was working on it, but there were flaky tests blocking it. Perhaps those are resolved now?

roidelapluie:
See https://github.com/prometheus/alertmanager/pull/2557

bimmerkiev:
Hi all.
I'm trying to use that feature with release version 0.22.0-rc.1, but I have no luck with it.
My alertmanager.yml:
```
mute_time_intervals:
  - name: business_hours
    time_intervals:
      - weekdays: ['monday:friday']
        times:
        - start_time: "09:00"
          end_time: "11:00"

templates:
- '/etc/alertmanager/templates/*.tmpl'
receivers:
- name: pagerduty
  pagerduty_configs:
  - client_url: http://x.x.x.x/
    description: '{{ if .CommonAnnotations.summary }}{{ .CommonAnnotations.summary
      }}{{ end }}'
    routing_key: xxx
    severity: '{{ if .CommonLabels.severity }}{{ .CommonLabels.severity | toLower
      }}{{ end }}'

route:
  group_by:
  - alertname
  - cluster
  - service
  - env
  group_interval: 5m
  group_wait: 30s
  receiver: pagerduty
  repeat_interval: 4h
  routes:
  - group_wait: 10s
    match:
      severity: low
      alertname: InstanceDowntest
      mute_time_intervals:
        - business_hours
      receiver: pagerduty
```

But  with that config alertmanager don't want to start.
Part of log:
```
level=error ts=2021-05-14T13:38:06.667Z caller=coordinator.go:118 component=configuration msg="Loading configuration file failed" file=/etc/alertmanager/alertmanager.yml err="yaml: unmarshal errors:\n  line 44: cannot unmarshal !!seq into string"
```
Can anyone help me?
Thanks

bimmerkiev:
the error about line:
" - business_hours"

brunocriado:
Hi 👋 
Before sending an issue I wanted to confirm if I'm doing it right. I have a test configuration where I want to mute the alerts for for 1 minute on Mondays for a specific receiver as you can see in below.

```
...
  - group_by:
    - alertname
    - severity
    - env
    match:
      type: aws-alarm
      mute_times: test
    receiver: aws-alarm
    continue: true
...
mute_time_intervals:
- name: test
  time_intervals:
  - weekdays: ['monday']
    times:
    - start_time: '09:00'
      end_time: '09:01'
```
I'm testing with amtool and when I send the alert I didn't receive it. If I remove that time interval it go off as expected. BTW today is not Monday and I'm testing for more than 1 minute.

brunocriado:
Just found the issue. I should use `mute_time_intervals` and not inside match
```
  - group_by:
    - alertname
    - severity
    - env
    match:
      type: cw-sns-cw-alarm
    mute_time_intervals:
      - Monday
    receiver: cw-sns-cw-alarm-alerts-dev
```

rschirin:
sorry team, is there any chance to see this feature within the docker image?

beorn7:
It has been released in v0.22. Any Docker image with that or a later release should support it.

PR #2402, opened 2020-10-23
notify: update notify/wechat

BinacsLee:
Accoding to [WeChat Work API](https://work.weixin.qq.com/api/doc/90000/90135/91039), we will get response such as:

```json
{
   "errcode": 0,
   "errmsg": "ok",
   "access_token": "accesstoken000001",
   "expires_in": 7200
}
```

instead of:

```json
{
   "code": 0,
   "error": "ok",
   "access_token": "accesstoken000001",
}
```

And sometimes the request fails because it uses an expired token

**So I did these things to update the code:**

1. Update `weChatResponse` and `tokenResponse`;
2. Use `expireIn` instead of `2 * time.Hour`;
3. Record the time before we send the request, not after we receive the response;

**For test, I wrote the following code:**
```go
func TestNotify(t *testing.T) {
        // Use `*` to protect information
	secret := "d2mS-c3*********************************dwKLMbDI"
	u, err := url.Parse("https://qyapi.weixin.qq.com/cgi-bin/")
	require.NoError(t, err)
	n, err := New(
		&config.WechatConfig{
			APIURL:      &config.URL{URL: u},
			HTTPConfig:  &commoncfg.HTTPClientConfig{},
			ToUser:      "binacs",
			AgentID:     "*********",
			CorpID:      "******************",
			MessageType: "text",
			Message:     "binacs: send from prometheus/alertmanager unit test",
			APISecret:   config.Secret(secret),
		},
		test.CreateTmpl(t),
		log.NewJSONLogger(os.Stdout),
	)
	require.NoError(t, err)
	ok, err := n.Notify(context.Background(), []*types.Alert{
		&types.Alert{
			Alert: model.Alert{
				Labels: model.LabelSet{
					"lbl1": "val1",
				},
				StartsAt: time.Now(),
				EndsAt:   time.Now().Add(time.Hour),
			},
		},
	}...)
	t.Log(ok, err)
}
```
And it works:
notice
message

**Thanks for your review.**


notify/wechat/wechat.go
```diff
@@ -94,9 +97,8 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 	if err != nil {
 		return false, err
 	}
-
-	// Refresh AccessToken over 2 hours
-	if n.accessToken == "" || time.Since(n.accessTokenAt) > 2*time.Hour {
+	// Refresh AccessToken over `expireIn`
+	if n.accessToken == "" || time.Since(n.accessTokenAt) > n.expireIn {
```
simonpasquier, line 101: It may be safer to renew the token before it expires (for instance when it reached 90% of its validity).

notify/wechat/wechat.go
```diff
@@ -132,7 +136,8 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 
 		// Cache accessToken
 		n.accessToken = wechatToken.AccessToken
-		n.accessTokenAt = time.Now()
+		n.accessTokenAt = startTime
+		n.expireIn = time.Duration(wechatToken.ExpiresIn) * time.Second
```
simonpasquier, line 140: instead of storing the creation time of the token and its period, you could record its expiration time (minus a safety interval as described above).

simonpasquier (commented):
Thanks, just a few remarks! Could you add tests too please?

BinacsLee:
Thank you very much for your advice!

I have submitted the latest code. Could you please review it again? @simonpasquier

And the unit test works like this:

1C972E920C46F5EAAF451D1219A4B6D9

notify/wechat/wechat.go
```diff
@@ -34,30 +34,32 @@ import (
...
 
-	accessToken   string
-	accessTokenAt time.Time
-}
-
-// token is the AccessToken with corpid and corpsecret.
-type token struct {
-	AccessToken string `json:"access_token"`
+	accessToken string
+	expireAt    time.Time
```
simonpasquier, line 52:
```suggestion
	tokenRenewAt    time.Time
```

notify/wechat/wechat_test.go
```diff
@@ -26,6 +28,35 @@ import (
 	"github.com/prometheus/alertmanager/notify/test"
 )
 
+// TestNotify can test the entire notification process as long as you fill in
+// the correct configuration: secret and WechatConfig.
+// You can get a full description of the WechatConfig's fields on
+// https://work.weixin.qq.com/api/doc/90001/90143/91199
+func TestNotify(t *testing.T) {
```
simonpasquier, line 35: The unit test can't access the real WeChat endpoint. We need to mock the WeChat API to simulate the various responses. Since it isn't trivial, I can have a look at it. Also this code doesn't test anything.

SoloJacobs:
Hi @binacs ,

it appears to me that we have not heard back from you after the latest review. In the meantime, a fix has been submitted in https://github.com/prometheus/alertmanager/pull/3330 .

Hopefully, this addresses the issue. If not, I'm happy to review a PR, which is based of the state of `main`. Thank you for your contribution.

Kind regards

PR #2403, opened 2020-10-26
Make filter labels consistent with Prometheus

### LINKED ISSUE #2342: Missing and empty label queries act differently to Prometheus
## What did you do?
Query alertmanager for alerts with the label `foo=~"bar|"` or `foo=~".*"`

## What did you expect to see?
I expected the alerts with the label foo set to bar and the alerts with foo unset to be returned. This is the behaviour when querying the `ALERTS` metric through prometheus.

## What did you see instead? Under which circumstances?
The alerts with the foo label unset weren't returned for any of the regex matchers (although they were returned for a query like `foo=""`)

This behaviour makes providing dropdown filters for alert labels in Grafana impossible as far as I can tell, as it will always filter out the alerts which don't have that label set.

## Environment

* System information: `Linux 4.4.225-1.el7.elrepo.x86_64 x86_64`
* Alertmanager version: `0.16.1`
* Prometheus version: `2.14.0`
```
global:
      resolve_timeout: 5m

    receivers:
    - name: 'default-receiver'
    - name: 'SNMP-Alert-Sender'
      webhook_configs:
      - send_resolved: true
        url: 'http://localhost:25443/v1/target/www.xxx.yyy.zzz/162'
      - send_resolved: true
        url: 'http://localhost:25443/v1/target/www.xxx.yyy.zzz/162'
      - send_resolved: true
        url: 'http://localhost:25443/v1/target/www.xxx.yyy.zzz/162'

    route:
      receiver: 'default-receiver'
      group_wait: 30s
      group_interval: 1m
      group_by: ['...']
      repeat_interval: 1h
      routes:
      - receiver: 'SNMP-Alert-Sender'
        repeat_interval: 1h
        continue: true
        group_by: ['...']
```

vears91:
Filtering the alert out when the label is missing precludes a
possible match for an empty value. This change allows the
match to be evaluated.

Closes #2342

Signed-off-by: Victor Araujo <vear91@gmail.com>

### DIFF
--- a/api/v1/api.go
+++ b/api/v1/api.go
@@ -662,7 +662,7 @@ func matchFilterLabels(matchers []*labels.Matcher, sms map[string]string) bool {
 			if string(m.Value) == "" && !prs {
 				continue
 			}
-			if !prs || !m.Matches(string(v)) {
+			if !m.Matches(string(v)) {
 				return false
 			}
 		}
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -551,7 +551,7 @@ func matchFilterLabels(matchers []*labels.Matcher, sms map[string]string) bool {
 			if m.Value == "" && !prs {
 				continue
 			}
-			if !prs || !m.Matches(v) {
+			if !m.Matches(v) {
 				return false
 			}
 		}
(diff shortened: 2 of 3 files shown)


simonpasquier (commented):
Thanks! Could you add tests for this? At least for `api/v2`, `api/v1` being deprecated I'd be ok without it.

vears91:
Added tests :+1:

simonpasquier (approved):
Thanks!

PR #2414, opened 2020-11-04
Add amtool command for rendering templates.

FUSAKLA:
Hi, I found quite difficult for users to write the Alertmanager templates.
Mainly it's quite difficult to test them. 

Now user has to:
1. run the alertmanager with the new configuration
1. send the test aler probably using amtool
1. have some receiver configured where he can check the result

I do not think this is much user-friendly and I hope we can do better.

This PR is just draft, because the changes required for it are quite big, so I'm not polishing it and finalizing it but just to show the idea:

The command would generate specified number of synthetic alerts which could be somehow parametrized (add custom annotation, label etc). These will be used for rendering of the templates. 

It's bit complicated to specify which templates should be rendered, by default it could render al of them possibly user could filter just specified type, id or name of receiver to render.

The rendered output as should be mainly human readable IMHO (but also reproducible in case someone would want to use it for testing for example.) That's why as initial implementation I did the templating as rendering the receiver config as yaml. That unfortunately required bigger refactoring of the receivers configs templating. 

Example tool help
```bash
$ ./amtool config templates render --config.file examples/ha/alertmanager.yml --help
usage: amtool config templates render

Flags:
  -h, --help                     Show context-sensitive help (also try --help-long and --help-man).
      --config.file=CONFIG.FILE  Config file to be used.
      --receiver.name=RECEIVER.NAME  
                                 Name of receiver to be used for templating.
      --receiver.type=RECEIVER.TYPE  
                                 Type of receiver to be used for templating.
      --receiver.id=-1           Id of receiver to be used for templating.
      --additional.labels="label-name=label-value"  
                                 Labels of the alert used for rendering of the templates.
      --additional.annotations="annotation-name=annotation-value"  
                                 Annotations of the alert used for rendering of the templates.
      --alert.count="annotation-name=annotation-value"  
                                 Number of alerts to be passed to the rendering.
```

Example usage output:
```bash
$ ./amtool config templates render --config.file examples/ha/alertmanager.yml 
--- # receiver:  type: opsgenie index: 1

send_resolved: true
http_config: {}
api_key: <secret>
api_url: https://api.opsgenie.com/
message: '[FIRING:1]  (bar)'
description: |+
  bar
  Alerts Firing:
  Labels:
   - foo = bar
  Annotations:
   - foo = bar
  Source: http://foo.bar
```

I'll be happy to finish this but I'm not 100% sure about the ideal form of output and input.
Would you be willing to accept such thing and if so, how should it look like ideally?

Thanks for any ideas!

### DIFF
--- a/cli/config.go
+++ b/cli/config.go
@@ -35,6 +35,7 @@ func configureConfigCmd(app *kingpin.Application) {
 	configCmd := app.Command("config", configHelp)
 	configCmd.Command("show", configHelp).Default().Action(execWithTimeout(queryConfig)).PreAction(requireAlertManagerURL)
 	configureRoutingCmd(configCmd)
+	configureTemplatingCmd(configCmd)
 }
 
 func queryConfig(ctx context.Context, _ *kingpin.ParseContext) error {
--- a/cli/templating.go
+++ b/cli/templating.go
@@ -1,0 +1,129 @@
+// Copyright 2018 Prometheus Team
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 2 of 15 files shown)


FUSAKLA:
@simonpasquier if you find a while would you give it a look? :pray:

GiedriusS:
Attempting to implement a more generic variant here: https://github.com/prometheus/alertmanager/pull/2538

roidelapluie:
#2538 has been merged. Thanks!

PR #2422, opened 2020-11-12
fix: do not validate silence if expiring it

### LINKED ISSUE #2415: Can't expire silence with empty matcher present from versions before the check was there.
We just bumped into deadlock caused by silence with empty matchers.

Version `0.20.0` added the check that `Check that at least one silence matcher matches a non-empty string` in #2081

We had such silence present already and upgraded our cluster.
After quite long time we wanted to delete this silence, but we couldn't because the check prevented us event from expiring or updating it. Resulting to 400 on update and 500 on expire.

We had to patch the current source code to disable the check, delete the silence and than remove the patch to resolve the issue.

I know the change was marked in the CHANGELOG as `Changed`, but there was no warning to drop those before the upgrade event though it might be obvious. 

I'd say this check maybe should not happen when expiring the silence? 
That would be probably valid fix and would avoid the 500 error.

FUSAKLA:
fixes #2415 , see it's description for details

Trivial fix, not sure if this is the right way to do it but it was the most straight forward :) 

Tested with build without validation to create silence and than using the generated `silences` file run the current master and fixed version.

Master returns `500 internal server error` with log
```
level=error ts=2020-11-12T22:41:17.199Z caller=api.go:678 component=api version=v2 path=/silence/e78ca547-5d3d-4385-b93a-e2d4184203fc method=DELETE msg="Failed to expire silence" err="silence invalid: at least one matcher must not match the empty string"
```

This version succeeds with the expiration.

The silences file for testing is attached here

[silences.zip](https://github.com/prometheus/alertmanager/files/5533735/silences.zip)

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -492,11 +492,13 @@ func (s *Silences) getSilence(id string) (*pb.Silence, bool) {
 	return msil.Silence, true
 }
 
-func (s *Silences) setSilence(sil *pb.Silence, now time.Time) error {
+func (s *Silences) setSilence(sil *pb.Silence, now time.Time, validate bool) error {
 	sil.UpdatedAt = now
 
-	if err := validateSilence(sil); err != nil {
-		return errors.Wrap(err, "silence invalid")
+	if validate {
+		if err := validateSilence(sil); err != nil {
+			return errors.Wrap(err, "silence invalid")
+		}
 	}
 
 	msil := &pb.MeshSilence{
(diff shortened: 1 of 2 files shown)


FUSAKLA:
@simonpasquier PTAL

roidelapluie:
I think we should instead document in the release notes that such bad silences must be expired before upgrading.

FUSAKLA:
Yes we could definitely do that. On the other hand the change is minimal and validation of the silence while expiring it does not feel right nonetheless and returning 500 also does not feel right for any server as a known behavior.

PR #2430, opened 2020-12-06
check alerts length

bigHave:
check alerts length in  FanoutStage.Exec , which is done in MultiStage.Exec.
```
if len(alerts) == 0
```
No need to create goroutines for each stages.


bigHave:
@simonpasquier PTAL

Issue #1023, opened 2017-10-04
Negative matchers for routing and inhibition


beorn7:
Currently, we use matchers at two different places in the config:

1. For the routing tree:
```
# A set of equality matchers an alert has to fulfill to match the node.
match:
  [ <labelname>: <labelvalue>, ... ]

# A set of regex-matchers an alert has to fulfill to match the node.
match_re:
  [ <labelname>: <regex>, ... ]
```
2. For inhibition rules:
```
# Matchers that have to be fulfilled in the alerts to be muted.
target_match:
  [ <labelname>: <labelvalue>, ... ]
target_match_re:
  [ <labelname>: <regex>, ... ]

# Matchers for which one or more alerts have to exist for the
# inhibition to take effect.
source_match:
  [ <labelname>: <labelvalue>, ... ]
source_match_re:
  [ <labelname>: <regex>, ... ]
```

Negative matching comes in handy in various situations. Since the RE2 library doesn't support straight-forward negative matching (e.g. `(?!re)`) as part of the regexp syntax, I see two ways of implementing them:

1. Add a `not_match` clause for each case. This is straight forward but increases the number of config options:
```
# Routing:
match:
  [ <labelname>: <labelvalue>, ... ]
not_match:
  [ <labelname>: <labelvalue>, ... ]
match_re:
  [ <labelname>: <regex>, ... ]
not_match_re:
  [ <labelname>: <regex>, ... ]

# Inhibition:
target_match:
  [ <labelname>: <labelvalue>, ... ]
target_not_match:
  [ <labelname>: <labelvalue>, ... ]
target_match_re:
  [ <labelname>: <regex>, ... ]
target_not_match_re:
  [ <labelname>: <regex>, ... ]
source_match:
  [ <labelname>: <labelvalue>, ... ]
source_not_match:
  [ <labelname>: <labelvalue>, ... ]
source_match_re:
  [ <labelname>: <regex>, ... ]
source_not_match_re:
  [ <labelname>: <regex>, ... ]
```
2. A possibly more elegant solution, which would even _reduce_ the number of config options, is to use the Promethean style matchers like we already do in the Alertmanager UI. The new config options could be used side-by-side with the old ones for a while for a smooth transition:
```
# Routing:
matchers:
  [ - <string>, ...  ]

# Inhibition:
target_matchers:
  [ - <string>, ...  ]
source_matchers:
  [ - <string>, ...  ]
```

[...]

PR #2434, opened 2020-12-14
Add negative matchers for routing.

aSquare14:
Signed-off-by: aSquare14 <atibhi.a@gmail.com>
Related issue : #1023 
Note : I am adding negative matchers for routing in this PR. Will add support for inhibition rules in next PR.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -21,6 +21,7 @@ import (
 	"net/url"
 	"path/filepath"
 	"regexp"
+	"sort"
 	"strings"
 	"time"
 
@@ -28,6 +29,8 @@ import (
 	commoncfg "github.com/prometheus/common/config"
 	"github.com/prometheus/common/model"
 	"gopkg.in/yaml.v2"
+
+	"github.com/prometheus/alertmanager/pkg/labels"
 )
 
 const secretToken = "<secret>"
@@ -576,11 +579,13 @@ type Route struct {
 	GroupByStr []string          `yaml:"group_by,omitempty" json:"group_by,omitempty"`
 	GroupBy    []model.LabelName `yaml:"-" json:"-"`
 	GroupByAll bool              `yaml:"-" json:"-"`
-
-	Match    map[string]string `yaml:"match,omitempty" json:"match,omitempty"`
-	MatchRE  MatchRegexps      `yaml:"match_re,omitempty" json:"match_re,omitempty"`
-	Continue bool              `yaml:"continue" json:"continue,omitempty"`
-	Routes   []*Route          `yaml:"routes,omitempty" json:"routes,omitempty"`
+	// Deprecated. Remove before v1.0 release.
+	Match map[string]string `yaml:"match,omitempty" json:"match,omitempty"`
+	// Deprecated. Remove before v1.0 release.
+	MatchRE  MatchRegexps `yaml:"match_re,omitempty" json:"match_re,omitempty"`
+	Matchers Matchers     `yaml:"matchers,omitempty" json:"matchers,omitempty"`
+	Continue bool         `yaml:"continue" json:"continue,omitempty"`
+	Routes   []*Route     `yaml:"routes,omitempty" json:"routes,omitempty"`
 
 	GroupWait      *model.Duration `yaml:"group_wait,omitempty" json:"group_wait,omitempty"`
 	GroupInterval  *model.Duration `yaml:"group_interval,omitempty" json:"group_interval,omitempty"`
@@ -775,4 +780,34 @@ func (re Regexp) MarshalJSON() ([]byte, error) {
 		return json.Marshal(re.original)
 	}
 	return nil, nil
+}
+
+// Matchers is label.Matchers with an added UnmarshalYAML method to implement the yaml.Unmarshaler interface
+// and MarshalYAML to implement the yaml.Marshaler interface.
+type Matchers labels.Matchers
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface for Matchers.
+func (m *Matchers) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var lines []string
+	if err := unmarshal(&lines); err != nil {
+		return err
+	}
+	for _, line := range lines {
+		pm, err := labels.ParseMatchers(line)
+		if err != nil {
+			return err
+		}
+		*m = append(*m, pm...)
+	}
+	sort.Sort(labels.Matchers(*m))
+	return nil
+}
+
+// MarshalYAML implements the yaml.Marshaler interface for Matchers.
+func (m Matchers) MarshalYAML() (interface{}, error) {
+	result := make([]string, len(m))
+	for i, matcher := range m {
+		result[i] = matcher.String()
+	}
+	return result, nil
 }
--- a/dispatch/route.go
+++ b/dispatch/route.go
@@ -23,7 +23,7 @@ import (
 	"github.com/prometheus/common/model"
 
 	"github.com/prometheus/alertmanager/config"
-	"github.com/prometheus/alertmanager/types"
+	"github.com/prometheus/alertmanager/pkg/labels"
 )
 
 // DefaultRouteOpts are the defaulting routing options which apply
@@ -43,9 +43,9 @@ type Route struct {
 	// The configuration parameters for matches of this route.
 	RouteOpts RouteOpts
 
-	// Equality or regex matchers an alert has to fulfill to match
+	// Matchers an alert has to fulfill to match
 	// this route.
-	Matchers types.Matchers
+	Matchers labels.Matchers
 
 	// If true, an alert matches further routes on the same level.
 	Continue bool
@@ -88,14 +88,31 @@ func NewRoute(cr *config.Route, parent *Route) *Route {
 	}
 
 	// Build matchers.
-	var matchers types.Matchers
+	var matchers labels.Matchers
 
+	// cr.Match will be deprecated. This for loop appends matchers.
 	for ln, lv := range cr.Match {
-		matchers = append(matchers, types.NewMatcher(model.LabelName(ln), lv))
+		matcher, err := labels.NewMatcher(labels.MatchEqual, ln, lv)
+		if err != nil {
+			// This error must not happen because the config already validates the yaml.
+			panic(err)
+		}
+		matchers = append(matchers, matcher)
 	}
+
+	// cr.MatchRE will be deprecated. This for loop appends regex matchers.
 	for ln, lv := range cr.MatchRE {
-		matchers = append(matchers, types.NewRegexMatcher(model.LabelName(ln), lv.Regexp))
+		matcher, err := labels.NewMatcher(labels.MatchRegexp, ln, lv.String())
+		if err != nil {
+			// This error must not happen because the config already validates the yaml.
+			panic(err)
+		}
+		matchers = append(matchers, matcher)
 	}
+
+	// We append the new-style matchers. This can be simplified once the deprecated matcher syntax is removed.
+	matchers = append(matchers, cr.Matchers...)
+
 	sort.Sort(matchers)
 
 	route := &Route{
@@ -122,7 +139,7 @@ func NewRoutes(croutes []*config.Route, parent *Route) []*Route {
 // Match does a depth-first left-to-right search through the route tree
 // and returns the matching routing nodes.
 func (r *Route) Match(lset model.LabelSet) []*Route {
-	if !r.Matchers.Match(lset) {
+	if !r.Matchers.Matches(lset) {
 		return nil
 	}
 
--- a/dispatch/route_test.go
+++ b/dispatch/route_test.go
@@ -60,7 +60,6 @@ routes:
     repeat_interval: 1h
     group_by: ['job']
 
-
 - match_re:
     owner: 'team-(B|C)'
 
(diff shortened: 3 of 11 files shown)


types/match.go
```diff
@@ -103,6 +120,24 @@ func NewRegexMatcher(name model.LabelName, re *regexp.Regexp) *Matcher {
...
+	m := &Matcher{
+		Type:            t,
+		Name:            string(n),
+		Value:           v,
+		NewRouteMatcher: true,
+	}
+	if t == labels.MatchRegexp || t == labels.MatchNotRegexp {
+		re, err := regexp.Compile("^(?:" + v + ")$")
+		if err != nil {
+			return nil
```
beorn7, line 134: If this error happens, it is completely silent to the user. The matcher will just not exist (as you return `nil`). That's a big no-no.

As suggested above, we need to check that the regexp (and also the label name and value) is valid during config parsing. Then the compilation cannot fail here.

However, as also suggested (in my option (3) from above), we should probably not need to touch this code here anyawy.

beorn7 (changes requested):
Big picture: In principle, things should work as you have done here, but I think we should go one step further and fully embrace the `labels.Matcher` for matchers in the config (but not (yet) for matchers in silences, see comments for details). We can talk 1:1 about it if you have questions. (I have found a few issues with the `labels.Matcher` while looking at the code, which we can also discuss in person.)

"Operational" note: In this PR, you effectively ignore a parsing error twice. It's a good general rule for Go code to only ignore errors if you know exactly why, and then add a comment explaining why it is fine to ignore the error. (Spoiler: In these two cases, it's not fine…)

types/match.go
```diff
@@ -19,16 +19,18 @@ import (
...
-	Name    string `json:"name"`
-	Value   string `json:"value"`
-	IsRegex bool   `json:"isRegex"`
-
-	regex *regexp.Regexp
+	Type            labels.MatchType
+	Name            string `json:"name"`
+	Value           string `json:"value"`
+	IsRegex         bool   `json:"isRegex"`
+	NewRouteMatcher bool
```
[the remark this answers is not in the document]
aSquare14: I tried implementing (3) but ran into some issues. It would be really helpful if you could help me out with them.

I created a new type in config.go which unmarshals the yaml and uses labels.matchers. However, I am not able to understand how that translates to not modifying any code in[ match.go](https://github.com/prometheus/alertmanager/blob/master/types/match.go), since route accepts Matchers of type, [types.Matchers] (https://github.com/prometheus/alertmanager/blob/master/dispatch/route.go#L91) , we would need a new type.matcher which means creating a new constructor anyways. 

Apart from this, checking error while unmarshalling sounds good but I am confused when UnmarshalYAML is called. I am guessing it is called automatically but am not able to understand the flow of the code. 

cc @beorn7

types/match.go
```diff
@@ -103,6 +120,24 @@ func NewRegexMatcher(name model.LabelName, re *regexp.Regexp) *Matcher {
 	}
 }
 
+// NewRouteMatchers returns a new matcher that compares values against values or regular expressions.
+func NewRouteMatchers(t labels.MatchType, n, v string) *Matcher {
```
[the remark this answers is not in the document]
aSquare14: As mentioned above, case 3 would require a new constructor as well imho.

dispatch/route.go
```diff
@@ -96,6 +97,13 @@ func NewRoute(cr *config.Route, parent *Route) *Route {
 	for ln, lv := range cr.MatchRE {
 		matchers = append(matchers, types.NewRegexMatcher(model.LabelName(ln), lv.Regexp))
 	}
+	for _, ln := range cr.RouteMatchers {
+		labelMatchers, _ := labels.ParseMatchers(ln)
```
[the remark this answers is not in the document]
aSquare14: > And then perhaps already hold on to the parsed matchers for later, see other comments.)
This is not very clear to me. Could you please explain more ?

types/match.go
aSquare14: Yep, it is not good to ignore this error and I have moved it to config. But I am still a bit confused about unmarshalyaml and where that function will be called.

types/match.go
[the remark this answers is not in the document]
beorn7: Yes, we would need to switch over the routing to use `labels.Matchers`. With #2441 included, this should keep all existing configs in working state. `types.Matcher` would only be used by silences (and possibly get kicked out there eventually, but let's not think about that now).

I guess my cunning plan might collide with some devil in the detail. I could play with it now, but how about some pair programming tomorrow, if the pager permits? Let's sync up once I'm awake…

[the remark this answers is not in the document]
aSquare14: Yep sure. Sent you an invite !

config/config.go
```diff
@@ -746,6 +750,39 @@ func (re *Regexp) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return nil
 }
 
+type Matchers []labels.Matchers
```
beorn7, line 753: I think now that we have figured out the unmarshaling below, this can just be `type Matchers labels.Matchers` (we unmarshal the whole list of strings in L758 now).

config/config.go
```diff
@@ -746,6 +750,39 @@ func (re *Regexp) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	}
+
+	for _, line := range lines {
+		labelMatchers, err := labels.ParseMatchers(line)
+		if err != nil {
+			return err
+		}
+		for i := range labelMatchers {
+			if labelMatchers[i].Type == labels.MatchRegexp || labelMatchers[i].Type == labels.MatchNotRegexp {
+				re, err := regexp.Compile("^(?:" + labelMatchers[i].Value + ")$")
```
beorn7, line 769: No need to compile the regexp. That should all be done by `labels.ParseMatchers(line)` above. I think the only thing left to do here is to append all the `labelMatchers` to `&m`.

dispatch/route.go
```diff
@@ -88,14 +88,36 @@ func NewRoute(cr *config.Route, parent *Route) *Route {
...
-		matchers = append(matchers, types.NewRegexMatcher(model.LabelName(ln), lv.Regexp))
+		matcher, err := labels.NewMatcher(labels.MatchRegexp, ln, lv.String())
+		if err != nil {
+			// This error must not happen because the config already validates the yaml.
+			panic(err)
+		}
+		matchers = append(matchers, matcher)
 	}
+
+	// for ln, lv := range cr.RouteMatchers {
```
beorn7, line 111: And here we only need to do something like the following:
```Go
matchers = append(matchers, cr.Matchers...)
```

aSquare14: done

config/config.go
aSquare14: makes sense

types/match.go
aSquare14: Resolved this

dispatch/route.go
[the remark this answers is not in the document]
aSquare14: Discussed over pair programming

config/config.go
```diff
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"github.com/prometheus/alertmanager/pkg/labels"
```
beorn7, line 28: Let's put this into a third block further down. That's how we usually do it with packages from the same repository (see https://github.com/prometheus/alertmanager/blob/master/cmd/alertmanager/main.go ).

config/config.go
```diff
@@ -746,6 +748,32 @@ func (re *Regexp) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return nil
 }
 
+// Matchers encapsulates labels.Matchers and makes it YAML marshalable.
+type Matchers labels.Matchers
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface for Matchers.
+func (m *Matchers) UnmarshalYAML(unmarshal func(interface{}) error) error {
+
```
beorn7, line 756: Nitpicking: I wouldn't put a blank line here. It consumes screen space without a clear benefit.

config/config.go
```diff
@@ -746,6 +748,32 @@ func (re *Regexp) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+		if err != nil {
+			return err
+		}
+		for _, matcher := range labelMatchers {
+			matchers, err := labels.NewMatcher(matcher.Type, matcher.Name, matcher.Value)
+			if err != nil {
+				return err
+			}
+			*m = append(*m, matchers)
+		}
```
beorn7, line 772: This whole `for` loop can be replaced by `*m = append(*m, labelMatchers...)`.

config/config.go
```diff
@@ -746,6 +748,32 @@ func (re *Regexp) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+
+// UnmarshalYAML implements the yaml.Unmarshaler interface for Matchers.
+func (m *Matchers) UnmarshalYAML(unmarshal func(interface{}) error) error {
+
+	var lines []string
+	if err := unmarshal(&lines); err != nil {
+		return err
+	}
+	for _, line := range lines {
+		labelMatchers, err := labels.ParseMatchers(line)
```
beorn7, line 762: Not really happy with `labelMatchers`. How about `parsedMatchers`? Since the scope is so narrow here, you could even go for something like `mm` or `ms` or `pm`.

dispatch/route.go
```diff
@@ -45,7 +45,7 @@ type Route struct {
 
 	// Equality or regex matchers an alert has to fulfill to match
 	// this route.
-	Matchers types.Matchers
+	Matchers labels.Matchers
```
beorn7, line 48: The doc comment has to be updated now. "Equality or regex matchers…" → "Matchers…"

dispatch/route.go
```diff
@@ -88,14 +88,28 @@ func NewRoute(cr *config.Route, parent *Route) *Route {
 	}
 
 	// Build matchers.
-	var matchers types.Matchers
+	var matchers labels.Matchers
 
 	for ln, lv := range cr.Match {
```
beorn7, line 93: I'd comment on this `for` loop and the next that this handles the deprecated case.

dispatch/route.go
```diff
@@ -88,14 +88,28 @@ func NewRoute(cr *config.Route, parent *Route) *Route {
...
-		matchers = append(matchers, types.NewRegexMatcher(model.LabelName(ln), lv.Regexp))
+		matcher, err := labels.NewMatcher(labels.MatchRegexp, ln, lv.String())
+		if err != nil {
+			// This error must not happen because the config already validates the yaml.
+			panic(err)
+		}
+		matchers = append(matchers, matcher)
 	}
+
+	matchers = append(matchers, cr.Matchers...)
```
beorn7, line 111: And comment here something like: `// Finally append the new-style matchers. This can be simplified once the deprecated matcher syntax is removed.`

config/config.go
```diff
@@ -746,6 +748,32 @@ func (re *Regexp) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+		}
+		for _, matcher := range labelMatchers {
+			matchers, err := labels.NewMatcher(matcher.Type, matcher.Name, matcher.Value)
+			if err != nil {
+				return err
+			}
+			*m = append(*m, matchers)
+		}
+	}
+	return nil
```
beorn7, line 774: I guess it cannot harm to do a `sort.Sort(*m)` before returning for consistency. (And later, once the deprecated matchers are gone, we don't need to sort in `dispatch/route.go` again.)

types/match.go
```diff
@@ -23,6 +23,8 @@ import (
 )
 
 // Matcher defines a matching rule for the value of a given label.
+//
+// TODO(asquare14): This is now only used for silences and might be replaced by labels.Matchers.
```
beorn7, line 27: Strictly speaking, with just this PR in, it's also still used by inhibitions (which you plan to tackle in a separate PR). Perhaps mention it here and update the comment in the next PR.

dispatch/route.go
```diff
@@ -122,7 +136,8 @@ func NewRoutes(croutes []*config.Route, parent *Route) []*Route {
 // Match does a depth-first left-to-right search through the route tree
 // and returns the matching routing nodes.
 func (r *Route) Match(lset model.LabelSet) []*Route {
-	if !r.Matchers.Match(lset) {
+
```
beorn7, line 139: Again, I'd prefer to save whitespace and not have this blank line here, which doesn't really help to understand the structure of the method. (Note that there was no blank line before).

config/config.go
```diff
@@ -579,6 +580,7 @@ type Route struct {
 
 	Match    map[string]string `yaml:"match,omitempty" json:"match,omitempty"`
 	MatchRE  MatchRegexps      `yaml:"match_re,omitempty" json:"match_re,omitempty"`
```
beorn7, line 582: Let's append a `// Deprecated. Remove before v1.0 release.` to these two lines.

dispatch/route_test.go
```diff
@@ -383,3 +383,238 @@ routes:
 	require.Equal(t, child1.RouteOpts.GroupByAll, true)
 	require.Equal(t, child2.RouteOpts.GroupByAll, false)
 }
+
+func TestRouteMatchers(t *testing.T) {
```
beorn7, line 387: Perhaps also add a test that tries out a combination of new-style and old-style matchers? Essentially just duplicate this test and write some of the matchers in the old YAML syntax.

beorn7 (changes requested):
Looks great in general, just minor things to tweak.

Keep in mind that we also need to update the documentation, but we can do that after the next PR that updates the inhibition rules.

pkg/labels/matcher.go
```diff
@@ -86,3 +89,70 @@ func (m *Matcher) Matches(s string) bool {
...
+		buf.WriteString(m.String())
+	}
+	buf.WriteByte('}')
+
+	return buf.String()
+}
+
+// Match checks whether the label of the matcher has the specified
+// matching value.
+func (m *Matcher) Match(lset model.LabelSet) bool {
```
beorn7, line 138: This is only used in line 153 below, but it gives `Matcher` a method `Match(model.labelSet)` in addition to `Matches(string)`. That's a bit confusing. At least the naming should be fixed, perhaps to `MatchesLabelSet` or something`, and the method should be moved up in the file to the other methods of `Matcher`.

But perhaps, for now, it's better to inline this code into the method below (line 153, will suggest something there) and not have this method until it's needed elsewhere.

A also think this method is incorrect for negative matching. And the tests should actually catch that bug. Could you investigate why not?

pkg/labels/matcher.go
```diff
@@ -86,3 +89,70 @@ func (m *Matcher) Matches(s string) bool {
...
+	}
+	return string(v) == m.Value
+}
+
+// Match checks whether all matchers are fulfilled against the given label set.
+func (ms Matchers) Match(lset model.LabelSet) bool {
+	for _, m := range ms {
+		if !m.Match(lset) {
+			return false
+		}
```
beorn7, line 155:
```suggestion
		// Unset labels are treated as unset labels globally. Thus, if a
		// label is not set we retrieve the empty label which is correct
		// for the comparison below.
		v := lset[model.LabelName(m.Name)]
		if !m.Matches(string(v)) {
			return false
		}
```

pkg/labels/matcher.go
```diff
@@ -86,3 +89,70 @@ func (m *Matcher) Matches(s string) bool {
...
+func (m *Matcher) Match(lset model.LabelSet) bool {
+	// Unset labels are treated as unset labels globally. Thus, if a
+	// label is not set we retrieve the empty label which is correct
+	// for the comparison below.
+	v := lset[model.LabelName(m.Name)]
+
+	if m.Type == MatchNotRegexp || m.Type == MatchRegexp {
+		return m.re.MatchString(string(v))
+	}
+	return string(v) == m.Value
```
beorn7, line 147: You should just call `m.Matches(string(v))` here instead of replicating code. Also, as said above, negative matchers are not treated as inverted here, IIUC.

beorn7 (changes requested):
I found an actual issue after all, see comments.

pkg/labels/matcher.go
```diff
@@ -86,3 +89,70 @@ func (m *Matcher) Matches(s string) bool {
...
+		return true
+	}
+	if ms[i].Value > ms[j].Value {
+		return false
+	}
+	return ms[i].Value < ms[j].Value
+}
+
+// Matches checks whether all matchers are fulfilled against the given label set.
+func (ms Matchers) Matches(lset model.LabelSet) bool {
```
beorn7, line 112: Sorry for noticing only now: Here is the correct method!

pkg/labels/matcher.go
beorn7: I have found out now: The correct method is already above, with the name `Matches`.

beorn7: So just delete this method entirely, together with the Matchers.Match method below.

pkg/labels/matcher.go
beorn7: Actually, this suggestion is already implemented in the `Matches` method above. So just delete this entire method and update the reference in `routing.go` (Line 108) to `Matches(...)`

config/config.go
```diff
@@ -746,6 +748,32 @@ func (re *Regexp) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	return nil
 }
 
+// Matchers encapsulates labels.Matchers and makes it YAML marshalable.
```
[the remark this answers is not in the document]
aSquare14: Yes sounds better.

pkg/labels/matcher.go
beorn7: The code in Alertmanager itself uses the `Matches` method already. There is a reference to `Match` in the CLI code. This simply has to be changed to `Matches`, see below.

config/config.go
aSquare14: Yes, sounds good

config/config.go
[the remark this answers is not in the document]
beorn7: Actually, we also need a `MarshalYAML` method for `Matchers`. We marshal the config back into "clean" and normalized YAML, but that doesn't really work without the `MarshalYAML` method missing.

E.g., right now, if you look at the config in the web UI, matchers look like:
```
  routes:
  - matchers:
    - type: 0
      name: env
      value: foo
    - type: 3
      name: bar
      value: '''.*bla.*'''
```

So you need a `MarshalYAML` method based on the `String()` method of the matcher. Cf. the existing `MarshalYAML` method of the `Regexp` type in this file.

pkg/labels/matcher.go
aSquare14: Yep, so i deleted match and updated it in routing.go

pkg/labels/matcher.go
aSquare14: Ok makes sense

pkg/labels/matcher.go
aSquare14: Since I am totally removing this method, this is resolved by default

config/config.go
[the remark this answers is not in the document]
aSquare14: ok 👍  done

config/config.go
```diff
@@ -776,3 +782,39 @@ func (re Regexp) MarshalJSON() ([]byte, error) {
 	}
 	return nil, nil
 }
+
+// Matchers is label.Matchers with an added UnmarshalYAML method to implement the yaml.Unmarshaler interface.
```
beorn7, line 786: And now it also has `MarshalYAML` to implement the `yaml.Marshaler` interface.

config/config.go
```diff
@@ -776,3 +782,39 @@ func (re Regexp) MarshalJSON() ([]byte, error) {
...
+	sort.Sort(labels.Matchers(*m))
+	return nil
+}
+
+// MarshalYAML implements the yaml.Marshaler interface for Matchers.
+func (m *Matchers) MarshalYAML() (interface{}, error) {
+	var buf bytes.Buffer
+
+	buf.WriteByte('{')
+	for i, ms := range *m {
```
beorn7, line 811: `ms` is kind of misleading, sounds like the plural of m. I would either use `ms` as the name for the `Matchers` receiver (but then needs to be changed throughout all methods of `Matchers`), or use something else here, e.g. `matcher`.

config/config.go
```diff
@@ -776,3 +782,39 @@ func (re Regexp) MarshalJSON() ([]byte, error) {
...
+			return err
+		}
+		*m = append(*m, pm...)
+	}
+	sort.Sort(labels.Matchers(*m))
+	return nil
+}
+
+// MarshalYAML implements the yaml.Marshaler interface for Matchers.
+func (m *Matchers) MarshalYAML() (interface{}, error) {
```
beorn7, line 807: This has to be `func (m Matcher) ...` (i.e. no pointer receiver, otherwise the yaml package will not recognize this method).

config/config.go
```diff
@@ -776,3 +782,39 @@ func (re Regexp) MarshalJSON() ([]byte, error) {
...
+	buf.WriteByte('{')
+	for i, ms := range *m {
+		if i > 0 {
+			buf.WriteByte(',')
+		}
+		buf.WriteString(ms.String())
+	}
+	buf.WriteByte('}')
+
+	return buf.String(), nil
```
beorn7, line 819: This is just replicating the code in `Matchers.String`. You can just write
```Go
    return m.String(), nil
```
instead.

However, since we are marshaling a Yaml list type here, we have to return a slice to make it work properly. You could return `[]string{m.String()}`, but I would just put one matcher into each element of the Yaml list, i.e. do the following:
```Go
func (m Matchers) MarshalYAML() (interface{}, error) {
	result := make([]string, len(m))
	for i, matcher := range m {
		result[i] = matcher.String()
	}
	return result, nil
}
```

dispatch/route_test.go
```diff
@@ -383,3 +382,474 @@ routes:
...
+			t.Errorf("\nexpected:\n%v\ngot:\n%v", test.result, matches)
+		}
+
+		if !reflect.DeepEqual(keys, test.keys) {
+			t.Errorf("\nexpected:\n%v\ngot:\n%v", test.keys, keys)
+		}
+	}
+}
+
+func TestRouteMatchersAndMacth(t *testing.T) {
```
beorn7, line 621: Type: "Macth"

beorn7 (changes requested):
Almost there, just an issue with marshaling.

pkg/labels/matcher.go
```diff
@@ -86,3 +89,46 @@ func (m *Matcher) Matches(s string) bool {
...
+	if ms[i].Name > ms[j].Name {
+		return false
+	}
+	if ms[i].Name < ms[j].Name {
+		return true
+	}
+	if ms[i].Value > ms[j].Value {
+		return false
+	}
+	return ms[i].Value < ms[j].Value
```
vladimiroff, line 108:
```suggestion
	return ms[i].Value < ms[j].Value
```

Also, does it make sense to sort by `.Type` if both `.Name` and `.Value` are equal?

aSquare14: I don't see any use case for it...

beorn7: But it would be legal. To be conservative, I'd actually prefer if we included Type as the ultimate sorting criteria. Imagine someone accidentally puts in `{foo="bar", foo!="bar"}`. At least the rendered config should always use the same order to help that poor fellow to find their mistake.

beorn7:
Cool. Besides that minor issue about sorting by `Type` also, I think we are done here.

@simonpasquier could you give this a final look? Note that ideally, this gets merged together with #2441, where I'd appreciate your review, too.

pkg/labels/matcher.go
aSquare14: Done

pkg/labels/matcher.go
```diff
@@ -105,7 +105,13 @@ func (ms Matchers) Less(i, j int) bool {
...
 		return false
 	}
-	return ms[i].Value < ms[j].Value
+	if ms[i].Value < ms[j].Value {
+		return true
+	}
+	if ms[i].Type > ms[j].Type {
+		return false
+	}
+	return ms[i].Type < ms[j].Type
```
beorn7, line 114: Similar to what @vladimiroff said before, it's enough to just `return ms[i].Type < ms[j].Type`. The other three lines are redundant.

beorn7 (approved):

beorn7:
In the same spirit as #2441, I'll merge this now.

PR #2437, opened 2020-12-18
Fix and improve AlertmanagerClusterFailedToSendAlerts

beorn7:
The alert was just looking at the minimum across integrations. So a
complete failure of one integration would be masked by a still working
other integration. With this fix, the `integration` label is retained
(as it was already expected by the `description`), and thus any
one failing integration will trigger the alert.

In addition, an `alertmanagerCriticalIntegrationsRegEx` is provided
that allows to mark integrations as critical. Integrations that are
not used to deliver critical alerts, or those that are just there for
auditing and logging purposes can now be configured to only trigger a
warning alert if they fail.

@duologic relevant to the recent firing of AlertmanagerClusterFailedToSendAlerts in our production environment.

### DIFF
--- a/doc/alertmanager-mixin/alerts.libsonnet
+++ b/doc/alertmanager-mixin/alerts.libsonnet
@@ -60,10 +60,10 @@
           {
             alert: 'AlertmanagerClusterFailedToSendAlerts',
             expr: |||
-              min by (%(alertmanagerClusterLabels)s) (
-                rate(alertmanager_notifications_failed_total{%(alertmanagerSelector)s}[5m])
+              min by (%(alertmanagerClusterLabels)s, integration) (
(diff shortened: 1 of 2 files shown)


Duologic (approved):
I'm probably not autorative to approve this but very much welcome the change.

From the jsonnet perspective, it would be more interesting to have this as a mixin-function instead of a config option. But I understand that would diverge too much from the current paradigm in the alertmanager-mixin.

beorn7:
…and in fact how mixins in general do it in the wild. I'd like to push for a mixin-function approach in a concerted way one day.

beorn7:
@simonpasquier @roidelapluie I'd prefer if you could have a look. I'll give it a few more days.

But since this is essentially "just documentation", I hope it's OK to merge soon if you cannot find the time.

simonpasquier:
Thanks! LGTM

PR #2439, opened 2020-12-22
Use UTC when rendering alert and silence dates

tbregolin:
This mirrors the behaviour of the Alertmanager UI, which converts the
inputed string into UTC. This is better because otherwise you get mixed
timezones in the string representations when you query the
/api/v2/silences endpoint, for instance. It is also more "correct"
because Alertmanager uses UTC for presenting and updating dates, and so
amtool should do the same.


roidelapluie (approved):

simonpasquier:
I'm confused how it changes anything to the current implementation. AFAICT this PR would only change the representation of what's being sent to the API.

roidelapluie:
it means that the bug is that we do not convert the date when returning it to the user.

PR #2441, opened 2020-12-23
Improve labels.Matcher

beorn7:
@aSquare14 is working on using the same `labels.Matcher` that we already use in the UI and API for the matchers in the configuration, i.e. routing tree and inhibition rules. This will also introduce negative matching as requested in #1023 and simplify the syntax. See #2434. This will also help with making our code more DRY as we currently have two matcher implementation that do more or less the same: `labels.Matcher` and `types.Matcher`. We will still need the latter for silences, as their structure is directly marshaled into the HTTP API responses, so replacing them is a bit more involved.

Anyway, that's just for context.

While discussing all of this with @aSquare14, I realized that `labels.Matcher` has a few issues:
- It doesn't validate that the name is a valid Prometheus label name.
- It doesn't validate that the value is a valid UTF-8 string.
- It does not allow `=`, `~`, or `!` in the value, but these are perfectly valid parts of a value.
- There is no way to put a literal `"` into the value.
- While a line break is in principle fine to include, the UI doesn't allow it, and there is no escaping like `\n`.
- It kind of tries to be liberal (curly braces are optional, quoting is optional), but has quite strict requirements at other places (e.g. no whitespace around the comparison operator).

This PR adds more validation, allows all legal value elements, and at the same time becomes even more liberal.
It introduces OpenMetrics-style escaping (but keeps accepting single `\` where there is no ambiguity, so that this is very unlikely to break existing usage, like most commonly something like `\.` in a regexp).

It also adds doc comments to precisely explain the behavior. Once this is in, we should make sure the same is documented in the user docs. (Possibly, we need adjustments in the Elm UI code to allow the new features. But for use in the config, see #2434, it should be fine out of the box.)

### DIFF
--- a/pkg/labels/matcher.go
+++ b/pkg/labels/matcher.go
@@ -90,7 +90,7 @@ func (m *Matcher) Matches(s string) bool {
 
 // openMetricsEscape is similar to the usual string escaping, but more
 // restricted. It merely replaces a new-line character with '\n', a double-quote
-// character with '\"', and a backslack with '\\', which is the escaping used by
+// character with '\"', and a backslash with '\\', which is the escaping used by
 // OpenMetrics.
 func openMetricsEscape(s string) string {
 	r := strings.NewReplacer(
--- a/pkg/labels/matcher.go
+++ b/pkg/labels/matcher.go
@@ -16,6 +16,7 @@ package labels
 import (
 	"fmt"
 	"regexp"
+	"strings"
 )
 
 // MatchType is an enum for label matching types.
@@ -69,7 +70,7 @@ func NewMatcher(t MatchType, n, v string) (*Matcher, error) {
 }
 
 func (m *Matcher) String() string {
-	return fmt.Sprintf("%s%s%q", m.Name, m.Type, m.Value)
+	return fmt.Sprintf(`%s%s"%s"`, m.Name, m.Type, openMetricsEscape(m.Value))
 }
 
 // Matches returns whether the matcher matches the given string value.
@@ -85,4 +86,17 @@ func (m *Matcher) Matches(s string) bool {
 		return !m.re.MatchString(s)
 	}
 	panic("labels.Matcher.Matches: invalid match type")
+}
+
+// openMetricsEscape is similar to the usual string escaping, but more
+// restricted. It merely replaces a new-line character with '\n', a double-quote
+// character with '\"', and a backslack with '\\', which is the escaping used by
+// OpenMetrics.
+func openMetricsEscape(s string) string {
+	r := strings.NewReplacer(
+		`\`, `\\`,
+		"\n", `\n`,
+		`"`, `\"`,
+	)
+	return r.Replace(s)
 }
--- a/pkg/labels/parse.go
+++ b/pkg/labels/parse.go
@@ -38,7 +38,7 @@ var (
 // ParseMatchers parses a comma-separated list of Matchers. A leading '{' and/or
 // a trailing '}' is optional and will be trimmed before further
 // parsing. Individual Matchers are separated by commas outside of quoted parts
-// of the input string. Those commas may be followed by whitespace. Parts of the
+// of the input string. Those commas may be sorrounded by whitespace. Parts of the
 // string inside unescaped double quotes ('"…"') are considered quoted (and
 // commas don't act as separators there). If double quotes are escaped with a
 // single backslash ('\"'), they are ignored for the purpose of identifying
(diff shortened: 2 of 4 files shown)


beorn7:
@simonpasquier could you have a look? Or delegate this review to someone? I can also work on the matchers just between @aSquare14 and myself, if you agree with the general direction.

pkg/labels/matcher.go
```diff
@@ -86,3 +87,16 @@ func (m *Matcher) Matches(s string) bool {
 	}
 	panic("labels.Matcher.Matches: invalid match type")
 }
+
+// openMetricsEscape is similar to the usual string escaping, but more
+// restricted. It merely replaces a new-line character with '\n', a double-quote
+// character with '\"', and a backslack with '\\', which is the escaping used by
```
aSquare14, line 93: Typo : backslash

pkg/labels/matcher.go
```diff
@@ -86,3 +87,16 @@ func (m *Matcher) Matches(s string) bool {
...
+// OpenMetrics.
+func openMetricsEscape(s string) string {
```
aSquare14, line 95: OpenMetricsEscape maybe for consistency ?

pkg/labels/parse.go
```diff
@@ -30,27 +35,58 @@ var (
 	}
 )
 
+// ParseMatchers parses a comma-separated list of Matchers. A leading '{' and/or
+// a trailing '}' is optional and will be trimmed before further
+// parsing. Individual Matchers are separated by commas outside of quoted parts
+// of the input string. Those commas may be sorrounded by whitespace. Parts of the
```
aSquare14, line 41: Typo : surrounded

pkg/labels/parse.go
```diff
@@ -30,27 +35,58 @@ var (
...
+// string inside unescaped double quotes ('"…"') are considered quoted (and
+// commas don't act as separators there). If double quotes are escaped with a
+// single backslash ('\"'), they are ignored for the purpose of identifying
+// quoted parts of the input string. If the input string, after trimming the
+// optional trailing '}', ends with a comma, followed by optional whitespace,
+// this comma and whitespace will be trimmed.
+//
+// Examples for valid input strings:
+//   {foo = "bar", dings != "bums", }
+//   foo=bar,dings!=bums
```
aSquare14, line 51: maybe we could add an example
foo=bar, dings!=bums (whitespace after comma)

pkg/labels/parse.go
```diff
@@ -30,27 +35,58 @@ var (
...
+//   {quote="She said: \"Hi, ladies! That's gender-neutral…\""}
+//   statuscode=~"5.."
+//
+// See ParseMatcher for details how an individual Matcher is parsed.
```
aSquare14, line 55:
```suggestion
// See ParseMatcher for details on how an individual Matcher is parsed.
```

pkg/labels/parse.go
```diff
@@ -30,27 +35,58 @@ var (
...
 	s = strings.TrimPrefix(s, "{")
 	s = strings.TrimSuffix(s, "}")
 
-	var insideQuotes bool
-	var token string
-	var tokens []string
+	var (
+		insideQuotes bool
+		escaped      bool
+		token        strings.Builder
```
aSquare14, line 64: Just a doubt : Why did we change token's data type ?

aSquare14 (changes requested):
The code looks good to me. Just a few nits in the comments.

pkg/labels/matcher.go
beorn7: You see how damaged I am by Slack already… :-/

pkg/labels/matcher.go
beorn7: It would export this function, showing up in the Go doc etc. I see it more as an internal helper function that I don't want to expose.

pkg/labels/parse.go
beorn7: Done.

pkg/labels/parse.go
beorn7: Done.

pkg/labels/parse.go
beorn7: Done.

pkg/labels/parse.go
beorn7: Previously, the infamous string concatenation with quadratic runtime was used: `token += string(r)` (line 47 prior to this PR). `strings.Builder` is the straightforward way to avoid it.

beorn7:
@aSquare14 seems happy with the code. @simonpasquier we would really like to move forward with this. You have my full sympathy if it's hard for you to find time, but could you give any kind of feedback or delegate this to someone else? We now have also @vladimiroff onboard to push through with negative matchers for silences, too. I'm happy to lead the effort, but it would pain me if I had to interpret your silence as consent.

aSquare14 (approved):
LGTM !

beorn7:
As @aSquare14 has approved this and @simonpasquier hasn't reacted in 23 days, I assume it's fine to merge.

PR #2443, opened 2020-12-26
api: check silence matching by string comparison in getSilences

### LINKED ISSUE #2283: GET /silences API filters out regex silences counter-intuitively
**What did you do?**

913bfc7371918b1772cc721605ff6974

I created two silences in Alertmanager UI

- `alertname=~(foo|bar)`
- `alertname=baz`

and tried to filter them out by clicking each label (internally it uses `GET /api/v2/silences` API).

When we click `alertname=~(foo|bar)` label, it should show the list of silences which have the corresponding matcher, as it does so when we click `alertname=baz` label.

When we click `alertname=~(foo|bar)` label, it doesn't show silences that have the corresponding label (as shown in the screen capture).

---

In [`api.gettableSilenceMatchesFilterLabels`](https://github.com/prometheus/alertmanager/blob/12da9d6570ce0f487f98eed1be32e9c7e53da6b1/api/v2/api.go#L641), it judges whether a silence passes a filter by checking that the silence label value as a raw text matches the filter value as a regex pattern. That is, it checks whether a text `(foo|bar)` matches the regex pattern `(foo|bar)`, but obviously it doesn't.

IMO the right behavior should be that it won't use regex match but only checks the equivalence of labels, values, and the use of regex between a silence and a filter. I'm willing to fix the issue but also would like to discuss the way to fix here (or ML).

---

**Environment**

* System information:

```
$ uname -srm
Darwin 19.4.0 x86_64
```

* Alertmanager version:

```
$ docker run  prom/alertmanager --version
alertmanager, version 0.20.0 (branch: HEAD, revision: f74be0400a6243d10bb53812d6fa408ad71ff32d)
  build user:       root@00c3106655f8
  build date:       20191211-14:13:14
  go version:       go1.13.5
```

<details>
<summary>
alertmanager.yml
</summary>
```
global:
  resolve_timeout: 5m
  http_config: {}
  smtp_hello: localhost
  smtp_require_tls: true
  pagerduty_url: https://events.pagerduty.com/v2/enqueue
  hipchat_api_url: https://api.hipchat.com/
  opsgenie_api_url: https://api.opsgenie.com/
  wechat_api_url: https://qyapi.weixin.qq.com/cgi-bin/
  victorops_api_url: https://alert.victorops.com/integrations/generic/20131114/alert/
route:
  receiver: web.hook
  group_by:
  - alertname
  group_wait: 10s
  group_interval: 10s
  repeat_interval: 1h
inhibit_rules:
- source_match:
    severity: critical
  target_match:
    severity: warning
  equal:
  - alertname
  - dev
  - instance
receivers:
- name: web.hook
  webhook_configs:
  - send_resolved: true
    http_config: {}
    url: http://127.0.0.1:5001/
templates: []
```
</details>

ajalab:
Fixes #2283.

This PR changes how `gettableSilenceMatchesFilterLabels` in api/v2 checks whether a silence matches a given filter, especially for regex cases. In short,

- before: a silence matches a filter if a pattern value in the silence matches pattern strings of the filter by regex.
- after: a silence matches a filter if a pattern value in the silence and pattern strings of the filter are equivalent.

For instance, as shown in figure #2283, a silence `alertname~=(foo|bar)` didn't match a filter of the same pattern `alertname~=(foo|bar)`, since the regex match fails. After the fix, the silence and filter will match as we use string comparison instead. 

d0d03c3c0cdc3d9ee27a3c30cac4fb70

### DIFF
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -584,13 +584,13 @@ func (api *API) getSilencesHandler(params silence_ops.GetSilencesParams) middlew
 
 	sils := open_api_models.GettableSilences{}
 	for _, ps := range psils {
+		if !checkSilenceMatchesFilterLabels(ps, matchers) {
+			continue
+		}
 		silence, err := gettableSilenceFromProto(ps)
 		if err != nil {
 			level.Error(logger).Log("msg", "Failed to unmarshal silence from proto", "err", err)
 			return silence_ops.NewGetSilencesInternalServerError().WithPayload(err.Error())
-		}
-		if !gettableSilenceMatchesFilterLabels(silence, matchers) {
-			continue
 		}
 		sils = append(sils, &silence)
 	}
@@ -638,18 +638,21 @@ func sortSilences(sils open_api_models.GettableSilences) {
 	})
 }
 
-// gettableSilenceMatchesFilterLabels returns true if
+// checkSilenceMatchesFilterLabels returns true if
 // a given silence matches a list of matchers.
 // A silence matches a filter (list of matchers) if
 // for all matchers in the filter, there exists a matcher in the silence
 // such that their names, types, and values are equivalent.
-func gettableSilenceMatchesFilterLabels(s open_api_models.GettableSilence, matchers []*labels.Matcher) bool {
+func checkSilenceMatchesFilterLabels(s *silencepb.Silence, matchers []*labels.Matcher) bool {
 	for _, matcher := range matchers {
 		found := false
 		for _, m := range s.Matchers {
-			if matcher.Name == *m.Name &&
-				(matcher.Type == labels.MatchEqual && !*m.IsRegex || matcher.Type == labels.MatchRegexp && *m.IsRegex) &&
-				matcher.Value == *m.Value {
+			if matcher.Name == m.Name &&
+				(matcher.Type == labels.MatchEqual && m.Type == silencepb.Matcher_EQUAL ||
+					matcher.Type == labels.MatchRegexp && m.Type == silencepb.Matcher_REGEXP ||
+					matcher.Type == labels.MatchNotEqual && m.Type == silencepb.Matcher_NOT_EQUAL ||
+					matcher.Type == labels.MatchNotRegexp && m.Type == silencepb.Matcher_NOT_REGEXP) &&
+				matcher.Value == m.Pattern {
 				found = true
 				break
 			}
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -638,13 +638,28 @@ func sortSilences(sils open_api_models.GettableSilences) {
 	})
 }
 
+// gettableSilenceMatchesFilterLabels returns true if
+// a given silence matches a list of matchers.
+// A silence matches a filter (list of matchers) if
+// for all matchers in the filter, there exists a matcher in the silence
+// such that their names, types, and values are equivalent.
 func gettableSilenceMatchesFilterLabels(s open_api_models.GettableSilence, matchers []*labels.Matcher) bool {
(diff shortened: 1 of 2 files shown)


api/v2/api.go
```diff
@@ -638,13 +638,28 @@ func sortSilences(sils open_api_models.GettableSilences) {
...
+				found = true
+				break
+			}
+		}
+		if !found {
+			return false
+		}
 	}
 
-	return matchFilterLabels(matchers, sms)
```
hooten, line 647: If the new logic you have here were refactored into `matchFilterLabels`, would it also improve alert filtering (since `matchFilterLabels` seems to be used by that endpoint too)?

ajalab: Hmm, I think it wouldn't. IMO we should apply different logic to filter alerts or silences.

For instance, suppose we'd like to filter out alerts coming from `*.example.com` -- we can use regexp `instance=~".*.example.com"` for this purpose. In this case, we should rather apply the existing `matchFilterLabels` logic, which checks whether an alert label matches with the given matcher by using regexp.

On the other hand, unlike alerts, a label of silence may contain regex like `instance=~".*.example.com"` (silence all alerts from *.example.com). In this case, we should _not_ use regexp to check whether the silence label matches with the given matcher, but apply a simple string comparison.

beorn7: I tend to agree with @ajalab .

beorn7 (changes requested):
I like it. 

@ajalab by now, we support silences with negative matchers, at least in the "backend". Could you check what updates need to be done in your PR to deal with them correctly?

ajalab:
Sure, I'll work on it!

ajalab:
Currently, the OpenAPI model for silence doesn't support negative matchers, while the Protobuf model does:

<details>
<summary>OpenAPI</summary>

https://github.com/prometheus/alertmanager/blob/e6a1bede89403920a9875f43e6981ed46dd2ab05/api/v2/models/matcher.go#L32-L45

</details>


<details>
<summary>Protobuf</summary>

https://github.com/prometheus/alertmanager/blob/e6a1bede89403920a9875f43e6981ed46dd2ab05/silence/silencepb/silence.pb.go#L65-L75

https://github.com/prometheus/alertmanager/blob/e6a1bede89403920a9875f43e6981ed46dd2ab05/silence/silencepb/silence.pb.go#L35-L40

</details>

`getSilences` API checks matching based on OpenAPI models (`gettableSilenceMatchesFilterLabels`). So updates would be either

1. Update `gettableSilenceMatchesFilterLabels` to check matching based on Protobuf models, instead of OpenAPI
2. Fix `Matcher` OpenAPI model so that it supports negation

We may have to work on 2 at some point, but that needs more consideration. I believe the first solution should be reasonable and has less impact for now.

ajalab:
> Update gettableSilenceMatchesFilterLabels to check matching based on Protobuf models, instead of OpenAPI

I tried updating the PR in this direction. Now silences with negative matchers will start working immediately after OpenAPI models and clients/frontends are updated. Could you take a look again? @beorn7

beorn7:
The changes for alertmanager/api/v2/models/matcher.go are in #2471 (which is not yet merged).

beorn7:
The test failure might be due to flakiness. I have rerun the test.

(The test_frontend is known to be broken at the moment.)

beorn7:
OK, now we are down to only the expected test_frontend error.

beorn7 (approved):
I think we should go with this, but I'd much prefer if @simonpasquier could also chime in. I'll leave this open for another day or so to give him the opportunity.

beorn7:
I take silence as consent and merge this now.

PR #2445, opened 2021-01-03
Fix extended printing of regex sign

treydock:
Fixed extended printing to use same regex that is expected to be used when creating silences.


simonpasquier (approved):
thanks!

PR #2448, opened 2021-01-07
Implement and expose negative re matching

### LINKED ISSUE #1682: Allow negative (re) matching in silences

vladimiroff:
Closes #1682


beorn7:
Thanks @vladimiroff . Note that we are currently working on the negative matchers in the config, see #2434. We want to unify on the `labels.Matcher` while doing so, see also #2441. I hope we are done with this work very soon. After that, I'd much prefer if we could get rid of `types.Matcher` completely while introducing the negative matching for silences.

vladimiroff:
That makes a lot of sense. On several occasions I got confused whether I'm looking at the correct matcher. Given that this is already in the working instead of some nice-to-have-eventually task, there isn't a lot of sense to merge my PR now, change the API, only to have it changed in a completely different way a bit later.

Should I close this PR?

beorn7:
It would be highly appreciated if you would still be willing to work on this once we are done with the negative matchers in the config (for routing and inhibition). To do so, you could leave this PR open and change it to use `labels.Matcher`, or you can create an entirely new PR. Whatever is easier for you.

beorn7:
#2434 and #2441 are merged now. @aSquare14 will now work on negative matchers for inhibition rules.

@vladimiroff based on what we have now merged, you can now build the silence matchers on top of that. To reiterate:
- Silences should then also use `labels.Matcher(s)` so that we can remove `types.Matcher(s)` altogether. This will change the API to retrieve and create silences, but I assume that's inevitable with the introduction of negative matchers anyway. (But maybe we can keep it compatible or have some smooth transition strategy? To be investigated…) 
- The UI should follow suit. I imagine the silence UI will then look more similar to the filtering of alerts, where we already use `labels.Matcher`. Note that this will again mean some breaking changes in the URLs representing silences. I think here we really have to keep compatibility because very many users like to create links like `https://alertmanager.example.org/#/silences/new?filter=...` e.g. from alert templates, which we really don't want to break.
- Silences are also persisted and gossip'd. While obviously an old AM won't be able to understand new-style silences with negative matchers, we need to be able to load and understand old-style silences for a while.

vladimiroff:
I've prepared a commit which completely deletes `types.Matcher(s)?` and doesn't break any tests **on my machine**, without touching UI and the only change in the protobuf API is adding `labels.Matcher`'s two types for negative matching.

Will open a new PR in a bit, because I'm frankly in denial. This change shouldn't be that straightforward. Hopefully you and/or CI's jobs will show me what's really wrong.

vladimiroff:
Closing this in favor of #2456.

beorn7:
Great, will have a look ASAP.

And WRT tests: they might not test every detail of the API.

Issue #2447, opened 2021-01-07
Alertmanager of other nodes resends same alerts again after resolve notification 


joernkleinbub:
**What did you do?**
I setted up a Prometheus and Alertmanager in HA on 3 Vagrant nodes on my local machine for testing purposes.
I established a single DemoAlert "sum(prometheus_build_info) < 3" via WebHook and provoke a raise of the alert by killing one of the prometheus. 

**What did you expect to see?**
Chronological: 
1. Prometheus is killed
2. The alert is fired.
3. Prometheus is up again
4. An resolved message.

**What did you see instead? Under which circumstances?**
Chronological (Notifications and Logs please see below): 
1. Prometheus is killed
2. An alert notification is fired by Alertmanager "http://n3:9093" 2021-01-07T09:38:11.059+01:00
3. Prometheus is up again
2. An resolved notification by Alertmanager "http://n3:9093":  2021-01-07T09:39:00.175+01:00
3. The same alert notification (2.) is fired again by Alertmanger "http://n2:9093":  2021-01-07T09:39:10.572+01:00
4. The same alert notification (2.) is fired again by Alertmanger "http://n1:9093": 2021-01-07T09:39:10.627+01:00
4. A resolve message is fired by  Alertmanger "http://n1:9093": 2021-01-07T09:39:25.801+01:00

**Environment**

* System information:
Linux 4.9.0-12-amd64 x86_64
Vagrant/Virtualbox nodes

* Alertmanager version:
vagrant@n3:~$ alertmanager-0.21.0.linux-amd64/alertmanager --version
alertmanager, version 0.21.0 (branch: HEAD, revision: 4c6c03ebfe21009c546e4d1e9b92c371d67c021d)
  build user:       root@dee35927357f
  build date:       20200617-08:54:02
  go version:       go1.14.4

-  Prometheus version:
vagrant@n3:~$ prometheus-2.24.0-rc.0.linux-amd64/prometheus --version
prometheus, version 2.24.0-rc.0 (branch: HEAD, revision: 0cb133aaa5c3bcb0fcbfb0b34c256fd3483a9bfd)
  build user:       root@fdb12576957f
  build date:       20201230-15:28:10
  go version:       go1.15.6
  platform:         linux/amd64

* Alertmanager configuration file:
[...]

PR #2450, opened 2021-01-08
Reproduce alertmanager resending alerts

simonswine:
This reproduces the issue mentioned in #2447:

With some combinations of `group_interval` and `group_delay` ratios alerts in a clustered alertmanager can be resend multiple times

### DIFF
--- a/test/with_api_v2/acceptance/cluster_test.go
+++ b/test/with_api_v2/acceptance/cluster_test.go
@@ -28,6 +28,8 @@ func TestClusterDeduplication(t *testing.T) {
 	t.Parallel()
 
 	conf := `
+global:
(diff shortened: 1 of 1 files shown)


gotjosh:
Maybe @beorn7 has an opinion on this.

beorn7:
Thanks for reproducing this.

I am actually not an expert for Alertmanager. I needed some thinking to understand this behavior completely.

@simonpasquier is the maintainer. Perhaps he will understand this more readily or knows someone who will.

simonpasquier:
I've commented on #2447 why this may happen (not saying that we shouldn't improve the situation :smiley:). I think we can close this PR since we have all data in the originating issue. What do you think @simonswine?

PR #2452, opened 2021-01-09
cosmetics improvement on if statement and other few changes

yabetsu93:
What is changes:
cosmetics changes and few changes on the code

### DIFF
--- a/cli/alert_add.go
+++ b/cli/alert_add.go
@@ -117,6 +117,9 @@ func (a *alertAddCmd) addAlert(ctx context.Context, _ *kingpin.ParseContext) err
 
 	amclient := NewAlertmanagerClient(alertmanagerURL)
 
-	_, err = amclient.Alert.PostAlerts(alertParams)
-	return err
+	if _, err = amclient.Alert.PostAlerts(alertParams); err != nil {
+		return err
+	}
+
+	return nil
 }
--- a/cli/silence_add.go
+++ b/cli/silence_add.go
@@ -29,11 +29,11 @@ import (
 )
 
 func username() string {
-	user, err := user.Current()
-	if err != nil {
-		return ""
+	if user, err := user.Current(); err == nil {
+		return user.Username
 	}
-	return user.Username
+
+	return ""
 }
 
 type silenceAddCmd struct {
@@ -169,6 +169,9 @@ func (c *silenceAddCmd) add(ctx context.Context, _ *kingpin.ParseContext) error 
 	if err != nil {
 		return err
 	}
-	_, err = fmt.Println(postOk.Payload.SilenceID)
-	return err
+	if _, err = fmt.Println(postOk.Payload.SilenceID); err != nil {
+		return err
+	}
+
+	return nil
 }
--- a/cli/silence_import.go
+++ b/cli/silence_import.go
@@ -90,8 +90,7 @@ func (c *silenceImportCmd) bulkImport(ctx context.Context, _ *kingpin.ParseConte
 
 	dec := json.NewDecoder(input)
 	// read open square bracket
-	_, err = dec.Token()
-	if err != nil {
+	if _, err = dec.Token(); err != nil {
 		return errors.Wrap(err, "couldn't unmarshal input data, is it JSON?")
 	}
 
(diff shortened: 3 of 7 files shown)


yabetsu93 (commented):
@beorn7 can you take a look? a simple cosmetic changes if you think is not needed i will close this pr :)

cli/alert_add.go
```diff
@@ -117,6 +117,9 @@ func (a *alertAddCmd) addAlert(ctx context.Context, _ *kingpin.ParseContext) err
 
 	amclient := NewAlertmanagerClient(alertmanagerURL)
 
-	_, err = amclient.Alert.PostAlerts(alertParams)
-	return err
+	if _, err = amclient.Alert.PostAlerts(alertParams); err != nil {
+		return err
+	}
+
+	return nil
```
beorn7, line 124: I prefer how it was done before. This code here seems needlessly wordy.

cli/silence_add.go
```diff
@@ -169,6 +169,9 @@ func (c *silenceAddCmd) add(ctx context.Context, _ *kingpin.ParseContext) error
 	if err != nil {
 		return err
 	}
-	_, err = fmt.Println(postOk.Payload.SilenceID)
-	return err
+	if _, err = fmt.Println(postOk.Payload.SilenceID); err != nil {
+		return err
+	}
+
+	return nil
```
beorn7, line 176: Same as above. I prefer the old state.

cli/silence_import.go
```diff
@@ -103,7 +102,7 @@ func (c *silenceImportCmd) bulkImport(ctx context.Context, _ *kingpin.ParseConte
 		wg.Add(1)
 		go func() {
 			addSilenceWorker(ctx, amclient.Silence, silencec, errc)
-			wg.Done()
+			defer wg.Done()
```
beorn7, line 105: What's the use of deferring here? This is the last thing in the function. It will be executed immediately anyway.

client/client.go
```diff
@@ -238,8 +238,11 @@ func (h *httpAlertAPI) Push(ctx context.Context, alerts ...Alert) error {
 		return fmt.Errorf("error creating request: %v", err)
 	}
 
-	_, _, err = h.client.Do(ctx, req)
-	return err
+	if _, _, err = h.client.Do(ctx, req); err != nil {
+		return err
+	}
+
+	return nil
```
beorn7, line 245: See above, I'm against this.

client/client.go
```diff
@@ -294,8 +297,11 @@ func (h *httpSilenceAPI) Expire(ctx context.Context, id string) error {
...
```
beorn7, line 304: See above, I'm against this.

client/client.go
```diff
@@ -319,9 +325,12 @@ func (h *httpSilenceAPI) Set(ctx context.Context, sil types.Silence) (string, er
...
 		SilenceID string `json:"silenceId"`
 	}
-	err = json.Unmarshal(body, &res)
 
-	return res.SilenceID, err
+	if err = json.Unmarshal(body, &res); err != nil {
+		return "", err
+	}
+
+	return res.SilenceID, nil
```
beorn7, line 333: Again, this seems needlessly wordy. I would only do it if there is any reason to ensure the string return is empty in case of an error. But I don't think there is. On the contrary.

client/client.go
```diff
@@ -343,7 +352,9 @@ func (h *httpSilenceAPI) List(ctx context.Context, filter string) ([]*types.Sile
 	}
 
 	var sils []*types.Silence
-	err = json.Unmarshal(body, &sils)
+	if err = json.Unmarshal(body, &sils); err != nil {
+		return nil, err
+	}
 
-	return sils, err
+	return sils, nil
```
beorn7, line 359: See above.

beorn7 (commented):
Personally, I prefer to return errors directly where that's possible. The pattern
```Go
if err != nil {
    return err
}
return nil
```
is needlessly verbose and IMHO even confusing. My brain immediately starts to waste cycles thinking why the code doesn't just read `return err`, suspecting that the latter would be a different (and wrong) behavior.

I'm fine with the changes that use the `if _, err := foo(); err != nil {return err}` pattern. It's more compact and arguably what Go wants (because this syntax pattern exists in the language). However, I know that some dislike it and prefer to not condense lines in this way.

@simonpasquier is the maintainer and should make the call for both cases.

yabetsu93:
Thanks for the review perhaps will close this pr,

PR #2456, opened 2021-01-15
Replace types.Matcher(s)? with labels.Matcher(s)?

### LINKED ISSUE #1682: Allow negative (re) matching in silences

vladimiroff:
Closes #1682.


beorn7:
cc @aSquare14

client/client_test.go
```diff
@@ -136,11 +137,11 @@ func TestAPI(t *testing.T) {
...
 		ID: "abc",
-		Matchers: []*types.Matcher{
+		Matchers: []*labels.Matcher{
 			{
-				Name:    "label1",
-				Value:   "test1",
-				IsRegex: false,
+				Name:  "label1",
+				Value: "test1",
+				Type:  labels.MatchEqual,
```
aSquare14, line 144: Maybe add more tests for labels.MatchNotEqual, labels.MatchRegexp and labels.MatchNotRegexp

test/with_api_v1/mock.go
```diff
@@ -106,17 +107,18 @@ func (s *TestSilence) nativeSilence(opts *AcceptanceOpts) *types.Silence {
...
 			Value: s.match[i+1],
 		})
 	}
 	for i := 0; i < len(s.matchRE); i += 2 {
-		nsil.Matchers = append(nsil.Matchers, &types.Matcher{
-			Name:    s.matchRE[i],
-			Value:   s.matchRE[i+1],
-			IsRegex: true,
-		})
+		m, err := labels.NewMatcher(labels.MatchRegexp, s.matchRE[i], s.matchRE[i+1])
```
aSquare14, line 117: maybe rename "m"

types/types.go
```diff
@@ -409,7 +410,7 @@ type Silence struct {
 	ID string `json:"id"`
 	// A set of matchers determining if a label set is affect
```
aSquare14, line 411: Typo: affected

aSquare14 (changes requested):

client/client_test.go
beorn7: This PR just moves from `types.Matcher` to `labels.Matcher`, so `labels.MatchNotEqual`and `labels.MatchNotRegexp` aren't supported yet.

In the next step, that will be done (but it also requires updates to the UI), and in that PR, you are right to ask for those tests.

beorn7: Actually, via the UI, you can probably already use the new silences, but we have no way to display them in the UI. So you could arguably already add the tests in the API tests. Mmmmh…

beorn7:
Thanks @aSquare14 for looking into this. Will also review tomorrow.

test/with_api_v1/mock.go
vladimiroff: Given that the scope of this variable is 4 lines, I would claim that the name is fine. Open to suggestions, though.

client/client_test.go
vladimiroff: I agree some tests covering silences with the new matcher types are in place. Here in `client` tests cover nothing more than the CRUD operations, though. How about some tests (mostly) regarding matching in the `silences` package, instead?

vladimiroff: See 4c0f5dfcc6fde0466c8d0f63630b06d38eff92c7

types/types.go
vladimiroff: See 452deafb611d425ae7e0362926c9d4f9a9ffcc33

beorn7:
I'm sorry, but I won't find time to review this today. I'll try again tomorrow.

inhibit/inhibit.go
```diff
@@ -167,22 +168,30 @@ type InhibitRule struct {
...
-		sourcem = append(sourcem, types.NewMatcher(model.LabelName(ln), lv))
+		sourcem = append(sourcem, &labels.Matcher{
+			Type:  labels.MatchEqual,
+			Name:  ln,
+			Value: lv,
+		})
 	}
 	for ln, lv := range cr.SourceMatchRE {
-		sourcem = append(sourcem, types.NewRegexMatcher(model.LabelName(ln), lv.Regexp))
+		sourcem = append(sourcem, labels.NewRegexpMatcher(ln, lv.Regexp))
```
beorn7, line 183: I see what you are doing here. You are avoiding to recompile the regexp. I think this is not in a performance critical path, plus this will go away once the deprecated way of configuring inhibition rules is gone. So I suggest to not create a specialized `labels.NewRegexpMatcher` and do it as @aSquare14 did it before for the routing: https://github.com/prometheus/alertmanager/blob/master/dispatch/route.go#L105

inhibit/inhibit.go
```diff
@@ -167,22 +168,30 @@ type InhibitRule struct {
...
-		targetm = append(targetm, types.NewMatcher(model.LabelName(ln), lv))
+		targetm = append(targetm, &labels.Matcher{
+			Type:  labels.MatchEqual,
+			Name:  ln,
+			Value: lv,
+		})
 	}
 	for ln, lv := range cr.TargetMatchRE {
-		targetm = append(targetm, types.NewRegexMatcher(model.LabelName(ln), lv.Regexp))
+		targetm = append(targetm, labels.NewRegexpMatcher(ln, lv.Regexp))
```
beorn7, line 194: Same here, see L183.

inhibit/inhibit.go
```diff
@@ -167,22 +168,30 @@ type InhibitRule struct {
...
+		sourcem = append(sourcem, labels.NewRegexpMatcher(ln, lv.Regexp))
 	}
 
 	for ln, lv := range cr.TargetMatch {
-		targetm = append(targetm, types.NewMatcher(model.LabelName(ln), lv))
+		targetm = append(targetm, &labels.Matcher{
+			Type:  labels.MatchEqual,
+			Name:  ln,
+			Value: lv,
+		})
```
beorn7, line 191: Just to be conservative, could you use `NewMatcher(labels.MatchEqual, ln, lv)` here? I am aware that this requires"handling" the error (just panic in case it err's out because that "must not happen" (TM)). But there might be changes to the labels.Matcher type in the future where more work is required in the constructor.

pkg/labels/matcher.go
```diff
@@ -152,3 +152,11 @@ func (ms Matchers) String() string {
...
 	return buf.String()
 }
+
+// NewRegexpMatcher returns a matcher with already compiled regexp.Regexp.
+//
+// TODO(vladimiroff): Get rid of this function once migration from
+// types.Matcher is complete.
+func NewRegexpMatcher(n string, re *regexp.Regexp) *Matcher {
+	return &Matcher{Type: MatchRegexp, Name: n, Value: re.String(), re: re}
+}
```
beorn7, line 162: As discussed above, I think we should not add this function and do a bit of extra work for the time the deprecated parts are still in.

test/with_api_v1/mock.go
beorn7: Yeah, I think it's fine as is.

beorn7 (changes requested):
@aSquare14 and I looked at this together. We understand now, why it doesn't break persistence, gossiping, and the v2 API. That's because this all goes via the protobuf representation.

However, the v1 API marshals the matcher type into JSON directly, and that changes. I'm just not sure if that matters, most people should use the v2 API, and I guess we want to kick out the v1 API before releasing v1.0 of Alertmanager. @simonpasquier , could you provide context on the plans here?

beorn7:
For more context, this is how a silence looks now in the v1 API:
```JSON
{
  "status": "success",
  "data": [
    {
      "id": "c8c3dcd6-3832-4107-8b9c-8f40ce200da2",
      "matchers": [
        {
          "Type": 2,
          "Name": "foo",
          "Value": "bar"
        }
      ],
      "startsAt": "2021-01-20T17:23:19.624814507Z",
      "endsAt": "2021-01-20T19:23:03.566Z",
      "updatedAt": "2021-01-20T17:23:19.624814507Z",
      "createdBy": "beorn",
      "comment": "blablab la",
      "status": {
        "state": "active"
      }
    }
  ]
}
```

This is how it looked before:

```json
{
  "status": "success",
  "data": [
    {
      "id": "c8c3dcd6-3832-4107-8b9c-8f40ce200da2",
      "matchers": [
        {
          "name": "foo",
          "value": "bar",
          "isRegex": true
        }
      ],
      "startsAt": "2021-01-20T17:23:19.624814507Z",
      "endsAt": "2021-01-20T19:23:03.566Z",
      "updatedAt": "2021-01-20T17:23:19.624814507Z",
      "createdBy": "beorn",
      "comment": "blablab la",
      "status": {
        "state": "active"
      }
    }
  ]
}
```

beorn7:
So in principle, we could provide a customized MarshalJSON method to emulate the old behavior. Not sure if it is worth it, as the v1 API might not be in much use anymore. But it would definitely be the conservative move.

vladimiroff:
@beorn7 totally missed that. Given how trivial is the change to ensure backwards compatibility, I don't see a reason NOT to do it. For the sake of supporting `MatchNotEqual` and `MatchNotRegexp` matcher types, will add  `isEqual` key to the object.

beorn7:
It's easy to miss because it's not covered by tests. And it's probably not covered because it's the old deprecated v1 API.

But if you could keep the backwards compatibility easily, that would certainly be the best solution.

vladimiroff:
Actually, now looking at it... the v2 api doesn't support **all** matcher types, either:

```yaml
  matcher:
    type: object
    properties:
      name:
        type: string
      value:
        type: string
      isRegex:
        type: boolean
    required:
      - name
      - value
      - isRegex
```

beorn7:
Yes, but it doesn't break by simply replacing `types.Matchers` with `labels.Matchers`. (The v1 API uses the JSON marshaling of the respective `Matchers` type, while v2 has Swagger marshaling of the protobuf type.)

So yeah, to truly support negative matchers for silences, the v2 code has to be adjusted, too, which means getting into the Swagger definitions and regenerate the code etc.

_But_ I think the scope of this PR is merely to replace `types.Matchers` with `labels.Matchers`, without changing anything else yet, i.e. back the current feature set with `labels.Matchers`. Next step would be to extend the APIs _and_ the web UI to actually support negative matching. (And in the UI, that's actually a bit more involved because we probably want to replace the current two text inputs and a checkbox for regexp with just one text input similar to what we have for alert filtering. Also, deep-linking into a filled silence creation form needs to support the new matching, but the old links should still keep working.)

beorn7:
`amtool` also needs to support negative matching with silences. (Luckily, it is already using `labels.Matchers`.)

beorn7:
So yeah, as said, just keep this PR to what the headline says: Remove `types.Matcher(s)` from the codebase without changing anything feature-wise yet.

vladimiroff:
> Given how trivial is the change to ensure backwards compatibility

Famous last words, right there. I've tried plugging `MarshalJSON` on three different places and every time I end up breaking the same test - `./test/with_api_v1/acceptance.TestSilenceDelete`. Will try digging deeper tomorrow.

beorn7:
Good luck. (o:

vladimiroff:
Argh, I obviously need to run `make all` in order for the acceptance tests to be against latest changes. @beorn7 adding a new key shouldn't break compatibility, so I've added the `isEqual` key in order not to serve wrong data to v1 clients. Is that okay?

beorn7:
Sounds good. I'll have a look ASAP.

vladimiroff:
Let me just rebase on top of master given that #2460 just got merged.

beorn7 (approved):
Great. I think we can ship it.

@aSquare14 would you like to give it a final check?

aSquare14:
LGTM 🥳

aSquare14 (approved):

PR #2463, opened 2021-01-22
Add JSON marshaling to config.Matchers

beorn7:
I have found out that we still need this for the v1 API (and possibly
elsewhere).

@aSquare14 could you have a look?

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -763,7 +763,7 @@ func (re Regexp) MarshalYAML() (interface{}, error) {
 	return nil, nil
 }
 
-// UnmarshalJSON implements the json.Marshaler interface for Regexp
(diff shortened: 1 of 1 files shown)


aSquare14 (approved):

roidelapluie:
Note that v1 can not be deprecated as it is used by prometheus server by default

aSquare14:
LGTM. Thank you

beorn7:
Wasn't the plan to move to using v2 by default in prometheus/prometheus?

roidelapluie:
> Wasn't the plan to move to using v2 by default in prometheus/prometheus?

Created an issue in prometheus to discuss this.

PR #2467, opened 2021-01-27
fix(notify/victorops): Catch routing_key templating errors

jdbaldry:
The error used during the templating of the routing key is not checked before it is later reassigned.


simonpasquier (approved):
Thanks!

PR #2475, opened 2021-02-04
Make MaxGossipPacketSize public

gotjosh:
I understand `MaxGossipPacketSize` has several implications in the way clustering works (oversized messages are always propagated via TCP which influence the number of parallel connections we make) but also downstream implementations might want to configure this.

I don't think the alertmanager offers any guarantees if downstream implementations change this value so it should be safe to make public? 😄


simonpasquier (approved):
sure, no issue making it public

PR #2478, opened 2021-02-05
Update github.com/gogo/protobuf to v1.3.2

simonpasquier:
Fix for CVE-2021-3121

Signed-off-by: Simon Pasquier <spasquie@redhat.com>


beorn7 (approved):
The frontend CI test has recently started to fail everywhere (`/bin/sh -c npm install -g   elm@0.19.0   elm-format@0.8.0   elm-test@0.19.0-beta5   uglify-js@3.4.7` doesn't work anymore).

Perhaps we should do the same (or do it first) in release-0.21 so that we can cut a bugfix release before releasing the more invasive changes like silences with negative matchers (also see https://github.com/prometheus/alertmanager/pull/2479 )?

simonpasquier:
> The frontend CI test has recently started to fail everywhere

yes I've noticed too as I was confused why this PR would break anything with the frontend :)

simonpasquier:
and yes to path release-0.21 first :)

beorn7:
Is what @roidelapluie said in https://github.com/prometheus/prometheus/pull/8446 also true for this PR? In that case, I'd say we shouldn't put it in a bugfix release.

simonpasquier:
To be honest, I have a hard time assessing how you can exploit the vulnerability and how the new version of the generated code is fundamentally different from what we have today. I'm looking for answers from our security team. Without further information, I agree that we don't have to patch v0.21.0.

roidelapluie:
Sorry for the screenshot, that was the more pratical I could find

skippy

As you can see, some of the code would be generate withot checking that `index+skippy < 0`.
I guess the change below was moving both checks `skippy < 0`  and `index+skippy < 0` on the same line to avoid copy/paste omissions in the future.

simonpasquier:
@roidelapluie :bow: it was the missing piece to me, thanks a lot!

beorn7:
Just for the record: The generated code that is actually used in Alertmanager is not affected by CVE-2021-3121. Therefore, we don't need to cut a bugfix release for 0.21.

PR #2479, opened 2021-02-05
Catch unknown matcher types in gossipped silences

beorn7:
I believe that gossibping between AM instances with different versions
of the silence protobuf could create a situation where an unknown
matcher type won't be recognized, thereby possibly inverting the
behavior of the silence.

This makes the matching more robust by error'ing out when facing an
unknown matcher type.

This has come to my attention during the recent work on silences with
negative matching. To make cluster upgrades more robust, this fix is
needed, and therefore I'm proposing it for a bugfix release of v0.21.

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -79,6 +79,8 @@ func (c matcherCache) add(s *pb.Silence) (types.Matchers, error) {
 			mt.IsRegex = false
 		case pb.Matcher_REGEXP:
 			mt.IsRegex = true
+		default:
+			return nil, errors.Errorf("unknown matcher type %q", m.Type)
 		}
 		err := mt.Init()
 		if err != nil {


beorn7:
cc @vladimiroff @aSquare14 @QuentinBisson

vladimiroff (approved):

roidelapluie (approved):

QuentinBisson (approved):

aSquare14 (approved):

beorn7:
Thanks everyone for your review.

I have played with this, the unmodified AM, and the new silences with negative matchers. There is good and bad news:

The good news is that even without this change, the AM isn't just inverting the meaning of a silence. A silence with a negative matcher will make it into the storage via gossipping, but it will not silence anything.

The bad news is that with or without this change, the UI just breaks because new silences with negative matching cannot be unmarshaled from protobuf, so the API returns a 500, which breaks the display of silences as a whole.

We could certainly tweak v0.21 to treat the new silences more gracefully, but I think the important part is that the new silences don't silence anything silently (no pun intended) they are not supposed to silence. Even if we created a more graceful degradation, the solution would still be to update all instances in the cluster. So arguably, breaking less gracefully is even better because it is easier to notice.

Therefore, I'll close this and won't cut a v0.21.1. (Conveniently, the protobuf CVE-2021-3121 has just turned out to not affect Alertmanager.).

If you disagree with my verdict, please follow up.

PR #2480, opened 2021-02-07
Allow filtering silences by creator

### LINKED ISSUE #2476: Allow filtering silences by creator
**What did you do?**
I tried to filter silences which were created by me (to see if I forgot to expire them). As this didn't work, I tried to filter by the exact ID, which also didn't work. I asked in the Martrix channel 
https://matrix.to/#/!HaYTjhTxVqshXFkNfu:matrix.org$161244270149569UcYSG:matrix.org?via=matrix.org
and it seems this is not possible. 
I mean it could be that this is a strange use case.

**Environment**
Alertmanager version: 0.21

hooten:
Allow using the filter `createdBy` for the get silences endpoint.

https://user-images.githubusercontent.com/6402556/107154279-4bbd2b00-692f-11eb-96af-cc3b7b67b863.mov

Closes #2476

### DIFF
--- a/api/v2/api_test.go
+++ b/api/v2/api_test.go
@@ -68,11 +68,10 @@ func assertEqualStrings(t *testing.T, expected string, actual string) {
 
 var (
 	testComment = "comment"
-	createdBy   = "test"
 )
 
 func gettableSilence(id string, state string,
-	updatedAt string, start string, end string,
+	updatedAt string, start string, end string, createdBy string,
 ) *open_api_models.GettableSilence {
 
 	updAt, err := strfmt.ParseDateTime(updatedAt)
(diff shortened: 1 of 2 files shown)


api/v2/api_test.go
```diff
@@ -210,3 +209,46 @@ func TestMatchFilterLabels(t *testing.T) {
 		require.Equal(t, tc.expected, matchFilterLabels(ms, sms))
 	}
 }
+
+func TestFilterByCreator(t *testing.T) {
+	startTime := "2021-02-7T12:00:00+00:00"
+	endTime := "2021-02-7T11:00:00+00:00"
+	updateTime := "2021-02-7T12:00:00+00:00"
```
beorn7, line 216: `-7` → `-07`. I think this is creating the test failure.

beorn7 (commented):
While I like the feature, I have concerns that we are now mixing up filtering for matchers with other filtering criteria. What would happen if someone had an actual label `createdBy`, and then places a silence using this label in a matcher, and then wants to filter for that one?

Another problem is that users will probably ask next for filters for the start or end date (with "before" and "after" and maybe even for words in the comment field. We will have a hard time to come up with more "magic" pseudo-labels. So perhaps we should extend the filter form, possibly hidden behind an "Show advanced filter option" button?

hooten:
Thanks for the comments @beorn7. I completely agree. I'll rethink the PR.

PR #2481, opened 2021-02-09
docs for negative matchers for route, inhibition

aSquare14:
Related issue : #1023 
Related Pull requests : #2434 #2460

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -594,33 +594,33 @@ Here are some examples of valid string matchers :
 
 1. Shown below are two equality matchers combined in a long form YAML list.
 
-```
+```yaml
   matchers :
    - foo = bar
-   - dings !=bums
+   - dings !=bums 
 ```
 
 2. Similar to example 1, shown below are two equality matchers combined in a short form YAML list.
 
-```
+```yaml
 matchers: [ "foo = bar", "dings != bums" ]
 ```
 
 As shown below, in the short-form, it's generally better to quote the list elements to avoid problems with special characters like commas:
 
-```
+```yaml
 matchers: [ '{foo="bar",dings!="bums"}' ]
 ```
 
 3. You can also put both matchers into one PromQL-like string. Single quotes for the whole string work best here.
 
-```
+```yaml
 matchers: ['{env=~"produ.*"}', '{baz!~".*quux"}']
 ```
 
 4. To avoid any confusion about YAML string quoting and escaping, you can use YAML block quoting and then only worry about the OpenMetrics escaping inside the block. A complex example with a regular expression and different quotes inside the label value is shown below:
 
-```
+```yaml
 matchers:
   - |
       {quote=~"She said: \"Hi, all!( How're you…)?\""}
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -576,6 +576,10 @@ value: <tmpl_string>
 [ short: <boolean> | default = slack_config.short_fields ]
 ```
 
+## `<matcher>`
+
+A matcher is a string with a syntax inspired by PromQL label matchers.
+
 ## `<opsgenie_config>`
 
 OpsGenie notifications are sent via the [OpsGenie API](https://docs.opsgenie.com/docs/alert-api).
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -146,19 +146,19 @@ current node.
 # Whether an alert should continue matching subsequent sibling nodes.
 [ continue: <boolean> | default = false ]
 
+# DEPRECATED: Use matchers below.
 # A set of equality matchers an alert has to fulfill to match the node.
 match:
   [ <labelname>: <labelvalue>, ... ]
 
+# DEPRECATED: Use matchers below.
 # A set of regex-matchers an alert has to fulfill to match the node.
 match_re:
   [ <labelname>: <regex>, ... ]
   
-# A combined set of equality and regex with support of negative matching 
-# that an alert has to fulfill to match the node.  
+# A list of matchers that an alert has to fulfill to match the node. 
 matchers:
-  [ <labelname>: <labelvalue>, <labelname>: <regex>, ... ]
-
+  [ - <matcher> ... ]
 
 # How long to initially wait to send a notification for a group
 # of alerts. Allows to wait for an inhibiting alert to arrive or collect
@@ -226,28 +226,32 @@ source matchers in a way that alerts never match both sides. It is much easier
 to reason about and does not trigger this special case.
 
 ```yaml
+# DEPRECATED: Use matchers below.
 # Matchers that have to be fulfilled in the alerts to be muted.
 target_match:
   [ <labelname>: <labelvalue>, ... ]
+# DEPRECATED: Use matchers below.
 target_match_re:
   [ <labelname>: <regex>, ... ]
   
-# Matchers that have to be fulfilled in the alerts to be muted 
-# with support of negative matching. 
+# A list of matchers that have to be fulfilled by the target 
+# alerts to be muted.
 target_matchers:
-  [ <labelname>: <labelvalue>, <labelname>: <regex>, ... ]
+  [ - <matcher> ... ]
 
+# DEPRECATED: Use matchers below.
 # Matchers for which one or more alerts have to exist for the
 # inhibition to take effect.
 source_match:
   [ <labelname>: <labelvalue>, ... ]
+# DEPRECATED: Use matchers below.
 source_match_re:
   [ <labelname>: <regex>, ... ]
   
-# Matchers for which one or more alerts have to exist for the
-# inhibition to take effect with support of negative matching.
+# A list of matchers for which one or more alerts have 
+# to exist for the inhibition to take effect.
 source_matchers:
-  [ <labelname>: <labelvalue>, <labelname>: <regex>, ... ]
+   [ - <matcher> ... ]
 
 # Labels that must have an equal value in the source and target
 # alert for the inhibition to take effect.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -226,11 +226,11 @@ source matchers in a way that alerts never match both sides. It is much easier
 to reason about and does not trigger this special case.
 
 ```yaml
-# DEPRECATED: Use target matchers below.
+# DEPRECATED: Use target_matchers below.
 # Matchers that have to be fulfilled in the alerts to be muted.
 target_match:
   [ <labelname>: <labelvalue>, ... ]
-# DEPRECATED: Use target matchers below.
+# DEPRECATED: Use target_matchers below.
 target_match_re:
   [ <labelname>: <regex>, ... ]
   
@@ -239,12 +239,12 @@ target_match_re:
 target_matchers:
   [ - <matcher> ... ]
 
-# DEPRECATED: Use source matchers below.
+# DEPRECATED: Use source_matchers below.
 # Matchers for which one or more alerts have to exist for the
 # inhibition to take effect.
 source_match:
   [ <labelname>: <labelvalue>, ... ]
-# DEPRECATED: Use source matchers below.
+# DEPRECATED: Use source_matchers below.
 source_match_re:
   [ <labelname>: <regex>, ... ]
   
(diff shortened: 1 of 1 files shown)


aSquare14:
cc @beorn7

beorn7:
Note: failing test_frontend is a known issue. If you rebase this, it should go away.

beorn7:
I hope I can review this tomorrow.

aSquare14:
> Note: failing test_frontend is a known issue. If you rebase this, it should go away.

I've rebased, could you rerun the test again.

docs/configuration.md
```diff
@@ -153,6 +153,12 @@ match:
 # A set of regex-matchers an alert has to fulfill to match the node.
 match_re:
   [ <labelname>: <regex>, ... ]
+  
+# A combined set of equality and regex with support of negative matching 
+# that an alert has to fulfill to match the node.  
+matchers:
+  [ <labelname>: <labelvalue>, <labelname>: <regex>, ... ]
```
beorn7, line 160: That's actually not the syntax. From the YAML perspective, `matchers` is just a list of strings. So it should simply be:

```
  [ - <matcher> ... ]
```

(See how it is done below for `routes:`.)

And then we need a separate section below that looks like:

```
## `<matcher>`

A matcher is a string with a syntax inspired by PromQL label matchers.
...
```

Then you have to explain the syntax in detail, which you can compile from https://github.com/prometheus/alertmanager/blob/master/pkg/labels/parse.go#L38-L56 and https://github.com/prometheus/alertmanager/blob/master/pkg/labels/parse.go#L38-L56.

You should probably provide plenty of examples how things look in YAML, including how a single `matcher` can contain many of the matching operations, i.e.
`'{foo = "bar", dings != "bums", }'` but that you can also list them individually as separate list items, e.g.

```YAML
  matchers:
    - '{foo = "bar", dings != "bums", }'
```

is equivalent to

```YAML
  matchers:
    - foo = "bar"
    - dings != "bums"
```

docs/configuration.md
```diff
@@ -225,13 +231,23 @@ target_match:
   [ <labelname>: <labelvalue>, ... ]
 target_match_re:
   [ <labelname>: <regex>, ... ]
+  
+# Matchers that have to be fulfilled in the alerts to be muted 
+# with support of negative matching.
```
beorn7, line 236: As above: "A list of matchers that have to be fulfilled by the target alerts to be muted."

docs/configuration.md
```diff
@@ -225,13 +231,23 @@ target_match:
   [ <labelname>: <labelvalue>, ... ]
 target_match_re:
   [ <labelname>: <regex>, ... ]
+  
+# Matchers that have to be fulfilled in the alerts to be muted 
+# with support of negative matching. 
+target_matchers:
+  [ <labelname>: <labelvalue>, <labelname>: <regex>, ... ]
```
beorn7, line 238: As above, the syntax is:
```
  [ - <matcher> ... ]
```

docs/configuration.md
```diff
@@ -225,13 +231,23 @@ target_match:
...
 # inhibition to take effect.
 source_match:
   [ <labelname>: <labelvalue>, ... ]
 source_match_re:
   [ <labelname>: <regex>, ... ]
+  
+# Matchers for which one or more alerts have to exist for the
+# inhibition to take effect with support of negative matching.
+source_matchers:
+  [ <labelname>: <labelvalue>, <labelname>: <regex>, ... ]
```
beorn7, line 250: As above, the syntax is:
```
  [ - <matcher> ... ]
```

docs/configuration.md
```diff
@@ -225,13 +231,23 @@ target_match:
...
 
 # Matchers for which one or more alerts have to exist for the
 # inhibition to take effect.
 source_match:
   [ <labelname>: <labelvalue>, ... ]
 source_match_re:
   [ <labelname>: <regex>, ... ]
+  
+# Matchers for which one or more alerts have to exist for the
+# inhibition to take effect with support of negative matching.
```
beorn7, line 248: As above: "A list of matchers for which one or more alerts have to exist for the inhibition to take effect."

docs/configuration.md
```diff
@@ -225,13 +231,23 @@ target_match:
...
 target_match_re:
   [ <labelname>: <regex>, ... ]
+  
+# Matchers that have to be fulfilled in the alerts to be muted 
+# with support of negative matching. 
+target_matchers:
+  [ <labelname>: <labelvalue>, <labelname>: <regex>, ... ]
 
 # Matchers for which one or more alerts have to exist for the
 # inhibition to take effect.
```
beorn7, line 241: Needs a deprecation note as above.

docs/configuration.md
```diff
@@ -225,13 +231,23 @@ target_match:
   [ <labelname>: <labelvalue>, ... ]
 target_match_re:
   [ <labelname>: <regex>, ... ]
```
beorn7, line 233: Needs a deprecation note as above.

beorn7 (changes requested):
Finally, I got to review this.

I think we really need to explain carefully the PromQL-style syntax for the new matchers. See comments, i.e. we need a separate section for that.

The example section needs adjustments, it still contains the old-style matching in lines 200-201 and 207-208.

docs/configuration.md
```diff
@@ -153,6 +153,12 @@ match:
 # A set of regex-matchers an alert has to fulfill to match the node.
```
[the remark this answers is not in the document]
aSquare14: Done

aSquare14:
@beorn7 Sorry for the delay. Have resumed work on this.

beorn7:
Great, thanks.

Is this ready for the next round of review, or are you still working on it?

aSquare14:
Have made changes, but some part is still left. Am working on it.

docs/configuration.md
beorn7: Note: I updated the source code links above. (Previously, I gave the same link twice.)

aSquare14: Thank you. Will update.

aSquare14: > which you can compile from

@beorn7 What does this mean ? Could you elaborate a bit

beorn7: Sorry for using a heavily overloaded word. (o:

That's "compile" in the old sense (before there were compilers): Take the two doc comments I linked above and, based on what's written there, create documentation about the config file (as described above).

(Remember that those that write the config probably aren't aware of the godoc documentation.)

aSquare14: Okay got it :) Updated the PR, sorry for the delay. Please feel free to take a look whenever you get some time :)

docs/configuration.md
```diff
@@ -556,6 +576,28 @@ value: <tmpl_string>
 [ short: <boolean> | default = slack_config.short_fields ]
 ```
 
+## `<matcher>`
+
+A matcher is a string with a syntax inspired by PromQL and Openmetrics. The syntax of a matcher consists of three tokens: 
+- A valid Prometheus label name. 
+- One of '=', '!=', '=~', or '!~', with the same meaning as known from PromQL selectors.
```
beorn7, line 583: Perhaps use backticks rather than single quotes to render it nicely in Markdown?

docs/configuration.md
```diff
@@ -556,6 +576,28 @@ value: <tmpl_string>
...
 ```
 
+## `<matcher>`
+
+A matcher is a string with a syntax inspired by PromQL and Openmetrics. The syntax of a matcher consists of three tokens: 
+- A valid Prometheus label name. 
+- One of '=', '!=', '=~', or '!~', with the same meaning as known from PromQL selectors. 
+- A UTF-8 string, which may be enclosed in double quotes. Before or after each token, there may be any amount of whitespace. 
+
+The 3rd token may be the empty string. Within the 3rd token, OpenMetrics escaping rules apply: '\"' for a double-quote, '\n' for a line feed, '\\' for a literal backslash. Unescaped '"' must not occur inside the 3rd token (only as the 1st or last character). However, literal line feed characters are tolerated, as are single '\' characters not followed by '\', 'n', or '"'. They act as a literal backslash in that case.
```
beorn7, line 586: Again consider backticks rather than single quotes.

docs/configuration.md
```diff
@@ -556,6 +576,28 @@ value: <tmpl_string>
...
+
+A comma separated list of matcher forms matchers. 
+For matchers, a leading '{' and/or a trailing '}' is optional and will be trimmed before further parsing. Individual Matchers are separated by commas outside of quoted parts of the input string. Those commas may be surrounded by whitespace. Parts of the string inside unescaped double quotes ('"…"') are considered quoted (and commas don't act as separators there). If double quotes are escaped with a single backslash ('\"'), they are ignored for the purpose of identifying quoted parts of the input string. If the input string, after trimming the optional trailing '}', ends with a comma, followed by optional whitespace, this comma and whitespace will be trimmed.
```
beorn7, line 589: Hmm, it's kind of more nuanced, because you could use a Yaml list or just multiple matchers all in one string.

How about the following:

In the configuration, multiple matchers are combined in a YAML list. However, it is also possible to combine multiple matchers within a single YAML string, again using syntax inspired by PromQL. In such a string, a leading `{` and/or a trailing `}` is optional and will be trimmed before further parsing. Individual matchers are separated by commas outside of quoted parts of the string. Those commas may be surrounded by whitespace. Parts of the string inside unescaped double quotes (`"…"`) are considered quoted (and commas don't act as separators there). If double quotes are escaped with a single backslash (`\"`), they are ignored for the purpose of identifying quoted parts of the input string. If the input string, after trimming the optional trailing `}`, ends with a comma, followed by optional whitespace, this comma and whitespace will be trimmed.

docs/configuration.md
```diff
@@ -556,6 +576,28 @@ value: <tmpl_string>
...
+
+Examples for valid input strings:
```
beorn7, line 591: Suggestion:

All of the following is therefore valid:

docs/configuration.md
beorn7: And perhaps give a short explanation what the operators mean. Not every person writing Alertmanager configs knows PromQL.

beorn7 (changes requested):
Thanks for following up.

It looks quite good but I think we need to write even a bit more. I have added some suggestions.

docs/configuration.md
aSquare14: Looks much better. Have updated it 👍

docs/configuration.md
aSquare14: Done

docs/configuration.md
aSquare14: Done

docs/configuration.md
aSquare14: Done

beorn7:
Sorry for delay. Will hopefully get to this today.

aSquare14:
No worries :)

docs/configuration.md
```diff
@@ -556,6 +576,30 @@ value: <tmpl_string>
...
+
+In the configuration, multiple matchers are combined in a YAML list. However, it is also possible to combine multiple matchers within a single YAML string, again using syntax inspired by PromQL. In such a string, a leading { and/or a trailing } is optional and will be trimmed before further parsing. Individual matchers are separated by commas outside of quoted parts of the string. Those commas may be surrounded by whitespace. Parts of the string inside unescaped double quotes `"…"` are considered quoted (and commas don't act as separators there). If double quotes are escaped with a single backslash `\`, they are ignored for the purpose of identifying quoted parts of the input string. If the input string, after trimming the optional trailing }, ends with a comma, followed by optional whitespace, this comma and whitespace will be trimmed.
```
beorn7, line 591: I'd put the `{` and `}` in backticks, similar to the quoted characters in the previous paragraph.

docs/configuration.md
```diff
@@ -556,6 +576,30 @@ value: <tmpl_string>
...
+
+All of the following is therefore valid:
+
+``` yaml
+matchers: [ '{foo = "bar", dings != "bums", }' ]
+matchers: [ foo=bar,dings!=bums ]
+matchers: [ foo=bar, dings!=bums ]
+matchers: [ '{quote="She said: \"Hi, ladies! That's gender-neutral…\""}' ]
+matchers: [ statuscode=~"5.." ]
+```
```
beorn7, line 601: Oh noes. I remember I wrote a whole lot about this section with a lot of examples, that I all tried out, but somehow GitHub has eaten it.

 The problem here is that a short-form list in Yaml uses other string quoting than a long-form list. Also, the above is technically not valid Yaml because it reuses the same key "matchers" in a map. I think we should have several separate examples and explanations in between.

I'll try to recreate what I wrote last week:

Two matchers combined in a long-form YAML list:

```yaml
matchers:
  - foo = bar
  - dings != bums
```

The same in a short-form YAML list:

```yaml
matchers: [ foo = bar, dings != bums ]
```

In the short-form, it's generally better to quote the list elements to avoid problems with special characters like commas:

```yaml
matchers: [ "foo = bar,baz", "dings != bums" ]
```

Or you put both matchers into one PromQL-like string. Single quotes for the whole string work best here:

```yaml
matchers: [ '{foo="bar,baz", dings!="bums"}' ]
```

To avoid any confusion about YAML string quoting and escaping, you can use YAML block quoting and then only worry about the OpenMetrics escaping inside the block. A complex example with a regular expression and different quotes inside the label value:

```yaml
matchers:
  - |
    {quote=~"She said: \"Hi, ladies!( That's gender-neutral…)?\""}
```

beorn7 (changes requested):
Some more nits, and then one large comment of mine got deleted in my last review. I tried to recreate it.

docs/configuration.md
aSquare14: done

docs/configuration.md
aSquare14: This sound good. Updated it.

aSquare14:
@beorn7 Sorry for the delay on this. Have updated it. Please take a look again. And this time will try completing it soon :)

docs/configuration.md
```diff
@@ -556,6 +576,57 @@ value: <tmpl_string>
...
+2. Similar to example 1, shown below are two equality matchers combined in a short form YAML list.
+
+```yaml
+matchers: [ "foo = bar", "dings != bums" ]
+```
+
+As shown below, in the short-form, it's generally better to quote the list elements to avoid problems with special characters like commas:
+
+```yaml
+matchers: [ '{foo="bar",dings!="bums"}' ]
```
beorn7, line 612: This is one single PromQL-like string, with single quotes all around it. In other words, this example should be the one for the next point (3) below.

Here, I'd suggest the following example (where there is actually a comma in one of the label values):

```yaml
matchers: [ "foo = bar,baz", "dings != bums" ]
```

docs/configuration.md
```diff
@@ -556,6 +576,57 @@ value: <tmpl_string>
...
+```yaml
+  matchers :
+   - foo = bar
+   - dings !=bums 
+```
+
+2. Similar to example 1, shown below are two equality matchers combined in a short form YAML list.
+
+```yaml
+matchers: [ "foo = bar", "dings != bums" ]
```
beorn7, line 606: This has already quoted list elements. I would use no quotes here so that the difference to the next example becomes more obvious:

```yaml
matchers: [ foo = bar, dings != bums ]
```

docs/configuration.md
```diff
@@ -556,6 +576,57 @@ value: <tmpl_string>
...
+As shown below, in the short-form, it's generally better to quote the list elements to avoid problems with special characters like commas:
+
+```yaml
+matchers: [ '{foo="bar",dings!="bums"}' ]
+```
+
+3. You can also put both matchers into one PromQL-like string. Single quotes for the whole string work best here.
+
+```yaml
+matchers: ['{env=~"produ.*"}', '{baz!~".*quux"}']
```
beorn7, line 618: This is the wrong example. As said above, line 612 should go here.

This example is also a good one, but it demonstrates having two PromQL-like strings in separate Yaml list elements (which is perfectly valid). And it demonstrates regexp matchers. But you have a regexp in example 4 below. I guess you could just remove this particular example. Or keep it, but then you need to give it its own explanation text (that actually matches what the example demonstrates).

beorn7 (changes requested):
I think you got the examples a bit shuffled.

docs/configuration.md
aSquare14: Done

docs/configuration.md
aSquare14: Done

aSquare14:
> I think you got the examples a bit shuffled.

Yes, I did. Sorry about it. Have updated it.

beorn7 (approved):
We made it! (o:

aSquare14:
Finally ! :) Thank you so much for the reviews and patience.

PR #2482, opened 2021-02-09
ui: update versions for the Elm components

simonpasquier:
The frontend CI job started to fail for some obscure reasons, most
probably because we use a very old version of NodeJS.

To make it work again, this change does the following:

* Bump to the latest NodeJS LTS version and use the buster variant.
* Use Yarn instead of npm to install Javascript packages.
* Bump Elm components to their latest versions.

There should be no functional change to the UI code.

Signed-off-by: Simon Pasquier <spasquie@redhat.com>


w0rm (approved):
Thanks for taking care of this!

beorn7 (approved):
Thanks for successfully troubleshooting this.

roidelapluie (approved):
Thanks!

PR #2484, opened 2021-02-09
Catch unknown matcher types in gossipped silences

beorn7:
This has been discussed in #2479. Even if the conclusion there was
that we don't need this in a bugfix release, it's still better to have
this kind of robustness. So this introduces the same check into the
main branch.

@vladimiroff @aSquare14


aSquare14 (approved):

simonpasquier (approved):
thanks!

PR #2487, opened 2021-02-19
Clustering: Interface for Peers in other packages

gotjosh:
A Peer as defined by the `cluster` package represents the node in the
cluster. It is used in other packages to know the status of all of the
members or how long should we wait to know if a notification has already fired.

In Cortex, we'd like to implement a slightly different way of
clustering (using gRPC for communication and a
hash ring for node discovery).

This is a small change to support that by changing the consumer of other
packages to an interface.

~~Silences and Notification channels don't need an interface as they take
a `func([]byte) error` as a parameter.~~

Edit: Ended up creating an interface for it as I think it's somewhat useful.

Signed-off-by: gotjosh <josue@grafana.com>

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -33,22 +33,20 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
-// ClusterPeer helps consumers gather information about the peer(s) in the cluster
+// Peer is a single peer in a gossip cluster.
 type ClusterPeer interface {
 	// Name returns the unique identifier of this peer in the cluster.
 	Name() string
-	// Address returns the IP address of this peer in the cluster.
-	Address() string
 	// Status returns a status string representing the peer state.
 	Status() string
 	// Peers returns the peer nodes in the cluster.
-	Peers() []Node
+	Peers() []ClusterMember
 }
 
-// Node interface that represents node peers in a cluster
-type Node interface {
-	// String returns the name of the node
-	String() string
+// ClusterMember interface that represents node peers in a cluster
+type ClusterMember interface {
+	// Name returns the name of the node
+	Name() string
 	// Address returns the IP address of the node
 	Address() string
 }
@@ -565,10 +563,6 @@ func (p *Peer) Name() string {
 	return p.mlist.LocalNode().Name
 }
 
-func (p *Peer) Address() string {
-	return p.mlist.LocalNode().Addr.String()
-}
-
 // ClusterSize returns the current number of alive members in the cluster.
 func (p *Peer) ClusterSize() int {
 	return p.mlist.NumMembers()
@@ -615,11 +609,22 @@ func (p *Peer) Self() *memberlist.Node {
 	return p.mlist.LocalNode()
 }
 
+// Member represents a member in the cluster.
+type Member struct {
+	*memberlist.Node
+}
+// Name implements cluster.ClusterMember
+func (m Member) Name() string { return m.Node.Name }
+// Address implements cluster.ClusterMember
+func (m Member) Address() string { return m.Node.Address() }
+
 // Peers returns the peers in the cluster.
-func (p *Peer) Peers() []Node {
-	peers := make([]Node, 0, len(p.mlist.Members()))
+func (p *Peer) Peers() []ClusterMember {
+	peers := make([]ClusterMember, 0, len(p.mlist.Members()))
 	for _, member := range p.mlist.Members() {
-		peers = append(peers, member)
+		peers = append(peers, Member{
+			Node: member,
+		})
 	}
 	return peers
 }
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -613,8 +613,10 @@ func (p *Peer) Self() *memberlist.Node {
 type Member struct {
 	*memberlist.Node
 }
+
 // Name implements cluster.ClusterMember
 func (m Member) Name() string { return m.Node.Name }
+
 // Address implements cluster.ClusterMember
 func (m Member) Address() string { return m.Node.Address() }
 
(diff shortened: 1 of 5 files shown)


cluster/cluster.go
```diff
@@ -33,6 +33,26 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+// ClusterPeer helps consumers gather information about the peer(s) in the cluster
+type ClusterPeer interface {
```
gotjosh, line 37: I had to place this in the cluster package as both API `v1` and `v2` consume this interface. Once we remove the v1 code I'll move to where it's used in the API package.

gotjosh: Also, I'm happy to take suggestions on a better name.

cluster/cluster.go
```diff
@@ -33,6 +33,26 @@ import (
...
+	// Status returns a status string representing the peer state.
+	Status() string
+	// Peers returns the peer nodes in the cluster.
+	Peers() []Node
+}
+
+// Node interface that represents node peers in a cluster
+type Node interface {
+	// String returns the name of the node
+	String() string
```
pracucci, line 51: I would call it `Name()` for clarity. You would have to wrap `*memberlist.Node` to do it.

cluster/cluster.go
```diff
@@ -33,6 +33,26 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+// ClusterPeer helps consumers gather information about the peer(s) in the cluster
```
pracucci, line 36:
```suggestion
// Peer is a single peer in a gossip cluster.
```

cluster/cluster.go
```diff
@@ -33,6 +33,26 @@ import (
...
+	// Address returns the IP address of this peer in the cluster.
+	Address() string
+	// Status returns a status string representing the peer state.
+	Status() string
+	// Peers returns the peer nodes in the cluster.
+	Peers() []Node
+}
+
+// Node interface that represents node peers in a cluster
+type Node interface {
```
pracucci, line 49: Maybe `ClusterMember` could be a better naming, given it represent a member of cluster?

cluster/cluster.go
```diff
@@ -545,6 +565,10 @@ func (p *Peer) Name() string {
 	return p.mlist.LocalNode().Name
 }
 
+func (p *Peer) Address() string {
```
pracucci, line 568: Who needs / uses this?

cluster/cluster.go
gotjosh: I like this, thanks!

cluster/cluster.go
gotjosh: This is used by the API to expose the node IP, but also internally to manage the peers as they join/leave/reconnect.

I have added a comment to signal what this returns.

gotjosh: Now I realise what you meant, the actual address is always used from the `Peers()` which returns the `Node` it is not needed in the base peer. Great catch!

pracucci (approved):
LGTM

cluster/cluster.go
```diff
@@ -591,14 +614,31 @@ func (p *Peer) Self() *memberlist.Node {
 	return p.mlist.LocalNode()
 }
 
+// Member represents a member in the cluster.
+type Member struct {
```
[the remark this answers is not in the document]
gotjosh: The linter complains about Exported functions that return unexported types. I thought this was not acceptable to a certain extend?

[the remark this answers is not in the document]
pracucci: You can just give a name to `*memberlist.Node`, like:

`node *memberlist.Node`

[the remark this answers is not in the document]
gotjosh: oh, you meant specifically the inner attribute - that makes sense.

cluster/cluster.go
```diff
@@ -33,6 +33,29 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 )
 
+// Peer is a single peer in a gossip cluster.
```
gouthamve, line 36: `Peer` --> `ClusterPeer`

cluster/cluster.go
```diff
@@ -591,14 +614,31 @@ func (p *Peer) Self() *memberlist.Node {
...
+			node: member,
+		})
+	}
+	return peers
 }
 
 // Position returns the position of the peer in the cluster.
 func (p *Peer) Position() int {
-	all := p.Peers()
+	all := p.mlist.Members()
```
gouthamve, line 641: Why are we not using `p.Peers()` here?

cluster/cluster.go
```diff
@@ -512,15 +535,15 @@ func (p *Peer) peerUpdate(n *memberlist.Node) {
...
 // broadcast messages for the state can be sent.
-func (p *Peer) AddState(key string, s State, reg prometheus.Registerer) *Channel {
+func (p *Peer) AddState(key string, s State, reg prometheus.Registerer) ClusterChannel {
 	p.states[key] = s
 	send := func(b []byte) {
 		p.delegate.bcast.QueueBroadcast(simpleBroadcast(b))
 	}
 	peers := func() []*memberlist.Node {
-		nodes := p.Peers()
+		nodes := p.mlist.Members()
```
gouthamve, line 544: Same here, when to use `Peers` and when to use `mlist.Members`?

gouthamve (approved):
LGTM with nits!

cluster/cluster.go
gotjosh: Does the above make sense?

gouthamve: YUSSS

cluster/cluster.go
gotjosh: Is this better?

PR #2488, opened 2021-02-20
Negative matchers on the Silence form page

### LINKED ISSUE #1682: Allow negative (re) matching in silences

w0rm:
Closes #1682

This implements the ability to create silences with negative matchers using the UI element from the alerts page.

Screenshot 2021-02-20 at 13 26 58

Signed-off-by: w0rm unsoundscapes@gmail.com


beorn7:
Great. Will have a closer look ASAP.

beorn7:
Awesome. I cannot judge the Elm code, but I played with it, and it does what it should do. Let's see if @simonpasquier or anyone with Elm foo wants to chime in (for another day or so), but otherwise, I think we should just merge this.

beorn7 (approved):

beorn7:
@vladimiroff you might be interested in this, too.

vladimiroff (approved):
Looking good. Also, edit and recreate forms don't forget `isEqual`, which is something I've spent a lot more time, that I am willing to admit. Extra points for the proper cleanup in `SilenceForm/Types.elm`, taught me a lot.

PR #2509, opened 2021-03-09
Clustering: Change WaitReady to accept a Context.

stevesg:
WaitReady is a blocking call and so should accept a Context in order to
be responsive to cancellation of the notification pipeline for any reason.

Signed-off-by: Steve Simpson <steve.simpson@grafana.com>

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -44,7 +44,7 @@ type ResolvedSender interface {
 // Peer represents the cluster node from where we are the sending the notification.
 type Peer interface {
 	// WaitReady waits until the node silences and notifications have settled before attempting to send a notification.
-	WaitReady()
+	WaitReady(context.Context) error
 }
 
 // MinTimeout is the minimum timeout that is set for the context of a call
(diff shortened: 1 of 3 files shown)


pracucci (approved):
LGTM! This would be helpful in Cortex.

gotjosh (approved):

gouthamve (approved):

notify/notify.go
```diff
@@ -44,7 +44,7 @@ type ResolvedSender interface {
 // Peer represents the cluster node from where we are the sending the notification.
 type Peer interface {
 	// WaitReady waits until the node silences and notifications have settled before attempting to send a notification.
-	WaitReady()
+	WaitReady(context.Context)
```
pstibrany, line 47: When accepting context, I would expect that method also returns `error`, so that client calling it knows whether method has succeded or not.

roidelapluie:
Thanks!

I agree with @pstibrany 's comment

notify/notify.go
stevesg: Also added a test for the non-blocking path.

PR #2511, opened 2021-03-10
Replace satori/go.uuid with gofrs/uuid

codesome:
github.com/satori/go.uuid seems to have some fixes/enhancement since it's last tagged release. And it has been a while since it had any commit into master (last was Oct 2018), so I am not expecting a tagged release to be out anytime soon.

So I am proposing we upgrade it to master (it is also blocking other upgrades in grafana/grafana).

EDIT: replaced the package based on suggestion here https://github.com/prometheus/alertmanager/pull/2511#issuecomment-795419514


gouthamve (approved):
LGTM!

beorn7 (approved):
OK as band aid. But preferably, we avoid those hacks.

Have you tried to ping the maintainers of the upstream repository about tagging a version?

roidelapluie:
We could also move to a maintained fork, https://github.com/gofrs/uuid

beorn7:
That sounds like a plan!

codesome:
I have now replaced satori/go.uuid with gofrs/uuid, PTAL

beorn7 (approved):
And wow, TIL that github.com/gofrs/uuid has reverted their Go Modules support ("until Go Modules is more mature").

PR #2512, opened 2021-03-10
Update go-openapi to latest

gouthamve:
go-swagger 0.25.0+ are failing though :/

Did this because importing AM downstream requires us to `replace go-openapi` components, thought I'd fix that. But holy wow, this is a massive diff, not sure if we want to do this now.


beorn7 (approved):
From my essentially not existent knowledge about Swagger, this looks good to me. (In other words: It would be better if someone more qualified could also review this.)

simonpasquier:
:+1: will take a look early next week but if tests are green, it's probably fine :)

go.mod
```diff
@@ -2,36 +2,50 @@ module github.com/prometheus/alertmanager
 
 require (
 	github.com/alecthomas/units v0.0.0-20190924025748-f65c72e2690d
-	github.com/cenkalti/backoff/v4 v4.0.2
+	github.com/armon/go-metrics v0.3.6 // indirect
```
simonpasquier, line 5: Could you remove all the `// indirect` dependencies? I don't think it makes sense to bump them explicitly...

simonpasquier (commented):
The changes look good to me, expect one remark on `go.mod`.

go.mod
gouthamve: Done! Ah they were automatically added, but removing them nuked a lot of vendored code! Nice!

Issue #2493, opened 2021-02-26
Alertmanager/Opsgenie: AM dosen't update alerts fileds (message and description) with the new informations


HamidEDD:
**What did you do?**
I have created a group of alerts based on the \_\_name\_\_ with the rule bellow : 
 
```
groups:
- name: Monitoring Probes 
  rules:
  - alert: "Test"
    description: "Test Probes"
    expr: "label_replace({__name__=~'test_.+',job='node-exporter'} >= 290,'metric', '$1', '__name__', '(.*)')"
    labels:
      alert_type: "alerting"
      severity: 'P1'
    annotations:
      summary: "{{ $labels.metric }} = {{ $value }}"
```

In the Alertmanager side, I selected to group_by (instance,alertname,severity) with this configuration:
```
global:
 opsgenie_api_url: 'https://api.eu.opsgenie.com/'
route:
 group_by: ['alertname','instance','severity']
 group_wait: 1m
 group_interval: 2m
 repeat_interval: 5m
 routes:
  - match:
      alert_type: 'alerting'
    receiver: opsgenie
receivers:
- name: opsgenie
  opsgenie_configs:
  - send_resolved: true
    source: 'Alertmanager'
    message: '{{ .GroupLabels.alertname }} {{ .GroupLabels.instance }} {{if eq .GroupLabels.alertname "Test"}}{{ range .Alerts }}{{ .Labels.metric }},{{ end }}{{ end }}'
    description: "Alerts:\n{{ range .Alerts }}{{ .Annotations.summary }}\n{{ end }}"
    priority: '{{ .GroupLabels.severity }}'
    responders:
    - name: 'alertmanager-responder'
      type: 'team'
```

**What did you expect to see?**
When Test rings (for example: test_service1, test_service2), the alerts arrive within 1m (goup_wait) and everything works fine.
Alertmanager notifies opsgenie, the message and description fields contain the correct information

`message :  Test   host1.localhost.local   test_service1,test_service2`

**What did you see instead? Under which circumstances?**
But when another Test alert (test_service3) arrives later (after 3m), the message/description in opsgenie seem to always have the first value: 
`message :  Test   host1.localhost.local   test_service1,test_service2`
instead of : 
`message :  Test   host1.localhost.local   test_service1,test_service2,test_service3`

it seems that Alertmanager does not update the message and description fields on the opsgenie side, when an alert is updated.
I reproduce the same issue with several different alert rules and with a group_by (instance, severity)

**Environment**

[...]

PR #2519, opened 2021-03-19
Add the option to update message and description when sending alerts to opsgenie

tomasfreund:
#2493

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -843,13 +843,9 @@ responders:
 # Priority level of alert. Possible values are P1, P2, P3, P4, and P5.
 [ priority: <tmpl_string> ]
 
-# Whether or not to send a request to update the alert message every time an alert is sent to OpsGenie
-# By default, the message of the alert is never updated in OpsGenie, the new message only appears in activity log.
-[ update_message: <boolean> | default = false ]
-
-# Whether or not to send a request to update alert description every time an alert is sent to OpsGenie
-# By default, the description of the alert is never updated in OpsGenie
-[ update_description: <boolean> | default = false ]
+# Whether or not to update message and description of the alert in OpsGenie if it already exists
+# By default, the alert is never updated in OpsGenie, the new message only appears in activity log.
+[ update_alerts: <boolean> | default = false ]
 
 # The HTTP client's configuration.
 [ http_config: <http_config> | default = global.http_config ]
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -843,6 +843,14 @@ responders:
 # Priority level of alert. Possible values are P1, P2, P3, P4, and P5.
 [ priority: <tmpl_string> ]
 
+# Whether or not to send a request to update alert message every time every time an alert is sent to OpsGenie
+# By default, the message of the alert is never updated in OpsGenie, the new message only appears in activity log
+[ update_message: <boolean> | default = false ]
+
+# Whether or not to send a request to update alert description every time every time an alert is sent to OpsGenie
+# By default, the description of the alert is never updated in OpsGenie
+[ update_description: <boolean> | default = false ]
+
 # The HTTP client's configuration.
 [ http_config: <http_config> | default = global.http_config ]
 ```
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -843,8 +843,8 @@ responders:
 # Priority level of alert. Possible values are P1, P2, P3, P4, and P5.
 [ priority: <tmpl_string> ]
 
-# Whether or not to send a request to update alert message every time every time an alert is sent to OpsGenie
-# By default, the message of the alert is never updated in OpsGenie, the new message only appears in activity log
+# Whether or not to send a request to update the alert message every time an alert is sent to OpsGenie
+# By default, the message of the alert is never updated in OpsGenie, the new message only appears in activity log.
 [ update_message: <boolean> | default = false ]
 
 # Whether or not to send a request to update alert description every time every time an alert is sent to OpsGenie
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -847,7 +847,7 @@ responders:
 # By default, the message of the alert is never updated in OpsGenie, the new message only appears in activity log.
 [ update_message: <boolean> | default = false ]
 
-# Whether or not to send a request to update alert description every time every time an alert is sent to OpsGenie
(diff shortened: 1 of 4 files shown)


HamidEDD (commented):
LGTM

dgsb:
any plan to merge that soon ?

docs/configuration.md
```diff
@@ -678,6 +678,14 @@ responders:
 # Priority level of alert. Possible values are P1, P2, P3, P4, and P5.
 [ priority: <tmpl_string> ]
 
+# Whether or not to send a request to update alert message every time every time an alert is sent to OpsGenie
```
roidelapluie, line 681:
```suggestion
# Whether or not to send a request to update the alert message every time an alert is sent to OpsGenie
```

docs/configuration.md
```diff
@@ -678,6 +678,14 @@ responders:
...
+# By default, the message of the alert is never updated in OpsGenie, the new message only appears in activity log
```
roidelapluie, line 682:
```suggestion
# By default, the message of the alert is never updated in OpsGenie, the new message only appears in activity log.
```

docs/configuration.md
```diff
@@ -678,6 +678,14 @@ responders:
...
+[ update_message: <boolean> | default = false ]
+
+# Whether or not to send a request to update alert description every time every time an alert is sent to OpsGenie
+# By default, the description of the alert is never updated in OpsGenie
+[ update_description: <boolean> | default = false ]
```
roidelapluie, line 687: Can we have maybe a single boolean for both?

notify/opsgenie/opsgenie.go
```diff
@@ -180,6 +204,55 @@ func (n *Notifier) createRequest(ctx context.Context, as ...*types.Alert) (*http
 			Note:        tmpl(n.conf.Note),
 			Priority:    tmpl(n.conf.Priority),
 		}
+		var buf bytes.Buffer
+		if err := json.NewEncoder(&buf).Encode(msg); err != nil {
+			return nil, false, err
+		}
+		req, err := http.NewRequest("POST", createEndpointURL.String(), &buf)
+		if err != nil {
+			return nil, true, err
```
roidelapluie, line 213: Can we detect here if an alert has been created or just grouped?

roidelapluie (commented):
Thanks! I have added a few comments/questions to start with.

notify/opsgenie/opsgenie.go
tomasfreund: unfortunately not, OpsGenie just returns 202 and processes the alert asynchronously

docs/configuration.md
tomasfreund: we could, I just don't know if everyone uses OpsGenie the same way, maybe not everybody cares about both message and description

alistarle: I personally agree with @roidelapluie, the functionnal need is to update or not the alert, not each field independently. 
Example If we want to update a new field in the future (maybe the severity idk), we don't want to add a third configuration option.

tomasfreund: I agree in principle but unfortunately the OpsGenie api needs a separate request for each one, if somebody is hitting api rate limits, they might want to only update some fields, updating all of them on every alert means 3x the request volume.

notify/opsgenie/opsgenie.go
tomasfreund: @roidelapluie can I resolve this thread?

docs/configuration.md
tomasfreund: @roidelapluie @alistarle If you think a single boolean is the way to go even when a request is needed to update each field, I don't really have a problem with it. Let me know so that I can update the PR

docs/configuration.md
```diff
@@ -678,6 +678,14 @@ responders:
 # Priority level of alert. Possible values are P1, P2, P3, P4, and P5.
 [ priority: <tmpl_string> ]
 
+# Whether or not to send a request to update the alert message every time an alert is sent to OpsGenie
+# By default, the message of the alert is never updated in OpsGenie, the new message only appears in activity log.
+[ update_message: <boolean> | default = false ]
+
+# Whether or not to send a request to update alert description every time every time an alert is sent to OpsGenie
```
alistarle, line 685:
```suggestion
# Whether or not to send a request to update alert description every time an alert is sent to OpsGenie
```

docs/configuration.md
alistarle: Indeed I understand this point, for me you can leave this as is, if it can make alertmanager more accessible for small business because it can save opsgenie API call

alistarle: Finally I think it is better to stick to the functionnal need, which is to just update the alert, because it is something who will never change, and we should avoid configuration change in the future.

We cannot know what will happen in the future regarding opsgenie API to allow bulk update, billing plan to allow more API call, customer who will upgrade to an higher plan... We will be able to update the code, but if it require config change it would be a pain for users.

tomasfreund: ok, I will update the PR to only use one option for both, do I call it `update_alerts` or just `update`?

tomasfreund: changed to a single `update_alerts` option

tomasfreund:
This PR should be ready to merge if nobody has any other comments. We've been running alertmanager with these changes in production and haven't noticed any issues. The alerts get updated as expected.

HamidEDD:
it has also been deployed in our production for about 2 months, work as expected and no issue to report !

rbizos:
Hello :wave: this features was requested by our users, it would be great to see it merged :)

roidelapluie:
Thanks, LGTM!

GustavJaner:
@tomasfreund thanks for this PR! However when I'm adding `update_alerts: true` to an existing working alertmanager config, the operator gives an error on the alertmanager config
- prometheus operator version: v0.56.0
- prometheus alertmanager version: v0.24.0

Any idea?
operator logs
```
rancher-monitoring-operator-6c58844c48-jvngr rancher-monitoring level=error ts=2022-05-06T08:09:49.626875745Z caller=klog.go:116 component=k8s_client_runtime func=ErrorDepth msg="Sync \"cattle-monitoring-system/rancher-monitoring-alertmanager\" failed: provision alertmanager configuration: failed to initialize from secret: yaml: unmarshal errors:\n  line 64: field update_alerts not found in type config.plain\n```

PR #2523, opened 2021-03-23
Export `newMetrics` function and `metrics` struct

codesome:
We have `NewRetryStage` method which takes this `metrics` struct, and given both the struct and the method to create it is unexported, it is impossible to use vendored `NewRetryStage`. We would like to use it in Grafana. Can we export them?

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -231,7 +231,7 @@ type NotificationLog interface {
 	Query(params ...nflog.QueryParam) ([]*nflogpb.Entry, error)
 }
 
-type metrics struct {
+type Metrics struct {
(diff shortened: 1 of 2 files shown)


gotjosh (changes requested):
I think we shouldn't need to pass a namespace, we can always use `prometheus.WrapRegistererWithPrefix`.

beorn7 (approved):
I think the whole metrics handling could be organized in a more streamlined fashion, which would also make it clearer what has to be exported and what not. But that would be more involved refactoring. I think this is fine as an incremental change.

gotjosh (approved):

Issue #2498, opened 2021-03-01
Generally enable reading secrets from files


beorn7:
This is the same as https://github.com/prometheus/prometheus/issues/8551 , just for Alertmanager.

roidelapluie:
Some fields we are missing:

smtp_auth_password_file
smtp_auth_secret_file
slack_api_url_file
victorops_api_key_file
opsgenie_api_key_file
wechat_api_secret_file

gr8Adakron:
When can we expect to have configuration for `victorops_api_key_file ` ??

beorn7:
I guess the implementation for all the other fields will follow the same pattern as #2534. PRs welcome. :smiley:

sinkingpoint:
Would anyone object to a general case of something like `api_key: file:/tmp/foo` to read the value for `api_key` from /tmp/foo? Seems like the most generic solution, if a bit "magic" but I'd be happy to PR it

beorn7:
That would collide with any secret that happens to start with `file:` (unlikely, but who knows…).

I guess including a "schema" from the beginning would have been the best solution (`file:/tmp/foo` vs `passwd:9euo9.y.3t3`).

We could switch to such a generic solution with AM, in principle, because we are still pre 1.x, but on the other hand, keeping things consistent with Prometheus (where we cannot change easily) has its value, too.

Just my random thoughts…

Duologic:
There are a few more not mentioned above:

- pagerduty.service_key_file
- pagerduty.routing_key_file
- pushover.user_key_file
- pushover.token_file

roidelapluie:
I like to have multiple keys. Having file: would just lead to more yaml issues for new users.

jkroepke:
I found a workaround for Opsgenie, Pagerduty and Pushover:

While coding https://github.com/prometheus/alertmanager/pull/2728 I detect a hidden functionality inside Alertmanager.

I saw this functionally for pagerduty and pushover

 * pagerduty.service_key_file
    https://github.com/prometheus/alertmanager/blob/70abcccedba5b111f9215e79f2842066db0ffbec/notify/pagerduty/pagerduty.go#L157
 * pagerduty.routing_key_file
https://github.com/prometheus/alertmanager/blob/70abcccedba5b111f9215e79f2842066db0ffbec/notify/pagerduty/pagerduty.go#L215
 * pushover.user_key_file
 * pushover.token_file
[...]

PR #2533, opened 2021-04-01
Support resolving environment variables in configuration

julienduchesne:
Super simple solution. Mostly opening this to get some feedback. Is such a solution good enough to ever be accepted?

Based on Kubernetes' resolution for container args: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#use-environment-variables-to-define-arguments
The expansion code is directly taken from there: https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/pkg/kubelet/container/helpers.go#L163

Probably fixes https://github.com/prometheus/alertmanager/issues/2498 partially

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -19,6 +19,7 @@ import (
 	"io/ioutil"
 	"net"
 	"net/url"
+	"os"
 	"path/filepath"
 	"regexp"
 	"sort"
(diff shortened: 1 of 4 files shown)


roidelapluie:
This is different than #2498. Expanding env variable is something that was considered in the past and not accepted.

Can I know what is your usecase?

julienduchesne:
Injecting Kubernetes secrets as environment variables to be used in alertmanager configurations

roidelapluie:
#2498 is about having secrets as files that can be read by alertmanager. Based on previous votes and even the last public prometheus developers summit, we are not allowing environment variable substitutions from the complete file.

I will add more details to #2498.

julienduchesne:
Alright, I did find the previous issues about it (should've looked better 😄 ). Thanks for answering

PR #2536, opened 2021-04-05
Add silence logging option

m-masataka:
Past silences are really important to postmortem our alert management operations.
We need to get "when" and "what kind of" create/edit/delete silences.
So this PR added optional function that logging silence change.

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -186,6 +186,7 @@ func run() int {
 		dataDir         = kingpin.Flag("storage.path", "Base path for data storage.").Default("data/").String()
 		retention       = kingpin.Flag("data.retention", "How long to keep data for.").Default("120h").Duration()
 		alertGCInterval = kingpin.Flag("alerts.gc-interval", "Interval between alert GC.").Default("30m").Duration()
+		silenceLogging  = kingpin.Flag("silences.logging", "Enable logging silences. If it is enabled, the status change of silence will be logged").Bool()
 
 		webConfig      = webflag.AddFlags(kingpin.CommandLine)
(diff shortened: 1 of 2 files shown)


cmd/alertmanager/main.go
```diff
@@ -186,6 +186,7 @@ func run() int {
 		dataDir         = kingpin.Flag("storage.path", "Base path for data storage.").Default("data/").String()
 		retention       = kingpin.Flag("data.retention", "How long to keep data for.").Default("120h").Duration()
 		alertGCInterval = kingpin.Flag("alerts.gc-interval", "Interval between alert GC.").Default("30m").Duration()
+		silenceLogging  = kingpin.Flag("silences.logging", "Enable logging silences. If it is enabled, the status change of silence will be logged").Bool()
```
SuperQ, line 189: This should match the existing log-related flags like `--log.level`, so how about `"log.silences"`?

SuperQ (commented):
Nice idea

roidelapluie:
Can we have a distinction between the silences gossipped and the ones created in the current instance?

cmd/alertmanager/main.go
anarcat: done in #4163

anarcat:
> Can we have a distinction between the silences gossipped and the ones created in the current instance?

could you expand on this? like tag them with a string of some sort, or disable logging for one or the other?

so far I haven't adressed this in #4163 but if it's the former, that should be fairly simple to do.

grobinson-grafana:
I'm going to close this in favor of #4163 as I believe this PR is stale.

PR #2540, opened 2021-04-11
[mixins] Alertmanager Overview dashboard

### LINKED ISSUE #2529: Add dashboards for Alertmanager's mixin
## Feature request

It would be lovely if Alertmanager's mixins could include two dashboards:

1. Alertmanager monitoring: To monitor Alertmanager itself, it would include metrics that are exposed by Alertmanager.
2. Alertmanager alerts: An informational dashboard that tells us what alerts are currently at the `firing` state, split by `severity`. For this one, I think it would be necessary to have [Alertmanager's Grafana Datasource plugin](https://grafana.com/grafana/plugins/camptocamp-prometheus-alertmanager-datasource/), so I'm not sure how viable it could be.

ArthurSens:
Fixes #2529


ArthurSens:
Hi @beorn7, apologies for the delay!

What do you think of the result? Anything you'd like me to change?

doc/alertmanager-mixin/.gitignore
```diff
@@ -0,0 +1,2 @@
+vendor
+dashboards_out
```
beorn7, line 2: EOL missing

doc/alertmanager-mixin/config.libsonnet
```diff
@@ -32,5 +32,13 @@
     // integration that is itself not used for critical alerts.
     // Example: @'pagerduty|webhook'
     alertmanagerCriticalIntegrationsRegEx: @'.*',
+
+
```
beorn7, line 36: Let's only have one blank line here.

doc/alertmanager-mixin/dashboards/overview.libsonnet
```diff
@@ -0,0 +1,310 @@
+local grafana = import 'github.com/grafana/grafonnet-lib/grafonnet/grafana.libsonnet';
+local dashboard = grafana.dashboard;
+local row = grafana.row;
+local prometheus = grafana.prometheus;
+local template = grafana.template;
+local graphPanel = grafana.graphPanel;
+
+
```
beorn7, line 8: You have a bunch of those double-blank-line gaps in this file. I could see a reason for them right here, but further down they seem fairly random. My apologies if there is an idea behind it that I didn't see. It's just a nit, but generally, I prefer minimizing vertical whitespace, i.e. sure, use a blank line for structuring, but _two_ blank lines probably doesn't really carry additional meaning for most, but takes away screen real estate.

doc/alertmanager-mixin/jsonnetfile.json
```diff
@@ -0,0 +1,15 @@
...
+        "git": {
+          "remote": "https://github.com/grafana/grafonnet-lib.git",
+          "subdir": "grafonnet"
+        }
+      },
+      "version": "master"
+    }
+  ],
+  "legacyImports": false
+}
```
beorn7, line 15: EOL missing

beorn7 (changes requested):
Just some formatting nits in the comments.

doc/alertmanager-mixin/dashboards/overview.libsonnet
```diff
@@ -0,0 +1,310 @@
...
+{
+  grafanaDashboards+:: {
+    local clusterTemplate =
+      template.new(
+        name='cluster',
+        datasource='$datasource',
+        query='label_values(alertmanager_alerts, %s)' % $._config.clusterLabel,
```
beorn7, line 15: I'm not quite sure if this is doing what you want.

An _Alertmanager cluster_ is a different thing from a _Kubernetes cluster_ (or generally a cluster of nodes you run services on). In fact, for HA, an _Alertmanager cluster_ will commonly span multiple _K8s clusters_. Ideally, you have only one (global) Alertmanager cluster in your org, but should you have multiple (e.g. you have a dev cluster, or you have strictly separate network partitions, each of which needs its own Alertmanager cluster), then you want your dashboard to switch between those, and when viewing at a particular Alertmanager cluster, you want to see _all_ instances included there, even if they run in different K8s clusters.

In fact, the mixin already has this concept of an Alertmanager cluster, see `alertmanagerClusterLabels` and `alertmanagerClusterName` in `config.libsonnet`. So ideally, your multi-cluster support utilizes those. The mixin allows for multiple labels to define the Alertmanager cluster, which makes the templating here hard. But I think it's possible.

doc/alertmanager-mixin/dashboards/overview.libsonnet
```diff
@@ -0,0 +1,310 @@
...
+          stack=false,
+          fill=1,
+          legend_show=false,
+        )
+        .addTarget(prometheus.target('sum(rate(alertmanager_alerts_received_total{%(alertmanagerSelector)s, %(clusterLabel)s="$cluster"}[5m]))' % $._config, legendFormat='Received'))
+        .addTarget(prometheus.target('sum(rate(alertmanager_alerts_invalid_total{%(alertmanagerSelector)s, %(clusterLabel)s="$cluster"}[5m]))' % $._config, legendFormat='Invalid'));
+
+      local notifications =
+        graphPanel.new(
+          'Rate',
```
beorn7, line 52: Shouldn't this be "Notifications send rate"?

doc/alertmanager-mixin/dashboards/overview.libsonnet
```diff
@@ -0,0 +1,310 @@
...
+          'Alerts receive rate',
+          datasource='$datasource',
+          span=6,
+          format='ops',
+          stack=false,
+          fill=1,
+          legend_show=false,
+        )
+        .addTarget(prometheus.target('sum(rate(alertmanager_alerts_received_total{%(alertmanagerSelector)s, %(clusterLabel)s="$cluster"}[5m]))' % $._config, legendFormat='Received'))
+        .addTarget(prometheus.target('sum(rate(alertmanager_alerts_invalid_total{%(alertmanagerSelector)s, %(clusterLabel)s="$cluster"}[5m]))' % $._config, legendFormat='Invalid'));
```
beorn7, line 48: Since a Prometheus server sends each alert to each Alertmanager instance, this sum will increase proportionally by the number of instances in the cluster. For example, if you add a 4th instance to your formerly three-instance AM cluster, this number will suddenly go up by 33%. That might or might not be what you want.

On the other hand, it points to a problem if some alerts only reach some of the AM instances.

Just brainstorming here, but since the number of instances in an AM cluster is usually low (three to five or something), perhaps don't sum but show the graph for each instance? In that case, it might be better to have a separate panel for the invalid alerts (which are, hopefully, tiny anyway and might make more sense in a separate panel anyway).

doc/alertmanager-mixin/dashboards/overview.libsonnet
```diff
@@ -0,0 +1,310 @@
...
+        graphPanel.new(
+          'Rate',
+          datasource='$datasource',
+          span=4,
+          format='ops',
+          stack=false,
+          fill=1,
+          legend_show=true,
+        )
+        .addTarget(prometheus.target('sum(rate(alertmanager_notifications_total{%(alertmanagerSelector)s, %(clusterLabel)s="$cluster"}[5m])) by (integration)' % $._config, legendFormat='{{integration}}'));
```
beorn7, line 60: Now here things are very much different from the alerts above. Ideally, the whole cluster sends each notification only once. And most will be sent by a single instance (the one with the highest sending priority). If suddenly a lower-priority instance sends a lot of notifications, it signals a possible problem. So perhaps again, but for different reasons, it might be brake it out to have one graph per instance in this panel? Of course, here you already brake it up by integration, too. Perhaps one separate panel per integration would make sense, as ingerations usually fail separately from each other and also might have very different notifications rates?

doc/alertmanager-mixin/dashboards/overview.libsonnet
```diff
@@ -0,0 +1,310 @@
...
+        graphPanel.new(
+          'Error',
+          datasource='$datasource',
+          span=4,
+          format='ops',
+          stack=false,
+          fill=1,
+          legend_show=true,
+        )
+        .addTarget(prometheus.target('sum(rate(alertmanager_notifications_failed_total{%(alertmanagerSelector)s, %(clusterLabel)s="$cluster"}[5m])) by (integration)' % $._config, legendFormat='{{integration}}'));
```
beorn7, line 72: Same as above, I think it makes more sense to have it broken up per instance.

doc/alertmanager-mixin/dashboards/overview.libsonnet
```diff
@@ -0,0 +1,310 @@
...
+          format='s',
+          stack=false,
+          fill=1,
+          legend_show=true,
+        )
+        .addTarget(prometheus.target(
+          |||
+            histogram_quantile(0.99,
+              rate(alertmanager_notification_latency_seconds_bucket{%(alertmanagerSelector)s, %(clusterLabel)s="$cluster"}[5m])
+            )
```
beorn7, line 88: So this _is_ actually broken up per instance, isn't it? But then it should also be in the legend.

And speaking about the "per instance" legend: This is actually where the K8s cluster label (usually "cluster") comes into the game again.

For example, here at Grafana Labs, we call our global AM cluster `global-alertmanager/alertmanager`, and this name is used as the job name, in each of the two K8s clusters where we run two instances each, all participating in the global AM cluster. The instances are then called `alertmanager-0` and `alertmanager-1` in _each_ K8s cluster, so the legend needed to include both the `instance` and the `cluster` label, so that it reads "alertmanager-0 in us-east-3" etc.

doc/alertmanager-mixin/dashboards/overview.libsonnet
```diff
@@ -0,0 +1,310 @@
...
+        graphPanel.new(
+          'Silences',
+          datasource='$datasource',
+          span=4,
+          format='none',
+          stack=false,
+          fill=1,
+          legend_show=false,
+        )
+        .addTarget(prometheus.target('sum(alertmanager_silences{%(alertmanagerSelector)s, %(clusterLabel)s="$cluster"}) by (state)' % $._config, legendFormat='{{ state }}'));
```
beorn7, line 117: Each instance should report the same number of silences. So again, summing them up will artificially inflate the number. You could again break up by instance, as different silence counts signal a problem.

doc/alertmanager-mixin/dashboards/overview.libsonnet
```diff
@@ -0,0 +1,310 @@
...
+
+      local silenceQueries =
+        graphPanel.new(
+          'Queries',
```
beorn7, line 121: This is only about queries for silences, so "Queries" seems a bit too general.

In different news, I'm not sure if this is interesting enough to be on the dashboard. It's a low traffic endpoint, driven by occasional user requests. In rare cases, I might need to look at it, but I can do that with an ad-hoc query. I doubt that it makes sense for a general overview.

doc/alertmanager-mixin/dashboards/overview.libsonnet
```diff
@@ -0,0 +1,310 @@
...
+          fill=1,
+          legend_show=false,
+        )
+        .addTarget(prometheus.target('rate(alertmanager_silences_queries_total{%(alertmanagerSelector)s, %(clusterLabel)s="$cluster"}[5m])' % $._config, legendFormat='Queries'))
+        .addTarget(prometheus.target('rate(alertmanager_silences_query_errors_total{%(alertmanagerSelector)s, %(clusterLabel)s="$cluster"}[5m])' % $._config, legendFormat='Errors'));
+
+
+      local silenceQueryDuration =
+        graphPanel.new(
+          'Queries Duration',
```
beorn7, line 135: Same concerns as above.

doc/alertmanager-mixin/dashboards/overview.libsonnet
```diff
@@ -0,0 +1,310 @@
...
+          |||
+            rate(alertmanager_silences_query_duration_seconds_sum{%(alertmanagerSelector)s, %(clusterLabel)s="$cluster"}[5m])
+            /
+            rate(alertmanager_silences_query_duration_seconds_count{%(alertmanagerSelector)s, %(clusterLabel)s="$cluster"}[5m])
+          ||| % $._config, legendFormat='Average'
+        ));
+
+      local silencesGossips =
+        graphPanel.new(
+          'Gossips',
```
beorn7, line 167: This is only about gossiping silences, which is fairly rare (only when a silence is created or updated by a user). `alertmanager_nflog_gossip_messages_propagated_total` is probably more interesting, but perhaps still too low-level to put it on the dashboard.

How about first refining the panels above and leave out gossiping for now?

doc/alertmanager-mixin/dashboards/overview.libsonnet
```diff
@@ -0,0 +1,310 @@
...
+          format='reqps',
+          stack=false,
+          fill=1,
+          legend_show=false,
+        )
+        .addTarget(prometheus.target('rate(alertmanager_silences_gossip_messages_propagated_total{%(alertmanagerSelector)s, %(clusterLabel)s="$cluster"}[5m])' % $._config, legendFormat='Messages propagated'));
+
+      local silenceGcDuration =
+        graphPanel.new(
+          'Garbage Collection Duration',
```
beorn7, line 179: That's fairly low level and presumably extremely rare to ever be a problem. Again, I'd say let's first refine the panels above and leave this out in the first iteration.

doc/alertmanager-mixin/dashboards/overview.libsonnet
```diff
@@ -0,0 +1,310 @@
...
+            rate(alertmanager_silences_snapshot_duration_seconds_sum{%(alertmanagerSelector)s, %(clusterLabel)s="$cluster"}[5m])
+            /
+            rate(alertmanager_silences_snapshot_duration_seconds_count{%(alertmanagerSelector)s, %(clusterLabel)s="$cluster"}[5m])
+          ||| % $._config, legendFormat='Duration'
+        ));
+
+
+      local HTTPRequestRate =
+        graphPanel.new(
+          'Request Rate',
```
beorn7, line 216: The requests here are mostly UI driven, and then there is the constant traffic for scraping /metrics. Not sure if this is of enough interest for the overview dashboard.

doc/alertmanager-mixin/dashboards/overview.libsonnet
```diff
@@ -0,0 +1,310 @@
...
+          legend_show=true,
+          legend_current=true,
+          legend_alignAsTable=true,
+          legend_rightSide=true,
+        )
+        .addTarget(prometheus.target('sum(rate(alertmanager_http_request_duration_seconds_count{%(alertmanagerSelector)s, %(clusterLabel)s="$cluster"}[5m])) by (handler)' % $._config, legendFormat='{{handler}}'));
+
+      local HTTPRequestDuration =
+        graphPanel.new(
+          'Request Duration',
```
beorn7, line 231: Same concern as above.

beorn7:
Tests were flaky, I re-ran them, and they succeeded.

doc/alertmanager-mixin/Makefile
```diff
@@ -1,10 +1,13 @@
 JSONNET_FMT := jsonnetfmt -n 2 --max-blank-lines 2 --string-style s --comment-style s
 ALERTMANAGER_ALERTS := alertmanager_alerts.yaml
 
-default: build
+default: vendor build
 
 all: fmt build
```
paulfantom, line 6: It would be nice if dashboards were produced as a result of `make` (without specifying any target).

doc/alertmanager-mixin/dashboards/overview.libsonnet
ArthurSens: Particularly, my setup consists of multiple k8s clusters with one alertmanager per cluster that do not communicate between themselves. I understand that my setup is not the usual Alertmanager-HA that we all should aim for, but I think I won't be able to test the filter using `alertmanagerClusterLabels` 😬

ArthurSens:
Thanks a lot for the detailed review! 

I think the main problem with my first implementation is because I don't use an HA setup for my alertmanagers. It also made me think about why I don't have an HA setup in the first place... 😅

I'll remove all the low-level metrics added and focus on the "meat", i.e. alert delivery.

ArthurSens:
Turning the PR into a draft until I think of a good way to show HA set ups...

doc/alertmanager-mixin/dashboards/overview.libsonnet
beorn7: In your case, you have "one-instance clusters", and the `alertmanagerClusterLabels` could indeed be just `cluster`.

Maybe the only problem is that the mixin as-is allows multiple labels for `alertmanagerClusterLabels` (which I introduced because it was easy at the time, and I think it's needed in some use cases). In fact, here at GL, we use `job, namespace` as the `alertmanagerClusterLabels`. `job` is usually something like `global-alertmanager` while `namespace` can be used to, for example, have a production global AM cluster and a separate test cluster.

I think this should all work, just that you need to jump through some jsonnet hoops to iterate through all the labels in `alertmanagerClusterLabels` and dynamically create the corresponding template variables for Grafana. (This will be easier if `alertmanagerClusterLabels` is a list. But we can totally make it one.)

The Grafana `query_result(query)` might also help here. See https://grafana.com/docs/grafana/latest/datasources/prometheus/#query-variable

ArthurSens:
Hi @beorn7 and @paulfantom,

I've refactored the PR to support multiple Alertmanager HA-clusters instead of multiple k8s clusters. Please take another look when possible 🙂 

I think I haven't done breaking changes to the mixin, but please pay some extra attention to the changes done in `config.libsonnet`. Thanks to the flexibility when identifying alertmanager clusters, I focused on identifying on labels names and boilerplate those labels into Grafana queries, legends and Prometheus alerting rules.

---
Here is an example dashboard:

image

beorn7:
Thanks, @ArthurSens . I'll have a look as soon as possible (which might not be very soon – very sorry, too much backlog).

@paulfantom , your feedback would be very appreciated.

doc/alertmanager-mixin/config.libsonnet
```diff
@@ -12,13 +13,25 @@
     // to keep for resulting cluster-level alerts.
     alertmanagerClusterLabels: 'job',
 
-    // alertmanagerName is inserted into annotations to name the Alertmanager
-    // instance affected by the alert.
-    alertmanagerName: '{{$labels.instance}}',
+    // alertmanagerQuerySelector is a matcher selector built from 'alertmanagerClusterLabels'.
+    // It is used to filter queries on grafana dashboards
+    alertmanagerQuerySelector: std.join(',', ['%s="$%s"' % [label, label] for label in std.split(c.alertmanagerClusterLabels, ',')]),
```
beorn7, line 18: Does this have to be in `_config`? I would think that this programmatic construction should always work.

doc/alertmanager-mixin/config.libsonnet
```diff
@@ -12,13 +13,25 @@
...
+
+    // alertmanagerNameLabels is a string with comma-separated
+    // labels used to identify different alertmanagers within the same
+    // alertmanager HA cluster.
     // If you run Alertmanager on Kubernetes with the Prometheus
     // Operator, you can make use of the configured target labels for
     // nicer naming:
-    // alertmanagerName: '{{$labels.namespace}}/{{$labels.pod}}'
+    // alertmanagerName: 'namespace,pod'
```
beorn7, line 26: This should probably be `// alertmanagerNameLabels: 'namespace,pod'`

doc/alertmanager-mixin/config.libsonnet
```diff
@@ -12,13 +13,25 @@
...
+    alertmanagerNameLabels: 'instance',
+
+    // alertmanagerName is an identifier for alerts that is built from 'alertmanagerNameLabels'
+    alertmanagerName: std.join('/', ['{{$labels.%s}}' % [label] for label in std.split(c.alertmanagerNameLabels, ',')]),
+
+    // alertmanagerNameDashboards is used to identify alertmanagers on panels legends,
+    // and is built from the 'alertmanagerNameLabels'
+    alertmanagerNameDashboards: std.join('/', ['{{%s}}' % [label] for label in std.split(c.alertmanagerNameLabels, ',')]),
```
beorn7, line 34: As above, I'd say we don't put these inte `_config` for now. If people feel the need to customize those, we can still pull them back up here.

beorn7 (changes requested):
I'm not a real expert in jsonnet, mixins, and dashboards, so I don't feel qualified to improve details on the dashboard is it is now. I think it's a good starting point, and we can iterate on it based on results from people actually using it in different scenarios. (But if any of the experts have something to improve right now, please go ahead.)

Below just a few comments on the `_config` fields. I realize that some of my suggestions will break people that have set `alertmanagerName`. But since we are now using parallel naming for dashboards, with slightly different generation, I'd say let's better break them noticeable than causing a surprise that the names on the dashboard suddenly look different than the one in the alerts.

.gitpod.yml
```diff
@@ -0,0 +1,3 @@
+vscode:
+  extensions:
+    - heptio.jsonnet@0.1.0:woEDU5N62LRdgdz0g/I6sQ==
```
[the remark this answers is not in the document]
ArthurSens: Whoops this file was added by mistake 😬

doc/alertmanager-mixin/config.libsonnet
ArthurSens: Yeah, good point. After your comment, I think it will be better as a local variable in the dashboard

ArthurSens:
Thanks again for the review Beorn, I've moved the mentioned variables from config to the dashboard libsonnet

doc/alertmanager-mixin/config.libsonnet
```diff
@@ -12,13 +13,17 @@
...
+    // alertmanager HA cluster.
     // If you run Alertmanager on Kubernetes with the Prometheus
     // Operator, you can make use of the configured target labels for
     // nicer naming:
-    // alertmanagerName: '{{$labels.namespace}}/{{$labels.pod}}'
+    // alertmanagerNameLabels: 'namespace,pod'
+    alertmanagerNameLabels: 'instance',
+
+    // alertmanagerName is an identifier for alerts that is built from 'alertmanagerNameLabels'
+    alertmanagerName: std.join('/', ['{{$labels.%s}}' % [label] for label in std.split(c.alertmanagerNameLabels, ',')]),
```
beorn7, line 26: I would either move this out of `_config`, too, or change the comment to "alertmanagerName is an identifier for alerts. By default, it is built from 'alertmanagerNameLabels'."

doc/alertmanager-mixin/config.libsonnet
```diff
@@ -12,13 +13,17 @@
     // to keep for resulting cluster-level alerts.
     alertmanagerClusterLabels: 'job',
 
-    // alertmanagerName is inserted into annotations to name the Alertmanager
-    // instance affected by the alert.
-    alertmanagerName: '{{$labels.instance}}',
+    // alertmanagerNameLabels is a string with comma-separated
+    // labels used to identify different alertmanagers within the same
+    // alertmanager HA cluster.
```
beorn7, line 18: Nit: We usually capitalize "Alertmanager".

beorn7 (changes requested):
Just a few more nits.

doc/alertmanager-mixin/config.libsonnet
ArthurSens: I'd like to leave this in config just for backward compatibility. PrometheusRules created by this mixin use this field in alert descriptions.

ArthurSens:
Thanks, Beorn 🙂 

I think the PR is getting to its last moments, would you like me to squash commits?

beorn7 (approved):
I think the backwards compatibility is an issue anyway because those that used to override `alertmanagerName` now probably need to set `alertmanagerNameLabels`. I'd say better break in a noticeable way.

Furthermore, having `alertmanagerName` in alert descriptions doesn't really require to have it as a configurable variable.

But on the other hand, users might ask for it anyway. The whole mixin is experimental anyway, so let's not try too hard to get it perfect.

I'll squash and merge, and we'll see what feedback we get.

paulfantom:
Sorry, this was completely lost in an influx of GitHub notifications (I really need to fix my setup for this). But I am happy you managed to finalize this. Let's include it in kube-prometheus and gather community feedback.

@ArthurSens would you mind bringing this into kube-prometheus?

ArthurSens:
Yep, I'll try to open the PR on kube-prometheus today or tomorrow 🙂

PR #2541, opened 2021-04-11
Allow filtering silences by creator 

m-masataka:
Add silence filter by "createdBy" item.

https://user-images.githubusercontent.com/15244170/114311273-10a2c800-9b29-11eb-9434-02ee46497002.mov


roidelapluie:
Hello @w0rm

Could you fancy a chance to review this from  a Elm perspective? Thanks!

w0rm:
@m-masataka @roidelapluie apologies for the delay.

 I wonder about the necessity to serialise the list of creators to a string and back: `{ creator1, creator2 }` vs passing things in URL encoded query vars, e.g. `creator=creator1&creator=creator2`. The current implementation feels a bit error prone to me, and we have to deal with things like string escaping and the allow list of characters, whilst the original input field for the silence creator doesn't have these restrictions:

```
isVarCharAndSymbol char =
     Char.isLower char
         || Char.isUpper char
         || String.contains (String.fromChar char) "_ .&-^[]:;*+?!#$%&()"
         || Char.isDigit char
```

Another question I have is usability point of view. 

* Do we need to filter by multiple creators? The current created by filter is joined with "OR", that works differently when filtering alerts by matches, that are joined with "AND".
* The creator field can be any text, there may be typos and extra spaces, not sure this can work reliably. Maybe allowing filtering by regex would mitigate some of that.

From the UI point of view, the button that toggles the created by bar looks a bit weird to me, but I think it would be OK to get this merged, and then I can improve (would like to play with a few ideas myself).

PR #2543, opened 2021-04-13
Clustering: Fix unsynchronised access

gotjosh:
In Cortex, we use the clustering implementation to achieve HA for each tenant. Under this scenario, we create one state per tenant and have that be synchronised through gossip.

Under this scheme, the state map can end up being called concurrently if we try to add a state while gossiping to another node. Overall, I feel it's safer to just make sure we synchronise access upstream to avoid downstream consumers worrying about this.


SuperQ (approved):
LGTM, makes sense to me.

gouthamve (approved):
LGTM

PR #2547, opened 2021-04-16
Add HTTP client options to receiver integrations

pracucci:
In Cortex we have a use case where we would need to customize the dialer used by the HTTP client in the upstream Prometheus alertmanager. Specifically, we're trying to customise the dialer used by HTTP client used by Alertmanager notifiers (eg. webhook). Unfortunately we can't do it without changing it upstream so, in this PR, I'm proposing to add HTTP options to receveirs `New()` functions.


notify/opsgenie/opsgenie.go
```diff
@@ -43,8 +43,8 @@ type Notifier struct {
 }
 
 // New returns a new OpsGenie notifier.
-func New(c *config.OpsGenieConfig, t *template.Template, l log.Logger) (*Notifier, error) {
-	client, err := commoncfg.NewClientFromConfig(*c.HTTPConfig, "opsgenie", false, false)
```
pracucci, line 47: The boolean flags in the previous version of `commoncfg.NewClientFromConfig()` were related to `disableKeepAlives, enableHTTP2 bool`, so `false, false` did mean "keepalive enabled, HTTP2 disabled". Keepalive and HTTP2 are now enabled by default so, to keep the same exact behaviour, I'm adding `WithHTTP2Disabled()`.

_Same comment applies to all other receiver integrations._

roidelapluie:
Thanks!

PR #2555, opened 2021-04-24
Integrate elm-review in CI

w0rm:
This PR integrates elm-review in CI, it also introduces another rule to simplify Elm code, you can read about it here: https://jfmengels.net/simplify/

I realised we are already pretty strict about code formatting, so decided to give it a try! 

Also running `elm-review --fix` goes case by case and proposes to fix all the problems it has discovered.


beorn7 (approved):
Disclaimer: Limited Elm foo. I just don't think we can hope for a timely review by @simonpasquier .

PR #2558, opened 2021-04-27
Dispatch: Make sure mutex gets unlocked on call to Stop

aknuds1:
It looks as if `Dispatch.Stop` doesn't unlock the mutex if `d.cancel == nil`. I'm rectifying it to unlock it also if `d.cancel == nil` and the method returns early. It seems to fix a deadlock in Grafana (on server shutdown), during our test suite.

[Here](https://gist.github.com/aknuds1/a9dcff50dd8cba2b82f570ed6b942923)'s a stacktrace of the involved goroutines when the deadlock occurs.

Let me know if you want me to write a test also.

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -258,12 +258,13 @@ func (d *Dispatcher) Stop() {
 		return
 	}
 	d.mtx.Lock()
-	defer d.mtx.Unlock()
 	if d.cancel == nil {
+		d.mtx.Unlock()
 		return
 	}
 	d.cancel()
 	d.cancel = nil
+	d.mtx.Unlock()
 
 	<-d.done
 }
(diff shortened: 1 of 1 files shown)


pstibrany (approved):
LGTM.

gouthamve (approved):

stevesg:
LGTM

stevesg (approved):

gotjosh (approved):
LGTM

beorn7 (approved):

beorn7:
Interestingly, the unlock was [correctly suggested](https://github.com/prometheus/alertmanager/pull/2208/files#r394372692) in the review for the PR that introduced the bug, but then somehow slipped under the radar.

PR #2560, opened 2021-04-29
Add OAuth 2.0 Config

LeviHarrison:
This PR upgrades the common version so that OAuth 2.0 is supported and properly documents the config.

cc @roidelapluie

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -360,6 +360,11 @@ authorization:
   # It is mutually exclusive with `credentials`.
   [ credentials_file: <filename> ]
 
+# Optional OAuth 2.0 configuration.
+# Cannot be used at the same time as basic_auth or authorization.
+oauth2:
+  [ <oauth2> ]
+
 # Optional proxy URL.
 [ proxy_url: <string> ]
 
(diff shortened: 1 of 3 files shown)


docs/configuration.md
```diff
@@ -371,6 +376,32 @@ tls_config:
   [ <tls_config> ]
 ```
 
+### `oauth2`
+
+OAuth 2.0 authentication using the client credentials grant type.
+Alertmanager fetches an access token from the specified endpoint with
+the given client access and secret keys.
```
LeviHarrison, line 383: I realize that in Prometheus I put `oauth2` after `tls_config`, but given that `oauth2` goes along with the other authentication methods and is therefore first, and decided to put it ahead of `tls_config` here like it's supposed to. That being said, it does make the ordering inconsistent.

roidelapluie: tls_config also contains authentication.

LeviHarrison: I guess, although the [order in common](https://github.com/prometheus/common/blob/b4304c504af1cf993f7b6873c5455e0ac51726a7/config/http_config.go#L132-L153) has OAuth 2.0 first anyways. I'm probably making too big of a deal out of this though.

roidelapluie: In prometheus we could reorder everything because the order is messy overall.

In common we could split some of them into separated files.

PR #2561, opened 2021-04-29
Lift moratorium on AlertManager receivers with long-term maintenance plan

maxbrodin:
This documentation change is intended to lift moratorium on new AlertManager receivers with long-term maintenance plan.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -426,6 +426,8 @@ A `tls_config` allows configuring TLS connections.
 
 Receiver is a named configuration of one or more notification integrations.
 
+Note: As part of lifting the past moratorium on new receivers it was agreed that, in addition to the existing requirements, new notification integrations will be required to have a committed maintainer with push access.
+
 ```yaml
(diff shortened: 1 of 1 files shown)


roidelapluie:
We probably want to reuse the phrasing in prometheus:

Note: As part of lifting the past moratorium on new SD implementations it was agreed that, in addition to the existing requirements, new service discovery implementations will be required to have a committed maintainer with push access (i.e., on -team).

maxbrodin:
Right. Added a note: 

>Note: As part of lifting the past moratorium on new receivers it was agreed that, in addition to the existing requirements, new notification integrations will be required to have a committed maintainer with push access.

roidelapluie (approved):
Thanks, I will merge this in a few days to make sure we have lazy consensus.

gotjosh (approved):
LGTM

QuentinBisson (approved):

SuperQ (approved):
LGTM

PR #2563, opened 2021-04-30
Fix race condition causing 1st alert to not be immediately delivered when group_wait is 0s

### LINKED ISSUE #2562: Race condition when group_wait = 0s causing the 1st alert to not be delivered immediately
**What did you do?**
While investigating a [flaky test in Cortex](https://github.com/cortexproject/cortex/pull/4150) I've found out a race condition (not easy to trigger) which can cause the 1st alert in a group to not be delivered immediately when `group_wait` is set to `0s`.

The 1st alert of a group to be always delivered immediately.

**Environment**
* Alertmanager version: `master`

**Root cause**
When the 1st alert of a group is processed by `Dispatcher.processAlert()`, this function creates a new `aggrGroup`, then calls the `ag.run()` in a dedicated goroutine and finally call `ag.insert()`:
https://github.com/prometheus/alertmanager/blob/f4b176fa5e51d8373f305fd089b7c4ff935f8080/dispatch/dispatch.go#L294-L316

The `ag.insert()` is expected to trigger an immediately flush if it hasn't flushed yet and the alert start time + `group_wait` is in the past:
https://github.com/prometheus/alertmanager/blob/f4b176fa5e51d8373f305fd089b7c4ff935f8080/dispatch/dispatch.go#L445-L447

We would expect the flushing to be always triggered for the 1st alert with `group_wait: 0s`. However, given ` ag.run()` has been previously started in a dedicated goroutine, there are cases when it already executed the 1st run and so `ag.hasFlushed` is already set to `true` once `ag.insert()` is executed, causing the flushing to not being triggered until the `group_interval` is hit.

pracucci:
This PR should fix #2562.

_Please see the issue for more details. I also suggest to review this with enabling "Hide whitespace changes"._

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -290,30 +290,36 @@ func (d *Dispatcher) processAlert(alert *types.Alert, route *Route) {
 		d.aggrGroups[route] = group
 	}
 
-	// If the group does not exist, create it.
 	ag, ok := group[fp]
-	if !ok {
-		ag = newAggrGroup(d.ctx, groupLabels, route, d.timeout, d.logger)
-		group[fp] = ag
-		d.metrics.aggrGroups.Inc()
-
-		go ag.run(func(ctx context.Context, alerts ...*types.Alert) bool {
-			_, _, err := d.stage.Exec(ctx, d.logger, alerts...)
-			if err != nil {
-				lvl := level.Error(d.logger)
-				if ctx.Err() == context.Canceled {
-					// It is expected for the context to be canceled on
-					// configuration reload or shutdown. In this case, the
-					// message should only be logged at the debug level.
-					lvl = level.Debug(d.logger)
-				}
-				lvl.Log("msg", "Notify for alerts failed", "num_alerts", len(alerts), "err", err)
-			}
-			return err == nil
-		})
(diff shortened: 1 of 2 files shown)


pracucci:
> Thanks for this, can we have a unit test to cover this?

I tried it but it's very difficult to trigger because depends on the goroutines scheduling (which one is executed first). Ideas on how to write such test in a robust way?

roidelapluie:
If you create 100 aggr groups in a test this could be robust enough ?

pracucci:
> If you create 100 aggr groups in a test this could be robust enough ?

I tried to make it reproduceable it in a unit test and it's not easy. I've committed a test that sometimes is able to reproduce it in `master` (but not always) while looks like never fails in this PR (I've run it with `-count=50`).

pracucci:
> > If you create 100 aggr groups in a test this could be robust enough ?
> 
> I tried to make it reproduceable in a unit test and it's not easy. I've committed a test that sometimes is able to reproduce it in `master` (but not always) while looks like never fails in this PR (I've run it with `-count=50`).

But CI doesn't like it:
`race: limit on 8128 simultaneously alive goroutines is exceeded, dying`

PR #2566, opened 2021-05-02
amtool: Add `silence display` subcommand, add silence-ids validation, fix nil pointer dereference.

yurkeen:
This change brings missing amtool functionality, which allows to request and display silence information by user-specified silence ID. Silence IDs are verified to be valid UUIDs, and when they are, retrieved from alertmanager. Before this change a user would need to filter silences by silence ID using other tools (think grep, jq, etc).

Naming for the subcommand, i.e. _display_, was choses over _show_ to not bring confusion to users, given that amtool already has _query_. Happy to change if any objections.

This PR also adds validation for *silence-ids*  supplied to the `silence display`, `silence update` and `silence expire` subcommands, so when supplied values are not valid UUIDs, error is returned.

Additionally, this PR fixes a nil pointer dereference  bug which happens due to [the change](https://github.com/prometheus/alertmanager/commit/2b6315f399b95c4e327dcd5f5eb10866ef844252#diff-4975284c2279cb0d4dbc76c701ea7c35725245a92caaf2e8be8e5b5e56cf1ab0) in #2471, where the _IsEqual_  `*bool` field of the _Matcher_ model [is optional](https://github.com/prometheus/alertmanager/blob/2b6315f399b95c4e327dcd5f5eb10866ef844252/api/v2/models/matcher.go#L35) and at the same time is intended to have `true` as its default value. Therefore, when a _Matcher_ with its _IsEqual_ field set to _nil_ is received by the _format/labelsMatcher()_ func, amtool panics.


yurkeen:
@simonpasquier or @w0rm – may I seek your reviews, when possible? Thanks!

angelinelee22:
Hi, 
I've been running into `panic: runtime error: invalid memory address or nil pointer dereference` whenever I run `amtool silence query` with either simple or extended formatting. Wanted to confirm if this is related to the above, although it runs fine with the `-q` quiet tag or `-o json` otherwise. If so, the issue has rendered `amtool silence query` unusable. It'd be great if this could be pushed along. Thanks!

yurkeen:
Hi @angelinelee22,

It is very likely that you are facing the bug covered by this PR, if your alertmanager version is v0.22.0 or older, where #2602 was not yet fixed.

angelinelee22:
Hi @yurkeen,
Sorry for missing this. I am actually using 0.22.2 on the darwin/amd64 platform and can confirm I'm still facing this issue.

PR #2571, opened 2021-05-05
Added group limit to dispatcher.

pstibrany:
This PR introduces limit for maximum number of per-route groups that Alertmanager's dispatcher can have. Dispatcher cannot report error back when limit is hit, so it just logs this and increments metric.

Similar to https://github.com/prometheus/alertmanager/pull/2569, we would like to use this limit in Cortex to protect Cortex' multitenant alertmanager against misconfigurations. In Prometheus Alertmanager itself this limit is not used. If there is interest to have this limit configurable in Prometheus Alertmanager, I'm happy to send new PR with that change.

(TODO: I'd like to add unit test for cleanup of groups, but other than that PR is ready for review / discussion about the approach.)

### DIFF
--- a/dispatch/dispatch_test.go
+++ b/dispatch/dispatch_test.go
@@ -504,7 +504,7 @@ route:
 	logger := log.NewNopLogger()
 	route := NewRoute(conf.Route, nil)
 	marker := types.NewMarker(prometheus.NewRegistry())
-	alerts, err := mem.NewAlerts(context.Background(), marker, time.Hour, logger)
+	alerts, err := mem.NewAlerts(context.Background(), marker, time.Hour, nil, logger)
 	if err != nil {
 		t.Fatal(err)
 	}
--- a/dispatch/dispatch_test.go
+++ b/dispatch/dispatch_test.go
@@ -375,7 +375,7 @@ route:
 	timeout := func(d time.Duration) time.Duration { return time.Duration(0) }
 	recorder := &recordStage{alerts: make(map[string]map[model.Fingerprint]*types.Alert)}
 	lim := limits{groups: 6}
-	m := NewDispatcherMetrics(prometheus.NewRegistry())
+	m := NewDispatcherMetrics(true, prometheus.NewRegistry())
 	dispatcher := NewDispatcher(alerts, route, recorder, marker, timeout, lim, logger, m)
 	go dispatcher.Run()
 	defer dispatcher.Stop()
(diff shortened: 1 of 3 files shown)


pracucci (approved):
Good job!

roidelapluie:
Can we not have the extra metric in alertmanager?

pstibrany:
> Can we not have the extra metric in alertmanager?

Then the only way to find that something went wrong would be a log line :( We would like to use this metric in Cortex, and expose it per user.

roidelapluie:
I mean, can we call register only if the limit is set?

pstibrany:
> I mean, can we call register only if the limit is set?

I see. I've done the change to avoid registering new metric in vanilla Alertmanager.

pstibrany:
Ping. Anything I can do to help with review and merging?

pstibrany:
Failing test seems to be unrelated to my changes, and I cannot reproduce the failure locally :(

dispatch/dispatch_test.go
```diff
@@ -373,7 +374,9 @@ route:
 
 	timeout := func(d time.Duration) time.Duration { return time.Duration(0) }
 	recorder := &recordStage{alerts: make(map[string]map[model.Fingerprint]*types.Alert)}
-	dispatcher := NewDispatcher(alerts, route, recorder, marker, timeout, logger, NewDispatcherMetrics(prometheus.NewRegistry()))
+	lim := limits{groups: 6}
```
roidelapluie, line 377: With the current implementation of the code, this is fine. However, I could see in the future optimizations that would change the way we deal with the limit.

I would encourage you to split this test to avoid regressions in both paths - with and without limits.

roidelapluie (approved):

dispatch/dispatch_test.go
pstibrany: That's a good idea. I've extracted (and simplified) `TestGroupsWithLimits` test from previous `TestGroups`.

roidelapluie:
Thanks, will merge on green

pstibrany:
Closed/reopened to trigger stuck CI.

pstibrany:
Can I poke CI somehow? (closing/reopening didn't work here 😞)

roidelapluie:
I have contacted circleci about this :)

roidelapluie:
dispatch/dispatch_test.go:507:72: cannot use logger (variable of type log.Logger) as mem.AlertStoreCallback value in argument to mem.NewAlerts: missing method PostDelete (typecheck)

pstibrany:
Unrelated test is failing now: `TestDispatcherRaceOnFirstAlertNotDeliveredWhenGroupWaitIsZero` :(

roidelapluie:
Thanks!

PR #2581, opened 2021-05-14
API: Only pass cluster peer if empty

roidelapluie:


gotjosh (approved):
Thanks @roidelapluie, are you OK with me creating a test to catch this regression? I'm happy to do it as a follow up PR to unblock the RC.

roidelapluie:
Yes please go ahead

Issue #2426, opened 2020-11-20
Alert manager silence feature is working partially


arnavbose:
Hi,

We came across an issue with our alert manager recently. We are able to create silences successfully and are also able to see the created silences in the UI. The issue is that the silence is working partially. 

Eg: Suppose there are 10 servers for a host with a label host_name. If I create a silence for the host_name, it should silence all alerts coming from all those 10 servers belonging to the host during the specified duration. What is happening is that sometimes 1 or 2 server alerts from the host are spilling into the slack channel. The rest majority of the alerts are silenced, but just a few of them at random are not silenced. Has this been observed previously in any version of alert manager? Was this a known issue in an older version of alert manager? 

Thanks,
Arnav

sam-celer:
amtool reports the same issue. 

Silence on client_name 'fern' currently active
7b023dad-0f95-4a68-b8a2-41d9d0df463c  client_name~=fern                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                   2020-11-21 05:00:00 UTC    2020-11-28 05:01:00 UTC    2020-11-21 04:59:17 UTC  prometheus-silencer  [fern]silence fern

Yet there are alerts with client_name fern
[...]

PR #2586, opened 2021-05-17
fix bug:The silence rule does not work in some situations

waitstory:
When the state of the created silence rule is pending,
the silence rule cannot be silent alert.For this change,
I validated it ensure that the function of the system is
not affected system.

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -123,21 +123,33 @@ func (s *Silencer) Mutes(lset model.LabelSet) bool {
 	if markerVersion == s.silences.Version() {
 		// No new silences added, just need to check which of the old
 		// silences are still relevant.
-		if len(ids) == 0 {
-			// Super fast path: No silences ever applied to this
-			// alert, none have been added. We are done.
-			return false
-		}
+
 		// This is still a quite fast path: No silences have been added,
 		// we only need to check which of the applicable silences are
 		// currently active. Note that newVersion is left at
 		// markerVersion because the Query call might already return a
 		// newer version, which is not the version our old list of
 		// applicable silences is based on.
-		sils, _, err = s.silences.Query(
-			QIDs(ids...),
-			QState(types.SilenceStateActive),
-		)
+		if len(ids) != 0 {
+			// Super fast path: No silences ever applied to this
+			// alert, none have been added. We are done.
+			sils, _, err = s.silences.Query(
+				QIDs(ids...),
+				QState(types.SilenceStateActive),
+			)
+		} else {
+			// Check if there is a silence rule that changes from pending to
+			// active state.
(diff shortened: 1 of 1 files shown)


beorn7:
Hi @waitstory , thanks for your contribution. Your description is hard to understand. I tried to find out what bug you are addressing here, and I think it is the following:

If a silence is pending, i.e. it is created with a start date in the future, and then we do not create new silences, we'll never notice when the silence becomes active.

However, I don't think your fix fixes all possible ways the problem can occur. It only works if there are no IDs, but if old IDs exist, but they are all expired by now, you would still run into the same problem. Also, your approach nullifies the whole point of versioning the silence state, which is that we don't have to run a full query each time we call the method.

Maybe a better approach would be to not only save the IDs fur currently active silences but also separately for pending silences so that we check those even if the version of the silences did not change.

The best starting point is to create a unit test that exposes the bug.

beorn7:
I think what we discuss here causes #2426 .

waitstory:
The problem I've encountered is creating a silence rule that takes effect at a future time, and there are no silent matching alerts.
Take the following case, When the current time is "02:03:11, 2021-05-19 (UTC)", create a silence rule starting with the time "02:05:29, 2021-05-19 (UTC)".The state of silence rule is "pedding".But creating this silence rule does not silently match alerts within the time range in which it is in effect.That's a real problem.
image

beorn7:
Yes, that's precisely the problem I tried to understand. And yes, that's a real problem.
However, as said, I don't think you code fixes the problem properly, and at the same time it still defeats the whole purpose of versioning.

As said, I think we need to track pending silences affecting an alert in the future separately.

And as also said, the first step would be to create a unit test exposing the problem.

Do you think you could create such a unit test?

waitstory:
I think it's hard to do a unit test to expose the bug. Also, I would like to ask when this problem will be fixed. This problem is in the use of the process will appear the problem.

Besides, I don't quite understand the meaning of this sentence—"I think we need to track pending silences affecting an alert in the future".

beorn7:
I can probably best explain it by actually fixing the issue. I'll try as soon as possible.

It is, indeed, a bug hard to fix. And the unit test is probably quite involved, too. And yes, I agree we need to fix this, it's a rather nasty bug.

So if anyone with intimate knowledge of this codebase could beat me to fixing this, even better.

beorn7:
OK, I hacked together #2590. Still needs testing.

beorn7:
I have now created a test case exposing the problem and a working fix in PR #2590.

Therefore, I'm closing this. @waitstory , many thanks for your attempt. It made me aware of the issue and enabled me to create the fix in #2590.

PR #2588, opened 2021-05-18
Add Syslog Notifier

### LINKED ISSUE #2468: Syslog Receiver
GCP logging and alerting has come a long way recently. Sending Syslog messages to GCP is super easy, which opens up the power of GCP to be used for alert channels to human beings and also to centralized logging. So +1 for syslog receiver as suggested in #448 

For those of us standardizing on GCP, it's not even optional. We would like to use Prometheus, but will not without using GCP for logging and alerting (much to gain by using GCP's tools and reducing number of third parties), and the best collection method GCP has is google-fluentd which pushes syslog messages to GCP.

LeviHarrison:
This PR adds a Syslog notifier for unix-based systems.

Fixes #2468


template/default.tmpl
```diff
@@ -217,3 +217,12 @@ Alerts Resolved:
...
 {{ define "pushover.default.url" }}{{ template "__alertmanagerURL" . }}{{ end }}
+
+
+{{ define "__text_alert_list_oneline" }}{{ range . -}}| Labels:{{ range .Labels.SortedPairs }} - {{ .Name }} = {{ .Value }}{{ end }}; Annotations:{{ range .Annotations.SortedPairs }} - {{ .Name }} = {{ .Value }}{{ end }}; Source:{{ .GeneratorURL }}{{- end }}{{- end }}
+
+
+{{ define "__subject_oneline" }}[{{ .Status | toUpper }}{{ if eq .Status "firing" }}:{{ .Alerts.Firing | len }}{{ end }} {{ .GroupLabels.SortedPairs.Values | join " " }}{{ if gt (len .CommonLabels) (len .GroupLabels) }}({{ with .CommonLabels.Remove .GroupLabels.Names }}{{ .Values | join " " }}{{ end }}){{ end }} {{ .CommonAnnotations.SortedPairs.Values | join " " -}}]{{ end }}
+
+
+{{ define "syslog.default.message" }}{{ template "__subject_oneline" . }} {{ if gt (len .Alerts.Firing) 0 -}}Alerts Firing: {{ template "__text_alert_list_oneline" .Alerts.Firing }}{{- end }}{{ if gt (len .Alerts.Resolved) 0 -}}Alerts Resolved: {{ template "__text_alert_list_online" .Alerts.Resolved }}{{- end }} AlertmanagerUrl: {{ template "__alertmanagerURL" . }} |{{- end }}
```
LeviHarrison, line 228: Syslog doesn't support the `\n` character, and will convert it to something else weird.

LeviHarrison:
I'll take a look at these test failures tomorrow.

docs/configuration.md
```diff
@@ -912,4 +918,33 @@ API](http://admin.wechat.com/wiki/index.php?title=Customer_Service_Messages).
...
+daemon:
+  [ <daemon_config> ... ]
+```
+
+## `<daemon_config>`
+
+Configure an alternate Syslog daemon.
+
+```yaml
+# Formats as 'tcp://localhost:1234'
```
roidelapluie, line 946: I think we could provide the end-form directly, without indirection.

config/notifiers.go
```diff
@@ -127,6 +127,22 @@ var (
 		Expire:   duration(1 * time.Hour),
 		HTML:     false,
 	}
+
+	// DefaultSyslogConfig defines default values for Syslog configurations.
+	DefaultSyslogConfig = SyslogConfig{
+		NotifierConfig: NotifierConfig{
+			VSendResolved: false,
```
roidelapluie, line 134: All the other SD's have send_resolved to true.

roidelapluie:
Thank you for this.

I did not test it yet. Some questions/comments:

- I think we miss setting the facility.
- Which implementation did you use to test this?
- Can we have the daemon set as a simple string?

LeviHarrison:
> I think we miss setting the facility.

Looks so, thanks for bringing that to my attention.

> Which implementation did you use to test this?

I tested this on my Linux machine, running Linux Mint 20.1 (Ubuntu Focal) with the Linux Kernel 5.4.0. Logs were viewed in the GNOME Logs application. Hopefully, that answers your question.

> Can we have the daemon set as a simple string?

Yup, that sounds like a good idea!

Also, I believe that the tests are failing because of permissions, given the error `error initializing syslog Unix syslog delivery error`. I got the same when running a provided example on play.golang.org. Although it might be an option to elevate the permissions in CircleCI, I think that we can probably just remove the two tests that send logs altogether. The `Notify` function is pretty simple and the only real logic is in the `New` function and the template, which can be tested separately. The package itself has tests for the actual logs anyway.

roidelapluie:
We should look if we can set a max size of messages (and default to 64k), otherwise, crop the message.

It might be interesting, in another PR, to add TLS support as well.

Concerting the format of the message, I would like to use a format directly recognized by syslong ng: cee ( https://www.rsyslog.com/json-elasticsearch/ )


`@cee:{"alertmanager":{aef732f:{status:firing,labels:labels, annotations:annotations}}}}`

@ cee is a prefix and alertmanager is a top level key. I am unsure lists are supported, so I have crafted an ID here.


What do you think?

LeviHarrison:
Although it seems rsyslog is widely used, I don't think it's the default, and neither are its formats. For example, in #2468 the use case was GCP's logging agent, which is fluentd. Fluentd doesn't seem to support formats like cee, only the ones defined in the RFCs (https://docs.fluentd.org/input/syslog#message_format). Although a JSON-based format like cee is definitely better suited for this task, I think that the only common formats among all the log agents will be the ones in the RFCs. Also, from my (somewhat cursory) research it seems that the only Syslog library for Go is `log/syslog`; all others link back to it. So, our options are limited unless we want to write our own (for example I don't think it supports TLS).

https://datatracker.ietf.org/doc/html/rfc3164
https://datatracker.ietf.org/doc/html/rfc5424

roidelapluie:
CEE is backwards compatible with the formats you have shown, as only the message is changing. I think it would be a nice default, which could be changed with templates anyway.

LeviHarrison:
Ok, well in that case it sounds like a great idea! I do think that we should provide an easier way to switch back to the standard format because writing templates is a little bit arduous (in my opinion). Maybe like a property `format`, where the default is `cee`, but it can be switched to `standard` (or possibly the other way around) and the one-line template I wrote (which may have to be tweaked) can be used unless specified otherwise.

I believe that the log part of the cee message can just be used like regular JSON, including lists, so that shouldn't be any problem. I wonder how (or if) we could verify the output is correct through tests because I can't find any cee packages for Go. Going off on a tangent, maybe cee support should land somewhere over in common (probably promlog) in case we want to reuse it elsewhere. I know there's some Syslog stuff already there.

I think it also might be nice to provide a pre-written custom template in the documentation somewhere because it seems that you need to write one of those to have other properties (see bottom of https://www.rsyslog.com/json-elasticsearch/).

roidelapluie:
Yes, let's not try to write a go template for this :')

LeviHarrison:
Actually, maybe we just encode it in JSON regardless of if it's cee or standard. I think that's the best human-readable format if we can't use newlines.

LeviHarrison:
Also, what do you think about this?

> Going off on a tangent, maybe cee support should land somewhere over in common (probably promlog) in case we want to reuse it elsewhere. I know there's some Syslog stuff already there.

roidelapluie:
There is no syslog elsewhere, let's only keep this in alertmanager.

LeviHarrison:
While I don't think this necessarily is a reason not to use cee, it does seem that the project has stopped: https://cee.mitre.org/. Project Lumberjack, which was mentioned in the rsyslog article, has gone offline and I can't find any mention of it elsewhere (https://fedorahosted.org/lumberjack/).

LeviHarrison:
The proposal has fallen flat (https://docs.google.com/document/d/1Oevu2stHVGAupzmc9C7_wW5nTb_CJ6Ut72viXfve6zI/edit?disco=AAAAMi05yRM). This code probably wasn't going to be used anyway.

LeviHarrison:
The proposal has fallen flat (https://docs.google.com/document/d/1Oevu2stHVGAupzmc9C7_wW5nTb_CJ6Ut72viXfve6zI/edit?disco=AAAAMi05yRM). This code probably wasn't going to be used anyway.

PR #2594, opened 2021-05-22
Fix example silence query parameter

marcel-tov:
The flag **-p** ist on the wrong position in your example silence query command.


roidelapluie (approved):

roidelapluie:
Thanks!

PR #2603, opened 2021-05-27
Default the isEqual flag to true in alertmanager

sinkingpoint:
Default the isEqual flag to true in alertmanager                        
                                                                        
Solves #2602 - the new version of alertmanager introduces an isEqual    
flag to support != matchers in alertmanager silences. Clients set this  
to true in the JSON blob sent to the api to indicate a foo=bar matcher, 
and to false to indicate foo!=bar. Due to gos unmarshalling, anything   
that _doesn't_ include this flag will default to false (i.e. a !=       
matcher), so any clients that aren't aware of this flag (such as amtools
before negative matchers and the new api) will not send it, and when you
think you are making a foo=bar matcher, alertmanager will interpret that
as a not equals.                                                        
                                                                        
This commit changes the Unmarshaling of the v1matcher struct to default 
the IsEqual flag to true, to keep the old behaviour for clients not     
setting the flag                                                        
                                                                        
### DIFF
--- a/pkg/labels/matcher.go
+++ b/pkg/labels/matcher.go
@@ -110,7 +110,10 @@ func (m Matcher) MarshalJSON() ([]byte, error) {
 }
 
 func (m *Matcher) UnmarshalJSON(data []byte) error {
-	var v1m apiV1Matcher
+	v1m := apiV1Matcher{
+		IsEqual: true,
+	}
+
 	if err := json.Unmarshal(data, &v1m); err != nil {
 		return err
 	}
--- a/pkg/labels/matcher_test.go
+++ b/pkg/labels/matcher_test.go
@@ -195,7 +195,7 @@ line`,
 	}
 }
 
-func TestMatcherJSON(t *testing.T) {
+func TestMatcherJSONMarshal(t *testing.T) {
 	tests := []struct {
 		name  string
 		op    MatchType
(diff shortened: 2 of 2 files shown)


pkg/labels/matcher.go
```diff
@@ -115,15 +116,22 @@ func (m *Matcher) UnmarshalJSON(data []byte) error {
 		return err
 	}
 
+	if v1m.IsEqual == nil {
```
roidelapluie, line 119: Can we instead of the pointer, do :

```
var v1m apiV1Matcher {
  isEqual: true
}
f err := json.Unmarshal(data, &v1m); err != nil {
```

That would probably solve the issue more simply.

sinkingpoint: Yeah sure - that sounds saner

sinkingpoint: Done

roidelapluie (commented):
Thank you. I have tested the fix and it works.

One last nit: can you rename the existing test TestMatcherJSONMarshal and add your tests to a TestMatcherJSONUnmarshal?

roidelapluie:
Also, if you could rebase on top of the branch release-0.22 this would be great - while not mandatory: if you do not rebase, I will simply cherry pick into the release branch.

sinkingpoint:
Done and done

roidelapluie:
> Done and done. Apologies for the initial scare - I'm glad the scope turned out to be less than I originally proposed

I expect that many people use the api v1 - without amtool. This is still an important bug.

roidelapluie:
Thank you!

beorn7:
Thanks for catching this. I thought we had covered it in the original change. But I vaguely remember that we thought about v1 API not as carefully, assuming it's rarely used and soon to be removed. But as @roidelapluie said, it should certainly keep working as expected as long as it is there.

asquare14:
Thank you for the catch and the speedy fix @sinkingpoint.

PR #2609, opened 2021-06-03
OpsGenie: refer to alert instead of incident

crepetl:
* OpsGenie provides management of both alerts and incidents.
* package opsgenie uses OpsGenie alert api, so let's refer to
  alert instead of incident in logs and doc to avoid confusion.


roidelapluie (approved):

roidelapluie:
Thanks!

PR #2613, opened 2021-06-10
Improve the AlertmanagerMembersInconsistent alert

fpetkovski:
The expression `alertmanager_cluster_members{job="alertmanager"}[5m])` is assumed to return
one series for each alertmanager instance in the cluster. When running inside Kubernetes,
alertmanager pods can get evicted and rescheduled. This can change the instance label and
produce a new series for that alertmanager instance.

When this happens, there will be a short interval in which Prometheus returns values from both
the new series and the old series. As a result, counting the number of
series for the `alertmanager_cluster_members` metric will overestimate the
number of instances in the given cluster.

This commit modifies the the AlertmanagerMembersInconsistent alert to group series by
`alertmanagerNameLabels` when calculating the number of alertmanagers in
the cluster.

Please refer to the following link for more information on how Prometheus deals with staleness: https://www.robustperception.io/staleness-and-promql

Signed-off-by: fpetkovski <filip.petkovsky@gmail.com>

### DIFF
--- a/doc/alertmanager-mixin/alerts.libsonnet
+++ b/doc/alertmanager-mixin/alerts.libsonnet
@@ -29,7 +29,7 @@
               < on (%(alertmanagerClusterLabels)s) group_left
                 count by (%(alertmanagerClusterLabels)s) (max_over_time(alertmanager_cluster_members{%(alertmanagerSelector)s}[5m]))
             ||| % $._config,
-            'for': '10m',
+            'for': '15m',
             labels: {
               severity: 'critical',
             },


doc/alertmanager-mixin/alerts.libsonnet
```diff
@@ -25,9 +25,11 @@
             expr: |||
               # Without max_over_time, failed scrapes could create false negatives, see
               # https://www.robustperception.io/alerting-on-gauges-in-prometheus-2-0 for details.
-                max_over_time(alertmanager_cluster_members{%(alertmanagerSelector)s}[5m])
+                max by (%(alertmanagerNameLabels)s) (max_over_time(alertmanager_cluster_members{%(alertmanagerSelector)s}[5m]))
               < on (%(alertmanagerClusterLabels)s) group_left
-                count by (%(alertmanagerClusterLabels)s) (max_over_time(alertmanager_cluster_members{%(alertmanagerSelector)s}[5m]))
+                count by (%(alertmanagerClusterLabels)s) (
+                    max by (%(alertmanagerNameLabels)s) (max_over_time(alertmanager_cluster_members{%(alertmanagerSelector)s}[5m]))
+                )
```
beorn7, line 32: I don't think this is a generally valid solution. If `alertmanagerNameLabels` are not a superset of `alertmanagerClusterLabels`, the `max by (%(alertmanagerNameLabels)s)` will aggregate away the labels we need for the matching `on (%(alertmanagerClusterLabels)s)`.

I think the issue you describe is real, but is it relevant? The overcounting will only happen for a short while, not much more than 5m. And that's presumably the reason why the `for` clause below is `10m`. Have you actually seen the alert firing falsely? Would it have helped to use an even longer `for` duration?

fpetkovski: Thanks for the review @beorn7, I see now that the additional `max` aggregation might be introducing a bug in the alert.

To your question on whether the alert fires in our case, the answer is yes, we see it happen. I've attached a screenshot counting the number of series coming from alertmanager. We have a cluster of 3 instances and the count between 1:42:00 and 1:53:12 is greater than 3 as seen in the screenshot.
Screenshot from 2021-06-15 17-09-49

A longer `for` clause would have helped in our case and is a viable solution. The downside is that such a change might affect the false negative rate, or prolong the time it takes to detect.

fpetkovski: We managed to track down the reason for the overcounting. In our case, alertmanager pods were getting evicted several times in a short period of time which produced multiple stale series. This is why the graph attached above shows a peak which lasts longer than 10 minutes.

In order to avoid complicating the alert expression, I can change the PR to increase the duration to 15m in order to reduce the chance of a false positive. What do you think about this solution?

beorn7: I think trying a longer `for` duration is probably the best. After all, those frequent evictions can actually lead to weird situations where different AM cluster members have different ideas about their peers, and that's what this alert is about. However, it's not an urgent alert. If those things happen, alerts should still reach their receivers "at least once" (which is the desired semantics). So in doubt, we could go even longer. But 15m sounds like a good start.

fpetkovski: That makes complete sense. I've extended the `for` duration to `15m`.

beorn7 (approved):

beorn7:
Thank you!

Issue #2559, opened 2021-04-27
Prometheus SNS receiver proposal


maxbrodin:
# Prometheus SNS receiver

## Problem

AlertManager allows to define receivers - notification integration with [email](https://prometheus.io/docs/alerting/latest/configuration/#email_config), [webhook](https://prometheus.io/docs/alerting/latest/configuration/#webhook_config) and third-party integrations like [PagerDuty](https://prometheus.io/docs/alerting/latest/configuration/#pagerduty_config), [OpsGenie](https://prometheus.io/docs/alerting/latest/configuration/#opsgenie_config), [Slack](https://prometheus.io/docs/alerting/latest/configuration/#slack_config) and [others](https://prometheus.io/docs/alerting/latest/configuration/#receiver). 
Currently there is no integration with [Amazon Simple Notification Service](https://aws.amazon.com/sns/?whats-new-cards.sort-by=item.additionalFields.postDateTime&whats-new-cards.sort-order=desc) which provides fully managed pub/sub messaging, SMS, email, and mobile push notifications. There is a workaround with [webhook](https://prometheus.io/docs/alerting/latest/configuration/#webhook_config) receiver as a proxy, but it lacks support of AlertManager templates and requires setup and maintenance of additional component.

## Proposed solution

This proposal is to add Prometheus SNS receiver - native support of notification integration with [Amazon SNS](https://aws.amazon.com/sns/)

### Message destinations

Prometheus SNS receiver can publish messages to the following destinations:

* Amazon SNS topic
* SMS message directly to a phone number
* notification to a mobile platform endpoint
* email endpoint

### SNS Publish API

In order to publish message to an SNS topic the following [HTTP request parameters](https://docs.aws.amazon.com/sns/latest/api/API_Publish.html#API_Publish_RequestParameters) are required:
Common:

* API Regional URL
* API Version

Specific for each request

* SNS Topic ARN (or Phone number for SMS or TartgetARN for mobile notifications)
* Message
* Request Signature Version 4

Optional

* Subject (when the message is delivered to email endpoints)

### <sns_config>

```
# Whether or not to notify about resolved alerts.
[ send_resolved: <boolean> | default = false ]

# The SNS API URL i.e. https://sns.us-east-2.amazonaws.com
api_url: <tmpl_string>

# The SNS API version i.e. 
[ api_version: <string> | default = sns.default.api_version ]

# Configures AWS's Signature Verification 4 signing process to sign requests.
sigv4:
[...]

PR #2615, opened 2021-06-10
Add AWS SNS receiver

treid314:
Work for issue: https://github.com/prometheus/alertmanager/issues/2559

Work Remaining:

- [x] Support AWS ARN auth
- [x] Get AWS keys from environment variables
- [x] Properly truncate SNS messages
- [x] Add email templates
- [x] Document new config options for SNS

Signed-off-by: Tyler Reid <tyler.reid@grafana.com>


roidelapluie:
Yes, preferably SIGv4 would move to its own go module in common.

config/notifiers.go
```diff
@@ -579,3 +589,46 @@ func (c *PushoverConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
 	}
 	return nil
 }
+
+// TODO: Move to common?
```
pracucci, line 593: What is this comment about? `SigV4Config` only? If so, please move the TODO right before its `type` definition.

config/notifiers.go
```diff
@@ -127,6 +127,16 @@ var (
 		Expire:   duration(1 * time.Hour),
 		HTML:     false,
 	}
+
+	// DefaultSNSConfig defines default values for SNS configurations.
+	DefaultSNSConfig = SNSConfig{
+		NotifierConfig: NotifierConfig{
+			VSendResolved: true,
+		},
+		APIVersion:  "sns.default.api_version",
```
pracucci, line 136: What's `sns.default.api_version`? Did you want it to be a template?

config/notifiers.go
```diff
@@ -579,3 +589,46 @@ func (c *PushoverConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
...
+
+	HTTPConfig *commoncfg.HTTPClientConfig `yaml:"http_config,omitempty" json:"http_config,omitempty"`
+
+	APIUrl      string            `yaml:"api_url" json:"api_url"`
+	APIVersion  string            `yaml:"api_version,omitempty" json:"api_version,omitempty"`
+	Sigv4       SigV4Config       `yaml:"sigv4" json:"sigv4"`
+	TopicARN    string            `yaml:"topic_arn,omitempty" json:"topic_arn,omitempty"`
+	PhoneNumber string            `yaml:"phone_number,omitempty" json:"phone_number,omitempty"`
+	TargetARN   string            `yaml:"target_arn,omitempty" json:"target_arn,omitempty"`
+	IsFIFOTopic bool              `yaml:"is_fifo_topic,omitempty" json:"is_fifo_topic,omitempty"`
```
pracucci, line 617: Looks a weird naming to me. Also I believe it's not much future proof, cause if a 3rd SNS queue type will be added it will not fit very well into a boolean. What's about a `queue_type` whose values can be either `standard` (default) or `fifo`?

config/testdata/conf.sns-test.yml
```diff
@@ -0,0 +1,16 @@
+route:
```
pracucci, line 1: Who is using this file?

notify/sns/sns.go
```diff
@@ -0,0 +1,171 @@
...
+			copy(truncated, message)
+			return string(truncated), true, nil
+		}
+		return message, false, nil
+	}
+	return "", false, fmt.Errorf("non utf8 encoded message string")
+}
+
+// New returns a new SNS notification handler.
+func New(c *config.SNSConfig, t *template.Template, l log.Logger, httpOpts ...commoncfg.HTTPClientOption) (*Notifier, error) {
```
pracucci, line 157: [nit] I would move the `New()` right below `type Notifier struct` definition.

notify/sns/sns.go
```diff
@@ -0,0 +1,171 @@
...
+
```
pracucci, line 158:
```suggestion
```

notify/sns/sns.go
```diff
@@ -0,0 +1,171 @@
...
+type Notifier struct {
+	conf    *config.SNSConfig
+	tmpl    *template.Template
+	logger  log.Logger
+	client  *http.Client
+	retrier *notify.Retrier
+}
+
+func (n Notifier) Notify(ctx context.Context, alert ...*types.Alert) (bool, error) {
+	credentials := credentials.NewStaticCredentials(n.conf.Sigv4.AccessKey, string(n.conf.Sigv4.SecretKey), "")
```
pracucci, line 45: credentials collides with the imported package name. You should rename the variable.

notify/sns/sns.go
```diff
@@ -0,0 +1,171 @@
...
+	if n.conf.Sigv4.AccessKey == "" {
+		credentials = nil
+	}
```
pracucci, line 48: Looks a bit weird. Why calling `credentials.NewStaticCredentials()` at all if then we would reset it to `nil`. You could reverse the logic here, and call `credentials.NewStaticCredentials()` only if you're going to use it.

notify/sns/sns.go
```diff
@@ -0,0 +1,171 @@
...
+
+	sess, err := session.NewSessionWithOptions(session.Options{
+		Config: aws.Config{
+			Region:      aws.String(n.conf.Sigv4.Region),
```
pracucci, line 52: Leaving the comment here, but applies to many config options. If you look at the specs in #2559, you can see many config options should actually be `tmpl_string`. The string is not evaluated as a template here (correct me if I'm wrong), but I think you should. As an example, look at `tmplText()` usage in the Slack integration.

notify/sns/sns.go
```diff
@@ -0,0 +1,171 @@
...
+		publishInput.SetSubject(n.conf.Subject)
+	}
+
+	// Deduplication key is only added if it's a FIFO SNS Topic.
+	if n.conf.IsFIFOTopic {
+		key, err := notify.ExtractGroupKey(ctx)
+		if err != nil {
+			return false, err
+		}
+		publishInput.SetMessageDeduplicationId(key.Hash())
```
pracucci, line 125: I see that's part of the original specifications but I'm not sure I fully understand why we set it to the group key's hash. To my understanding (but I may be wrong) is that the 5-minutes deduplication time window done by SNS will conflict with the Alertmanager `group_interval` if it's set less to 5 minutes because we may loose new alerts added to the group because of the SNS deduplication.

notify/sns/sns.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+
+	publishOutput, err := client.Publish(publishInput)
+	if err != nil {
+		// AWS Response is bad, probably a config issue.
```
pracucci, line 130: [nit] Not sure this comment is valuable.

notify/sns/sns.go
```diff
@@ -0,0 +1,171 @@
...
+		return false, err
+	}
+
+	err = n.logger.Log(publishOutput.String())
+	if err != nil {
+		return false, err
+	}
+
+	// Response is good and does not need to be retried.
```
pracucci, line 139: [nit] Not sure this comment is valuable.

notify/sns/sns.go
```diff
@@ -0,0 +1,171 @@
...
+		if err != nil {
+			return false, err
+		}
+		publishInput.SetMessageDeduplicationId(key.Hash())
+	}
+
+	publishOutput, err := client.Publish(publishInput)
+	if err != nil {
+		// AWS Response is bad, probably a config issue.
+		return false, err
```
pracucci, line 131: Why `return false` (no retry)? I think the most common use case could be a network issue, for which we want retries.

notify/sns/sns.go
```diff
@@ -0,0 +1,171 @@
...
+	commoncfg "github.com/prometheus/common/config"
+)
+
+// Notifier implements a Notifier for SNS notifications.
+type Notifier struct {
+	conf    *config.SNSConfig
+	tmpl    *template.Template
+	logger  log.Logger
+	client  *http.Client
+	retrier *notify.Retrier
```
pracucci, line 41: The `retrier` doesn't look used right now, but I think we should.

notify/sns/sns_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"github.com/go-kit/kit/log"
+	"github.com/prometheus/alertmanager/config"
+	"github.com/prometheus/alertmanager/notify/test"
+	"github.com/prometheus/alertmanager/types"
+	commoncfg "github.com/prometheus/common/config"
+	"github.com/prometheus/common/model"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNotifier_Notify(t *testing.T) {
```
pracucci, line 29: How is this test expected to work? We can't really push to SNS endpoint.

pracucci (commented):
Thanks @treid314 for working on this! I left some comments to hopefully help you progressing on it.

notify/sns/sns_test.go
treid314: This test has been used for my local testing and will be removed.

pracucci: Does the AWS SDK allows you to mock a service (SNS in this case)? Can we actually write a unit test on `Notify()`?

config/notifiers.go
treid314: Another option is to use the topic/target ARN to decide if the topic is FIFO or Standard. ARN's for topics that are FIFO will end in `.fifo`. Using the suffix to determine queue type is noted here in the aws docs https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html. I think using the topic/target arn would be the right path forward here rather than the config option.

notify/sns/sns.go
treid314: Yeah, I confirmed the group key hash isn't unique enough to handle a message with the same title but different labels send in the 5 minute SNS de-duplication window. I'll ask in the issue, we're also missing a message group id for FIFO queues that is required by AWS.

config/testdata/conf.sns-test.yml
treid314: We'll use it for a config test in config_test.go.

config/notifiers.go
jeromeinsf: What is the FIFO info used for?
Shouldn't the topic be created by the tenant and used by the receiver independently whether it is FIFO or not??

treid314: The FIFO information is used to decide if we need to send a FIFO required de-duplication key and message group id as part of the SNS publish action. But we can figure that out from the topic ARN.

config/notifiers.go
treid314: I'm not sure about changing the api version; the API Version is hard coded [in the SNS SDK client metadata currently.](https://github.com/aws/aws-sdk-go/blob/v1.38.35/service/sns/service.go#L66) Is this something we would want to change?

notify/sns/sns.go
treid314: We don't get a response code back in the event the publish is successful,[ for the final return ](https://github.com/treid314/alertmanager/blob/sns-reciever/notify/sns/sns.go#L159)should we be returning `false, nil` or `retrier(200, <response body from successful message>)`

config/notifiers.go
pracucci: > The FIFO information is used to decide if we need to send a FIFO required de-duplication key and message group id as part of the SNS publish action

What happens if we set it for standard queues too? Does the Publish API call fails?

config/notifiers.go
pracucci: > I'm not sure about changing the api version

What's part of the specifications. If you suggest to remove it, I would raise the topic in the issue.

Getting back to my comment: it was about the default value being `sns.default.api_version`. That's just a static string, but who does interpret it? I'm not sure I understand it.

notify/sns/sns.go
pracucci: > We don't get a response code back in the event the publish is successful

Is the error a custom type exposing extra information, like if it's retryable or not?

config/notifiers.go
```diff
@@ -590,11 +589,11 @@ func (c *PushoverConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
 	return nil
 }
 
-// TODO: Move to common?
 
 // SigV4Config is the configuration for signing remote write requests with
 // AWS's SigV4 verification process. Empty values will be retrieved using the
 // AWS default credentials chain.
+// TODO: Move to common.
```
pracucci, line 596: Can we make this happen?

notify/sns/sns.go
```diff
@@ -107,39 +137,35 @@ func (n Notifier) Notify(ctx context.Context, alert ...*types.Alert) (bool, erro
...
 	if err != nil {
 		return false, err
 	}
 
-	// Response is good and does not need to be retried.
 	return false, nil
 }
 
+func isFIFOTopic(topicARN string) bool {
+	if topicARN[len(topicARN)-5:] == ".fifo" {
```
pracucci, line 163: Not sure how much robust this check it.

notify/sns/sns.go
```diff
@@ -107,39 +137,35 @@ func (n Notifier) Notify(ctx context.Context, alert ...*types.Alert) (bool, erro
...
-		}
-		publishInput.SetMessageDeduplicationId(key.Hash())
+		publishInput.SetSubject(tmpl(n.conf.Subject))
 	}
 
 	publishOutput, err := client.Publish(publishInput)
 	if err != nil {
-		// AWS Response is bad, probably a config issue.
-		return false, err
+		return n.retrier.Check(err.(awserr.RequestFailure).StatusCode(), strings.NewReader(err.(awserr.RequestFailure).Message()))
```
pracucci, line 151: If the error is not a `awserr.RequestFailure` then it will panic. To have the code more robust, you should check for the data type first.

notify/sns/sns.go
```diff
@@ -41,17 +43,38 @@ type Notifier struct {
...
-	credentials := credentials.NewStaticCredentials(n.conf.Sigv4.AccessKey, string(n.conf.Sigv4.SecretKey), "")
-	if n.conf.Sigv4.AccessKey == "" {
-		credentials = nil
+	var(
+		err error
+		data = notify.GetTemplateData(ctx, n.tmpl, alert, n.logger)
+		tmpl = notify.TmplText(n.tmpl, data, &err)
+		creds *credentials.Credentials = nil
+	)
+	if n.conf.Sigv4.AccessKey != "" && n.conf.Sigv4.SecretKey != "" {
```
pracucci, line 69: I would suggest to have the config unmarshal failing if only one of the two is set.

treid314: Not sure if we can do that, since it is also valid to pull the credentials from environment variables and they would be empty when we unmarshal the config yml.

config/notifiers.go
treid314: Yes, if we send a de-duplication key to a standard type queue the publish fails with: `Invalid parameter: MessageDeduplicationId Reason: The request includes MessageDeduplicationId parameter that is not valid for this topic type`

notify/sns/sns.go
treid314: The error appears to be just `nil` returned from the sender in the success case and not wrapping anything.

notify/sns/sns.go
treid314: In the event this isn't an awserr, should we be using a retrier? or just `return true, err`

roidelapluie: just return. In the next few seconds, this would be retried by other instances in the HA pair.

config/notifiers.go
pracucci: Is using `GetTopicAttributes()` an option (and the cache it in-memory on a per ARN basis)? I think would be safer than parsing the ARN.

treid314: I switched the check to use `GetTopicAttributes()` but I couldn't find a good place to cache the result, any suggestions on where to cache Topics to isFifo booleans?

notify/sns/sns.go
```diff
@@ -0,0 +1,198 @@
...
+	err = n.logger.Log(publishOutput.String())
+	if err != nil {
+		return false, err
+	}
+
+	return false, nil
+}
+
+func checkTopicFifoAttribute(client *sns.SNS, topicARN string) (bool, error) {
+	fmt.Println("Checking Attributes")
```
pracucci, line 175: This should be removed :)

config/notifiers.go
pracucci: Discussed offline. We made it work. Will be fixed in the next commit.

config/notifiers.go
treid314: >  If you suggest to remove it, I would raise the topic in the issue.
I think since it's hard coded in the API, it makes the most sense to remove it. I'll make that change and add it to the issue.

notify/sns/sns.go
```diff
@@ -0,0 +1,200 @@
...
+	if n.conf.TargetARN != "" {
+		publishInput.SetTargetArn(tmpl(n.conf.TargetARN))
+		messageToSend, isTrunc, err := validateAndTruncateMessage(tmpl(n.conf.Message))
+		if err != nil {
+			return false, err
+		}
+		if isTrunc {
+			n.conf.Attributes["truncated"] = "true"
+		}
+		publishInput.SetMessage(messageToSend)
```
pracucci, line 145: This piece of code is shared with TopicARN. Also, I think the truncation logic for the PhoneNumber case can be based on `validateAndTruncateMessage()` too because the spec says "For SMS, each message can contain up to 140 characters" and assuming the ASCII encoding it's basically 1 byte per rune (`validateAndTruncateMessage()` truncates to bytes).

Given that, I think the message truncation logic + `publishInput.SetMessage()` can be generic. What we need to know for each case (TopicARN, TargetARN, PhoneNumber) is the max chars limit.

notify/sns/sns.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	if len(n.conf.Attributes) > 0 {
+		attributes := map[string]*sns.MessageAttributeValue{}
```
pracucci, line 149: Pre-allocate it with the right size.

```suggestion
		attributes := make(map[string]*sns.MessageAttributeValue{}, len(c.conf.Attributes)
```

notify/sns/sns.go
```diff
@@ -0,0 +1,200 @@
...
+	publishOutput, err := client.Publish(publishInput)
+	if err != nil {
+		if e, ok := err.(awserr.RequestFailure); ok {
+			return n.retrier.Check(e.StatusCode(), strings.NewReader(e.Message()))
+		} else {
+			return true, err
+		}
+	}
+
+	err = n.logger.Log(publishOutput.String())
```
pracucci, line 169: I guess you want to log the output, but that's not the right way.

You should log it with debug level and add a message too. Something like this:
`level.Debug(n.logger).Log("msg", "SNS notification successfully sent", "message_id", publishOutput.MessageId, "sequence_number", publishOutput.SequenceNumber)`

notify/sns/sns.go
```diff
@@ -0,0 +1,200 @@
...
+func checkTopicFifoAttribute(client *sns.SNS, topicARN string) (bool, error) {
+	topicAttributes, err := client.GetTopicAttributes(&sns.GetTopicAttributesInput{TopicArn: aws.String(topicARN)})
+	if err != nil {
+		return false, err
+	}
+	ta := topicAttributes.Attributes["FifoTopic"]
+	if ta != nil && *ta == "true" {
+		return true, nil
+	}
+	return false, nil
```
pracucci, line 186: This can be simplified with the `aws.BoolValue()` utility.

notify/sns/sns.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func validateAndTruncateMessage(message string) (string, bool, error) {
+	if utf8.ValidString(message) {
```
pracucci, line 190: [nit] You can reverse the logic using a guard here:

```suggestion
	if !utf8.ValidString(message) {
	    return "", false, fmt.Errorf("non utf8 encoded message string")
	}
```

notify/sns/sns.go
```diff
@@ -0,0 +1,200 @@
...
+		// if the message is larger than 256KB we have to truncate.
+		if len(message) > 256*1024 {
```
pracucci, line 192: A guard here too:

```suggestion
		if len(message) <= 256*1024 {
		    return message, false, nil
		}
```

notify/sns/sns.go
```diff
@@ -0,0 +1,200 @@
...
+			publishInput.SetMessageDeduplicationId(key.Hash())
+			publishInput.SetMessageGroupId(key.Hash())
+		}
+
+		publishInput.SetMessage(messageToSend)
+	}
+	if n.conf.PhoneNumber != "" {
+		publishInput.SetPhoneNumber(tmpl(n.conf.PhoneNumber))
+		// If SMS message is over 1600 chars, SNS will reject the message.
+		_, isTruncated := notify.Truncate(tmpl(n.conf.Message), 1600)
```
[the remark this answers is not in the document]
treid314: SNS will truncate to that for us, the publish limit is 1600 charactors

> If you publish a message that exceeds this size limit, Amazon SNS sends the message as multiple messages, each fitting within the size limit. Messages aren't truncated mid-word but are cut off at whole-word boundaries.
> 
> The total size limit for a single SMS Publish action is 1,600 characters.

config/notifiers.go
treid314: Once this is in I think we're good to remove that: https://github.com/prometheus/common/pull/309

notify/sns/sns_test.go
treid314: There's not a built in mocking in the AWS SDK. If I re-work the the way we create the publish input and move it to a function we could check that logic against the publish input we expect to generate. But I'm not seeing a way to test Notify() directly.

notify/sns/sns.go
treid314: We actually get a *string back from aws, but it makes sense to use the aws utility.

docs/configuration.md
```diff
@@ -691,6 +691,60 @@ value: <tmpl_string>
...
+[ message: <tmpl_string> | default = '{{ template "sns.default.message" .}}' ] 
+
+# SNS message attributes.
+attributes: 
+  [key : value]
+
+# The HTTP client's configuration.
+[ http_config: <http_config> | default = global.http_config ]
+```
+###`<sigv4_config>`
```
pracucci, line 731: [nit]
```suggestion

###`<sigv4_config>`
```

notify/sns/sns.go
```diff
@@ -0,0 +1,190 @@
...
+		data                           = notify.GetTemplateData(ctx, n.tmpl, alert, n.logger)
+		tmpl                           = notify.TmplText(n.tmpl, data, &err)
+		creds *credentials.Credentials = nil
+	)
+	if n.conf.Sigv4.AccessKey != "" && n.conf.Sigv4.SecretKey != "" {
+		creds = credentials.NewStaticCredentials(n.conf.Sigv4.AccessKey, string(n.conf.Sigv4.SecretKey), "")
+	}
+
+	attributes := make(map[string]*sns.MessageAttributeValue, len(n.conf.Attributes))
+	if len(n.conf.Attributes) > 0 {
```
pracucci, line 75: [nit] You can simplify the code removing this `if`. The `range n.conf.Attributes` will just skip iterating if `len(n.conf.Attributes) == 0`.

notify/sns/sns.go
```diff
@@ -0,0 +1,190 @@
...
+		for k, v := range n.conf.Attributes {
+			attributes[tmpl(k)] = &sns.MessageAttributeValue{DataType: aws.String("String"), StringValue: aws.String(tmpl(v))}
+		}
+
+	}
+
+	sess, err := session.NewSessionWithOptions(session.Options{
+		Config: aws.Config{
+			Region:      aws.String(tmpl(n.conf.Sigv4.Region)),
```
pracucci, line 84: Region is not documented as `<tmpl_string>` so either update the doc or remove the template parsing from here.

notify/sns/sns.go
```diff
@@ -0,0 +1,190 @@
...
+	if _, err := sess.Config.Credentials.Get(); err != nil {
+		return false, fmt.Errorf("could not get SigV4 credentials: %w", err)
+	}
+
+	if n.conf.Sigv4.RoleARN != "" {
+		sess.Config.Credentials = stscreds.NewCredentials(sess, n.conf.Sigv4.RoleARN)
+	}
+
+	// Max message size for a message in a SNS publish request is 256KB, except for SMS messages where the limit is 1600 characters/runes.
+	messageSize := 256 * 1024
```
pracucci, line 100: I would call it `messageSizeLimit` since that's what it is.

notify/sns/sns.go
```diff
@@ -0,0 +1,190 @@
...
+	client := sns.New(sess, &aws.Config{Credentials: creds})
+	publishInput := &sns.PublishInput{}
+
+	if n.conf.TopicARN != "" {
+		publishInput.SetTopicArn(tmpl(n.conf.TopicARN))
+
+		if n.isFifo == nil {
+			checkFifo, err := checkTopicFifoAttribute(client, n.conf.TopicARN)
```
pracucci, line 108: You have to evaluate the template here as well. You may do it once and then use the compiled one both for `SetTopicArn()` and `checkTopicFifoAttribute()`.

```suggestion
			checkFifo, err := checkTopicFifoAttribute(client, tmpl(n.conf.TopicARN))
```

notify/sns/sns.go
```diff
@@ -0,0 +1,190 @@
...
+		publishInput.SetTargetArn(tmpl(n.conf.TargetARN))
+
+	}
+
+	messageToSend, isTrunc, err := validateAndTruncateMessage(tmpl(n.conf.Message), messageSize)
+	if err != nil {
+		return false, err
+	}
+	if isTrunc {
+		attributes[tmpl("truncated")] = &sns.MessageAttributeValue{DataType: aws.String("String"), StringValue: aws.String(tmpl("true"))}
```
pracucci, line 143: Not sure I understand why you need to call `tmpl()` here.
```suggestion
		attributes["truncated"] = &sns.MessageAttributeValue{DataType: aws.String("String"), StringValue: aws.String("true")}
```

notify/sns/sns.go
```diff
@@ -0,0 +1,190 @@
...
+	publishOutput, err := client.Publish(publishInput)
+	if err != nil {
+		if e, ok := err.(awserr.RequestFailure); ok {
+			return n.retrier.Check(e.StatusCode(), strings.NewReader(e.Message()))
+		} else {
+			return true, err
+		}
+	}
+
+	level.Debug(n.logger).Log("msg", "SNS publish successfully sent", "message_id", publishOutput.MessageId, "sequence number", publishOutput.SequenceNumber)
```
pracucci, line 162: [nit]
```suggestion
	level.Debug(n.logger).Log("msg", "SNS message successfully sent", "message_id", publishOutput.MessageId, "sequence number", publishOutput.SequenceNumber)
```

notify/sns/sns.go
```diff
@@ -0,0 +1,190 @@
...
+func checkTopicFifoAttribute(client *sns.SNS, topicARN string) (bool, error) {
+	topicAttributes, err := client.GetTopicAttributes(&sns.GetTopicAttributesInput{TopicArn: aws.String(topicARN)})
+	if err != nil {
+		return false, err
+	}
+	ta := topicAttributes.Attributes["FifoTopic"]
+	if aws.StringValue(ta) == "true" {
+		return true, nil
+	}
+	return false, nil
```
pracucci, line 176: [nit] Can be simplified.

```suggestion
	return aws.StringValue(ta) == "true", nil
```

notify/sns/sns.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func validateAndTruncateMessage(message string, sizeInBytes int) (string, bool, error) {
```
pracucci, line 179: [nit] What we take in input is the limit
```suggestion
func validateAndTruncateMessage(message string, maxSizeInBytes int) (string, bool, error) {
```

pracucci (commented):
Good job! I think we're very close. I left few more comments. Mostly nits.

docs/configuration.md
```diff
@@ -691,6 +691,61 @@ value: <tmpl_string>
...
+
+# SNS message attributes.
+attributes: 
+  [key : value]
+
+# The HTTP client's configuration.
+[ http_config: <http_config> | default = global.http_config ]
+
+```
+###`<sigv4_config>`
```
pracucci, line 732: [nit] That's the suggestion I meant. Basically adding a newline before the `<sigv4_config>` title.

```suggestion
```

###`<sigv4_config>`
```

**Review by pracucci — APPROVED** [review]
Good job! The logic LGTM. From my perspective the only missing thing is getting the sigv4 config merged into prometheus/common and vendor it here.

File: config/notifiers.go (hunk 16 above)
Comment (treid314) [reply config/notifiers.go:595]: Ok, maybe with this one: https://github.com/prometheus/common/pull/310

File: notify/sns/sns.go [hunk 38]
```diff
@@ -0,0 +1,185 @@
...
+	attributes := make(map[string]*sns.MessageAttributeValue, len(n.conf.Attributes))
+	for k, v := range n.conf.Attributes {
+		attributes[tmpl(k)] = &sns.MessageAttributeValue{DataType: aws.String("String"), StringValue: aws.String(tmpl(v))}
+	}
+
+	sess, err := session.NewSessionWithOptions(session.Options{
+		Config: aws.Config{
+			Region:      aws.String(n.conf.Sigv4.Region),
+			Credentials: creds,
+			Endpoint:    aws.String(tmpl(n.conf.APIUrl)),
```
alanprot, line 83: This session is being used to call STS and SNS (and so we are calling both services with the same URL).

I think the intention was to configure the SNS endpoint. I think we can remove this config from here and set it on a new session to be used by the SNS client.

config/notifiers.go
```diff
@@ -579,3 +588,46 @@ func (c *PushoverConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
...
+	Profile   string `yaml:"profile,omitempty"`
+	RoleARN   string `yaml:"role_arn,omitempty"`
+}
+
+type SNSConfig struct {
+	NotifierConfig `yaml:",inline" json:",inline"`
+
+	HTTPConfig *commoncfg.HTTPClientConfig `yaml:"http_config,omitempty" json:"http_config,omitempty"`
+
+	APIUrl      string            `yaml:"api_url" json:"api_url"`
```
alanprot, line 609: Can we make this optional?

notify/sns/sns.go
```diff
@@ -0,0 +1,185 @@
...
+		},
+		Profile: n.conf.Sigv4.Profile,
+	})
+
+	if _, err := sess.Config.Credentials.Get(); err != nil {
+		return false, fmt.Errorf("could not get SigV4 credentials: %w", err)
+	}
```
alanprot, line 90: Do we need this? Seems that the result is being ignored and making an extra call to STS?

config/notifiers.go
alvinlin123: Had some chat with @treid314 offline. Here is my take on calling `GetTopicAttributes()` to tell if a topic is FIFO. 

While it works, but it has the drawback that:
1. it's extra API call
2. Customer has to explicitly give sender permission to call `GetTopicAttributes()` on top of `Publish()`, the `GetTopicAttribute()` returns a lot of detailed information on the topic (e.g. IAM policy, KMS Key ID) that customer may not want to expose to the sender. 

The `.fifo` suffix was employed by SNS for senders to easily tell the topic type without making extra API call. So I would suggest we just use the suffix to tell the topic type.

notify/sns/sns.go
```diff
@@ -0,0 +1,182 @@
...
+	if n.conf.Sigv4.AccessKey != "" && n.conf.Sigv4.SecretKey != "" {
+		creds = credentials.NewStaticCredentials(n.conf.Sigv4.AccessKey, string(n.conf.Sigv4.SecretKey), "")
+	}
+
+	attributes := make(map[string]*sns.MessageAttributeValue, len(n.conf.Attributes))
+	for k, v := range n.conf.Attributes {
+		attributes[tmpl(k)] = &sns.MessageAttributeValue{DataType: aws.String("String"), StringValue: aws.String(tmpl(v))}
+	}
+
+	sess, err := session.NewSessionWithOptions(session.Options{
```
cstyan, line 79: should you be checking the error here?

cstyan (approved):
Looks sane to me overall, though I've never used AWS SNS.

@simonpasquier @w0rm can either of you take a look?

notify/sns/sns.go
pracucci: Oh good point! We missed this error checking.

treid314: Good catch!

docs/configuration.md
```diff
@@ -693,6 +693,64 @@ value: <tmpl_string>
...
 ```
 
+## `<sns_configs>`
+```yaml
+# Whether or not to notify about resolved alerts.
+[ send_resolved: <boolean> | default = false ]
+
+# The SNS API URL i.e. https://sns.us-east-2.amazonaws.com.
+#  If not specified, the SNS API URL from the SNS SDK will be used.
+[api_url: <tmpl_string>]
```
roidelapluie, line 703:
```suggestion
[ api_url: <tmpl_string> ]
```

docs/configuration.md
```diff
@@ -693,6 +693,64 @@ value: <tmpl_string>
...
+
+# Configures AWS's Signature Verification 4 signing process to sign requests.
+sigv4:
+  [ <sigv4_config> ]
+  
+# SNS topic ARN, i.e. arn:aws:sns:us-east-2:698519295917:My-Topic
+# If you don't specify this value, you must specify a value for the phone_number or target_arn.
+# If you are using a FIFO SNS topic you should set a message group interval longer than 5 minutes
+# to prevent messages with the same group key being deduplicated by the SNS default deduplication window
+[ topic_arn: <tmpl_string>]
```
roidelapluie, line 713:
```suggestion
[ topic_arn: <tmpl_string> ]
```

docs/configuration.md
```diff
@@ -693,6 +693,64 @@ value: <tmpl_string>
...
+# The  mobile platform endpoint ARN if message is delivered via mobile notifications.
+# If you don't specify this value, you must specify a value for the topic_arn or phone_number.
+[ target_arn: <tmpl_string> ] 
+
+# The message content of the SNS notification.
+[ message: <tmpl_string> | default = '{{ template "sns.default.message" .}}' ] 
+
+# SNS message attributes.
+attributes: 
+  [key : value]
```
roidelapluie, line 731:
```suggestion
  [ <string>: <string> ... ]
```

as seen in `<oauth2>`

docs/configuration.md
```diff
@@ -693,6 +693,64 @@ value: <tmpl_string>
...
+
+# The HTTP client's configuration.
+[ http_config: <http_config> | default = global.http_config ]
```
roidelapluie, line 734: Should we have a global sigv4 config? anfd global api keys etc?

config/config_test.go
```diff
@@ -1020,6 +1020,20 @@ func TestSlackGlobalAPIURLFile(t *testing.T) {
...
+func TestValidSNSConfig(t *testing.T) {
+	_, err := LoadFile("testdata/conf.sns-topic-arn.yml")
+	if err != nil {
+		t.Fatalf("Error parsing %s: %s", "testdata/conf.sns-topic-arn.yml\"", err)
+	}
+}
+
+func TestInvalidSNSConfig(t *testing.T) {
+	_, err := LoadFile("testdata/conf.sns-invalid.yml")
+	if err == nil {
```
roidelapluie, line 1032: We should test the error message

notify/sns/sns.go
```diff
@@ -0,0 +1,189 @@
...
+	return &Notifier{
+		conf:    c,
+		tmpl:    t,
+		logger:  l,
+		client:  client,
+		retrier: &notify.Retrier{},
+	}, nil
+}
+
+func (n *Notifier) Notify(ctx context.Context, alert ...*types.Alert) (bool, error) {
```
roidelapluie, line 63: Is there a way to make this function more readable? For now this is quite long and has a lot of branches and indentation levels.

docs/configuration.md
treid314: I added add a global sigV4 config, for each field in the config the hierarchy in order of precedence is: 
1) ` sns.sigv4.< access_key | secret_key | region | profile | roleARN > `
2) `global.sigv4.< access_key | secret_key | region | profile | roleARN > `
3) environmental variables if the field is empty in both configs listed above

I think this is a little complicated, but it does follow the same pattern we use for the other global api keys (Slack, PagerDuty, VictorOps). I'm not sure if you have any ideas on how to make it clearer, or if this is a non-issue. The other option I can think of is flattening the global sigv4 object into `global.sigv4_access_key, global.sigv4_secret_key, ... `

roidelapluie:
did you see my comment about the notify method? can we split it to make it more readable/reviewable?

docs/configuration.md
treid314: We decided to only use the local sns.sigv4 config, and not add a global.

treid314:
@roidelapluie Thank you for the review! I updated the notify method to make it more readable. Let me know if you have any other comments!

alanprot (approved):

roidelapluie (approved):
Thanks! There would be some clean-up to be done on the imports and one comment does not start with a capital letter but it has no consequences on the code, so let's merge this and create a followup PR later.

kevinayres:
This is great!  You guys rock! Can't wait to try it.

just1900:
Hi, @treid314 I'm wondering if there is a way to send alerts to sns without ak/sk?

jeromeinsf:
@just1900 SNS does not support unauthenticated calls against its APIs

just1900:
> @just1900 SNS does not support unauthenticated calls against its APIs

Thanks for your reply, I'm currently running `alertmanager` under eks, I mean using the identity token instead of `ak/sk` directly.

jeromeinsf:
I see, so https://docs.aws.amazon.com/emr/latest/EMR-on-EKS-DevelopmentGuide/setting-up-enable-IAM.html would not help?

just1900:
I've configured serviceaccount with IAM role. This does work finally :)
I dive a little bit into the implementation of sns notifier and figure out how to send sns request with identity token. The final config would be like 
```yaml
  sns_configs:
  - topic_arn: arn:aws:sns:us-east-1:552185537300:am-test
#    api_url: https://sns.us-east-1.amazonaws.com # do not set this
    sigv4:
      region: us-east-1
#      role_arn: <role> # do not set this
```

The first thing that confuse me is that if setting the `role_arn`, aws sdk will try to assume the role of `role_arn` with a `AssumeRoleWithWebIdentity` role, and return an unauthorized error. I'm not sure how to bypass this. https://github.com/prometheus/alertmanager/blob/dbc86188c043352e13c4c812f373ad42d27082c6/notify/sns/sns.go#L119-L130

Then I get `role_arn` commented, however, alertmanager will use the `api_url` as an endpoint when retrieving sts credentials.
https://github.com/prometheus/alertmanager/blob/dbc86188c043352e13c4c812f373ad42d27082c6/notify/sns/sns.go#L102-L106

which may cause the AssumeRoleWithWebIdentity not work as documented in https://github.com/aws/aws-sdk-go/issues/3972.

PR #2618, opened 2021-06-15
Allow Reading Secrets from Files

sinkingpoint:
As discussed in #2498 , it would be nice to be able to read secrets from files. It's a rainy day here, so in this PR we introduce that support for all secrets in the global and individual notifier configs, based on the existing support for this in the SlackAPIKeyFile config var.

On a sort of personal node: This is useful for any alertmanager deployed in K8s as you can now use Kubernetes Secrets to manage individual config components rather than the whole config file

Comments/Criticism welcome - It's been a really interesting traipse around the codebase

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -397,8 +397,14 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 			if ec.AuthPassword == "" {
 				ec.AuthPassword = c.Global.SMTPAuthPassword
 			}
+			if ec.AuthPasswordFile == "" {
+				ec.AuthPasswordFile = c.Global.SMTPAuthPasswordFile
+			}
 			if ec.AuthSecret == "" {
 				ec.AuthSecret = c.Global.SMTPAuthSecret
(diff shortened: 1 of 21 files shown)


sinkingpoint:
Note that I mimicked the Slack precedent for this that loads the secret at notify time - in high turn over alertmanager instances this feels like overkill for a value that might never change. We should probably make this load the secrets at config unmarshal time and handle reloads nicely

roidelapluie:
The values should be read from file every time they are used. I wonder if we really need to have all this custom code or if we can pass http client credentials to achieve a lot of them.

I will have a look, but it will take a long time. What about we split this pull request per receiver? Maybe start with one to see if we have the right approach?

sinkingpoint:
Happy to close this and reopen with just one notifier. Probably Pagerduty because that's the one I care about the most. It seems that most every provider handles Auth differently, so as much as we can hope for a unified solution I don't think it's going to be that easy (as much as I would love to add a bearer token and call this done)

PR #2622, opened 2021-06-15
Alert store not calling callback when an alert is deleted

### LINKED ISSUE #2619: Metric `alertmanager_alerts` report incorrect number of alerts
**What did you do?**

Setup the alertmanager with clustering, sent alerts to _all_ Alertmanagers.

The `alertmanager_alerts` metric always converge across replicas.

As time passes the alertmanager alerts metric drift apart although no difference in API results - they are consistent. 

As you can see in the screenshot, as time progresses the `alertmanager_alerts` metric drift apart in the range of 10s per replica. 

Screenshot 2021-06-15 at 11 36 08

Verifying the API responses across replicas, I can see that no inconsistent results are being provided:

```
$ cat alertmanager-0 | jq '.data | length'
956
$ cat alertmanager-1 | jq '.data | length'
956
```

Looking at the metric implementation, we can see that the [`Marker` is responsible for giving us the information of the current alerts](https://github.com/prometheus/alertmanager/blob/master/types/types.go#L109-L148) but the in-memory alerting store coming from [memAlerts](https://github.com/prometheus/alertmanager/blob/master/provider/mem/mem.go#L80-L100) is responsible for setting the callback that will keep the marker synced with the current alerts held in the `Store`. However, it is very apparent that the `Store` would execute the callback [ONLY when we garbage collect](https://github.com/prometheus/alertmanager/blob/master/store/store.go#L84) and [NOT when we delete an alert directly](https://github.com/prometheus/alertmanager/blob/master/store/store.go#L110-L116).

We believe the fix is to also execute the callback when an alert is directly deleted.

https://github.com/prometheus/alertmanager/blob/58169c14126074bf45cce3e168641ede9eb23e47/store/store.go#L73-L116

gotjosh:
The garbage collection process within the store is in charge of
determining if an alert is resolved, deleting it, and then communicating
this back to the callback set.

When an alert was explicitly deleted, these were not being communicated
back to the callback and caused the metric to report incorrect results.

Fixes #2619

### DIFF
--- a/provider/mem/mem.go
+++ b/provider/mem/mem.go
@@ -77,7 +77,7 @@ func NewAlerts(ctx context.Context, m types.Marker, intervalGC time.Duration, al
 		logger:    log.With(l, "component", "provider"),
 		callback:  alertCallback,
 	}
-	a.alerts.SetGCCallback(func(alerts []*types.Alert) {
+	a.alerts.SetResolvedCallback(func(alerts []*types.Alert) {
 		for _, alert := range alerts {
 			// As we don't persist alerts, we no longer consider them after
 			// they are resolved. Alerts waiting for resolved notifications are
--- a/store/store_test.go
+++ b/store/store_test.go
@@ -15,6 +15,7 @@ package store
 
 import (
 	"context"
+	"sync"
 	"testing"
 	"time"
 
@@ -42,7 +43,6 @@ func TestDelete(t *testing.T) {
 		UpdatedAt: time.Now(),
 	}
 	require.NoError(t, a.Set(alert))
-
 	fp := alert.Fingerprint()
 
 	err := a.Delete(fp)
@@ -53,7 +53,8 @@ func TestDelete(t *testing.T) {
 	require.Equal(t, ErrNotFound, err)
 }
 
-func TestGC(t *testing.T) {
+func TestGCAndResolvedAlerts(t *testing.T) {
+	ac := &alertsCounter{}
 	now := time.Now()
 	newAlert := func(key string, start, end time.Duration) *types.Alert {
 		return &types.Alert{
@@ -72,21 +73,34 @@ func TestGC(t *testing.T) {
 		newAlert("a", -10, -5),
 		newAlert("d", -10, -1),
 	}
+	deleted := []*types.Alert{
+		newAlert("e", -10, 30),
+	}
 	s := NewAlerts()
 	var (
-		n           int
 		done        = make(chan struct{})
 		ctx, cancel = context.WithCancel(context.Background())
 	)
-	s.SetGCCallback(func(a []*types.Alert) {
-		n += len(a)
-		if n >= len(resolved) {
+	s.SetResolvedCallback(func(a []*types.Alert) {
+		ac.Inc(len(a))
+		if ac.Count() >= len(resolved) {
 			cancel()
 		}
 	})
-	for _, alert := range append(active, resolved...) {
+	// Merge all of alerts into a single slice.
+	all := []*types.Alert{}
+	all = append(all, active...)
+	all = append(all, resolved...)
+	all = append(all, deleted...)
+
+	for _, alert := range all {
 		require.NoError(t, s.Set(alert))
 	}
+
+	for _, alert := range deleted {
+		require.NoError(t, s.Delete(alert.Fingerprint()))
+	}
+
 	go func() {
 		s.Run(ctx, 10*time.Millisecond)
 		close(done)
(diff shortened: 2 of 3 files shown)


store/store_test.go
```diff
@@ -108,5 +122,24 @@ func TestGC(t *testing.T) {
...
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	c.counter += n
+}
+
+func (c *alertsCounter) Count() int {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
+	return c.counter
 }
```
pracucci, line 145: I think all of this can be simplified using an atomic int.

store/store_test.go
```diff
@@ -53,7 +53,8 @@ func TestDelete(t *testing.T) {
 	require.Equal(t, ErrNotFound, err)
 }
 
-func TestGC(t *testing.T) {
+func TestGCAndResolvedAlerts(t *testing.T) {
+	ac := &alertsCounter{}
```
pracucci, line 57: [nit] I think calling it `resolvedCount` may clarify.

store/store_test.go
```diff
@@ -108,5 +122,24 @@ func TestGC(t *testing.T) {
 			t.Errorf("alert %v should have been gc'd", alert)
 		}
 	}
-	require.Equal(t, len(resolved), n)
+	// If we manually delete an alert it should be sent back to the callback as it should be considered resolved.
+	// So in total, the callback should have received 3 alerts. 2 that were GC and 1 that was manually deleted.
+	require.Equal(t, 3, ac.Count())
```
pracucci, line 127: I would make it more explicit. This probably removes the need of the whole comment.
```suggestion
	require.Equal(t, len(resolved) + len(deleted), ac.Count())
```

provider/mem/mem.go
```diff
@@ -77,7 +77,7 @@ func NewAlerts(ctx context.Context, m types.Marker, intervalGC time.Duration, al
 		logger:    log.With(l, "component", "provider"),
 		callback:  alertCallback,
 	}
-	a.alerts.SetGCCallback(func(alerts []*types.Alert) {
+	a.alerts.SetResolvedCallback(func(alerts []*types.Alert) {
```
pracucci, line 80: A part from `m.Delete(alert.Fingerprint())`, is it correct to execute the rest of the logic for the case this callback is called because an explicit call to `Alerts.Delete()`? Could this lead to any side effect? Have you checked it?

gotjosh: I verified that `Delete` is only called from [within dispatch](https://github.com/prometheus/alertmanager/blob/master/dispatch/dispatch.go#L529-L533).

But I think you might be right, given we now have a chance of calling this callback more than once concurrently I should double-check that the operation within it is safe. I'll do that tomorrow and let you know.

gotjosh: Thanks for raising this, I've now taken a look. The way I understand it, an explicit call to `Alerts.Delete()` would only occur if (due to a timing issue) the alert would resolve **just before** we're about to send it.

https://github.com/prometheus/alertmanager/blob/fd0929ba9fc58737a9c91f24771862692fa72d17/dispatch/dispatch.go#L529-L533

While (to the best of my knowledge) the deletion of no longer used listeners at a more regular interval (each receiver/inhibitions subscribes to the alerts store) doesn't produce any side effects, I don't think a single alert resolving should trigger the garbage collection of this.

But, there's more... After testing more thoroughly I realised that this doesn't really solve what we wanted in the first place (accurate reporting of the `alertmanager_alerts`) because if an alert is resolved naturally it would be GCed until the next run anyway.

An example of this behaviour (with this change included) is that despite seeing 67 alerts in the UI and the metric reporting 72 alerts, it was not until the GC cycle ran (every 30 minutes by default) that I could see the accurate measurement:

Metrics
Screenshot 2021-06-23 at 18 18 15

API (all filters are enabled by default)
Screenshot 2021-06-23 at 18 20 07

Given there's no way to see resolved alerts through the UI, My proposal now is that we actually create a new label of "resolved" for this metric which would allow us to accurately track the value over time in accordance with what we see in the UI. 


With that we can do 

```promQL
sum(alertmanager_alerts{state=~"active|supressed"}) - sum(alertmanager_alerts{state="resolved"})
```
to match what we see in the UI.

roidelapluie:
The metrics in alertmanager are for pure operational reasons IIRC and are not meant to reflect the UI or the state of alertmanager. "Fixing" them might break users. (Note: I could be wrong)

roidelapluie:
My point is that if you want accurate metrics it might be better to implement a new collector.

gotjosh:
Still valid, I'll come back to this soon.

PR #2623, opened 2021-06-16
Allow reading Pagerduty secrets from files

sinkingpoint:
This commit adds two optional config values to pagerduty routing
configurations - service_key_file and routing_key_file. These function
the same as their non _file variants (and infact only one can be
specified), but allows reading the value from a file instead of
embedding the secret in the config file itself

c.c. @roidelapluie as discussed, this trims that PR down to just the Pagerduty stuff

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -46,6 +46,43 @@ func init() {
 	secretTokenJSON = string(b)
 }
 
+// ResolveFileConfig takes two values - a config value and a file path and resolves it to an absolute value
+// if the value is set, that gets returned, otherwise if the file path is set, we return the contents of the file
+// otherwise, we error
+func ResolveFileConfigSecret(configValue Secret, filePath string) (Secret, error) {
+	if configValue != "" || filePath == "" {
+		return configValue, nil
+	}
+
+	val, err := ioutil.ReadFile(filePath)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


sinkingpoint:
Partially solves #2618

roidelapluie:
I am wondering if we should add the extra helper function or not. It is really case by case if it should be read from file because we want to use the content of the file every time we make a request. So we could miss a place later on where we would use a "static" secret instead of a dynamic one.

sinkingpoint:
I think the helper function is still useful - itherwise we would have a bunch of redundant IO logic everywhere. There's nothing stopping you calling the helper every time you use the secret. Particularly in a world where all the secrets are file readable, having this "if file read file, otherwise use plain text secret" logic scattered about everywhere sounds bad when we can abstract it

gecgooden:
Hi, I'm interested in this change too. Is there anything that needs to be done for it to be ready to be reviewed, that I can help with?

mattchrist:
I'm also interested in this

d7volker:
I am also interested it this PR. I will ease the configuration of pagerduty, heavily.

simonpasquier:
Closed by https://github.com/prometheus/alertmanager/pull/3107

Issue #2630, opened 2021-06-22
labels.ParseMatcher allows unbalanced surrounding quotes


squat:
**What did you do?**
Using `labels.ParseMatcher` with strings that have unbalanced surrounding `"`s results in successful parsing, e.g. `labels.ParseMatcher("foo=\"bar")`.

**What did you expect to see?**
Error due to unbalanced surrounding quotes

**What did you see instead? Under which circumstances?**
Successful parsing of quotes

**Environment**
CI for prom-label-proxy when bumping dependencies (https://app.circleci.com/pipelines/github/prometheus-community/prom-label-proxy/100/workflows/aa8e2071-9392-43b9-8471-2505fc38b54f/jobs/231)

* Alertmanager version:
v0.20.0

This seems quite clear from the logic in the `ParseMatcher` func. We strip any opening surrounding quotes (https://github.com/prometheus/alertmanager/blob/99e4a4fbae5938f4fd31b8bc90755826f650d9da/pkg/labels/parse.go#L126) and do not enforce a matching quote (after all, this case is optional https://github.com/prometheus/alertmanager/blob/99e4a4fbae5938f4fd31b8bc90755826f650d9da/pkg/labels/parse.go#L159 ).

This parsing is pretty liberal and is a breaking change with the previous logic. It's somewhat tricky: the previous parser allowed a closing quote without a starting quote whereas this one allows unbalanced quotes in either position. Is it intended? If so let's close this and I can fix the test-case in prom-label-proxy to not expect a failure. Otherwise, maybe we want to take stance on the balancing.

bwplotka:
I am surprised we have this source code in alertmanager, not in prometheus common or something. Let me investigate if this was simply copied, and we can copy fixed version from github.com/prometheus/common or other spaces

bwplotka:
* common does not have matchers: https://github.com/prometheus/common/blob/4240322d2138024beddb57544851d77a08afe19c/model/silence.go
* Prometheus looks like have optimized matches, but they do not have parsing logic for string: https://github.com/prometheus/prometheus/blob/39d79c3cfb86c47d6bc06a9e9317af582f1833bb/pkg/labels/matcher.go#L44

*Side question:* Should we move matchers + parsing to common?

Fix: https://github.com/prometheus/alertmanager/pull/2632

beorn7:
[...]

PR #2632, opened 2021-06-23
matchers: Parse Matcher now expects consistent enclosing with quotes.

bwplotka:
Fixes https://github.com/prometheus/alertmanager/issues/2630

Signed-off-by: Bartlomiej Plotka <bwplotka@gmail.com>

### DIFF
--- a/pkg/labels/parse.go
+++ b/pkg/labels/parse.go
@@ -22,11 +22,9 @@ import (
 )
 
 var (
-	re = regexp.MustCompile(
-		// '=~' has to come before '=' because otherwise only the '='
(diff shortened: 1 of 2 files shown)


beorn7:
The current behavior was kind of deliberate, following the idea to tolerate as much as reasonably possible. But I don't feel strongly. Unbalanced quotes is certainly nothing that people would expect to work, and I can see the argument of unbalanced quotes pointing towards a mistake by the user we should make them aware of.

beorn7 (approved):

bwplotka:
Thanks!

PR #2634, opened 2021-06-25
Fix segmentation violation when IsEqual == nil

jmdacruz:
Running `amtool silence query` is currently broken with a segmentation fault. This PR adds a default value for IsEqual on the `labelsMatcher` function. The `model.Matcher` struct defines `IsEqual` as `omitempty`: 
https://github.com/prometheus/alertmanager/blob/fd0929ba9fc58737a9c91f24771862692fa72d17/api/v2/models/matcher.go#L35

which means that this pointer could be nil. The `labelsMatcher` function is not checking for nil, which causes a segmentation fault:

https://github.com/prometheus/alertmanager/blob/fd0929ba9fc58737a9c91f24771862692fa72d17/cli/format/format.go#L53-L67

### DIFF
--- a/cli/format/format.go
+++ b/cli/format/format.go
@@ -52,14 +52,22 @@ func FormatDate(input strfmt.DateTime) string {
 
 func labelsMatcher(m models.Matcher) *labels.Matcher {
 	var t labels.MatchType
+	isEqual := true
+	isRegex := false
+	if m.IsEqual != nil {
+		isEqual = *m.IsEqual
+	}
+	if m.IsRegex != nil {
+		isRegex = *m.IsRegex
+	}
 	switch {
-	case !*m.IsRegex && *m.IsEqual:
+	case !isRegex && isEqual:
 		t = labels.MatchEqual
-	case !*m.IsRegex && !*m.IsEqual:
+	case !isRegex && !isEqual:
 		t = labels.MatchNotEqual
-	case *m.IsRegex && *m.IsEqual:
+	case isRegex && isEqual:
 		t = labels.MatchRegexp
-	case *m.IsRegex && !*m.IsEqual:
+	case isRegex && !isEqual:
 		t = labels.MatchNotRegexp
 	}
 
--- a/cli/format/format_test.go
+++ b/cli/format/format_test.go
@@ -1,0 +1,95 @@
+// Copyright 2018 Prometheus Team
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package format
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/prometheus/alertmanager/api/v2/models"
+	"github.com/prometheus/alertmanager/pkg/labels"
+)
+
+func TestLabelsMatcher(t *testing.T) {
+	trueValue := true
+	falseValue := false
+
+	tests := map[string]struct {
+		value    string
+		input    models.Matcher
+		expected *labels.Matcher
+	}{
+		"IsRegex:true IsEqual:nil": {
+			value:    "some value",
+			input:    models.Matcher{IsEqual: nil, IsRegex: &trueValue},
+			expected: &labels.Matcher{Type: labels.MatchRegexp},
+		},
+		"IsRegex:true IsEqual:true": {
+			value:    "some value",
+			input:    models.Matcher{IsEqual: &trueValue, IsRegex: &trueValue},
+			expected: &labels.Matcher{Type: labels.MatchRegexp},
+		},
+		"IsRegex:true IsEqual:false": {
+			value:    "some value",
+			input:    models.Matcher{IsEqual: &falseValue, IsRegex: &trueValue},
+			expected: &labels.Matcher{Type: labels.MatchNotRegexp},
+		},
+
+		"IsRegex:false IsEqual:nil": {
+			value:    "some value",
+			input:    models.Matcher{IsEqual: nil, IsRegex: &falseValue},
+			expected: &labels.Matcher{Type: labels.MatchEqual},
+		},
+		"IsRegex:false IsEqual:true": {
+			value:    "some value",
+			input:    models.Matcher{IsEqual: &trueValue, IsRegex: &falseValue},
+			expected: &labels.Matcher{Type: labels.MatchEqual},
+		},
+		"IsRegex:false IsEqual:false": {
+			value:    "some value",
+			input:    models.Matcher{IsEqual: &falseValue, IsRegex: &falseValue},
+			expected: &labels.Matcher{Type: labels.MatchNotEqual},
+		},
+
+		"IsRegex:nil IsEqual:nil": {
+			value:    "some value",
+			input:    models.Matcher{IsEqual: nil, IsRegex: &falseValue},
+			expected: &labels.Matcher{Type: labels.MatchEqual},
+		},
+		"IsRegex:nil IsEqual:true": {
+			value:    "some value",
+			input:    models.Matcher{IsEqual: &trueValue, IsRegex: &falseValue},
+			expected: &labels.Matcher{Type: labels.MatchEqual},
+		},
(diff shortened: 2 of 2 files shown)


roidelapluie:
Thank you. How can we reproduce this? 1. in tests and 2. pratically when using the command line?

Thanks!

jmdacruz:
For (1), I would suggest a unit test such as this one (format_test.go`). I can add this to the PR if it sounds reasonable, and we can add coverage for the other cases (note that `IsRegex` could also be nil on this function, so we could fix that too):

```golang
package format

import (
	"reflect"
	"testing"

	"github.com/prometheus/alertmanager/api/v2/models"
	"github.com/prometheus/alertmanager/pkg/labels"
)

func TestTabelsMatcher(t *testing.T) {
	trueValue := true

	tests := map[string]struct {
		value    string
		input    models.Matcher
		expected *labels.Matcher
	}{
		"nil IsEqual": {
			value:    "some value",
			input:    models.Matcher{IsEqual: nil, IsRegex: &trueValue},
			expected: &labels.Matcher{Type: labels.MatchNotRegexp},
		},
	}

	for name, test := range tests {
		test.input.Name = &name
		test.expected.Name = name
		test.input.Value = &test.value
		test.expected.Value = test.value
		t.Run(name, func(t *testing.T) {
			got := labelsMatcher(test.input)
			if !reflect.DeepEqual(got, test.expected) {
				t.Fail()
			}
		})
	}

}
```

For (2), I was able to reproduce the issue by simply running `amtool --alertmanager.url=... silence query` in an environment where I don't have any sort of configuration file.

jmdacruz:
@roidelapluie I just amended the commit by also fixing the case of a nil value for `IsRegex`, and added a test for it

jmdacruz:
@roidelapluie This is interesting... I created the following `docker-compose.yml` file in order to test this, and I'm getting different results:

```yaml
version: "3.8"
services:
  alertmanager:
    image: prom/alertmanager:v0.21.0
    ports: 
    - 9093:9093
  test:
    image: prom/alertmanager:v0.22.2
    entrypoint: amtool
    depends_on:
    - alertmanager
    command:
    - --alertmanager.url=http://alertmanager:9093
    # - --alertmanager.url=https://alertmanager.my-internal-domain.com
    - silence
    - query
```

Version `v0.21.0` is the one I have on my internal server. I run this test file as follows:
```
# Bring up the alert manager instance on the docker-compose
docker-compose up -d alertmanager
# Run the test
docker-compose up test
```

Running the file with the first alert manager URL (the one launched by the docker-compose itself) works just fine, but when I replace it with the URL pointing to my internal instance I get the segmentation fault (I can reproduce this with both `v0.22.2` and `v0.21.0`):

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x8da822]

goroutine 1 [running]:
github.com/prometheus/alertmanager/cli/format.labelsMatcher(...)
      /app/cli/format/format.go:56
github.com/prometheus/alertmanager/cli/format.simpleFormatMatcher(0x0, 0xc0002b44d8, 0xc00064b250, 0xc00064b260, 0xc0000a45d0, 0x28)
      /app/cli/format/format_simple.go:98 +0x102
github.com/prometheus/alertmanager/cli/format.simpleFormatMatchers(0xc00026e120, 0x1, 0x4, 0x1, 0x1)
      /app/cli/format/format_simple.go:92 +0x98
github.com/prometheus/alertmanager/cli/format.(*SimpleFormatter).FormatSilences(0xc000323130, 0xc000386000, 0xb, 0x10, 0xc000330408, 0x134bc01)
      /app/cli/format/format_simple.go:48 +0x1de
github.com/prometheus/alertmanager/cli.(*silenceQueryCmd).query(0xc000485d40, 0xeb7de8, 0xc0004051a0, 0xc0005a4240, 0xc0004051a0, 0xc000403030)
      /app/cli/silence_query.go:143 +0x7fa
github.com/prometheus/alertmanager/cli.execWithTimeout.func1(0xc0005a4240, 0x0, 0x0)
      /app/cli/utils.go:154 +0xa4
gopkg.in/alecthomas/kingpin%2ev2.(*actionMixin).applyActions(0xc000162618, 0xc0005a4240, 0x0, 0x0)
      /go/pkg/mod/gopkg.in/alecthomas/kingpin.v2@v2.2.6/actions.go:28 +0x6d
gopkg.in/alecthomas/kingpin%2ev2.(*Application).applyActions(0xc0005a00f0, 0xc0005a4240, 0x0, 0x0)
      /go/pkg/mod/gopkg.in/alecthomas/kingpin.v2@v2.2.6/app.go:557 +0xdf
gopkg.in/alecthomas/kingpin%2ev2.(*Application).execute(0xc0005a00f0, 0xc0005a4240, 0xc00040e9e0, 0x2, 0x2, 0x0, 0x0, 0x0, 0x0)
      /go/pkg/mod/gopkg.in/alecthomas/kingpin.v2@v2.2.6/app.go:390 +0x95
gopkg.in/alecthomas/kingpin%2ev2.(*Application).Parse(0xc0005a00f0, 0xc000030110, 0x3, 0x3, 0x3, 0x0, 0x0, 0x10)
      /go/pkg/mod/gopkg.in/alecthomas/kingpin.v2@v2.2.6/app.go:222 +0x228
      /app/cli/root.go:122 +0x925
main.main()
      /app/cmd/amtool/main.go:19 +0x25
```

jmdacruz:
@roidelapluie I've updated my previous comment with the proper way of reproducing this issue

cli/format/format.go
```diff
@@ -52,14 +52,21 @@ func FormatDate(input strfmt.DateTime) string {
 
 func labelsMatcher(m models.Matcher) *labels.Matcher {
 	var t labels.MatchType
+	var isEqual, isRegex bool
+	if m.IsEqual != nil {
+		isEqual = *m.IsEqual
```
roidelapluie, line 57: isEqual should be equal to true

roidelapluie (commented):
Thanks, I have left a comment.

cli/format/format.go
jmdacruz: got it, fixed!

roidelapluie:
Now I am actually wondering if we should support it. You could create silences to 0.21 that would not be supported. So you would ask amtool to create a silence for env!=prod and it would create a silence env=prod.

I think we should to our best so that old clients can talk to new alertmanager, I do not think we should do the opposite.

What do you think?

roidelapluie:
So maybe instead of guessing or panicking we should error properly.

jmdacruz:
I think it makes sense @roidelapluie. I'm quite new to `amtool`, but it looks like in the long run it might be beneficial for the client to check the version of the server for compatibility (if there's a known breaking change). In the meantime, erroring out is probably the best alternative. I can give this a try on this PR.

That said, I noticed something interesting while trying different combinations, let me know if this behavior is expected: I modified my `docker-compose.yml` test file so that the version of alertmanager and amtool is the same using `TAG` as an environment variable for the image version, and I added a matcher on the test:

```yaml
version: "3.8"
services:
  alertmanager:
    image: prom/alertmanager:${TAG}
    ports: 
    - 9093:9093
  setup:
    image: prom/alertmanager:${TAG}
    entrypoint: amtool
    depends_on:
    - alertmanager
    command:
    - --alertmanager.url=http://alertmanager:9093
    - silence
    - add
    - --author=jmdacruz
    - --duration=90d
    - --comment="foobar"
    - namespace=~foobar
  test:
    image: prom/alertmanager:${TAG}
    entrypoint: amtool
    depends_on:
    - alertmanager
    command:
    - --alertmanager.url=http://alertmanager:9093
    - silence
    - query
    - namespace=~.+
```

I get two distinct results when running with version `v0.21.0` and `v0.22.2`:
```
export TAG=v0.21.0
docker-compose up -d alertmanager
docker-compose up setup
docker-compose up test
```
The above finds the silence:
```
ID                                    Matchers           Ends At                  Created By  Comment   
ffaa2b92-86ef-4569-b63b-94fd297991fb  namespace=~foobar  2021-10-04 17:06:34 UTC  jmdacruz    "foobar"
```

but with `v0.22.2` I get an empty result:
```
export TAG=v0.22.2
docker-compose up -d alertmanager
docker-compose up setup
docker-compose up test
```

roidelapluie:
Note: In the long run we will release a 1.0 release and the clients should be compatible across 1.0.

jmdacruz:
In that case, then maybe surfacing the error for now should be enough. I'll work on changing the PR, see what I can come up with.

roidelapluie:
The silence not found is expected. The silence has - namespace=~foobar, not - namespace=~.+
not sure why it works on previous releases

roidelapluie:
Done in #2668, thanks!

jmdacruz:
No problem! glad to be a contributor!

PR #2640, opened 2021-06-30
fix: silence mismatch by empty label matcher

bigHave:
If a label name does't exist in matchers, directly return false. 
otherwise, empty matcher might mismatch the label set

for example:
```
labelset := {foo="foo",bar="bar"} 
matchers := {foo="foo",baz=""}
```

matchers.Matches(labelset) == true
which is not expected.

this caused a mismatch problem in our production environment, all alerts had been silenced for about 2 hours until we found the reason: a silence that contains empty matcher

In my opinion, as empty matcher is necessary, a check before match could solve this problem 

appreciated for reviews and suggestions

Signed-off-by: wangyingtao <wangyingtao@bilibili.com>

### DIFF
--- a/pkg/labels/matcher.go
+++ b/pkg/labels/matcher.go
@@ -178,6 +178,17 @@ func (ms Matchers) Less(i, j int) bool {
 // Matches checks whether all matchers are fulfilled against the given label set.
 func (ms Matchers) Matches(lset model.LabelSet) bool {
 	for _, m := range ms {
+		if !m.Matches(string(lset[model.LabelName(m.Name)])) {
+			return false
+		}
+	}
+	return true
+}
+
+// MatchesSilence checks matchers against the given label set used by silence situation
+// to avoid empty matcher mismatch
+func (ms Matchers) MatchesSilence(lset model.LabelSet) bool {
+	for _, m := range ms {
 		value, ok := lset[model.LabelName(m.Name)]
 		if !ok {
 			return false
--- a/pkg/labels/matcher.go
+++ b/pkg/labels/matcher.go
@@ -178,7 +178,11 @@ func (ms Matchers) Less(i, j int) bool {
 // Matches checks whether all matchers are fulfilled against the given label set.
 func (ms Matchers) Matches(lset model.LabelSet) bool {
 	for _, m := range ms {
-		if !m.Matches(string(lset[model.LabelName(m.Name)])) {
+		value, ok := lset[model.LabelName(m.Name)]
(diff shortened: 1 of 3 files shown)


roidelapluie:
Hello, yes, that sounds like a bug.

Can you please add a test?

bigHave:
> Hello, yes, that sounds like a bug.
> 
> Can you please add a test?

sure, I'll add a test soon

bigHave:
I realized that matcher is used by other components other than silence

bigHave:
I realized that matchers is used by other components other than silence, so I think add a particular
```
func (ms Matchers) MatchesSilence(lset model.LabelSet) bool
```
would solve the problem in silence situation.

besides, I add a test for empty silence matcher situation

bigHave:
@roidelapluie hello, please take a look for further opinions.

roidelapluie:
Why is this limited to silences?

bigHave:
That's because empty matcher is used for router match and inhibit match, which make use of default empty matcher.

In a word, silences and router/inhibit expect different match result for the same matchers/labelset pair. (and that's why my first commit failed the ci test.)

roidelapluie:
I need to review this, thanks.

bigHave:
Hello, how's it going? 
appreciated for any news!:smiley:

roidelapluie:
It seems llike this would break the negative matcher case. I am investigating.

roidelapluie:
I actually think that the current behaviour is **correct**.

labelset := {foo="foo",bar="bar"} 
matchers := {foo="foo",baz=""}


This **does** match because baz="" and not having the baz label is **equivalent**. I do not see a bug here.

bigHave:
IMHO, there is a slightly difference between "empty" and "nonexistent", especially when it comes to silence scenario.  
For example, in our k8s environment(mainly test environment), apps can be deployed with different versions, known as "deploy_color".

roidelapluie:
> A label with an empty label value is considered equivalent to a label that does not exist.

https://prometheus.io/docs/concepts/data_model/

This is an essential part of Prometheus data model, and changing this is out of scope for the alertmanager.

roidelapluie:
Empty matchers *only* are disallowed. With another label, they are allowed. I already had to use empty matchers in production once.

What I would recommend is that you pick other labels for your silences, which would be more effective.

bigHave:
I see. Empty matchers with another label works fine, nothing need to be fixed then.
Thanks for your time! :smiley:

PR #2643, opened 2021-07-06
autorefresh Alertmanager UI every minute

Hayk96:
Allow Alertmanager to refresh every minute, to keep alerts up to date.

### DIFF
--- a/ui/app/index.html
+++ b/ui/app/index.html
@@ -2,6 +2,7 @@
 <html lang="en">
     <head>
         <meta charset="utf-8">
+        <meta http-equiv="refresh" content="60">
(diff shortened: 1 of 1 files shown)


SuperQ:
Sorry, this would cause sudden reloads while a user is editing a silence. This would be very bad.

Hayk96:
I agree with you. Thank you!
p.s. Having an auto-refresh feature will be good.

roidelapluie:
Most browsers have plugins to enable auto-refreshing a tab.

SuperQ:
Doing refersh via the javascript UI would probably be better, this way users can toggle it, and the UI can block refersh when editing.

Hayk96:
You're right @SuperQ, that's will be better.

PR #2648, opened 2021-07-13
Ensure time interval comparisons are in UTC

benridley:
The mute time mechanism uses the internal timer of alert aggregation groups to determine the current time, which is then used to determine whether a route should be muted. By default, this time fires in the local time of the server, not UTC as desired. 

This PR ensures comparisons are only made in UTC.

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -807,7 +807,7 @@ Loop:
 			return ctx, alerts, errors.Errorf("mute time %s doesn't exist in config", mtName)
 		}
 		for _, ti := range mt {
-			if ti.ContainsTime(now) {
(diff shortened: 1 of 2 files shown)


roidelapluie (approved):

roidelapluie:
thanks

benridley:
@roidelapluie Just wanted to make sure too that we mention this in the release notes for the next version. Some people might already have mute times setup that assume local time and this will break them :(.

roidelapluie:
This was released in 2.23

benridley:
Oh! I didn't realize we had a release already. No worries, at least this change makes the behavior consistent with the documentation so hopefully it won't surprise too many users.

PR #2663, opened 2021-08-01
Add ability to skip TLS verification for amtool

nedvna:
Just like with prometheus or alertmanager endpoints, we sometimes need to disable TLS certificate check. We can do that providing `<tls_config>` section there.  
amtool is a simpler thing but it does provide us some basic features like providing custom port and performing basic auth. I think skipping tls verification will fit in here too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,6 +315,9 @@ output: extended
 
 # Set a default receiver
 receiver: team-X-pager
+
+# Skip TLS certificate verification
+skip.verify: true
 ```
 
 ### Routes
--- a/README.md
+++ b/README.md
@@ -317,7 +317,7 @@ output: extended
 receiver: team-X-pager
 
 # Skip TLS certificate verification
-skip.verify: true
+tls.insecure.skip.verify: true
 ```
 
 ### Routes
--- a/README.md
+++ b/README.md
@@ -315,9 +315,6 @@ output: extended
 
 # Set a default receiver
 receiver: team-X-pager
-
-# Skip TLS certificate verification
-tls.insecure.skip.verify: true
 ```
 
 ### Routes
--- a/cli/root.go
+++ b/cli/root.go
@@ -81,7 +81,7 @@ func NewAlertmanagerClient(amURL *url.URL) *client.Alertmanager {
 
 	cr := clientruntime.New(address, path.Join(amURL.Path, defaultAmApiv2path), schemes)
 
-	if amURL.Scheme == "https" && skipVerify {
+	if skipVerify {
 		transport := http.Transport{
 			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
 		}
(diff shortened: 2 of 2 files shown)


cli/root.go
```diff
@@ -98,6 +108,7 @@ func Execute() {
 	app.Flag("alertmanager.url", "Alertmanager to talk to").URLVar(&alertmanagerURL)
 	app.Flag("output", "Output formatter (simple, extended, json)").Short('o').Default("simple").EnumVar(&output, "simple", "extended", "json")
 	app.Flag("timeout", "Timeout for the executed command").Default("30s").DurationVar(&timeout)
+	app.Flag("skip.verify", "Skip TLS certificate verification").BoolVar(&skipVerify)
```
roidelapluie, line 111:
```suggestion
	app.Flag("tls.insecure.skip.verify", "Skip TLS certificate verification").BoolVar(&skipVerify)
```

To match go tls name.

cli/root.go
```diff
@@ -78,6 +81,13 @@ func NewAlertmanagerClient(amURL *url.URL) *client.Alertmanager {
 
 	cr := clientruntime.New(address, path.Join(amURL.Path, defaultAmApiv2path), schemes)
 
+	if amURL.Scheme == "https" && skipVerify {
```
roidelapluie, line 84: Now I notice this. We should not check the scheme, but apply this all the time. You could go to an HTTP endpoint that redirects you (HTTP 302) to HTTPS.

nedvna: Didn't think we automatically followed them.  
Thanks, removed it.

cli/root.go
```diff
@@ -151,5 +162,9 @@ static configuration:
 
 	date.format
 		Sets the output format for dates. Defaults to "2006-01-02 15:04:05 MST"
+
+	skip.verify
```
roidelapluie, line 166: This was not updated

README.md
```diff
@@ -315,6 +315,9 @@ output: extended
 
 # Set a default receiver
 receiver: team-X-pager
+
+# Skip TLS certificate verification
+tls.insecure.skip.verify: true
```
roidelapluie, line 320: Let's get this out of the example configuration to keep it simple and not promote insecure settings.

cli/root.go
```diff
@@ -35,6 +37,7 @@ var (
 	alertmanagerURL *url.URL
 	output          string
 	timeout         time.Duration
+	skipVerify      bool
```
roidelapluie, line 40: Can we have tlsInsecureSkipVerify ?

roidelapluie (commented):
Sorry, I noticed one thing that was not updated, so I added comments from 2 other small things :) Then it will be OK :)

nedvna:
> Sorry, I noticed one thing that was not updated, so I added comments from 2 other small things :) Then it will be OK :)

No problem.  
No I hope I'm not breaking any more naming conventions :)

roidelapluie (approved):

roidelapluie:
Thanks!

PR #2665, opened 2021-08-03
Docs: target_match and source_match are DEPRECATED

7olstoy:
Via original docs https://prometheus.io/docs/alerting/latest/configuration/#inhibit_rule

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,9 +149,9 @@ route:
 # We use this to mute any warning-level notifications if the same alert is
 # already critical.
 inhibit_rules:
-- source_match:
+- source_matchers:
     severity: 'critical'
-  target_match:
+  target_matchers:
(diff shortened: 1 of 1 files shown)


roidelapluie:
Thanks! This does  not use the new matcherc, in the form of severity=warning etc.

7olstoy:
@roidelapluie Well, I guess we need a new one sample with not deprecated syntax. You can correct me, if I wrong.
This is the main reason why I opened PR.

roidelapluie:
yes, but this is an example with a mix between the old and the new syntax.

7olstoy:
@roidelapluie yeah, my bad. I just have looked at my working config. Check it.

roidelapluie:
Thanks!

PR #2672, opened 2021-08-06
amtool: Detect version drift and warn users

roidelapluie:
This change detects the alertmanager version when initiating the client.
It ignores most errors since I expect amtool to fail later.

If amtool is not compiled with proper version, we do not do anything
either.

We use MajorMinor for now as we have not reach 1.0, but we still allow
the bugfix version number (Z in x.y.Z) to differ.

### DIFF
--- a/cli/root.go
+++ b/cli/root.go
@@ -15,6 +15,7 @@ package cli
 
 import (
 	"crypto/tls"
+	"fmt"
 	"net/http"
 	"net/url"
 	"os"
@@ -23,6 +24,7 @@ import (
 
 	"github.com/go-openapi/strfmt"
 	"github.com/prometheus/common/version"
+	"golang.org/x/mod/semver"
 	kingpin "gopkg.in/alecthomas/kingpin.v2"
 
 	"github.com/prometheus/alertmanager/api/v2/client"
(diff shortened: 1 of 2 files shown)


cli/root.go
```diff
@@ -93,7 +95,19 @@ func NewAlertmanagerClient(amURL *url.URL) *client.Alertmanager {
 		cr.DefaultAuthentication = clientruntime.BasicAuth(amURL.User.Username(), password)
 	}
 
-	return client.New(cr, strfmt.Default)
+	c := client.New(cr, strfmt.Default)
+
```
gotjosh, line 99: Given this is now executed for _every_ command, can we an option to skip it? 

Something like:

```golang
var skipVersionCheck bool

app.Flag("skip.version-check", "Skip version checks when executing commands.").BoolVar(&skipVersionCheck)

// Then, we can do something like
if skipVersionCheck {
	return c
}

```

cli/root.go
```diff
@@ -93,7 +95,19 @@ func NewAlertmanagerClient(amURL *url.URL) *client.Alertmanager {
...
+	status, err := c.General.GetStatus(nil)
+	if err != nil || status.Payload.VersionInfo == nil || version.Version == "" {
+		// We can not get version info, or we do not know our own version. Let amtool continue.
```
gotjosh, line 102: Now that we have an option to cancel the check, I don't think we should fail silently. How do you feel about logging that we were unable to compute the version and tell the user that should upgrade/downgrade?

cli/root.go
```diff
@@ -93,7 +95,19 @@ func NewAlertmanagerClient(amURL *url.URL) *client.Alertmanager {
...
+		return c
+	}
+
+	if semver.MajorMinor("v"+*status.Payload.VersionInfo.Version) != semver.MajorMinor("v"+version.Version) {
```
gotjosh, line 106: This is quite nice 🎉

gotjosh (commented):
Approach LGTM, some quick questions on whenever we should have a way to make this optional.

cli/root.go
roidelapluie: If we can not run the check, it is because alertmanager does not work at all, and the error will be printed in the actual call later on.

roidelapluie:
updated

cli/root.go
gotjosh: I was thinking more in the [case of Grafana](https://github.com/grafana/grafana/blob/b78a67cec745b11b05ba2cf7ff12d848798b07a5/pkg/services/ngalert/api/tooling/definitions/alertmanager.go#L178-L197) but this is not a concern for us here to be honest.

gotjosh (approved):
LGTM.

```bash
$ ./amtool --alertmanager.url=http://localhost:9093 config show
Warning: amtool version (0.22.2) and alertmanager version (0.21.0) are different.

# and then

$ ./amtool --alertmanager.url=http://localhost:9093 config show --no-version-check
```

Both work as expected.

PR #2680, opened 2021-08-16
Fix some typo

machinly:
1. The `configuration.md` use both of `’` and `'`. I unified them as `'`.
2. I think `case-insentive` should probably be `case-insensitive`.
3. `sender's address.` makes a little more sense.


gotjosh (approved):
LGTM.

Thank you for your contribution!

roidelapluie (approved):

roidelapluie:
thanks!

Issue #1987, opened 2019-08-02
Feature request: support for dynamically determining responders in opsgenie configuration


hrancic:
Alertmanager v0.18.0 moved from `team` field to `responders` for alert routing in opsgenie. 

Having `team` field was useful for an use case where teams were determined dynamically after the alert fired, but before the alert was sent to opsgenie. If the list of teams responsible for the alert is not know before the alert is created, `responders` approach is not useful. 

Something along the lines of the following solution would be much appreciated:

```

#names contain comma separated list of teams

responders:
 - names: <tmpl_string>
   type: teams
```

iobestar:
any updates?

anandsinghkunwar:
@simonpasquier This seems like a very fair usecase/request and it was possible to do it before the migration. We've been using the previous alertmanager for the same reason, it'd be nice if we could get the same featureset in the newer version as well.

simonpasquier:
I agree but I'm not sure yet how it should translate in the configuration file. Splitting unconditionally on commas seems a bit rough.

MarkoRelic:
Any updates?

s3kkt:
We needs this feature too.
It will be good to determine responders list when alert is created.

MarkoRelic:
Can we expect this to be implemented?

multani:
Is there anything missing from #2685 that prevents this issue from being closed?

PR #2685, opened 2021-08-25
Dynamic opsgenie responders

markoposavec:
Hi!
I added an option to include multiple teams in alert as requested in this issue: https://github.com/prometheus/alertmanager/issues/1987

My current solution adds a new type option for Opsgenie responders - "teams" that are safe split by comma and then changed back into a list of teams. Another option would be to split the current "team" type by comma.
Any suggestions on how to approach this are welcome!


notify/opsgenie/opsgenie_test.go
```diff
@@ -112,7 +112,7 @@ func TestOpsGenie(t *testing.T) {
...
+			expectedBody: `{"alias":"6b86b273ff34fce19d6b804eff5a3f5747ada4eaa22f1d49c01e52ddb7875b4b","message":"message","description":"description","details":{"Description":"description","Message":"message","Note":"this is a note","Priority":"P1","ResponderName1":"TeamA","ResponderName2":"EscalationA","ResponderName3":"TeamA,TeamB","ResponderType1":"team","ResponderType2":"escalation","ResponderType3":"teams","Source":"http://prometheus","Tags":"tag1,tag2"},"source":"http://prometheus","responders":[{"name":"TeamA","type":"team"},{"name":"EscalationA","type":"escalation"}],"tags":["tag1","tag2"],"note":"this is a note","priority":"P1"}
```
roidelapluie, line 115:
```suggestion
			expectedBody: `{"alias":"6b86b273ff34fce19d6b804eff5a3f5747ada4eaa22f1d49c01e52ddb7875b4b","message":"message","description":"description","details":{"Description":"description","Message":"message","Note":"this is a note","Priority":"P1","ResponderName1":"TeamA","ResponderName2":"EscalationA","ResponderName3":"TeamA,TeamB","ResponderType1":"team","ResponderType2":"escalation","ResponderType3":"team","Source":"http://prometheus","Tags":"tag1,tag2"},"source":"http://prometheus","responders":[{"name":"TeamA","type":"team"},{"name":"EscalationA","type":"escalation"}],"tags":["tag1","tag2"],"note":"this is a note","priority":"P1"}
```

i am not sure why it is teams and not team here.

markoposavec: Hi, that "teams" is the label value used to create a responder type in tests - so in the first part of the body, there are label values, and in the second part, you can see that those teams become two different responders. This is an approach with adding a new type "teams," as I mentioned in the initial commit.
Another approach is to put all of this logic under "team" type but I think this is better because "team" doesn't have to be split by comma. Should I change PR to "team"?

markoposavec: @roidelapluie can you take a look again please?

roidelapluie (approved):

roidelapluie:
Thanks!

WockaFlockaRist:
This has been sought after by my **teams**, awesome. Any idea when this will be released?

PR #2689, opened 2021-09-01
Enable support for custom callbacks as part of maintenance

gotjosh:
This enables support for custom Maintenance callbacks as part of the periodic maintenance of silences and notification logs.
Effectively a no-op for the Alertmanager but allows a downstream implementation to inject custom logic as part of it.

- [x] Does this requires a changelog? Given it's a no-op for anything alertmanager related.

### DIFF
--- a/nflog/nflog.go
+++ b/nflog/nflog.go
@@ -88,6 +88,7 @@ type Log struct {
 	mtx       sync.RWMutex
 	st        state
 	broadcast func([]byte)
+	mf        func() (int64, error)
 }
 
 type metrics struct {
@@ -190,7 +191,7 @@ func WithMetrics(r prometheus.Registerer) Option {
 //
 // The maintenance terminates on receiving from the provided channel.
 // The done function is called after the final snapshot was completed.
-func WithMaintenance(d time.Duration, stopc chan struct{}, done func()) Option {
+func WithMaintenance(d time.Duration, stopc chan struct{}, done func(), mf func() (int64, error)) Option {
 	return func(l *Log) error {
 		if d == 0 {
 			return errors.New("maintenance interval must not be 0")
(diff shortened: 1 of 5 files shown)


roidelapluie:
=== RUN   TestWithMaintenance_SupportsCustomCallback
==================
WARNING: DATA RACE

nflog/nflog.go
```diff
@@ -325,34 +327,39 @@ func (l *Log) run() {
...
+
+	if l.mf != nil {
+		inner = l.mf
+	}
+
+	if l.done != nil {
+		defer l.done()
+	}
+
+	f := func(cb func() (int64, error)) error {
```
gouthamve, line 356: `f` --> `runMaintainance`

silence/silence.go
```diff
@@ -349,34 +349,40 @@ func New(o Options) (*Silences, error) {
 // Maintenance garbage collects the silence state at the given interval. If the snapshot
 // file is set, a snapshot is written to it afterwards.
 // Terminates on receiving from stopc.
-func (s *Silences) Maintenance(interval time.Duration, snapf string, stopc <-chan struct{}) {
+func (s *Silences) Maintenance(interval time.Duration, snapf string, stopc <-chan struct{}, mf func() (int64, error)) {
```
gouthamve, line 352: Same comments on this file.

gouthamve (commented):
Looks good, but omg the code is hard to grok due to the small variable names. Can we make that better and document in the function comment that this new parameter is optional and advanced usage?

gouthamve (approved):
LGTM!

nflog/nflog_test.go
```diff
@@ -15,9 +15,11 @@ package nflog
 
 import (
 	"bytes"
+	"github.com/prometheus/client_golang/prometheus"
```
codesome, line 18: This should be grouped with the below imports

codesome (approved):
LGTM with a nit :)

juliusv:
This introduced a crash bug for the default silence maintenance function, see fix in https://github.com/prometheus/alertmanager/pull/2701.

PR #2694, opened 2021-09-07
export ValidateMatcher for DI

kylebrandt:
so third parties, Grafana in particular, can over ride the validation.

Grafana wants to do this because other data sources will have label keys with things like spaces, periods, or other characters - and looking for a better integration with alert manager.

goes with https://github.com/grafana/grafana/pull/38629


silence/silence.go
```diff
@@ -425,7 +425,10 @@ func (s *Silences) GC() (int, error) {
 	return n, nil
 }
 
-func validateMatcher(m *pb.Matcher) error {
+// ValidateMatcher runs validation on the matcher Name, type, and pattern.
+// It is exported so it can be changed via dependency injection for third party
```
roidelapluie, line 429: No need to explain why it's exported.

config/config.go
```diff
@@ -701,11 +702,11 @@ func (r *Route) UnmarshalYAML(unmarshal func(interface{}) error) error {
 		if l == "..." {
 			r.GroupByAll = true
 		} else {
-			labelName := model.LabelName(l)
-			if !labelName.IsValid() {
-				return fmt.Errorf("invalid label name %q in group_by list", l)
-			}
-			r.GroupBy = append(r.GroupBy, labelName)
+			// labelName := model.LabelName(l)
```
kylebrandt, line 705: @gotjosh Do you have a recommendation for how to disable this as some sort of code option? (instead of commenting it out :-) )

roidelapluie:
it is really unclear to me what direction grafana is taking. At some point, it might be better to just use a fork of alertmanager rather than adding abstraction layers over all the code...

i think that you could just work with the alertmanager objects, preprare them before passing them here.

it will be *very* confusing for users if grafana's "alertmanager" behaves differently than alertmanager.

roidelapluie:
Especially since Alertmanager has not reached 1.x yet. If we miss some concepts or if there is room for improvements, we could discuss and possibly integrate those features directly in alertmanager :)

kylebrandt:
@roidelapluie Sorry I didn't mean to make a draft PR in this repo yet (meant to use Grafana's but had to many GH tabs), I can close if it is a distraction - as this is still a bit exploratory (just let me know).

The objective with this branch is to allow Label Keys to be able to contain characters beyond the Open API / Prometheus label key restrictions. This is to make so alert manager can be used with other TSDBs (or things acting like TSDBs). Or more on point, other data sources that Grafana supports.

So far in my exploration this seems to work if I: 
 1) Bypass label key validation
 2) Take the matchers in as objects in the config so they do not need to be parsed (like the silence pb already seems to be).

@gotjosh is far more familiar with the alert manage side as well so can maybe add some context here.

roidelapluie:
I have boostrapped a discussion on the grafana-developers mailing list.

config/config_test.go
```diff
@@ -1127,3 +1129,26 @@ func TestNilRegexp(t *testing.T) {
...
+	cfg, err := Load(in)
+	require.NoError(t, err)
+
+	m1, err := labels.NewMatcher(labels.MatchRegexp, "a c", ".+")
+	require.NoError(t, err)
+	m2, err := labels.NewMatcher(labels.MatchEqual, "a.c", "z")
+	require.NoError(t, err)
+
+	expectedMatchers := ObjectMatchers{m1, m2}
+	spew.Dump(cfg)
```
kylebrandt, line 1152: remove

kylebrandt:
replaced by https://github.com/prometheus/alertmanager/pull/2716

PR #2701, opened 2021-09-13
Correctly call default silence maintenance function

juliusv:
https://github.com/prometheus/alertmanager/pull/2689 introduced a
regression where the default maintenance function would no longer be
called even if no override was specified. The Alertmanager now crashes
on any silence maintenance run without this fix.


codesome (approved):
Thanks! LGTM

gotjosh:
Thanks for picking this up @juliusv.

roidelapluie:
Looks like it did not hit a stable release.

juliusv:
Yep, luckily not, all good :)

PR #2702, opened 2021-09-13
Update configuration.md

creganFL:
Clarified behaviour of default route
@simonpasquier @w0rm
Signed-off-by: Corey Regan <cregan@freelancer.com>

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -130,9 +130,11 @@ must match all alerts (i.e. not have any configured matchers).
 It then traverses the child nodes. If `continue` is set to false, it stops
 after the first matching child. If `continue` is true on a matching node, the
 alert will continue matching against subsequent siblings.
+If an alert matches a child node with `continue` set to true, even if it is the
+last sub-route, it will not trigger the default route. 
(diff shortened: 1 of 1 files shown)


docs/configuration.md
```diff
@@ -130,9 +130,11 @@ must match all alerts (i.e. not have any configured matchers).
 It then traverses the child nodes. If `continue` is set to false, it stops
 after the first matching child. If `continue` is true on a matching node, the
 alert will continue matching against subsequent siblings.
+If an alert matches a child node with `continue` set to true, even if it is the
+last sub-route, it will not trigger the default route.
```
roidelapluie, line 134: This behavior is correct but it is the same as with any other routes. I think we should not add this wording, because this is what happens with any other route. We should not let readers believe that the default route is "special". (or do we have a bug?)

creganFL:
I'm just trying to save somebody else hours of troubleshooting. In my new
role I am interacting with Prometheus/Alertmanager for the first time, and
it was not immediately obvious that the default route will never be
triggered if a sub-route matches. It makes sense after the fact, but in
that moment it did not.

SuperQ:

PR #2706, opened 2021-09-14
Fix HTTP client options in SNS receiver

pracucci:
The HTTP client options passed to SNS receiver are ignored because the HTTP client is never passed to the SNS client. This PR fixes it.


roidelapluie:
Thanks!

PR #2707, opened 2021-09-14
Mattermost channel with upper case letter not accepted

igaskin:
feel free to close as "won't fix".  This is arguably user-error.


roidelapluie:
Is this also the case with slack? I think we can not know what future updates of those recipients should accept. unless it is documented to be upper case, I don't think it belongs here. Thanks!

igaskin:
Yes, I believe this is also the case with Slack, since Mattermost uses the same incoming webhook format as Slack.

I'll go ahead and close this then.  Thanks @roidelapluie

PR #2709, opened 2021-09-15
Update xxhash to v2.1.1 and improve pooling

bboreham:
I started on this because I didn't like to see my program linking two versions of xxhash.
The way it's used in Alertmanager it makes no practical difference; I just wanted to tidy up.

Then, while I was looking at it, I decided the buffer pooling could be made more idiomatic.
By putting pointer-valued objects into the pool we avoid boxing a slice value; this saves one memory allocation every time.

I added a benchmark to illustrate:
```
name         old time/op    new time/op    delta
HashAlert-4     529ns ± 1%     490ns ± 2%   -7.29%  (p=0.008 n=5+5)

name         old alloc/op   new alloc/op   delta
HashAlert-4     96.0B ± 0%     72.0B ± 0%  -25.00%  (p=0.008 n=5+5)

name         old allocs/op  new allocs/op  delta
HashAlert-4      3.00 ± 0%      2.00 ± 0%  -33.33%  (p=0.008 n=5+5)
```

(I also tried re-coding using `xxhash.New()`, but that is slower, probably because it uses the Go implementation and not the assembly one).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ require (
 	github.com/alecthomas/units v0.0.0-20210208195552-ff826a37aa15
 	github.com/aws/aws-sdk-go v1.40.11
 	github.com/cenkalti/backoff/v4 v4.1.1
-	github.com/cespare/xxhash v1.1.0
+	github.com/cespare/xxhash/v2 v2.1.1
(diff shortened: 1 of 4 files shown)


roidelapluie:
thanks!

PR #2710, opened 2021-09-16
Close file descriptor after snapshot file was read

yuri-tceretian:
Fixes a tiny bug in the constructor of `Silences`. If options specify a path to snapshot file the constructor opens it and it is never closed.
The proposed change closes the file descriptor at the end of the constructor.


silence/silence.go
```diff
@@ -327,6 +327,9 @@ func New(o Options) (*Silences, error) {
 			}
 		} else {
 			o.SnapshotReader = r
+			defer func(fileDescriptor *os.File) {
```
gerobinson, line 330: I think this could be just `defer r.Close()` or does that fail linting because the error is not checked?

yuri-tceretian: It could, yes. I think this is more a question of style. I think this is more explicit.

gotjosh (approved):
LGTM

PR #2711, opened 2021-09-17
Add slack.NewWithErrDetails function

plauko:
This allows passing in a handler function for custom formatting of slack error messages when alertmanager is used as a library.

cc @simonpasquier, @w0rm


roidelapluie:
Error messages should be a detail. What are we doing wrong? We generally do not add dead code.

plauko:
You do all right :) It's Slack that returns a huge 404 page which our ui cannot tolerate. Having able to pass in a custom handler that trims the message on 404 errors would be more elegant than the regex magic we have now.

roidelapluie:
we do not add dead code as it is a cost for us. It also happens that this seems quite niche. is this normal slack behaviour or because of a user's misconfiguration?

roidelapluie:
to clarify: there are little chances that we accept this pull request as is, but there could be other ways and we could find a solution that works for you and us.

plauko:
Slack 404 is caused by misconfiguration, but returning a http page this case is normal behaviour.
We have a feature to allow testing the notification config the user entered in the UI on the fly, before deploying it to alertmanager, and this is where we use AM as library, and this is where this type of error is disturbing. I admit this is niche use case, yet having a configurable error handler can be useful to others too.
I chose to introduce a new constructor because adding errDetails or the retrier to the existing New felt a worse pick to me. IMO, considering AM can be used as a library this 2nd constructor is not dead code, but an additional option to instantiate a notifier, but I'm happy to change this pr in a way that better suits all our needs.
How about adding errDetails to config.SlackConfig? Or do you have any better in mind?

SoloJacobs (commented):
Hi @plauko ,

as already stated, we can't accept this PR. Sorry you did not receive a response to your question in a timely manor. It seems that you are already discussing the implementation details here. A better approach would be to open an issue and describe in detail the feature you need. From there we can discuss whether something like that is suitable for the project.

Personally, I feel that customizing error messages for Slack misconfiguration should be done by forking alertmanager, but I am to be convinced otherwise.

Kind regards, Solomon Jacobs

PR #2716, opened 2021-09-23
Export ValidateMatcher for DI (#2)

kylebrandt:
So third parties, Grafana in particular, can override the validation.

Grafana wants to do this because other data sources will have label keys with things like spaces, periods, or other characters - and looking for a better integration with alert manager.

goes with grafana/grafana#38629
replaces https://github.com/prometheus/alertmanager/pull/2694 with more minimal changes, as most of that was able to be moved into Grafana's code base.


gotjosh (approved):
LGTM

roidelapluie:
is that the only way to do this? exporting a function as global variable is not that great.

kylebrandt:
> is that the only way to do this? exporting a function as global variable is not that great.

Perhaps it come be done so `ValidateMatcherFunc func(m *pb.Matcher) error` is added to the `Options` struct. Then it could be passed as private property on the `Silences` struct, and `validateSilence` could take the function as an additional argument. If it is `nil` call the builtin `validateMatcher`, if not nil call what was passed.

Does that sound better?

PR #2730, opened 2021-10-01
Generalize User-Agent header in receiver requests

theag3nt:
The request helper functions in the `notify` package now set the user agent for all requests. All applicable receivers were transitioned to these helpers, except opsgenie. In the case of opsgenie where requests are pre-generated the user agent is set before sending each request.


roidelapluie:
Thanks!

Issue #2449, opened 2021-01-07
Update alertmanager docs to specify legacy slack webhook integration. 


vanugrah:
Hey team, 

Just wanted to bring an issue to awareness. The link titled "Slack webhooks" the points to https://api.slack.com/incoming-webhooks in the [alertmanager docs](https://prometheus.io/docs/alerting/latest/configuration/#slack_config) now gets redirected to https://api.slack.com/messaging/webhooks instead. 

Although this may not seem major, this is actually the new method that requires creating a custom slack app. There is a serious limitation in that this method only supports sending messages to one slack channel per webhook URI. 

The legacy method was in fact to use a standard app in the app directory called "Incoming WebHooks" which supports the use case of sending alerts to multiple channels with a single webhook URI.

PR #2749, opened 2021-10-28
Update outdated url link with slack

zhyChesterCheung:
Related: #2449 
- Update the outdated url link with slack in document.


roidelapluie (approved):
Thanks!

PR #2758, opened 2021-11-07
Document that matchers are ANDed together

mac-chaffee:
When trying to create more complex matchers, I wasn't sure whether the alert's labels had to match all matchers (AND) or just one (OR) in order to be sent. Looks like the source code indicates matchers are always ANDed together anywhere the `Matchers` type is used: https://github.com/prometheus/alertmanager/blob/fd0929ba9fc58737a9c91f24771862692fa72d17/pkg/labels/matcher.go#L179-L185

Also I see my editor trimmed trailing whitespace. I figure that's desirable, but I can undo it if need-be.

It does feel like "ANDed" is a weird word, but I guess it's somewhat common: https://english.stackexchange.com/a/79744

If there's a less obscure way to describe logical AND, I'm open to changing it.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -161,8 +161,8 @@ match:
 # A set of regex-matchers an alert has to fulfill to match the node.
 match_re:
   [ <labelname>: <regex>, ... ]
-  
-# A list of matchers that an alert has to fulfill to match the node. 
+
+# A list of matchers (ANDed together) that an alert has to fulfill to match the node.
 matchers:
   [ - <matcher> ... ]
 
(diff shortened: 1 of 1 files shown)


roidelapluie:
I also feel like ANDed is a strange word. The current wording "A list of matchers that an alert has to fulfill to match the node." says that you must match the matchers in the list already, I am unsure this reword is needed.

mac-chaffee:
When I read it, my thinking is that "fulfill" doesn't necessarily imply "AND" or "OR". 

One could say that an expression like `x AND y` is fulfilled by `x=1,y=1`, just like `x OR y` is fulfilled by `x=0,y=1`. The "list of matchers" would be the variables x and y.

Maybe an alternate phrasing could be "A list of matchers that must all evaluate to true for an alert to match the node"?

roidelapluie:
`A list of all matchers that an alert has to fulfill to match the node.` maybe?

mac-chaffee:
I guess I'm not too concerned about that specific sentence changing, as long as there is a re-stating of it somewhere that's more explicit.

Guess we could remove the "(ANDed together) from the other sections, but leave the extra sentence and example in the "matchers" section, since they're linked together anyway?

roidelapluie:
Yes that sounds better.

mac-chaffee:
Following up on this PR :)

simonpasquier (approved):
LGTM, thanks!

Issue #2762, opened 2021-11-09
Upgrading Client Lib to API v2?


sinkingpoint:
**What did you do?**

Used the client library in this repo (https://github.com/prometheus/alertmanager/blob/main/client/client.go)

**What did you expect to see?**

It use the v2 api

**What did you see instead? Under which circumstances?**

It's hardcoded to use the v1 api (https://github.com/prometheus/alertmanager/blob/main/client/client.go#L32)

Is this maintained? I note that it hasn't been touched in a year and a bit. Considering the expected deprecation of the v1 api we should probably update it or deprecate it along with the API (I would be in favour of the former)

roidelapluie:
The v2 client is https://github.com/prometheus/alertmanager/blob/main/api/v2/client/alertmanager_client.go

roidelapluie:
Would you send a PR to remove the v1 client?

sinkingpoint:
Ah. Good to know. Yes sure

PR #2763, opened 2021-11-09
Remove v1 client lib

sinkingpoint:
This is confusing as it sits in the root of the repo. Considering we
have a v2 client lib at (https://github.com/prometheus/alertmanager/blob/main/api/v2/client/alertmanager_client.go) this commit removes the v1 one to try and avoid
that confusion

Solves #2762

### DIFF
--- a/test/with_api_v1/acceptance.go
+++ b/test/with_api_v1/acceptance.go
@@ -31,8 +31,6 @@ import (
 
 	"github.com/prometheus/client_golang/api"
 	"github.com/prometheus/common/model"
-
-	"github.com/prometheus/alertmanager/client"
 )
 
 // AcceptanceTest provides declarative definition of given inputs and expected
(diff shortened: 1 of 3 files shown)


sinkingpoint:
Erk... This seems to be more ingrained that I anticipated. Not sure whether we want to do this @roidelapluie considering this might have a bit of an impact

roidelapluie:
This is indeed used in the acceptance tests.

What about we move client and client_test.go inside test/with_api_v1/acceptance.go and merge them into a file called helper_test.go ?

test/with_api_v1/acceptance.go
35:     "github.com/prometheus/alertmanager/client"

sinkingpoint:
Sounds reasonable. The source of this confusion comes from the prominence of this lib in the repo, so moving this out of that location makes sense

sinkingpoint:
Build error looks transient? 

```
error pulling image configuration: Get https://docker-images-prod.s3.dualstack.us-east-1.amazonaws.com/registry-...: dial tcp 52.217.66.96:443: i/o timeout
```

roidelapluie (commented):
I think only the client part shold go to helper.go, acceptance should stay in acceptance.go

sinkingpoint:
Fixed this. Yet again, the build failures appear to be docker pull timeouts which is weird

roidelapluie:
Thanks!

PR #2764, opened 2021-11-10
amtool to support http_config to access alertmanager

clyang82:
This PR is to enable amtool using bearerToken to access alertmanager. the use case is expose the alertmanager api via openshift route or k8s ingress, the external amtool can access the exposed alertmanager by providing bearerToken.

Signed-off-by: clyang82 <chuyang@redhat.com>


roidelapluie:
We should error if the user specifies multiple authentication mechanisms

cli/root.go
```diff
@@ -91,11 +92,19 @@ func NewAlertmanagerClient(amURL *url.URL) *client.Alertmanager {
 		cr.Transport = &transport
 	}
 
+	if amURL.User != nil && bearerToken != "" {
+		kingpin.Fatalf("could not specify multiple authentication mechanisms.")
```
roidelapluie, line 96:
```suggestion
		kingpin.Fatalf("basic authentication and bearer token are mutually exclusive")
```

clyang82:
Thanks @roidelapluie

roidelapluie:
Wait, I now realize that this is giving the token on the command line, which is insecure. Can we point to a bearer token file instead?

clyang82:
Sorry for late response. You can put the bearer_token in $HOME/.config/amtool/config.yml or /etc/amtool/config.yml. for example:
```
alertmanager.url: https://xxx
tls.insecure.skip.verify: true
bearer-token: sha256~xxxxxxx
```
Does it help? @roidelapluie

roidelapluie:
I am still against it but I do not want to be "on the way" so I have reached the community for more data/ideas/opinions: https://groups.google.com/g/prometheus-developers/c/-lXLx2nYKlk

clyang82:
Agree with your point that give the token on the command line is insecure. But it is a little complex for the user who needs to create 2 or more files (config.yml and http_client conf file, etc).

clyang82:
@roidelapluie Can we merge http_config with config.yml? the example of $HOME/.config/amtool/config.yml can be:
```
alertmanager.url: https://alertmanager
tls.insecure.skip.verify: true
require-comment: false
http_config:
  bearer_token_file: /tmp/bar
```

roidelapluie:
This is two different things. I would not mix them:

First, the style is different: one is dot.style.keys; the other is snake_case_keys.

Then, they map to different things. I would avois users thinking they can pass the parameters to the command line, like the other values.

cli/root.go
```diff
@@ -85,15 +84,46 @@ func NewAlertmanagerClient(amURL *url.URL) *client.Alertmanager {
...
+	if httpConfigFile != "" {
+		var err error
+		httpConfig, err = config.LoadHttpConfig(httpConfigFile)
+		if err != nil {
+			kingpin.Fatalf("could not load http config file: %v\n", err)
+		}
+
+		tlsConfig, err := promconfig.NewTLSConfig(&httpConfig.TLSConfig)
+		if err != nil {
+			kingpin.Fatalf("failed to create tls config: %v\n", err)
```
simonpasquier, line 106:
```suggestion
			kingpin.Fatalf("failed to create TLS config: %v\n", err)
```

cli/root.go
```diff
@@ -85,15 +84,46 @@ func NewAlertmanagerClient(amURL *url.URL) *client.Alertmanager {
...
 		password, _ := amURL.User.Password()
 		cr.DefaultAuthentication = clientruntime.BasicAuth(amURL.User.Username(), password)
 	}
 
+	var httpConfig *promconfig.HTTPClientConfig
+	if httpConfigFile != "" {
+		var err error
+		httpConfig, err = config.LoadHttpConfig(httpConfigFile)
+		if err != nil {
+			kingpin.Fatalf("could not load http config file: %v\n", err)
```
simonpasquier, line 101:
```suggestion
			kingpin.Fatalf("could not load HTTP config file: %v\n", err)
```

cli/config/http_config.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"io/ioutil"
+	"os"
+
+	promconfig "github.com/prometheus/common/config"
+	"gopkg.in/yaml.v2"
+)
+
+// LoadHttpConfig returns HTTPClientConfig for the given http_config file
+func LoadHttpConfig(httpConfigFile string) (*promconfig.HTTPClientConfig, error) {
```
simonpasquier, line 25:
```suggestion
func LoadHTTPConfig(httpConfigFile string) (*promconfig.HTTPClientConfig, error) {
```

cli/root.go
```diff
@@ -85,15 +84,46 @@ func NewAlertmanagerClient(amURL *url.URL) *client.Alertmanager {
...
+		cr.Transport = &http.Transport{
+			TLSClientConfig: tlsConfig,
+		}
+
+		if httpConfig.BasicAuth != nil {
+			password, _ := config.FetchBasicAuthPassword(httpConfig.BasicAuth)
+			cr.DefaultAuthentication = clientruntime.BasicAuth(httpConfig.BasicAuth.Username, password)
+		}
+
+		if httpConfig.Authorization.Type == "Bearer" {
```
simonpasquier, line 118: To be consistent with what the HTTP configuration file exposes, we shouldn't limit authorization type to bearer token only. The same is true for other settings like OAuth2 and proxy.
It would be better to create a custom HTTP client from the given configuration and use `clientruntime.NewWithClient()`?

clyang82: @simonpasquier Thanks for your comments. I would like to have this PR to cover bearer token firstly. I will have a following PR to address a custom HTTP client for other cases. Does it make sense? we have customers to wait for bearer token support.

simonpasquier: I'd prefer to have everything in the same PR otherwise it may confuse our users.

clyang82: @simonpasquier PTAL. Thanks.

cli/config/http_config_test.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+func TestInvalidHTTPConfig(t *testing.T) {
+	_, err := LoadHTTPConfig("testdata/http_config.bad.yml")
+	errMsg := `authorization type cannot be set to "basic", use "basic_auth" instead`
+	if !strings.Contains(err.Error(), errMsg) {
+		t.Errorf("Expected error for invalid HTTP client configuration to contain %q but got: %s", errMsg, err)
+	}
+}
+func TestValidHTTPConfig(t *testing.T) {
```
simonpasquier, line 32:
```suggestion
}

func TestValidHTTPConfig(t *testing.T) {
```

cli/config/http_config.go
```diff
@@ -0,0 +1,40 @@
...
+
+	promconfig "github.com/prometheus/common/config"
+	"gopkg.in/yaml.v2"
+)
+
+// LoadHTTPConfig returns HTTPClientConfig for the given http_config file
+func LoadHTTPConfig(httpConfigFile string) (*promconfig.HTTPClientConfig, error) {
+	if _, err := os.Stat(httpConfigFile); err != nil {
+		return nil, err
+	}
```
simonpasquier, line 28: I'm not sure that os.Stat is needed since ioutil.ReadFile would fail the same way.

```suggestion
```

cli/config/http_config_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+package config
+
+import (
+	"strings"
+	"testing"
+)
+
+const (
+	AuthorizationCredentials = "theanswertothegreatquestionoflifetheuniverseandeverythingisfortytwo"
```
simonpasquier, line 22: it doesn't seem to be used?

cli/config/http_config_test.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	proxyURL := "http://remote.host"
+	if cfg.ProxyURL.String() != proxyURL {
+		t.Fatalf("Expected proxy_url is %q but got: %s", proxyURL, cfg.ProxyURL.String())
+	}
+}
+
+func TestValidBasicAuthHTTPConfig(t *testing.T) {
+
```
simonpasquier, line 46:
```suggestion
```

cli/config/http_config_test.go
```diff
@@ -0,0 +1,60 @@
...
+func TestInvalidHTTPConfig(t *testing.T) {
+	_, err := LoadHTTPConfig("testdata/http_config.bad.yml")
+	errMsg := `authorization type cannot be set to "basic", use "basic_auth" instead`
+	if !strings.Contains(err.Error(), errMsg) {
+		t.Errorf("Expected error for invalid HTTP client configuration to contain %q but got: %s", errMsg, err)
+	}
+}
+func TestValidHTTPConfig(t *testing.T) {
+
```
simonpasquier, line 33:
```suggestion
```

cli/config/http_config.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"io/ioutil"
+	"os"
+
+	promconfig "github.com/prometheus/common/config"
+	"gopkg.in/yaml.v2"
+)
+
+// LoadHTTPConfig returns HTTPClientConfig for the given http_config file
+func LoadHTTPConfig(httpConfigFile string) (*promconfig.HTTPClientConfig, error) {
```
simonpasquier, line 25: (nit)

```suggestion
func LoadHTTPConfigFile(filename string) (*promconfig.HTTPClientConfig, error) {
```

cli/root.go
```diff
@@ -85,15 +83,30 @@ func NewAlertmanagerClient(amURL *url.URL) *client.Alertmanager {
...
 		password, _ := amURL.User.Password()
 		cr.DefaultAuthentication = clientruntime.BasicAuth(amURL.User.Username(), password)
 	}
 
+	var httpConfig *promconfig.HTTPClientConfig
+	if httpConfigFile != "" {
+		var err error
+		httpConfig, err = config.LoadHTTPConfig(httpConfigFile)
+		if err != nil {
+			kingpin.Fatalf("could not load HTTP config file: %v\n", err)
```
simonpasquier, line 100:
```suggestion
			kingpin.Fatalf("failed to load HTTP config file: %v", err)
```

cli/root.go
```diff
@@ -85,15 +83,30 @@ func NewAlertmanagerClient(amURL *url.URL) *client.Alertmanager {
...
+		}
+
+		httpclient, err := promconfig.NewClientFromConfig(*httpConfig, "amtool")
+		if err != nil {
+			kingpin.Fatalf("could not create a new HTTP client: %v\n", err)
```
simonpasquier, line 105:
```suggestion
			kingpin.Fatalf("failed to create a new HTTP client: %v", err)
```

cli/root.go
```diff
@@ -85,15 +83,30 @@ func NewAlertmanagerClient(amURL *url.URL) *client.Alertmanager {
 
 	cr := clientruntime.New(address, path.Join(amURL.Path, defaultAmApiv2path), schemes)
 
-	cr.Transport = &http.Transport{
-		TLSClientConfig: tlsConfig,
+	if amURL.User != nil && httpConfigFile != "" {
+		kingpin.Fatalf("basic authentication and http.config.file are mutually exclusive")
```
simonpasquier, line 87: I'm wondering what takes precedence if you specify credentials both in the URL and in the HTTP config file?

cli/root.go
```diff
@@ -85,15 +83,30 @@ func NewAlertmanagerClient(amURL *url.URL) *client.Alertmanager {
...
 	}
 
 	if amURL.User != nil {
 		password, _ := amURL.User.Password()
 		cr.DefaultAuthentication = clientruntime.BasicAuth(amURL.User.Username(), password)
 	}
 
+	var httpConfig *promconfig.HTTPClientConfig
```
simonpasquier, line 95: I'd find it more readable if we had distinct paths depending on whether a config file is given or not
```
var cr *clientruntime.Runtime
if httpConfigFile != "" {
   ...
} else {
   ...
}
```

simonpasquier (commented):
Minor comments, looks good otherwise. Thanks!

cli/root.go
clyang82: That is mutually exclusive. You cannot specify credentials both in the URL and in the HTTP config file. If you specify credentials both in the URL and in the HTTP config file, the error `basic authentication and http.config.file are mutually exclusive` thrown and exit.

cli/config/http_config.go
clyang82: Yes. it is unnecessary. removed

cli/root.go
clyang82: remove this variable declaration since we do not need to have this variable out of the if statement.

cli/root.go
simonpasquier: I get that but my question was more: how would `net/http` behave if we don't forbid this combination? I've tested locally and the credentials from the config file take precedence over the credentials passed in the URL. I'm fine with keeping this check though since it would avoid people losing time figuring out why their credentials aren't taken into account.

clyang82: Yes. keep this check make it clear to avoid such thing you mentioned. Thanks @simonpasquier

cli/root.go
```diff
@@ -85,15 +83,29 @@ func NewAlertmanagerClient(amURL *url.URL) *client.Alertmanager {
...
+	if httpConfigFile != "" {
+		var err error
+		httpConfig, err := config.LoadHTTPConfigFile(httpConfigFile)
+		if err != nil {
+			kingpin.Fatalf("failed to load HTTP config file: %v\n", err)
+		}
+
+		httpclient, err := promconfig.NewClientFromConfig(*httpConfig, "amtool")
+		if err != nil {
+			kingpin.Fatalf("failed to create a new HTTP client: %v\n", err)
```
simonpasquier, line 104:
```suggestion
			kingpin.Fatalf("failed to create a new HTTP client: %v", err)
```

cli/root.go
```diff
@@ -85,15 +83,29 @@ func NewAlertmanagerClient(amURL *url.URL) *client.Alertmanager {
...
 	if amURL.User != nil {
 		password, _ := amURL.User.Password()
 		cr.DefaultAuthentication = clientruntime.BasicAuth(amURL.User.Username(), password)
 	}
 
+	if httpConfigFile != "" {
+		var err error
+		httpConfig, err := config.LoadHTTPConfigFile(httpConfigFile)
+		if err != nil {
+			kingpin.Fatalf("failed to load HTTP config file: %v\n", err)
```
simonpasquier, line 99: Sorry I missed that one too:

```suggestion
			kingpin.Fatalf("failed to load HTTP config file: %v", err)
```

cli/config/http_config.go
```diff
@@ -0,0 +1,36 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	httpConfig := &promconfig.HTTPClientConfig{}
+	err = yaml.UnmarshalStrict(b, httpConfig)
+	if err != nil {
+		return nil, err
+	}
+	return httpConfig, nil
```
simonpasquier, line 35: We need also to call `httpConfig.SetDirectory(...)` to resolve relative paths.

```suggestion
	httpConfig.SetDirectory(filepath.Dir(filepath.Dir(filename)))

	return httpConfig, nil
```

clyang82: 👍 Thanks @simonpasquier Let me know if you have further comments.

clyang82:
@simonpasquier PTAL if you are OK to merge it in. Thanks.

simonpasquier (approved):
lgtm. @roidelapluie could you give a final look in case I've missed anything? thanks!

roidelapluie (approved):
LGTM

roidelapluie:
Thanks!

PR #2765, opened 2021-11-10
Expose the fields from MuteTimeInterval in JSON using the YAML field name

JohnnyQQQQ:
Before this PR, when marshalling to JSON, the go variable name would be taken (respective 'Name' and 'TimeIntervals') as JSON field name. With this PR we take instead the same name as for the YAML fields as everywhere else in the configuration.


gotjosh (approved):
LGTM

roidelapluie (approved):

Issue #2769, opened 2021-11-19
Feature Request: Add max_alerts to slack receiver


aned:
We get a lot of `400s` errors when the alertmanager tries to push 20+ alerts at once to slack. 
It would be nice to have max_alerts feature, similar to what's implemented for webhook receiver: https://prometheus.io/docs/alerting/latest/configuration/#webhook_config

roidelapluie:
is this 400s or 429? Where are the slack limits documented? Thanks!

aned:
We got 400s. 
Slack limits: https://api.slack.com/rtm#real-time-messaging-rtm-api__limits

PR #2774, opened 2021-11-26
Truncating fields that are sent to slack

prashbnair:
Signed-off-by: Prashant Balachandran <pnair@redhat.com>

Fixed issue https://github.com/prometheus/alertmanager/issues/2769

### DIFF
--- a/notify/slack/slack.go
+++ b/notify/slack/slack.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/go-kit/log/level"
 	"io/ioutil"
 	"net/http"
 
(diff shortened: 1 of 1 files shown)


roidelapluie:
Thanks for your contribution. I think we should limit the length of the fields we send to slack,  like we do for pageduty, instead of limiting the number of alerts.

prashbnair:
Removed the max alerts config parameter and truncated the field that is the longest as per https://github.com/prometheus/alertmanager/blob/main/template/default.tmpl#L15

notify/slack/slack.go
```diff
@@ -92,13 +93,22 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 		tmplText = notify.TmplText(n.tmpl, data, &err)
 	)
 	var markdownIn []string
+	key, err := notify.ExtractGroupKey(ctx)
+
+	if err != nil {
+		return false, err
+	}
```
simonpasquier, line 100: i would move this block inside the `if truncated { ... }` block.

prashbnair: Done

notify/slack/slack.go
```diff
@@ -92,13 +93,22 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
...
 	} else {
 		markdownIn = n.conf.MrkdwnIn
 	}
+	title, truncated := notify.Truncate(tmplText(n.conf.Title), 1024)
+	if truncated {
+		key, err := notify.ExtractGroupKey(ctx)
+		if err != nil {
+			return false, err
+		}
+		level.Debug(n.logger).Log("msg", "Truncated text", "text", title, "key", key)
```
simonpasquier, line 108:
```suggestion
		level.Debug(n.logger).Log("msg", "Truncated title", "text", title, "key", key)
```

prashbnair: Done

Issue #2217, opened 2020-03-26
Slack Blocks Notifications


mambax:
Hi Prometheus

Currently [`<slack-config>`](https://prometheus.io/docs/alerting/configuration/#slack_config) is using [Simple formatting](https://api.slack.com/reference/surfaces/formatting).

This is marked by Slack themselves as `deprecated` (or at least most hints dropped say: Format using Block Kit).

The new way to go seems to be [Block-Kit](https://api.slack.com/block-kit).

I don't want to ask "Is there a plan?" as this would be a question. I'd like to pose as feature request to add something like:

`<slack_config_blocks>`
which would then accept stuff like:

![image](https://user-images.githubusercontent.com/5472171/77641279-16e4f700-6f5c-11ea-8972-95e947c60f9d.png)

And configuration like so (I am new to Yaml and Go Tmpl):
```
blocks:
- type: section
  text:
    type: mrkdwn
    text: ":fire: `Critical` alert in *production*: `ELK running out of shards (2
      left)`"
- type: section
  fields:
  - type: mrkdwn
    text: |-
      *Value before:*
      22
  - type: mrkdwn
    text: |-
      *Value now:*
      2
  - type: mrkdwn
    text: |-
      *Action:*
      Contact operations.
  - type: mrkdwn
    text: |-
      *Outcome:*
      X will happen.
- type: actions
  elements:
  - type: button
    text:
      type: plain_text
      emoji: true
      text: Victorops
    style: primary
    value: click_me_123
  - type: button
    text:
      type: plain_text
      emoji: true
      text: Silence
    style: danger
    value: click_me_123
```
Obvisouly my hardcoded strings would be replaced by `{{ .alertProperties }}`

Adding this I believe would make our DevOps plans (and hopefully others) dreams come true ❤️

simonpasquier:
I'm worried that the configuration schema might quickly get out of control. But there's nothing preventing someone to try to come up with something reasonable.

vanugrah:
Block kit makes formatting of messages SO much cleaner. If I could +1000 I would :)

KarthikRangaraju:
+1 The possibilities are endless. These are all the things we can do: https://app.slack.com/block-kit-builder/T091CRSGH#%7B%22blocks%22:%5B%5D%7D

niko2:
[Simple Formatting](https://api.slack.com/reference/surfaces/formatting) seems to be broken on Slack Desktop (at least for buttons).
[...]

PR #2777, opened 2021-11-30
Add support for Slack Block Kit

johanfleury:
Implements #2217

This is a draft PR in which I started implementing support for Slack Block Kit. It’s still missing a lot (from actual implementation to unit and acceptance tests), but I think it’s a good first step.

I’m still willing to work on this but I find it a bit messy and I’d rather have a first review/validation from project maintainers especially concerning:
* what to do regarding the inability for `encoding/json` to marshal `map[interfaces{}]interfaces{}` (see comment on `SlackBlock.Config` at line 407 in `config/notifiers.go`) and
* how to iterate with at lines 151 and 170 in `notify/slack/slack.go` (my first idea was to use reflect to find string values and apply the templating function on them).

Discussion on a “design” have taken place in #2217 and this might not be the best way to implement this feature. Let me know.

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -68,17 +68,20 @@ var (
 		NotifierConfig: NotifierConfig{
 			VSendResolved: false,
 		},
-		Color:      `{{ if eq .Status "firing" }}danger{{ else }}good{{ end }}`,
-		Username:   `{{ template "slack.default.username" . }}`,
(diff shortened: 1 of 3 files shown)


carlosjgp:
Looking forward to see this implemented

johanfleury:
Closing this as I don’t have time to work on it.

Issue #2778, opened 2021-12-02
New feature: Mute when time is outside time_interval


dubyte:
Currently mute_intervals allow to express the interval where the alarm won't be sent. 
But would be useful to mute in the complement of the mute_time_interval.

Let's check at this example:
```
  routes:
  - receiver: devs
    mute_time_intervals:
      - offhours
      - holidays
    continue: true
  - receiver: monitoring
    active_time_intervals: 
      - offhours
      - holidays
  time_intervals:     
  - name: offhours
    time_intervals:
    - ...
  - name: holidays
    time_intervals:
    - ...

```
"mute_time_intervals" keyword to describe the intervals will be deprecated. Instead the configuration will use "time_intervals". When in a route there is active and mute time intervals, the mute_time_interval will have higher precedence.
For example:
```
 - receiver: group_a
    active_time_intervals:
    - officeHours
   mute_time_intervals:
   - lunchHour
```
Let's imagine that there is a time interval for officeHours 9am - 5pm and a lunch hour in the middle of the day. With this the alert will be sent during officeHours but it wont receive during the lunch.  

Updated about:
- Deprecation of mute_time_interval
- new time_intervals config
- mute_time_interval has higher precedence than active_time_intervals

benridley:
Nice @dubyte , I think this is a cool idea, and a natural extension of the time intervals concept so thanks for raising it. We should probably clarify the semantics of how this will work before considering further. Some of my rough thoughts:
- We should change the top level config section for `mute_time_intervals` to just `time_intervals`, while still allowing `mute_time_intervals` basically as an alias and note that its usage is deprecated.
- Mute time intervals will take precedence over active time intervals. This means that if a route provides both active and mute intervals, it will only be active during the active intervals unless that interval overlaps with a mute interval, in which case the route will be muted in the period of overlap.

Any thoughts?

dubyte:
I agree about deprecating mute_time_interval. 

About the precedence  you mentioned, that is the current behavior of the PR but config is not mention it.

I can update the PR or what would be the next steps?

dubyte:
Ahh i just notice that I reference a PR but this is a comment in the New Feature "issue"

Should I update the new feature to include your comments and continue talking Frome there?

Thanks

simonpasquier:
Closed by #2779

DanielZerihon:
sadsad

PR #2779, opened 2021-12-02
Adds: Active time interval

dubyte:
This is a PR that implements active time interval as in: https://github.com/prometheus/alertmanager/issues/2778

The point is to use active_time_interval to express that alertmanager only sent alerts in the interval mentioned.
so in other words active_time_interval mutes in the complement of mute_time_interval


dubyte:
Example:

```
  routes:
  - receiver: devs
    mute_time_intervals:
      - offhours
      - holidays
    continue: true
  - receiver: monitoring
    active_time_intervals: 
      - offhours
      - holidays

mute_time_intervals:
- name: offhours
  time_intervals:
  # Monday to Friday 5pm - 9:00 in UTC
  - weekdays: ['monday:friday']
    times:
    - start_time: '00:00'
      end_time: '13:00'
    - start_time: '21:00'
      end_time: '24:00'
  - weekdays: ['saturday', 'sunday']
- name: holidays
  time_intervals:
  - months: ['december']
    days_of_month: ['24']
  - months: ['december']
    days_of_month: ['25:31']
  - months: ['january']
    days_of_month: ['1:2']
  - ...
```

notify/notify.go
```diff
@@ -206,13 +211,20 @@ func ResolvedAlerts(ctx context.Context) ([]uint64, bool) {
 	return v, ok
 }
 
-// MuteTimeIntervalNames extracts a slice of mute time names from the context. Iff none exists, the
+// MuteTimeIntervalNames extracts a slice of mute time names from the context. If none exists, the
```
simonpasquier, line 214: I bet that Iff was for "if and only if"
```suggestion
// MuteTimeIntervalNames extracts a slice of mute time names from the context. If and only if none exists, the
```

notify/notify.go
```diff
@@ -785,6 +797,11 @@ func NewTimeMuteStage(mt map[string][]timeinterval.TimeInterval) *TimeMuteStage
 // Exec implements the stage interface for TimeMuteStage.
 // TimeMuteStage is responsible for muting alerts whose route is not in an active time.
 func (tms TimeMuteStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.Alert) (context.Context, []*types.Alert, error) {
```
simonpasquier, line 799: it might be more convenient to have a generic `TimeStage` that could work for both active and mute depending on a flag? And the pipeline would run the active interval stage before the mute interval stage.

simonpasquier (commented):
I agree that it makes sense.

notify/notify.go
dubyte: I see it make sense, thanks

dispatch/route.go
```diff
@@ -210,6 +211,9 @@ type RouteOpts struct {
 
 	// A list of time intervals for which the route is muted.
 	MuteTimeIntervals []string
+
+	// A list of time intervals for which the route is active
```
simonpasquier, line 215:
```suggestion
	// A list of time intervals for which the route is active.
```

config/config.go
```diff
@@ -516,9 +516,13 @@ func checkTimeInterval(r *Route, timeIntervals map[string]struct{}) error {
 			return err
 		}
 	}
-	if len(r.MuteTimeIntervals) == 0 {
-		return nil
+
+	for _, mt := range r.ActiveTimeIntervals {
```
simonpasquier, line 520: we need also a test in `*Config.UnmarshalYAML()` to prevent active time intervals on the root route.

dubyte: Agreed I will include the test, thanks

config/config.go
```diff
@@ -478,18 +498,33 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
 	if err := checkReceiver(c.Route, names); err != nil {
 		return err
 	}
 
 	tiNames := make(map[string]struct{})
-	for _, mt := range c.MuteTimeIntervals {
-		if _, ok := tiNames[mt.Name]; ok {
-			return fmt.Errorf("mute time interval %q is not unique", mt.Name)
+	// read mute time intervals until deprecated
+	if len(c.MuteTimeIntervals) > 0 {
```
simonpasquier, line 512: We should probably consider that `mute_time_intervals` and `time_intervals` can coexist and that `mute_time_intervals` can be used indifferently for mute and active intervals.
In which case, I would expect that we check name uniqueness across both slices.

notify/notify.go
```diff
@@ -309,7 +321,7 @@ func (pb *PipelineBuilder) New(
 	wait func() time.Duration,
 	inhibitor *inhibit.Inhibitor,
 	silencer *silence.Silencer,
-	muteTimes map[string][]timeinterval.TimeInterval,
+	times map[string][]timeinterval.TimeInterval,
```
simonpasquier, line 324: IIUC we're still passing only the mute time intervals?

notify/notify.go
```diff
@@ -794,24 +809,71 @@ func (tms TimeMuteStage) Exec(ctx context.Context, l log.Logger, alerts ...*type
...
+	// If the current time is not inside an active time, all alerts are removed from the pipeline
+	if !active {
+		level.Debug(l).Log("msg", "Notifications not sent, route is not within active time")
+		return ctx, nil, nil
+	}
+
+	return ctx, alerts, nil
+}
+
+func InTimeIntervals(now time.Time, intervals map[string][]timeinterval.TimeInterval, intervalNames []string) (bool, error) {
```
simonpasquier, line 863: (nit)
```suggestion
// inTimeIntervals returns true if the current time is contained in one of the given time intervals.
func inTimeIntervals(now time.Time, intervals map[string][]timeinterval.TimeInterval, intervalNames []string) (bool, error) {
```

config/config.go
dubyte: The test was added

config/config.go
dubyte: I see I will check this, thanks

docs/configuration.md
```diff
@@ -189,6 +189,15 @@ matchers:
 mute_time_intervals:
   [ - <string> ...]
```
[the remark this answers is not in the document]
dubyte: Ok, i will update the configuration doc

notify/notify.go
dubyte: sorry what means IIUC?

docs/configuration.md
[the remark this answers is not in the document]
dubyte: I update the documentation. I also add an example.

config/config.go
```diff
@@ -266,14 +266,34 @@ func (mt *MuteTimeInterval) UnmarshalYAML(unmarshal func(interface{}) error) err
...
-	Route             *Route             `yaml:"route,omitempty" json:"route,omitempty"`
-	InhibitRules      []*InhibitRule     `yaml:"inhibit_rules,omitempty" json:"inhibit_rules,omitempty"`
-	Receivers         []*Receiver        `yaml:"receivers,omitempty" json:"receivers,omitempty"`
-	Templates         []string           `yaml:"templates" json:"templates"`
+	Global       *GlobalConfig  `yaml:"global,omitempty" json:"global,omitempty"`
+	Route        *Route         `yaml:"route,omitempty" json:"route,omitempty"`
+	InhibitRules []*InhibitRule `yaml:"inhibit_rules,omitempty" json:"inhibit_rules,omitempty"`
+	Receivers    []*Receiver    `yaml:"receivers,omitempty" json:"receivers,omitempty"`
+	Templates    []string       `yaml:"templates" json:"templates"`
+	// Deprecated.
```
simonpasquier, line 294:
```suggestion
	// Deprecated. Remove before v1.0 release.
```

docs/configuration.md
```diff
@@ -116,9 +116,14 @@ receivers:
 inhibit_rules:
   [ - <inhibit_rule> ... ]
 
+# DEPRECATED: use time_intervals below
```
simonpasquier, line 119:
```suggestion
# DEPRECATED: use time_intervals below.
```

docs/configuration.md
```diff
@@ -189,6 +194,15 @@ matchers:
 mute_time_intervals:
   [ - <string> ...]
 
+# Times when the route should be active. These must match the name of a
+# mute time interval defined in the mute_time_intervals section.
```
simonpasquier, line 198: I would expect the same change for the `mute_time_intervals` field above.
```suggestion
# time interval defined in the time_intervals section. 
```

docs/configuration.md
```diff
@@ -189,6 +194,15 @@ matchers:
 mute_time_intervals:
   [ - <string> ...]
 
+# Times when the route should be active. These must match the name of a
+# mute time interval defined in the mute_time_intervals section. 
+# Additionally, the root node cannot have any active times.
+# Only when a route is active it will send any notifications, but
+# otherwise acts as muted (it ends the route-matching process
+# if the `continue` option is not set.)
```
simonpasquier, line 202:
```suggestion
# The route will send notifications only when active, but otherwise
# otherwise acts normally (including ending the route-matching process
# if the `continue` option is not set).
```

docs/configuration.md
```diff
@@ -189,6 +194,15 @@ matchers:
 mute_time_intervals:
   [ - <string> ...]
 
+# Times when the route should be active. These must match the name of a
```
simonpasquier, line 197: I think it's worth mentioning here that an empty values means that the route is always active.

docs/configuration.md
```diff
@@ -221,11 +235,31 @@ route:
...
+      - offhours
+      - holidays
 ```
 
-## `<mute_time_interval>`
+## `<time_interval>`
 
-A `mute_time_interval` specifies a named interval of time that may be referenced
+A `time_interval` specifies a named interval of time that may be referenced
 in the routing tree to mute particular routes for particular times of the day.
```
simonpasquier, line 263:
```suggestion
in the routing tree to mute/activate particular routes for particular times of the day.
```

notify/notify.go
simonpasquier: My bad, IIUC="if I understand correctly".

I meant to say that the main() function only passes `conf.MuteTimeIntervals` and that it should merge `MuteTimeIntervals` and `TimeIntervals` instead.

https://github.com/prometheus/alertmanager/blob/da9de1f540af4f01f6336d4424e76a70c547bcbb/cmd/alertmanager/main.go#L439-L443

docs/configuration.md
dubyte: Right, I will include this

dubyte: I update the doc.

dubyte: It is updated

notify/notify.go
dubyte: Yes, that was right I update that part to merge the both configs

docs/configuration.md
```diff
@@ -189,6 +194,16 @@ matchers:
 mute_time_intervals:
   [ - <string> ...]
 
+# Times when the route should be active. These must match the name of a
+# time interval defined in the time_intervals section. An empty value
+# means the alert is always active.
```
simonpasquier, line 199:
```suggestion
# means that the route is always active.
```

docs/configuration.md
```diff
@@ -189,6 +194,16 @@ matchers:
...
+# Additionally, the root node cannot have any active times.
+# The route will send notifications only when active, but otherwise
+# otherwise acts normally (including ending the route-matching process
```
simonpasquier, line 202:
```suggestion
# The route will send notifications only when active, but otherwise
# acts normally (including ending the route-matching process
```

docs/configuration.md
```diff
@@ -221,12 +236,32 @@ route:
     group_by: [product, environment]
     matchers:
     - team="frontend"
+
+  # All alerts with the service=inhouse-service label match this sub-route.
+  # the alert will be mute during offhours and holidays time intervals.
```
simonpasquier, line 241:
```suggestion
  # the route will be muted during offhours and holidays time intervals.
```

docs/configuration.md
```diff
@@ -221,12 +236,32 @@ route:
...
+  # even if it match, it will continue to the next sub-route
+  - receiver: 'dev-pager'
+    matchers:
+      - service:~"inhouse-service"
```
simonpasquier, line 245:
```suggestion
      - service="inhouse-service"
```

docs/configuration.md
```diff
@@ -221,12 +236,32 @@ route:
...
+    mute_time_interval:
+      - offhours
+      - holidays
+    continue: true
+
+    # All alerts with the service=inhouse-service label match this sub-route
+    # the alert will be active only during offhours and holidays time intervals.
+  - receiver: 'on-call-pager'
+    matchers:
+      - service:~"inhouse-service"
```
simonpasquier, line 255:
```suggestion
      - service="inhouse-service"
```

docs/configuration.md
```diff
@@ -221,12 +236,32 @@ route:
...
+  - receiver: 'dev-pager'
+    matchers:
+      - service:~"inhouse-service"
+    mute_time_interval:
+      - offhours
+      - holidays
+    continue: true
+
+    # All alerts with the service=inhouse-service label match this sub-route
+    # the alert will be active only during offhours and holidays time intervals.
```
simonpasquier, line 252:
```suggestion
    # the route will be active only during offhours and holidays time intervals.
```

cmd/alertmanager/main.go
```diff
@@ -436,10 +436,14 @@ func run() int {
 			integrationsNum += len(integrations)
 		}
 
-		// Build the map of time interval names to mute time definitions.
-		muteTimes := make(map[string][]timeinterval.TimeInterval, len(conf.MuteTimeIntervals))
+		// Build the map of time interval names to mute time definitions or.
```
simonpasquier, line 439:
```suggestion
		// Build the map of time interval names to time interval definitions.
```

cmd/alertmanager/main.go
```diff
@@ -436,10 +436,14 @@ func run() int {
...
+		timeIntervals := make(map[string][]timeinterval.TimeInterval, len(conf.MuteTimeIntervals))
```
simonpasquier, line 440:
```suggestion
		timeIntervals := make(map[string][]timeinterval.TimeInterval, len(conf.MuteTimeIntervals) + len(conf.TimeIntervals))
```

simonpasquier (commented):
a few nits, otherwise lgtm

dubyte:
Thank you @simonpasquier

docs/configuration.md
```diff
@@ -221,12 +236,32 @@ route:
     group_by: [product, environment]
     matchers:
     - team="frontend"
+
+  # All alerts with the service=inhouse-service label match this sub-route.
+  # the route will be muted during offhours and holidays time intervals.
+  # even if it match, it will continue to the next sub-route
```
simonpasquier, line 242:
```suggestion
  # even if it matches, it will continue to the next sub-route
```

docs/configuration.md
```diff
@@ -116,9 +116,14 @@ receivers:
 inhibit_rules:
   [ - <inhibit_rule> ... ]
 
+# DEPRECATED: use time_intervals below.
 # A list of mute time intervals for muting routes.
 mute_time_intervals:
   [ - <mute_time_interval> ... ]
+
+# A list of time intervals for muting routes.
```
simonpasquier, line 124:
```suggestion
# A list of time intervals for muting/activating routes.
```

simonpasquier (commented):
looks good to me (minus 2 last nits). I'd ask @roidelapluie to give it a review to make sure that I haven't missed anything critical. Thanks!

simonpasquier:
could you also rebase on top of the `main` branch (it should fix the CI failure with mixins)?

xkilian:
An issue has been identified related to this PR during testing with @dubyte 

If an active alert is muted, the only effect currently, is to mute notifications.
But the alert is still considered active, which causes duplicate alertes to be created. There is no way to distinguish between a real alert and a muted alert.

Suggested paths to correct this:

- Add a new possible state: active, silenced, inhibited, muted
  - change the state of the alert to muted
  - add the muted state to the @state LABEL.
  - add muted to the API filters when querying for alerts
- MAYBE: muted and unmuted alerts that are equivalent should be treated as the same alert for repeat-interval
- alerts should not clear/trigger at each time_interval cross.
- alertmanager UI should permit to filter between active/silenced/inhibited/muted alerts

Or
- add a LABEL to the alert, to indicated this alert is muted (like state=muted or muted=true); This could be a temporary fix while solution 1 is fleshed out.

Or
- consider active_interval and mute_interval in the matching stage and not in the notification processing stage. Though this may lead to clearing+re-firing alerts with no way to deal with them. So I do not think this is the right path.

dubyte:
Hi
Agreed that there is no way to know if the notification is muted -for certain receivers- or not, so I propose the next: https://github.com/prometheus/alertmanager/issues/2832

If all agreed I would like to move the discussion about the api  change to that feature request. So we can close this PR.
The tests show that the feature works for mute and active time intervals. So we are not blocked.

Thank you

simonpasquier:
> But the alert is still considered active, which causes duplicate alertes to be created in the alertmanager/karma UI and alertmanager API. This is due to the use of the "continue" keywork and the fact intervals are evaluated at the notification phase. There is no way to distinguish between a real alert and a muted alert.

Correct but the state of an alert (muted/active) is always relative to an alert group. I agree with @dubyte that it should be addressed at the API/UI level.

simonpasquier:
@benridley would you have time to review this PR by any chance (at least the UX)?

notify/notify.go
```diff
@@ -794,24 +809,69 @@ func (tms TimeMuteStage) Exec(ctx context.Context, l log.Logger, alerts ...*type
...
 	return ctx, alerts, nil
 }
+
+type TimeActiveStage timeStage
+
+func NewTimeActiveStage(ti map[string][]timeinterval.TimeInterval) *TimeActiveStage {
+	return &TimeActiveStage{ti}
+}
+
+// Exec implements the stage interface for TimeMuteStage.
```
benridley, line 831: Minor typo, I think this should be TimeActiveStage?

notify/notify.go
```diff
@@ -794,24 +809,69 @@ func (tms TimeMuteStage) Exec(ctx context.Context, l log.Logger, alerts ...*type
...
+// TimeStage is responsible for muting alerts whose route is not in an active time.
```
benridley, line 832: And here too

benridley:
Besides the minor typo above, this looks great. Thanks @dubyte, I think this will be a nice addition to AM.

notify/notify.go
dubyte: Thanks fixing the typo

notify/notify.go
dubyte: Thanks

dubyte:
Thanks @benridley

simonpasquier (commented):
sorry @dubyte I had 2 last pending comments that I forgot to submit. Hopefully they are minor and easy to fix :)

simonpasquier (approved):
:+1:

bollmann:
Hi,

I'm trying to use this feature and I'm defining my time intervals and corresponding routes as follows:

```
    time_intervals:
      - name: workinghours
        time_intervals:
          - times:
              - start_time: 08:00
                end_time: 18:00
            weekdays:
              - monday
              - tuesday
              - wednesday
              - thursday
              - friday
...
    route:
      routes:
        - receiver: oncall-pager
          matchers:
            - severity="workinghours"
          active_time_intervals:
            - workinghours
```

Furthermore, I have a prometheus rule with label `severity="workinghours"` defined. With this configuration, I would expect my prometheus rule to only be active during workinghours, i.e., Monday to Friday from 8am until 6pm. However, for some reason my prometheus rule also fires outside of the hours 8am until 6pm. That is, it seems to be active even when it should not be.

Did I do something wrong here in the configuration?

xkilian:
Here is a working example for you.

It is a best practive to define your alertes with severity levels. We use 0-5, 5 being the most critical.
The group label defines the administrative group responsible for handling the alert.

So based on the group, severity and time of day the alert will be routed to the appropriate receiver configuration. We use distinct receiver configurations as the email template (in our case) need to be different for each case. Ex. Offhours is an email to a specific box. Critical daytime hours will open a ticket automatically with a ticketing app. And severity 3 will be handled with an email regardless of time of day.

What you were doing wrong is not setting the mute_time_intervals. Which is when to NOT send a notification (ex. off-hours), Then the time interval indicates what to actually DO during offhours. This is documented, but may not be clear at first read.

`    - matchers:
        - group= system-admins
      receiver: receiver-system-admins
      routes:
        - matchers:
            - severity = 3
          receiver: receiver-system-admins-email
        - matchers:
            - "severity =~ 4|5"
          receiver: receiver-system-admins-critical-daytime
          mute_time_intervals: 
            - offhours
            - holidays
          continue: true
        - matchers:
            - "severity =~ 4|5"
            - env = PR
          receiver: receiver-system-admins-critical-offhours
          active_time_intervals: 
            - offhours
            - holidays       `
Hope this helps

xkilian:
As a bonus, you can see how to do a regex OR statement for the severity (had me stumped for a while) with the new matchers.

PR #2780, opened 2021-12-02
config: root route should have empty matchers

PhilipGough:
Unmarshal should validate that the root route does not contain any matchers. Prior to this change,
only the deprecated match structures were checked. It adds a check for matchers provided with the updated matching syntax.


jan--f (approved):
Thanks, this lgtm.

simonpasquier (commented):
could you add a unit test for it? `TestRootRouteHasNoMatcher()` can be extended.

simonpasquier (approved):

Issue #2771, opened 2021-11-22
mute_time_intervals in local time zone


arnitolog:
Hello,

Does it make sense to add an ability to specify mute_time_intervals start and end times in the local time zone instead of UTC?
Let's imaging that we deploy the same service in 4 regions (US, EU, APAC, AE), And we have a maintenance job for this service starting at 1:00 AM local time. 
Also, we have a local Prometheus/Alertmanager in every region.
With the current approach, we have to have 4 different configurations of Alertmanager depending on the target region because the start time will be different in UTC. Also, onboarding a new region will also require Alertmanager's config modification.
Having the configuration in local time will allow having the same config across all regions, and do not care about time conversion

xkilian:
Same here, we are in a time zone that changes year by year for daylight saving. 

Only having UTC makes it much more annoying to maintain offhour schedules.

There must be a go library that handles timezones?! Is there interest from the devs for a pull request to implement this?

benridley:
Hi @arnitolog, 

Yes there is interest in this, and it was discussed when the feature was first merged. Initially the mute time interval specification allowed a timezone parameter as part of the definition, however this had some issues. The main problem is that Go does not use the Windows timezone database, so conversion between timezones simply does not work on Windows which is a platform that Prometheus supports. 

That being said, Go is capable of retrieving the servers local time from Windows, just not arbitrarily converting between timezones on Windows. Perhaps the easiest option here is to just allow using the servers local timezone as a boolean flag in the timezone definition rather than allowing arbitrary timezones as a parameter. 

I'm happy to add this if @simonpasquier or @w0rm are ok with the concept.

roidelapluie:
We could load the default timezone go provides if we also provide the user the ability to load a fresher timezone file.

benridley:
I've just opened #2782 which allows a user-specified timezone in time interval definitions. For Windows, users can pass in 'Local' which works without a timezone database and (I would guess) addresses most users who just want to use a local time.

ohadasulin:
Anything specific that prevents #2782 from being merged?
It's a great feature to have for global companies, we eagerly wait for this merge :)

megamorf:
I'm also interested in seeing this merged because I'd also like to define time zone specific times.

PR #2782, opened 2021-12-07
Add timezone support to time intervals.

benridley:
This PR addresses #2771 by adding a `time_zone` field to time interval definitions, which accepts an IANA time zone name (like 'Australia/Sydney') and if supplied means that the time intervals are defined inside that time zone.

On Windows machines the only time zones that are accepted are 'Local' and 'UTC', but I think that is probably enough for 99% of users who just want to use their local server time zone. I have noted this in the documentation, and also added some more info to the error message when a timezone fails to load on Windows mentioning that the user may need to supply their own time zone database.

I think it's also probably worth changing the default to be the server's local time rather than UTC as this would probably be unexpected for most users, but it is breaking behavior so I'm not sure how we should proceed with that but open to the ideas.

Feedback on the approach (particularly dealing with Windows) is more than welcome. Cheers!

### DIFF
--- a/notify/notify_test.go
+++ b/notify/notify_test.go
@@ -775,10 +775,19 @@ func TestTimeMuteStage(t *testing.T) {
 			shouldMute: true,
 		},
 		{
-			// Ensure comparisons with other time zones work as expected.
+			fireTime:   "14 Nov 21 21:30 +0000",
+			labels:     model.LabelSet{"mute": "utc"},
+			shouldMute: true,
+		},
+		{
 			fireTime:   "15 Nov 22 14:30 +0900",
 			labels:     model.LabelSet{"kst": "dont_mute"},
 			shouldMute: false,
+		},
+		{
+			fireTime:   "15 Nov 21 02:00 -0500",
+			labels:     model.LabelSet{"mute": "0500"},
+			shouldMute: true,
 		},
 	}
 	var intervals []timeinterval.TimeInterval
--- a/notify/notify_test.go
+++ b/notify/notify_test.go
@@ -724,16 +724,20 @@ func TestMuteStageWithSilences(t *testing.T) {
 }
 
 func TestTimeMuteStage(t *testing.T) {
-	// Route mutes alerts outside business hours if it is a mute_time_interval
+	// Route mutes alerts outside business hours in November, using the AEDT timezone.
 	muteIn := `
 ---
 - weekdays: ['monday:friday']
+  time_zone: 'Australia/Sydney'
+  months: ['November']
   times:
    - start_time: '00:00'
      end_time: '09:00'
    - start_time: '17:00'
      end_time: '24:00'
-- weekdays: ['saturday', 'sunday']`
+- weekdays: ['saturday', 'sunday']
+  months: ['November']
+  time_zone: 'Australia/Sydney'`
 
 	cases := []struct {
 		fireTime   string
@@ -742,37 +746,37 @@ func TestTimeMuteStage(t *testing.T) {
 	}{
 		{
 			// Friday during business hours
-			fireTime:   "01 Jan 21 09:00 +0000",
+			fireTime:   "19 Nov 21 13:00 AEDT",
 			labels:     model.LabelSet{"foo": "bar"},
 			shouldMute: false,
 		},
 		{
 			// Tuesday before 5pm
-			fireTime:   "01 Dec 20 16:59 +0000",
+			fireTime:   "16 Nov 21 16:59 AEDT",
 			labels:     model.LabelSet{"dont": "mute"},
 			shouldMute: false,
 		},
 		{
 			// Saturday
-			fireTime:   "17 Oct 20 10:00 +0000",
+			fireTime:   "20 Nov 21 10:00 AEDT",
 			labels:     model.LabelSet{"mute": "me"},
 			shouldMute: true,
 		},
 		{
 			// Wednesday before 9am
-			fireTime:   "14 Oct 20 05:00 +0000",
+			fireTime:   "17 Nov 21 05:00 AEDT",
 			labels:     model.LabelSet{"mute": "me"},
 			shouldMute: true,
 		},
 		{
-			// Ensure comparisons are UTC only. 12:00 KST should be muted (03:00 UTC)
-			fireTime:   "14 Oct 20 12:00 +0900",
+			// Ensure comparisons with other time zones work as expected.
+			fireTime:   "14 Nov 21 20:00 KST",
 			labels:     model.LabelSet{"mute": "kst"},
 			shouldMute: true,
 		},
 		{
-			// Ensure comparisons are UTC only. 22:00 KST should not be muted (13:00 UTC)
-			fireTime:   "14 Oct 20 22:00 +0900",
+			// Ensure comparisons with other time zones work as expected.
+			fireTime:   "14 Nov 21 22:00 KST",
 			labels:     model.LabelSet{"kst": "dont_mute"},
 			shouldMute: false,
 		},
@@ -788,7 +792,7 @@ func TestTimeMuteStage(t *testing.T) {
 	outAlerts := []*types.Alert{}
 	nonMuteCount := 0
 	for _, tc := range cases {
-		now, err := time.Parse(time.RFC822Z, tc.fireTime)
+		now, err := time.Parse(time.RFC822, tc.fireTime)
 		if err != nil {
 			t.Fatalf("Couldn't parse fire time %s %s", tc.fireTime, err)
 		}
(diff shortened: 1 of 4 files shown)


benridley:
It seems I've fixed the tests that were failing, but I'm not sure what's going wrong with the 'mixin' part of the CI pipeline, it looks unrelated to my changes...

PhilipGough:
@benridley looks like this was addressed in #2785 so a rebase on top of main should resolve this for you

benridley:
Thanks @PhilipGough, I've just rebased so we'll see how it goes.

timeinterval/timeinterval.go
```diff
@@ -405,6 +457,13 @@ func clamp(n, min, max int) int {
 
 // ContainsTime returns true if the TimeInterval contains the given time, otherwise returns false.
 func (tp TimeInterval) ContainsTime(t time.Time) bool {
+	var location *time.Location
+	if tp.TimeZone == nil {
+		location = time.UTC
+	} else {
+		location = tp.TimeZone.Location
+	}
+	t = t.In(location)
```
sylr, line 466:
```suggestion
	if tp.TimeZone != nil {
		t = t.In(tp.TimeZone.Location)
	}
```

timeinterval/timeinterval_test.go
```diff
@@ -434,7 +480,7 @@ func TestYamlUnmarshal(t *testing.T) {
 			t.Errorf("Error unmarshalling %s: Want %+v, got %+v", tc.in, tc.intervals, ti)
 		}
 		for _, ts := range tc.contains {
-			_t, _ := time.Parse(time.RFC822, ts)
+			_t, _ := time.Parse(time.RFC822Z, ts)
```
sylr, line 483: I wouldn't have changed that. The purpose of those tests is to ensure non regression so if we change all the test suit we can't be sure we didn't break things. Also I don't see the benefit to use numeric timezones instead of the original short codes. I would revert this and use AEDT / AEST for the new test cases related to timezones.

timeinterval/timeinterval.go
benridley: Nice catch, thanks

timeinterval/timeinterval_test.go
benridley: I changed these because the original format (RFC822) doesn't actually work the way I thought it did (I wrote the original tests as well). It doesn't actually support timezones like "AEDT". I realized this because it was failing on my new tests, until I read the RFC and as it turns out it only supports 4 static timezones. So I changed to this format so I could test with more timezones, including changing between daylight savings etc.

timeinterval/timeinterval.go
```diff
@@ -165,6 +172,31 @@ var monthsInv = map[int]string{
...
+	if err := unmarshal(&str); err != nil {
+		return err
+	}
+
+	loc, err := time.LoadLocation(str)
+	if err != nil {
+		if runtime.GOOS == "windows" && str != "Local" && str != "UTC" {
+			return fmt.Errorf("unable to load time location. Timezones other than 'Local' and 'UTC' may not be supported on Windows without using a custom timezone file: %w", err)
+		}
+		return err
```
simonpasquier, line 187: I'm wondering if we could make the error message more targeted.

```suggestion
		if runtime.GOOS == "windows" {
			if zoneinfo := os.GetEnv("ZONEINFO"); zoneinfo != "" {
				return fmt.Errorf("%w (ZONEINFO=%q)", err, zoneinfo)
			}
			return fmt.Errorf("%w (on Windows platforms, you may have to pass the time zone database using the ZONEINFO environment variable, see https://pkg.go.dev/time#LoadLocation for details)", err)
		}
		return err
```

benridley: Thanks @simonpasquier. 

Yeah I think that's a good idea. I've also been thinking more, and I'm wondering if we just go the extra distance and embed the timezone database on Windows platforms? Obviously it's not ideal because we can have stale timezone information between releases of AM, but I think that's probably better than nothing. What do you think?

beorn7:
As discussed elsewhere, while "timezone" in the context of the TZ database indeed refers to a location (e.g. "Australia/Sydney"), the general understanding of "timezone" is more ambiguous. In many other contexts, a timezone is something like "PDT" or "PST" (or even "PST/PDT" :scream: ). As a concrete example particularly relevant for Go developers, look at the terminology in https://pkg.go.dev/time#Time .

Therefore, I suggest to use a word like "location". Even if you come from a TZ database context, you will still understand what that is, while those from other contexts will have a much easier life (not trying to enter "PDT" and wondering why it doesn't work).

benridley:
Sure @beorn7, I think that's a good suggestion and keeps the terminology consistent with Go's stdlib. Happy to add this.

timeinterval/timeinterval.go
simonpasquier: I would defer it for a later PR if people feel that it's really needed. As you said, we're at risk of providing stale information.

simonpasquier:
Great feedback from @beorn7 (as usual) on location vs. time zone, +1 for me.

piotr1212:
Hey, wondering if this is still being worked on as we'd need this feature. Also assuming this works well with daylight saving?

benridley:
Hey @piotr1212,

Yeah it's coming - I was mainly waiting on #2779 because I knew I'd need to rebase and fix this code up once that was merged. I'll probably work on it this weekend and get it tidied up.

And yes, it will work fine with daylight savings - Provided you aren't running AM on Windows.

benridley:
I've just rebased onto main, and also added in the following:
- Renamed Timezone/time_zone to Location/location as its a more accurate description and consistent with Go's stdlib (thanks @beorn7)
- Added in @simonpasquier's suggestion to be more specific with the Windows error and check.
- Fixed up the docs to use the 'location' keyword, and also added some more details about how it works.

timeinterval/timeinterval_test.go
benridley: To add some clarification to this (in case it's a blocker for merging), using RFC822 with timezones like AEDT / AEST just falls back to UTC, because it doesn't understand those time zones at all. The new format (RFC822Z) allows us to test the behaviour correctly by using numeric offsets.

sylr:
@simonpasquier could we go ahead and merge this please ?

notify/notify_test.go
```diff
@@ -724,16 +724,20 @@ func TestMuteStageWithSilences(t *testing.T) {
 }
 
 func TestTimeMuteStage(t *testing.T) {
-	// Route mutes alerts outside business hours if it is a mute_time_interval
+	// Route mutes alerts outside business hours in November, using the +1100 timezone.
 	muteIn := `
 ---
 - weekdays: ['monday:friday']
+  time_zone: 'Australia/Sydney'
```
sylr, line 731:
```suggestion
  location: 'Australia/Sydney'
```

notify/notify_test.go
```diff
@@ -724,16 +724,20 @@ func TestMuteStageWithSilences(t *testing.T) {
...
+  months: ['November']
   times:
    - start_time: '00:00'
      end_time: '09:00'
    - start_time: '17:00'
      end_time: '24:00'
-- weekdays: ['saturday', 'sunday']`
+- weekdays: ['saturday', 'sunday']
+  months: ['November']
+  time_zone: 'Australia/Sydney'`
```
sylr, line 740:
```suggestion
  location: 'Australia/Sydney'`
```

timeinterval/timeinterval.go
```diff
@@ -33,6 +35,7 @@ type TimeInterval struct {
 	DaysOfMonth []DayOfMonthRange `yaml:"days_of_month,flow,omitempty" json:"days_of_month,omitempty"`
 	Months      []MonthRange      `yaml:"months,flow,omitempty" json:"months,omitempty"`
 	Years       []YearRange       `yaml:"years,flow,omitempty" json:"years,omitempty"`
+	Location    *Location         `yaml:"time_zone,flow,omitempty" json:"time_zone,omitempty"`
```
sylr, line 38:
```suggestion
	Location    *Location         `yaml:"location,flow,omitempty" json:"location,omitempty"`
```

timeinterval/timeinterval_test.go
```diff
@@ -378,6 +409,21 @@ var yamlUnmarshalTestCases = []struct {
 			},
 		},
 	},
+	{
+		// Time zones may be specified by location.
+		in: `
+---
+- years: ['2020:2022']
+  time_zone: 'Australia/Sydney'
```
sylr, line 417:
```suggestion
  location: 'Australia/Sydney'
```

benridley:
Thanks for those test corrections @sylr, I've just merged them.

simonpasquier (approved):
lgtm.

We could maybe add a blank import to https://pkg.go.dev/time/tzdata in cmd/alertmanager/main.go to embed a copy of the timezone database but it can be done later if we see fit.

PR #2798, opened 2021-12-22
Automate CSS-inlining for default HTML email template

### LINKED ISSUE #1939: template/default.tmpl and template/email.tmpl are out of sync
Presumably there is some premailer or inline-styler used to build `template/default.tmpl`'s `email.default.html` value from `template/email.html`. However, from a cursory glance, it appears out of sync on https://github.com/prometheus/alertmanager/pull/1475. Specifically that PR didn't update the default template HTML.

Are they out of sync by intention? What tool is used to generate the inline-style versioned? I came across this when working with my own templates.

bison:
The original HTML email template was added in `template/email.html`.
It looks like the CSS was manually inlined.  Most likely using the
premailer.dialect.ca web form, which is mentioned in the README for
the Mailgun transactional-email-templates project.  The resulting HTML
with inlined CSS was then copied into `template/default.tmpl`.  This
has resulted in `email.html` and `default.tmpl` diverging at times.

This commit adds build automation to inline the CSS automatically
using [juice][1].  The Go template containing the resulting HTML has
been moved into its own file to avoid the script that performs the CSS
inlining having to parse the `default.tmpl` file to insert it there.

Fixes #1939.

[1]: https://www.npmjs.com/package/juice

Here are examples of HTML emails after inlining the CSS with Juice:

juice-firing
juice-resolved

And the same emails with the existing template for comparison:

current-firing
current-resolved


simonpasquier:
thanks, lgtm in general but I'm not sure why the CI is `Waiting for status to be reported` :thinking:

bison:
Not sure what the deal was with CI, but it looks like it's passed now.

PR #2808, opened 2021-12-30
add SNS receiver validation

qinxx108:
1. Add validation for subject, if not a validate input, we will replace it with error message
2. Add validation for message attribute, if not a validate attribute, we will remove the attribute
3. Add validation for message, if not a validate message, we will replace the message with error message
4. Add truncation logic that will include message attributes as SNS receiver count 256KB size limit with message body+attribute

### DIFF
--- a/notify/sns/sns.go
+++ b/notify/sns/sns.go
@@ -15,9 +15,13 @@ package sns
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"github.com/pkg/errors"
 	"net/http"
+	"regexp"
 	"strings"
+	"unicode"
 	"unicode/utf8"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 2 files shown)


simonpasquier:
:wave: could you split the change into several pull requests. There are things that make sense to me but others that would need prior discussion and agreement.

> Add validation for subject, if not a validate input, we will replace it with error message

I'd be fine to truncate the subject when we detect that it's larger than the max size. I'm less convinced that replacing an invalid subject with an error message is a good approach.

> Add validation for message attribute, if not a validate attribute, we will remove the attribute

that might be acceptable.

> Add validation for message, if not a validate message, we will replace the message with error message

Same remark as for the subject.

> Add truncation logic that will include message attributes as SNS receiver count 256KB size limit with message body+attribute

There could be a discussion about which fields should have priority.

PR #2815, opened 2022-01-12
api/v2: Delete silence respond with 404 when silence is not found

hateeyan:
Hi,
When a client attempts to delete a non-existent silence, respond with 404 instead of 500.

Signed-off-by: hateeyan <hateeyan@gmail.com>

PR #2817, opened 2022-01-12
Return no error when deleting expired silence

soonping-amzn:
Delete silence currently returns HTTP 500 if called on an expired silence.  This makes it non-idempotent, as the following call sequence can result:

```
T0 initial: DELETE /api/v1/silence/MySilence => Timeout (but actually succeeded)
T1 retry:   DELETE /api/v1/silence/MySilence => 500
```

After this sequence, the user is left uncertain whether there's a problem in AlertManager/Prometheus.  Even fetching the silence to verify its state won't confirm whether the 500 was indicative of a separate underlying problem that could manifest in other ways later.

Prometheus exposes the delete_series API which does exhibit idempotent behaviour -- calling delete_series on an already deleted series returns the same status as the initial delete_series call:  HTTP 200.

This PR changes Silences.expire(id) to return nil for expired silences, which impacts APIs as follows:
- DELETE silence will no longer fail if the silence was already expired
- PUT silence will no longer fail if the silence exists and was already expired
  https://github.com/prometheus/alertmanager/blob/c91717bc89aadd3253063ddca9f56a7c4a4c2736/silence/silence.go#L561-L562

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -622,7 +622,7 @@ func (s *Silences) expire(id string) error {
 
 	switch getState(sil, now) {
 	case types.SilenceStateExpired:
-		return errors.Errorf("silence %s already expired", id)
+		return nil
 	case types.SilenceStateActive:
 		sil.EndsAt = now
 	case types.SilenceStatePending:
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -622,6 +622,7 @@ func (s *Silences) expire(id string) error {
 
 	switch getState(sil, now) {
 	case types.SilenceStateExpired:
+		// returning nil ensures idempotent behaviour, at least in the short term before the silence is gc'd
 		return nil
 	case types.SilenceStateActive:
 		sil.EndsAt = now
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -622,7 +622,7 @@ func (s *Silences) expire(id string) error {
 
 	switch getState(sil, now) {
 	case types.SilenceStateExpired:
-		// Returning nil ensures idempotent behaviour, at least in the short term before the silence is gc'd
+		// Returning nil ensures idempotent behaviour, at least in the short term before the silence is gc'd.
 		return nil
 	case types.SilenceStateActive:
 		sil.EndsAt = now
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -612,6 +612,8 @@ func (s *Silences) Expire(id string) error {
 }
 
 // Expire the silence with the given ID immediately.
+// It is idempotent, nil is returned if the silence already expired before it is GC'd.
+// If the silence is not found an error is returned.
 func (s *Silences) expire(id string) error {
 	sil, ok := s.getSilence(id)
 	if !ok {
@@ -622,7 +624,6 @@ func (s *Silences) expire(id string) error {
 
 	switch getState(sil, now) {
 	case types.SilenceStateExpired:
-		// Returning nil ensures idempotent behaviour, at least in the short term before the silence is gc'd.
 		return nil
 	case types.SilenceStateActive:
 		sil.EndsAt = now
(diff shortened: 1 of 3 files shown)


soonping-amzn:
Note: a PR (#2815) was posted shortly before this one addressing a related but distinct issue:  return a 404 instead of a 500 for non-existent silences.

The 2 PRs address _different_ issues arising from delete silence returning 500 on any error.

roidelapluie (commented):
I think it makes sense, I would suggest to add a comment in the code to make it clear that we want this method to be indempotent.

Note that this is not guaranteed to work indefinitvely because expired silences get garbage collected at some point.

cc @simonpasquier @gotjosh  for a second look.

roidelapluie (approved):

alvinlin123:
@simonpasquier @gotjosh, sorry to interrupt, but would you be able to take some time to take a look at this PR soon?

silence/silence.go
```diff
@@ -622,7 +622,8 @@ func (s *Silences) expire(id string) error {
 
 	switch getState(sil, now) {
 	case types.SilenceStateExpired:
-		return errors.Errorf("silence %s already expired", id)
+		// returning nil ensures idempotent behaviour, at least in the short term before the silence is gc'd
```
roidelapluie, line 625:
```suggestion
		// Returning nil ensures idempotent behaviour, at least in the short term before the silence is gc'd.
```

silence/silence.go
```diff
@@ -622,7 +622,8 @@ func (s *Silences) expire(id string) error {
 
 	switch getState(sil, now) {
 	case types.SilenceStateExpired:
-		return errors.Errorf("silence %s already expired", id)
+		// Returning nil ensures idempotent behaviour, at least in the short term before the silence is gc'd
```
roidelapluie, line 625:
```suggestion
		// Returning nil ensures idempotent behaviour, at least in the short term before the silence is gc'd.
```

roidelapluie:
Sorry for the nits, our styleguide for comments is capitalized sentences with full stop. Once correct I'd merge this pull request.

soonping-amzn:
No problem. Apologies for missing the full stop in your previous recommended change.

silence/silence.go
```diff
@@ -622,7 +622,8 @@ func (s *Silences) expire(id string) error {
 
 	switch getState(sil, now) {
 	case types.SilenceStateExpired:
-		return errors.Errorf("silence %s already expired", id)
```
gotjosh, line 625: Apologies for coming in late - I feel that this kind of comment about the semantics of the function is best put as part of the function description.

How do you feel about changing the top-level comment from:

```
// Expire the silence with the given ID immediately.
```

to

```
// Expire the silence with the given ID immediately. It is idempotent, nil is returned if the silence already expired before it is GC'd.
// If the silence is not found an error is returned.
```

gotjosh:
I'm a bit surprised that the CI passes here - if we don't have a test at an API level, we should create one.

silence/silence.go
soonping-amzn: Makes sense. I'll make the change.

simonpasquier (commented):
agreed with all the comments from @roidelapluie and @gotjosh

soonping-amzn:
I've updated the function comment, and added tests for DeleteSilence and PostSilences.

Found during testing that PostSilences actually probably won't change in behaviour, as expire() is only called if the existing silence is unexpired.  i.e. The test passed even without my change to the behaviour of expire().  If you want me to remove that test, let me know.

https://github.com/prometheus/alertmanager/blob/c91717bc89aadd3253063ddca9f56a7c4a4c2736/silence/silence.go#L559-L564

soonping-amzn:
@simonpasquier @gotjosh Please let me know if you think I need to add to or change the tests I've added, and if anything else is needed.  Thank you

simonpasquier:
LGTM

simonpasquier:
thanks!

PR #2819, opened 2022-01-21
notify/sns: detect FIFO topic based on the rendered value

simonpasquier:
Since the TopicARN field is a template string, it's safer to check for
the ".fifo" suffix in the rendered string.

@roidelapluie does it make sense?


roidelapluie (approved):

roidelapluie:
LGTM

PR #2822, opened 2022-01-27
provider: Avoid receive alerts after the listener closed

johncming:
The closed listener takes 30m to `gc` by default, and it is possible to append an alert to the closed listener during the `Put` operation in this time.

The pr should not be a bug, just optimization.

### DIFF
--- a/provider/mem/mem.go
+++ b/provider/mem/mem.go
@@ -202,6 +202,12 @@ func (a *Alerts) Put(alerts ...*types.Alert) error {
 		a.mtx.Lock()
 		for _, l := range a.listeners {
 			select {
+			case <-l.done:
+				continue
+			default:
+			}
+
+			select {
 			case l.alerts <- alert:
 			case <-l.done:
 			}


provider/mem/mem.go
```diff
@@ -201,6 +201,12 @@ func (a *Alerts) Put(alerts ...*types.Alert) error {
 
 		a.mtx.Lock()
 		for _, l := range a.listeners {
+			select {
+			case <-l.done:
+				continue
+			default:
+			}
+
```
sthaha, line 209: Would it be better to rely only on alerts channel being open? e.g.

```suggestion
		for _, l := range a.listeners {
			if a.alerts == nil {
			  continue
			}

			select {
			case l.alerts <- alert:
			}
```
and elsewhere when the channel is closed, 
```
select {
case <-l.done:
  close(l.alerts)
  l.alerts = nil
  // perhaps even  l.done = nil
}
```

johncming: Because of the lock protection, the problem described by the pr will not occur, so I  closed this pr. Thanks.

PR #2823, opened 2022-02-01
Fix markdown list describing Alertmanager matchers

juliusv:
The Markdown variant used on prometheus.io is different from the GitHub one and
requires code blocks to be indented by four spaces to be recognized as part of
a list item vs. terminating the list.

<hr/>

**Before:**

broken-format

<hr/>

**After:**

fixed-format


juliusv:
@asquare14

asquare14:
LGTM ! :)

PR #2835, opened 2022-02-14
validate sns message body

qinxx108:
Add validation for message, if not a validate message, we will replace the message with error message
Separate from https://github.com/prometheus/alertmanager/pull/2808

### DIFF
--- a/notify/sns/sns.go
+++ b/notify/sns/sns.go
@@ -36,6 +36,12 @@ import (
 	"github.com/prometheus/alertmanager/types"
 )
 
+const (
+	// The errors
+	MessageNotValidUtf8 = "Error - not a valid UTF-8 encoded string"
+	MessageIsEmpty      = "Error - the message should not be empty"
+)
+
 // Notifier implements a Notifier for SNS notifications.
 type Notifier struct {
 	conf    *config.SNSConfig
@@ -189,7 +195,10 @@ func (n *Notifier) createPublishInput(ctx context.Context, tmpl func(string) str
 
 func validateAndTruncateMessage(message string, maxMessageSizeInBytes int) (string, bool, error) {
 	if !utf8.ValidString(message) {
-		return "", false, fmt.Errorf("non utf8 encoded message string")
+		return MessageNotValidUtf8, false, fmt.Errorf("non utf8 encoded message string")
+	}
+	if len(message) == 0 {
+		return MessageIsEmpty, false, fmt.Errorf("message is empty")
 	}
 	if len(message) <= maxMessageSizeInBytes {
 		return message, false, nil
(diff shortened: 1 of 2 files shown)


simonpasquier (commented):
The code could use [`strings.ToValidUTF8`](https://pkg.go.dev/strings#ToValidUTF8) to eliminate invalid UTF-8 byte sequence rather than dropping everything.

qinxx108:
> The code could use [`strings.ToValidUTF8`](https://pkg.go.dev/strings#ToValidUTF8) to eliminate invalid UTF-8 byte sequence rather than dropping everything.

I wonder which will be better in case of user experience. The subject is replaced by some "error" alerting your subject is invalid VS the subject is changed without your notice


On the other hand https://pkg.go.dev/strings#ToValidUTF8 can also return you an empty string as described here: "ToValidUTF8 returns a copy of the string s with each run of invalid UTF-8 byte sequences replaced by the replacement string, which may be empty."

alvinlin123:
My opinion is dropping everything is safer option because it will be very obvious to the receiver that something is wrong. For example, in the case where message body is a JSON, replacing the whole body would choke the receiver's JSON parser - a very clear signal something is wrong; if we replace invalid character, then the JSON parsing would succeed but the receiver consumes incomplete information (imagine we replace a invalid "0", in someone's bank account).

However, I do think there are cases where replacing invalid character would be the right thing, but by default I feel we should go with the safest behaviour. In the future, we can have configuration options to let users configure different behaviour upon invalid SNS message body.

roidelapluie:
in prometheus we generally drop when there is an error rather than potentially misleading the users, I think I prefer to have the explicit error as well.

qinxx108:
Thanks @roidelapluie for your comments. I wonder if everybody is okay with current approach, can i get approved for this PR?

Thank you so much

Yijie

notify/sns/sns.go
```diff
@@ -36,6 +36,12 @@ import (
 	"github.com/prometheus/alertmanager/types"
 )
 
+const (
+	// The errors
```
LeviHarrison, line 40:
```suggestion
	// Error messages to replace an invalid message with.
```

LeviHarrison (commented):
Why are we replacing messages with error messages if those messages will never be sent? Or maybe I've misread the code.

Other than that this looks good (after rebase).

SoloJacobs (commented):
Hello @qinxx108 ,

this PR cannot be merged in its current form, since the comments made by @LeviHarrison need to be addressed. To paraphrase his review: Your change does not pass the CI validation. You can check this by running `make` from the root of the repository. More importantly, if `validateAndTruncateMessage` returns an error, then the `publishInput` is never sent. Thus, your change does not affect any behaviour. What would really help here is a clear description of what you are trying to achieve.

SoloJacobs:
Since some time has passed since  the review of @LeviHarrison, I'll close this PR. Feel free to open a new PR, if you feel that you can address the above issues. Thank you for your time.

Kind regards

PR #2837, opened 2022-02-28
Add Webex notifier support

JTBlanchard:
This adds support for sending notifications to Webex.

This should follow the prior notifier examples pretty closely, but there is one thing that may stand out:  The Webex attachment content field expects a JSON string rather than an object.  Since the Adaptive Card format used in attachments doesn't have fixed fields and has an evolving schema, using a string makes templating easier/less fragile than the the recursion and use of reflect to apply templating to each field.  However, since most users will likely be copying and pasting the JSON from the output of a UI editor, this is likely to be simpler for the user.


gotjosh:
We already have a Webex notifier.

I think I completely missed this contribution, sorry.

PR #2848, opened 2022-03-15
chore: fix build of alertmanager without docker

jmichalek132:
Hi, so we build alertmanager from source because of some internal requirements. We do this inside docker as part of multi stage build. We also don't have docker in docker available.
We were using the `NO_DOCKER` env variable to make the build not use docker. 
However this does work on line 58, where `DOCKER_RUN_CURRENT_USER` is used instead `DOCKER_CMD` this one is not set to empty when `NO_DOCKER` env variable is set to true.
This PR fixes that and makes the build work without docker available again.


jmichalek132:
Hi, @simonpasquier could you please take a look at this small change once you have some time?

Thank you in andvance.

metalmatze (approved):
LGTM

simonpasquier (approved):
lgtm

simonpasquier:
oh wait! Would this work as you expect? When I tried locally, I get `rm: invalid option -- 'g'` which makes sense looking at this part.

https://github.com/prometheus/alertmanager/blob/ec2f5f589402d8b45e327555d1330ce780354221/ui/app/Makefile#L58-L61

jmichalek132:
> oh wait! Would this work as you expect? When I tried locally, I get `rm: invalid option -- 'g'` which makes sense looking at this part.
> 
> https://github.com/prometheus/alertmanager/blob/ec2f5f589402d8b45e327555d1330ce780354221/ui/app/Makefile#L58-L61

That's interesting, yeah I what you are point out by for some reason it doesn't happen for us in our pipeline.
Basically when I use the docker image we use for building alertmanager, set NO_DOCKER to true and run make scriptjs it works.

PR #2849, opened 2022-03-15
Add CLI args for snapshot intervals

### LINKED ISSUE #2660: Add command line argument for silences snapshot period
Currently the period for taking silences snapshots is fixed on 15 minutes.
For non-HA it could be useful to be able to reduce this value.

https://github.com/prometheus/alertmanager/blob/3d86bd709df88e2b95a09636cf36b77a9d79af51/cmd/alertmanager/main.go#L300

sed-i:
This PR adds two new CLI args to support custom snapshot intervals for nflogs and silences.

Fixes #2660.

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -195,6 +195,9 @@ func run() int {
 		retention       = kingpin.Flag("data.retention", "How long to keep data for.").Default("120h").Duration()
 		alertGCInterval = kingpin.Flag("alerts.gc-interval", "Interval between alert GC.").Default("30m").Duration()
 
+		silencesInterval = kingpin.Flag("silences.interval", "Interval between writing silences snapshot to disk.").Default("15m").Duration()
+		nflogsInterval   = kingpin.Flag("nflogs.interval", "Interval between writing nflogs snapshot to disk.").Default("15m").Duration()
+
 		webConfig      = webflag.AddFlags(kingpin.CommandLine)
 		externalURL    = kingpin.Flag("web.external-url", "The URL under which Alertmanager is externally reachable (for example, if Alertmanager is served via a reverse proxy). Used for generating relative and absolute links back to Alertmanager itself. If the URL has a path portion, it will be used to prefix all HTTP endpoints served by Alertmanager. If omitted, relevant URL components will be derived automatically.").String()
 		routePrefix    = kingpin.Flag("web.route-prefix", "Prefix for the internal routes of web endpoints. Defaults to path of --web.external-url.").String()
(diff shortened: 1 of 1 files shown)


cmd/alertmanager/main.go
```diff
@@ -195,6 +195,9 @@ func run() int {
 		retention       = kingpin.Flag("data.retention", "How long to keep data for.").Default("120h").Duration()
 		alertGCInterval = kingpin.Flag("alerts.gc-interval", "Interval between alert GC.").Default("30m").Duration()
 
+		silencesInterval = kingpin.Flag("silences.interval", "Interval between writing silences snapshot to disk.").Default("15m").Duration()
```
simonpasquier, line 198:
```suggestion
		silencesInterval = kingpin.Flag("silences.maintenance-interval", "Interval between garbage collection and snapshotting to disk of the silences.").Default("15m").Duration()
```

cmd/alertmanager/main.go
```diff
@@ -195,6 +195,9 @@ func run() int {
...
+		nflogsInterval   = kingpin.Flag("nflogs.interval", "Interval between writing nflogs snapshot to disk.").Default("15m").Duration()
```
simonpasquier, line 199:
```suggestion
		nflogsInterval   = kingpin.Flag("nflogs.maintenance-interval", "Interval between garbage collection and snapshotting to disk of the notification logs.").Default("15m").Duration()
```

gotjosh:
Overall, this looks good. Do you have a use-case for separating (and having to add two new flags, albeit optional), or can we stick with just one for both? I can definitely see the use case for changing the interval but not why you would have separate ones.

sed-i:
> Overall, this looks good. Do you have a use-case for separating (and having to add two new flags, albeit optional), or can we stick with just one for both? I can definitely see the use case for changing the interval but not why you would have separate ones.

For my use-case we can stick with one flag only. It makes sense to me. I wasn't sure the two always go hand-in-hand.

Some ideas for a name for a single flag:
1. `snapshot.interval`
2. `maintenance.interval`
3. `maintenance.gc-interval`

simonpasquier:
That's a good point @gotjosh :) 
I've don't have a strong feeling but since `--data.retention` already applies to silences and nflogs, it would make sense to have `--data.maintenance-interval`. What do you think?

gotjosh:
> --data.maintenance-interval

👍 on the idea of this name.

sed-i:
Changed to `data.maintenance-interval`.
@gotjosh would you be able to rerun the failed test? Seems like an unrelated timeout error.

simonpasquier (approved):
lgtm

gotjosh (approved):
Sadly, I can't kick off the CI but @simonpasquier was here to do so.

LGTM.

simonpasquier:
thanks!

PR #2853, opened 2022-03-18
Update golangci-lint and fix complaints

metalmatze:
I've copied the `.golangci.yml` from prometheus/prometheus and then updated the Go files according to the given output.

The only change is that I've added all default rules for revive but enabled `disableStutteringCheck` to prevent the linter from complaining about the struct names such as `cluster.ClusterPeer` and more. I thought that might be too invasive in the end.

Let me know if there's anything that should be changed on the golangci-lint side rather than in the code, though keep in mind it's what Prometheus does too.


template/template.go
```diff
@@ -20,12 +20,12 @@ import (
 	"net/url"
 	"path"
 	"path/filepath"
-	"regexp"
 	"sort"
 	"strings"
 	tmpltext "text/template"
 	"time"
 
+	"github.com/grafana/regexp"
```
simonpasquier, line 28: :thinking:

test/cli/acceptance.go
```diff
@@ -33,6 +32,7 @@ import (
 
 	httptransport "github.com/go-openapi/runtime/client"
 	"github.com/go-openapi/strfmt"
+	"github.com/grafana/regexp"
```
simonpasquier, line 35: same

simonpasquier (commented):
yamllint is nit-picky on `.golangci.yml`. I've spotted a few import replacements that don't look correct but otherwise :+1:

template/template.go
metalmatze: This is in line with https://github.com/prometheus/prometheus/pull/10251

metalmatze:
All green now @simonpasquier :relaxed: :tada:

gotjosh (approved):
LGTM

template/template.go
gotjosh: I believe this is only temporary, has Bryan has submitted all of the patches of grafana/regexp to go already.

roidelapluie: I don't know if that is needed for alertmanager.

gotjosh: Right, I don't think we've dealt with performance issues in the area so I'm indifferent about the change.

The only thing to consider here would be @metalmatze's point on the alignment with Prometheus which I guess is a net positive?

gotjosh: That's a good shout @roidelapluie. @metalmatze can we remove this dependency and stick with Go's regexp then?

metalmatze: Yes, this is only temporary indeed. I'll remove it then and stick with stdlib. No problem. 
AFAIK the package is mostly for Loki dealing with lots of regexp matches in each log line.

gotjosh (approved):
LGTM

simonpasquier (commented):
:+1: very cool, thanks!

PR #2855, opened 2022-03-20
Provide shell environment variable to yml file

Maissacrement:


gotjosh:
Sorry for coming across as a bit abrasive here, but what exactly is the goal of this PR?

From a quick look at other repositories (Prometheus, Cortex) I don't think this is something that we do so precedent is not there. Am I missing something @simonpasquier @roidelapluie?

roidelapluie:
Prometheus only supports environment variables in very specific areas of the configuration.

Maissacrement:
I encountered some issue with alertmanager who is already provide by prometheus. I talking about smtp authentication where i need to rewrite alertmanager yml configuration to add secrets, urls and more in my repository for production.

I thinks we can challenged data with `curly brace bracket` to distinguish data as SHELL environment ?

peterbourgon:
No reason for the config file to support this natively.

```
export FOO=bar
cat alertmanager.yaml.template | envsubst > alertmanager.yaml
```

Wilderone:
@peterbourgon this is not so simple then you deploy alertmanager within kubernetes through helm chart. Yes, it is still possible, but with huge overhead just for simple substitution

roidelapluie:
The way to deal with this is to put env variables into files and read from files. A lot of useful fields already support that. What is your use case?

Maissacrement:
pragmatically in devops we are confronted with problems which require to quickly operate solutions for our customers in a non-binding way.
So I worked on this POC in order to inject the secrets into a container at a moment in order to provide the configuration files as data and to automatically deploy to a target container
https://github.com/Maissacrement/secret-manager

simonpasquier (commented):
I think that I understand what you're trying to achieve here: it's something that we've already discussed before and we've always considered that it should be the responsibility for your config management system rather than Alertmanager.
A change of direction would need to be discussed for all the Prometheus projects (e.g. Prometheus + exporters).

SuperQ:

PR #2867, opened 2022-03-29
Issue #2850: Add benbjohnson/clock to the silences package.

### LINKED ISSUE #2850: Feature Request: Support clock injection for testing and mocking.
In Grafana we [import some prometheus alertmanager libraries.](https://github.com/grafana/grafana/blob/93390b5a1ee430670ba1c7db7e06bbe80cd252f0/go.mod#L82)

When testing our usage of the alertmanager, it'd be convenient to be able to manipulate the clock. For example, silence maintenance happens on a time.Ticker, and silence expirations are based on real wall-clock time. If we could manipulate time, we can test Silence maintenance (we use a custom maintenance function) quickly in unit tests rather than waiting on real clock time.

In [the clock library,](https://pkg.go.dev/github.com/benbjohnson/clock) you create a clock.Clock interface - in real systems this is a wrapper around the time libraries, and during testing you use a MockClock to manipulate time.

Proposal: Depend on benbjohnson/clock and inject a clock.Clock any where a `time.Now()` or `time.Ticker` is used. The library is MIT-licensed.

There are 6 usages of time.Ticker and 43 appearances of time.Now, including in tests. It's not necessary to instantiate only a single clock.Clock and thread it down to all dependencies or anything like that. Clocks could be instantiated at the construction of the objects that make time calls.

(I would be willing to do this change if the proposal is acceptable)

joeblubaugh:
github.com/benbjohnson/clock provides a time interface to programs
rather than using the stdlib time package. This allows mocking time in
programs and tests. In this commit, the clock is used to speed up and
simplify testing of the silences package.

Fixes #2850

### DIFF
--- a/silence/silence_test.go
+++ b/silence/silence_test.go
@@ -20,7 +20,6 @@ import (
 	"os"
 	"runtime"
 	"sort"
-	"sync/atomic"
 	"testing"
 	"time"
 
@@ -30,6 +29,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/atomic"
 
 	pb "github.com/prometheus/alertmanager/silence/silencepb"
 	"github.com/prometheus/alertmanager/types"
@@ -211,9 +211,9 @@ func TestSilences_Maintenance_SupportsCustomCallback(t *testing.T) {
 	s := &Silences{st: state{}, logger: log.NewNopLogger(), clock: clock, metrics: newMetrics(nil, nil)}
 	stopc := make(chan struct{})
 
-	var calls int32 = 0
+	calls := atomic.NewInt32(0)
 	go s.Maintenance(100*time.Millisecond, f.Name(), stopc, func() (int64, error) {
-		atomic.AddInt32(&calls, 1)
+		calls.Add(1)
 		return 0, nil
 	})
 	runtime.Gosched()
@@ -224,7 +224,7 @@ func TestSilences_Maintenance_SupportsCustomCallback(t *testing.T) {
 	close(stopc)
 
 	require.Eventually(t, func() bool {
-		return atomic.LoadInt32(&calls) == 2
+		return calls.Load() == 2
 	}, 1*time.Millisecond, 50*time.Microsecond)
 }
 
--- a/silence/silence_test.go
+++ b/silence/silence_test.go
@@ -225,7 +225,7 @@ func TestSilences_Maintenance_SupportsCustomCallback(t *testing.T) {
 
 	require.Eventually(t, func() bool {
 		return calls.Load() == 2
-	}, 1*time.Millisecond, 50*time.Microsecond)
+	}, 100*time.Millisecond, 1*time.Millisecond)
 }
 
 func TestSilencesSetSilence(t *testing.T) {
--- a/silence/silence_test.go
+++ b/silence/silence_test.go
@@ -18,11 +18,13 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"runtime"
 	"sort"
-	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
+	"github.com/benbjohnson/clock"
 	"github.com/go-kit/log"
 	"github.com/matttproud/golang_protobuf_extensions/pbutil"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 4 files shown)


silence/silence_test.go
```diff
@@ -974,23 +981,40 @@ func TestSilenceExpireWithZeroRetention(t *testing.T) {
...
 	require.NoError(t, s.Expire("expired"))
 
+	// Advance time again. Despite what the function name says, s.Expire() does
+	// not expire a silence. It sets the silence to EndAt the current time. This
+	// means that the silence is active immediately after calling Expire.
+	clock.Add(1 * time.Millisecond)
+
+	// Make sure we can still find our silences.
 	_, err = s.QueryOne(QIDs("pending"))
 	require.NoError(t, err)
```
simonpasquier, line 1008: do we really need this check? I believe that counting the number of silenes per state is enough.

simonpasquier (commented):
looks good! just a few nits/remarks :)

silence/silence_test.go
```diff
@@ -260,18 +264,16 @@ func TestSilencesSetSilence(t *testing.T) {
...
-	// setSilence() is always called with s.mtx locked()
-	go func() {
+	// setSilence() is always called with s.mtx locked() in the application code
+	func() {
 		s.mtx.Lock()
-		require.NoError(t, s.setSilence(sil, now))
-		s.mtx.Unlock()
+		defer s.mtx.Unlock()
+		require.NoError(t, s.setSilence(sil, nowpb))
 	}()
```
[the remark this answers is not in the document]
joeblubaugh: This errors out while the mutex is still held. I suppose that's OK in a test context.

silence/silence_test.go
joeblubaugh: I agree with you. If this check was exhaustive you'd do it three times, or include all three QIDs.

silence/silence_test.go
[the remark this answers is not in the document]
simonpasquier: I see your point now. Maybe this as a middle ground? (but really not a showstopper, I was more puzzled by the "unfamiliar" presence of the anonymous function here)

```suggestion
        s.mtx.Lock()
	err := s.setSilence(sil, nowpb))
	s.mtx.Unlock()
	require.NoError(t, err)
```

joeblubaugh:
@simonpasquier is there another Prom maintainer I could ask for a review? @codesome?

codesome:
I am not the best person for Alertmanager, sorry. I see @w0rm as a maintainer who can possibly help.

w0rm:
@codesome sorry, I can only help reviewing the ui code in Elm.

joeblubaugh:
The test failure here seems to be a transient timeout. I can't reproduce it when running the race detector locally and I can't re-run the CI.

joeblubaugh:
@gotjosh I've rebased on main - hopefully this fixes the test issues.

silence/silence_test.go
```diff
@@ -203,27 +207,25 @@ func TestSilences_Maintenance_DefaultMaintenanceFuncDoesntCrash(t *testing.T) {
...
+	runtime.Gosched()
+
+	clock.Add(100 * time.Millisecond)
 
-	time.Sleep(200 * time.Millisecond)
+	// Stop the maintenance loop. We should get exactly one more execution of the maintenance func.
 	close(stopc)
 
 	require.Eventually(t, func() bool {
-		mtx.Lock()
```
gotjosh, line 223: I was surprised to see the use of an `eventually` here now that we have a mock-able clock. You can get rid of the eventually call by making sure you wait until the maintenance function returns:

```golang
func TestSilences_Maintenance_SupportsCustomCallback(t *testing.T) {
	f, err := ioutil.TempFile("", "snapshot")
	require.NoError(t, err, "creating temp file failed")
	clock := clock.NewMock()
	s := &Silences{st: state{}, logger: log.NewNopLogger(), clock: clock, metrics: newMetrics(nil, nil)}
	stopc := make(chan struct{})

	calls := atomic.NewInt32(0)
	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
		s.Maintenance(100*time.Millisecond, f.Name(), stopc, func() (int64, error) {
			calls.Add(1)
			return 0, nil
		})
		wg.Done()
	}()
	runtime.Gosched()

	clock.Add(100 * time.Millisecond)

	// Stop the maintenance loop. We should get exactly one more execution of the maintenance func.
	close(stopc)
	<-stopc
	wg.Wait()

	require.EqualValues(t, calls.Load(), 2)
}
```

gotjosh: Definitely, a nit and can be addressed as part of a second PR if you wish.

gotjosh (approved):
LGTM

silence/silence_test.go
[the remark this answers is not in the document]
gotjosh: I'll merge off the basis that @simonpasquier called out this wasn't a showstopper, but please do let @joeblubaugh if this is worth following upon.

gotjosh:
@joeblubaugh thank you for your contribution!

Issue #2868, opened 2022-03-29
Feature request: Allow plain forwarding of alerts in routes


codesome:
### What

I would like to propose a feature (via config) to mention `plain_forward: true` (or some other name) under routing config which totally ignores group interval, group wait, repeat interval, and group by, and simply forwards all alerts that it comes across to the configured receivers.

### Why 

Grafana Cloud and Amazon Managed Prometheus (and probably more) uses Prometheus Alertmanager in the cloud offerings to send alerts. This adds hurdles in testing the Prometheus alert-generator compliance because alerts cannot be send back to the test suite properly as expected. This can be fixed up to some extent by having `groupinterval=1s, groupwait=1s, repeatinterval=100y, groupby=...`. But it won't forward duplicate alerts that are sent by alert-generator every minute. (setting repeatinterval=1m also has its problems).

There might also be other use cases with this that I don't know of yet.

beorn7:
Sounds almost like "no-op mode"…

Pardon me if I'm just naive here (I don't have a lot of context on the compliance test suite), but is Alertmanager really needed for compliance testing if it isn't really doing anything? Couldn't the test suite itself emulate the Alertmanager API and be configured as an Alertmanager in the alert-generator to test?

codesome:
The test suite acts like an alertmanager which receives alerts from the alert-generator (for example a Prometheus instance). There is no alertmanager in picture here.

The specific problem here is for the cloud providers; there is an alertmanager sitting in between the alert-generator and the test suite which is creating problems and it is not trivial to let alerts bypass this alertmanager in the cloud. Hence this feature request.

codesome:
I have opened https://github.com/prometheus/alertmanager/pull/2873 as a reference implementation for this

gotjosh:
> but is Alertmanager really needed for compliance testing if it isn't really doing anything?

Originally, compliance testing was meant to be only for the alert generator (Prometheus) as opposed to the alert receiver (Alertmanagert). However, after looking at the broader landscape compliance switched its approach to include the Alertmanager.

**Why?**

If we look at how cloud offerings (AWS as a non-Grafana example) ship "Prometheus-based alerts" as a product, they don't distinguish between the alert generator and alert receiver. Furthermore, they don't allow the configuration of an external alert receiver by its users - with reason (added product complexity and in most cases final users don't care about it).

In a cloud offering, you set up an alert and eventually, it gets delivered (more often than not the delivery is through the Alertmanager).

[...]

PR #2873, opened 2022-04-01
Introduce 'forward_alerts' option for routes

codesome:
This is a reference implementation for https://github.com/prometheus/alertmanager/issues/2868

cc @gotjosh @simonpasquier


config/config.go
```diff
@@ -742,6 +742,7 @@ type Route struct {
 	GroupWait      *model.Duration `yaml:"group_wait,omitempty" json:"group_wait,omitempty"`
 	GroupInterval  *model.Duration `yaml:"group_interval,omitempty" json:"group_interval,omitempty"`
 	RepeatInterval *model.Duration `yaml:"repeat_interval,omitempty" json:"repeat_interval,omitempty"`
+	ForwardAlerts  bool            `yaml:"forward_alerts,omitempty" json:"forward_alerts,omitempty"`
```
gotjosh, line 745: I don't think this should be a stand-alone option - I'd rather have semantics in place using the existing `GroupInterval` and `RepeatInterval`.

Let's finish the discussion on https://github.com/prometheus/alertmanager/issues/2868 before we discuss specifics.

PR #2878, opened 2022-04-05
Fix resolved header in email template

nileger:
# Background
We use a custom email template based on `alertmanager/template/email.tmpl`.

# Problem
The email template for resolved messages currently has a styling issue. Due to an extra `</td>`, the green heading is only shown in the top right corner, instead of covering the full width.

image

# Solution
Removing the extra `</td>` solves the issue.

image

### DIFF
--- a/template/email.tmpl
+++ b/template/email.tmpl
@@ -91,7 +91,7 @@ h4 {
             {{ if gt (len .Alerts.Firing) 0 }}
             <td class="alert alert-warning" style="margin: 0; font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif; box-sizing: border-box; vertical-align: top; font-size: 16px; color: #fff; font-weight: 500; padding: 20px; text-align: center; border-radius: 3px 3px 0 0; background-color: #E6522C;" valign="top" align="center" bgcolor="#E6522C">
             {{ else }}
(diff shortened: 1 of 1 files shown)


NilsLeger:
Can anyone help with the DCO check, please?

simonpasquier:
`template/email.tmpl` gets generated from `template/email.html` via the [template/inline-css.js](https://github.com/prometheus/alertmanager/blob/main/template/inline-css.js) script which uses `juice` under the hood to inline the CSS properies into the `style` attributes. 

IIUC `juice` is confused by this part and adds the extra `</td>` 
https://github.com/prometheus/alertmanager/blob/dcc92eda2ecbb96948c1e864338bfbf32ce8885b/template/email.html#L326-L334

One way to fix it would be to duplicate the range loop like this:

```
 {{ if gt (len .Alerts.Firing) 0 }} 
 <td class="alert alert-warning"> 
   {{ .Alerts | len }} alert{{ if gt (len .Alerts) 1 }}s{{ end }} for {{ range .GroupLabels.SortedPairs }} 
     {{ .Name }}={{ .Value }}  
   {{ end }} 
 </td>
 {{ else }} 
 <td class="alert alert-good"> 
   {{ .Alerts | len }} alert{{ if gt (len .Alerts) 1 }}s{{ end }} for {{ range .GroupLabels.SortedPairs }} 
     {{ .Name }}={{ .Value }}  
   {{ end }} 
 </td>
 {{ end }} 
```

cod-r:
Can this be merged? We are having the same problem and is annoying.

simonpasquier:
@cod-r this should be fixed in the latest v0.25.0 thanks to #3166

PR #2827, opened 2022-02-04
Adding telegram support 

timmilesdw:
Added telegram integration

Features:
- Custom API URL
- Disable Notifications
- Message templating

image

PR #2879, opened 2022-04-05
notify/telegram: Set API URL and ParseMode defaults

### LINKED ISSUE #2866: Issue with Telegram bot ID
I just updated to Alertmanager 0.24 and tried switching to the new native Telegram integration but stumbled over an issue.
My bot has a "-" character in it's ID (I do not know how common this is) which leads to the following error:

`alertmanager[10628]: ts=2022-03-28T14:28:45.218Z caller=notify.go:732 level=warn component=dispatcher receiver=telegram_native integration=telegram[0] msg="Notify attempt failed, will retry later" attempts=1 err="telegram: Bad Request: can't parse entities: Character '-' is reserved and must be escaped with the preceding '\\' (400)"`

When escaped, there's no error, but also no notifications being sent. My current Telegram integration (Sachet) is working, so the bot and IDs itself are correct.

Also, a minor thing: the `api_url` parameter seems to be optional, but not providing it leads to a validation error:
`Checking '/etc/alertmanager/alertmanager.yml'  FAILED: missing api_url on telegram_config`
I had to specify the URL in the receiver config.

```
receivers:
  - name: 'telegram_native'
    telegram_configs:
    - bot_token: 'token'
      chat_id: ID
      send_resolved: true
      api_url: "https://api.telegram.org"
```

### LINKED ISSUE #2876: telegram_config.api_url should be optional
I just updated to alertmanager-0.24.0 and went to use the new Telegram integration. Works great for the most part!

However, [the docs](https://prometheus.io/docs/alerting/latest/configuration/#telegram_config) suggest that setting the `api_url` is optional. However, when validating a config that doesn't explicitly set `api_url`, `amtool check-config alertmanager.yml` reports `FAILED: missing api_url on telegram_config`.

@timmilesdw @simonpasquier

* System information:

`Linux 3.10.108 armv7l`

* Alertmanager version:
```
alertmanager, version 0.24.0 (branch: HEAD, revision: f484b17fa3c583ed1b2c8bbcec20ba1db2aa5f11)
  build user:       root@45f97b18b623
  build date:       20220325-09:29:04
  go version:       go1.17.8
  platform:         linux/arm
```

metalmatze:
For Telegram, there is really just one API URL, and other than for testing I never saw anything changing it. Therefore, as asked in #2876 we can set the default.
The default Telegram template that's shipped with Alertmanager uses HTML and if send to Telegram without the parseMode set to `HTML` Telegram's API will return an error:
```
Bad Request: can't parse entities: Character '-' is reserved and must be escaped with the preceding '\\' (400)"
```
  
Closes #2866
Closes #2876

### DIFF
--- a/notify/telegram/telegram.go
+++ b/notify/telegram/telegram.go
@@ -16,6 +16,7 @@ package telegram
 import (
 	"context"
 	"net/http"
+	"net/url"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
@@ -42,6 +43,19 @@ func New(conf *config.TelegramConfig, t *template.Template, l log.Logger, httpOp
 	httpclient, err := commoncfg.NewClientFromConfig(*conf.HTTPConfig, "telegram", httpOpts...)
 	if err != nil {
 		return nil, err
+	}
+
+	if conf.APIUrl == nil {
+		apiURL, err := url.Parse("https://api.telegram.org")
+		if err != nil {
+			return nil, err
+		}
+		conf.APIUrl = &config.URL{URL: apiURL}
+	}
+	if conf.ParseMode == "" {
+		// The default Telegram template is in HTML,
+		// so we need to set the parse mode to HTML for Telegram not to return a parse error.
+		conf.ParseMode = "HTML"
 	}
 
 	client, err := createTelegramClient(conf.BotToken, conf.APIUrl.String(), conf.ParseMode, httpclient)
--- a/notify/telegram/telegram.go
+++ b/notify/telegram/telegram.go
@@ -16,7 +16,6 @@ package telegram
 import (
 	"context"
 	"net/http"
-	"net/url"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
@@ -43,19 +42,6 @@ func New(conf *config.TelegramConfig, t *template.Template, l log.Logger, httpOp
 	httpclient, err := commoncfg.NewClientFromConfig(*conf.HTTPConfig, "telegram", httpOpts...)
 	if err != nil {
 		return nil, err
-	}
-
-	if conf.APIUrl == nil {
-		apiURL, err := url.Parse("https://api.telegram.org")
-		if err != nil {
-			return nil, err
-		}
-		conf.APIUrl = &config.URL{URL: apiURL}
-	}
-	if conf.ParseMode == "" {
-		// The default Telegram template is in HTML,
-		// so we need to set the parse mode to HTML for Telegram not to return a parse error.
-		conf.ParseMode = "HTML"
 	}
 
 	client, err := createTelegramClient(conf.BotToken, conf.APIUrl.String(), conf.ParseMode, httpclient)
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -143,7 +143,7 @@ var (
 		},
 		DisableNotifications: false,
 		Message:              `{{ template "telegram.default.message" . }}`,
-		ParseMode:            "MarkdownV2",
+		ParseMode:            "HTML",
 	}
 )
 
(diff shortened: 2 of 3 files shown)


notify/telegram/telegram.go
```diff
@@ -44,6 +45,19 @@ func New(conf *config.TelegramConfig, t *template.Template, l log.Logger, httpOp
 		return nil, err
 	}
 
+	if conf.APIUrl == nil {
+		apiURL, err := url.Parse("https://api.telegram.org")
```
gotjosh, line 49: nit: Can this be a `const`?

gotjosh (approved):
LGTM

Can you please fix the title of the PR so that it specifies that what we're fixing is the URL?

From `notify/telegram: Set API and ParseMode defaults` to `Telegram: Set API URL and ParseMode defaults`

notify/telegram/telegram.go
roidelapluie: This should be done during unmarhalling of the configuration, not when calling New.

notify/telegram/telegram.go
```diff
@@ -44,6 +45,19 @@ func New(conf *config.TelegramConfig, t *template.Template, l log.Logger, httpOp
...
+		if err != nil {
+			return nil, err
+		}
+		conf.APIUrl = &config.URL{URL: apiURL}
+	}
+	if conf.ParseMode == "" {
+		// The default Telegram template is in HTML,
+		// so we need to set the parse mode to HTML for Telegram not to return a parse error.
+		conf.ParseMode = "HTML"
```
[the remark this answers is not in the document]
roidelapluie: This should be done during unmarhalling of the configuration, not when calling New.

notify/telegram/telegram.go
gotjosh: Good shout Julien.

On a second look at https://github.com/prometheus/alertmanager/blob/71d61c9c9149420209c973014e0d6c981b183611/config/notifiers.go#L140-L147

It seems like both of these belong within `DefaultTelegramConfig`

notify/telegram/telegram.go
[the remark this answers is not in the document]
gotjosh: Our current default is actually `MarkdownV2` this would effectively be a breaking change and I don't think it was the intent of #2827.

Looking at the current template - I don't see how it is HTML, am I missing something? https://github.com/prometheus/alertmanager/blob/main/template/default.tmpl#L105-L114

gotjosh (changes requested):
On a second look, it seems like some parts of these are not correct.

I suggest we split this into two PRs.

1. The default URL is correct and seems straightforward - we need to add it to the docs though.

2. The default parse mode is currently Markdown, and setting the default to HTML would be a breaking change based on what's currently here.

roidelapluie:
Actually the API URl should come from DefaultGlobalConfig. i do not know why it is not the case.

notify/telegram/telegram.go
roidelapluie: nit really because the API should come from DefaultGlobalConfig

metalmatze:
I've looked into the parseMode issue again. It seems that Telegram doesn't allow the `-` in Markdown messages.
This character is used many times in messages.

Screenshot from 2022-06-03 11-46-20

Most notably the `-` is used to list all labels of an alert:
https://github.com/prometheus/alertmanager/blob/f958b8be84b870e363f7dafcbeb807b463269a75/template/default.tmpl#L8-L10

Unless we want to change the default template for all receivers, we should probably make `HTML` the default parseMode...

metalmatze:
Not sure why the tests fail. I can try rebasing later.

roidelapluie (approved):

gotjosh:
Once you rebase this should be fixed - it relates to #2899

PR #2880, opened 2022-04-06
UI: add 'Link' button to alerts in list

ccmtaylor:
adds a button to each alert that allows linking to a single alert.
This works by generating a link that filters on _all_ of the alert's
label values, similar to the button to create a new silence from an
alert.


gotjosh:
Visually, the changes look good and seem like a decent improvement to link individual alerts. However, I'm in no capacity to review the FE changes - perhaps @w0rm you can take a look?

Here are a couple of screenshots to confirm what it looks like:

### Before clicking on "Link"
Screenshot 2022-05-03 at 09 48 13

### After clicking on "Link"
Screenshot 2022-05-03 at 09 48 20

ccmtaylor:
thanks for taking a look @gotjosh! Originally, I wanted this to copy the resulting URL to the clipboard, but I wasn't able to figure out how to access the clipboard (I'm not experienced with Elm). This change at least makes it easy to copy the link from the browser's address bar.

ccmtaylor:
is there any interest in this? If not, I'd close the PR in a week or so

w0rm:
@ccmtaylor apologies! LGTM

PR #2897, opened 2022-04-29
Remove unneeded stopping of nil Dispatcher

juliusv:
The function value and parameters of a defer statement are immediately
evaluated, so this "disp" value is always nil, and calling Stop() on a nil
dispatcher is a no-op, so this does nothing, but is confusing.

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -347,7 +347,9 @@ func run() int {
 	defer alerts.Close()
 
 	var disp *dispatch.Dispatcher
-	defer disp.Stop()
+	defer func() {
+		disp.Stop()
+	}()
 
 	groupFn := func(routeFilter func(*dispatch.Route) bool, alertFilter func(*types.Alert, time.Time) bool) (dispatch.AlertGroups, map[model.Fingerprint][]string) {
 		return disp.Groups(routeFilter, alertFilter)


gotjosh (approved):
LGTM.

Took a while to review this one, I wasn't sure I understood why didn't we need to move this one but after a few rounds of testing and a re-read, this is safe to remove.

I think that we need to add is a way to explicitly stop all components similarly, as an example both silence and nflog maintenance are governed by the `stopc` but `mem.NewAlerts` nor `Dispatcher` aren't - would good to make sure _all_ components are cleanly shutting down as we exit.

Let me know if you'd like to create a follow issue+PR to make sure we don't forget as I think this can be separate from this PR.

juliusv:
This PR just resulted from brief tourism into the AM codebase for the purpose of refreshing my memory about it :) I filed an issue (https://github.com/prometheus/alertmanager/issues/2906), but can't promise I'll get to it myself.

Now actually looking at this PR here again, I noticed that indeed, the Dispatcher is also not shut down any other way when main.go exits (only upon config reload), so the right short-term fix is probably not to remove this ineffectual line, but to fix it by wrapping the `disp.Stop()` in a deferred closure. Look again :)

gotjosh:
> but can't promise I'll get to it myself.

I'm happy to take it for a spin.

> disp.Stop() in a deferred closure

That also works 😄

gotjosh (approved):
LGTM

roidelapluie:
Thanks!

PR #2901, opened 2022-04-29
Sort receiver stage constructions by pipeline order

juliusv:
This is less confusing to read if the receiver stages are constructed in the
same order as they are used in the pipeline below.


gotjosh (approved):
LGTM

PR #2908, opened 2022-05-04
Add support for custom Go template delimiters

siavashs:
Alertmanager uses Go templates in different configuration parameters.
Many users generate the configuration using tools like Helm, ExternalSecret, etc.
Most of these tools are written in Go, and therfore use Go templates.
This creates problems when the configuration tool needs to skip the Alertmanager templates.
As a result users usually have to add escaping to the templates:
```go
{{ `{{ ... }}` }}
```

This change allows setting the template delimiters used in Alertmanager configuration.
This is already configurable in Go templates using `Delims()` functions for both `text` and `html`.
The newly added global configuration parameters allows users to set the left and right delimiters
used in Alertmanager templates and therfore avoid conflict with other configuration management tools:

```yaml
template_config:
  delim_left: "[["
  delim_right: "]]"
```

This change is backward compatible as empty delimiter parameters stands for the corresponding default: `{{` or `}}`.
The builtin templates are not affected by this change as custom delimiters are only applied to:
- `<tmpl_string>`
- `<tmpl_secret>`
- custom templates loaded by `templates: []`

### DIFF
--- a/cli/check_config.go
+++ b/cli/check_config.go
@@ -82,9 +82,14 @@ func CheckConfig(args []string) error {
 			}
 			fmt.Printf(" - %d inhibit rules\n", len(cfg.InhibitRules))
 			fmt.Printf(" - %d receivers\n", len(cfg.Receivers))
-			fmt.Printf(" - %d templates\n", len(cfg.Templates))
-			if len(cfg.Templates) > 0 {
-				_, err = template.FromGlobs(cfg.Templates...)
+			fmt.Println(" - builtin templates")
+			tmpl, err := template.New()
+			if err != nil {
+				fmt.Printf("  FAILED: %s\n", err)
+				failed++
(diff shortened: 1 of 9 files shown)


siavashs:
I was planning a similar change for Prometheus but usually we don't use any secrets there and most of the scrape config is generated by the operator and `ServiceMonitor`s, etc.

Also there is a similar change in this PR for Helm https://github.com/helm/helm/pull/10299

siavashs:
> I am concerned because this will break all built-in templates.

Ouch, I missed the internal templates.
I'll try to find a solution for that. If you have any suggestions please let me know.

siavashs:
I just did a test and we can keep using `{{ }}` for default templates and custom delimiters for other templates https://go.dev/play/p/l_wElt_dYfw

I'll update the PR accordingly to make sure we only apply the configured delimiters to non-default templates.

Issue #2899, opened 2022-04-29
Mixin tests failing


juliusv:
On recent PRs, the mixin tests are failing with:

```
#!/bin/bash -eo pipefail
cd doc/alertmanager-mixin; make lint
jb install
GET https://github.com/grafana/grafonnet-lib/archive/55cf4ee53ced2b6d3ce96ecce9fb813b4465be98.tar.gz 200
mixtool generate alerts mixin.libsonnet > alertmanager_alerts.yaml
find . -name 'vendor' -prune -o -name '*.libsonnet' -print -o -name '*.jsonnet' -print | \
	while read f; do \
		jsonnetfmt -n 2 --max-blank-lines 1 --string-style s --comment-style s "$f" | diff -u "$f" -; \
	done
mixtool lint mixin.libsonnet
[target-instance-rule] 'Alertmanager / Overview': Dashboard 'Alertmanager / Overview', panel 'Alerts', target idx '0' invalid PromQL query 'sum(alertmanager_alerts{job=~"$job"}) by (job,instance)': instance selector not found
[target-instance-rule] 'Alertmanager / Overview': Dashboard 'Alertmanager / Overview', panel 'Alerts receive rate', target idx '0' invalid PromQL query 'sum(rate(alertmanager_alerts_received_total{job=~"$job"}[$__rate_interval])) by (job,instance)': instance selector not found
[target-instance-rule] 'Alertmanager / Overview': Dashboard 'Alertmanager / Overview', panel 'Alerts receive rate', target idx '1' invalid PromQL query 'sum(rate(alertmanager_alerts_invalid_total{job=~"$job"}[$__rate_interval])) by (job,instance)': instance selector not found
[target-instance-rule] 'Alertmanager / Overview': Dashboard 'Alertmanager / Overview', panel '$integration: Notifications Send Rate', target idx '0' invalid PromQL query 'sum(rate(alertmanager_notifications_total{job=~"$job", integration="$integration"}[$__rate_interval])) by (integration,job,instance)': instance selector not found
[target-instance-rule] 'Alertmanager / Overview': Dashboard 'Alertmanager / Overview', panel '$integration: Notifications Send Rate', target idx '1' invalid PromQL query 'sum(rate(alertmanager_notifications_failed_total{job=~"$job", integration="$integration"}[$__rate_interval])) by (integration,job,instance)': instance selector not found
[...]

PR #2910, opened 2022-05-04
config: fix validation of OpsGenie configuration

simonpasquier:
The validation should fail if both `api_key` and `api_key_file` are
defined. I think there was a typo in the original PR (#2728) that
enforced `api_url` and `api_key_file` not being defined at the same
time.

cc @jkroepke since you did the initial implementation.

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -492,7 +492,7 @@ func (c *OpsGenieConfig) UnmarshalYAML(unmarshal func(interface{}) error) error 
 		return err
 	}
 
-	if c.APIURL != nil && len(c.APIKeyFile) > 0 {
+	if c.APIKey != "" && len(c.APIKeyFile) > 0 {
 		return fmt.Errorf("at most one of api_key & api_key_file must be configured")
 	}
 
--- a/config/notifiers_test.go
+++ b/config/notifiers_test.go
@@ -590,6 +590,72 @@ func TestOpsgenieTypeMatcher(t *testing.T) {
 	}
 }
 
+func TestOpsGenieConfiguration(t *testing.T) {
+	for _, tc := range []struct {
+		name string
+		in   string
+
+		err bool
+	}{
+		{
+			name: "valid configuration",
+			in: `api_key: xyz
+responders:
+- id: foo
+  type: scheDule
+- name: bar
+  type: teams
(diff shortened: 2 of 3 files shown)


gotjosh:
Can you rebase from `main`? I think this the failure relates to #2899 which was fixed a few days ago.

simonpasquier:
Thanks! I wanted the change against `release-0.24` to cut v0.24.1 right after since `api_key_file` was added in v0.24.0 :)
I've cherry-picked 35bf59f1829ed998aaf577d56681cf24039fd6a8 instead...

gotjosh:
Silly me! I didn't notice that the target branch was different.

config/notifiers.go
```diff
@@ -492,7 +492,7 @@ func (c *OpsGenieConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
 		return err
 	}
 
-	if c.APIURL != nil && len(c.APIKeyFile) > 0 {
+	if c.APIKey != "" && len(c.APIKeyFile) > 0 {
```
gotjosh, line 495:
```suggestion
        if c.APIURL == nil {
          return fmt.Errorf("missing URL in config")
        }
        
	if c.APIKey != "" && len(c.APIKeyFile) > 0 {
```

gotjosh: I think we need to validate the URL as well, `DefaultOpsGenieConfig` does not contain a default URL.

config/notifiers_test.go
```diff
@@ -590,6 +590,72 @@ func TestOpsgenieTypeMatcher(t *testing.T) {
 	}
 }
 
+func TestOpsGenieConfiguration(t *testing.T) {
```
gotjosh, line 593: Should we also include a test here if we decide to include the URL validation?

config/notifiers.go
simonpasquier: there's a default OpsGenie URL in the global config.
https://github.com/prometheus/alertmanager/blob/23f961ec6a9c8a15bf99c83815d35e13d66a4d5d/config/config.go#L591

gotjosh: 🤦

gotjosh (approved):
LGTM

PR #2919, opened 2022-05-11
adding support for accessing template.Data in prometheus rules/alerts

jrkt:
Right now, the only data that is usable in alert notification templates is the [following struct](https://github.com/prometheus/prometheus/blob/c40e269c3e514953299e9ba1f6265e067ab43e64/template/template.go#L348):
```go
struct {
	Labels         map[string]string
	ExternalLabels map[string]string
	Value          float64
}
```

It would be useful to have the ability to use the [template.Data](https://github.com/prometheus/alertmanager/blob/23f961ec6a9c8a15bf99c83815d35e13d66a4d5d/template/template.go#L236) struct in these alert notification templates.

One use case for this is being able to time-box a URL to the applications logs. We are currently adding an annotation that links to GCP logs for the application that the alert is for. The problem with this is, there's no way to drill down to the specific time the alert fired. So, unless that URL is clicked on right away, the logs that show are not at the time of the alert. 

What I was trying to do initially was something like this:
```yaml
groups:
- name: TestGroup
  rules:
  - alert: TestAlert
    expr: test-expr
    annotations:
      logs_url: https://www.our-logs-domain?app-name=test-application&startTime={{ (index .Alerts 0).StartsAt.Format "2006-01-02T15:04:05.999999999Z07:00" }}
```

This obviously doesn't work because the only fields available are `Labels`, `ExternalLabels`, and `Value`.

I'm proposing introducing a new annotation, `extra_delims`, that tells the template to evaluate the text one additional time with the delimiters provided in the annotation.

This would allow something like this in the prometheus alert definitions:
```yaml
groups:
- name: TestGroup
  rules:
  - alert: TestAlert
    expr: test-expr
    annotations:
      extra_delims: "<<,>>"
      logs_url: https://www.our-logs-domain?app-name=test-application&startTime=<< (index .Alerts 0).StartsAt.Format "2006-01-02T15:04:05.999999999Z07:00" >>
```

The other thing that this fixes is nested templates. For example, here is an alert manager config that injects data to a specific receiver that comes from the annotations of the prometheus rule/alert:
```yaml
receivers:
- name: opsgenie
  opsgenie_configs:
  - api_key: XXX
    details:
      logs_url: '{{ .CommonAnnotations.logs_url }}'
    ...
```

Once that is evaluated, it is at it's final state. So, if my annotation `logs_url` has data that I want to be evaluated by a template, it will never be evaluated. Currently, this evaluates the `logs_url` to the final value of: 
```
https://www.our-logs-domain?app-name=test-application&startTime=<< (index .Alerts 0).StartsAt.Format "2006-01-02T15:04:05.999999999Z07:00" >>
```

But, with the `extra_delims` annotation, this would be evaluated one additional time resulting in something like this:
```
https://www.our-logs-domain?app-name=test-application&startTime=2022-05-12T22:47:04.535890220Z
```

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -20,6 +20,9 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
+
 	commoncfg "github.com/prometheus/common/config"
 	"github.com/prometheus/common/sigv4"
 )
@@ -189,7 +192,7 @@ func (c *EmailConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	// Header names are case-insensitive, check for collisions.
 	normalizedHeaders := map[string]string{}
 	for h, v := range c.Headers {
-		normalized := strings.Title(h)
+		normalized := cases.Title(language.AmericanEnglish).String(h)
 		if _, ok := normalizedHeaders[normalized]; ok {
 			return fmt.Errorf("duplicate header %q in email config", normalized)
 		}
--- a/go.mod
+++ b/go.mod
@@ -36,6 +36,7 @@ require (
 	go.uber.org/atomic v1.9.0
 	golang.org/x/mod v0.5.1
 	golang.org/x/net v0.0.0-20220225172249-27dd8689420f
+	golang.org/x/text v0.3.7
 	golang.org/x/tools v0.1.9
 	gopkg.in/alecthomas/kingpin.v2 v2.2.6
 	gopkg.in/telebot.v3 v3.0.0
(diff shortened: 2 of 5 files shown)


jrkt:
@simonpasquier or @w0rm could I get a review on this?

Nexucis:
Maybe @gotjosh will have time to review it

Nexucis:
cool, thanks @gotjosh for your input and review !! Hope it will help you @jrkt

gotjosh:
If I understood correctly, your problem is about using the time of alert as part of the annotation as described here:

> One use case for this is being able to time-box a URL to the applications logs. We are currently adding an annotation that links to GCP logs for the application that the alert is for. The problem with this is, there's no way to drill down to the specific time the alert fired. So, unless that URL is clicked on right away, the logs that show are not at the time of the alert.

Am I missing something?

jrkt:
Correct, so we have a PrometheusRule that defines an alert that looks something like this:
```yaml
    - alert: devx-reference-alerts-TESTALERT
      annotations:
        description: |
          This is a test alert
        summary: This is a test alert
      expr: my-expr
      for: 1m
      labels:
        service_metrics_dashboard: https://mylogs.com?timestamp={{ WHAT-GOES-HERE }}
```

I would like to timebox the label that gets propagated to the alert. If there's a builtin way to do that in this alert definition that's awesome. I'll just do that. I just could not figure out a way to make that work.

jrkt:
That is EXACTLY what I needed. Thank you :pray: One thing though.. where does the `humanizeTimestamp` function come from? Is there a list of what functions are provided?

jrkt:
Nevermind, found it. Thank you for your help! I'll go ahead and close this.

jrkt:
@gotjosh one question.. how can I convert the timestamp that looks like this `1.656534216234e%2009` to `1656534216234009`

roidelapluie:
`printf "%f"`

jrkt:
Something is off when piping the value into `printf "%f"`. The value I get back is 3 decimal places shy of being valid. For example, `1.657043646234e+09` gets converted to `1657043646.234000`. But, to be valid it needs to be `1657043646234.000`.

PR #2922, opened 2022-05-13
Fix Markdown formatting bugs / typos in the config docs

juliusv:
I noticed that some of the formatting on
https://prometheus.io/docs/alerting/latest/configuration/#time_interval-0 is
broken (including some words being formatted as italic due to underscores). This should fix that.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -291,9 +291,9 @@ to match a complete time interval, all fields must match.
 Some fields support ranges and negative indices, and are detailed below. All definitions are
 taken to be in UTC, no other timezones are currently supported.
 
-`time_range` Ranges inclusive of the starting time and exclusive of the end time to
+`time_range` ranges are inclusive of the starting time and exclusive of the end time to
 make it easy to represent times that start/end on hour boundaries.
-For example, start_time: '17:00' and end_time: '24:00' will begin at 17:00 and finish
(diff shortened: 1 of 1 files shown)


juliusv:
@benridley

gotjosh (approved):
LGTM

docs/configuration.md
```diff
@@ -291,9 +291,9 @@ to match a complete time interval, all fields must match.
 Some fields support ranges and negative indices, and are detailed below. All definitions are
 taken to be in UTC, no other timezones are currently supported.
 
-`time_range` Ranges inclusive of the starting time and exclusive of the end time to
+`time_range` ranges are inclusive of the starting time and exclusive of the end time to
```
benridley, line 294: I think maybe this just needs a colon after the ‘time_range’ part? The other descriptions uses the section name as a heading, rather than using them as part of the description.

benridley:
Thanks for fixing these @juliusv, I left a comment about one section but the rest look fine to me.

docs/configuration.md
juliusv: Ahhh, that's how that was meant, yeah. Pushed a change to just add the semicolon here.

benridley:
Thanks @juliusv, LGTM

PR #2929, opened 2022-05-23
Trim contents of slack api urls from files

srhb:
Trailing newlines will almost always be present in a posixish file, so trim it away at first opportunity to prevent having invalid URLs we won't discover until a notification is sent.

### DIFF
--- a/notify/slack/slack.go
+++ b/notify/slack/slack.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strings"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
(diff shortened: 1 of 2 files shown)


gotjosh (approved):
LGTM

gotjosh:
Can you please fix the DCO and create a test case for this? Thanks!

srhb:
Sure, here you go :smile_cat:

simonpasquier:
you'd need to fix the test

```
notify/slack/slack_test.go:88:12: undeclared name: `ioutil` (typecheck)
	f, err := ioutil.TempFile("", "slack_test_newline")
	          ^
make: *** [Makefile.common:198: common-lint] Error 1

Exited with code exit status 2
```

simonpasquier:
thanks!

PR #2932, opened 2022-05-25
notify/webhook: allow configuring HTTP status codes to retry on

jvrplmlmn:
Introduces a new webhook config `retry_codes` which is a list of
additional HTTP codes that the Webhook notifier should retry on.

A specific use case: This will allow users to honor HTTP 429 (rate limiting),
which is currently not supported.

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -399,6 +399,9 @@ type WebhookConfig struct {
 	// Alerts exceeding this threshold will be truncated. Setting this to 0
 	// allows an unlimited number of alerts.
 	MaxAlerts uint64 `yaml:"max_alerts" json:"max_alerts"`
+
+	// Additional HTTP status codes that should be retried.
+	RetryCodes []int `yaml:"retry_codes,omitempty" json:"retry_codes,omitempty"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
--- a/notify/webhook/webhook.go
+++ b/notify/webhook/webhook.go
@@ -56,6 +56,7 @@ func New(conf *config.WebhookConfig, t *template.Template, l log.Logger, httpOpt
 			CustomDetailsFunc: func(int, io.Reader) string {
 				return conf.URL.String()
 			},
+			RetryCodes: conf.RetryCodes,
 		},
 	}, nil
 }
--- a/notify/webhook/webhook_test.go
+++ b/notify/webhook/webhook_test.go
@@ -15,6 +15,7 @@ package webhook
 
 import (
 	"fmt"
+	"net/http"
 	"net/url"
 	"testing"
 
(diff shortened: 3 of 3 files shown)


jvrplmlmn:
Hey folks, seems like CI is failing on an acceptance test: `TestResolvedFilter`

I am not entirely familiar with the test process for Alertmanager, for completeness this is what I have been running for both `main` and this branch, and in both cases the test suite fails for me:
- `make build-all`
- `make common-test`

Could I get a hint on how I could troubleshoot if that is related to this changes? Much appreciated!

iainlane:
Just to say quickly: 429 errors can come with a [`Retry-After header`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) telling you how long to wait before retrying. Perhaps we could extend the retrying logic to look for this? I see it's using an [exponential ticker](https://github.com/prometheus/alertmanager/blob/f958b8be84b870e363f7dafcbeb807b463269a75/notify/notify.go#L694-L697) now though, so might be slightly hairy to implement - if so it could legitimately be delayed to a followup perhaps.

The rationale is that it would be nice if you're being rate limited to back off for the amount of time the server says to, more chance of not being limited the next time.

roidelapluie:
Yes. However, they are cloud services.

For alertmanager webhooks, we expect them to be written specifically for alertmanager, therefore never return 429's.

gotjosh:
> Due to the clustering nature of alertmanager,

Out of curiosity, what am I missing from this rationale? I can't picture how the clustering logic creates a problem to retry. However, I can see how this conflicts in non-intuitive ways with things like the `repeat_interval`.

With clustering, under partitions, the price to pay will always be duplicate notifications. If the downstream system is under pressure that's up to the operator's configuration and you have all the knobs you need to tweak and adjust as necessary.  

I want to hear's @roidelapluie opinion first to make sure I understand his angle, but in my eyes, this kind of change makes sense (mostly [because](https://github.com/prometheus/alertmanager/blob/main/notify/opsgenie/opsgenie.go#L57) [there's precedent with other integrations](https://github.com/prometheus/alertmanager/blob/main/notify/pagerduty/pagerduty.go#L274)) - My understanding is that the Alertmanager philosophy is to allow custom integrations via Webhooks. 

By including this functionality, we're allowing webhook integrations to have similar features to other receivers we support. I see this is a net positive, as there'd be less of a need to include proprietary receivers in the Alertmanager and thus less maintenance overhead.

gotjosh:
After taking a closer look, I do see a small problem with this - and is the fact that we'll try indefinitely (see [here](https://github.com/prometheus/alertmanager/blob/f958b8be84b870e363f7dafcbeb807b463269a75/notify/notify.go#L694-L697)) so for the implementation side, I think this might need to be paired (for the Webhooks at least) with a maximum elapsed time that we're willing to retry the notification for. 

But as I said, I'll wait to hear @roidelapluie opinion's before we move there.

gotjosh:
@jvrplmlmn I've discussed this with @roidelapluie offline and have agreed that we can move forward. 

Can you please extend this Pull Request with a maximum try time for webhooks? I don't think we need to make it configurable from the get-go and for now can be just a sensible default.

jvrplmlmn:
Unfortunately, this has gone stale and I don't have bandwidth to work on it. So I am closing the PR.

PR #2944, opened 2022-06-15
Mixin: `template-job-rule` now only validates `job` and not `instance` and `job` labels

gotjosh:
With https://github.com/grafana/dashboard-linter/pull/49 `template-job-rule` no longer validates both `instance` and `job` labels. Add the new rule of `template-instance-rule` to the exclusions to preserve the previous behaviour.

This should fix the CI for the mixin.


gotjosh:
@simonpasquier @roidelapluie would appreciate a pair of 👀 so that we can fix the CI? 🙏

roidelapluie (approved):

PR #2947, opened 2022-06-17
Update configuration doc to reflect tls_config in oauth2

majolo:
Updated for this change: https://github.com/prometheus/common/pull/331

Bringing in line with Prometheus docs: https://prometheus.io/docs/prometheus/latest/configuration/configuration/#oauth2


roidelapluie:
Signed off needs to me made by git so it is correct

majolo:
Yes since this was a simple change I was doing it via the Github UI rather than cloning the repo locally, will try and get round to it later.

PR #2949, opened 2022-06-17
Force close mock servers as part of v2 acceptance tests

gotjosh:
While merging #2944, I noticed the CI failed: https://app.circleci.com/pipelines/github/prometheus/alertmanager/2686/workflows/b6f87b0a-20c3-455b-b706-432c38a77511/jobs/12028.

It seemed like a deadlock between uncoordinated routines but I couldn't pin point (or reproduce, I tried with -race and -count) the exact problem. However, from the logs, I could point out where the problem originated and kind of have a hunch it had to do with the way net listeners are handled by the TODO removed.

The more worrying bit of the CI failure is that it took 10m to timeout, with this change we'll force close the connection with a 5s deadline so at the very least we'll get the feedback faster.

Along the way, I fixed a couple of instances of variables shadowing package names which should be pretty harmless. All the functionality of the tests should remain intact.

### DIFF
--- a/test/with_api_v2/acceptance/api_test.go
+++ b/test/with_api_v2/acceptance/api_test.go
@@ -58,7 +58,7 @@ receivers:
 		Tolerance: 1 * time.Second,
 	})
 	co := at.Collector("webhook")
-	wh := a.NewWebhook(co)
+	wh := a.NewWebhook(t, co)
 
 	amc := at.AlertmanagerCluster(fmt.Sprintf(conf, wh.Address()), 1)
 	require.NoError(t, amc.Start())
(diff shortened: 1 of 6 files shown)


test/with_api_v2/acceptance/api_test.go
```diff
@@ -89,8 +89,8 @@ receivers:
 	require.NoError(t, err)
 	// No silence has been created or inhibiting alert sent, alert should
 	// be active.
-	for _, alert := range resp.Payload {
```
gotjosh, line 92: There's a couple more like this but these it's just fixing package names matching variables.

test/with_api_v2/mock.go
```diff
@@ -256,32 +257,28 @@ func equalTime(a, b time.Time, opts *AcceptanceOpts) bool {
...
 	// Func is called early on when retrieving a notification by an
 	// Alertmanager. If Func returns true, the given notification is dropped.
 	// See sample usage in `send_test.go/TestRetry()`.
 	Func func(timestamp float64) bool
 }
 
-func NewWebhook(c *Collector) *MockWebhook {
-	l, err := net.Listen("tcp4", "localhost:0")
-	if err != nil {
-		// TODO(fabxc): if shutdown of mock destinations ever becomes a concern
```
gotjosh, line 270: `t.Cleanup` was introduced for this exact same purpose hence the TODO deletion.

gotjosh:
@simonpasquier @roidelapluie WDYT?

roidelapluie (approved):

Issue #2610, opened 2021-06-04
Inconsistent behaviour between create and preview alert in UI


thyzzv:
**What did you do?**
Tried to create a catch all silence using `job =~ .*`. Via the Alertmanager UI

When I  do 'Preview alerts', al is good.
However when I try to 'create' the silence it fails with a '400 bad request'.
In the network console of the browser I see: `silence invalid: at least one matcher must not match the empty string`

**What did you expect to see?**
That either the 'preview' failed.
Or
The create succeeded. 
Either way is fine. But they should be consistent

**Environment**

* System information:

* Alertmanager version:
   Version: 0.22.2

* Prometheus version:
   Version: 2.27.1

* Logs:
```
GET | https://alertmanager.xxxx.nl/api/v2/alerts?filter=job=~".*"&amp;silenced=true&amp;inhibited=true&amp;active=true
returns status 200 with 'alerts'
vs
POST | https://alertmanager.xxxx.nl/api/v2/silences
{"matchers":[{"name":"job","value":".*","isRegex":true}],"startsAt":"2021-06-04T06:35:18.694Z","endsAt":"2021-06-04T08:35:18.694Z","createdBy":"Thyzzv","comment":"test","id":null}
```
In alertmanager log I also see:
```
level=error ts=2021-06-04T06:22:26.193Z caller=api.go:753 component=api version=v2 path=/silences method=POST msg="Failed to create silence" err="silence invalid: at least one matcher must not match the empty string"
```

wbh1:
I just hit this, too. Silly enough, you can create a silence for `job=~.+` since that requires 1 or more characters, rather than 0 or more.

thyzzv:
#1897 Sounds reasonable. I thought as much after hitting the limitation. That's why I described the behavior as inconsistent.

TheMeier:
This issue was not updated for quite a while. Reading through the comments I got the impression that this is sufficiently addressed therefore I am closing this issue. If I am wrong in my assessment please re-open this issue, possibly with additional information.

PR #2968, opened 2022-06-28
Ensure matcher values are present when parsing matchers from strings

### LINKED ISSUE #2936: Panic when using amtool
```
$ ./amtool --alertmanager.url=http://127.0.0.1:9093 alert add foo=
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
github.com/prometheus/alertmanager/pkg/labels.ParseMatcher({0x7fffcacb016f, 0x4})
        /home/roidelapluie/dev/alertmanager/pkg/labels/parse.go:130 +0x67f
github.com/prometheus/alertmanager/cli.parseLabels({0xc000569c38, 0x1, 0xc00007c570?})
        /home/roidelapluie/dev/alertmanager/cli/utils.go:110 +0x7b
github.com/prometheus/alertmanager/cli.(*alertAddCmd).addAlert(0xc00005d080, {0x103c0b0?, 0xc00080c0c0}, 0x6fc23ac00?)
        /home/roidelapluie/dev/alertmanager/cli/alert_add.go:76 +0x97
github.com/prometheus/alertmanager/cli.execWithTimeout.func1(0x40b4d4?)
        /home/roidelapluie/dev/alertmanager/cli/utils.go:154 +0x70
gopkg.in/alecthomas/kingpin%2ev2.(*actionMixin).applyActions(0xc000286150?, 0xcfd2c0?)
        /home/roidelapluie/go/pkg/mod/gopkg.in/alecthomas/kingpin.v2@v2.2.6/actions.go:28 +0x5c
gopkg.in/alecthomas/kingpin%2ev2.(*Application).applyActions(0xc0000de870?, 0xc0002ce090)
        /home/roidelapluie/go/pkg/mod/gopkg.in/alecthomas/kingpin.v2@v2.2.6/app.go:557 +0xed
gopkg.in/alecthomas/kingpin%2ev2.(*Application).execute(0xc0000de870, 0xc0002ce090?, {0xc000598040, 0x2, 0x2})
        /home/roidelapluie/go/pkg/mod/gopkg.in/alecthomas/kingpin.v2@v2.2.6/app.go:390 +0x66
gopkg.in/alecthomas/kingpin%2ev2.(*Application).Parse(0xc0000de870, {0xc0000300b0?, 0xc0000300b0?, 0x4?})
        /home/roidelapluie/go/pkg/mod/gopkg.in/alecthomas/kingpin.v2@v2.2.6/app.go:222 +0x14f
        /home/roidelapluie/dev/alertmanager/cli/root.go:162 +0xd45
main.main()
        /home/roidelapluie/dev/alertmanager/cmd/amtool/main.go:19 +0x17

```

When building `main`

gotjosh:
Fixes #2936

### DIFF
--- a/pkg/labels/parse.go
+++ b/pkg/labels/parse.go
@@ -120,8 +120,12 @@ func ParseMatcher(s string) (_ *Matcher, err error) {
 		return nil, errors.Errorf("bad matcher format: %s", s)
 	}
 
+	rawValue := ms[3]
+	if len(rawValue) == 0 {
+		return nil, errors.Errorf("matcher value is not present: %s", s)
+	}
+
 	var (
-		rawValue            = ms[3]
 		value               strings.Builder
 		escaped             bool
 		expectTrailingQuote bool
--- a/pkg/labels/parse.go
+++ b/pkg/labels/parse.go
@@ -120,18 +120,14 @@ func ParseMatcher(s string) (_ *Matcher, err error) {
 		return nil, errors.Errorf("bad matcher format: %s", s)
 	}
 
-	rawValue := ms[3]
-	if len(rawValue) == 0 {
-		return nil, errors.Errorf("matcher value is not present: %s", s)
-	}
-
 	var (
+		rawValue            = ms[3]
 		value               strings.Builder
 		escaped             bool
 		expectTrailingQuote bool
 	)
 
-	if rawValue[0] == '"' {
+	if strings.HasPrefix(rawValue, "\"") {
 		rawValue = strings.TrimPrefix(rawValue, "\"")
 		expectTrailingQuote = true
 	}
--- a/pkg/labels/parse_test.go
+++ b/pkg/labels/parse_test.go
@@ -191,7 +191,10 @@ func TestMatchers(t *testing.T) {
 		},
 		{
 			input: `job=`,
-			err:   `matcher value is not present: job=`,
+			want: func() []*Matcher {
+				m, _ := NewMatcher(MatchEqual, "job", "")
+				return []*Matcher{m}
+			}(),
 		},
 		{
 			input: `job="value`,
--- a/pkg/labels/parse_test.go
+++ b/pkg/labels/parse_test.go
@@ -190,6 +190,10 @@ func TestMatchers(t *testing.T) {
 			}(),
 		},
 		{
+			input: `job=`,
+			err:   `matcher value is not present: job=`,
+		},
+		{
 			input: `job="value`,
 			err:   `matcher value contains unescaped double quote: "value`,
 		},


roidelapluie (commented):
Should we fail? Or just it should match the empty value?

gotjosh:
> Should we fail? Or just it should match the empty value?

I pondered a bit on that as that was my initial approach, but it felt like allowing it might have an unwanted side effect (and we fail on a _lot_ of cases) - I have no good reference points and couldn't find any in the code.

That being said, I'm happy to rely on your expertise and experience for this one - do you think we should fall back to `""`?

roidelapluie:
Empty labels are valid in Prometheus and alertmanager.

cc @beorn7 WDYT here?

beorn7:
I think it should be possible to create a silence for an empty label, which is equivalent to an non-existing label, or in other words, it gives you a straightforward way of silencing all alerts that _do not have_ that label (with a non-empty value).

If we forbid `foo=""`, it becomes kind of tedious to formulate that, e.g. `foo!~".+"`.

Maybe we could allow `foo=""` and error on `foo=`? But this feels a bit inconsistent.

roidelapluie:
yeah, "" will be dropped by shell for most users anyway. I suggest we consider it as an empty value.

simonpasquier:
> Should we fail? Or just it should match the empty value?

not feeling strongly on this one but as the parse is already quite liberal ([example](https://github.com/prometheus/alertmanager/blob/3f3e2ce4dac4090a5ce7361e50d09b14c7dea313/pkg/labels/parse_test.go#L168)), I would ok with `foo=` being equivalent to `foo=""`.

gotjosh:
> Maybe we could allow foo="" and error on foo=? But this feels a bit inconsistent.

I thought that's what I was doing with this PR, but after a close second look at the problem (not just the solution), I realise what's currently going on.

If you look closely at the command executed as part of https://github.com/prometheus/alertmanager/issues/2936 it's:

```shell
./amtool --alertmanager.url=http://127.0.0.1:9093 alert add foo=
```

Which I believe it's a mistake. It doesn't include the alertname as the first argument - I think this should have resulted in an error so that the user is aware.

## Alertname must be present

Now, for this case, I propose the following: Allow the first argument to be used as a whole `string` if it doesn't contain the `	model.AlertNameLabel` as part of the `string` or we're unable to parse the label throw an error.

Case #1: ✅
```
./amtool --alertmanager.url=http://127.0.0.1:9093 alertname=AlwaysFiring
# results in alertname=AlwaysFiring

```

Case #2: ✅
```
./amtool --alertmanager.url=http://127.0.0.1:9093 AlwaysFiring
# results in alertname=AlwaysFiring
```

Case #3: ❌
```
./amtool --alertmanager.url=http://127.0.0.1:9093 alertname=
error: unable to determine alertname
```

Case #4: ❌
```
./amtool --alertmanager.url=http://127.0.0.1:9093 foo=
error: unable to determine alertname
```

## Empty value labels

Now, this brings us to the second question... The consensus I take from this discussion seems to be that: `foo=` is the same as `foo=""`. I think we're able to keep that behaviour for labels that are not the alertname, see below.


Case #1: ✅
```
./amtool --alertmanager.url=http://127.0.0.1:9093 AlwaysFiring foo=
```

Case #2:  ✅
```
./amtool --alertmanager.url=http://127.0.0.1:9093 AlwaysFiring foo=""
```

## Josh's opinion

Personally, I would prefer to error on `foo=` but as @simonpasquier pointed out - the parser is already quite flexible with this so I'm happy to accommodate. I'll have to look at a few more places where we use `ParseMatchers` to make sure I'm not breaking something but overall sounds doable.

If we agree with the above, I'm happy to take it for another spin.

roidelapluie:
Why would alertname need to be present? I can be relabeled away by Prometheus alert_relabel_config ..

gotjosh:
After Julien's comment, I took another holistic look and these are my findings:

1. `alertname` as a label is not required by the API (I thought it was - my bad 🤦)
2. Empty labels don't make it to the alertmanager, [they're removed by the add alerts API endpoint](https://github.com/prometheus/alertmanager/blob/main/api/v2/api.go#L332)
3. Silences allow empty matchers meaning `foo=""` is a valid matcher as long as there is at least one non-empty matcher. See [here](https://github.com/prometheus/alertmanager/blob/main/silence/silence.go#L489)

To top the above, it seems like users have also reported confusion on different behaviour between silence matchers vs alert matchers with https://github.com/prometheus/alertmanager/issues/2610. This is because the "filters" params on that endpoints treats it like any other matcher, however, silences when set have additional validation on them. 

I'm struggling to come up with suggestions on how to best proceed forward that a) don't incur a breaking change and b) solve the problem for both adding alerts and creating silences - any context to share here?

Concretely, I'm struggling with:

- I see no reason why the CLI should allow empty labels for adding alerts if they're going to end up being removed anyways. However, I see the appeal for silences.
- If `alertname` is not required why do we have special handling for it in the CLI? IMO this is better suited for a CLI flag that when specified it helps you create the `alertname` e.g. `./amtool add alert -n AlwaysFiring foo=bar`. As opposed to inferring that [an error in parsing the first label means you're trying to set the `alertname` directly](https://github.com/prometheus/alertmanager/blob/main/cli/alert_add.go#L76-L78). 

tl;dr: Does it feel like we're at a point where a silence matcher has different semantics from a label matcher even if they have the same implementation (even if theoretically they're to match just key/value pairs)? I can see the appeal of separating the matching logic and documenting/validating the semantics of them concretely so that the CLIs, APIs and UI benefit from a consistent implementation. 

We don't have to turn this into a long discussion async, I'll add this to the WG so that we can discuss it.

beorn7:
WRT the `amtool` behavior: I agree that the special handling of the alertname feels weird. I guess that has historically evolved.

WRT at least one matcher that doesn't match the empty string: I'm not sure what the actual reason for this is, but I assume it's an efficiency concern. Any matcher matching the empty string essentially requires the equivalent of a "full table scan". If there is at least one matcher not matching the empty string, at least that "full table scan" is limited to whatever that one "proper" matcher returns rather than all alerts. To further speculate, I assume that the filter parameter on the API endpoint is more liberal because it's a user initiated action. If it is too expensive, well, then it times out eventually, but that's it. However, silences have to be checked continuously. If that's too expensive, the AM is permanently overloaded. (Again, this is all educated guessing. My knowledge of the AM internals is limited.) I do think that the preview should implement the same restrictions as the silence itself (cf. #2610). (The reason for the different behavior can be explained by my speculation above: One-time user initiated action vs. permanently running queries. Still, of course, we should do the same evaluation for the former in this case.)

WRT allowing empty labels for adding alerts: While an empty label is the same as a non-existing label, it is still valid and has well defined semantics throughout the Prometheus ecosystem. It would be really confusing if the only place where you ban this kind of redundancy would be the API for adding alerts.

In summary, I don't think that the matchers have different semantics for silences vs. alerts. I would even say that what we have is all rooted in the intention to keep the semantics the same. The one thing that sticks out (at least one matcher not matching the empty string) isn't really semantics of the matchers themselves but a higher level query requirement (which exists in the same way in Prometheus itself).

gotjosh:
After recently experiencing a panic in one of our largest environments, I realised this is not worth blocking on some sort of theoretical consistency across silences and alerts APIs.

The general comments have a tendency on `foo=` being equal to `foo=""` and that's what this PR does. Effectively avoiding panic and providing the user with a consistent error message.

As an example, in the case of the reported issue:

```
 ~> ./amtool --alertmanager.url=http://127.0.0.1:9093 alert add foo=
amtool: error: [POST /alerts][400] postAlertsBadRequest  at least one label pair required
```

instead of a panic.

I think we should move forward and merge this as this problem has already been reported several times on multiple fronts.

roidelapluie (approved):
LGTM

PR #2970, opened 2022-06-29
Deprecate and remove api/v1/

### LINKED ISSUE #2469: Remove Alertmanager API v1
Alertmanager API v2 is available since v0.16.0 and it's probably fair to say that it's (at least) feature-equivalent to the API v1 since v0.19.0. At the time when v0.16.0 was released, the [CHANGELOG](https://github.com/prometheus/alertmanager/blob/master/CHANGELOG.md#0160--2019-01-17) stated that v1 would be removed in v0.18.0. Obviously it hasn't happened yet and this issue is to discuss if/when it should happen.

cc @roidelapluie @beorn7 

xref https://github.com/prometheus/prometheus/issues/8398

gotjosh:
This is intentionally incomplete and is open mostly as a guideline to see what breaks.

The description will be updated once the work is finished and serves as a checklist to give visibility of a plan for reviewers.

fixes #2469 

The steps to take are:

- [x] Allow `/api/v1/alerts` to behave like `/api/v2/alerts`.
- [x] Update comments and resolve TODOs if applicable - I don't have any context on them yet but I'll find out.
- [x] Revise any configuration options to see if there are any applicable only to `api/v1`

**Edit:**

- Instead of making `api/v1/*` redirect to the `/api/v2` I opted for adding a deprecation notice in JSON to all of the v1 routes. On top of that, each time a v1 route is called it'll log a warning letting the operator know that something is not going right. 
- I checked, TODOs and the removed code to see if there was anything I was missing but I couldn't find anything.


api/v1/api.go
```diff
@@ -1,808 +0,0 @@
...
-	}
-
-	return &API{
-		alerts:         alerts,
-		silences:       silences,
-		getAlertStatus: sf,
-		uptime:         time.Now(),
-		peer:           peer,
-		logger:         l,
-		m:              metrics.NewAlerts("v1", r),
```
gotjosh, line 106: Technically, we can remove this distinction now but it would be a rather critical breaking change - I would rather not risk it.

gotjosh: nvm, I made it permament.

gotjosh:
@simonpasquier I think this ready for review - this is currently blocking the progress on #3567 as making API v1 compatible with that work is a large endeavour for something we don't need.

Would appreciate a speedy review - I think the changes to the deprecation router are good enough to let users know that something that we deprecated back in 0.18 is now gone.

api/v1_deprecation_router.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (dr *V1DeprecationRouter) deprecationHandler(w http.ResponseWriter, req *http.Request) {
+	level.Warn(dr.logger).Log("msg", "v1 API received a request on a removed endpoint", "path", req.URL.Path, "method", req.Method)
+
+	resp := struct {
+		Status string `json:"status"`
+		Error  string `json:"error"`
+	}{
+		"deprecated",
+		"The Alertmanager v1 API was deprecated in version 0.16.0 and entirely removed since version 0.28.0 - please use the equivalent route in the v2 API",
```
grobinson-grafana, line 58: Small nit, feel free to ignore 😄 

```suggestion
		"The Alertmanager v1 API was deprecated in version 0.16.0 and is removed as of version 0.28.0 - please use the equivalent route in the v2 API",
```

api/v1_deprecation_router.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(410)
```
grobinson-grafana, line 62: Nice! ❤️

api/v1_deprecation_router.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// We don't care about errors for this route.
+	b, _ := json.Marshal(resp)
```
grobinson-grafana, line 65: You don't need to change this, but you can also use:

```go
if err := json.NewEncoder(w).Encode(resp); err != nil {
    level.Error(dr.logger).Log("msg", "failed to write response", "err", err)
}
```

grobinson-grafana (approved):
Tested and LGTM! ❤️

beorn7:
Looks great in general, but you are so deep into the weeds here that I don't feel qualified to evaluate the details. :)

gotjosh:
Thanks @beorn7 - With the amount of context you have, you'll always be overqualified to review anything we throw at you.

alanprot:
Should we follow up with a change on prometheus to now allow configuring v1 endpoints anymore?

https://github.com/prometheus/prometheus/blob/9b7de4778732ca2f2ab5028e9d1955109f440c4c/config/config.go#L910-L921

grobinson-grafana:
> Chore: Update Prometheus cortexproject/cortex#5841

Yes please Alan! Would you be happy to open a PR for that?

alanprot:
Np: https://github.com/prometheus/prometheus/pull/13883

PR #2971, opened 2022-06-29
Silence creator from HTTP header or Basic Auth username

### LINKED ISSUE #1196: Feature request: restrict the identity of silence creator to one passed by some HTTP header

gotjosh:
The discussion is long as part of #1196, so instead of building this in isolation - I'm putting this out there to gather some feedback as I know there's some historic context and I want people to voice their opinions.

The first step is the building block for the server part (I'd argue this is the easy part):

1. Try to infer the username from basic authentication
2. If no basic auth is present, try the header which is specified via a configuration option

The other bit is the front-end part (the actual hard one), a couple of thoughts:

- When the silence form is rendered, there's no HTTP call as part of that and we all know that is not possible to [access headers from javascript](https://stackoverflow.com/questions/220231/accessing-the-web-pages-http-headers-in-javascript).
- There are several approaches to solving the above, it can range from: Including the parsed username as part of the response to every HTTP call, creating a dedicated endpoint e.g. `/api/v1/me`, including it as part of the status endpoint and calling that as we render the silence form, etc. any opinions on what route should we take? I can start to see what Brian meant that this can very quickly turn into a scope creep for authentication things.
- The front-end experience that I have in my head is the following:
    - Sever tells you your current username (try basic auth first, if not then header)
    - If there's any, created by field gets filed in and disabled (greyed-out)
    - Any thoughts?

Fixes #1196

TODO

- [ ] tests
- [ ] front-end

### DIFF
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -626,6 +626,17 @@ func (api *API) deleteSilenceHandler(params silence_ops.DeleteSilenceParams) mid
 func (api *API) postSilencesHandler(params silence_ops.PostSilencesParams) middleware.Responder {
 	logger := api.requestLogger(params.HTTPRequest)
 
+	username := api.usernameFromHeaderOrBasicAuth(params.HTTPRequest)
+	if username != "" {
+		if params.Silence.CreatedBy == nil {
+			params.Silence.CreatedBy = &username
+		}
+
+		if params.Silence.CreatedBy != &username {
+			return silence_ops.NewPostSilencesBadRequest().WithPayload(fmt.Sprintf("created_by does not match HTTP basic authentication or value of %s HTTP header", api.alertmanagerConfig.Global.UserHTTPHeader))
+		}
+	}
+
 	sil, err := PostableSilenceToProto(params.Silence)
 	if err != nil {
 		level.Error(logger).Log("msg", "Failed to marshal silence to proto", "err", err)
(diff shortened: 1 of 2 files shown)


api/v2/api.go
```diff
@@ -660,6 +671,18 @@ func (api *API) postSilencesHandler(params silence_ops.PostSilencesParams) middl
 	})
 }
 
+// usernameFromHeaderOrBasicAuth returns the username specified as part of Basic Authentication. If no auth is provided,
+// it tries the specified HTTP header of UserHTTPHeader. If none are found it returns an empty string "".
+func (api *API) usernameFromHeaderOrBasicAuth(request *http.Request) string {
+	// First, let's try getting the username from Basic Authentication.
+	if user, _, ok := request.BasicAuth(); ok {
+		return user
```
roidelapluie, line 679: This is not correct, we should have this configured. if a specific header is specified in config, we should always use it, even if a user adds a custom authorization header.

use case for this:

- you put an open id connect proxy in front of alertmanager, which pass the username as http header
- communication between the proxy and alertmanager is done via static basic auth and TLS

m-masataka:
This is my sample front-end program with this PR. [Here](https://github.com/m-masataka/alertmanager/pull/1)
I'm implementing this based on the following idea

> creating a dedicated endpoint e.g. /api/v1/me

However, there are still some points to be discussed.
- Should the creator be overwritten on the **edit silence view**?
- When the front-end get username? (I think it should be called on reload.) 
- How to know whether this auth option is enable or not in front-end?
  - I think we should get the status with the response of ``api/v../me`` .


[demo]

https://user-images.githubusercontent.com/15244170/177025919-1b2010f6-0336-4238-8424-edd3895bd051.mov

gotjosh:
Wow! Thank you so much @m-masataka this looks great. I'm not sure a new API endpoint is the direction we'll take but the front-end changes look 👌 

As for your questions:

> Should the creator be overwritten on the edit silence view?

Yes, if there is a value available for the creator we should always use it. Validation will fail if the user creating/editing the silence does not match the current value. This effectively turns the feature is a type of last-write wins semantic where the creator is actually a last modified. 

> When the front-end get username? (I think it should be called on reload.)

Assuming that reload means on every time the silence form is rendered then we agree. 

> How to know whether this auth option is enabled or not in front-end?

We'll return a 204 status code instead of a 200. 204 non-content basically means we have unable to a value.

AakarshitAgarwal:
Can we restrict for many and recheck the Restrictors/Creator by either username or email?

m-masataka:
> Can we restrict for many and recheck the Restrictors/Creators by either username or email?

If you means set username **or** email I think it depends on your identity proxy. @AakarshitAgarwal 

We can set one custom header to use  restrict Creator in this PR. so we can't set many creators.

genofire:
Any update, i would really love this feature ;)

PR #2972, opened 2022-06-29
Mixin: Fix linter erorrs

gotjosh:
In accordance with a new rule introduced as part of https://github.com/grafana/dashboard-linter/pull/79 this is now required. However, for the new rule of `panel-unit-rule` we don't reap any benefits from specifying a particular unit for our panels, the defaults work perfectly fine so I've chosen to ignore that rule instead.


simonpasquier (approved):
lgtm

Issue #2961, opened 2022-06-23
Split retention in 2 parts or multiple nflogs


roidelapluie:
We have a default of `120h` retention.

While this default seems fine for the silences, it seems a lot too high for the nflog. Indeed, the nflog should ideally be kept only for ~110% of `x=max( group_wait, group_interval, repeat_interval)`. When having a large number of alerts and a low `x`, alertmanager un-necessarily uses a lot of memory, because the state is broadcasted perpetually.

Here a heap of such a case: https://share.polarsignals.com/73d955e/

I see multiple ways forward:

- Multiple nflogs, with exponential durations (e.g. 1m, 1h, 12h, 10h), based on the duration the notification should stay available in the log.
- Allow setting a nflog-specific data retention policy.

roidelapluie:
Maybe we can write a new nflog where we remove the duplicates as well as part of some garbage collection process.

PR #2982, opened 2022-07-05
Expires notify log sooner when possible

roidelapluie:
It seems useless to keep the notifications in the nflog for longer than
twice the repeat interval. This should help reduce memory usage of
clustered alertmanagers.

- [x] needs tests

Signed-off-by: Julien Pivotto <roidelapluie@o11y.eu>

### DIFF
--- a/nflog/nflog.go
+++ b/nflog/nflog.go
@@ -399,7 +399,7 @@ func stateKey(k string, r *pb.Receiver) string {
 	return fmt.Sprintf("%s:%s", k, receiverKey(r))
 }
 
-func (l *Log) Log(r *pb.Receiver, gkey string, firingAlerts, resolvedAlerts []uint64) error {
+func (l *Log) Log(r *pb.Receiver, gkey string, firingAlerts, resolvedAlerts []uint64, expiry time.Duration) error {
 	// Write all st with the same timestamp.
 	now := l.now()
 	key := stateKey(gkey, r)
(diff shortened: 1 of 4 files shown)


roidelapluie:
Attempt to fix #2961

roidelapluie:
@gotjosh We have ran this and noticed a 30% memory decrease with alertmanager with low repeat interval. Any chances to get this merged?

notify/notify.go
```diff
@@ -785,7 +785,12 @@ func (n SetNotifiesStage) Exec(ctx context.Context, l log.Logger, alerts ...*typ
 		return ctx, nil, errors.New("resolved alerts missing")
 	}
 
-	return ctx, alerts, n.nflog.Log(n.recv, gkey, firing, resolved)
+	var expiry time.Duration
+	if n, ok := RepeatInterval(ctx); ok {
```
simonpasquier, line 789: The repeat interval should always be present in the context? At least  `*DedupStage.Exec()` returns an error if this is the case. I suggest that we do the same here.

roidelapluie: Done!

nflog/nflog.go
```diff
@@ -415,6 +415,11 @@ func (l *Log) Log(r *pb.Receiver, gkey string, firingAlerts, resolvedAlerts []ui
 		}
 	}
 
+	expiresAt := now.Add(l.retention)
+	if expiry > 0 && l.retention > expiry {
```
simonpasquier, line 419: wouldn't it make sense to use the default retention only if expiry > 0?
If repeat_interval is 10d (because why not), it would make sense to keep the log for longer than the default 120h retention. We could also get rid of this check:

https://github.com/prometheus/alertmanager/blob/d034f116d52912f00e444b1b07bb0e97a168074e/cmd/alertmanager/main.go#L488-L500

simonpasquier (approved):
lgtm. I have a question on how the data retention interacts with repeat interval.

roidelapluie:
This would be change of behaviour, let's merge this while we think on the further change

PR #2983, opened 2022-07-05
Simplify Silence test to remove unnecessary wait.

joeblubaugh:
As noted in #2867, there is an unnecessary require.Eventually in a
silence test. This PR addresses that by using a channel to signal that
that the maintenance loop has completed.

@gotjosh as a follow-up


gotjosh:
Thanks!

PR #2985, opened 2022-07-06
Speed up acceptance tests

joeblubaugh:
This change marks two unrelated acceptance tests in each path as
parallel so that they can run concurrently. It also sets a
higher-than-default parallelism argument in the CircleCI config. The
higher config works well for these tests because they spend a large
amount of time in a halted state, so there's no CPU contention to worry
about.

On an M1 Macbook, test execution time drops from ~30s to ~15s.

Signed-off-by: Joe Blubaugh <joe.blubaugh@grafana.com>

cc @gotjosh


joeblubaugh:
Sorry for the large # of pushes. Validating CircleCI locally is tricky.

SuperQ:

PR #3005, opened 2022-07-14
Fix wechat secret may contain newlines

lou-lan:


lou-lan:
ping @roidelapluie

lou-lan:
ping @simonpasquier

lou-lan:
@juliusv ping

simonpasquier (commented):
Thanks for the PR. Can you add a unit test? I'm also wondering whether `corpid` would need the same treatment?

lou-lan:
> Thanks for the PR. Can you add a unit test? I'm also wondering whether `corpid` would need the same treatment?

I'll add the test later, but corp id won't, because he didn't add it from secret.

SuperQ:
This seems like it's a bug in the operator, not in the configuration parsing.

Issue #2474, opened 2021-02-02
Repeated resolved alerts after 'repeat_interval'


noevhlev:
After updating alertmanager from version 0.19 to 0.20, I had a problem. When I group alerts, the resolved alerts are not disappear anymore in the case that there are already firing alerts in the group. When upgrading to version 0.21, the problem is still reproducible.

It seems this problem appeared after [PR#2040](https://github.com/prometheus/alertmanager/pull/2040).
In the annotation to the PR, it was written that the aggregation group is responsible for removing the resolved alerts. However, this does not happen - perhaps I've missed something.

**What did you expect to see?(That's how it worked before):**
- I get 10 firing alerts
- After sometime I get 3 resolved alerts - So there are 7 firing alerts and 3 resolved alerts
- After repeat_interval, there are 7 firing alerts

**What did you see instead?**
- I get 10 firing alerts
- After sometime I get 3 resolved alerts - So there are 7 firing alerts and 3 resolved alerts
- After repeat_interval I get 7 firing alerts and 3 resolved alerts. Nothing changes.

**Environment**

* System information:

	Linux 4.15.0-115-generic x86_64

* Alertmanager version:

	alertmanager, version 0.20.0 (branch: HEAD, revision: f74be0400a6243d10bb53812d6fa408ad71ff32d)
          build user:       root@00c3106655f8
          build date:       20191211-14:13:14
          go version:       go1.13.5

* Prometheus version:

	prometheus, version 2.18.1 (branch: HEAD, revision: ecee9c8abfd118f139014cb1b174b08db3f342cf)
          build user:       root@2117a9e64a7e
          build date:       20200507-16:51:47
          go version:       go1.14.2

* Alertmanager configuration file:
```
global:
  resolve_timeout: 3m
...

route:
  group_by:
  - environment
  - alertname
  - severity
  group_wait: 20s
  group_interval: 80s
  repeat_interval: 3m
...
```

* Logs:
```
Feb 02 02:49:44 pupa alertmanager[18119]: level=debug ts=2021-02-01T23:49:44.364Z caller=dispatch.go:138 component=dispatcher msg="Received alert" alert=DiskReplace[1d2cb32][active] 
Feb 02 02:49:44 pupa alertmanager[18119]: level=debug ts=2021-02-01T23:49:44.364Z caller=dispatch.go:138 component=dispatcher msg="Received alert" alert=DiskReplace[281d27a][active] 
[...]

PR #3006, opened 2022-07-15
Fix repeated resolved alerts

damnever:
Assumptions:
1. Prometheus/XXRuler failed to send some active alerts, but succeeded to send all resolved alerts:
    alert-a: T1(active:notify-ok), T2(resolved:notify-ok), **T3(active:notify-failed)**, T4(resolved:notify-ok).
2. Or maybe the alertmanager is behind a proxy with some retry policy.

Reproduce:
1. A **repeated** `alert-a[resolved]` is received during [aggrGroup.flush](https://github.com/prometheus/alertmanager/blob/da6de1f1cc0c08a2e93c48c210954b3f0b34d954/dispatch/dispatch.go#L495).
2. We can not [delete `alert-a[resolved]`](https://github.com/prometheus/alertmanager/blob/da6de1f1cc0c08a2e93c48c210954b3f0b34d954/dispatch/dispatch.go#L528) since the alert has been changed.
3. We receive a new `alert-b[firing]` during `group_interval`, then both `alert-a[resolved]` and `alert-b[firing]` will be sent out.
4. Users may receive `alert-a[resolved]` multiple times.

There is another case that may trigger this issue, if the time is right, the repeated `alert-a[resolved]` is received right after [aggrGroup](https://github.com/prometheus/alertmanager/blob/da6de1f1cc0c08a2e93c48c210954b3f0b34d954/dispatch/dispatch.go#L184) gets deleted.

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -475,8 +475,12 @@ func (ag *aggrGroup) stop() {
 
 // insert inserts the alert into the aggregation group.
 func (ag *aggrGroup) insert(alert *types.Alert) {
-	if err := ag.alerts.Set(alert); err != nil {
-		level.Error(ag.logger).Log("msg", "error on set alert", "err", err)
+	if err := ag.alerts.SetOrReplaceResolved(alert); err != nil {
+		if errors.Is(err, store.ErrNotFound) {
+			level.Warn(ag.logger).Log("msg", "ignore resolved alert since there is no corresponding record in the store")
+		} else {
+			level.Error(ag.logger).Log("msg", "error on set alert", "err", err)
+		}
 	}
 
 	// Immediately trigger a flush if the wait duration for this
@@ -517,17 +521,15 @@ func (ag *aggrGroup) flush(notify func(...*types.Alert) bool) {
 
 	if notify(alertsSlice...) {
 		for _, a := range alertsSlice {
-			// Only delete if the fingerprint has not been inserted
+			// Only delete the resolved alert if the fingerprint has not been active
 			// again since we notified about it.
-			fp := a.Fingerprint()
-			got, err := ag.alerts.Get(fp)
-			if err != nil {
-				// This should never happen.
-				level.Error(ag.logger).Log("msg", "failed to get alert", "err", err, "alert", a.String())
+			if !a.Resolved() {
 				continue
 			}
-			if a.Resolved() && got.UpdatedAt == a.UpdatedAt {
-				if err := ag.alerts.Delete(fp); err != nil {
+			if err := ag.alerts.DeleteIfResolved(a.Fingerprint()); err != nil {
+				if errors.Is(err, store.ErrNotResolved) {
+					level.Debug(ag.logger).Log("msg", "resolved alert has been active again", "alert", a.String())
+				} else {
 					level.Error(ag.logger).Log("msg", "error on delete alert", "err", err, "alert", a.String())
 				}
 			}
--- a/dispatch/dispatch_test.go
+++ b/dispatch/dispatch_test.go
@@ -601,6 +601,23 @@ func (r *recordStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.A
 	return ctx, nil, nil
 }
 
+type notifyStage struct {
+	C chan struct{}
+}
+
+func newNotifyStage() *notifyStage {
+	return &notifyStage{C: make(chan struct{}, 1)}
+}
+
+func (s *notifyStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.Alert) (context.Context, []*types.Alert, error) {
+	select {
+	case <-ctx.Done():
+		return ctx, nil, ctx.Err()
+	case s.C <- struct{}{}:
+		return ctx, alerts, nil
+	}
+}
+
 var (
 	// Set the start time in the past to trigger a flush immediately.
 	t0 = time.Now().Add(-time.Minute)
@@ -651,7 +668,7 @@ func TestDispatcherRaceOnFirstAlertNotDeliveredWhenGroupWaitIsZero(t *testing.T)
 	route := &Route{
 		RouteOpts: RouteOpts{
 			Receiver:       "default",
-			GroupBy:        map[model.LabelName]struct{}{"alertname": {}},
+			GroupBy:        map[model.LabelName]struct{}{model.AlertNameLabel: {}},
 			GroupWait:      0,
 			GroupInterval:  1 * time.Hour, // Should never hit in this test.
 			RepeatInterval: 1 * time.Hour, // Should never hit in this test.
@@ -666,7 +683,7 @@ func TestDispatcherRaceOnFirstAlertNotDeliveredWhenGroupWaitIsZero(t *testing.T)
 
 	// Push all alerts.
 	for i := 0; i < numAlerts; i++ {
-		alert := newAlert(model.LabelSet{"alertname": model.LabelValue(fmt.Sprintf("Alert_%d", i))})
+		alert := newAlert(model.LabelSet{model.AlertNameLabel: model.LabelValue(fmt.Sprintf("Alert_%d", i))})
 		require.NoError(t, alerts.Put(alert))
 	}
 
@@ -682,6 +699,80 @@ func TestDispatcherRaceOnFirstAlertNotDeliveredWhenGroupWaitIsZero(t *testing.T)
 
 	// We expect all alerts to be notified immediately, since they all belong to different groups.
 	require.Len(t, recorder.Alerts(), numAlerts)
+}
+
+func TestDispatcherReceiveAndNotifyRepeatedResolvedAlerts(t *testing.T) {
+	// More information here: https://github.com/prometheus/alertmanager/pull/3006
+
+	logger := log.NewNopLogger()
+	// logger := log.NewLogfmtLogger(os.Stdout)
+	marker := types.NewMarker(prometheus.NewRegistry())
+	alerts, err := mem.NewAlerts(context.Background(), marker, time.Hour, nil, logger, nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer alerts.Close()
+
+	route := &Route{
+		RouteOpts: RouteOpts{
+			Receiver:       "default",
+			GroupBy:        map[model.LabelName]struct{}{model.AlertNameLabel: {}},
+			GroupWait:      0,
+			GroupInterval:  6 * time.Second,
+			RepeatInterval: 6 * time.Second,
+		},
+	}
+
+	recorder := &recordStage{alerts: make(map[string]map[model.Fingerprint]*types.Alert)}
+	notifier := newNotifyStage()
+	dispatcher := NewDispatcher(alerts, route, notify.FanoutStage{recorder, notifier}, marker, nil, nil, logger, NewDispatcherMetrics(false, prometheus.NewRegistry()))
+	go dispatcher.Run()
+	defer dispatcher.Stop()
+
+	// Here we simulate the case when the alertmanager receives resolved alerts
+	// right after the aggrGroup is deleted.
+	t.Run("repeated alerts after aggrGroup deleted", func(t *testing.T) {
+		alert := newAlert(model.LabelSet{model.AlertNameLabel: "test-repeated-resolved-alerts-1"})
+		alert.Alert.EndsAt = alert.StartsAt.Add(time.Second)
+		alert.UpdatedAt = alert.Alert.EndsAt
+		require.True(t, alert.Resolved())
+		require.NoError(t, alerts.Put(alert))
+		select {
+		case <-time.After(20 * time.Second):
+		case <-notifier.C:
+			t.Errorf("unexpected repeated resolved alerts")
+		}
+	})
+
+	// Alertmanager receives repeated resolved alerts after aggrGroup.flush.
+	t.Run("repeated alerts after aggrGroup flush", func(t *testing.T) {
+		alert := newAlert(model.LabelSet{model.AlertNameLabel: "test-repeated-resolved-alerts-2"})
+		require.NoError(t, alerts.Put(alert))
+		select {
+		case <-time.After(20 * time.Second):
+			t.Errorf("wait active alert timed out")
+		case <-notifier.C:
+		}
+
+		alert.Alert.EndsAt = alert.StartsAt.Add(time.Second)
+		alert.UpdatedAt = alert.Alert.EndsAt
+		require.True(t, alert.Resolved())
+		require.NoError(t, alerts.Put(alert))
+		select {
+		case <-time.After(20 * time.Second):
+			t.Errorf("wait resolved alert timed out")
+		case <-notifier.C:
+		}
+
+		alert.UpdatedAt = alert.Alert.EndsAt.Add(time.Second)
+		require.True(t, alert.Resolved())
+		require.NoError(t, alerts.Put(alert))
+		select {
+		case <-time.After(20 * time.Second):
+		case <-notifier.C:
+			t.Errorf("unexpected repeated resolved alerts")
+		}
+	})
 }
 
 type limits struct {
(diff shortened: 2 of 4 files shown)


dispatch/dispatch.go
```diff
@@ -516,17 +521,15 @@ func (ag *aggrGroup) flush(notify func(...*types.Alert) bool) {
...
-			got, err := ag.alerts.Get(fp)
-			if err != nil {
-				// This should never happen.
-				level.Error(ag.logger).Log("msg", "failed to get alert", "err", err, "alert", a.String())
+			if !a.Resolved() {
 				continue
 			}
-			if a.Resolved() && got.UpdatedAt == a.UpdatedAt {
-				if err := ag.alerts.Delete(fp); err != nil {
+			if err := ag.alerts.DeleteIfResolved(a.Fingerprint()); err != nil {
```
damnever, line 529: We need an atomic operation to ensure no race between Get and Delete.

roidelapluie:
I think this might be on purpose and written for simplicity. Alertmanager has at least once promise, it's better to send one extra alert than miss one.

damnever:
@roidelapluie Yes, alertmanager has at least once promise. I think that is because alertmanager will retry until the notification success, then remove those resolved alerts.

But in this case, we are sending repeated(orphan) resolved alerts. I'd consider this as a BUG rather than a "feature".

damnever:
[#2474](https://github.com/prometheus/alertmanager/issues/2474) could be the case.

damnever:
@roidelapluie  @simonpasquier @w0rm Please take a look in detail.

simonpasquier:
After reviewing the Alertmanager code and history, I kinda agree that there's room for improvement. As far as I can tell while https://github.com/prometheus/alertmanager/pull/2040 attempted to fix a race between the aggregation group and the alert store (e.g. a resolved alert could be removed before being notified) by disabling garbage-collection in the alerts store, it introduced a side effect that increased the likelihood of receiving several notifications with the same resolved alert.

Any change in the current logic would need careful attention though IMHO. Ideally a unit test highlighting the current issue in `dispatch.aggrGroup` would be a good starting point.

damnever:
@simonpasquier  I have added test cases for this issue, the main branch will fail the test:
```
gotest -v -race -run TestDispatcherReceiveAndNotifyRepeatedResolvedAlerts
=== RUN   TestDispatcherReceiveAndNotifyRepeatedResolvedAlerts
=== RUN   TestDispatcherReceiveAndNotifyRepeatedResolvedAlerts/repeated_alerts_after_aggrGroup_deleted
    dispatch_test.go:743: unexpected repeated resolved alerts
=== RUN   TestDispatcherReceiveAndNotifyRepeatedResolvedAlerts/repeated_alerts_after_aggrGroup_flush
    dispatch_test.go:773: unexpected repeated resolved alerts
--- FAIL: TestDispatcherReceiveAndNotifyRepeatedResolvedAlerts (12.01s)
    --- FAIL: TestDispatcherReceiveAndNotifyRepeatedResolvedAlerts/repeated_alerts_after_aggrGroup_deleted (0.00s)
    --- FAIL: TestDispatcherReceiveAndNotifyRepeatedResolvedAlerts/repeated_alerts_after_aggrGroup_flush (12.00s)
FAIL
exit status 1
FAIL    github.com/prometheus/alertmanager/dispatch     13.180s
```

damnever:
Ping someone..

bo-er:
@damnever  Hi! Apologize for asking, but could you please share some debug log here? I'm supposably having the same problem.

damnever:
@bo-er I do not have any logs to reproduce the original issue since we are currently using this patch in our production environment. However, I can provide some logs that were generated by this patch:
```
... msg="ignore resolved alert since there is no corresponding record in the store"
```

damnever:
@simonpasquier, could you please take another look at this? We have been running this in our production environment for over a year. I am confident that it has not introduced any side effects. In fact, it has even resolved some issues, especially for users who were using a very short repeat interval and so on.

dispatch/dispatch_test.go
```diff
@@ -684,6 +701,80 @@ func TestDispatcherRaceOnFirstAlertNotDeliveredWhenGroupWaitIsZero(t *testing.T)
...
+
+	recorder := &recordStage{alerts: make(map[string]map[model.Fingerprint]*types.Alert)}
+	notifier := newNotifyStage()
+	dispatcher := NewDispatcher(alerts, route, notify.FanoutStage{recorder, notifier}, marker, nil, nil, logger, NewDispatcherMetrics(false, prometheus.NewRegistry()))
+	go dispatcher.Run()
+	defer dispatcher.Stop()
+
+	// Here we simulate the case when the alertmanager receives resolved alerts
+	// right after the aggrGroup is deleted.
+	t.Run("repeated alerts after aggrGroup deleted", func(t *testing.T) {
```
grobinson-grafana, line 734: Hi! 👋 Josh has asked if I can help take a look at this issue. I'm looking at this test here but I'm not sure I understand what is being tested and how this is showing one of the reported race conditions (I understand there are at least two?).

Here we are adding an expired alert to the store and waiting for a flush. Your test expects no flush to occur, so when it does occur the test fails. What I don't understand is how this relates to the race condition? It is expected that when Alertmanager receives a resolved alert it still creates an aggregation group and flushes it.

dispatch/dispatch_test.go
```diff
@@ -684,6 +701,80 @@ func TestDispatcherRaceOnFirstAlertNotDeliveredWhenGroupWaitIsZero(t *testing.T)
...
+		alert.UpdatedAt = alert.Alert.EndsAt
+		require.True(t, alert.Resolved())
+		require.NoError(t, alerts.Put(alert))
+		select {
+		case <-time.After(20 * time.Second):
+			t.Errorf("wait resolved alert timed out")
+		case <-notifier.C:
+		}
+
+		alert.UpdatedAt = alert.Alert.EndsAt.Add(time.Second)
```
grobinson-grafana, line 767: This is the other case I don't understand. You are changing the `UpdatedAt` time, putting the alert, and expecting it be deleted. However, if the `UpdatedAt` time is incremented during a flush, the alert will not be deleted, and it will be flushed again. You won't get a second resolved notification though because the notification log will show that a resolved notification was sent. Your test skips that check though. Have I misunderstood?

grobinson-grafana, line 767: This is the other case I don't understand. You are changing the `UpdatedAt` time, putting the alert, and expecting it to be deleted. However, if the `UpdatedAt` time is incremented during a flush, the alert will not be deleted, and it will be flushed again. You won't get a second resolved notification though because the notification log will show that a resolved notification was sent. Your test skips that check though. Have I misunderstood?

dispatch/dispatch_test.go
grobinson-grafana, line 734: Hi! 👋 Josh has asked if I can help take a look at this issue. I'm looking at this test here but I'm not sure I understand what is being tested and how this is showing one of the reported race conditions (I understand there are at least two?).

Here you are adding an expired alert to the store and waiting for a flush. Your test expects no flush to occur, so when it does occur the test fails. What I don't understand is how this relates to the race condition? It is expected that when Alertmanager receives a resolved alert it still creates an aggregation group and flushes it.

dispatch/dispatch_test.go
grobinson-grafana, line 767: This is the other case I don't understand. You are changing the `UpdatedAt` time, putting the alert, and expecting it to be deleted. However, if the `UpdatedAt` time is incremented during a flush, the alert will not be deleted, and it will be flushed again. You won't get a second resolved notification though because the notification log will show that a resolved notification was sent. Your test skips that check though. Have I misunderstood?

dispatch/dispatch_test.go
grobinson-grafana, line 734: Hi! 👋 Josh has asked if I can help take a look at this issue. I'm looking at this test here but I'm not sure I understand what is being tested and how this is showing one of the reported race conditions (I understand there are at least two?).

Here you are adding an expired alert to the store and waiting for a flush. Your test expects no flush to occur, so when it does occur the test fails. What I don't understand is how this relates to the race condition? It is expected that when Alertmanager receives a resolved alert it still creates an aggregation group and flushes it.

dispatch/dispatch_test.go
grobinson-grafana, line 767: This is the other case I don't understand. You are changing the `UpdatedAt` time, putting the alert, and expecting it to be deleted. However, if the `UpdatedAt` time is incremented during a flush, the alert will not be deleted, and it will be flushed again. You won't get a second resolved notification though because the notification log will show that a resolved notification was sent. Your test skips that check though. Have I misunderstood?

damnever: If the notification log prevents the sending of repeated resolved alerts, then this is not the case.

damnever: Ok, I believe the notification log has its own garbage collection function that operates independently. This case will occur when the timing is right.

dispatch/dispatch_test.go
grobinson-grafana: >the race condition could result in normal alerts being deleted if the same alert becomes active again and is received at the same time as the deletion of the alert.

OK, I think I understand the issue!

Goroutine #1: You have a Resolved alert that is about to be deleted because the if statement is true:

```
if a.Resolved() && got.UpdatedAt == a.UpdatedAt {
	...
}
```

Goroutine #2: Between execution of the condition and execution of the code in the if statement, a firing alert is put into the store via `func (ag *aggrGroup) insert(alert *types.Alert)`, changing the alert back from resolved to firing.

Goroutine #1: The code in the if statement resumes execution and deletes the alert from the store. However, the alert has changed since the condition was execution, and `got` no longer refers to the alert in the store. The firing alert is deleted.

```
if err := ag.alerts.Delete(fp); err != nil {
	...
}
```

> We reject storing resolved alerts if the corresponding firing alert is not found, it may have been lost due to network conditions.

I don't think this fix works though. It has a bug where the Alertmanager does not accept resolved alerts following a crash or restart. Alertmanager does not write its state to disk, so there is no firing alert when the resolved alert is received.

grobinson-grafana: >the race condition could result in normal alerts being deleted if the same alert becomes active again and is received at the same time as the deletion of the alert.

OK, I think I understand the issue!

Goroutine #1: You have a Resolved alert that is about to be deleted because the if statement is true:

```
if a.Resolved() && got.UpdatedAt == a.UpdatedAt {
	...
}
```

Goroutine #2: Between execution of the condition and execution of the code in the if statement, a firing alert is put into the store via `func (ag *aggrGroup) insert(alert *types.Alert)`, changing the alert back from resolved to firing.

Goroutine #1: The code in the if statement resumes execution and deletes the alert from the store. However, the alert has changed since the condition was executed, and `got` no longer refers to the alert in the store. The firing alert is deleted.

```
if err := ag.alerts.Delete(fp); err != nil {
	...
}
```

> We reject storing resolved alerts if the corresponding firing alert is not found, it may have been lost due to network conditions.

I don't think this fix works though. It has a bug where the Alertmanager does not accept resolved alerts following a crash or restart. Alertmanager does not write its state to disk, so there is no firing alert when the resolved alert is received.

damnever: Then, I would prefer to avoid receiving an orphan resolved alert, and I think that might be why we have an option called `resolved_timeout`.

grobinson-grafana: What do you mean by orphaned alert?

damnever: In this case, AlertManager could deliver a resolved alert without a preceding firing alert, so I refer to it as an 'orphan resolved alert'.

damnever: I think the resolved alert should not be too late, so we can manage the restart scenario efficiently by comparing the process start time with the creation time of the aggregation group.

grobinson-grafana: I think that will be difficult to avoid for a number of reasons. Let's still fix the race condition though.

grobinson-grafana: OK! I am able to reproduce the issue with a unit test.

The test creates and starts a dispatcher, inserts a firing alert into the store, and waits for the alert to be flushed. It then inserts a resolved alert and waits for it to be flushed too.

Last it inserts the same firing alert that it inserted at the start of the test and waits for that to be flushed. However, the flush never happens because the firing alert is deleted.

```
=== RUN   TestDispatcherRaceFiringAlertDeleted
level=debug component=dispatcher msg="Received alert" alert=[3fff2c2][active]
level=debug component=dispatcher aggrGroup={}:{} msg=flushing alerts=[[3fff2c2][active]]
level=debug component=dispatcher msg="Received alert" alert=[3fff2c2][resolved]
level=debug component=dispatcher aggrGroup={}:{} msg=flushing alerts=[[3fff2c2][resolved]]
level=debug component=dispatcher msg="Received alert" alert=[3fff2c2][active]
level=debug component=dispatcher aggrGroup={}:{} msg="Deleted alert" alert=[3fff2c2][active]
    dispatch_test.go:693:
--- FAIL: TestDispatcherRaceFiringAlertDeleted (3.00s)
FAIL
FAIL	github.com/prometheus/alertmanager/dispatch	3.260s
FAIL
```

To make the race condition easier to observe I added a 1ms sleep to the dispatcher between the call to `Get(fp)` and `Delete(fp)`:

```go
index 104471e3..c2669496 100644
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -526,10 +526,16 @@ func (ag *aggrGroup) flush(notify func(...*types.Alert) bool) {
                                continue
                        }
+                       // Insert a 1ms sleep. This allows the scheduler to resume the test
+                       // between ag.alerts.Get(fp) and ag.alerts.Delete(fp) during which
+                       // it should replace the resolved alert with a firing alert.
+                       <-time.After(time.Millisecond)
                        if a.Resolved() && got.UpdatedAt == a.UpdatedAt {
-                               if err := ag.alerts.Delete(fp); err != nil {
+                               deleted, err := ag.alerts.Delete(fp)
+                               if err != nil {
                                        level.Error(ag.logger).Log("msg", "error on delete alert", "err", err, "alert", a.String())
                                }
+                               level.Info(ag.logger).Log("msg", "Deleted alert", "alert", deleted)
                        }
                }
        }
```

I also updated the store to return the deleted alert:

```go
diff --git a/store/store.go b/store/store.go
index 9b30542f..2d5e4666 100644
--- a/store/store.go
+++ b/store/store.go
@@ -115,12 +115,12 @@ func (a *Alerts) Set(alert *types.Alert) error {
 }

 // Delete removes the Alert with the matching fingerprint from the store.
-func (a *Alerts) Delete(fp model.Fingerprint) error {
+func (a *Alerts) Delete(fp model.Fingerprint) (*types.Alert, error) {
        a.Lock()
        defer a.Unlock()
-
+       tmp := a.c[fp]
        delete(a.c, fp)
-       return nil
+       return tmp, nil
 }
```

We can see that the the firing alert is deleted:

```
level=debug component=dispatcher aggrGroup={}:{} msg=flushing alerts=[[3fff2c2][resolved]]
level=debug component=dispatcher msg="Received alert" alert=[3fff2c2][active]
level=debug component=dispatcher aggrGroup={}:{} msg="Deleted alert" alert=[3fff2c2][active]
```

However, what should happen is no alert is deleted as the alert in the store was updated during the flush.

grobinson-grafana:
Thanks! I believe the issue is fixed with https://github.com/prometheus/alertmanager/pull/3826. I will take a look at https://github.com/prometheus/alertmanager/pull/3648 this week too!

gotjosh:
Should we close this PR then?

dispatch/dispatch_test.go
grobinson-grafana: > If the notification log prevents the sending of repeated resolved alerts, then this is not the case.

That is correct! The notification log will contain an entry for the resolved alert as it was the last successful notification sent. Entries in the notification log are deleted after the retention period (5 days). I'm still not sure I understand the bug though?

grobinson-grafana:
> Should we close this PR then?

Let's wait for @damnever. I think we fixed the issue, but I just want to confirm with them.

damnever:
I may have lost some memories as it's been too long. Let's close this

PR #3009, opened 2022-07-18
Remove ioutil

inosato:
`ioutil` is deprecated from go1.16 (ref: https://go.dev/doc/go1.16#ioutil).
So removed them.


SuperQ (approved):
LGTM, Thanks!

PR #3013, opened 2022-07-20
Move api error logs to debug

sinkingpoint:
This is a of a nitpicky change, but having error logs on bad requests
is a bit of a pain. This means that a bad client can spam the logs with
bad requests that are really not actually an issue for the server -
we just send back the error and move on. This commit moves a couple of logs
from `Error` to `Debug` so that they can be filtered a bit better


simonpasquier (approved):
Thanks, it seems legit to me.

Issue #2285, opened 2020-06-07
Feature Request: Keep cluster peer name on restart


margau:
**What did you do?**
Peer 

**What did you expect to see?**
When the alertmanager process is restarted, it should not be recognized as the old peer to the cluster. A possible solution would be an manual given peer name (like `--cluster.peer-name=`), or the use of a deterministic peer name generation.

**What did you see instead? Under which circumstances?**
When the process is restarted for whatever reason, the alertmanager joins as a new peer. The old peer is present as "failed", which is not correct.

**Environment**

* System information: n/a

* Alertmanager version: version 0.20.0 (branch: HEAD, revision: f74be0400a6243d10bb53812d6fa408ad71ff32d)

* Prometheus version: n/a

* Alertmanager configuration file: n/a

* Prometheus configuration file: n/a

* Logs: n/a

roidelapluie:
This sounds like it should not happen in the first place. Do you know why it is changing?

Overall this feature seems dangerous; it should probably work out of the box.

simonpasquier:
@roidelapluie the peer name has always been auto-generated since the clustering has been refactored to use hashicorp/memberlist. I'm not sure if there was a technical reason for this... I can only assume that not requiring an explicit peer name was deemed easier from a configuration point of view.

https://github.com/prometheus/alertmanager/blob/1895fde85692bd18dca69cbf96a7b67a3e519b22/cluster/cluster.go#L167-L170

dmitrime:
Wanted to ask about the state of this issue. Recently after restarting our Alertmanagers and looking at the dashboards, we noticed  the `alertmanager_cluster_failed_peers` metric from several peers was non-zero. This caused confusion until we realised that the restart caused this and the cluster was indeed okay, while some information about the old peers was retained. After about 6h the metric went back to zero, this seems to be controlled by `--cluster.reconnect-timeout` (we didn't override it).

In our case `alertmanager_cluster_failed_peers` was misleading, so having the option to keep old peer names after a restart would be nice to have.

SuperQ:
Fixed in #4636

machine424:
> When the process is restarted for whatever reason, the alertmanager joins as a new peer. The old peer is present as "failed", which is not correct.

[...]

PR #3014, opened 2022-07-21
Allow specifying persistent peer ids

sinkingpoint:
Currently the default behaviour is to generate a new, random peer id on
every startup. This makes clustering a bit weird, so this commit adds in
an optional cli arg that allows specifying a custom peer id that we can
persist across runs

This addresses #2285 


cmd/alertmanager/main.go
```diff
@@ -206,6 +206,7 @@ func run() int {
 		clusterBindAddr = kingpin.Flag("cluster.listen-address", "Listen address for cluster. Set to empty string to disable HA mode.").
 				Default(defaultClusterAddr).String()
 		clusterAdvertiseAddr   = kingpin.Flag("cluster.advertise-address", "Explicit address to advertise in cluster.").String()
+		clusterPeerName        = kingpin.Flag("cluster.peer-name", "Explicit name of the peer, rather than generating a random one").Default("").String()
```
simonpasquier, line 209: Maybe insist that peer names should be unique across all nodes of the cluster?

simonpasquier (commented):
Thanks! As I said in https://github.com/prometheus/alertmanager/issues/2285, I see no real reason why the peer names are auto-generated. And in kubernetes land, we could rely on the statefulset names to guarantee uniqueness.

It would be nice to have a new acceptance test to verify that it works. I've tested locally and it seems to work fine though name conflicts aren't easy to detect (they appear in the memberlist logs but those are only available at the debug level). I suppose that we'd need to implement the `ConflictDelegate` interface (https://pkg.go.dev/github.com/hashicorp/memberlist#ConflictDelegate) for the `delegate` struct?

sinkingpoint:
Thanks for the review @simonpasquier ! I've added in an acceptance test to test the behaviour. Detecting duplicate names is an interesting problem... Yes - we can implement the ConflictDelegate interface, but the question is what to do with that information - we can log it (easy), or attempt to shut down one of the duplicates, but that, in practice gets a bit messy. In particular, identifying which peer is the "invalid" one is kind of difficult - memberlist doesn't have the idea of an "authoritative" node for a given name

sinkingpoint:
Apologies for leaving this... I think that sounds reasonable. Will push that today

SuperQ:
Fixed in #4636

PR #3016, opened 2022-07-22
Capture receiver integration errors

gotjosh:
As an alerts traverses through the routing tree, there's a chance we fail to delivery the notification as we attempt to send it.

This creates a problem where a user that has a failure in their template or their receiver integration (e.g. webhook) is unavailable has no visibility on this problem.

This change, enhances the receiver API to include the last error, duration and time an alert went through the receiver.


api/v2/api.go
```diff
@@ -204,11 +208,40 @@ func (api *API) getStatusHandler(params general_ops.GetStatusParams) middleware.
...
+
+		for _, integration := range r.Integrations() {
+			notify, duration, err := integration.GetReport()
+			iname := integration.String()
+			sr := integration.SendResolved()
+
+			integrations = append(integrations, &open_api_models.Integration{
+				Name:               &iname,
+				SendResolve:        &sr,
+				LastNotify:         notify.UTC().String(),
```
JohnnyQQQQ, line 226: _As it's still a draft ignore it if you already planned to do so_: can we make sure that the date is in ISO 8601 format?

SuperQ:

PR #3017, opened 2022-07-23
Update build

SuperQ:
* Update Go to 1.18
* Update circleci machine image.
* Switch maildev to new upstream image location.
* Update Go modules to 1.17 format.
* Make dependabot monthly to match prometheus/prometheus.

Signed-off-by: SuperQ <superq@gmail.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,7 @@
 module github.com/prometheus/alertmanager
 
+go 1.17
+
 require (
 	github.com/alecthomas/units v0.0.0-20211218093645-b94a6e3cc137
 	github.com/aws/aws-sdk-go v1.43.11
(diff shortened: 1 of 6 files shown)


SuperQ:
It looks like MailDev completely dropped STARTLS. (https://github.com/maildev/maildev/issues/274)

go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/prometheus/alertmanager
 
+go 1.17
```
gotjosh, line 3: shouldn't this be 1.18? 🤔

gotjosh (commented):
LGTM, but please see my comment re the go version.

joeblubaugh:
This PR also included a functional change to the email notifier, using x/text to do casing, and x/text panics on some inputs. Here's a stack trace we've seen while running the mimir alertmanager:
```sh
/vendor/gopkg.in/yaml.v2/decode.go:270 +0xa7
gopkg.in/yaml%2ev2.(*decoder).callUnmarshaler(0xc20be3f980, 0xc20c116e70, {0x7f170dda42b8, 0xc0430126c0})
/vendor/github.com/prometheus/alertmanager/config/notifiers.go:196 +0x1b9
github.com/prometheus/alertmanager/config.(*EmailConfig).UnmarshalYAML(0xc0430126c0, 0x2340b20?)
	/vendor/golang.org/x/text/cases/cases.go:51
golang.org/x/text/cases.Caser.String(...)
	/vendor/golang.org/x/text/transform/transform.go:650 +0xbe5
golang.org/x/text/transform.String({0x7f170e30ea18, 0xc0002d2a00}, {0xc20fd3aeb8, 0x2})
...
runtime error: slice bounds out of range [11:7]
```

That wasn't mentioned in the commit, making this a very difficult to track regression.

I'm going to file an issue asking to revert the email notifier changes. I think they should be done in a separate PR, and I don't understand the problem that the casing call is solving.

I also think we can get away without doing Title casing - HTTP Header names are case *in*sensitive: https://www.rfc-editor.org/rfc/rfc7230#section-3.2 so the code could strings.ToLower() everything, which is a simpler transformation.

PR #3024, opened 2022-07-25
[UI] Silence creator from HTTP header or Basic Auth

m-masataka:
PR for following implementation
[Silence creator from HTTP header or Basic Auth](https://github.com/prometheus/alertmanager/pull/2971)


matzarah:
I would like to see this feature as we have a reverse proxy in front Alertmanager for authentication that could easily add a username header for Alertmanager to consume and prefill the author, or even enforce the author.

PR #2534, opened 2021-04-02
Add support to set the Slack URL in a file

julienduchesne:
- Added support for the file in both the global and the lower level
- Tried to follow configuration patterns I saw in prometheus
- The slack file is read on every request as mentioned in the prometheus issue to enable seamless switches

https://github.com/prometheus/alertmanager/issues/2498

PR #3038, opened 2022-08-15
SMTP config: add global and local password file fields

ericrrath:
Add config fields (for both global email config and route-specific email
config) that specify path to file containing SMTP password.  We don't
want the password in the config file itself, and reading the password
from a k8s-secret-backed file keeps the password itself "encrypted at
rest" in etcd, and cleanly separated from the rest of the AM config.

This is intended to partially address issue https://github.com/prometheus/alertmanager/issues/2498

I used the same approach as pull request https://github.com/prometheus/alertmanager/pull/2534


joywaves:
Can't wait to see this PR in place, thank you much @ericrrath

valvin1:
it looks this PR is what I need to avoid to do weird thing such as : 

* push an smtp password in plain in my config file on a git repository
* encrypt my whole alertmanager configuration file
* setup a local smtp to manage authentification

What this PR requires to do some progress?

config/notifiers.go
```diff
@@ -165,19 +165,20 @@ type EmailConfig struct {
...
-	Text         string              `yaml:"text,omitempty" json:"text,omitempty"`
-	RequireTLS   *bool               `yaml:"require_tls,omitempty" json:"require_tls,omitempty"`
-	TLSConfig    commoncfg.TLSConfig `yaml:"tls_config,omitempty" json:"tls_config,omitempty"`
+	To               string              `yaml:"to,omitempty" json:"to,omitempty"`
+	From             string              `yaml:"from,omitempty" json:"from,omitempty"`
+	Hello            string              `yaml:"hello,omitempty" json:"hello,omitempty"`
+	Smarthost        HostPort            `yaml:"smarthost,omitempty" json:"smarthost,omitempty"`
+	AuthUsername     string              `yaml:"auth_username,omitempty" json:"auth_username,omitempty"`
+	AuthPassword     Secret              `yaml:"auth_password,omitempty" json:"auth_password,omitempty"`
+	AuthPasswordFile Secret              `yaml:"auth_password_file,omitempty" json:"auth_password_file,omitempty"`
```
simonpasquier, line 174: We don't need the `*File` fields to be of type `Secret`, they can be plain string.

config/config.go
```diff
@@ -693,26 +702,27 @@ type GlobalConfig struct {
...
-	WeChatAPICorpID    string     `yaml:"wechat_api_corp_id,omitempty" json:"wechat_api_corp_id,omitempty"`
-	VictorOpsAPIURL    *URL       `yaml:"victorops_api_url,omitempty" json:"victorops_api_url,omitempty"`
-	VictorOpsAPIKey    Secret     `yaml:"victorops_api_key,omitempty" json:"victorops_api_key,omitempty"`
-	TelegramAPIUrl     *URL       `yaml:"telegram_api_url,omitempty" json:"telegram_api_url,omitempty"`
+	SMTPFrom             string     `yaml:"smtp_from,omitempty" json:"smtp_from,omitempty"`
+	SMTPHello            string     `yaml:"smtp_hello,omitempty" json:"smtp_hello,omitempty"`
+	SMTPSmarthost        HostPort   `yaml:"smtp_smarthost,omitempty" json:"smtp_smarthost,omitempty"`
+	SMTPAuthUsername     string     `yaml:"smtp_auth_username,omitempty" json:"smtp_auth_username,omitempty"`
+	SMTPAuthPassword     Secret     `yaml:"smtp_auth_password,omitempty" json:"smtp_auth_password,omitempty"`
+	SMTPAuthPasswordFile Secret     `yaml:"smtp_auth_password_file,omitempty" json:"smtp_auth_password_file,omitempty"`
```
simonpasquier, line 710: same here about the field's type.

config/config.go
```diff
@@ -365,8 +369,13 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 			if ec.AuthUsername == "" {
 				ec.AuthUsername = c.Global.SMTPAuthUsername
 			}
-			if ec.AuthPassword == "" {
+			// require a password only if a username is provided
```
simonpasquier, line 372: I don't see how this change relates to the original issue? Can you elaborate?

notify/email/email.go
```diff
@@ -353,3 +353,14 @@ func (a *loginAuth) Next(fromServer []byte, more bool) ([]byte, error) {
...
 	return nil, nil
 }
+
+func (n *Email) getPassword() string {
+	if len(n.conf.AuthPassword) > 0 {
+		return string(n.conf.AuthPassword)
+	}
+	content, err := os.ReadFile(string(n.conf.AuthPasswordFile))
+	if err != nil {
+		return ""
```
simonpasquier, line 363: we should return an error, rather than silently discarding the issue.

simonpasquier (commented):
thanks for doing this!

config/notifiers.go
ericrrath: No problem, fixed.

config/config.go
ericrrath: No problem, fixed.

notify/email/email.go
ericrrath: Good catch, fixed.  My first instinct was to have this new func treat both the "raw" and "file" fields being unset as an error, and return that error.  But the "upstream" code already has a check for that case, so I just had the new func return "" if both fields are unset.

config/config.go
ericrrath: Another good catch.  I thought username-but-no-password was an invalid config, but now see that it's [explicitly supported](https://github.com/prometheus/alertmanager/blob/v0.24.0/notify/email/email.go#L77).  Simplified.

ericrrath: Ugh, wait, sorry, I was rushing, and misread the code in that last part.  no-username is supported, but username-without-password still produces errors.  Hold on, let me get this sorted.  Sorry for the hassle.

simonpasquier (commented):
could you add sub-tests [here](https://github.com/prometheus/alertmanager/blob/3d624c0552e173aa57bbdad52a55788e38744252/notify/email/email_test.go#L419) that verify the happy (password file configured with good value) and sad (password file configured with wrong value) paths?

ericrrath:
> could you add sub-tests [here](https://github.com/prometheus/alertmanager/blob/3d624c0552e173aa57bbdad52a55788e38744252/notify/email/email_test.go#L419) that verify the happy (password file configured with good value) and sad (password file configured with wrong value) paths?

Another good catch; done!

ericrrath:
> Very cool! I'm being a bit nit-picky but could you also add a test that would configure an invalid file name and trigger an error when `getPassword()` gets called?

Done!

config/config.go
ericrrath: Okay, I finally remembered. Local configs inherit fields from the global config, and now that the SMTP password can come from two different kinds of field -- literal or from-file -- I wanted to avoid having a local config with one kind of field inheriting the other kind of field from global. I.e. if the global config has SMTPAuthPasswordFile set, and a local config has AuthPassword set, that local config should not inherit SMTPAuthPasswordFile.

I updated the code again to enforce that; local configs now only inherit the global SMTP password fields if both of the local fields are unset.

simonpasquier (commented):
thanks!

Issue #3012, opened 2022-07-19
Include receiver name in notification metrics


sinkingpoint:
Currently Alertmanager has metrics for various details around the behavior of receivers 

https://github.com/prometheus/alertmanager/blob/f30aef2c6990b9e8fd481f94bb5bd6293ddf589f/notify/notify.go#L256-L281

 These are broken down by an `integration` label, which indicates the type of receiver (`pagerduty`, `webhook` etc). 

The docs recommend extending Alertmanager with the `webhook` receiver (https://prometheus.io/docs/operating/integrations/#alertmanager-webhook-receiver) as opposed to adding new integrations (which I understand, it limits the maintenance burden of the main repo), but this means that a lot of receivers become the `webhook` integration. 

Because of the limited resolution of the receiver metrics, this means that it's impossible to tease different webhooks apart. For example, we have various webhooks that go to JIRA, Chat, and other internal systems. By just having the one metric, if one of those webhook receivers falls over, it becomes impossible from just the metrics to tell which one it was. 

I would like to propose a couple of solutions for this:

a) include the receiver name as a secondary label to the above metrics
b) include the hostname/port of the webhook either as another label, or replacing/extending the existing integration label

In order to allow more accurate alerting of receiver issues

sinkingpoint:
Also: I recognise that this would be a breaking change should we decide to go down this path - not sure how to handle it (or even if we should, considering alertmanager is still technically 0.x)

felipesere:
I would also be keen on this and happy to contribute a PR.

sinkingpoint:
@felipesere I have a patch at https://github.com/sinkingpoint/alertmanager/tree/sinkingpoint/receiver_name_metrics that I'm testing internally and then will PR if there's interest

felipesere:
How has that patch been working for you? 😄

sinkingpoint:
It's been working well for us. I'll PR it today

alimehrabikoshki:
@sinkingpoint I see the MR is still open, are you still happy to see that through to completion?

sinkingpoint:
@alimehrabikoshki I've just updated the PR

TheMeier:
Since this is available via feature-flag I consider this done.

PR #3045, opened 2022-08-22
Add the receiver name to notification metrics

sinkingpoint:
Solves #3012 

This PR adds in a second label to all the *Notification* metrics, so that we not only have the type of notification, but all the receiver_name that the notification belongs to. 

Because of the proliferation of webhook notifiers (as recommended by the docs), this is necessary to be able to properly diagnose failures from metrics - currently we get one metric for all web hooks, regardless of the fact that they might represent completely different notification pathways.

This is technically a breaking change, so this might have to wait for a bigger release, if its even worthwhile at all


roidelapluie:
Could we have this behind a cli flag?

Cli flag because I think we should issue it consistently over the lifetime of the app.

sinkingpoint:
Sorry for the hiatus. Will push a CLI flag today

sinkingpoint:
Fixed this up and it's ready for another review

alimehrabikoshki:
@roidelapluie could you pls take another look?

gotjosh:
Thank you very much for your contribution - you've been waiting for a long time on this so instead of dragging it back and forth, and went ahead and implemented a couple of suggestions I wanted on this PR.

The bulk of the change is around having the "feature flag management" on its package and applying a couple of patterns to keep them tidy for future additions.

The other big change is the fact that we can initialize ahead of time the metrics for the given receiver names. 

I'd love if @simonpasquier can take a look as I've written a big chunk of this code.

simonpasquier (commented):
My main remark is that old label values would be retained after a config reload in case some receivers are deleted/renamed. I guess that we need [DeleteLabelValues()](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#MetricVec.DeleteLabelValues) to prune them.

gotjosh:
> My main remark is that old label values would be retained after a config reload in case some receivers are deleted/renamed. I guess that we need [DeleteLabelValues()](https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#MetricVec.DeleteLabelValues) to prune them.

I can't use `https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#MetricVec.DeleteLabelValues` as I would need to know the old label values to prune them (and I can, but this would be a pretty invasive change). Instead, I chose to use `Reset()` to prune all of the metrics in the given vector and then recreate them again - given all of these are counters this shouldn't be a problem, as they would simply go back to 0 and would count as counter reset anyways.

simonpasquier (approved):
LGTM

gotjosh:
Thank you very much for your contribution @sinkingpoint.

mmikitka-arcticwolf:
@gotjosh Do you have an estimate for when the next official Alertmanager container image will be available that includes this change?

rtkwlf-nachiketrao:
@gotjosh Is there any update on when the new Alertmanager image will be available?

gotjosh:
Have you tested out the current rc.0 https://github.com/prometheus/alertmanager/releases/tag/v0.27.0-rc.0? I'm planning on promoting the the RC to 27.0 by the end of the week if nothing major is reported.

Issue #3050, opened 2022-08-30
Feature Request: Truncate messages with fewer characters.


joeblubaugh:
Alertmanager truncates messages using three full stops (U+002E): `...`. We could instead truncate with a single ellipsis (U+2026) `…`, freeing up two additional characters for the message contents.

https://github.com/prometheus/alertmanager/blob/c732372d7d3be49198398d34753080459f01749e/notify/util.go#L93

joeblubaugh:
(I'm happy to submit the PR)

yanksyoon:
Hi @joeblubaugh, I'm trying to familiarize with the codebase, would you mind if I work on it if this request gets approved? :)

joeblubaugh:
Sounds great!

simonpasquier:
sounds good to me

simonpasquier:
Closed by #3072

PR #3072, opened 2022-09-11
Truncate messages with fewer characters

yanksyoon:
Replace three dots characters `...` with ellipsis (U+2026) `…` characters to save 2 extra chars space.
Related issue: #3050

### DIFF
--- a/notify/util.go
+++ b/notify/util.go
@@ -90,7 +90,7 @@ func Truncate(s string, n int) (string, bool) {
 	if n <= 3 {
 		return string(r[:n]), true
 	}
-	return string(r[:n-3]) + "...", true
+	return string(r[:n-1]) + "…", true
 }
 
 // TmplText is using monadic error handling in order to make string templating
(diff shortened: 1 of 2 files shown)


simonpasquier (approved):
thanks!

notify/util.go
```diff
@@ -90,7 +90,7 @@ func Truncate(s string, n int) (string, bool) {
 	if n <= 3 {
 		return string(r[:n]), true
 	}
-	return string(r[:n-3]) + "...", true
+	return string(r[:n-1]) + "…", true
```
pstibrany, line 93: "…" is encoded as 3 bytes in utf-8, so this doesn't save anything, and makes string longer by 2 bytes compared to previous version.

Furthermore blindly cutting last 3 or 1 byte can actually break the utf-8 sequence of bytes. Ideally string would be truncated at rune boundary.

gotjosh: I fixed this eventually: https://github.com/prometheus/alertmanager/blob/main/notify/util.go#L88-L130

pstibrany: Oh nice, I missed that when checking changelog of alertmanager 0.26.0.

pstibrany: Thanks!

Issue #3037, opened 2022-08-11
Config reload causes notifications of new alerts within a group (doesn't respect group_interval)


andrejbranch:
**What did you do?**
- Setup alertmanager and add group by alertname and long group_interval 
- Send an alert and get notified
- Send another alert with the same alertname but different other labels 
- No alert is sent due to group_interval not being hit
- Send a post reload request to alertmanager
- Another notification is sent

**What did you expect to see?**
Group interval should be respected even when config is reloaded

**What did you see instead? Under which circumstances?**
Group interval is not respected when reloading config

* Alertmanager version:
v0.24.0

This seems to be caused because reload restarts the alerting group and restarts the group interval here https://github.com/prometheus/alertmanager/blob/main/dispatch/dispatch.go#L434

And if there are any changes in the group subset this returns true https://github.com/prometheus/alertmanager/blob/main/notify/notify.go#L570 causing a notification

zecke:
Sounds like a dup of #2492

andrejbranch:
yes it does thanks

Issue #2492, opened 2021-02-25
emails are resent on config reload if group contents has changed


asuffield:
**What did you do?**

Write a route with a group match, of the form:
```
group_by:
- alertname
group_wait: 10s
group_interval: 1h
repeat_interval: 4h
```

Perform these operations, in order:
```
amtool alert add alertname=test tag=1
sleep 11
# observe the notification being sent
amtool alert add alertname=test tag=2
killall -HUP alertmanager
```

**What did you expect to see?**

One hour later, notifications should be resent, when group_interval expires.

**What did you see instead? Under which circumstances?**

Notifications for tag=1 and tag=2 are resent immediately when SIGHUP is received.

This is notably specific to the email receiver, because the other ones can deduplicate on group_key, which hides the problem.

**Environment**

* Alertmanager version:

0.21

So, here's what's going on:

Grouping is implemented in the dispatcher, while deduplication is implemented in the notification pipeline. The notification pipeline is stateful, via nflog. The dispatcher only keeps its state in memory.

Config reloading is implemented by completely stopping the dispatcher and constructing a new one from config. This also clears all aggregation group state.

When the config is reloaded, the new dispatcher recreates all the aggregation groups from the in-memory alert state. This results in dispatching all the events into the notification pipeline again.

If the set of alerts is unchanged since the last time this occurred, the DedupStage in the notification pipeline will suppress them, based on the contents of nflog, respecting repeat_interval as expected. However, if there has been any change to the set of alerts in the group, then DedupStage can't handle that and the notifications will be sent to the receiver.

If the receiver is anything other than email, the group key and timestamps will not have changed so the duplicate notifications can be suppressed. The email receiver is unable to do this (because it's email...) so duplicate emails are sent.

Half of this story is taken up by explaining why we don't see this happen all the time, but the grouping functionality doesn't really work across config reloads, and probably also behaves badly across cluster gossip. This isn't noticed in most configurations because the other features are effective at hiding the problem.

I'm thinking that the correct solution here would be for aggregation grouping to be implemented in the notification pipeline based on nflog. However, a crude and immediate solution is likely achievable by having the dispatcher query nflog when creating an aggregation group, and adjusting its timings if this group key was successfully sent in the last group_interval; this would solve the config reload case, but I'm not sure it's the right way to go in the presence of cluster gossip.

Opinions?

zecke:
[...]

PR #3074, opened 2022-09-14
Fix config reloads not respecting group interval

andrejbranch:
We have been running this modification at Shopify for the last few weeks and it has solved two issues for us.

1. Config reload causes notifications of new alerts within a group (doesn't respect group_interval) 
2. When all alerts in a group are resolved, resolved notifications are sent quickly instead of waiting for group interval 

Related issues that this addresses:
- https://github.com/prometheus/alertmanager/issues/2492
- https://github.com/prometheus/alertmanager/issues/1660
- https://github.com/prometheus/alertmanager/issues/3037

This change does the following:
- Changes an alerts group ticker interval to group wait (theres a TODO here), this results in notification evaluation to happen every 30s (default group wait) for each alert group.
- Moves group interval logic to dedupe stage utilizing the persistent notification entries. If the last notification entry for an alert group was a firing event and it was sent within the group interval window, we prevent notification.

TODO:
This change adds to even further the confusion of group interval, and group_wait. Three separate configs seems more appropriate: group_wait (how long we wait to start the evaluation background job for an alert group), group_interval (how frequent an alert group is evaluated for dispatch), group_repeat_interval (how long to wait before sending a notification about new alerts that are added to a group of alerts. This would allow for more fine tuning alert dispatching.


notify/notify.go
```diff
@@ -560,14 +573,16 @@ func hashAlert(a *types.Alert) uint64 {
...
-func (n *DedupStage) needsUpdate(entry *nflogpb.Entry, firing, resolved map[uint64]struct{}, repeat time.Duration) bool {
+func (n *DedupStage) needsUpdate(entry *nflogpb.Entry, firing, resolved map[uint64]struct{}, repeat, groupInterval time.Duration) bool {
 	// If we haven't notified about the alert group before, notify right away
 	// unless we only have resolved alerts.
 	if entry == nil {
 		return len(firing) > 0
 	}
 
-	if !entry.IsFiringSubset(firing) {
+	groupIntervalMuted := len(entry.FiringAlerts) > 0 && entry.Timestamp.After(time.Now().Add(-groupInterval))
```
zecke, line 583: Any chance we want to use keyNow from the context here?

simonpasquier:
> When all alerts in a group are resolved, resolved notifications have to wait till group interval is complete

This is somehow a feature because sending resolved notifications immediately might exacerbate flapping notifications.

yuri-tceretian:
In https://github.com/prometheus/alertmanager/pull/3283 I try to solve another problem a similar way. There are some differences though:
- my PR drops the positive result of `needsUpdate` if the tick expires.
- I do not use group_interval and a `magic` constant 0.99. Why did you choose 0.99 and not 1.0 or 0.95?

andrejbranch:
I have modified this quite a bit in Shopify's alertmanager fork, i do not use 0.99 constant anymore and I also had to solve alert flapping with hard coded check if its a recent entry. 

for what its worth here is the needsUpdate in our forked version
```
func (n *DedupStage) needsUpdate(entry *nflogpb.Entry, firing, resolved map[uint64]struct{}, repeat, groupInterval time.Duration) bool {
	if entry == nil {
		return len(firing) > 0
	}

	if recentEntry := entry.Timestamp.After(time.Now().Add(-30 * time.Second)); recentEntry {
		return false
	}

	groupIntervalMuted := len(entry.FiringAlerts) > 0 && entry.Timestamp.After(time.Now().Add(-groupInterval))

	if !entry.IsFiringSubset(firing) && !groupIntervalMuted {
		return true
	}

	// Notify about all alerts being resolved.
	// This is done irrespective of the send_resolved flag to make sure that
	// the firing alerts are cleared from the notification log.
	if len(firing) == 0 {
		// If the current alert group and last notification contain no firing
		// alert, it means that some alerts have been fired and resolved during the
		// last interval. In this case, there is no need to notify the receiver
		// since it doesn't know about them.
		return len(entry.FiringAlerts) > 0
	}

	if n.rs.SendResolved() && !entry.IsResolvedSubset(resolved) && !groupIntervalMuted {
		return true
	}

	// Nothing changed, only notify if the repeat interval has passed.
	return entry.Timestamp.Before(n.now().Add(-repeat))
}
```

grobinson-grafana:
Hi! 👋 I found this issue from another PR, just adding 

> Config reload causes notifications of new alerts within a group (doesn't respect group_interval)

Config reload will re-create all existing aggregation groups. This causes an immediate flush because of this code [here](https://github.com/prometheus/alertmanager/blob/main/dispatch/dispatch.go#L486):

```go
// Immediately trigger a flush if the wait duration for this
// alert is already over.
ag.mtx.Lock()
defer ag.mtx.Unlock()
if !ag.hasFlushed && alert.StartsAt.Add(ag.opts.GroupWait).Before(time.Now()) {
	ag.next.Reset(0)
}
```

I think we should wait Group wait seconds. I'm not sure if we need to persist and wait the remainder of the previous Group interval?

> When all alerts in a group are resolved, resolved notifications have to wait till group interval is complete.

Like Simon said, I think this is intended behavior. The problem is while a group might contain all resolved alerts at time `x`, that might not be the case at the next group interval. If you have a group of size 1 (just a single alert) that is flapping between firing and resolved, won't this change make Alertmanager send a notification each time it flaps, rather than smooth it out?

grobinson-grafana:
Hi! 👋 I found this issue from another PR, just adding 

> Config reload causes notifications of new alerts within a group (doesn't respect group_interval)

Config reload will re-create all existing aggregation groups. This causes an immediate flush because of this code [here](https://github.com/prometheus/alertmanager/blob/main/dispatch/dispatch.go#L486):

```go
// Immediately trigger a flush if the wait duration for this
// alert is already over.
ag.mtx.Lock()
defer ag.mtx.Unlock()
if !ag.hasFlushed && alert.StartsAt.Add(ag.opts.GroupWait).Before(time.Now()) {
	ag.next.Reset(0)
}
```

I think we should wait Group wait seconds. I'm not sure if we need to persist and wait the remainder of the previous Group interval?

> When all alerts in a group are resolved, resolved notifications have to wait till group interval is complete.

Like Simon said, I think this is intended behavior. If you have a group of size 1 (i.e. grouping is disabled) that is flapping between firing and resolved, won't this change make Alertmanager send a notification each time it flaps, rather than smooth it out?

grobinson-grafana:
Hi! 👋 I found this issue from another PR.

> Config reload causes notifications of new alerts within a group (doesn't respect group_interval)

Config reload will re-create all existing aggregation groups. This causes an immediate flush because of this code [here](https://github.com/prometheus/alertmanager/blob/main/dispatch/dispatch.go#L486):

```go
// Immediately trigger a flush if the wait duration for this
// alert is already over.
ag.mtx.Lock()
defer ag.mtx.Unlock()
if !ag.hasFlushed && alert.StartsAt.Add(ag.opts.GroupWait).Before(time.Now()) {
	ag.next.Reset(0)
}
```

I think we should wait Group wait seconds. I'm not sure if we need to persist and wait the remainder of the previous Group interval?

> When all alerts in a group are resolved, resolved notifications have to wait till group interval is complete.

Like Simon said, I think this is intended behavior. If you have a group of size 1 (i.e. grouping is disabled) that is flapping between firing and resolved, won't this change make Alertmanager send a notification each time it flaps, rather than smooth it out?

PR #3078, opened 2022-09-16
notify/email: refactor tests

simonpasquier:
We don't need to test `getPassword()` separately as it can be covered with the tests already checking the various authentication methods.

Follow-up of #3038


PhilipGough (approved):
lgtm

gotjosh (approved):
LGTM

PR #3080, opened 2022-09-21
EmailConfig: Use CanonicalMIMEHeaderKey instead of TitleCasing for headers

### LINKED ISSUE #3064: Email Notifier panics when unmarshalling headers from email config
**What did you do?**
Updated the github.com/grafana/mimir dependency on Prometheus to v0.24.1-0.20220805150254-c732372d7d3b

Unmarshalled some email notifier configuration yaml that unmarshalled successfully in v0.24.0

The configuration would continue to unmarshal successfully.

A panic while title-casing HTTP header values.

**Environment**

* System information:

* Alertmanager version:
	v0.24.1-0.20220805150254-c732372d7d3b

This is a tenant-private configuration. We will update this issue with a reproducible string once we find one.

* Logs:
```
/vendor/gopkg.in/yaml.v2/decode.go:270 +0xa7
gopkg.in/yaml%2ev2.(*decoder).callUnmarshaler(0xc20be3f980, 0xc20c116e70, {0x7f170dda42b8, 0xc0430126c0})
/vendor/github.com/prometheus/alertmanager/config/notifiers.go:196 +0x1b9
github.com/prometheus/alertmanager/config.(*EmailConfig).UnmarshalYAML(0xc0430126c0, 0x2340b20?)
	/vendor/golang.org/x/text/cases/cases.go:51
golang.org/x/text/cases.Caser.String(...)
	/vendor/golang.org/x/text/transform/transform.go:650 +0xbe5
golang.org/x/text/transform.String({0x7f170e30ea18, 0xc0002d2a00}, {0xc20fd3aeb8, 0x2})
...
runtime error: slice bounds out of range [11:7]
```

This was introduced in #3017, where Title casing switched from `strings.Title` (deprecated) to `golang.org/x/text/case` title casing. 

One fix would be to revert to string.Title, which does not panic on these inputs, and filing a bug with x/text/case when we can determine the input that panics.

Another possible way of addressing this issue is to avoid title casing of header names. Header names are case insensitive: https://www.rfc-editor.org/rfc/rfc7230#section-3.2, so lower-casing every header with `strings.ToLower()` should be sufficient to check for collisions.

joeblubaugh:
Fixes #3064.


simonpasquier (approved):
thanks!

joeblubaugh:
@simonpasquier how and when do commits get tagged as releases? Is there's a process that we can help with? I'd like to import this change into some downstream projects.

simonpasquier:
@joeblubaugh we have no fixed schedule for releases but I'm planning on preparing a 0.25 release candidate soon-ish.

joeblubaugh:
Great, thank you.

PR #3092, opened 2022-10-05
Reduce memory utilization in downstream projects creating multiple Alertmanager instances

pracucci:
Grafana Mimir vendors the Prometheus Alertmanager. Specifically, Grafana Mimir (which is a multi-tenant system) creates a Prometheus Alertmanager instance for each tenant. This means that, in a given process, there could be hundreds or thousands of Prometheus Alertmanager instances (one for each tenant).

While investigating some high memory utilization, I noticed that in some clusters the memory allocated by `loads.Analyzed()` (swagger spec) takes about 50% of memory.

Screenshot 2022-10-05 at 14 03 45

In this PR I'm proposing to lazy load and then cache the swagger spec. To my understanding, the swagger spec is immutable, so shouldn't be an issue sharing it between different Aletmanager instances.

WDYT?

### DIFF
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -694,7 +694,7 @@ func getSwaggerSpec() (*loads.Document, error) {
 	// Load embedded swagger file.
 	swaggerSpec, err := loads.Analyzed(restapi.SwaggerJSON, "")
 	if err != nil {
-		return nil, fmt.Errorf("failed to load embedded swagger file: %v", err.Error())
+		return nil, fmt.Errorf("failed to load embedded swagger file: %w", err)
 	}
 
 	swaggerSpecCache = swaggerSpec
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -101,9 +101,9 @@ func NewAPI(
 	}
 
 	// Load embedded swagger file.
-	swaggerSpec, err := loads.Analyzed(restapi.SwaggerJSON, "")
+	swaggerSpec, err := getSwaggerSpec()
 	if err != nil {
-		return nil, fmt.Errorf("failed to load embedded swagger file: %v", err.Error())
+		return nil, err
 	}
 
 	// Create new service API.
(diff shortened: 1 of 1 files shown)


api/v2/api.go
```diff
@@ -672,3 +672,31 @@ func parseFilter(filter []string) ([]*labels.Matcher, error) {
 	}
 	return matchers, nil
 }
+
+var (
+	swaggerSpecCacheMx sync.Mutex
```
grobinson-grafana, line 677: We should be able to make this a `RWMutex`, although I'm not sure if we would get much out of it?

api/v2/api.go
```diff
@@ -672,3 +672,31 @@ func parseFilter(filter []string) ([]*labels.Matcher, error) {
...
+
+	// Check if a cached version exists.
+	if swaggerSpecCache != nil {
+		return swaggerSpecCache, nil
+	}
+
+	// Load embedded swagger file.
+	swaggerSpec, err := loads.Analyzed(restapi.SwaggerJSON, "")
+	if err != nil {
+		return nil, fmt.Errorf("failed to load embedded swagger file: %v", err.Error())
```
grobinson-grafana, line 697: Should we not keep the original `err`?

```suggestion
		return nil, fmt.Errorf("failed to load embedded swagger file: %v", err)
```

grobinson-grafana (commented):
Looks like a great improvement for the Mimir Alertmanager. I presume the specification is just read and not modified?

api/v2/api.go
pracucci: I just kept the previous error. If we want to switch it to error wrapping, then it should be:

```suggestion
		return nil, fmt.Errorf("failed to load embedded swagger file: %w", err)
```

(notice the `%w`).

api/v2/api.go
pracucci: I don't think the extra complexity of handling a RWMutex (first read lock to check if it's cached, then take write lock to cache it) is worth. We don't expect this lock to be hit very frequently (just once in Prometheus Alertmanager).

api/v2/api.go
grobinson-grafana: Perhaps the original code was written before `%w` was added to `fmt`. I think it would be nice to change it to `%w`, unless there is a good reason against it! :slightly_smiling_face:

simonpasquier (approved):
lgtm

api/v2/api.go
pracucci: Agree, let me do it.

PR #3093, opened 2022-10-06
Allow first day of week to be Sunday or Monday

Git-Jiro:
refers to 3073


Git-Jiro:
@simonpasquier 
who can give me feedback on this implementation of "choose week to start at monday or sunday"

simonpasquier:
@Git-Jiro thanks for doing this! I'm not really qualified for reviewing the code, @w0rm is definitely more qualified :)

For the other reviewers, this PR introduces a new "Settings" tab which allows to select in a drop-down list the first day of the week (Monday or Sunday):

image

From a feature standpoint, I wonder if the list shouldn't contain all week days?
Some padding between "Start of week:" and the drop-down list would be neat. It might also be good to mention in the Settings page that preferences are stored in the client's local storage (hence not shared between different clients).

Git-Jiro:
@simonpasquier  good points!
regarding "put all the weekdays in the dropdown": That might imply that setting the other days as "first day of week" works as well. Unfortunately, "first day of week" only works for Monday and Sunday at the moment.

ui/app/src/Views/Settings/Views.elm
```diff
@@ -0,0 +1,37 @@
...
+                    , selected
+                        (model.startOfWeek == 7)
+                    ]
+                    [ text "Sunday" ]
+                ]
+            , small [ class "form-text text-muted" ]
+                [ text "Note: This setting is saved in local storage of your browser"
+                ]
+            ]
+        ]
```
w0rm, line 37: Can you share the screenshot once you fix the spacing issue reported by @simonpasquier ? I would probably check the layout of the silence form for an inspiration.

ui/app/src/Main.elm
```diff
@@ -86,12 +86,24 @@ init flags url key =
 
             else
                 "/"
+
+        startWeekAtMonday =
+            flags
+                |> Json.decodeValue (Json.field "startWeekAtMonday" Json.bool)
+                |> Result.withDefault False
```
w0rm, line 93: Perhaps we should reverse the logic and make Monday the default if we fail to decode.

ui/app/index.html
```diff
@@ -20,6 +20,7 @@
             var app = Elm.Main.init({
                 flags: {
                     production: true,
+                    startWeekAtMonday: JSON.parse(localStorage.getItem('startWeekAtMonday')),
```
w0rm, line 23: Can we store `startWeekOnSunday` instead, to make sure Monday is the default?

ui/app/src/Views/Settings/Types.elm
```diff
@@ -0,0 +1,10 @@
+module Views.Settings.Types exposing (..)
+
+
+type alias Model =
+    { startOfWeek : Int
```
w0rm, line 5: Can we call this the same name as elsewhere, `firstDayOfWeek`, and use the custom type instead of the int?

w0rm:
We shouldn't put all days in the dropdown, it's Monday or Sunday. Perhaps we could switch to the radio buttons, so that the user doesn't have to click twice to select the option?

ui/app/src/Views/Settings/Views.elm
Git-Jiro: Current Layout of Settings Page

Screenshot of current layout of settings page

ui/app/index.html
Git-Jiro: Since the variable is now called "firstDayOfWeek" everywhere, I also refactored the flag to be called "firstDayOfWeek".
I hope this is fine as well.

ui/app/src/Utils/DateTimePicker/Views.elm
```diff
@@ -80,14 +81,19 @@ viewMonth : DateTimePicker -> Posix -> Html Msg
...
-            listDaysOfMonth justViewTime
+            listDaysOfMonth justViewTime dateTimePicker.firstDayOfWeek
 
         weeks =
             splitWeek days []
     in
     div [ class "row justify-content-center" ]
         [ div [ class "weekheader" ]
-            (List.map viewWeekHeader [ "Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat" ])
+            (if dateTimePicker.firstDayOfWeek == Sunday then
```
w0rm, line 91: Can we use `case of` here and other places where we check the firstDayOfWeek value. This way it would be clear that the only two options are Sunday and Monday.

Git-Jiro:
For me it looks like the current test failure is a false positive, is it possible to restart the test?

ui/app/src/Views/Settings/Views.elm
w0rm: I think it would look much nicer as two radio buttons:

🔘 Monday 🔘 Sunday

But if that's cumbersome, could you make the dropdown shorter?

ui/app/src/Utils/DateTimePicker/Utils.elm
```diff
@@ -33,17 +39,38 @@ listDaysOfMonth time =
...
                 |> (\wd ->
-                        if wd == 7 then
-                            0
+                        case firstDayOfWeek of
+                            Sunday ->
+                                if wd == 7 then
+                                    0
+
+                                else
+                                    wd
```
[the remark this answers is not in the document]
w0rm: Nevermind! I misunderstood the previous implementation

w0rm:
@Git-Jiro thanks for the hard work! I have one last suggestion for the design. Let me know what you think. I am also happy to do it myself if you don't have time.

ui/app/src/Main.elm
```diff
@@ -86,12 +87,25 @@ init flags url key =
 
             else
                 "/"
+
+        firstDayOfWeek =
+            flags
+                |> Json.decodeValue (Json.field "firstDayOfWeek" Json.string)
+                |> Result.withDefault "Monday"
```
w0rm, line 94: I'm sorry, I misunderstood the original issue, that `Sunday` indeed used to be the default.

Git-Jiro: no problem, since changing the default is now a single line change.

Git-Jiro:
> @Git-Jiro thanks for the hard work! I have one last suggestion for the design. Let me know what you think. I am also happy to do it myself if you don't have time.

I guess we are talking about radio buttons instead of the dropdown?
should not be that hard.

Git-Jiro:
Now with radio buttons:
Radio Buttons

ui/app/src/Views/Settings/Views.elm
```diff
@@ -0,0 +1,35 @@
...
+
+
+radio : String -> Bool -> (String -> msg) -> Html msg
+radio radioValue isChecked msg =
+    div [ class "form-check" ]
+        [ input [ type_ "radio", onInput msg, class "form-check-input", id radioValue, checked isChecked, value radioValue ] []
+        , label [ class "form-check-label", for radioValue ]
+            [ text radioValue
+            ]
+        ]
```
w0rm, line 35:
```suggestion
    label [ class "mt-1 ml-1 custom-control custom-radio" ]
        [ input
            [ type_ "checkbox"
            , class "custom-control-input"
            , checked isChecked
            , value radioValue
            , onInput msg
            ]
            []
        , span [ class "custom-control-indicator" ] []
        , span [ class "custom-control-description" ] [ text radioValue ]
        ]
```

Should look a bit nicer! 

Screenshot 2022-10-18 at 18 36 09

w0rm:
@Git-Jiro looks like `assets_vfsdata.go` needs to be regenerated. And then it is good to go! 🙏

PR #3094, opened 2022-10-07
add status code label to the numTotalFailedNotifications metric

qinxx108:
This is to follow up the discussion on issue:  https://github.com/prometheus/alertmanager/pull/2927

Instead of creating a new metric, we will add the status code label to the numTotalFailedNotifications metric. The default status code is going to be "5xx" and if the integration receivers defined the NewErrorWithStatusCode, we will translate the returned status code and put that in the label override


krishnateja325 (approved):

notify/notify.go
```diff
@@ -662,9 +665,17 @@ func NewRetryStage(i Integration, groupName string, metrics *Metrics) *RetryStag
...
 	if err != nil {
-		r.metrics.numTotalFailedNotifications.WithLabelValues(r.integration.Name()).Inc()
+		if e, ok := errors.Cause(err).(*ErrorWithStatusCode); ok {
+			result, interErr := getFailureStatusCodeCategory(e.StatusCode)
+			if interErr == nil {
+				statusCodeCategory = result
+			}
+		}
 	}
+	r.metrics.numTotalFailedNotifications.WithLabelValues(r.integration.Name(), statusCodeCategory).Inc()
```
rohang98, line 678: Shouldn't this increment still be inside the if statement?

qinxx108: Fixed

rohang98 (approved):

notify/notify.go
```diff
@@ -262,7 +265,7 @@ func NewMetrics(r prometheus.Registerer) *Metrics {
 			Namespace: "alertmanager",
 			Name:      "notifications_failed_total",
 			Help:      "The total number of failed notifications.",
-		}, []string{"integration"}),
+		}, []string{"integration", "statusCode"}),
```
bwplotka, line 268: Let's be concise. Perhaps `code` makes more sense?

notify/util.go
```diff
@@ -167,6 +167,20 @@ func readAll(r io.Reader) string {
...
+// other than 4xx and 5xx input status will return an error.
+func getFailureStatusCodeCategory(statusCode int) (string, error) {
+	if statusCode/100 == 4 {
+		return "4xx", nil
+	}
+	if statusCode/100 == 5 {
+		return "5xx", nil
+	}
+
+	return "", fmt.Errorf("unexpected status code %v", statusCode)
```
bwplotka, line 181: Do we really want to fail some application logic if status code is unknown? I would just return `unknown` string here.

notify/notify.go
```diff
@@ -293,7 +296,7 @@ func NewMetrics(r prometheus.Registerer) *Metrics {
 		"telegram",
 	} {
 		m.numNotifications.WithLabelValues(integration)
-		m.numTotalFailedNotifications.WithLabelValues(integration)
+		m.numTotalFailedNotifications.WithLabelValues(integration, "")
```
bwplotka, line 299: This does not make a lot of sense... I think we want to populate here counters with each possible 3 status codes: `4xx`, `5xx` and unknown.

bwplotka (approved):
Some ideas to improve this code, but generally looks amazing. Thanks for coming to Prometheus ContribFest and this contribution!

notify/notify.go
qinxx108: done

notify/util.go
```diff
@@ -167,6 +176,20 @@ func readAll(r io.Reader) string {
 	return string(bs)
 }
 
+// getFailureStatusCodeCategory return the status code category for failure request
+// the status starts with 4 will return 4xx and starts with 5 will return 5xx
+// other than 4xx and 5xx input status will return an error.
+func getFailureStatusCodeCategory(statusCode int) (string, error) {
```
bwplotka, line 182: Based on how we use this, I feel we should not return error here, but just return default value, WDYT? Would make more clean and simple.

bwplotka (commented):
LGTM just one nit (:

notify/util.go
qinxx108: I kind of wanting to return an error so that the user of the function has knowledge about if the status code has been successfully translated. Otherwise they will have to rely on comparing if the result = failureUnknownCategoryCode if failureUnknownCategoryCode is not the ideal result for them in error case. WDYT?

notify/util.go
```diff
@@ -30,9 +30,18 @@ import (
...
+	failure4xxCategoryCode     = "4xx"
+	failure5xxCategoryCode     = "5xx"
+	failureUnknownCategoryCode = "unknown"
+)
+
 // UserAgentHeader is the default User-Agent for notification requests
 var UserAgentHeader = fmt.Sprintf("Alertmanager/%s", version.Version)
 
+// PossibleFailureStatusCategory is a list of possible failure status code category
+var PossibleFailureStatusCategory = []string{failure4xxCategoryCode, failure5xxCategoryCode, failureUnknownCategoryCode}
```
bwplotka, line 43: Why it's exported? Can we make it private and move it next to `notify.go`?

notify/util.go
bwplotka: Hm, well there are no users of this function yet, and as YAGNI goes, we should assume there will be none. I would focus on our usage:

```
result, interErr := getFailureStatusCodeCategory(e.StatusCode)
			if interErr == nil {
				statusCodeCategory = result
			}
```

We don't want to know IF the status was translated correctly, but rather we want `defaultStatusCodeCategory` in this case. So I would just return `defaultStatusCodeCategory` 

Also, now when we talk about it, `util` is not a great place for it - it's pretty generic - we want it in util if we have more than one usage of it. Given the function is very small and used only in one place, why not:

1. Moving this function next to `notify.go` file? 
OR
2. Copy the content of this function to caller place? It's might be too shallow for a function, but up to you (:

bwplotka (commented):
Last nits, thanks!

notify/notify.go
```diff
@@ -662,8 +668,16 @@ func NewRetryStage(i Integration, groupName string, metrics *Metrics) *RetryStag
 func (r RetryStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.Alert) (context.Context, []*types.Alert, error) {
 	r.metrics.numNotifications.WithLabelValues(r.integration.Name()).Inc()
 	ctx, alerts, err := r.exec(ctx, l, alerts...)
+
+	statusCodeCategory := defaultStatusCodeCategory
```
simonpasquier, line 672: For integrations that don't implement `ErrorWithStatusCode`, this could be inaccurate (e.g. the receiving side might have returned a 400 status code). It's also not correct if Alertmanager couldn't even send the notification (template error for instance). And the email notifier is another case that doesn't fit here.

I wonder if we shouldn't abstract away from the HTTP situation and consider a more generic label like `reason` with a handful number of possible values: 
* `server error` (e.g. 5xx)
* `client error` (e.g. 4xx)
* `authentication error`
* `template`
* `other`

jeromeinsf: @simonpasquier would this work?

simonpasquier: I think that we can start with:
* ClientError (typically templating errors, 4xx responses)
* ServerError (typically 5xx responses from the server)
* Other (e.g. receivers for which we haven't implemented ClientError/ServerError logic yet)

And we can always add more in the future where we see fit (e.g. maybe we'll want to distinguish authn failures, serialization errors, templating errors, ...)

qinxx108: done

notify/util.go
```diff
@@ -245,3 +245,60 @@ func (r *Retrier) Check(statusCode int, body io.Reader) (bool, error) {
...
+		Err:    err,
+		Reason: reason,
+	}
+}
+
+func (e *ErrorWithReason) Error() string {
+	return e.Err.Error()
+}
+
+// Reason is the failure reason
```
gotjosh, line 267:
```suggestion
// Reason is the failure reason.
```

notify/util.go
```diff
@@ -245,3 +245,60 @@ func (r *Retrier) Check(statusCode int, body io.Reader) (bool, error) {
...
+		panic(fmt.Sprintf("unknown Reason: %d", s))
+	}
+}
+
+// possibleFailureReasonCategory is a list of possible failure reason
+var possibleFailureReasonCategory = []string{DefaultReason.String(), ClientErrorReason.String(), ServerErrorReason.String()}
+
+// GetFailureReasonFromStatusCode return the reason for failure request
+// the status starts with 4 will return 4xx and starts with 5 will return 5xx
+// other than 4xx and 5xx input status will return an 5xx.
```
gotjosh, line 294:
```suggestion
// GetFailureReasonFromStatusCode returns the reason for the failure based on the status code provided.
```

gotjosh: Not sure if this method has changed its signature as of lately, but the comment is no longer accurate - what do you think about this suggestion?

notify/util.go
```diff
@@ -245,3 +245,60 @@ func (r *Retrier) Check(statusCode int, body io.Reader) (bool, error) {
 	}
 	return retry, errors.New(s)
 }
+
+type ErrorWithReason struct {
+	Err error
+
+	// The reason of the failure.
```
gotjosh, line 252:
```suggestion
```

This seems redundant - I'd consider removing it.

notify/notify.go
```diff
@@ -662,8 +665,13 @@ func NewRetryStage(i Integration, groupName string, metrics *Metrics) *RetryStag
 func (r RetryStage) Exec(ctx context.Context, l log.Logger, alerts ...*types.Alert) (context.Context, []*types.Alert, error) {
 	r.metrics.numNotifications.WithLabelValues(r.integration.Name()).Inc()
 	ctx, alerts, err := r.exec(ctx, l, alerts...)
+
+	failureReason := DefaultReason.String()
```
gotjosh, line 669: I've filed https://github.com/prometheus/alertmanager/issues/3231 so that we can do the same for all other notifiers.

gotjosh (approved):
LGTM!

All my comments are nits - but please let's address them, and we're ready to merge. Assuming that at this point, both @simonpasquier and @bwplotka's comments are resolved and they have nothing else to add.

notify/util.go
qinxx108: fixed

notify/util.go
qinxx108: fixed

notify/util.go
```diff
@@ -245,3 +245,60 @@ func (r *Retrier) Check(statusCode int, body io.Reader) (bool, error) {
...
+	case ClientErrorReason:
+		return "clientError"
+	case ServerErrorReason:
+		return "serverError"
+	default:
+		panic(fmt.Sprintf("unknown Reason: %d", s))
+	}
+}
+
+// possibleFailureReasonCategory is a list of possible failure reason
```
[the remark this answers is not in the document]
qinxx108: fixed

gotjosh:
Build is broken right now but tests pass, I'll assume everything is proper order here until we get that fixed that merge this.

gotjosh:
Thanks very much for your contribution!

PR #3100, opened 2022-10-10
Add PagerDuty source field (PD-CEF)

### LINKED ISSUE #3079: Feature Request: PagerDuty source field (PD-CEF)
**What did you do?**
I would like to configure the PagerDuty receiver to set the **Source** field of the event, as per [PD-CEF](https://support.pagerduty.com/docs/pd-cef). 

I expected to configure the `source` field independently of the `client` field.

The `source` field is not directly configurable, however the `client` configuration field may be used to set the Source ([code reference](https://github.com/prometheus/alertmanager/blob/efc956c7f79ef3cd02d52104fb997a2f69235f06/notify/pagerduty/pagerduty.go#L212-L229)).  However, the `client` field is also used to set the `client` parameter of the eventv2 message, and I want to have a different value for **Client** and **Source**.

Looking at the [PagerDuty API Reference](https://developer.pagerduty.com/api-reference/368ae3d938c9e-send-an-event-to-pager-duty), the Client and Source have different purposes:
- **Client** - The name of the monitoring client that is triggering this event (default: `Alertmanager`). 
- **Source** - The unique location of the affected system, preferably a hostname or FQDN.

**Suggested Improvement**
My suggestion would be to:
1. Introduce a new `source` configuration field (templated). 
2. Use `client` as the default value for `source`, for backwards compatibility.

* Alertmanager version: v0.24.0

```
receivers:
  - name: example
    pagerduty_configs:
    - routing_key:  ...
      client: Alertmanager
      source: "{{ .CommonLabels.hostname }}"
```

OktarianTB:
Fix #3079 by adding a new `source` field to the PagerDuty config. Maintains backward compatibility by defaulting to the `client` field. Added some tests to ensure behaviour works as expected now and in the future. Updated docs as well.

*this is my first ever open source contribution! #hacktoberfest*


notify/pagerduty/pagerduty.go
```diff
@@ -209,6 +209,11 @@ func (n *Notifier) notifyV2(
 		level.Debug(n.logger).Log("msg", "Truncated summary", "summary", summary, "key", key)
 	}
 
+	payloadSource := n.conf.Client
+	if n.conf.Source != "" {
+		payloadSource = n.conf.Source
+	}
```
simonpasquier, line 215: The default value could be handled in config/notifiers.go when unmarshaling the YAML data. This would be more explicit when users look at the /api/v2/status endpoint.

simonpasquier (commented):
thanks! @EronWright since you've filed the original issue, does it look like the right approach for you?

notify/pagerduty/pagerduty.go
OktarianTB: Updated 👍

simonpasquier:
Can you sign your commits? the easiest solution is probably to squash all the commits:

```
git rebase -i main
git amend -s
```

OktarianTB:
I opened a new PR with the signing properly setup: https://github.com/prometheus/alertmanager/pull/3106 - sorry!

simonpasquier:
No problem!

PR #3102, opened 2022-10-11
bugfix/3101: skip one silence rule instead of skipping all silence ru…

bo-er:
…les in face of error.


SoloJacobs:
Hi,

thanks for the contribution. The correct way to solve this problem would be to reject the silence before it can be added to the silence store. I had a look at the `POST /api/v1/silences` and it appears to already handle that case correctly already. If you can add a reproduction of the issue, I'll be happy to review it.

Kind regards

PR #3103, opened 2022-10-11
Log Response Details for Notifier errors for Webhooks, Pushover and VictorOps

### LINKED ISSUE #3057: Webhook Notifier should log response details in case of an error
Hi,
I'm using the webhook configuration to send alerts to some webhook destination.

This destination requires some additional fields.
If there are missing mandatory fields, an 400 error is returned, with details about the missing fields.

I expected to see the error details in the logs, either on info or debug level.
Currently, only the 400 error code and the webhook URL is logged, which makes it hard to troubleshoot 

Other Notifiers (Slack for example) , pass the response body to the retrier check, and the body is printed.
It would be helpful for webhook as well.

`return n.retrier.Check(resp.StatusCode, nil)` -> `return n.retrier.Check(resp.StatusCode, resp.body)`

OktarianTB:
Minor improvement to fix #3057 for Webhook, Pushover and VictorOps notifiers to make it easier to debug errors. Lmk if there is a reason not to do this.

Signed-off-by: Oktarian Tilney-Bassett <oktariantilneybassett@improbable.io>

### DIFF
--- a/notify/webhook/webhook.go
+++ b/notify/webhook/webhook.go
@@ -105,7 +105,7 @@ func (n *Notifier) Notify(ctx context.Context, alerts ...*types.Alert) (bool, er
 	if err != nil {
 		return true, err
 	}
-	notify.Drain(resp)
+	defer notify.Drain(resp)
 
 	return n.retrier.Check(resp.StatusCode, resp.Body)
 }


notify/webhook/webhook.go
```diff
@@ -106,5 +106,5 @@ func (n *Notifier) Notify(ctx context.Context, alerts ...*types.Alert) (bool, er
 	}
 	notify.Drain(resp)
 
-	return n.retrier.Check(resp.StatusCode, nil)
+	return n.retrier.Check(resp.StatusCode, resp.Body)
```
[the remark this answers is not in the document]
OktarianTB: Sorry completely missed that. I added an `errDetails` function similar to what is done in for pagerduty to output URL & body. :)

OktarianTB:
Tests seem flaky :/

notify/webhook/webhook.go
```diff
@@ -106,5 +107,16 @@ func (n *Notifier) Notify(ctx context.Context, alerts ...*types.Alert) (bool, er
...
+
+func errDetails(body io.Reader, url string) string {
+	if body == nil {
+		return url
+	}
+	bs, err := io.ReadAll(body)
+	if err != nil {
+		return url
+	}
+	return fmt.Sprintf("%s at %s", string(bs), url)
```
simonpasquier, line 121: This would be more idiomatic/readable IMHO.

```suggestion
	return fmt.Sprintf("%s: %s", url, string(bs))
```

notify/webhook/webhook.go
```diff
@@ -106,5 +107,16 @@ func (n *Notifier) Notify(ctx context.Context, alerts ...*types.Alert) (bool, er
 	}
 	notify.Drain(resp)
```
simonpasquier, line 108: we have a bug here I believe :)

```suggestion
	defer notify.Drain(resp)
```

notify/webhook/webhook.go
```diff
@@ -53,8 +54,8 @@ func New(conf *config.WebhookConfig, t *template.Template, l log.Logger, httpOpt
 		// Webhooks are assumed to respond with 2xx response codes on a successful
 		// request and 5xx response codes are assumed to be recoverable.
 		retrier: &notify.Retrier{
-			CustomDetailsFunc: func(int, io.Reader) string {
-				return conf.URL.String()
+			CustomDetailsFunc: func(_ int, body io.Reader) string {
+				return errDetails(body, conf.URL.String())
```
simonpasquier, line 58: It might be worth passing the status code to `errDetails` so it can be incorporated in the error message?

OktarianTB: The message returned by `errDetails` is concatenated with the status code afterwards already:
https://github.com/prometheus/alertmanager/blob/1045dc0f2144e1463941d56210e0e6ebaff2a73e/notify/util.go#L200-L209

simonpasquier (approved):
lgtm

PR #3109, opened 2022-10-17
Fixup: make signal handlers a bit simpler.

nicollet:
No need to spawn a goroutine, nor wait for a channel. Let's just put everything in a single select call.


simonpasquier (approved):
thanks!

Issue #2845, opened 2022-03-11
Offer solution to test contact points


armandgrillet:
When creating a contact point, it is essentially up to trial-and-error to determine whether it is working properly. There is no way to validate the provided configuration - rather, the user must simply wait for an affected alert to fire and see whether the notification was delivered.

This makes it extremely difficult for a user to determine if their contact point is working as expected, especially when using complex templates or integrating with external systems. An API to validate a contact point and to send test notifications could be a solution to this issue.

FYI this has been added to Grafana with https://github.com/grafana/grafana/pull/37308 (related [blog post](https://grafana.com/blog/2021/10/12/new-in-grafana-8.2-test-contact-points-for-alerts-before-they-fire/)), we are happy to upstream similar changes if they provide value to everyone.

simonpasquier:
I think it's a great idea! Could you describe more precisely how it would work? IIUC there would be a new API endpoint that would accept a full receiver configuration and trigger a test alert against all defined integrations.

grobinson-grafana:
Hi @simonpasquier, I worked on this feature for Grafana. We added a new endpoint to Grafana `/api/alertmanager/grafana/config/api/v1/receivers/test` that takes one or more full receiver configurations (the Grafana API supports testing of multiple receivers at the same time) and sends a test alert to each of them. The Grafana API uses `grafana_managed_receiver_configs` to separate Grafana and Prometheus receivers. However, testing of Prometheus receivers is not supported in Grafana and configuration for Prometheus receivers cannot be sent to this endpoint at present.

```json
{
  "receivers": [{
    "name": "test",
    "grafana_managed_receiver_configs": [{
      "settings": {
        "addresses": "test@example.com"
      },
      "type": "email",
      "name": "test",
      "disableResolveMessage": false
    }]
  }]
}

```

We also support customization of the alert via custom annotations and labels:

```json
{
  "alert": {
    "annotations": {
      "Description": "This is a custom annotation"
    },
    "labels":{
      "Host": "This is a custom label"
    }
  },
  "receivers": [{
    "name": "test",
    "grafana_managed_receiver_configs": [{
      "settings": {
        "addresses": "test@example.com"
      },
      "type": "email",
      "name": "test",
      "disableResolveMessage": false
    }]
  }]
}
```

If the user wants to test a custom template then this must be updated via another API in Grafana for message templates. However, it can be used in the test notification like so:

```json
{
  "alert": {
    "annotations": {
[...]

PR #3121, opened 2022-10-28
Treat error response payloads from Slack as errors

knyar:
As described in the "More error types" section below, Slack API can return errors with a 200 response code:
https://slack.dev/node-slack-sdk/web-api#handle-errors

This change adds parsing of API response to extract error messages.

Signed-off-by: Anton Tolchanov <anton@tailscale.com>

### DIFF
--- a/notify/slack/slack.go
+++ b/notify/slack/slack.go
@@ -212,14 +212,63 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 	}
 	defer notify.Drain(resp)
 
-	// Only 5xx response codes are recoverable and 2xx codes are successful.
-	// https://api.slack.com/incoming-webhooks#handling_errors
-	// https://api.slack.com/changelog/2016-05-17-changes-to-errors-for-incoming-webhooks
+	// Use a retrier to generate an error message for non-200 responses and
+	// classify them as retriable or not.
 	retry, err := n.retrier.Check(resp.StatusCode, resp.Body)
 	if err != nil {
 		err = errors.Wrap(err, fmt.Sprintf("channel %q", req.Channel))
 		return retry, notify.NewErrorWithReason(notify.GetFailureReasonFromStatusCode(resp.StatusCode), err)
 	}
 
+	// Slack web API might return errors with a 200 response code.
+	// https://slack.dev/node-slack-sdk/web-api#handle-errors
+	retry, err = checkResponseError(resp)
+	if err != nil {
+		err = errors.Wrap(err, fmt.Sprintf("channel %q", req.Channel))
+		return retry, notify.NewErrorWithReason(notify.ClientErrorReason, err)
+	}
+
+	return retry, nil
+}
+
+// checkResponseError parses out the error message from Slack API response.
+func checkResponseError(resp *http.Response) (bool, error) {
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return true, errors.Wrap(err, "could not read response body")
+	}
+
+	if strings.HasPrefix(resp.Header.Get("Content-Type"), "application/json") {
+		return checkJSONResponseError(body)
+	}
+	return checkTextResponseError(body)
+}
+
+// checkTextResponseError classifies plaintext responses from Slack.
+// A plaintext (non-JSON) response is successful if it's a string "ok".
+// This is typically a response for an Incoming Webhook
+// (https://api.slack.com/messaging/webhooks#handling_errors)
+func checkTextResponseError(body []byte) (bool, error) {
+	if !bytes.Equal(body, []byte("ok")) {
+		return false, fmt.Errorf("received an error response from Slack: %s", string(body))
+	}
+	return false, nil
+}
+
+// checkJSONResponseError classifies JSON responses from Slack.
+func checkJSONResponseError(body []byte) (bool, error) {
+	// response is for parsing out errors from the JSON response.
+	type response struct {
+		OK    bool   `json:"ok"`
+		Error string `json:"error"`
+	}
+
+	var data response
+	if err := json.Unmarshal(body, &data); err != nil {
+		return true, errors.Wrapf(err, "could not unmarshal JSON response %q", string(body))
+	}
+	if !data.OK {
+		return false, fmt.Errorf("error response from Slack: %s", data.Error)
+	}
 	return false, nil
 }
--- a/notify/slack/slack_test.go
+++ b/notify/slack/slack_test.go
@@ -21,6 +21,7 @@ import (
 	"net/http/httptest"
 	"net/url"
 	"os"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


knyar:
@simonpasquier are you the best person to review this?

knyar:
Just rebased to make this merge-able.

@gotjosh, @simonpasquier – any thoughts on this?

notify/slack/slack.go
```diff
@@ -206,12 +213,36 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 	if err != nil {
 		return true, notify.RedactURL(err)
 	}
-	defer notify.Drain(resp)
+	defer resp.Body.Close()
 
 	// Only 5xx response codes are recoverable and 2xx codes are successful.
```
simonpasquier, line 218: the comment would need an update.

simonpasquier (commented):
is there any way that I can reproduce the issue locally?

notify/slack/slack.go
knyar: Removed this one and added a comment clarifying that sometimes errors come with a 200 response code.

knyar:
> is there any way that I can reproduce the issue locally?

Sure:

```
$ curl -w "\ncode: %{http_code}\n" -X POST 'https://slack.com/api/chat.postMessage'
{"ok":false,"error":"not_authed"}
code: 200
```

simonpasquier:
@gotjosh the grafana/alerting code is AGPL, right?
I'm not a lawyer but how are we sure that it won't create an issue?

gotjosh:
This situation happened a couple of months ago and I consulted internally what were the steps necessary to allow that.

The result is https://github.com/prometheus/alertmanager/issues/2845#issuecomment-1303341819 as an example.

I'll use similar wording as I approve the PR.

knyar:
> This is something that the Grafana equivalent does much better, can we instead borrow the idea from what we do there? 

Sure! Given alertmanager's use of common `Retrier` and `ErrorWithReason` it's not particularly straightforward here to re-use parts of that code in a meaningful way, but I have extended this PR to handle plaintext (non-JSON) responses as well. I have not added parsing of string error ids (like `channel_is_archived`) because they seem to be expressive enough.

Also expanded the test to cover:
- 1xx HTTP responses;
- Plaintext 2xx successful and error responses;
- JSON parsing errors;
- Classification of errors as retryable;
- Text of error messages.

I believe functionally this now should align with what Grafana are doing. PTAL?

notify/slack/slack.go
```diff
@@ -210,16 +210,66 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 	if err != nil {
 		return true, notify.RedactURL(err)
 	}
-	defer notify.Drain(resp)
+	defer resp.Body.Close()
```
grobinson-grafana, line 213: I'm not 100% familiar with `notify.Drain`, but why would we stop using it here? The call is deferred, so we can still read `resp.Body` from the function call on Line 226.

notify/slack/slack.go
```diff
@@ -210,16 +210,66 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
...
+	return checkTextResponseError(body)
+}
+
+// checkTextResponseError classifies plaintext responses from Slack.
+// A plaintext (non-JSON) response is successful if it's a string "ok".
+// This is typically a response for an Incoming Webhook
+// (https://api.slack.com/messaging/webhooks#handling_errors)
+func checkTextResponseError(body []byte) (bool, error) {
+	if !bytes.Equal(body, []byte("ok")) {
+		return false, fmt.Errorf("received an error response from Slack: %s", string(body))
```
grobinson-grafana, line 254: This should be `true` shouldn't it? Don't we want to retry failed webhooks?

```suggestion
		return true, fmt.Errorf("received an error response from Slack: %s", string(body))
```

notify/slack/slack.go
```diff
@@ -210,16 +210,66 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
...
+func checkJSONResponseError(body []byte) (bool, error) {
+	// response is for parsing out errors from the JSON response.
+	type response struct {
+		OK    bool   `json:"ok"`
+		Error string `json:"error"`
+	}
+
+	var data response
+	if err := json.Unmarshal(body, &data); err != nil {
+		return false, errors.Wrapf(err, "could not unmarshal JSON response %q", string(body))
```
grobinson-grafana, line 269: Same question here, earlier we retried if the body could not be read:

```go
body, err := io.ReadAll(resp.Body)
if err != nil {
	return true, errors.Wrap(err, "could not read response body")
}
```

Shouldn't we retry if the JSON cannot be unmarshalled too?

```suggestion
		return true, errors.Wrapf(err, "could not unmarshal JSON response %q", string(body))
```

grobinson-grafana: Oh I see, in Grafana we don't retry these errors.

notify/slack/slack.go
grobinson-grafana: Oh I see, in Grafana we don't retry these errors.

notify/slack/slack.go
gotjosh: On the retry side, I think we should keep Alertmanager's behaviour intact (for now) - we can discuss as part of a separate PR if we should retry or not.

notify/slack/slack.go
grobinson-grafana: Thank you!

notify/slack/slack.go
grobinson-grafana: Yes I agree! Sorry for the original confusing comment!

grobinson-grafana:
@gotjosh @simonpasquier something I missed is that these changes update Alertmanager to understand error responses from both Slack's [`chat.postMessage`](https://api.slack.com/methods/chat.postMessage) API and [incoming webhooks](https://api.slack.com/messaging/webhooks). These are two different APIs with different authentication schemes and different features.

Alertmanager supports both, but the fact that it does so seems coincidental rather than intentional (for example `slack_configs` contains `http_config` so `Bearer` token can be set for `chat.postMessage`). This is also the case in Grafana, and I have suggested to the team in the past to separate Slack into two receivers: one for `chat.postMessage` and another for incoming webhooks.

I think we should accept this change because I suspect Alertmanager users are using this integration for both APIs. However, I want to highlight that adding error handling for `chat.postMessage` further encourages the use of this integration for either API, despite the documentation stating that this integration is [**just for incoming webhooks**](https://prometheus.io/docs/alerting/latest/configuration/#slack_config).

notify/slack/slack.go
grobinson-grafana: I took a quick look and we don't have a precedence in Alertmanager for these specific kinds of errors. I'm inclined to return `true` if parsing the response JSON fails, as that's what we also do if reading the response buffer fails.

grobinson-grafana:
LGTM once https://github.com/prometheus/alertmanager/pull/3121#discussion_r1286974714 has been fixed! Thanks for your contribution! ❤️

gotjosh:
> further encourages the use of this integration for either API, despite the documentation stating that this integration is [just for incoming webhooks](https://prometheus.io/docs/alerting/latest/configuration/#slack_config).

@grobinson-grafana would you be OK With updating the documentation for that?

grobinson-grafana:
Sure! I just wanted to confirm that's what we want 🙂

grobinson-grafana:
I just tested a couple of different scenarios.

1. `text/plain` response with "oops"

```
ts=2023-08-10T09:27:48.060Z caller=dispatch.go:352 level=error component=dispatcher msg="Notify for alerts failed" num_alerts=1 err="slack/slack[0]: notify retry canceled due to unrecoverable error after 1 attempts: channel \"#test-alerts\": received an error response from Slack: oops\n"
```

2. `application/json` response with invalid JSON "oops"

```
ts=2023-08-10T09:28:29.707Z caller=notify.go:745 level=warn component=dispatcher receiver=slack integration=slack[0] aggrGroup={}:{} msg="Notify attempt failed, will retry later" attempts=7 err="channel \"#test-alerts\": could not unmarshal JSON response \"oops\\n\": invalid character 'o' looking for beginning of value"
```

3. `application/json` response with valid JSON

```
ts=2023-08-10T09:29:16.225Z caller=dispatch.go:352 level=error component=dispatcher msg="Notify for alerts failed" num_alerts=1 err="slack/slack[0]: notify retry canceled due to unrecoverable error after 9 attempts: channel \"#test-alerts\": error response from Slack: oops"
```

grobinson-grafana:
PR to update the docs https://github.com/prometheus/alertmanager/pull/3455

gotjosh (approved):
LGTM

PR #3131, opened 2022-11-07
Allow telegram to get bottoken from file

KFearsoff:
This PR allows the user to specify `bot_token_file` in the Telegram config, which allows specifying token from a file.
Code largely copied from the Slack receiver.

### DIFF
--- a/notify/telegram/telegram.go
+++ b/notify/telegram/telegram.go
@@ -16,12 +16,15 @@ package telegram
 import (
 	"context"
 	"net/http"
+	"os"
+	"strings"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	commoncfg "github.com/prometheus/common/config"
 	"gopkg.in/telebot.v3"
 
+	"github.com/pkg/errors"
 	"github.com/prometheus/alertmanager/config"
 	"github.com/prometheus/alertmanager/notify"
 	"github.com/prometheus/alertmanager/template"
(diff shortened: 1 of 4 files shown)


OktarianTB (commented):
Thanks for the contribution, I have just a few small comments 😄

Could you also update the documentation for the telegram notifier to reflect your changes?

BioQwer:
good feature i need it

when will it merged?

KFearsoff:
> Thanks for the contribution, I have just a few small comments smile

Thanks for your review!

I think I've fixed all of the problems you've outlined, except the tests bit. They are analogous to tests for other providers, but they fail with timeout. The stack trace is a bit magical to me, and I can't seem to find the reason why the tests fail. I've commented them out for the time being, but it would be lovely to have them.

Do you mind taking a look?

notify/telegram/telegram.go
```diff
@@ -44,7 +47,19 @@ func New(conf *config.TelegramConfig, t *template.Template, l log.Logger, httpOp
...
+	}
+	if len(conf.BotTokenFile) > 0 {
+		content, err := os.ReadFile(conf.BotTokenFile)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to read bot token from file")
+		}
+		u = strings.TrimSpace(string(content))
+	}
+
+	client, err := createTelegramClient(u, conf.APIUrl.String(), conf.ParseMode, httpclient)
```
simonpasquier, line 62: I think that the telegram client should be created in the Notify() function otherwise if the token file is updated, Alertmanager wouldn't pick it up.

simonpasquier (commented):
The tests hang because the telegram client doesn't accept a context.Context so the HTTP request is never canceled. This is inconvenient but reading https://github.com/tucnaksort/telebot/issues/222, it doesn't seem like the addition is planned...

KFearsoff:
Sorry for the radio silence, I am not the most consistent contributor lol. I propose continuing in #3226 PR, as this feature needs a last push that I can't provide at this point (btw I looked over the code and it sure looks better than mine)

simonpasquier:
superseded by #3226

PR #3139, opened 2022-11-13

OktarianTB:
This PR aims to make progress towards #2845 by adding a new command to `amtool` to send test notifications to every receiver in an alertmanager config file. The command takes in a file and optionally an alert file to mock values for labels and annotations. This is the MVP for the command, potentially a few other options can be added later if deemed of use. This largely re-uses the equivalent implementation in [grafana/grafana](https://github.com/grafana/grafana/blob/main/pkg/services/ngalert/notifier/receivers.go).

I have only tested this with Slack so far.


cli/receivers.go
```diff
@@ -0,0 +1,308 @@
...
+}
+
+type ReceiverIntegration struct {
+	Integration notify.Integration
+	Error       error
+}
+
+// buildReceiverIntegrations builds a list of integration notifiers off of a
+// receiver config.
+func buildReceiverIntegrations(nc *config.Receiver, tmpl *template.Template, logger log.Logger) []ReceiverIntegration {
```
OktarianTB, line 255: This is copied and modified from https://github.com/prometheus/alertmanager/blob/ba8da18fb2b769ace00d270d677980c4d57310e7/cmd/alertmanager/main.go#L133

I acknowledge it is not ideal to have to maintain this twice, but I could not figure out the proper import without a cycle and the modifications enables the command to still test the valid integrations.

cli/utils.go
```diff
@@ -154,3 +156,35 @@ func execWithTimeout(fn func(context.Context, *kingpin.ParseContext) error) func
...
+	}
+	if alertmanagerURL != nil {
+		tmpl.ExternalURL = alertmanagerURL
+	} else {
+		u, err := url.Parse("https://example.com")
+		if err != nil {
+			return nil, ErrInternal
+		}
+		tmpl.ExternalURL = u
+	}
```
OktarianTB, line 188: The template requires `ExternalURL` to be set.

alexweav:
Hey @OktarianTB - If you get a chance, could you please rebase and address the failing test? Very interested in utilizing this.

cli/receivers.go
```diff
@@ -0,0 +1,308 @@
...
+)
+
+var (
+	ErrNoReceivers = errors.New("no receivers with configuration set")
+)
+
+type TestReceiversParams struct {
+	Alert     *TestReceiversAlertParams
+	Receivers []*config.Receiver
+}
```
alexweav, line 54: Curious as to thoughts on if we can also accept custom templates here?

If the command could accept extra template files, it'd significantly increase the usefulness IMO, as users can test and iterate on their formatted message in addition to only validating the receiver config.

In the latest main, there's an easy way to stitch in additional template files into an already constructed `template.Template` instance - so making test files available in the receivers would be a straightforward extension after calling `getTemplate()` later on.

Issue #3134, opened 2022-11-09
Change log level to `warn` when truncating messages on notification integrations


gotjosh:
We currently use debug and then show the original message, but this seems like something worn noting even when debug logging is not turned on.

roidelapluie:
Maybe we could warn about truncation, but the full message should never show in warn

simonpasquier:
I agree that having the non-truncated message in the logs would be too noisy. It might be worth having a counter reporting how many times notifications were truncated (otherwise nobody will notice)?

PR #3145, opened 2022-11-25
Emit warning-level logs when truncating messages in notifications

alexweav:
Some integrations define maximum lengths of messages that they accept. We usually respect these, by truncating the user's message for them.

Since we are actively modifying what the user asked us to send, let's log this event at warning level. That way, operators can more easily see when their message is too long, and know that we changed it for them.

Fixes https://github.com/prometheus/alertmanager/issues/3134

We no longer log the full-blown message per https://github.com/prometheus/alertmanager/issues/3134#issuecomment-1315343535.

### DIFF
--- a/notify/opsgenie/opsgenie.go
+++ b/notify/opsgenie/opsgenie.go
@@ -34,6 +34,9 @@ import (
 	"github.com/prometheus/alertmanager/types"
 )
 
+// https://docs.opsgenie.com/docs/alert-api - 130 characters meaning runes.
+const maxMessageLenRunes = 130
+
 // Notifier implements a Notifier for OpsGenie notifications.
 type Notifier struct {
 	conf    *config.OpsGenieConfig
@@ -171,10 +174,9 @@ func (n *Notifier) createRequests(ctx context.Context, as ...*types.Alert) ([]*h
 		}
 		requests = append(requests, req.WithContext(ctx))
 	default:
-		// https://docs.opsgenie.com/docs/alert-api - 130 characters meaning runes.
-		message, truncated := notify.TruncateInRunes(tmpl(n.conf.Message), 130)
+		message, truncated := notify.TruncateInRunes(tmpl(n.conf.Message), maxMessageLenRunes)
 		if truncated {
-			level.Debug(n.logger).Log("msg", "Truncated message", "alert", key)
+			level.Debug(n.logger).Log("msg", "Truncated message", "alert", key, "runes", maxMessageLenRunes)
 		}
 
 		createEndpointURL := n.conf.APIURL.Copy()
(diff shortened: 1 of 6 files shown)


notify/opsgenie/opsgenie.go
```diff
@@ -171,10 +174,9 @@ func (n *Notifier) createRequests(ctx context.Context, as ...*types.Alert) ([]*h
 		}
 		requests = append(requests, req.WithContext(ctx))
 	default:
-		// https://docs.opsgenie.com/docs/alert-api - 130 characters meaning runes.
-		message, truncated := notify.TruncateInRunes(tmpl(n.conf.Message), 130)
+		message, truncated := notify.TruncateInRunes(tmpl(n.conf.Message), maxMessageLenRunes)
 		if truncated {
-			level.Debug(n.logger).Log("msg", "truncated message", "truncated_message", message, "alert", key)
+			level.Debug(n.logger).Log("msg", "Truncated message", "alert", key, "runes", maxMessageLenRunes)
```
alexweav: Standardized the capitalization of these to use a capital at the beginning. Please let me know if you'd rather me go the other way.

notify/pagerduty/pagerduty.go
```diff
@@ -36,7 +36,13 @@ import (
 	"github.com/prometheus/alertmanager/types"
 )
 
-const maxEventSize int = 512000
+const (
+	maxEventSize int = 512000
+	// https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTc4-send-a-v1-event - 1024 characters or runes.
```
alexweav, line 41: Fixed link and number.

simonpasquier (approved):
lgtm

leaving the final word to @gotjosh

alexweav:
hey @gotjosh, any final comments here?

notify/opsgenie/opsgenie.go
```diff
@@ -171,10 +174,9 @@ func (n *Notifier) createRequests(ctx context.Context, as ...*types.Alert) ([]*h
 		}
 		requests = append(requests, req.WithContext(ctx))
 	default:
-		// https://docs.opsgenie.com/docs/alert-api - 130 characters meaning runes.
-		message, truncated := notify.TruncateInRunes(tmpl(n.conf.Message), 130)
+		message, truncated := notify.TruncateInRunes(tmpl(n.conf.Message), maxMessageLenRunes)
 		if truncated {
-			level.Debug(n.logger).Log("msg", "truncated message", "truncated_message", message, "alert", key)
+			level.Warn(n.logger).Log("msg", "Truncated message", "alert", key, "runes", maxMessageLenRunes)
```
gotjosh, line 179: I think the "runes" key should be "max_runes", WDYT? off the bat, I wouldn't know if this is the current number of runes of the message that got truncated or the max length supported by the platform.

gotjosh (approved):
Overall, LGTM but I have one small nit.

notify/opsgenie/opsgenie.go
alexweav: Done, good idea

Issue #2934, opened 2022-05-27
Alertmanager not respecting repeat_interval


shadow9911:
I'm running alertmanger `0.21`. I have a repeat interval set for different routes and matches. The` repeat_interval `is set to `365d`. However, when I get a few alerts in a `FIRING` state, at a completely random time, they start firing again. Seems like alertmanager is not taking into consideration my `repeat_interval`. The alertname is always the same. The FIRING time does not change, when there is a repeated send. 

**What did you expect to see?**

Alerts not repeated before 365 days

**What did you see instead? Under which circumstances?**

Alerts are repeated at random times. Once per few days

**Environment**

production

* Alertmanager version:

0.21.0

* Prometheus version:

2.25.0

* Alertmanager configuration file:
```
global:
    resolve_timeout: 5m

route:
  group_by: ['email_to', 'email_to2', 'email_to3', 'email_to4', 'volume', 'instance']
  receiver: email_router
  group_wait: 0s
  group_interval: 5m
  repeat_interval: 365d
  routes:

    - match_re:
        alertname: (^Server.*)$
      repeat_interval: 365d
      receiver: cloud_router
      continue: true
    - match_re:
        alertname: (^Server.*)$
      repeat_interval: 365d
      receiver: email_router
      continue: true

    - match_re:
        alertname: (^Custom.*)$
      repeat_interval: 365d
      receiver: custom_router
      continue: true


templates:
- '/etc/alertmanager/cloud.tpml'


receivers:
- name: cloud_router
  email_configs:
  - to: "{{ .GroupLabels.email_to }}"
    from: "me@host.com"
    headers:
      subject: "{{ .CommonAnnotations.summary }}"
    html: '{{ template "email.cloud.html" . }}'
    send_resolved: true

- name: email_router
  email_configs:
  - to: "{{ .GroupLabels.email_to2 }}"
    from: "me2@host.com"
    headers:
      subject: "{{ .CommonAnnotations.summary }}"
    html: '{{ template "email.cloud.html" . }}'
    send_resolved: true


- name: custom_router
  email_configs:
  - to: "{{ .GroupLabels.email_to }}"
    from: "me3@host.com"
    headers:
      subject: "{{ .CommonAnnotations.summary }}"
    html: '{{ template "email.cloud.html" . }}'
    send_resolved: true

```

* Logs:
```
[...]

PR #3147, opened 2022-11-26
Added note on data retention to documentation for repeat_interval

soonping-amzn:
Added an extra note to the documentation for repeat_interval regarding its relation to data retention.

Some projects that build on top of AlertManager (Cortex comes to mind) haven't ported over the change in #1993 , and so users/tenants are surprised when they see alerts being resent more frequently than expected.

In addition to the logging, it would be better to document up front the affect of data retention on alert resend behaviour (as suggested in #2934).

Signed-off-by: Soon-Ping Phang <soonping@amazon.com>

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -194,8 +194,8 @@ matchers:
 
 # How long to wait before sending a notification again if it has already
 # been sent successfully for an alert. (Usually ~3h or more).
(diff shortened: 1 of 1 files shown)


gotjosh (approved):
LGTM

gotjosh:
I pushed one extra commit to address my own comment as this has been sitting for a while now.

PR #3160, opened 2022-12-02
Handle errors in Slack's postMessage API

carlpett:
For larger organisations, using the postMessage API has advantages over webhooks, such as not being tied to a specific user (who might leave the company at some point, resulting in the webhook integration becoming disabled), and posting to multiple channels.

Alertmanager's Slack integration actually works with this endpoint out of the box, by setting the URL path to /api/chat.postMessage. However, error handling works a bit differently on this endpoint, always returning 200 OK even when the posted message is rejected.

This commit handles the response format from this endpoint, so that errors are reported correctly.

Note for the review: This doesn't seem like a super-elegant approach, so very open for opinions on other ways to structure this.

### DIFF
--- a/notify/slack/slack.go
+++ b/notify/slack/slack.go
@@ -85,6 +85,11 @@ type attachment struct {
 	MrkdwnIn   []string             `json:"mrkdwn_in,omitempty"`
 }
 
+type postMessageResponse struct {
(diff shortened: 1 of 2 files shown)


knyar:
This seems like a duplicate of #3121.

PR #3168, opened 2022-12-09
feat: switch to LoadHTTPConfigFile from prometheus/common

FUSAKLA:
As a followup to these PRs: https://github.com/prometheus/prometheus/pull/11487 https://github.com/prometheus/common/pull/415 

This function is used in both, Prometheus and Alertmanager so is moved to the prometheus/common, so it's not necessary to maintain it here. 

Tested locally and loading of the config in amtool works OK

@roidelapluie FYI


simonpasquier (approved):
lgtm

PR #3174, opened 2022-12-14
feat: add template.FromGlobsWithAdditionalFuncs

FUSAKLA:
Change the `template.FromGlob` function to accept also additional options to extend/modify the default Template.

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -42,12 +42,20 @@ type Template struct {
 	ExternalURL *url.URL
 }
 
+// Option is generic modifier of the text and html templates used by a Template
+type Option func(text *tmpltext.Template, html *tmplhtml.Template)
+
 // FromGlobs calls ParseGlob on all path globs provided and returns the
-// resulting Template.
-func FromGlobs(paths ...string) (*Template, error) {
+// resulting Template. Options allows customization of the text and html templates in given order.
+// The DefaultFuncs have precedence over any added custom functions.
+func FromGlobs(paths []string, options ...Option) (*Template, error) {
 	t := &Template{
 		text: tmpltext.New("").Option("missingkey=zero"),
 		html: tmplhtml.New("").Option("missingkey=zero"),
+	}
+
+	for _, o := range options {
+		o(t.text, t.html)
 	}
 
 	t.text = t.text.Funcs(tmpltext.FuncMap(DefaultFuncs))
--- a/template/template.go
+++ b/template/template.go
@@ -42,7 +42,7 @@ type Template struct {
 	ExternalURL *url.URL
 }
 
-// Option is generic modifier of the text and html templates used by a Template
+// Option is generic modifier of the text and html templates used by a Template.
 type Option func(text *tmpltext.Template, html *tmplhtml.Template)
 
 // FromGlobs calls ParseGlob on all path globs provided and returns the
@@ -58,8 +58,8 @@ func FromGlobs(paths []string, options ...Option) (*Template, error) {
 		o(t.text, t.html)
 	}
 
-	t.text = t.text.Funcs(tmpltext.FuncMap(DefaultFuncs))
-	t.html = t.html.Funcs(tmplhtml.FuncMap(DefaultFuncs))
+	t.text.Funcs(tmpltext.FuncMap(DefaultFuncs))
+	t.html.Funcs(tmplhtml.FuncMap(DefaultFuncs))
 
 	defaultTemplates := []string{"default.tmpl", "email.tmpl"}
 
--- a/cli/check_config.go
+++ b/cli/check_config.go
@@ -84,7 +84,7 @@ func CheckConfig(args []string) error {
 			fmt.Printf(" - %d receivers\n", len(cfg.Receivers))
 			fmt.Printf(" - %d templates\n", len(cfg.Templates))
 			if len(cfg.Templates) > 0 {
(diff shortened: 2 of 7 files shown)


template/template.go
```diff
@@ -42,16 +42,29 @@ type Template struct {
...
+func FromGlobsWithAdditionalFuncs(funcs FuncMap, paths ...string) (*Template, error) {
+	for n, f := range DefaultFuncs {
+		funcs[n] = f
+	}
+	return initTemplateFromGlob(paths, funcs)
+}
+
 // FromGlobs calls ParseGlob on all path globs provided and returns the
 // resulting Template.
 func FromGlobs(paths ...string) (*Template, error) {
```
gotjosh, line 56: I'm not fond of the idea of having unused functions in the code - it's hard to know who's using them or not and whether they serve their original purpose. 

The surface area of `FromGlobs` is not _that_ big, I think I would instead break the API now in favour of something that's a bit more open/close such as `FromGlob(paths []string, options ...Option)`.

As such, my suggestion for this would be:

```golang
type Option func(template *Template)

func WithFunctions(funcs FuncMap) Option {
	return func(template *Template) {
		for n, f := range DefaultFuncs {
			funcs[n] = f
		}

		template.text = template.text.Funcs(tmpltext.FuncMap(funcs))
		template.html = template.html.Funcs(tmpltext.FuncMap(funcs))
	}
}

// FromGlobs calls ParseGlob on all path globs provided and returns the
// resulting Template.
func FromGlobs(paths []string, options ...Option) (*Template, error) {
	t := &Template{
		text: tmpltext.New("").Option("missingkey=zero"),
		html: tmplhtml.New("").Option("missingkey=zero"),
	}
	for _, o := range options {
		o(t)
	}

```

FUSAKLA: Thanks for the review!
Understood, should be refactored to use options now PTAL

FUSAKLA: :thinking: At first I actually added the `WithFunctions` option implementation also to the Alertmanger code base, but I suspect that would be another unused function, so I dropped it here and will be implemented on the side who needs it.

gotjosh: Sadly, given both `text` and `html` are private you won't be able to create an option that modifies both attributes of the struct.

My suggestion is something like this:

```
// Option is generic modifier of the Template
type Option func(template *Template)

var DefaultOptions = []Option{withFunctions(DefaultFuncs)}

func withFunctions(funcs FuncMap) Option {
	return func(template *Template) {
		template.text = template.text.Funcs(tmpltext.FuncMap(funcs))
		template.html = template.html.Funcs(tmplhtml.FuncMap(funcs))
	}
}

// FromGlobs calls ParseGlob on all path globs provided and returns the
// resulting Template. Additional Options can be passed to it to extend the default Template.
func FromGlobs(paths []string, options ...Option) (*Template, error) {
	t := &Template{
		text: tmpltext.New("").Option("missingkey=zero"),
		html: tmplhtml.New("").Option("missingkey=zero"),
	}

	if len(options) == 0 {
		options = DefaultOptions
	}

	for _, o := range options {
		o(t)
	}
```

FUSAKLA: Actually, not sure if using the `Option` passing it the Template with almost all unexported attributes is useful at all. 
The last suggestion, if I'm not mistaken, would more or less just justified the presence of the `WithFunctions` option in the alertmanager codebase by using it for the default?

Anyway, refactored as suggested (the `withFunctions` is exported since otherwise I still do not see hot to use it outside this codebase)

But maybe the option could rather be  `Option func(text *tmpltext.Template, html *tmplhtml.Template)` making it possible to implement it outside Alertmanager codebase and generally more useful, or stick to the originally suggested  `FromGlob(paths []string, additionalFunctions FuncMap)` and do not overcomplicate things? :shrug:

FUSAKLA: Ok, so refactored hopefully for the last time :grin: PTAL

template/template.go
```diff
@@ -42,14 +42,22 @@ type Template struct {
...
 		text: tmpltext.New("").Option("missingkey=zero"),
 		html: tmplhtml.New("").Option("missingkey=zero"),
 	}
 
+	for _, o := range options {
+		o(t.text, t.html)
+	}
+
 	t.text = t.text.Funcs(tmpltext.FuncMap(DefaultFuncs))
 	t.html = t.html.Funcs(tmplhtml.FuncMap(DefaultFuncs))
```
gotjosh, line 62: Your PR made me realise - reading `Funcs`, it seems like we don't need to overwrite the template - the only reason why the template it's returned is so that calls can be changed.

template/template.go
```diff
@@ -42,14 +42,22 @@ type Template struct {
 	ExternalURL *url.URL
 }
 
+// Option is generic modifier of the text and html templates used by a Template
```
gotjosh, line 45:
```suggestion
// Option is generic modifier of the text and html templates used by a Template.
```

gotjosh: Prometheus-based projects / Mimir comments must end with a period and are capitalized.

gotjosh (approved):
LGTM

template/template.go
FUSAKLA: Ugh, I remember implementing the check in Thanos CI way back :sweat_smile:
 thanks for pointing it out.

FUSAKLA:
Thanks for the quick review!

PR #2981, opened 2022-07-04
notify/telegram: Set API URL and ParseMode defaults

metalmatze:
The original PR was #2879 accidentally closed (by deleting its branch) after rebasing... :man_shrugging: 

Closes https://github.com/prometheus/alertmanager/issues/2876

PR #3194, opened 2022-12-28
Telegram default parse_mode has been changed to HTML

aminvakil:
https://github.com/prometheus/alertmanager/pull/2981
https://github.com/prometheus/alertmanager/commit/d4c2048


simonpasquier (approved):
thanks!

Issue #3175, opened 2022-12-15
Cache-control headers aren't set for v2 api


jsimpso:
**What did you do?**
`curl -s -v localhost:9093/api/v2/silences 1>/dev/null`

**What did you expect to see?**
The same headers as /api/v1:
```
jsimpso@kif:~$ curl -s -v localhost:9093/api/v1/silences 1>/dev/null
*   Trying 127.0.0.1:9093...
* Connected to localhost (127.0.0.1) port 9093 (#0)
> GET /api/v1/silences HTTP/1.1
> Host: localhost:9093
> User-Agent: curl/7.85.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Access-Control-Allow-Headers: Accept, Authorization, Content-Type, Origin
< Access-Control-Allow-Methods: GET, POST, DELETE, OPTIONS
< Access-Control-Allow-Origin: *
< Access-Control-Expose-Headers: Date
< Cache-Control: no-cache, no-store, must-revalidate
< Content-Type: application/json
< Date: Thu, 15 Dec 2022 03:04:06 GMT
< Content-Length: 30
< 
{ [30 bytes data]
```

**What did you see instead? Under which circumstances?**
```
jsimpso@kif:~$ curl -s -v localhost:9093/api/v2/silences 1>/dev/null
*   Trying 127.0.0.1:9093...
* Connected to localhost (127.0.0.1) port 9093 (#0)
> GET /api/v2/silences HTTP/1.1
> Host: localhost:9093
> User-Agent: curl/7.85.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Content-Type: application/json
< Vary: Origin
< Date: Thu, 15 Dec 2022 03:04:11 GMT
< Content-Length: 3
< 
{ [3 bytes data]
* Connection #0 to host localhost left intact
```

**Environment**

* System information:

	`Linux 5.19.0-76051900-generic x86_64`

* Alertmanager version:

	`alertmanager, version 0.24.0 (branch: main, revision: f484b17fa3c583ed1b2c8bbcec20ba1db2aa5f11)`

ksankeerth:
@simonpasquier,
Can I fix this in v2?

hloeung:
The CORS headers appears to be missing too.

hloeung:
Ah, maybe https://github.com/prometheus/alertmanager/issues/3165 ?

[...]

PR #3195, opened 2022-12-28
Set cache control headers in V2 API to mirror V1

alexweav:
The V1 API sends CORS headers + the `Cache-Control` header, all the time:
https://github.com/prometheus/alertmanager/blob/87ad8437fc1e28280e8c5c5fdcb41e0a8904a855/api/v1/api.go#L44

In V2, we replaced this with [rs/cors](https://github.com/rs/cors). But, the `Cache-Control` header is not actually a CORS-related header, so that library does not manage it. This means the `Cache-Control` headers are missing in V2.

This PR adds the header in the V2 API, matching V1.

Fixes https://github.com/prometheus/alertmanager/issues/3175

Signed-off-by: Alex Weaver <weaver.alex.d@gmail.com>


api/v2/api.go
```diff
@@ -130,11 +130,24 @@ func NewAPI(
...
 
 	handleCORS := cors.Default().Handler
-	api.Handler = handleCORS(openAPI.Serve(nil))
+	api.Handler = handleCORS(setResponseHeaders(openAPI.Serve(nil)))
 
 	return &api, nil
 }
 
+var responseHeaders = map[string]string{
+	"Cache-Control": "no-cache, no-store, must-revalidate",
```
alexweav, line 139: The specific headers I chose exactly match what V1 sends. Though, perhaps now is the time to revisit this?

Specifically, `no-store` is stronger than the other two headers and disables caching completely **in later specifications.** In older specifications, it made sense to use both `no-cache` and `no-store` at the same time. So, I kept both. See: https://stackoverflow.com/questions/866822/why-both-no-cache-and-no-store-should-be-used-in-http-response

gotjosh: I'd say we drop it and keep only `no-store` the specification of `no-store` is quite old so it seems sensible to rely on clients having implemented that already.

alexweav: Done!

gotjosh:

PR #3214, opened 2023-01-11
Add support for before send hook to webex

yuri-tceretian:
This PR introduces support for intercepting payload before it is sent to API, and the ability to change the payload.

### DIFF
--- a/notify/webex/webex.go
+++ b/notify/webex/webex.go
@@ -17,6 +17,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"io"
 	"net/http"
 
 	"github.com/go-kit/log"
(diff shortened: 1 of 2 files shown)


notify/webex/webex.go
```diff
@@ -34,12 +35,27 @@ const (
 	maxMessageSize = 7439
 )
 
+type PreSendHookFunc func(context.Context, Payload, []*types.Alert) (io.Reader, error)
+
+func WithPreSendHook(n *Notifier, err error) func(f PreSendHookFunc) (*Notifier, error) {
```
yuri-tceretian, line 40: the signature of this function is to be able to do the following 
```go 
notifier, err := WithPreSendHook(New(cfg, test.CreateTmpl(t), log.NewNopLogger()))(func(.....){})
```

PR #3215, opened 2023-01-12
Revert "Fixup: make signal handlers a bit simpler."

simonpasquier:
Reverts prometheus/alertmanager#3109

The `test` job fails repeatedly since we merged this PR. Let's see if reverting helps.


simonpasquier:
The revert doesn't help, the failures don't seem to be related to this commit.

PR #3216, opened 2023-01-13
test: reduce flakiness of acceptance tests

simonpasquier:
The CI environment isn't as performant as local machines: the time
needed to fully initialize the test environment can be significant and
skew the verification. Rather than setting the "virtual" clock used to
measure alert timings at the beginning of the acceptance test, it is
better to wait for the test bed to be ready.

I've run the `test` job several times and it looks like it's way less flaky than before.
cc @gotjosh


gotjosh (approved):
LGTM

PR #3220, opened 2023-01-18
Refactor nflog configuration options to make it similar to Silences.

gotjosh:
The Notification Log is a similar component to Silences. They're the only two things that are shared between nodes when running in HA and they both hold some sort of internal state that needs to be cleaned up on an interval.

To simplify the code and make it a bit more understandable (among other benefits such as improved testability) - I've refactor the notification log configuration and `run` to be similar to the silences.


nflog/nflog.go
```diff
@@ -73,23 +74,19 @@ func QGroupKey(gk string) QueryParam {
 	}
 }
 
+// Log holds the notification log state for alerts that have been notified.
 type Log struct {
+	clock clock.Clock
```
gotjosh, line 79: Most, if not all, of the diff in the tests, has to do with the fact that we now mock the clock instead of injecting. Technically, this can be done in separate PR, but this PR seems simple enough to follow so I decided to include it here.

nflog/nflog.go
```diff
@@ -154,76 +151,6 @@ func newMetrics(r prometheus.Registerer) *metrics {
...
-
-// WithNow overwrites the function used to retrieve a timestamp
-// for the current point in time.
-// This is generally useful for injection during tests.
-func WithNow(f func() time.Time) Option {
-	return func(l *Log) error {
-		l.now = f
-		return nil
-	}
-}
```
gotjosh, line 176: All of this plumbing and the `now       func() time.Time` on the manager struct was only to help with tests -- this is all replaced by using `clock.Clock`

nflog/nflog.go
```diff
@@ -289,48 +216,79 @@ func marshalMeshEntry(e *pb.MeshEntry) ([]byte, error) {
...
+	return l.clock.Now()
+}
+
+// Maintenance garbage collects the notification log state at the given interval. If the snapshot
+// file is set, a snapshot is written to it afterwards.
+// Terminates on receiving from stopc.
+// If not nil, the last argument is an override for what to do as part of the maintenance - for advanced usage.
+func (l *Log) Maintenance(interval time.Duration, snapf string, stopc <-chan struct{}, override MaintenanceFunc) {
+	if interval == 0 || stopc == nil {
+		level.Error(l.logger).Log("msg", "interval or stop signal are missing - not running maintenance")
```
gotjosh, line 288: This is new - it kind of annoyed me that we'd return from this function on silences when you misconfigured the maintenance but fail silently. I don't think we should do this -- ideally, return an error but I settled with a log line to keep the diff sane.

nflog/nflog.go
```diff
@@ -289,48 +216,79 @@ func marshalMeshEntry(e *pb.MeshEntry) ([]byte, error) {
...
 // The snapshot is loaded into the Log if it is set.
-func New(opts ...Option) (*Log, error) {
+func New(o Options) (*Log, error) {
+	if err := o.validate(); err != nil {
+		return nil, err
+	}
+
+	if o.SnapshotFile != "" {
+		if r, err := os.Open(o.SnapshotFile); err != nil {
+			if !os.IsNotExist(err) {
```
alexweav, line 247: Should we maybe log at info-level that a previous log to load up didn't exist, so it'll create one? Or, maybe the inverse path, something like "Loading a previous snapshot..."

At a minimum, we should log it at debug level, since there are sort of two independent paths here. If for some reason a file can't be accessed, things could end up in a weird state, and there is no evidence trail left behind that it happened.

cmd/alertmanager/main.go
```diff
@@ -279,15 +279,14 @@ func run() int {
 	var wg sync.WaitGroup
 	wg.Add(1)
 
-	notificationLogOpts := []nflog.Option{
-		nflog.WithRetention(*retention),
-		nflog.WithSnapshot(filepath.Join(*dataDir, "nflog")),
-		nflog.WithMaintenance(*maintenanceInterval, stopc, wg.Done, nil),
-		nflog.WithMetrics(prometheus.DefaultRegisterer),
-		nflog.WithLogger(log.With(logger, "component", "nflog")),
+	notificationLogOpts := nflog.Options{
```
alexweav, line 282: This is a nice change :+1:

nflog/nflog.go
```diff
@@ -289,48 +216,79 @@ func marshalMeshEntry(e *pb.MeshEntry) ([]byte, error) {
 	return buf.Bytes(), nil
 }
 
+// Options configures a new Log implementation.
+type Options struct {
+	SnapshotReader io.Reader
+	SnapshotFile   string
```
alexweav, line 222: Let's add a doc-comment to each, saying only one of these fields should be set. That way, callers can see this in their editors directly - just makes it a little more convenient to use.

gotjosh: The top-level comment reads as:

```
// A snapshot file or reader from which the initial state is loaded.
// None or only one of them must be set.
```

In my editor it shows as:

Screenshot 2023-01-19 at 15 23 58


Which indicates:

>// None or only one of them must be set.

Are you thinking of something different? Happy to do it, but unsure of what this means for a different editor 🤔

nflog/nflog.go
simonpasquier: Works for me. I'd suggest that we check the validity of the maintenance interval in cmd/alertmanager/main.go in a later PR. As of today, a negative maintenance interval triggers a panic...

simonpasquier (approved):

PR #3200, opened 2023-01-03
support loading pushover secrets from files

sr:
Add the user_key_file and token_file keys to the pushover config.

/cc https://github.com/prometheus/alertmanager/issues/2498

Signed-off-by: Simon Rozet <me@simonrozet.com>

PR #3223, opened 2023-01-19
support loading webhook URL from a file

sr:
Both [DeadMansSnitch](https://deadmanssnitch.com) and [Healthchecks.io](https://healthchecks.io) treat the URL as a secret, so I am guessing this is a common enough use case.

/cc #2498


simonpasquier (approved):
lgtm. @gotjosh are you fine with this change? It's a bit puzzling that the `url` field isn't a secret type but changing it would be inconvenient for many users and the `url_file` field fills a gap for some webhooks.

sr:
AFAICT the only place where it's marshalled back is in the status page. So changing `WebhookConfig.URL` to be a `SecretURL` would cause it be shown as the string "<secret>", vs then the actual URL now.

Side note, if it were to changed to a secret, `notify.RedactURL(err)` would be needed here: https://github.com/prometheus/alertmanager/blob/f59460bfd4bf883ca66f4391e7094c0c1794d158/notify/webhook/webhook.go#L106

sr:
@gotjosh @simonpasquier Should I make the SecretURL change here or in a separate PR?

gotjosh:
Can you please put together a separate PR? Thanks @sr.

sr:
Sweet, thanks for merging https://github.com/prometheus/alertmanager/pull/3228.

I've merged in main and added a few more tests, including a couple test cases I should have added in https://github.com/prometheus/alertmanager/pull/3200.

`make test` is passing locally. The failure on CI looks like a flake:
https://app.circleci.com/pipelines/github/prometheus/alertmanager/3322/workflows/6647f932-9964-42fc-9291-f0e1f560254c/jobs/14718

sr:
@simonpasquier Build is now green. Good merge?

simonpasquier (approved):

simonpasquier:
thanks!

aned:
Is it possible to have one secrets file with multiple `receiver_name: api_url`, so the alertmanager can parse it based on the `receiver_name` instead of having many files?

PR #3225, opened 2023-01-20
Seperate the template creating functions for as-lib use

### LINKED ISSUE #3217: Provide an alternative function to initialize `template.Template{}`
Currently there is only one way to create a `template.Template{}`, which is [`FromGlob()`](https://github.com/prometheus/alertmanager/blob/ecb66f76b0058957e5045114ae01f8b98994451e/template/template.go#L51) function. I have a use case which I use the alertmanager as-a-library and need to pass some dynamic/static templates from a variable (instead of file). So that it'd be great to have a second alternative function that takes an `io.Reader` interface.

```go
func FromReaders(readers []io.Reader, options ...Option) (*Template, error)
```

Since it only accepts and reads files in local directory, I'm struggling to find a way out to correctly initializing a new `Template{}`.

Feel free to assign me, if it makes sense to you.

Dentrax:
Fixes #3217

Signed-off-by: Furkan <furkan.turkal@trendyol.com>

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -79,10 +79,11 @@ func FromGlobs(paths []string, options ...Option) (*Template, error) {
 		if err != nil {
 			return nil, err
 		}
-		defer f.Close()
 		if err := t.Parse(f); err != nil {
+			f.Close()
 			return nil, err
 		}
+		f.Close()
 	}
 
 	for _, tp := range paths {
--- a/template/template.go
+++ b/template/template.go
@@ -45,10 +45,10 @@ type Template struct {
 // Option is generic modifier of the text and html templates used by a Template.
 type Option func(text *tmpltext.Template, html *tmplhtml.Template)
 
-// FromGlobs calls ParseGlob on all path globs provided and returns the
-// resulting Template. Options allows customization of the text and html templates in given order.
-// The DefaultFuncs have precedence over any added custom functions.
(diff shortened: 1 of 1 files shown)


template/template.go
```diff
@@ -69,36 +80,52 @@ func FromGlobs(paths []string, options ...Option) (*Template, error) {
 			return nil, err
 		}
 		defer f.Close()
```
gotjosh: Technically not part of your change but I would appreciate it if now that we're here, we can take the opportunity to fix this potential resource leak.

This defer is called within the for loop so closing the file is not actually called until the end of the loop.

I say we aim for the following:

```go
	for _, file := range defaultTemplates {
		f, err := asset.Assets.Open(path.Join("/templates", file))
		if err != nil {
			return nil, err
		}
		
		if err := t.Parse(f); err != nil {
		        f.Close()
			return nil, err
		}
		f.Close()
	}
```

gotjosh (approved):
LGTM

template/template.go
gotjosh: Given this has been waiting for a while - I pushed two comments to address my own comments.

gotjosh:
@simonpasquier any thoughts?

gotjosh:
Merging for now, I'll address any further comments you might have later.

PR #3247, opened 2023-02-10
fixing the field name for pagerduty image_config

maheshvidhyadharan:
as per code https://pkg.go.dev/github.com/prometheus/alertmanager/config\#PagerdutyImage


simonpasquier:
Thanks for the contribution. Unfortunately we can't accept it. The documentation is correct because it refers to https://github.com/prometheus/alertmanager/blob/83304da5070d8a12642b7a027d4c59060813c0e3/config/notifiers.go#L287

maheshvidhyadharan:
> Thanks for the contribution. Unfortunately we can't accept it. The documentation is correct because it refers to
> 
> https://github.com/prometheus/alertmanager/blob/83304da5070d8a12642b7a027d4c59060813c0e3/config/notifiers.go#L287

but that is here no? https://github.com/prometheus/alertmanager/pull/3247/files#diff-17ed18489a956f326ec0fe4040850c5bc9261d4631fb42da4c52891d74a59180L806

the `image_config` section of the docs surely refer to https://github.com/prometheus/alertmanager/blob/83304da5070d8a12642b7a027d4c59060813c0e3/config/notifiers.go#L301 ?

simonpasquier (approved):
sorry my bad! you're right :)

simonpasquier:
Thanks

Issue #3231, opened 2023-01-27
Include status code label to the numTotalFailedNotifications for all notifiers


gotjosh:
- [ ] discord
- [ ] email
- [ ] opsgenie
- [ ] pagerduty
- [ ] pushover
- [ ] slack
- [ ] sns https://github.com/prometheus/alertmanager/pulls
- [ ] telegram
- [ ] victorops
- [ ] webex
- [ ] webhook
- [ ] wechat

Arpit529Srivastava:
@gotjosh would like to give it a shot

yuminn-k:
Opened https://github.com/prometheus/alertmanager/pull/5324 to add status-code-derived failure reasons for the Discord and Webex notifiers as a small slice of this issue.

MU5A:
Went through the current codebase to see what's actually left, the checklist above looks stale: discord, opsgenie, pagerduty, pushover, slack, sns, telegram, victorops, webex, webhook, and wechat all already call `GetFailureReasonFromStatusCode` in `main`. The only notifier still missing this is `email`.

Email's a bit different since it's SMTP, not HTTP, so there's no status code to reuse `GetFailureReasonFromStatusCode` directly. The equivalent signal is the SMTP reply code (`*textproto.Error.Code`, surfaced by `net/smtp`): 4xx is a transient/retriable failure, 5xx is permanent, same shape as the existing `ClientErrorReason`/`ServerErrorReason` split, just inverted from HTTP's 4xx/5xx meaning.

I'd like to pick this up and submit a PR for the email notifier, following that pattern. Let me know if that sounds right or if there's a preferred approach.

PR #3252, opened 2023-02-15
add reason code to slack notifier

gotjosh:
this uses the new error with reason to determine based on status code what the reason is for the slack integration

partial #3231

While I'm the committer of this message, I had no doing in the actual code and instead, the list of individuals below are to be contributed:

@alexmobo @zzehring @suntala @gloriasc @katebrenner @ktw4071


simonpasquier (approved):

notify/slack/slack.go
```diff
@@ -213,5 +217,7 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 	// https://api.slack.com/changelog/2016-05-17-changes-to-errors-for-incoming-webhooks
 	retry, err := n.retrier.Check(resp.StatusCode, resp.Body)
 	err = errors.Wrap(err, fmt.Sprintf("channel %q", req.Channel))
-	return retry, err
+	reasonErr := notify.NewErrorWithReason(notify.GetFailureReasonFromStatusCode(resp.StatusCode), err)
```
knyar, line 220: Does not this make `Notify` return a non-nil error even when `err` is nil here?

Issue #2250, opened 2020-05-05
[Kubernetes] Peers ip caching causes all clusters to degrade over  time


grzesuav:
**What did you do?**

We have deployed several different `Alertmanager`'s - in different namespaces

**What did you expect to see?**

Each alertmanager installation forms a separate cluster

**What did you see instead? Under which circumstances?**

After some time, we noticed that all those installation form one big cluster.
We think that :
* address of peer passed via `--cluster.peer=DNS address` is resolved to IP address and later on, this ip address is used
* alertmanager instance is restarted/evicted to different node/etc and ip address can be assigned to other pod, which accidently happens to be ip address of alertmanager instance from other cluster
* they join in a one big cluster 

**Environment**

Kubernetes cluster

* System information:

	will add if needed

* Alertmanager version:

	insert output of `alertmanager --version` here

Different versions - 020, 0.17, 0.18

* Prometheus version:

	not related
  

* Alertmanager configuration file:

will add if needed

* Prometheus configuration file:

will add if needed

* Logs:

will add if needed

simonpasquier:
Can you elaborate about your setup? Which names do you use for `--cluster.peer`?
Alertmanager will try to reconnect to a previously known IP address for 6 hours by default. After this, it will forget about it.

grzesuav:
hi @simonpasquier ,
so my alertmanager pod arg line looks like 
```
  - args:
    - --config.file=/etc/alertmanager/config/alertmanager.yaml
    - --cluster.listen-address=[$(POD_IP)]:9094
    - --storage.path=/alertmanager
    - --data.retention=120h
    - --web.listen-address=:9093
    - --web.external-url=https://alertmanager-address
    - --web.route-prefix=/
    - --cluster.peer=alertmanager-main-0.alertmanager-operated.namespace.svc:9094
    - --cluster.peer=alertmanager-main-1.alertmanager-operated.namespace.svc:9094
    - --cluster.peer=alertmanager-main-2.alertmanager-operated.namespace.svc:9094
```

and yes, fresh after hard reset of all instances in the same time it starts with three peers. But previously it has cluster formed along  with all other alertmanager installations in our cluster (over 50 instances I guess ?)


So, from what you are saying, if during those 6 hours, ip gets re-assigned to completely different pod, AM will try to form cluster with this new pod ?

simonpasquier:
[...]

PR #3254, opened 2023-02-15
Support setting the Gossip cluster "label"

greed42:
This is an alternate mechanism for isolating Alertmanager clusters without having to set up the right components of TLS.

It should solve issues such as <https://github.com/prometheus/alertmanager/issues/2250>, although enabling this feature will lead to loss of non-persisted state. (For example, if you rely on alertmanager cluster peering to maintain silences instead of using persistent volume storage in Kubernetes.) The Gossip label serves as the "cluster ID" idea mentioned in <https://github.com/prometheus/alertmanager/issues/2250#issuecomment-961933082>.

You can enable with the command-line flag, `--cluster.gossip-label`; any non-empty string will form an effective namespace for gossip communication.

If you use Prometheus Operator, you can set the `ALERTMANAGER_CLUSTER_GOSSIP_LABEL` environment variable (as Prometheus Operator does not have a way of adding additional command-line flags). You would need to modify your Alertmanager object something like:

```
kind: Alertmanager
...
spec:
  ...
  containers:
    - name: alertmanager
      env:
        - name: ALERTMANAGER_CLUSTER_GOSSIP_LABEL
          value: infrastructure-eu-west-2
  ...
```

This is low-security mechanism, suitable for use with Alertmanager configuration where anyone can add or remove a silence. It protects against surprising cluster expansion due to IP:port re-use.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -142,6 +142,7 @@ func Create(
 	probeInterval time.Duration,
 	tlsTransportConfig *TLSTransportConfig,
 	allowInsecureAdvertise bool,
+	gossipLabel string,
 ) (*Peer, error) {
 	bindHost, bindPortStr, err := net.SplitHostPort(bindAddr)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


simonpasquier:
Superseded by #3354

PR #3256, opened 2023-02-16
Implement Stringer interface for Pairs and KVs

grobinson-grafana:
This commit implements the `Stringer` interface for Pairs and KVs. It changes how Pairs are printed in templates from `map[name1:value1 name2:value2]` to `name1=value1, name2=value2`. KVs work similar, but are first sorted into pairs before being printed.

This should mean that Alertmanager has better text when using `{{ .Annotations }}` and `{{ .Labels }}`, as an example, without having to write a `range` with custom delimitation.

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -187,6 +187,19 @@ func (ps Pairs) Values() []string {
 	return vs
 }
 
+func (ps Pairs) String() string {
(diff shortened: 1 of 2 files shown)


gotjosh (approved):
LGTM

PR #3260, opened 2023-02-24
fix gracefully-exiting stuck due to adding one more for `WaitGroup`

chengjoey:
In the main function, `WaitGroup` done twice, but add three times, so it will get stuck when exiting.

Currently, `AlertManager` cannot gracefully-exiting, because [wg.Wait()](https://github.com/prometheus/alertmanager/blob/6450c9ff18a4f7a6ebcefa83e157e80485014a4b/cmd/alertmanager/main.go#L333) for two done, but added three times

Signed-off-by:  chengjoey <zchengjoey@gmail.com>

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -277,7 +277,6 @@ func run() int {
 
 	stopc := make(chan struct{})
 	var wg sync.WaitGroup
-	wg.Add(1)
 
 	notificationLogOpts := nflog.Options{
 		SnapshotFile: filepath.Join(*dataDir, "nflog"),


chengjoey:
bug detail:
image
Here `wg.Wait()` is stuck and cannot end because it is added once more

after fix:
image

dgl (approved):
Note there's not an assignment bot in the prometheus org, also not sure where you found me, I'm not a maintainer on alertmanager.

The Add vs Done are indeed mismatched, so approving.

chengjoey:
> Note there's not an assignment bot in the prometheus org, also not sure where you found me, I'm not a maintainer on alertmanager.
> 
> The Add vs Done are indeed mismatched, so approving.

thanks @dgl , i was found you on readme doc [contributing](https://github.com/prometheus/alertmanager/blob/main/README.md#contributing), this is the same as the prometheus repository

tanks again for review

chengjoey:
@gotjosh

PR #3262, opened 2023-02-28
Initiate the React application with Status page

Nexucis:
Hello there,

Here is the PR that initiate a React application that would replace at some point the current UI.
The ui is available with the route prefix `/react-app`.

Hope you will be ok with this PR.

Once this PR is merged, we can migrate one by one each page and each feature and certainly make it look better as I'm not an expert in UX design and UI in general :) 

Here is the current display of the status page as a first draft:
image


gotjosh:
This is so nice @Nexucis! THANK YOU SO MUCH.

 Let me loop in @gillesdemey so that he can give you a more accurate review on this.

Can you please let me know what's your plan with all of this? I haven't fully reviewed this yet but from what I can see the UI doesn't look the same as Prometheus does that mean that we're using a different framework? I think we use bootstrap in Prometheus right?

Nexucis:
> Can you please let me know what's your plan with all of this? I haven't fully reviewed this yet but from what I can see the UI doesn't look the same as Prometheus does that mean that we're using a different framework? I think we use bootstrap in Prometheus right?

Yeah I'm trying a different framework (Material UI). If it nice enough Prometheus will follow ;). And maybe finally we will go away from every security and dependency issues.

The goal here is just to set up an initial React-App that would be accessible using the path `/react-app`. It will be the case until every pages are implemented.
Once it's done and there is no particular issue with the new UI, we will remove the old one and replace it by the react-app.

Is it clearer saying like that ?

Nexucis:
> Oh, I wasn't aware. Can you elaborate? What kind of issues?

Well today it's super hard to upgrade the UI dependencies in Prometheus. But it's probably more because we are using `react-create-app` than because of Bootstrap.

But `Bootstrap` is also quite annoying and old in the design.

> I'm a bit reluctant to diverge from Prometheus to be honest. We took the same approach with Elm and it set us back from the rest of the ecosystem in the Alertmanager. I'm not keen on taking the same route... 😭

yeah I agree on that. But I would push to move to `react-mui` as I have the feeling it's easier to make a simple UI with `mui`.
And after, I will perform the same migration to PromLens and Prometheus at the end.

Prometheus is a bit a special case. It requires to change many thing to finally have something easy to maintain. Like getting ride of JQuery. 
That's also why we will try to use the component from `Perses` in `Prometheus` that is also using `mui`. That's also a reason I would like to get ride of Bootstrap. Like that we will have a consistent framework across every Prometheus UI.

Probably Thanos will follow at some point. But it's another topic.

ui/react-app/src/components/navbar.tsx
```diff
@@ -0,0 +1,49 @@
...
+                marginRight: '1rem',
+                color: theme.palette.common.black,
+              })}
+            >
+              AlertManager
+            </Typography>
+          </Button>
+          <Button variant="text">Alerts</Button>
+          <Button variant="text">Silences</Button>
+          <Button
```
gillesdemey, line 31: I feel like this one definitely needs an indication when the route is active, unfortunately I couldn't figure out how to combine react-router's `<NavLink />` with mui's `<Button />` without jumping through a few hoops.

We can use the `matchPath` function from `react-router` or build a custom `NavButton` component.

ui/react-app/src/views/ViewStatus.tsx
```diff
@@ -0,0 +1,111 @@
...
+      </TableCell>
+      <TableCell>{content}</TableCell>
+    </TableRow>
+  );
+}
+
+export default function ViewStatus() {
+  const { data } = useAMStatus();
+  if (data === undefined || data === null) {
+    return <></>;
```
gillesdemey, line 48: Any particular reason why we're returning an empty fragment here iso `null`?

ui/react-app/src/utils/url-builder.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export default function buildURL(params: URLParams): string {
+  let url = params.apiPrefix === undefined ? apiPrefix : params.apiPrefix;
+  url = `${url}/${params.resource}`;
+
+  if (params.queryParams !== undefined) {
+    url = `${url}?${params.queryParams.toString()}`;
+  }
+  return url;
+}
```
gillesdemey, line 17: I feel like we can simplify this a tad by using de-structuring with a default value assignment (assuming `apiPrefix` was renamed to `API_PREFIX` to denote a const).

```suggestion
export default function buildURL({ apiPrefix = API_PREFIX, resource, queryParams }: URLParams): string {
  let url = `${apiPrefix}/${resource}`;

  if (queryParams !== undefined) {
    url = `${url}?${queryParams.toString()}`;
  }

  return url;
}
```

gillesdemey (commented):
Really solid foundation, made some comments (nits, really) where I felt I could contribute :)

ui/react-app/src/components/navbar.tsx
Nexucis: > I feel like this one definitely needs an indication when the route is active

By an indication do you mean like the button is deactivated when its route is active ?

ui/react-app/src/views/ViewStatus.tsx
Nexucis: not particularly reason :). I will change it by `null`

ui/react-app/src/utils/url-builder.ts
Nexucis: ah yeah indeed. Thanks for proposition !

ui/react-app/src/components/navbar.tsx
gillesdemey: Yea some visual indication of the active route so users know where they are. You could play around with the `variant` prop or just wrap the contents of the button with `<strong>` so long as there is some visual distinction between an inactive button and an active one.

Nexucis: I deactivated the button based on the location. Let me know what do you think about this way :).

gillesdemey: I'm not sure if using the `disabled` prop is the best way to go about it since it has a certain semantic meaning and users with screen readers might be confused about a navigation item being disabled :)

My suggestion here would be to simply make the text of the button bold but we can move forward with what we have and make another pass at the navigation later since we're only in the very early setup stages here 👍

gillesdemey (approved):

gotjosh:
Stupid question @gillesdemey @Nexucis - how do we feel about frameworks? e.g. [Next.js](https://nextjs.org/)

I've got a round of feedback to write up but haven't found the time to put it together. I want to start the conversation with the "why not a framework argument?

juliusv:
I have never used Next.js, but it reads a bit like it prefers to be used with a Node.js backend for all the SSR and similar features. Static HTML exports are only listed as an "advanced feature" (https://nextjs.org/docs/advanced-features/static-html-export), and that turns off support for a bunch of Next.js features: https://nextjs.org/docs/advanced-features/static-html-export#unsupported-features. Is it still common for people to use Next.js without a Node.js backend anyway?

Overall frameworks can be great if they solve a lot of the things you want to build anyway, but they can also introduce limitations or maintenance burdens (e.g. React updates + dependency versionings can already be a pain sometimes). I'd say go for it if there's just the right framework that is stable, popular, not too complex, and solves a lot of common issues.

Nexucis:
I have never used `Next.js` as well so I don't know if it's better or not :).
I would rather keep React since at least we know it a bit better and we might share component between the different UI if it makes sense at some point.

But if we really want to go with `Next.js` why not, I just won't be able to help

gillesdemey:
As @juliusv mentioned most SSR frameworks like Next or Remix work with a Node server and some allow static HTML exports but that comes with a plethora of caveats so it seems ill-advised to go down that road.

Is there a particular reason why you wanted to explore this option @gotjosh? As far as the React ecosystem is concerned a combinations of Create-React-App + React Router (as implemented in this PR) is really all you need if server-side rendering isn't a specific requirement.

roidelapluie:
I am concerned that we do not go with bootstrap. The goal was to align with prometheus as much as possible.

juliusv:
> I am concerned that we do not go with bootstrap. The goal was to align with prometheus as much as possible.

So I think the history here is that Perses uses Material UI, it seemed nice, and @Nexucis had plans to convert Prometheus as well and potentially use Perses components there. But I'll let him speak to that :) Other than that I agree that if we start using another framework here, we will want a HARD commitment (and maybe the beginning of a code implementation?) to convert Prometheus to Material UI as well. It would be tragic to not use the Alertmanager UI reimplementation as an opportunity to make Prometheus + Alertmanager UIs consistent.

simonpasquier:
Agree with everything's that has been said. My understanding is that @Nexucis's intent is to consolidate all UIs eventually.

gotjosh:
Let's split the problem into parts because I don't want to conflate different things.

## CSS Framework

My main argument (and of others) is that we should align with what Prometheus uses. In this case, it's currently bootstrap. 

The current proposal is that we want to "try-out" Material UI for that to become the de facto standard for all of Prometheus projects.

**With this idea, I am OK**. Choosing the right tool for the job is something I'm very much in favour of, if we think that bootstrap does not offer what we need and material UI is going to replace that, let's do it - but please make sure we have a plan in place to also replace this in Prometheus so that we eventually converge.

In my opinion changing a CSS framework for a project such as the Alertmanager should not be a big (famous last words) concern. If eventually, we realise that material UI does not offer what we need, I'm sure we can correct course without incurring a complete rewrite of the front-end. 

## React Framework

We don't have a dedicated front-end maintainer in the Alertmanager. Heck, we have a couple of bugs / blockers because we're unable to do even basic Elm (I've tried).

I like frameworks because they offer opinions. I've done some react in the past, and I can find my way around React as long as I get some opinion on how things such be done/structured (think of writing a whole new page from scratch). I suggested Next.js as it seems to be popular enough (and backed by some big companies ) that it doesn't look like it's going away anytime soon (again, famous last words).

In particular, you can look at [the fundamentals page to understand](https://beta.nextjs.org/docs/routing/fundamentals) what kind of opinions I think matter:

- How routing is done
- How components are organized
- How linking and navigation is done
- Caching

But more important is documentation. With popularity, we gain the community and documentation built around these opinions and now suddenly, patterns that (non-react experts like me) can copy to achieve things.

With that said, I'm not married to the idea of a framework, and I would very much favour consistency between projects over anything else, but if **we're going to diverge massively from what Prometheus currently does, I'd very much favour a framework instead of yet another snowflake.**

## Maintenance

I want to make sure my opinions don't come across as abrasive @Nexucis as I very much appreciate the effort that you have put into this, but I also hope you can appreciate my concerns.

I want to make sure we don't shoot ourselves in the foot by writing something that's extremely custom and then having you spread too thin across all of the projects.

It's unrealistic of me to expect that whenever issues arise I ping you to solve them for me and I'd like to contribute to this in whichever way I can.  

My promise to you is that once we have an agreement on what we're going to do, this will move fairly quickly 😄

Nexucis:
My wish is indeed to have something consistent between PromLens, Prometheus, AlertManager and eventually Perses.

I started with AlertManager because it’s the opportunity to start with a fresh stack with no technical debt and so one.
That’s also one I didn’t use `react-create-app` so we have no security issues when installing the npm dependencies. And also the upgrade of the dependencies are not blocked at all by `react-create-app` which I find it super annoying.

Talking about technical debt, it’s been a while I’m thinking how to remove the usage of `jQuery` in prometheus which shouldn’t be used anymore when `React` is involved. That doesn’t make a lot of sense to me.
But it’s not that easy to remove because it involves the display of the graph. That’s where potentially Perses will help. By using some components from Perses we might be able to actually remove the usage of `jQuery` and well a nicer display of the graph (`nicer` will depend on everyone taste of course). Which will reduce a lot the code in Prometheus (so less thing to maintain).

Perses like @juliusv said is using Material UI. So to avoid a display conflict I would go for Material UI so we don’t have an issue around that.

@sjcobb (who works on Perses) made a proof of concept around this topic available here: https://github.com/sjcobb/prometheus/blob/sjcobb/perses-graph-view-refactor/web/ui/react-app/src/components/perses/PersesQueryPanel.tsx

prometheus_ui_perses_panel

Once the work is done in AM, I would like to made some changes in PromLens to move as well to Material UI but also to export it as a component that we can reuse in Prometheus, so we are able to use/display PromLens directly in the Prometheus UI.

Starting with PromLens will highlight the steps for the migration since it’s more or less the same base of code than Prometheus. It’s just simpler and we might be able to break things if necessary.
Also it will be a good playground to see in live the integration of the Perses components.

The keys of these changes are : 
- Reducing the technical debt in the UI
- Having the same stack between each UI
- Potentially being able to share component between each of them. For example, PromLens component in Prometheus, Prometheus Alert display in AlertManager …etc.
- Avoiding the pain of the dependencies upgrade we are facing today in the UI

Finally regarding what kind of Typescript we should choose, if we based our argument on the popularity, `React` is beyond `Angular` and `View`. (based on some googling search like : https://www.browserstack.com/guide/angular-vs-react-vs-vue)
Moving with View will require much more work from my opinion than just replacing `Bootstrap` by `Material UI`. Which is already a huge work in Prometheus.
Moving from React to View in Prometheus is (again from my opinion) too much work. Besides I have the feeling that React is the JS/TS framework used in the Prometheus ecosystem. That’s also why in Perses we are using `React` and not `Angular` or `View`. To be closer to the Prometheus Ecosystem.

Besides in Prometheus, I'm not the only one that knows React. A lot of people from Grafana know it and @juliusv as well :).
I think the big question would be: 
**Does someone that is a maintainer of AlertManager is able to write some code using `ViewJS` ?**

Nexucis:
same argument for `NextJS` even if I thought we were talking about `ViewJS` 😓

gillesdemey:
> If eventually, we realise that material UI does not offer what we need, I'm sure we can correct course without incurring a complete rewrite of the front-end.

This is somewhat false since most design systems come with an opinionated set of components and rarely have I seen design systems that have compatible interfaces for what might be a visually similar component. It won't be a _complete_ rewrite but does take up a bunch of time unless anyone is familiar with codemods.

I personally believe it's important that we commit to one up front as opposed to changing our minds about this later.

sjcobb:
Hey @juliusv, [this branch](https://github.com/sjcobb/prometheus/pull/1) has the beginning of a code implementation to convert the Prometheus UI to MUI. I added some commits to the prototype @Nexucis mentioned. There’s now a MUI theme provider and I began rewriting the header (moved the Perses graph view to a hidden URL).

juliusv:
I just had a good long chat with @Nexucis not only about this PR, but also the long-term plans around Perses, PromLens, and Prometheus, and getting them all onto some consistent new UI. Overall I feel like we should give this a shot with Material UI and React, but keep it **experimental** for a while, alongside the old (main, default) UI, just as this PR does. This is also how I did it in Prometheus when I introduced the React UI.

That way we can at least see how it goes, and Material UI + React are definitely solid and very popular choices that improve both over Bootstrap, as well as Elm (in terms of contributor accessibility). There's of course no hard guarantee that all components are converted on a volunteer basis, but it's at least an aspiration. Given how little is otherwise happening in terms of UI contributions, I think it's worth trying it if @Nexucis feels motivated to do so. And overall I support the conversion of Prometheus (and PromLens) to Material UI as well.

@gotjosh Would that be ok with you to try this in parallel to the old UI?

gotjosh:
@juliusv absolutely! To be clear, I was never opposed to this work.

My only ask is that we use [NextJS](https://nextjs.org/) - I think the project can truly benefit from a framework and my understanding is that Material UI + React fit perfectly into NextJS.

It feels like the components already developed as part of this PR can be reused, we just need to fit them into the framework.

juliusv:
@gotjosh I started reading more NextJS documentation, and the impression I had in https://github.com/prometheus/alertmanager/pull/3262#issuecomment-1457925910 is not really going away yet (but I could be wrong). It looks really solid and nice, but it's also selling its tight integration with server-side rendering logic as one of its main features on basically every page. For example, data fetching: https://nextjs.org/docs/app/building-your-application/data-fetching/fetching-caching-and-revalidating - it talks about 4 ways, 2 of which are server-side (not usable by us), 1 of which is client-side and NextJS-specific (I don't understand yet if it brings us tangible benefits) and another 1 is just third-party libraries again (same as what we're currently doing). Similarly, the [routing docs](https://nextjs.org/docs/app/building-your-application/routing#the-app-router) talk a lot about [Server Components](https://nextjs.org/docs/getting-started/react-essentials#server-components) and how and when to use Server Components vs. Client Components and the thought models and boundaries - all stuff that feels like extra mental stuff to juggle that wouldn't be usable by us with a Go backend that has a static site compiled in.

I must admit that I don't understand much about NextJS yet though, so this is just a first impression. If someone knows NextJS better and can give more of an informed opinion of whether it makes sense to use over plain React in a static server backend scenario, it would actually be very interesting to hear! I just don't want us to pick it because it's a framework that works well for other scenarios. Right now I have a suspicion that introducing NextJS would mean having to learn a lot of new things that we're not familiar with yet, and I'm not sure we'd get a lot of the benefits of it.

gotjosh:
> I must admit that I don't understand much about NextJS yet though, so this is just a first impression

To be honest, I'm on the same boat. My preference for a framework comes from my Ruby on Rails background - most people will tell you it's bad, but for quickly developing what you need, it's perfect.

As I said before, we don't have dedicated active front-end maintainers in the Alertmanager and having to learn a custom React framework (you know how it goes with the flexibility of how components and routers are organized) seems extremely daunting for me - learning NextJS (and its added benefit of [having a place for everything](https://nextjs.org/docs/getting-started/project-structure)) seems less daunting than trying to understand custom React (but maybe I'm wrong?).

That being said, I think (?) this is nothing that documentation cannot solve? If you and @Nexucis feel that the current approach is the way to go and it's going to be homogeneous for all prom-related projects, then let's document the proposed structure + where goes what in order to move forward.

We've been stuck for a while here, and I would love to avoid the bystander effect - as such if you still feel strongly against nextJS, can we agree on the documentation bit?

juliusv:
@gotjosh Yes, documenting the overall structure and custom bits sounds very reasonable to me! At least before this becomes the default UI for Alertmanager.

Hopefully to allay worries a bit more even, the differences between Ruby on Rails vs. just Ruby are much larger in terms of abstraction level and built-in functionality than the difference between React and Next.js, with React almost already being a framework in itself that has a lot of pretty standard ways of doing things. Like for routing, this PR is just using the very standard `react-router`, which every React developer will know about. Generally I find those things relatively similar and easy to pick up on across React apps. The things that scare and confuse me more in any modern web app setup are the build and module infrastructure bits (webpack, eslint, typescript, etc., and building all those together correctly with weird different module standards and stuff that doesn't always fit together 😅 - @Nexucis understands more about that). Luckily most of the time that doesn't need to be touched on an ongoing basis...

@Nexucis Do you want to rebase the PR to get rid of conflicts so we can review in detail?

Nexucis:
Hey @gotjosh @juliusv 

I resolved the conflict so this PR is ready again to be review.

Concerning the question about NextJS let's give it a try and let's see if it brings enough things to be used. For the moment introducing a framework "just" for the routing sounds a bit light but perhaps there are other things that are solved with NextJS that are annoying without it.
I'm not really sure I will look at it really soon but let's see. For the moment I will more focus on adding other pages to prepare a bit a nice communication during the PromCon

juliusv:
@Nexucis Thanks, looks good to me for a first merge! Of course the UI is still ugly, but that can be fixed later. Making any kind of progress is the most important thing for now IMO :)

@gotjosh How would you like to handle reviews purely around the new experimental (and still hidden) UI? Is a :+1: from me sufficient at the moment for that, or would you always like an Alertmanager maintainer to give a :+1: as well?

ui/react-app/src/Router.tsx
```diff
@@ -0,0 +1,17 @@
+// Other routes are lazy-loaded for code-splitting
+import { Suspense, lazy } from 'react';
+import { Route, Routes } from 'react-router-dom';
+
+const ViewStatus = lazy(() => import('./views/ViewStatus'));
+
+function Router() {
+  return (
+    <Suspense>
```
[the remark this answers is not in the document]
Nexucis: I found that: https://stackoverflow.com/questions/63624950/react-suspense-lazy-loading-without-fallback
And it looks like it's ok to do it without a fallback indeed.
I think I took it from Perses (mainly) where we should have a fallback (like the home page). So probably when we will implement the silence page we will have the fallback then

gotjosh:
> @gotjosh How would you like to handle reviews purely around the new experimental (and still hidden) UI? Is a 👍 from me sufficient at the moment for that, or would you always like an Alertmanager maintainer to give a 👍 as well?

I'd say that if you don't get a response from a maintainer in 24h-48h (use best judgement on time frame to pick e.g. allow 48h for more complex pages) you're free to merge. I'll be pretty good with reviews for the next few months but want to make sure you're not blocked in case I take a while.

Once we're not hidden (or when we're ready to make the cut-over) that's a different story as you mentioned!

juliusv:
@gotjosh Sounds good to me! Will leave this PR open for your comments / :+1: then for now as well.

Nexucis:
looking at the different comment / review, I think we should be good to merge this one right @gotjosh @juliusv ?

PR #3278, opened 2023-03-03
Fix corruption in templates that use title function

### LINKED ISSUE #3277: Race condition in ExecuteTextString
**What did you do?**

I have been looking into a data corruption issue in Slack notifications in Grafana (https://github.com/grafana/grafana/issues/63584) and have found a race condition in `ExecuteTextString` in Alertmanager that I suspect to be the cause. I have been able to reproduce this bug in `main`, commit `41eb121`.

I have the following Alertmanager configuration:

```
receivers:
- name: test
  slack_configs:
  - api_url: redacted
    title: '{{ template "slack.title" . }}'
route:
  receiver: test
  group_by: ['...']
  group_wait: 5s
  group_interval: 10s
  repeat_interval: 15s
templates:
  - templates/*.tmpl
```

and the following template in `test.tmpl`:

```
{{ define "slack.title" }}
  {{ reReplaceAll ":" "-" .ExternalURL | title }}
{{ end }}
```

I created a single rule that has 4 alert instances:

Screenshot 2023-03-03 at 15 34 44

I expected to see the following in all messages:

```
Http-//Georges-Air.fritz.box-9093
Http-//Georges-Air.fritz.box-9093
Http-//Georges-Air.fritz.box-9093
Http-//Georges-Air.fritz.box-9093
```

However most messages are corrupted. For example:

```
HHtt//EegeeS-Air..tz.box-9093
Http-//Georges-Air.fritz.box-
Http-//Georges-Air.fritz.box-9093
Ht
```

I think the issue is related to how Alertmanager uses `text/template` across goroutine as the bug is no longer reproducible with the following changes:

```
diff --git a/template/template.go b/template/template.go
index a42cf03d..9e56bf1f 100644
--- a/template/template.go
+++ b/template/template.go
@@ -23,6 +23,7 @@ import (
        "regexp"
        "sort"
        "strings"
+       "sync"
        tmpltext "text/template"
        "time"

@@ -38,6 +39,7 @@ import (
 type Template struct {
        text *tmpltext.Template
        html *tmplhtml.Template
+       mu   sync.Mutex

        ExternalURL *url.URL
 }
@@ -131,6 +133,8 @@ func (t *Template) FromGlob(path string) error {

 // ExecuteTextString needs a meaningful doc comment (TODO(fabxc)).
 func (t *Template) ExecuteTextString(text string, data interface{}) (string, error) {
+       t.mu.Lock()
+       defer t.mu.Unlock()
        if text == "" {
                return "", nil
        }
```

grobinson-grafana:
This commit fixes data corruption in templates that use the title function as it used a shared cases.Title when casers should not be shared between goroutines. When templates that used the title function were executed at the same time, data corruption would occur in the text that was being cased. This is fixed using a separate caser for each function call.

Here is an example of the issue as explained [here](https://github.com/prometheus/alertmanager/issues/3277#issuecomment-1453800831):

```
2023/03/03 16:40:51 Waltz, BaD
2023/03/03 16:40:51 Waltz, Bad Nymph, For Quick Jigs Vex
2023/03/03 16:40:51 Waltz, Bad Nymph, For Quick Jigs Vex
2023/03/03 16:40:51 Waltz, Bad Nymph, For Quick Jigs Vex
2023/03/03 16:40:51 Waltz, Bad Nymph, For Quick Jigs Vex
2023/03/03 16:40:51 Waltz, Bad Nymph, For Quick Jigs Vex
2023/03/03 16:40:51  Nym
2023/03/03 16:40:51 Wph, For icc Jigs Vex
2023/03/03 16:40:51 Waltz, Bad Nymph, For Quick Jigs Vex
2023/03/03 16:40:51 Waltz, Bad Nymph, For Quick Jigs Vex
```

and with the fix:

```
2023/03/03 21:33:15 Waltz, Bad Nymph, For Quick Jigs Vex
2023/03/03 21:33:15 Waltz, Bad Nymph, For Quick Jigs Vex
2023/03/03 21:33:15 Waltz, Bad Nymph, For Quick Jigs Vex
2023/03/03 21:33:15 Waltz, Bad Nymph, For Quick Jigs Vex
2023/03/03 21:33:15 Waltz, Bad Nymph, For Quick Jigs Vex
2023/03/03 21:33:15 Waltz, Bad Nymph, For Quick Jigs Vex
2023/03/03 21:33:15 Waltz, Bad Nymph, For Quick Jigs Vex
2023/03/03 21:33:15 Waltz, Bad Nymph, For Quick Jigs Vex
2023/03/03 21:33:15 Waltz, Bad Nymph, For Quick Jigs Vex
2023/03/03 21:33:15 Waltz, Bad Nymph, For Quick Jigs Vex
```

Fixes #3277

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -171,8 +171,11 @@ var DefaultFuncs = FuncMap{
 	"toUpper":   strings.ToUpper,
 	"toLower":   strings.ToLower,
 	"trimSpace": strings.TrimSpace,
-
-	"title": cases.Title(language.AmericanEnglish).String,
+	"title": func(text string) string {
+		// casers should not be shared between goroutines, instead
+		// create a new caser each time this function is called
+		return cases.Title(language.AmericanEnglish).String(text)
+	},
 	// join is equal to strings.Join but inverts the argument order
 	// for easier pipelining in templates.
 	"join": func(sep string, s []string) string {
--- a/template/template.go
+++ b/template/template.go
@@ -168,14 +168,14 @@ func (t *Template) ExecuteHTMLString(html string, data interface{}) (string, err
 type FuncMap map[string]interface{}
 
 var DefaultFuncs = FuncMap{
-	"toUpper":   strings.ToUpper,
-	"toLower":   strings.ToLower,
-	"trimSpace": strings.TrimSpace,
+	"toUpper": strings.ToUpper,
+	"toLower": strings.ToLower,
 	"title": func(text string) string {
 		// casers should not be shared between goroutines, instead
 		// create a new caser each time this function is called
 		return cases.Title(language.AmericanEnglish).String(text)
 	},
+	"trimSpace": strings.TrimSpace,
 	// join is equal to strings.Join but inverts the argument order
 	// for easier pipelining in templates.
 	"join": func(sep string, s []string) string {
--- a/template/template_test.go
+++ b/template/template_test.go
@@ -16,6 +16,7 @@ package template
 import (
 	tmplhtml "html/template"
 	"net/url"
+	"sync"
 	"testing"
 	tmpltext "text/template"
 	"time"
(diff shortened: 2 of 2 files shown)


gotjosh (approved):
LGTM

Can you please commit the test case you used? Ideally, we iterate trough the list of functions and execute the test for each.

template/template.go
```diff
@@ -170,7 +170,11 @@ type FuncMap map[string]interface{}
 var DefaultFuncs = FuncMap{
 	"toUpper": strings.ToUpper,
 	"toLower": strings.ToLower,
-	"title":   cases.Title(language.AmericanEnglish).String,
+	"title": func(text string) string {
+		// casers should not be shared between goroutines, instead
+		// create a new caser each time this function is called
+		return cases.Title(language.AmericanEnglish).String(text)
```
gotjosh, line 176:
```suggestion
		// Casers should not be shared between goroutines, instead
		// create a new caser each time this function is called.
		return cases.Title(language.AmericanEnglish).String(text)
```

simonpasquier (approved):

taisph:
Is a fix release coming? I've started seeing alert emails with weird or missing subjects and today even an empty email that seem to have been cut off at the subject header. I'll remove the use of `title` as a workaround in the meantime.

gotjosh:
It is; we're planning to regularize the release schedule.

PR #3285, opened 2023-03-07
Metrics: Silence maintenance success and failure

gotjosh:
Due to various reasons, we've observed different kind of errors on this area. From read-only disks to silly code bugs. Errors during maintenance are effectively a _data loss_ and therefore we should encourage proper monitoring of this area.

This PR Introduces a total and failure metric for silence maintenance. If agreed, I'll do the same for the nflog and fix the flaky test like I did for silences while I'm there.


silence/silence.go
```diff
@@ -409,7 +421,9 @@ Loop:
 		case <-stopc:
 			break Loop
 		case <-t.C:
+			s.metrics.maintenanceTotal.Inc()
```
simonpasquier, line 424: should we move the counter updates to runMaintenance()?

gotjosh: Good shout, I considered neglected the idea and now that you mention it - I think it's better.e

gotjosh:
Any other thoughts on this @simonpasquier?

simonpasquier (approved):
:+1:

PR #3289, opened 2023-03-08
Add receiver name to the alerts page

roidelapluie:
cc @gotjosh


gotjosh (approved):
LGTM!

This is great, thank you so much 🙇

gotjosh:
Screenshot 2023-03-08 at 18 23 21

roidelapluie:
I have improved this a lot. Can you test again?

gotjosh:
Aesthetically looks much better 👌 

Two things:

- The alert list will now look like a staircase because it is dependant on the name of the receiver 😭 I'd rather have it on the other side next to the alert count.
- The whitespace to indicate the indication when you open the group is a bit big

Neither of us are Elm experts so I'm more than happy to proceed with this as is and improve later as this is already a big step up from what we got 👍

roidelapluie:
What do you mean about The alert list will now look like a staircase because it is dependant on the name of the receiver ?

roidelapluie:
Do you mean the labels are unaligned?

gotjosh:
> The alert list will now look like a staircase because it is dependent on the name of the receiver 😭 I'd rather have it on the other side next to the alert count.

Notice how the "team-db-pager" receiver it's just a few extra letters more than the others, in the case where your receiver name letter count is varied, the indentation will look all over the place. Whereas before, every would look uniform as part of the pills. 

Screenshot 2023-03-09 at 14 01 52

roidelapluie:
My recommendation is that we use this implementation for now and consider any future improvements later. Currently, the page isn't completely informative and adding some labels may assist, which would move labels around anyway.

gotjosh:
I 100% agree agree with that sentiment

> Neither of us are Elm experts so I'm more than happy to proceed with this as is and improve later as this is already a big step up from what we got 👍

gotjosh (approved):
LGTM

PR #3299, opened 2023-03-17
Github actions

jan--f:
This is a first draft to move from CircleCI to Github Actions.
Generally I tried to trigger workflows only for relevant changes, e.g. the mixin-lint is only triggered if a file in the mixins change.

Two things are tbd:

1. This doesn't yet store the frontend asset tarball anywhere. I wasn't exactly sure how this is currently handled in the CircleCI setup, any pointer are appreciated.
2. ~This currently uses my fork of `maildev` at https://github.com/jan--f/maildev/tree/hide-starttls-when-user-password, which simply re-enables `STARTTLS`, which was disabled in [v1.1.1](https://github.com/maildev/maildev/pull/276/). There is a [PR](https://github.com/maildev/maildev/pull/435) up that looks promising in order for Alertmanager to switch back to upstream `maildev`. The current PR version doesn't work, since I believe it no longer sets up self-signed certificates. Otoh we could simply create those certs ourselves and pass them via the proposed arguments.~ This still requires a fix to maildev, see https://github.com/maildev/maildev/pull/469. So I'll keep my forked image in here for now.

Looking forward to any feedback.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,41 +17,14 @@ jobs:
       - run: git diff --exit-code
 
   test:
-    name: Test with maildev
+    name: Test
     runs-on: ubuntu-latest
     # Whenever the Go version is updated here, .promu.yml
     # should also be updated.
     container:
       image: quay.io/prometheus/golang-builder:1.19-base
-    services:
-      # run two maildev containers for auth and noauth
-      maildev-noauth:
-        # image: docker.io/maildev/maildev
-        image: quay.io/jfajersk/prom-maildev:latest
-      maildev-auth:
-        # image: docker.io/maildev/maildev
-        image: quay.io/jfajersk/prom-maildev:latest
-        env:
-          MAILDEV_INCOMING_USER: user
-          MAILDEV_INCOMING_PASS: pass
-          MAILDEV_INCOMING_SECURE: true
-          MAILDEV_HIDE_STARTTLS: false
-    env:
-      EMAIL_NO_AUTH_CONFIG: /tmp/smtp_no_auth.yml
-      EMAIL_AUTH_CONFIG: /tmp/smtp_auth.yml
     steps:
       - uses: actions/checkout@v3
       - uses: prometheus/promci@v0.0.2
       - uses: ./.github/promci/actions/setup_environment
-      - run: |
-          cat <<EOF > "$EMAIL_NO_AUTH_CONFIG"
-          smarthost: maildev-noauth:1025
-          server: http://maildev-noauth:1080/
-          EOF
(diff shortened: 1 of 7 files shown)


jan--f:
cc @roidelapluie

simonpasquier:
> This doesn't yet store the frontend asset tarball anywhere. I wasn't exactly sure how this is currently handled in the CircleCI setup, any pointer are appreciated.

The test_frontend job will persist the tarball in the workflow's workspace at the same directory (e.g. `.tarballs` ) where the binaries will be generated. When running `promu release ...`, everything will be uploaded to the release page (https://github.com/prometheus/alertmanager/releases/tag/v0.26.0).

I believe that we need to use `./.github/promci/actions/save_artifacts` and `./.github/promci/actions/restore_artifacts`.

jan--f:
This is currently blocked by https://github.com/actions/runner/issues/2783.

jan--f:
> I believe that we need to use `./.github/promci/actions/save_artifacts` and `./.github/promci/actions/restore_artifacts`.

Thanks! I added `./.github/promci/actions/save_artifacts` to the `test_frontend` job. `./.github/promci/actions/restore_artifacts` is then used by the `./.github/promci/actions/publish_[main|release]` actions in the `publish` and `release` workflows.

gotjosh:
Hey @jan--f can you walk me through the how is this blocked by https://github.com/actions/runner/issues/2783?

jan--f:
@gotjosh sure thing. This intends to run [maildev as](https://github.com/prometheus/alertmanager/pull/3299/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR26-R38) a [service containers](https://docs.github.com/en/actions/using-containerized-services/about-service-containers).
If you look at the [action runs in my clone](https://github.com/jan--f/alertmanager/actions/runs/6049606082/job/16612928173), the maildev related tests complain about not being able to resolve the service container. according to the docs, this should work exactly as specified though. https://github.com/actions/runner/issues/2783 describes exactly that.

gotjosh:
@jan--f I think we shouldn't block this migration on these tests.

My suggestion is that you can add a new make command that runs all of the excepts except the ones in `email_tests.go` (we'll run both test suites in parallel for some time before cutting over anyways) and I'll be happy to change the tests to run the container as part of the tests directly (this is something we rely on heavily in mimir).

What do you think?

jan--f:
@gotjosh yeah that sounds good. Happy to take stab at pulling maildev into the test myself if you're not already working on it. We probably still want https://github.com/maildev/maildev/pull/469 though? Or simply keep using the ancient version like in the existing tests?

gotjosh (approved):
LGTM

PR #3300, opened 2023-03-20
replace github.com/gofrs/uuid with github.com/google/uuid

DanCech:
While cleaning up dependencies in https://github.com/grafana/grafana we noticed that `github.com/gofrs/uuid` is being brought in as an indirect dependency of `prometheus/alertmanager`.  We can easily swap that out for `github.com/google/uuid` which is used by a number of the other modules both projects depend on, and remove the dependency on `github.com/gofrs/uuid` altogether.

See also grafana/grafana#65016


SuperQ:
Would you mind rebasing this?

DanCech:
outdated by #4765, closing.

PR #3307, opened 2023-03-28
add reason code to notifiers

qinxx108:
this uses the new error with reason to determine based on status code what the reason is for the notifiers:
* Pagerduty
* Opsgenie
* Pushover
* Victorops
* Webhook
* Wechat


simonpasquier (approved):

simonpasquier:
Thanks!

PR #3320, opened 2023-04-07
if 9093/9094 port is in use, test case will fail

rongyi:
change hardcode port to random available port

### DIFF
--- a/cluster/tls_transport_test.go
+++ b/cluster/tls_transport_test.go
@@ -29,11 +29,11 @@ import (
 
 var logger = log.NewNopLogger()
 
-func freeport() int {
+func freeport() uint16 {
 	lis, _ := net.Listen(network, "127.0.0.1:0")
 	defer lis.Close()
 
-	return lis.Addr().(*net.TCPAddr).Port
+	return uint16(lis.Addr().(*net.TCPAddr).Port)
 }
 
 func newTLSTransport(file, address string, port int) (*TLSTransport, error) {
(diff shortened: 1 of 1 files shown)


dswarbrick:
Finally! This has been the bane of packaging Alertmanager for a long time. The Debian autopkgtests fail if we do not patch the test to avoid port 9093, since the tests are executed in an environment where the very same Alertmanager is running and has bound to that port.

Just one point the sticks out however - wouldn't it make more sense to use uint16 for port variables, since they cannot be anything other than unsigned 16-bit values anyway?

gotjosh:
Thank you very much for the contribution - any chance you can address @dswarbrick's concerns and fix the DCO?

rongyi:
thanks for your kindly suggestions, @gotjosh  @dswarbrick

cluster/tls_transport_test.go
```diff
@@ -29,7 +29,16 @@ import (
 
 var logger = log.NewNopLogger()
 
+func freeport() int {
```
dswarbrick, line 32: Wouldn't it make more sense for this function to return a uint16?

cluster/tls_transport_test.go
```diff
@@ -89,7 +99,7 @@ func TestFinalAdvertiseAddr(t *testing.T) {
 			require.Nil(t, err)
 			if tc.expectedPort == 0 {
 				require.True(t, tc.expectedPort < port)
-				require.True(t, uint32(port) <= uint32(1<<32-1))
+				require.True(t, uint16(port) <= uint16(1<<16-1))
```
dswarbrick, line 102: This assertion is irrelevant, since casting a var to uint16 will _by definition_ ensure that it is <= 1<<16-1.

gotjosh:

Issue #3319, opened 2023-04-04
Support any symbol in label names


yuri-tceretian:
Currently, Alertmanager supports only valid Prometheus label names, i.e. the ones that match the following [regular expression](https://github.com/prometheus/common/blob/main/model/labels.go#L94)  `^[a-zA-Z_][a-zA-Z0-9_]*$`. 
This is a quite restrictive pattern and limits usage of the Alertmanager outside the Prometheus eco-system. 

For example. Grafana uses Alertmanager for notifications, and because Grafana supports many different data sources (like Elasticsearch, Graphite, SQL, etc) where the format of label names usually does fit into a narrow range of characters Alertmanager accepts, it [works around](https://github.com/grafana/grafana/blob/aba91d3053524f3e40f69453cf494d0975d6971f/pkg/services/ngalert/api/tooling/definitions/alertmanager.go#L744) the limitation by overriding API level and config parsing logic. 

I propose a feature request to expand the range of the allowed symbols in label names to accept any UTF-8 characters. 

Looking at the code I did not find any technical limitations that would prevent Alertmanager from working with label names that contain any UTF-8 characters, and change is fairly simple. 
The only problem spot is the parsing of Matchers https://github.com/prometheus/alertmanager/blob/432ee01c8c4f723aa438a48b6249e9e634162545/pkg/labels/parse.go#L47-L56 I think that we can support quoted label names. For example, the examples above will look like 

```
"foo"=bar,dings!=bums
{"quote"="She said: \"Hi, ladies! That's gender-neutral…\""}
```
The idea is to distinguish between matcher type symbols `(=~|=|!=|!~)` and those sequences in label names. For example
`"~=labeltest=~"=~"good label"`

siavashs:
Fixed in #3453

PR #3321, opened 2023-04-10
Support any UTF-8 string as label name

yuri-tceretian:
Currently, Alertmanager supports only valid Prometheus labels (i.e. ones that match the following regular expression `^[a-zA-Z_][a-zA-Z0-9_]*$)`. This PR expands the range of valid symbols to any symbol in UTF-8 range.

The only limitation to the label name is that it should not include _only_ whitespace symbols. 

1. It replaces all usage of Prometheus' model.LabelName method `Valid` to a new function `labels.IsValidName` that accepts `model.LabelName`.

2. Override types.Alert method `Valid` that is derived from Prometheus' model.Alert and changes validation of Labels and Annotations. The tests are copied from the Prometheus' Alert tests and expanded with a few more test cases.

3. Update ParseMatcher function that is used to parse string to `labels.Matcher`. The regular expression was updated to match any character if it is wrapped by double quotes and only Prometheus-compatible names unquoted. 

Fixes https://github.com/prometheus/alertmanager/issues/3319

Notes for reviewer: The PR can be reviewed by commit.

- [ ] Update validation in UI


grobinson-grafana:
Nice work Yuri!

I am a little concerned about adding inconsistent rules to label matchers. For example, with this PR UTF-8 is supported without double quotes for values, but not names. For example, the following matcher is accepted:

```
{foo=bar😊}
```

but this is not:

```
{foo😊=bar}
```

as it instead must be:

```
{"foo😊"=bar}
```

I don't like that there are different rules for each side of the expression. I'm interested to hear what Josh and Simon think?

docs/configuration.md
```diff
@@ -420,13 +421,15 @@ Label matchers are used both in routes and inhibition rules to match certain ale
...
-- A valid Prometheus label name.
+- A UTF-8 string enclosed in double quotes that contains at least one non-whitespace character. If string is a valid Prometheus label name then double quotes can be omitted. 
 
 - One of  `=`, `!=`, `=~`, or `!~`. `=` means equals, `!=` means that the strings are not equal, `=~` is used for equality of regex expressions and `!~` is used for un-equality of regex expressions. They have the same meaning as known from PromQL selectors.
 
-- A UTF-8 string, which may be enclosed in double quotes. Before or after each token, there may be any amount of whitespace.
+- A UTF-8 string, which may be enclosed in double quotes. Can be an empty string.
 
-The 3rd token may be the empty string. Within the 3rd token, OpenMetrics escaping rules apply: `\"` for a double-quote, `\n` for a line feed, `\\` for a literal backslash. Unescaped `"` must not occur inside the 3rd token (only as the 1st or last character). However, literal line feed characters are tolerated, as are single `\` characters not followed by `\`, `n`, or `"`. They act as a literal backslash in that case.
+Before or after each token, there may be any amount of whitespace.
```
grobinson-grafana, line 430: This was moved to a newline, just wanted to check if that was intentional?

pkg/labels/matcher.go
```diff
@@ -74,6 +74,9 @@ func NewMatcher(t MatchType, n, v string) (*Matcher, error) {
 }
 
 func (m *Matcher) String() string {
+	if !model.LabelName(m.Name).IsValid() {
```
grobinson-grafana, line 77: I'm not sure I understand what this does, would it be possible to explain it in a comment?

pkg/labels/labels.go
```diff
@@ -0,0 +1,48 @@
...
+func IsValidName(ln model.LabelName) bool {
+	lns := string(ln)
+	allSpaces := true
+	for _, i := range lns {
+		if !unicode.IsSpace(i) {
+			allSpaces = false
+			break
+		}
+	}
+	return !allSpaces && utf8.ValidString(lns)
```
grobinson-grafana, line 34: Do we need `utf8.ValidString` if its also checked on Line 138 of `parse.go`?

pkg/labels/parse.go
```diff
@@ -122,21 +126,55 @@ func ParseMatcher(s string) (_ *Matcher, err error) {
...
+	}
+
+	value, err := unescapeMatcherString(rawValue)
+	if err != nil {
+		return nil, fmt.Errorf("matcher value %w", err)
+	}
+
+	name := rawName
+	// if name is quoted, then it can contain any UTF-8 character. Unescape some escape sequences.
+	if strings.HasPrefix(rawName, `"`) {
```
grobinson-grafana, line 152: I thought we want to escape open metrics for unquoted strings, not for double quoted strings? Double quoted strings have different escape sequences right?

```suggestion
	if !strings.HasPrefix(rawName, `"`) {
```

pkg/labels/parse.go
```diff
@@ -122,21 +126,55 @@ func ParseMatcher(s string) (_ *Matcher, err error) {
...
+
+// unescapeMatcherString unescapes sequences that are escaped by openMetricsEscape: \n, \\, \", and removes leading and trailing double quotes.
+func unescapeMatcherString(raw string) (string, error) {
+	var (
+		escaped             bool
+		value               strings.Builder
+		rawValue            = raw
+		expectTrailingQuote = false
+	)
+	if strings.HasPrefix(rawValue, `"`) {
```
grobinson-grafana, line 174: Should be removed before the call to `unescapeMatcherString` I think.

pkg/labels/parse.go
```diff
@@ -122,21 +126,55 @@ func ParseMatcher(s string) (_ *Matcher, err error) {
...
+		rawValue = rawValue[1:]
+		expectTrailingQuote = true
```
grobinson-grafana, line 176: Like above, I think checking that a double quoted string is well terminated should be done somewhere else? I think I would argue that checking for terminating `"` and checking escape sequences inside a double quoted string are different operations?

pkg/labels/matcher.go
yuri-tceretian: Sure! The idea is to check whether the name is Prometheus-compatible and return it without quotes (same as it does now), and wrap it in double quotes otherwise.

yuri-tceretian:
> I am a little concerned about adding inconsistent rules to label matches. 

Currently, label values can be quoted and unquoted and also can contain UTF-8 characters in both cases. 

I agree that different rules for different parts of the expression can be confusing, and I can change that if maintainers agree. However, the goal of this PR is to introduce an extension to the current syntax with as less changes as possible and without breaking current configurations.

I did try to apply similar rules to both parts of the expression, and that's why I added support for escaped special characters (\n, \t, \") to label names.

Ideally, I think a matcher should just be a structured object rather than a string that needs to be parsed. That would drastically simplify code.

pkg/labels/labels.go
yuri-tceretian: This function is used in many other places to validate incoming labels: API calls, config parsing\validation. `parse.go` only parses matchers.

pkg/labels/parse.go
yuri-tceretian: Unquoted names can be only Prometheus-compatible, and therefore no escaping is allowed. Double quoted in contrast, can contain any UTF-8 characters, and therefore it can be escaped the same way as value.

pkg/labels/parse.go
grobinson-grafana: What I mean here is that `unescapeMatcherString` is doing two operations: unescaping escaped sequences **and** also checking if a double quoted string has both start and end quotes. What I was proposing was separate those out into separate functions / code.

yuri-tceretian:
Closing it as it does not seem that anyone is interested in reviewing it, and it will be superseded by https://github.com/prometheus/alertmanager/pull/3353.

gotjosh:
I have taken a look, but I have a strong preference for the approach taken in #3353 as is clearer in terms of explanation  - as much we'll focus our efforts on that one.

For the future though, I would have expected to see a much better documentation (in the PR) for such a critical change - as an example I have this https://github.com/prometheus/alertmanager/pull/3353#issuecomment-1549511757 to be very useful to understand to what degree this is a breaking change.

yuri-tceretian:
@gotjosh this PR did not break any current behavior but extended it and therefore did not require any supplemental documentation you referred to. 
All other documentation was updated according to the change.

> I have taken a look,

Good. For the future though, the comment would be appreciated :)

gotjosh:
> @gotjosh this PR did not break any current behavior but extended it and therefore did not require any supplemental documentation you referred to.

The expansion in character-set meant that previously rejected matchers would now be accepted - this can be considered a breaking change.

Issue #3267, opened 2023-03-01
time_intervals configuration differences in /status


cbryant42:
**What did you do?**
Updated to v0.25.0 to get support for locations in time_intervals.
**What did you expect to see?**
time_intervals represented in normal yaml format in the /#/status alertmanager endpoint. (ie. http://alertmanager:9093/#/status)
**What did you see instead? Under which circumstances?**
The first time_interval represented in yaml format, and the rest represented as dictionary objects. This is the relevant output from the alertmanager status endpoint.

```
time_intervals:
- name: time_interval_1
  time_intervals:
  - times:
    - start_time: "02:45"
      end_time: "23:45"
    weekdays: ['monday:friday']
    location: America/Chicago
- {name: time_interval_2, time_intervals: [{times: [{start_time: "05:15", end_time: "19:50"}],
      weekdays: ['monday:friday'], location: America/Chicago}]}
- {name: time_interval_3, time_intervals: [{times: [{start_time: "02:10", end_time: "18:20"}],
      weekdays: ['monday:friday'], location: America/Chicago}]}
```

**Environment**

* System information:

	Linux 4.18.0-348.7.1.el8_5.x86_64 x86_64

* Alertmanager version:

	v0.25.0

* Prometheus version:

	v2.32.1

* Alertmanager configuration file: 
  I've tried several different configuration formats for these time_intervals (changing time/weekday/location ordering, different spacing, etc.) and all give the same output to status. I'm not sure if I've misconfiguration something, this is intentional representation, or a bug. 
```
time_intervals:
- name: time_interval_1
  time_intervals:
  - weekdays: ['monday:friday']
    times:
    - start_time: '02:45'
      end_time: '23:45'
    location: 'America/Chicago'

- name: time_interval_2
  time_intervals:
  - weekdays: ['monday:friday']
    times:
    - start_time: '05:15'
      end_time: '19:50'
    location: 'America/Chicago'

- name: time_interval_3
  time_intervals:
  - weekdays: ['monday:friday']
    times:
    - start_time: '02:10'
      end_time: '18:20'
    location: 'America/Chicago'
```

* Prometheus configuration file:
```
insert configuration here (if relevant to the issue)
```
N/A
* Logs:
```
insert Prometheus and Alertmanager logs relevant to the issue here
```
N/A

simonpasquier:
This looks like an artefact of `gopkg.in/yaml.v2`?

chengjoey:
[...]

PR #3322, opened 2023-04-11
chore: upgrade gopkg.in/yaml to v3

chengjoey:
use `decoder.KnownFields` instead of `UnmarshalStrict`
add `MarshalYaml` func for `TimeInterval`

#3267


roidelapluie:
We have good reasons not to upgrade to v3, as common and prometheus are stuck with v2.

chengjoey:
> We have good reasons not to upgrade to v3, as common and prometheus are stuck with v2.

never mind, there is really no problem if don't upgrade to the v3

Issue #3331, opened 2023-04-20
amtool: /etc/amtool.yml or $home/.config/amtool/ on windows server


coatico:
The amtool documentation states that the default configuration file paths are $HOME/.config/amtool/config.yml or /etc/amtool/config.yml. When using amtool on a windows server, this poses a problem since $HOME is always relative to the current user and /etc does not exist (and can't exist).

How do I put default configuration for all users on a windows server?

PR #3334, opened 2023-04-24
add windows config path directory

coatico:
To address this issue: https://github.com/prometheus/alertmanager/issues/3331, I added a file path 'C:\Program Files\amtool' where the config.yml and credentials.yml can be stored. 
Since currently only `$HOME/.config/amtool/config.yml` or `/etc/amtool/config.yml` are used as sources for the config files, which do not exist on windows instances.

### DIFF
--- a/cli/root.go
+++ b/cli/root.go
@@ -19,6 +19,7 @@ import (
 	"os"
 	"path"
 	"time"
+	"runtime"
 
 	"github.com/alecthomas/kingpin/v2"
 	"github.com/go-openapi/strfmt"
@@ -40,8 +41,8 @@ var (
 	timeout         time.Duration
 	httpConfigFile  string
 	versionCheck    bool
+	configFiles     []string
 
-	configFiles = []string{os.ExpandEnv("$HOME/.config/amtool/config.yml"), "/etc/amtool/config.yml"}
 	legacyFlags = map[string]string{"comment_required": "require-comment"}
 )
 
@@ -127,6 +128,11 @@ func NewAlertmanagerClient(amURL *url.URL) *client.AlertmanagerAPI {
 
 // Execute is the main function for the amtool command
 func Execute() {
+	if runtime.GOOS == "windows" {
+		configFiles = []string{"C:/program files/amtool/config.yml"}
+	} else {
+		configFiles = []string{os.ExpandEnv("$HOME/.config/amtool/config.yml"), "/etc/amtool/config.yml"}
+	}
 	app := kingpin.New("amtool", helpRoot).UsageWriter(os.Stdout)
 
 	format.InitFormatFlags(app)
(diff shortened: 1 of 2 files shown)


cli/root.go
```diff
@@ -127,6 +128,11 @@ func NewAlertmanagerClient(amURL *url.URL) *client.AlertmanagerAPI {
 
 // Execute is the main function for the amtool command
 func Execute() {
+	if runtime.GOOS == "windows" {
+		configFiles = []string{"C:/program files/amtool/config.yml"}
```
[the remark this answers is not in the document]
coatico: Hey @alexweav , I found this https://pkg.go.dev/github.com/kirsle/configdir#SystemConfig and found it to work great for what is needed here I think. It's using the global or user configuration path for windows or linux.

coatico:
Hey @alexweav , I found this https://pkg.go.dev/github.com/kirsle/configdir#SystemConfig and found it to work great for what is needed here I think. It's using the global or user configuration path for windows or linux.

bboreham:
Reading the CI failures, maybe you need to push your changes to `go.mod` and `go.sum` also.

coatico:
Thanks for the feedback! I added it, still pretty new to Go.

coatico:
@bboreham the only thing I notice now is that the ci job test fails, not sure why though. I installed the latest version of go (1.20.3), referring to the help message in the ci log.

coatico:
Great! Now I know, makes sense in the end.  Updated it, not sure if I should checkout the acceptance test files, seeing the latest test run?

coatico:
@bboreham Hey, small gentle reminder here. Anything else need for the PR to be considered?

cli/root.go
```diff
@@ -41,7 +42,7 @@ var (
 	httpConfigFile  string
 	versionCheck    bool
 
-	configFiles = []string{os.ExpandEnv("$HOME/.config/amtool/config.yml"), "/etc/amtool/config.yml"}
+	configFiles = configdir.SystemConfig("amtool", "config.yml")
```
simonpasquier, line 45: This could introduce a change for existing users. I'd rather use `os.UserHomeDir()` to be platform agnostic and keep the same directory structure.

```suggestion
        var configFiles []string
	homeDir, err := os.UserHomeDir()
	if err == nil {
		configFiles = append(configFiles, filepath.Join(homeDir, ".config", "amtool" "config.yml")
	}
	configFiles = append("/etc/amtool/config.yml")
```

go.mod
```diff
@@ -70,6 +70,7 @@ require (
 	github.com/josharian/intern v1.0.0 // indirect
 	github.com/jpillora/backoff v1.0.0 // indirect
 	github.com/julienschmidt/httprouter v1.3.0 // indirect
+	github.com/kirsle/configdir v0.0.0-20170128060238-e45d2f54772f
```
siavashs, line 73: Not a fan of the extra dependency, why do we need this?

PR #3209, opened 2023-01-09
Changes receiver and inhibition rules arrays to slices of value items instead of pointer items

rohang98:
I found an issue described in #3204.

To prevent panics caused by null-pointer exceptions for empty array objects, I used @simonpasquier's suggestion by changing the arrays to slices of value items instead of pointer items.

PR #3338, opened 2023-04-25
fix: change receiver model pointer to value

fgouteroux:
Hello, 

There is a bug when calling `/api/v2/receivers`, it return the same receiver name multiple times.

**Expected:**
```
curl http://localhost:9093/api/v2/receivers
[{"name":"web.hook"},{"name":"web.hook2"}]
```

**Got:**
```
$ curl http://localhost:9093/api/v2/receivers
[{"name":"web.hook2"},{"name":"web.hook2"}]
```
I think it was related to this change https://github.com/prometheus/alertmanager/pull/3209.

### DIFF
--- a/api/v2/api_test.go
+++ b/api/v2/api_test.go
@@ -470,7 +470,7 @@ func TestMatchFilterLabels(t *testing.T) {
 	}
 }
 
-func TestGetReceivers(t *testing.T) {
+func TestGetReceiversHandler(t *testing.T) {
 	in := `
 route:
     receiver: team-X
@@ -506,9 +506,7 @@ receivers:
 		responder.WriteResponse(w, p)
 		body, _ := io.ReadAll(w.Result().Body)
 
-		bodyStr := string(body)
-
 		require.Equal(t, tc.expectedCode, w.Code)
-		require.Equal(t, tc.body, bodyStr)
+		require.Equal(t, tc.body, string(body))
 	}
 }


api/v2/api.go
```diff
@@ -224,7 +224,7 @@ func (api *API) getReceiversHandler(params receiver_ops.GetReceiversParams) midd
 
 	receivers := make([]*open_api_models.Receiver, 0, len(api.alertmanagerConfig.Receivers))
 	for _, r := range api.alertmanagerConfig.Receivers {
-		receivers = append(receivers, &open_api_models.Receiver{Name: &r.Name})
```
gotjosh, line 227: This tripped me up for a moment, but I get it now.

The problem here is that we're storing a pointer to `r` where `r` will be the last value that's used. Instead of changing the API spec, I think we should do:

```golang
receivers = append(receivers, &open_api_models.Receiver{Name: &api.alertmanagerConfig.Receivers[i].Name})
```

Which defefintely solves the problem.

Can we also please add some tests to this endpoint?

gotjosh (changes requested):
Thanks! Please see my comment.

api/v2/api.go
fgouteroux: I applied your suggested changes and add a test.

simonpasquier (approved):

api/v2/api_test.go
```diff
@@ -468,3 +469,46 @@ func TestMatchFilterLabels(t *testing.T) {
 		require.Equal(t, tc.expected, matchFilterLabels(ms, sms))
 	}
 }
+
+func TestGetReceivers(t *testing.T) {
```
gotjosh, line 473:
```suggestion
func TestGetReceiversHandler(t *testing.T) {
```

gotjosh (approved):
LGTM

I just have one small nit, can you please address it really quickly? 😄

api/v2/api_test.go
fgouteroux: yes

gotjosh:

PR #3350, opened 2023-05-02
Slack Notifier: Only return an error when it actually occurs

gotjosh:
This was introduced as part of #3252 and its a case I didn't really took into account.


alexweav (approved):
LGTM

PR #3354, opened 2023-05-05
Cluster: Add memberlist label configuration option

JohnnyQQQQ:
This PR makes it possible to set a label for the Memberlist cluster. This label is used to block any traffic that is not meant for the cluster.

For instance, in a situation where there are multiple alertmanager clusters, such as in K8s where IP addresses are shared and reused throughout the cluster, this label helps to prevent a wrong alertmanager cluster from receiving traffic meant for another cluster.

### DIFF
--- a/README.md
+++ b/README.md
@@ -372,6 +372,7 @@ be configured to communicate with each other. This is configured using the
 - `--cluster.probe-interval` value: interval between random node probes (default "1s")
 - `--cluster.reconnect-interval` value: interval between attempting to reconnect to lost peers (default "10s")
 - `--cluster.reconnect-timeout` value: length of time to attempt to reconnect to a lost peer (default: "6h0m0s")
+- `--cluster.label` value: unique label to prevent cross-communication issues when running multiple clusters (default:"")
 
 The chosen port in the `cluster.listen-address` flag is the port that needs to be
 specified in the `cluster.peer` flag of the other peers.
--- a/README.md
+++ b/README.md
@@ -372,7 +372,7 @@ be configured to communicate with each other. This is configured using the
 - `--cluster.probe-interval` value: interval between random node probes (default "1s")
 - `--cluster.reconnect-interval` value: interval between attempting to reconnect to lost peers (default "10s")
 - `--cluster.reconnect-timeout` value: length of time to attempt to reconnect to a lost peer (default: "6h0m0s")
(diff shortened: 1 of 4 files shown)


cmd/alertmanager/main.go
```diff
@@ -226,6 +226,7 @@ func run() int {
 		peerReconnectTimeout   = kingpin.Flag("cluster.reconnect-timeout", "Length of time to attempt to reconnect to a lost peer.").Default(cluster.DefaultReconnectTimeout.String()).Duration()
 		tlsConfigFile          = kingpin.Flag("cluster.tls-config", "[EXPERIMENTAL] Path to config yaml file that can enable mutual TLS within the gossip protocol.").Default("").String()
 		allowInsecureAdvertise = kingpin.Flag("cluster.allow-insecure-public-advertise-address-discovery", "[EXPERIMENTAL] Allow alertmanager to discover and listen on a public IP address.").Bool()
+		label                  = kingpin.Flag("cluster.label", "A unique identifier for the cluster used to prevent cross-communication issues when sending gossip messages.").Default("").String()
```
gotjosh, line 229:
```suggestion
		label                  = kingpin.Flag("cluster.label", "The cluster label is an optional string to include on each packet and stream. It uniquely identifies the cluster and prevents cross-communication issues when sending gossip messages.").Default("").String()
```

 	The cluster label is an optional string to include in outbound packets and gossip streams. Other members in the memberlist cluster will discard any message whose label doesn't match the configured one, unless the 'cluster-label-verification-disabled' configuration option is set to true

gotjosh (approved):
LGTM

README.md
```diff
@@ -372,6 +372,7 @@ be configured to communicate with each other. This is configured using the
 - `--cluster.probe-interval` value: interval between random node probes (default "1s")
 - `--cluster.reconnect-interval` value: interval between attempting to reconnect to lost peers (default "10s")
 - `--cluster.reconnect-timeout` value: length of time to attempt to reconnect to a lost peer (default: "6h0m0s")
+- `--cluster.label` value: the label is an optional string to include on each packet and stream. It uniquely identifies the cluster and prevents cross-communication issues when sending gossip messages(default:"")
```
JohnnyQQQQ, line 375:
```suggestion
- `--cluster.label` value: the label is an optional string to include on each packet and stream. It uniquely identifies the cluster and prevents cross-communication issues when sending gossip messages (default:"")
```

gotjosh:

PR #3362, opened 2023-05-15
Add more tests to TestMatchers

grobinson-grafana:
As part of #3353, I'd like confidence in the current test suite to catch any regressions from introducing a new parser.

This adds two cases where we expect to parsing to fail.


gotjosh (approved):
LGTM

Thanks for your contribution!

PR #3372, opened 2023-05-31
Add the route ID to uuid

qinxx108:
Adding a UUID to identify the route in general 
The route can be UUID by itself matchers+parents matchers + itself position 

```
                                                                         route1(matcher="1")
         
                    route2(matcher="1")                   route3(matcher="1")           route4(matcher="1")

route5(matcher="1")      route6(matcher="1")            

```

For example to UUID route 6, it will be (matcher="1")/(matcher="1")/(matcher="1")/1

### DIFF
--- a/dispatch/route.go
+++ b/dispatch/route.go
@@ -180,7 +180,8 @@ func (r *Route) Key() string {
 	return b.String()
 }
 
-// ID returns a key for the route. It should uniquely identify the route in general.
+// ID returns a key for the route. It should uniquely identify the route in general,
+// it is different than Key() as it adds the route's position on its parent's children.
 func (r *Route) ID() string {
 	b := strings.Builder{}
 
--- a/dispatch/route.go
+++ b/dispatch/route.go
@@ -180,8 +180,7 @@ func (r *Route) Key() string {
 	return b.String()
 }
 
-// ID returns a key for the route. It should uniquely identify the route in general,
-// it is different than Key() as it adds the route's position on its parent's children.
+// ID returns a unique identifier for the route, unlike Key().
 func (r *Route) ID() string {
 	b := strings.Builder{}
 
--- a/dispatch/route.go
+++ b/dispatch/route.go
@@ -180,25 +180,26 @@ func (r *Route) Key() string {
 	return b.String()
 }
 
-// ID returns a unique identifier for the route, unlike Key().
+// ID returns a unique identifier for the route, unlike Key(). This need to be called after full route tree
+// finish generated as it need to know the position of this route in the whole tree
 func (r *Route) ID() string {
 	b := strings.Builder{}
 
-	var position *int
+	position := -1
 	if r.parent != nil {
 		// Find the position in the same level leaf.
 		for i, cr := range r.parent.Routes {
 			if cr == r {
-				position = &i
+				position = i
 				break
 			}
 		}
 	}
 	b.WriteString(r.Key())
 
-	if position != nil {
+	if position > -1 {
 		b.WriteRune('/')
-		b.WriteString(fmt.Sprint(*position))
+		b.WriteString(fmt.Sprint(position))
 	}
 	return b.String()
 }
--- a/dispatch/route.go
+++ b/dispatch/route.go
@@ -180,8 +180,7 @@ func (r *Route) Key() string {
 	return b.String()
 }
 
-// ID returns a unique identifier for the route, unlike Key(). This need to be called after full route tree
-// finish generated as it need to know the position of this route in the whole tree
+// ID returns a unique identifier for the route.
 func (r *Route) ID() string {
 	b := strings.Builder{}
 
--- a/dispatch/route.go
+++ b/dispatch/route.go
@@ -180,6 +180,29 @@ func (r *Route) Key() string {
 	return b.String()
 }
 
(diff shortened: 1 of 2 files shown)


rohang98 (approved):

rajagopalanand (approved):

dispatch/route_test.go
```diff
@@ -853,3 +853,78 @@ routes:
 		}
 	}
 }
+
+func TestRouteID(t *testing.T) {
```
simonpasquier, line 857: if we're only testing the top-level routes, there's no need for having sub-routes?
also it might be good to have routes with same key but different ids?

dispatch/route.go
```diff
@@ -180,6 +180,30 @@ func (r *Route) Key() string {
 	return b.String()
 }
 
+// ID returns a key for the route. It should uniquely identify the route in general,
+// it is different than Key() as it adds the route's position on its parent's children.
```
simonpasquier, line 184: (nit) not sure if we need to go into the implementation details here. 
```suggestion
// ID returns a unique identifier for the route, unlike Key().

**Review by simonpasquier — COMMENTED** [review]
I'm missing the context about this change. Is it linked to the pagination API?

File: dispatch/route_test.go (hunk 1 above)
Comment (qinxx108) [reply dispatch/route_test.go:857]: "I'm missing the context about this change. Is it linked to the pagination API?"

-> It is related as we need to UUID an aggregation group, aggregation group can be uuid by group_labels + routeId

File: dispatch/route.go (hunk 2 above)
Comment (qinxx108) [reply dispatch/route.go]: fixed

File: dispatch/route_test.go (hunk 1 above)
Comment (qinxx108) [reply dispatch/route_test.go:857]: => if we're only testing the top-level routes, there's no need for having sub-routes?
also it might be good to have routes with same key but different ids?

Added the test

**Comment by gotjosh** [comment]
Overall direction looks good, but I have a couple of pieces of feedback:

1. I think we should directly make the `ID` an attribute of the Route struct. Routes structs are only created once once the Alertmanager starts or its configuration its updated and computing the IDs every single time is not needed as they don't change.

2. I think using the string representation of matches + position is a bit less human-friendly and provide poor UX. In particular, because we expect these to be used in URL that are meant to be be copy / pasteable by users. I suggest we do the following (and use a fnv hash on its hex representation)

```golang
// ID returns a unique identifier for the route, unlike Key().
func (r *Route) ID() string {
	hash := fnv.New64()

	var position *int
	if r.parent != nil {
		// Find the position in the same level leaf.
		for i, cr := range r.parent.Routes {
			if cr == r {
				position = &i
				break
			}
		}
	}

	_, _ = hash.Write([]byte(r.Key()))

	if position != nil {
		_, _ = hash.Write([]byte("/"))
		_, _ = hash.Write([]byte(fmt.Sprint(*position)))
	}

	return fmt.Sprintf("%016x", hash.Sum64())
}
```

3. I have a few comments on the readability of the test, but let's tackle the other two first, and then we can move into the test (which it's 👍 at an assertion level) later as those are mostly nits.

qinxx108:
> 



> Overall direction looks good, but I have a couple of pieces of feedback:
> 
> 1. I think we should directly make the `ID` an attribute of the Route struct. Routes structs are only created once once the Alertmanager starts or its configuration its updated and computing the IDs every single time is not needed as they don't change.
> 2. I think using the string representation of matches + position is a bit less human-friendly and provide poor UX. In particular, because we expect these to be used in URL that are meant to be be copy / pasteable by users. I suggest we do the following (and use a fnv hash on its hex representation)
> 
> ```go
> // ID returns a unique identifier for the route, unlike Key().
> func (r *Route) ID() string {
> 	hash := fnv.New64()
> 
> 	var position *int
> 	if r.parent != nil {
> 		// Find the position in the same level leaf.
> 		for i, cr := range r.parent.Routes {
> 			if cr == r {
> 				position = &i
> 				break
> 			}
> 		}
> 	}
> 
> 	_, _ = hash.Write([]byte(r.Key()))
> 
> 	if position != nil {
> 		_, _ = hash.Write([]byte("/"))
> 		_, _ = hash.Write([]byte(fmt.Sprint(*position)))
> 	}
> 
> 	return fmt.Sprintf("%016x", hash.Sum64())
> }
> ```
> 
> 3. I have a few comments on the readability of the test, but let's tackle the other two first, and then we can move into the test (which it's 👍 at an assertion level) later as those are mostly nits.

1. I think we should directly make the `ID` an attribute of the Route struct. 
=> When a route gets created, we don't know its position as the other route in the same level in the tree hasn't finished creation yet. Add some comments about the ID function

2.

gotjosh:
@qinxx108 if I remember correctly, this is the first PR that we're suppose to merge as part of the API - right? If so, could you rebase to fix the diff? 

Thanks!

qinxx108:
@gotjosh Yes this is the change
Rebased !

Thank you!

qinxx108:
@gotjosh  wonder if we can merge this pr?

gotjosh:
Yep, we can. Thanks again @qinxx108!

grobinson-grafana:
Hi! 👋 I found a bug where `route.ID()` can return conflicting IDs. I've opened an issue here https://github.com/prometheus/alertmanager/issues/3802.

PR #3381, opened 2023-06-04
Add support for Block Kit

grobinson-grafana:
### Background

In #2217 @mambax and others proposed a number of options for supporting Slack's Block Kit UI framework in Alertmanager. This feature seems to have gained some traction with 88 👍 , 9 🎉 , 12 ❤️ and 5 👀 at the time of opening this pull request.

### What this pull request does

This pull request is a draft of how I think support for Slack's Block Kit UI framework could be added to Alertmanager. You will find in this pull request the most simple implementation of Block Kit, missing validation of user configuration and a number of blocks such as the `Actions` block, and almost all Element blocks.

### How it works

The configuration file is structured as follows:

```yml
receivers:
- name: test
  slack_configs:
    - api_url: https://example.com/hooks/test
      blocks:
        - header:
            text:
              text: "{{ len .Alerts.Firing }} :fire: alerts, {{ len .Alerts.Resolved }} :white_check_mark: alerts"
              emoji: true
        - context:
            elements:
              - text:
                  text: "You have {{ len .Alerts.Firing }} firing alert(s), and {{ len .Alerts.Resolved }} alert(s) for *{{ .CommonLabels }}*"
                  markdown: true
        - section:
            text:
              text: "This is the message"
              markdown: true
route:
  receiver: test
  group_by: [alertname]
  group_wait: 30s
  group_interval: 1m
  repeat_interval: 4h
```

and here is a screenshot of a notification I sent using the above configuration:

Screenshot 2023-06-04 at 17 06 59

You'll notice that, unlike a number of the other proposals in #2217, I have put the type of each block as the name of the object rather than using `type: header`:

```yml
blocks:
  - type: header
    text:
      type: plain_text
      text: "{{ len .Alerts.Firing }} :fire: alerts, {{ len .Alerts.Resolved }} :white_check_mark: alerts"
```

The reason for this is so we can decode the Yaml into Go structs such as the following:

```go
type SlackBlock struct {
	Context *SlackBlockContext `yaml:"context,omitempty" json:"context,omitempty"`
	Divider *SlackBlockDivider `yaml:"divider,omitempty" json:"divider,omitempty"`
	Header  *SlackBlockHeader  `yaml:"header,omitempty" json:"header,omitempty"`
	Image   *SlackBlockImage   `yaml:"image,omitempty" json:"image,omitempty"`
	Section *SlackBlockSection `yaml:"section,omitempty" json:"section,omitempty"`
}
```

This is not just consistent with how all other configuration is represented in Alertmanager, but it also doesn't break strict Yaml decoding which complains when using a generic struct such as the following:

```go
type SlackBlock struct {
	Type string `yaml:"type,omitempty" json:"type,omitempty"`
	// Config contains the configuration for the block. For example, if this
	// is a header block then Config is a *SlackBlockHeader.
	Config interface{} `yaml:"-" json:"-"`
}
```

### Mockups

Here is a mockup of what notifications from Alertmanager could look like with full support of Slack's Block Kit UI framework:

Screenshot 2023-06-04 at 17 27 51

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -470,6 +470,7 @@ type SlackConfig struct {
 	LinkNames   bool           `yaml:"link_names" json:"link_names,omitempty"`
 	MrkdwnIn    []string       `yaml:"mrkdwn_in,omitempty" json:"mrkdwn_in,omitempty"`
 	Actions     []*SlackAction `yaml:"actions,omitempty" json:"actions,omitempty"`
+	Blocks      []*SlackBlock  `yaml:"blocks,omitempty" json:"blocks,omitempty"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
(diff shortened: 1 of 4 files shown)


grobinson-grafana:
A problem that we need to solve for Block Kit to be useful is how to write loops. For example, to have separate `section` and `action` blocks for each alert.

I think we can address with additional control blocks:

1. Range block
2. If block

The range and if block are special-case blocks that are not available in Block Kit, but instead are evaluated within Alertmanager to create the full Block Kit message. For example:

```yml
blocks:
  - range:
      iter: ".Alerts"
      blocks:
        - section:
            fields:
              - text:
                  text: "*Labels*: {{ .Labels }}"
                  markdown: true
        - actions:
            elements:
              - button:
                  text: Silence
                  url: "{{ .SilenceURL }}"
              - button:
                  text: View in Alertmanager
                  url: "{{ .GeneratorURL }}"
```

Here `iter` is a Go template pipeline that can be used inside a `range`. For example:

```
{{ range .Alerts }}
{{ end }}
```

The `range` and `end` statements would be inserted automatically when the configuration is loaded.

grobinson-grafana:
I'm going to close this PR because I believe BlockKit should be added to a `slack_v2` receiver, along with first class support for Bot Tokens, similar to what we did with `msteams_v2`.

PR #3383, opened 2023-06-05
Document difference between tab and newline in tests

grobinson-grafana:
I found a possible bug (unexpected behaviour?) in the current parser for label matchers where `\n` is interpolated and `\t` is not. I've chosen to document this with a test while I figure out if this is intentional or not.

Related: https://github.com/prometheus/alertmanager/pull/3353

### DIFF
--- a/pkg/labels/parse_test.go
+++ b/pkg/labels/parse_test.go
@@ -283,6 +283,22 @@ func TestMatchers(t *testing.T) {
 			}(),
 		},
 		{
+			input: `{foo=bar\t}`,
+			want: func() []*Matcher {
+				ms := []*Matcher{}
+				m, _ := NewMatcher(MatchEqual, "foo", "bar\\t")
+				return append(ms, m)
+			}(),
+		},
+		{
+			input: `{foo=bar\n}`,
+			want: func() []*Matcher {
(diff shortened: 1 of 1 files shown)


grobinson-grafana:
At least, from the PromQL documentation, I would expect `\t` and \n` to behave the same:

```
"this is a string"
'these are unescaped: \n \\ \t'
`these are not unescaped: \n ' " \t`
```

gotjosh:
Tagging @beorn7 and @simonpasquier to see if they have any reckons. In the meantime, I'll merge this as tests that document existing behaviour are always a great thing to have.

beorn7:
This is probably inspired by the behavior in the [exposition format](https://prometheus.io/docs/instrumenting/exposition_formats/#comments-help-text-and-type-information), in particular: “… the backslash (\), double-quote ("), and line feed (\n) characters have to be escaped as \\, \", and \n, respectively” (i.e. this is not using the usual Go string escaping, but a very specific limited one).

This is certainly a mixup as the matchers are supposed to be inspired by PromQL, not by the exposition format. So your expectation is spot-on, the matchers should follow escaping rules of PromQL wherever reasonably possible.

Issue #3396, opened 2023-06-17
Add num firing and resolved alerts to webhook payload


Konstantinov-Innokentii:
Hey, I'm working on integrating AlertManager with Grafana OnCall.  We are using webhook receiver for that. However,  we would like to have number of firing and resolved alerts in payload, because if _max_alerts_ is set it is impossible to calculate it correctly.

PR #3397, opened 2023-06-17
Add num Alerts to webhook notifier

Konstantinov-Innokentii:
Implement https://github.com/prometheus/alertmanager/issues/3396


Konstantinov-Innokentii:
By the way, Should I increase [Version](https://github.com/prometheus/alertmanager/pull/3397/files#diff-b0a60eed46a4546f91a8e7d13816b87b2e23758b34d37e484d86acc5d82a39b2R72)?

notify/webhook/webhook.go
```diff
@@ -96,6 +112,8 @@ func (n *Notifier) Notify(ctx context.Context, alerts ...*types.Alert) (bool, er
 		Data:            data,
 		GroupKey:        groupKey.String(),
 		TruncatedAlerts: numTruncated,
+		FiringAlerts:    numFiring,
```
simonpasquier, line 115: If we're concerned about efficiency, I'd rather add some helper functions to the `template` package.
```suggestion
		FiringAlerts:    len(alerts.Firing()),
```

simonpasquier (commented):
No need to bump the Version field as it's not a breaking change ([source](https://github.com/prometheus/alertmanager/pull/2274#issuecomment-638120849)) but it would be nice to document that in code and docs.

gotjosh:
I have some thoughts on this area.

> If we're concerned about efficiency, I'd rather add some helper functions to the template package.

I don't think we can - what the author of this PR is trying to get the numbers before truncation, and the template package turns the alerts into the type we need *after the truncation* (e.g. `data.Alerts.Firing()` or `data.Alerts.Resolved()`).

With that, I also don't love the fact that we would end up with `firingAlerts`, `truncatedAlerts` and `resolvedAlerts` but it is not clear that fire and resolved fields are pre-truncation and not what you actually see. The user of the payload would be puzzled at why this is giving different numbers that I have no way of matching.

Instead, that leaves us with three options:


no.1 w/ breaking change
```json
{
  "truncatedFired": 1,
   "truncatedResolved": 2
   "firing": 0,  // refers to alerts included in the payload
   "resolved": 1, // refers to alerts included in the payload
}
```

no.2 no breaking change to the format, but breaking change for the data
```json
{
  "truncatedAlerts": 1, // refers to firing alerts
   "truncatedAlertsResolved": 2 // refers to resolved alerts
   "firing": 0, // refers to alerts included in the payload
   "resolved": 1, // refers to alerts included in the payload
}
```

no. 3 /w breaking change

```json
{
  "truncatedTotals": { "firing": 0, "resolved": 0 },
  "totals": { "firing": 0, "resolved": 0 }
}
```

gotjosh:
@simonpasquier WDYT?

Konstantinov-Innokentii:
@gotjosh thanks for the feedback! I think that truncated totals might be an excess, since it's always possible to count resolved and firing alerts numbers after truncation using alerts array. What do you think about renaming firingAlerts to firingTotalAlerts or using only 
```
{
  "totals": { "firing": 0, "resolved": 0 }
}
```

Issue #3407, opened 2023-06-27
Reloading the config leads to incorrect notifications being sent due to a race condition


artli:
**What did you do?**

1. Run a single instance of Alertmanager: `rm -r alertmanager-data; alertmanager-0.25.0.linux-amd64/alertmanager --config.file=alertmanager.yml --web.listen-address=:9093 --cluster.listen-address=127.0.0.1:9094 --storage.path=alertmanager-data --log.level=debug`
2. Post a bunch of alerts to it: `for i in {1..20}; do alertmanager-0.25.0.linux-amd64/amtool alert add alertname=test label=$i --start $(date -u +'%Y-%m-%dT%H:%M:%SZ') --end 2024-01-01T00:00:00Z --alertmanager.url=http://localhost:9093/ --generator-url=http://localhost/ & done`
3. Wait for at least GroupWait so that the alerts are flushed
4. Issue a config reload: `curl -XPOST http://localhost:9093/-/reload`

**What did you expect to see?**

Alertmanager should respect RepeatInterval and not issue any notifications until at least RepeatInterval passes since the previous notification and then issue an identical notification.

**What did you see instead? Under which circumstances?**

Instead, a notification is sent ~immediately after the config reload. This notification contains only a small subset of the active alerts, often just one or two. After GroupInterval, a full notification is sent with all active alerts.

This is pretty bad, especially if GroupInterval is large, because it hides almost all active alerts.

**Environment**

* Alertmanager version:

```
$ alertmanager-0.25.0.linux-amd64/alertmanager --version
alertmanager, version 0.25.0 (branch: HEAD, revision: 258fab7cdd551f2cf251ed0348f0ad7289aee789)
  build user:       root@abe866dd5717
  build date:       20221222-14:51:36
  go version:       go1.19.4
  platform:         linux/amd64
```

* Alertmanager configuration file:

```
$ cat alertmanager.yml 
global:
  resolve_timeout: 500000s

route:
  group_by: ["alertname"]
  group_wait: 5s
  group_interval: 30s
  repeat_interval: 30h
  receiver: main

receivers:
  - name: main
```

Here, no notification configs are specified and so I look at the "flushing alerts" log lines as an indication of the bug, but in production we've seen this issue occur with actual receivers and actual incorrect notifications being sent (and with as few as 3 alerts in the group).

**Logs**

Startup:
```
[...]

PR #3419, opened 2023-07-05
Remove immediate flush on reload/restart

grobinson-grafana:
### What this PR does

This pull request changes Alertmanager so it no longer flushes aggregation groups on configuration reload or restart of Alertmanager as this behavior causes a number of issues:

1. Alertmanager will send notifications for inhibited alerts if the inhibited alert is sent to Alertmanager before the inhibiting alert following a restart (https://www.grobinson.net/best-practices-for-avoiding-race-conditions-in-inhibition-rules.html)

2. Reloading Alertmanager via /-/reload can cause incomplete flushes of aggregation groups (https://github.com/prometheus/alertmanager/issues/3407)

3. Reloading or restarting an Alertmanager while sending a notification can cause a race between the reloaded/restarted Alertmanager and the next peer in the Alertmanager cluster. This can, in some cases, cause firing and resolved notifications to be sent out of order. For example, resolved, firing, resolved.

A potential issue with this change is that following a reload or restart of Alertmanager, alerts that were waiting for group_wait will have to wait from the beginning of group_wait again. If group_wait is large then notifications could take longer to send then expected. Frequent reloads in combination with a large group_wait could even prevent alerts from being flushed at all.

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -381,8 +381,6 @@ type aggrGroup struct {
 	done    chan struct{}
 	next    *time.Timer
 	timeout func(time.Duration) time.Duration
-
-	mtx sync.RWMutex
 }
 
 // newAggrGroup returns a new aggregation group.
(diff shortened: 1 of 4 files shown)


grobinson-grafana:
I've just read a comment on another issue that suggests this might be causing issues for users:

> Any progress on this? We're running with a patch right now that just delays the start of the dispatcher because we were getting lots of false alarms for alerts that should be inhibited when we reloaded configs.
> https://github.com/prometheus/alertmanager/pull/3167#issuecomment-1783525248

verejoel:
We've been running in production with a patched version of AlertManager that includes this PR since a week. Is there any chance we can get some momentum behind merging this?

PR #3420, opened 2023-07-07
Update email.html to use "https://schema.org" instead of "https://schema.org"

alvinlin123:
According to schema.org's faq both https and http are fine, but https is preferred going forward: https://schema.org/docs/faq.html#19

Why am I making this change? To make some security code scanner happy :)


gotjosh:

PR #3430, opened 2023-07-27
[OpsGenie] Update message on resolve alert

jkroepke:
If `update_alerts` is enable for OpsGenie receiver, it will be only update message and description while the alerting is in firing state. This the alert goes to resolved, the message will be not updated and in OpsGenie, the FIRING prefix remains.


simonpasquier (commented):
Is the added code different from the firing case? Can we share the same between the 2 cases?

jkroepke:
@simonpasquier good call. I move both into a method. I also validate this PR again our OpsGenie instance.

image

jkroepke:
@simonpasquier No clue, why ci is faliures. One has a go panic, the other job failes on go version

notify/opsgenie/opsgenie.go
```diff
@@ -297,3 +272,50 @@ func (n *Notifier) createRequests(ctx context.Context, as ...*types.Alert) ([]*h
 
 	return requests, true, nil
 }
+
+func (n *Notifier) updateAlert(ctx context.Context, tmpl func(string) string, key notify.Key, alias string, requests *[]*http.Request) (bool, error) {
```
simonpasquier, line 276: (suggestion) rather than passing a pointer to a slice, it would be more idiomatic to return a slice and let the caller append it to the existing slice (e.g. `requests = append(requests, additionalRequests...)`)
```suggestion
func (n *Notifier) updateAlert(ctx context.Context, tmpl func(string) string, key notify.Key, alias string) ([]requests, bool, error) {
```

notify/opsgenie/opsgenie.go
```diff
@@ -297,3 +272,50 @@ func (n *Notifier) createRequests(ctx context.Context, as ...*types.Alert) ([]*h
...
+	updateMsgMsg := &opsGenieUpdateMessageMessage{
+		Message: message,
+	}
+	var updateMessageBuf bytes.Buffer
+	if err := json.NewEncoder(&updateMessageBuf).Encode(updateMsgMsg); err != nil {
+		return false, err
+	}
+	req, err := http.NewRequest("PUT", updateMessageEndpointURL.String(), &updateMessageBuf)
+	if err != nil {
+		return true, err
```
simonpasquier, line 296: I'm not quite sure why an error from NewRequest() returned should be retried. Looking at the [source code](https://cs.opensource.google/go/go/+/refs/tags/go1.20.7:src/net/http/request.go;l=838-854), we should probably not retry.

simonpasquier (commented):
Can you rebase on top of the main branch please?

notify/opsgenie/opsgenie.go
jkroepke: I did not change the existing logic here. I will change it

notify/opsgenie/opsgenie.go
jkroepke: good idea!

jkroepke:
@simonpasquier Can you take an additional review here? It looks almost complete from my point of view

jkroepke:
@simonpasquier is there any chance to get this into 0.26.0? There there any ways that I can support or assists here?

jkroepke:
Rebased

jkroepke:
@simonpasquier how we can proceed here?

jkroepke:
@simonpasquier how we can proceed here?

jkroepke:
@simonpasquier Can I help here to review this?

kbudde (approved):
Hey,

@simonpasquier  I'm also an Opsgenie user and worked on the code yesterday #3639.
Can we support you on getting some MRs reviewed and merged? 
@jkroepke already rebased his code and it looks good to me.

What do you think, can we merge it?

kbudde:
I did some tests on our opsgenie instance, this change works as expected.

Without updating message and description on resolve, the alert will always show an outdaten state
Here I compared the old and new alertmanager. The final state was Firing:0 and resolved: 2

image

jkroepke:
@simonpasquier could you please take a look here? Thanks you.

PR #3411, opened 2023-06-29
Build with Go 1.20

kellinm:
Go 1.19 will be EOL come August.

https://go.dev/doc/devel/release

Verified 1.20 tag available: https://quay.io/repository/prometheus/golang-builder?tab=tags

```
Build with Go 1.20

- Start using Go 1.20 toolchain to build
- Remove deprecated reference to updating .travis.yml
```

PR #3439, opened 2023-08-03
CI: Pin the mixtool version in CircleCI and update the golang CircleCI orb

gotjosh:
In mixtool, the tip of master broke for our mixin - I have managed to trace it down and opened a PR (see https://github.com/grafana/dashboard-linter/pull/143) but for now, let's pin the version to make sure our CI is not affected.

There's also the fact that I golang circleCI orb hasn't been updated in ages - updating it solved the CI problem we hand on tests_frontend where the goland version was failing to install. I think this might have been a result of #3411

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@
 version: 2.1
 orbs:
   prometheus: prometheus/prometheus@0.16.0
-  go: circleci/go@1.7.0
+  go: circleci/go@1.7.3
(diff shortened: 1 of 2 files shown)


gotjosh:
This one at least fixes 1 of the 2 failures - I'd say let's merge this one and tackle the other one independently.

simonpasquier (approved):
Thanks!

PR #3440, opened 2023-08-03
Update email assets after #3420

simonpasquier:
_I'm not quite sure why the CI didn't catch the missing updates._


simonpasquier:
Added dddb6e45b7e6660630ccb2eb12b441862726ebbe as part of this PR to ensure that the CI catches future issues.

PR #3442, opened 2023-08-04
Add template/email.tmpl to the clean target

simonpasquier:
The change ensures that all assets can be regenerated from scratch in the CI.


simonpasquier:
included in #3440

Issue #3446, opened 2023-08-07
Re-add group key to alert groups endpoint


Spaceman1701:
I'm working on some integrations which would benefit from associating webhook notifications with alerts returned from the results from `alerts/groups`. It'd be super useful if this endpoint included the group key for each group.

It looks like the old v1 api _did_ include the group key in `AlertGroup` (#576) but when the new v2 api was added this field was omitted: #1791 

There doesn't seem to be any discussion of the group key in #1791  so my guess is that it was just an accidental removal. Are there any concerns about re-adding this?

simonpasquier:
I participated in the discussions you've mentioned but can't remember if the choice of not including the group key was deliberate or not :thinking: 
I've got no strong objection of adding the information but I'd like to hear from @gotjosh too.

fihufil:
Hello everyone!

Reintroduction of this feature would be really welcome as it would allow us to do some more automation around Alertmanager, is it possible to reconsider this PR?

KR,
Filip Dorosz

PR #3452, opened 2023-08-08
Add group key to AlertGroup and groups API

Spaceman1701:
This change adds the group key back to the `AlertGroup` struct so that it's available in the `groups` API endpoint. I started a discussion about this with #3446.

For the most part I think this is pretty straightforward. The one bit which might need a little more work is the use of `AlertGroup` in the UI. Since it uses a type generated from the OpenAPI spec, the local construction of groups in the UI requires a dummy value for the key. I considered making the value optional in the OpenAPI spec, but that wouldn't be true to the behavior of the API.

### DIFF
--- a/dispatch/dispatch_test.go
+++ b/dispatch/dispatch_test.go
@@ -415,6 +415,7 @@ route:
 				"alertname": "OtherAlert",
 			},
 			Receiver: "prod",
+			Key:      "{}:{alertname=\"OtherAlert\"}",
 		},
 		&AlertGroup{
 			Alerts: []*types.Alert{inputAlerts[1]},
@@ -423,6 +424,7 @@ route:
 				"service":   "api",
 			},
 			Receiver: "testing",
+			Key:      "{}/{env=\"testing\"}:{alertname=\"TestingAlert\", service=\"api\"}",
 		},
 		&AlertGroup{
 			Alerts: []*types.Alert{inputAlerts[2], inputAlerts[3]},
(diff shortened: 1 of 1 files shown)


api/v2/openapi.yaml
```diff
@@ -481,10 +481,13 @@ definitions:
         type: array
         items:
           $ref: '#/definitions/gettableAlert'
+      key:
+        type: string
     required:
       - labels
       - receiver
       - alerts
+      - key
```
siavashs, line 490: This will break clients like https://github.com/prymitive/karma

siavashs (commented):
I think changing the API v2 which has been stable for a long time can cause issues for clients.
We could think about a new v3 API (graphql?) which is extendable while ensuring backwards compatibility.

api/v2/openapi.yaml
Spaceman1701: Eh yeah, it has some potential to break clients. At the time I opened this PR, API v2 was new 😓 .

While it doesn't speak for everything, this does not break Karma in practice, we have some teams who use Karma with our fork (which has this patch).

W.r.t API v3: This is something I'd be super interested in discussing. We have a some much more significant API breaks which we would upstream if we could.

siavashs: Good to know it doesn't break Karma, since we also use it.
We have many other internal clients but I'm not sure if any depend on this endpoint.

Spaceman1701: I doubt that there are clients which depend on the API not returning extra fields, but it obviously is possible. I think the question for us is whether or not a client which breaks when a field is added to a response object is well behaved or not. I'd argue that type of client is not well behaved - if an extra field causes a problem, that's a client bug.

PR #3458, opened 2023-08-14
Capture and expose notification delivery errors (#31)

grobinson-grafana:
This PR makes it possible to store the last error for each receiver in case of notification delivery failure. These errors are exposed via the `/api/v2/receivers` endpoint.


grobinson-grafana:
Not sure what happened here, please ignore this PR.

PR #3461, opened 2023-08-16
Add AlertLifeCycleObserver that allows consumers to hook into Alert life cycle

emanlodovice:
## What this pull request does

This pull requests introduces a new `AlertLifeCycleObserver` interface that is accepted in the API, Dispatcher, and the notification pipeline. This interface contains methods to allow tracking what happens to an alert in alert manager. 

## Motivation
Currently, when a customer complains “I think my alert is delayed”, we currently have no straightforward way to troubleshoot. At minimum, we should be able to quickly identify if the problem is post-notification (we sent to the receiver on time but the receiver has some delay) or pre-notification.

By introducing a new interface that allows to hook into the alert life cycle, consumers of the alert manager package would be able to implement whatever observability solution works best for them.


api/v1/api.go
```diff
@@ -447,6 +451,9 @@ func (api *API) insertAlerts(w http.ResponseWriter, r *http.Request, alerts ...*
 		if err := a.Validate(); err != nil {
 			validationErrs.Add(err)
 			api.m.Invalid().Inc()
+			if api.alertLCObserver != nil {
```
qinxx108, line 454: probably we can instance a no op observer so that you don't have to check if this is nil everytime

api/v1/api.go
```diff
@@ -447,6 +451,9 @@ func (api *API) insertAlerts(w http.ResponseWriter, r *http.Request, alerts ...*
...
+				api.alertLCObserver.Rejected("Invalid", a)
```
qinxx108, line 455: can we change the invalid to the actual error?

api/v1/api.go
```diff
@@ -456,8 +463,14 @@ func (api *API) insertAlerts(w http.ResponseWriter, r *http.Request, alerts ...*
 			typ: errorInternal,
 			err: err,
 		}, nil)
+		if api.alertLCObserver != nil {
+			api.alertLCObserver.Rejected("Failed to create", validAlerts...)
```
qinxx108, line 467: why this is rejecting?

api/v1/api_test.go
```diff
@@ -153,6 +154,20 @@ func TestAddAlerts(t *testing.T) {
 		body, _ := io.ReadAll(res.Body)
 
 		require.Equal(t, tc.code, w.Code, fmt.Sprintf("test case: %d, StartsAt %v, EndsAt %v, Response: %s", i, tc.start, tc.end, string(body)))
+
+		observer := alertobserver.NewFakeAlertLifeCycleObserver()
```
qinxx108, line 158: nit: maybe create a separate test case?

dispatch/dispatch_test.go
```diff
@@ -568,6 +569,56 @@ route:
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	// Let alerts get processed.
+	for i := 0; len(recorder.Alerts()) != 1 && i < 10; i++ {
+		time.Sleep(200 * time.Millisecond)
+	}
+	require.Equal(t, 1, len(recorder.Alerts()))
+	require.Equal(t, inputAlerts[0].Fingerprint(), observer.AggrGroupAlerts[0].Fingerprint())
+	o, ok := notify.AlertLCObserver(dispatcher.ctx)
```
qinxx108, line 617: can we create a fake observer for example increment a counter, then verify if the observer's function get called?

dispatch/dispatch.go
```diff
@@ -136,7 +140,11 @@ func (d *Dispatcher) Run() {
 	d.aggrGroupsPerRoute = map[*Route]map[model.Fingerprint]*aggrGroup{}
 	d.aggrGroupsNum = 0
 	d.metrics.aggrGroups.Set(0)
-	d.ctx, d.cancel = context.WithCancel(context.Background())
+	ctx := context.Background()
+	if d.alertLCObserver != nil {
+		ctx = notify.WithAlertLCObserver(ctx, d.alertLCObserver)
```
qinxx108, line 145: should we put the observer into the stages rather than in ctx?

api/v1/api.go
emanlodovice: I did it this way because there are cases where calling the observer method is not free. Like when an alert is added to the aggregation group we compute the `GroupKey` of the aggregation group and pass it in the method. Checking for `nil` helps us avoid this unnecessary calls when its not really providing any value. That is my only reasoning behind this, but if you think otherwise I can change it and use noop as default.

api/v1/api.go
emanlodovice: This is when `alerts.Put` failed. Since we don't end up recording the alert I considered it as rejected.

dispatch/dispatch.go
emanlodovice: You mean pass it as one of the arguments in the Exec call instead of adding it in the context?

dispatch/dispatch_test.go
emanlodovice: Yes we already do that. In line 598 we create a fake observer and in like 616 we verify that the function was called by checking the recorded alert.

api/v1/api.go
emanlodovice: updated

api/v1/api_test.go
emanlodovice: updated

grobinson-grafana:
This is great! I've been thinking about doing something similar, for the exact reasons mentioned:

> when a customer complains “I think my alert is delayed”, we currently have no straightforward way to troubleshoot. At minimum, we should be able to quickly identify if the problem is post-notification (we sent to the receiver on time but the receiver has some delay) or pre-notification.

alertobserver/alertobserver.go
```diff
@@ -0,0 +1,30 @@
...
+
+package alertobserver
+
+import (
+	"context"
+
+	"github.com/prometheus/alertmanager/types"
+)
+
+type AlertLifeCycleObserver interface {
```
grobinson-grafana, line 22: Instead of having a large interface with a method per event, have you considered having a generic Observe method that accepts metadata?

For example:

```suggestion
type LifeCycleObserver interface {
	Observe(event string, alerts []*types.Alert, meta Metadata)
}
```

The metadata could be something as simple as:

```go
type Metadata map[string]interface{}
```

grobinson-grafana (approved):

simonpasquier (commented):
I'm not 100% sure to understand how it would be used outside of prometheus/alertmanager. Can you share some code?
Also though not exactly the same, I wonder if we shouldn't implement tracing inside Alertmanager to provide this visibility about "where's my alert?".

alertobserver/alertobserver.go
emanlodovice: Sure, I can update the code as suggested. Thanks for checking 🙇

emanlodovice:
> I'm not 100% sure to understand how it would be used outside of prometheus/alertmanager. Can you share some code? Also though not exactly the same, I wonder if we shouldn't implement tracing inside Alertmanager to provide this visibility about "where's my alert?".

The use that we are thinking of is just adding logs for these events. It sort of becomes an alert history that we can query when the customer comes in. We would like to have the flexibility in implementing how we collect and format the logs and how we will store them.

alertobserver/alertobserver.go
emanlodovice: updated 🙇

dispatch/dispatch.go
```diff
@@ -338,6 +345,9 @@ func (d *Dispatcher) processAlert(alert *types.Alert, route *Route) {
 	// function, to make sure that when the run() will be executed the 1st
 	// alert is already there.
 	ag.insert(alert)
+	if d.alertLCObserver != nil {
```
qinxx108, line 348: do we need an event at d.metrics.aggrGroupLimitReached.Inc()?

notify/notify.go
```diff
@@ -797,6 +830,15 @@ func (r RetryStage) exec(ctx context.Context, l log.Logger, alerts ...*types.Ale
 			if err != nil {
 				r.metrics.numNotificationRequestsFailedTotal.WithLabelValues(r.labelValues...).Inc()
 				if !retry {
+					if r.alertLCObserver != nil {
+						m := alertobserver.AlertEventMeta{
+							"ctx":         ctx,
+							"msg":         "Unrecoverable error",
+							"integration": r.integration.Name(),
```
qinxx108, line 837: do we care about each retry? should we just record final fail or final success  at func (r RetryStage) Exec()

qinxx108:
Just some nits but overall looks good!

notify/notify.go
emanlodovice: I don't think we should care about retry here, currently we only record the final fail.

qinxx108: i mean why not put this into line 758 Exec function

emanlodovice: I updated the code the log the `sent` alerts instead because it is the correct list of alerts that was sent. I think because we don't return the `sent` alerts we have to keep the code where it currently is.

qinxx108 (approved):

emanlodovice:
@grobinson-grafana @simonpasquier could you have a look at this PR when you have time? Thank you

emanlodovice:
Rebased PR and fixed conflicts

emanlodovice:
@simonpasquier this draft PR in cortex gives the general idea of our use case for this feature https://github.com/cortexproject/cortex/pull/5602/commits

emanlodovice:
@gotjosh good day. Can you take a look at this one?

siavashs:
API v1 is deprecated, if still valid please reimplement this on top of API v2.
From my perspective investing in tracing, emitting events might be a better option.

rajagopalanand:
I can take a look at this

siavashs:
Checking this again, I think we can close it:

- it uses the deprecated API v1
- there is no garbage collection implemented, meaning this will just keep allocating memory until OOM
- this will be inconsistent cross cluster nodes, it is not guaranteed if all nodes will observe the same alert
- only the first peer will send most of the notifications, querying other peers will not show correct notification information unless `nflog` is queried (I did not analyze all the added notify logic to see if `nflog` is checked)
- following the discussion on #4732 the current approach we are trying is tracing which is being added in #4745

PR #3462, opened 2023-08-16
Add huawei welink notifier

dongjiang1989:
Support Huawei WeLink notify

see: https://www.huaweicloud.com/intl/en-us/product/welink/welink-download.html

open api: https://open.welink.huaweicloud.com/docs/#/990hh0/whokyc/mmkx2n


gotjosh:
Thank you very much for your contribution @dongjiang1989!

I don't think there's enough demand for us to support this - I've created https://github.com/prometheus/alertmanager/issues/3508 to gather interest. If enough support comes forward, I'm more than willing to have this conversation again.

PR #3465, opened 2023-08-17
Release: Fix version format

gotjosh:
Made a mistake in the version format.


VERSION
```diff
@@ -1 +1 @@
-0.26.0-rc.0
+0.26.0.rc-0
```
simonpasquier, line 1: v0.26.0-rc.0

PR #3467, opened 2023-08-17
Release: fix the version formatting

gotjosh:
It should be `<major>.<minor>.<patch>-rc.<candidate>` both the `VERSION` file and the `CHANGELOG` header must match.


gotjosh:
Discussed with @simonpasquier - I'll be merging this w/o an approval.

PR #3468, opened 2023-08-17
Fix year in 0.26.0-rc.0 entry

iamleot:
Signed-off-by: Leonardo Taccari <leot@NetBSD.org>

---

Minor PR that just fixes a thinko/pasto in the year (it should be 2023, not 2022).


gotjosh (approved):

gotjosh:
Thanks!

PR #3483, opened 2023-08-24
Support UTF-8 label matchers: Add compat package with feature flag and use in amtool

grobinson-grafana:
This pull request adds the adapter package that will allow users to switch between the new matchers/parse parser and the old pkg/labels parser. The new matchers/parse parser uses a fallback mechanism where if the input cannot be parsed in the new parser it then attempts to use the old parser. If an input is parsed in the old parser but not the new parser then a warning log is emitted.

To be reviewed after #3453, #3507 and #3523.

### DIFF
--- a/cli/alert_add.go
+++ b/cli/alert_add.go
@@ -15,6 +15,7 @@ package cli
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"time"
@@ -24,6 +25,8 @@ import (
 
 	"github.com/prometheus/alertmanager/api/v2/client/alert"
 	"github.com/prometheus/alertmanager/api/v2/models"
+	"github.com/prometheus/alertmanager/matchers/compat"
+	"github.com/prometheus/alertmanager/pkg/labels"
 )
 
 type alertAddCmd struct {
@@ -74,29 +77,45 @@ func (a *alertAddCmd) addAlert(ctx context.Context, _ *kingpin.ParseContext) err
 	if len(a.labels) > 0 {
 		// Allow the alertname label to be defined implicitly as the first argument rather
 		// than explicitly as a key=value pair.
-		if _, err := parseLabels([]string{a.labels[0]}); err != nil {
+		if _, err := compat.Matcher(a.labels[0]); err != nil {
 			a.labels[0] = fmt.Sprintf("alertname=%s", strconv.Quote(a.labels[0]))
 		}
 	}
 
-	labels, err := parseLabels(a.labels)
-	if err != nil {
-		return err
+	ls := make(models.LabelSet, len(a.labels))
+	for _, l := range a.labels {
+		matcher, err := compat.Matcher(l)
+		if err != nil {
+			return err
+		}
+		if matcher.Type != labels.MatchEqual {
+			return errors.New("labels must be specified as key=value pairs")
+		}
+		ls[matcher.Name] = matcher.Value
 	}
 
-	annotations, err := parseLabels(a.annotations)
-	if err != nil {
-		return err
+	annotations := make(models.LabelSet, len(a.annotations))
+	for _, a := range a.annotations {
+		matcher, err := compat.Matcher(a)
+		if err != nil {
+			return err
+		}
+		if matcher.Type != labels.MatchEqual {
+			return errors.New("annotations must be specified as key=value pairs")
+		}
+		annotations[matcher.Name] = matcher.Value
 	}
 
 	var startsAt, endsAt time.Time
 	if a.start != "" {
+		var err error
 		startsAt, err = time.Parse(time.RFC3339, a.start)
 		if err != nil {
 			return err
 		}
 	}
 	if a.end != "" {
+		var err error
 		endsAt, err = time.Parse(time.RFC3339, a.end)
 		if err != nil {
 			return err
@@ -106,7 +125,7 @@ func (a *alertAddCmd) addAlert(ctx context.Context, _ *kingpin.ParseContext) err
 	pa := &models.PostableAlert{
 		Alert: models.Alert{
 			GeneratorURL: strfmt.URI(a.generatorURL),
-			Labels:       labels,
+			Labels:       ls,
 		},
 		Annotations: annotations,
 		StartsAt:    strfmt.DateTime(startsAt),
@@ -117,6 +136,6 @@ func (a *alertAddCmd) addAlert(ctx context.Context, _ *kingpin.ParseContext) err
 
 	amclient := NewAlertmanagerClient(alertmanagerURL)
 
-	_, err = amclient.Alert.PostAlerts(alertParams)
+	_, err := amclient.Alert.PostAlerts(alertParams)
 	return err
 }
--- a/cli/alert_add.go
+++ b/cli/alert_add.go
@@ -16,6 +16,7 @@ package cli
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/alecthomas/kingpin/v2"
@@ -74,7 +75,7 @@ func (a *alertAddCmd) addAlert(ctx context.Context, _ *kingpin.ParseContext) err
 		// Allow the alertname label to be defined implicitly as the first argument rather
 		// than explicitly as a key=value pair.
 		if _, err := parseLabels([]string{a.labels[0]}); err != nil {
-			a.labels[0] = fmt.Sprintf("alertname=%s", a.labels[0])
+			a.labels[0] = fmt.Sprintf("alertname=%s", strconv.Quote(a.labels[0]))
 		}
 	}
 
--- a/cli/alert_query.go
+++ b/cli/alert_query.go
@@ -22,7 +22,7 @@ import (
 
 	"github.com/prometheus/alertmanager/api/v2/client/alert"
 	"github.com/prometheus/alertmanager/cli/format"
-	"github.com/prometheus/alertmanager/matchers/adapter"
+	"github.com/prometheus/alertmanager/matchers/compat"
 )
 
 type alertQueryCmd struct {
@@ -80,7 +80,7 @@ func (a *alertQueryCmd) queryAlerts(ctx context.Context, _ *kingpin.ParseContext
 		// the user wants alertname=<arg> and prepend `alertname=` to
 		// the front.
 		m := a.matcherGroups[0]
-		_, err := adapter.ParseMatcher(m)
+		_, err := compat.ParseMatcher(m)
 		if err != nil {
 			a.matcherGroups[0] = fmt.Sprintf("alertname=%s", m)
 		}
--- a/cli/alert_query.go
+++ b/cli/alert_query.go
@@ -80,7 +80,7 @@ func (a *alertQueryCmd) queryAlerts(ctx context.Context, _ *kingpin.ParseContext
 		// the user wants alertname=<arg> and prepend `alertname=` to
 		// the front.
 		m := a.matcherGroups[0]
-		_, err := compat.ParseMatcher(m)
+		_, err := compat.Matcher(m)
 		if err != nil {
 			a.matcherGroups[0] = fmt.Sprintf("alertname=%s", m)
 		}
--- a/cli/alert_query.go
+++ b/cli/alert_query.go
@@ -22,7 +22,7 @@ import (
 
 	"github.com/prometheus/alertmanager/api/v2/client/alert"
 	"github.com/prometheus/alertmanager/cli/format"
-	"github.com/prometheus/alertmanager/pkg/labels"
+	"github.com/prometheus/alertmanager/matchers/adapter"
 )
 
 type alertQueryCmd struct {
@@ -80,7 +80,7 @@ func (a *alertQueryCmd) queryAlerts(ctx context.Context, _ *kingpin.ParseContext
 		// the user wants alertname=<arg> and prepend `alertname=` to
 		// the front.
 		m := a.matcherGroups[0]
-		_, err := labels.ParseMatcher(m)
+		_, err := adapter.ParseMatcher(m)
 		if err != nil {
 			a.matcherGroups[0] = fmt.Sprintf("alertname=%s", m)
 		}
--- a/cli/root.go
+++ b/cli/root.go
@@ -18,9 +18,11 @@ import (
 	"net/url"
 	"os"
 	"path"
+	"strings"
 	"time"
 
 	"github.com/alecthomas/kingpin/v2"
+	"github.com/go-kit/log"
 	"github.com/go-openapi/strfmt"
 	promconfig "github.com/prometheus/common/config"
 	"github.com/prometheus/common/version"
@@ -29,6 +31,8 @@ import (
 	"github.com/prometheus/alertmanager/api/v2/client"
 	"github.com/prometheus/alertmanager/cli/config"
 	"github.com/prometheus/alertmanager/cli/format"
+	"github.com/prometheus/alertmanager/featurecontrol"
+	"github.com/prometheus/alertmanager/matchers/adapter"
 
 	clientruntime "github.com/go-openapi/runtime/client"
 )
@@ -40,6 +44,7 @@ var (
 	timeout         time.Duration
 	httpConfigFile  string
 	versionCheck    bool
+	featureFlags    string
 
 	configFiles = []string{os.ExpandEnv("$HOME/.config/amtool/config.yml"), "/etc/amtool/config.yml"}
 	legacyFlags = map[string]string{"comment_required": "require-comment"}
@@ -137,6 +142,17 @@ func Execute() {
 	app.Flag("timeout", "Timeout for the executed command").Default("30s").DurationVar(&timeout)
 	app.Flag("http.config.file", "HTTP client configuration file for amtool to connect to Alertmanager.").PlaceHolder("<filename>").ExistingFileVar(&httpConfigFile)
 	app.Flag("version-check", "Check alertmanager version. Use --no-version-check to disable.").Default("true").BoolVar(&versionCheck)
+	app.Flag("enable-feature", fmt.Sprintf("Experimental features to enable. The flag can be repeated to enable multiple features. Valid options: %s", strings.Join(featurecontrol.AllowedFlags, ", "))).Default("").StringVar(&featureFlags)
+
+	logger := log.NewLogfmtLogger(os.Stdout)
+	featureConfig, err := featurecontrol.NewFlags(logger, featureFlags)
+	if err != nil {
+		kingpin.Fatalf(":error parsing the feature flag list: %v\n", err)
+	}
+	if featureConfig.DisableNewLabelMatchers() {
+		adapter.ParseMatcher = adapter.OldMatcherParser(logger)
+		adapter.ParseMatchers = adapter.OldMatchersParser(logger)
+	}
 
 	app.Version(version.Print("amtool"))
 	app.GetFlag("help").Short('h')
--- a/cli/root.go
+++ b/cli/root.go
@@ -148,10 +148,7 @@ func Execute() {
 	if err != nil {
 		kingpin.Fatalf(":error parsing the feature flag list: %v\n", err)
 	}
-	if featureConfig.DisableNewLabelMatchers() {
-		compat.ParseMatcher = compat.OldMatcherParser(logger)
-		compat.ParseMatchers = compat.OldMatchersParser(logger)
-	}
+	compat.InitFromFlags(logger, featureConfig)
 
 	app.Version(version.Print("amtool"))
 	app.GetFlag("help").Short('h')
--- a/cli/root.go
+++ b/cli/root.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/alecthomas/kingpin/v2"
 	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
 	clientruntime "github.com/go-openapi/runtime/client"
 	"github.com/go-openapi/strfmt"
 	promconfig "github.com/prometheus/common/config"
@@ -48,6 +49,21 @@ var (
 	configFiles = []string{os.ExpandEnv("$HOME/.config/amtool/config.yml"), "/etc/amtool/config.yml"}
 	legacyFlags = map[string]string{"comment_required": "require-comment"}
 )
+
+func initMatchersCompat(_ *kingpin.ParseContext) error {
+	logger := log.NewLogfmtLogger(os.Stdout)
+	if verbose {
+		logger = level.NewFilter(logger, level.AllowDebug())
+	} else {
+		logger = level.NewFilter(logger, level.AllowInfo())
+	}
+	featureConfig, err := featurecontrol.NewFlags(logger, featureFlags)
+	if err != nil {
+		kingpin.Fatalf("error parsing the feature flag list: %v\n", err)
+	}
+	compat.InitFromFlags(logger, featureConfig)
+	return nil
+}
 
 func requireAlertManagerURL(pc *kingpin.ParseContext) error {
 	// Return without error if any help flag is set.
@@ -143,13 +159,6 @@ func Execute() {
 	app.Flag("version-check", "Check alertmanager version. Use --no-version-check to disable.").Default("true").BoolVar(&versionCheck)
 	app.Flag("enable-feature", fmt.Sprintf("Experimental features to enable. The flag can be repeated to enable multiple features. Valid options: %s", strings.Join(featurecontrol.AllowedFlags, ", "))).Default("").StringVar(&featureFlags)
 
-	logger := log.NewLogfmtLogger(os.Stdout)
-	featureConfig, err := featurecontrol.NewFlags(logger, featureFlags)
-	if err != nil {
-		kingpin.Fatalf("error parsing the feature flag list: %v\n", err)
-	}
-	compat.InitFromFlags(logger, featureConfig)
-
 	app.Version(version.Print("amtool"))
 	app.GetFlag("help").Short('h')
 	app.UsageTemplate(kingpin.CompactUsageTemplate)
@@ -165,6 +174,8 @@ func Execute() {
 	configureClusterCmd(app)
 	configureConfigCmd(app)
 	configureTemplateCmd(app)
+
+	app.Action(initMatchersCompat)
 
 	err = resolver.Bind(app, os.Args[1:])
 	if err != nil {
--- a/cli/root.go
+++ b/cli/root.go
@@ -146,7 +146,7 @@ func Execute() {
 	logger := log.NewLogfmtLogger(os.Stdout)
 	featureConfig, err := featurecontrol.NewFlags(logger, featureFlags)
 	if err != nil {
-		kingpin.Fatalf(":error parsing the feature flag list: %v\n", err)
+		kingpin.Fatalf("error parsing the feature flag list: %v\n", err)
 	}
 	compat.InitFromFlags(logger, featureConfig)
 
--- a/cli/root.go
+++ b/cli/root.go
@@ -23,12 +23,12 @@ import (
 
 	"github.com/alecthomas/kingpin/v2"
 	"github.com/go-kit/log"
+	clientruntime "github.com/go-openapi/runtime/client"
 	"github.com/go-openapi/strfmt"
 	promconfig "github.com/prometheus/common/config"
 	"github.com/prometheus/common/version"
 	"golang.org/x/mod/semver"
 
-	clientruntime "github.com/go-openapi/runtime/client"
 	"github.com/prometheus/alertmanager/api/v2/client"
 	"github.com/prometheus/alertmanager/cli/config"
 	"github.com/prometheus/alertmanager/cli/format"
--- a/cli/root.go
+++ b/cli/root.go
@@ -28,13 +28,12 @@ import (
 	"github.com/prometheus/common/version"
 	"golang.org/x/mod/semver"
 
+	clientruntime "github.com/go-openapi/runtime/client"
 	"github.com/prometheus/alertmanager/api/v2/client"
 	"github.com/prometheus/alertmanager/cli/config"
 	"github.com/prometheus/alertmanager/cli/format"
 	"github.com/prometheus/alertmanager/featurecontrol"
-	"github.com/prometheus/alertmanager/matchers/adapter"
-
-	clientruntime "github.com/go-openapi/runtime/client"
+	"github.com/prometheus/alertmanager/matchers/compat"
 )
 
 var (
@@ -150,8 +149,8 @@ func Execute() {
 		kingpin.Fatalf(":error parsing the feature flag list: %v\n", err)
 	}
 	if featureConfig.DisableNewLabelMatchers() {
-		adapter.ParseMatcher = adapter.OldMatcherParser(logger)
-		adapter.ParseMatchers = adapter.OldMatchersParser(logger)
+		compat.ParseMatcher = compat.OldMatcherParser(logger)
+		compat.ParseMatchers = compat.OldMatchersParser(logger)
 	}
 
 	app.Version(version.Print("amtool"))
--- a/cli/utils.go
+++ b/cli/utils.go
@@ -26,22 +26,8 @@ import (
 	"github.com/prometheus/alertmanager/api/v2/client/general"
 	"github.com/prometheus/alertmanager/api/v2/models"
 	"github.com/prometheus/alertmanager/config"
-	"github.com/prometheus/alertmanager/matchers/compat"
 	"github.com/prometheus/alertmanager/pkg/labels"
 )
-
-// parseMatchers parses a list of matchers (cli arguments).
-func parseMatchers(inputMatchers []string) ([]labels.Matcher, error) {
-	matchers := make([]labels.Matcher, 0, len(inputMatchers))
-	for _, s := range inputMatchers {
-		matcher, err := compat.Matcher(s)
-		if err != nil {
-			return []labels.Matcher{}, err
-		}
-		matchers = append(matchers, *matcher)
-	}
-	return matchers, nil
-}
 
 // getRemoteAlertmanagerConfigStatus returns status responsecontaining configuration from remote Alertmanager
 func getRemoteAlertmanagerConfigStatus(ctx context.Context, alertmanagerURL *url.URL) (*models.AlertmanagerStatus, error) {
@@ -90,25 +76,6 @@ func convertClientToCommonLabelSet(cls models.LabelSet) model.LabelSet {
 		mls[model.LabelName(ln)] = model.LabelValue(lv)
 	}
 	return mls
-}
-
-// parseLabels parses a list of labels (cli arguments).
-func parseLabels(inputLabels []string) (models.LabelSet, error) {
-	labelSet := make(models.LabelSet, len(inputLabels))
-
-	for _, l := range inputLabels {
-		matcher, err := compat.Matcher(l)
-		if err != nil {
-			return models.LabelSet{}, err
-		}
-		if matcher.Type != labels.MatchEqual {
-			return models.LabelSet{}, errors.New("labels must be specified as key=value pairs")
-		}
-
-		labelSet[matcher.Name] = matcher.Value
-	}
-
-	return labelSet, nil
 }
 
 // TypeMatchers only valid for when you are going to add a silence
--- a/cli/utils.go
+++ b/cli/utils.go
@@ -26,7 +26,7 @@ import (
 	"github.com/prometheus/alertmanager/api/v2/client/general"
 	"github.com/prometheus/alertmanager/api/v2/models"
 	"github.com/prometheus/alertmanager/config"
-	"github.com/prometheus/alertmanager/matchers/adapter"
+	"github.com/prometheus/alertmanager/matchers/compat"
 	"github.com/prometheus/alertmanager/pkg/labels"
 )
 
@@ -35,7 +35,7 @@ func parseMatchers(inputMatchers []string) ([]labels.Matcher, error) {
 	matchers := make([]labels.Matcher, 0, len(inputMatchers))
 
 	for _, v := range inputMatchers {
-		matcher, err := adapter.ParseMatcher(v)
+		matcher, err := compat.ParseMatcher(v)
 		if err != nil {
 			return []labels.Matcher{}, err
 		}
@@ -100,7 +100,7 @@ func parseLabels(inputLabels []string) (models.LabelSet, error) {
 	labelSet := make(models.LabelSet, len(inputLabels))
 
 	for _, l := range inputLabels {
-		matcher, err := adapter.ParseMatcher(l)
+		matcher, err := compat.ParseMatcher(l)
 		if err != nil {
 			return models.LabelSet{}, err
 		}
--- a/cli/utils.go
+++ b/cli/utils.go
@@ -35,7 +35,7 @@ func parseMatchers(inputMatchers []string) ([]labels.Matcher, error) {
 	matchers := make([]labels.Matcher, 0, len(inputMatchers))
 
 	for _, v := range inputMatchers {
-		matcher, err := compat.ParseMatcher(v)
+		matcher, err := compat.Matcher(v)
 		if err != nil {
 			return []labels.Matcher{}, err
 		}
@@ -100,7 +100,7 @@ func parseLabels(inputLabels []string) (models.LabelSet, error) {
 	labelSet := make(models.LabelSet, len(inputLabels))
 
 	for _, l := range inputLabels {
-		matcher, err := compat.ParseMatcher(l)
+		matcher, err := compat.Matcher(l)
 		if err != nil {
 			return models.LabelSet{}, err
 		}
--- a/cli/utils.go
+++ b/cli/utils.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"net/url"
 	"os"
-	"strings"
 
 	"github.com/alecthomas/kingpin/v2"
 	"github.com/prometheus/common/model"
@@ -35,9 +34,6 @@ import (
 func parseMatchers(inputMatchers []string) ([]labels.Matcher, error) {
 	matchers := make([]labels.Matcher, 0, len(inputMatchers))
 	for _, s := range inputMatchers {
-		if strings.HasPrefix(s, "{") || strings.HasSuffix(s, "}") {
-			return []labels.Matcher{}, fmt.Errorf("unexpected open or close brace: %s", s)
-		}
 		matcher, err := compat.Matcher(s)
 		if err != nil {
 			return []labels.Matcher{}, err
@@ -101,9 +97,6 @@ func parseLabels(inputLabels []string) (models.LabelSet, error) {
 	labelSet := make(models.LabelSet, len(inputLabels))
 
 	for _, l := range inputLabels {
-		if strings.HasPrefix(l, "{") || strings.HasSuffix(l, "}") {
-			return models.LabelSet{}, fmt.Errorf("unexpected open or close brace: %s", l)
-		}
 		matcher, err := compat.Matcher(l)
 		if err != nil {
 			return models.LabelSet{}, err
--- a/cli/utils.go
+++ b/cli/utils.go
@@ -101,6 +101,9 @@ func parseLabels(inputLabels []string) (models.LabelSet, error) {
 	labelSet := make(models.LabelSet, len(inputLabels))
 
 	for _, l := range inputLabels {
+		if strings.HasPrefix(l, "{") || strings.HasSuffix(l, "}") {
+			return models.LabelSet{}, fmt.Errorf("unexpected open or close brace: %s", l)
+		}
 		matcher, err := compat.Matcher(l)
 		if err != nil {
 			return models.LabelSet{}, err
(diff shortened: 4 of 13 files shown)


grobinson-grafana:
I was thinking of adding the actual feature flag in a follow up PR to keep this one small and easier to review @gotjosh.

matchers/adapter/parse.go
```diff
@@ -0,0 +1,142 @@
...
+			// it cannot be valid input.
+			m, invalidErr = labels.ParseMatchers(s)
+			if invalidErr != nil {
+				return nil, invalidErr
+			}
+			// The input is valid in the old pkg/labels parser, but not the
+			// new matchers/parse parser.
+			level.Warn(l).Log(
+				"msg",
+				"Failed to parse input with matchers/parse, falling back to pkg/labels parser",
```
grobinson-grafana, line 133: @gotjosh I'm thinking if I should make this message more actionable for the end user. Perhaps something like:
```suggestion
				"Alertmanager is moving to a new parser for label matchers, and it looks like this matcher will be incompatible. To fix it make sure the regular expression or value is double-quoted.",
```

It's not true that it **has** to be double-quoted, but double-quoting it should guarantee to fix the issue, so I think it's the best recommendation to make.

grobinson-grafana, line 133: @gotjosh I'm thinking if I should make this message more actionable for the end user. Perhaps something like:
```suggestion
				"Alertmanager is moving to a new parser for label matchers, and it looks like this matcher will be incompatible. To fix it make sure the regular expression or value is double-quoted.",
```

It's not true that the regular expression or value **has** to be double-quoted, but double-quoting should fix the issue in all cases, so I think it's the right suggestion to put in the error message.

cli/root.go
```diff
@@ -137,6 +142,17 @@ func Execute() {
 	app.Flag("timeout", "Timeout for the executed command").Default("30s").DurationVar(&timeout)
 	app.Flag("http.config.file", "HTTP client configuration file for amtool to connect to Alertmanager.").PlaceHolder("<filename>").ExistingFileVar(&httpConfigFile)
 	app.Flag("version-check", "Check alertmanager version. Use --no-version-check to disable.").Default("true").BoolVar(&versionCheck)
+	app.Flag("enable-feature", fmt.Sprintf("Experimental features to enable. The flag can be repeated to enable multiple features. Valid options: %s", strings.Join(featurecontrol.AllowedFlags, ", "))).Default("").StringVar(&featureFlags)
+
+	logger := log.NewLogfmtLogger(os.Stdout)
```
grobinson-grafana, line 147: Not sure I should be creating a logger here, but then where else should I do it? :thinking:

featurecontrol/featurecontrol.go
```diff
@@ -22,24 +22,31 @@ import (
...
-	fcReceiverNameInMetrics = "receiver-name-in-metrics"
+	fcReceiverNameInMetrics    = "receiver-name-in-metrics"
+	fcDisabledNewLabelMatchers = "disable-new-label-matchers"
 )
 
 var AllowedFlags = []string{fcReceiverNameInMetrics}
 
 type Flagger interface {
 	EnableReceiverNamesInMetrics() bool
+	DisableNewLabelMatchers() bool
```
grobinson-grafana, line 33: Feel free to suggest a better name here

featurecontrol/featurecontrol.go
```diff
@@ -22,24 +22,31 @@ import (
 )
 
 const (
-	fcReceiverNameInMetrics = "receiver-name-in-metrics"
+	fcReceiverNameInMetrics    = "receiver-name-in-metrics"
+	fcDisabledNewLabelMatchers = "disable-new-label-matchers"
```
gotjosh, line 26:
```suggestion
	fcUTF8MatcherParsing = "utf8-matchers-parsing"
	fcStrictMatcherParsing = "strict-matchers-parsing"
```

gotjosh: One is more restrictive than the other, so that's the name I came up with.

The feature flags have wording around them in terms of enablement - as such, I think enabling and disabling sounds counterintuitive.

cli/root.go
```diff
@@ -137,6 +142,17 @@ func Execute() {
...
+	featureConfig, err := featurecontrol.NewFlags(logger, featureFlags)
+	if err != nil {
+		kingpin.Fatalf(":error parsing the feature flag list: %v\n", err)
+	}
+	if featureConfig.DisableNewLabelMatchers() {
+		adapter.ParseMatcher = adapter.OldMatcherParser(logger)
+		adapter.ParseMatchers = adapter.OldMatchersParser(logger)
```
gotjosh, line 154: I think I would prefer if we made the compact package stateful that way we wouldn't need to do global variables.

```golang
compatibleParser := compat.NewParser(featureConfig) 
```

Then callers can do: 

```golang
compatibleParser.ParseMatcher()
```

matchers/adapter/parse.go
gotjosh: > It's not true that the regular expression or value has to be double-quoted

What would be the exceptions?

If that's the case, then I'd suggest you include the wording "Have you tried to double quote the regular expression or value?" to indicate that this can be a solution but not the only solution and perhaps prompt the user to open an issue.

gotjosh (changes requested):
Looks good! I have a few suggestions, let me know what you think.

No matter what I tried, I couldn't get the message about the `Disabled new label matchers` to show up. Can you give me an example command that would make the show up?

I tried:

```golang
./amtool --enable-feature="disable-new-label-matchers" --alertmanager.url=http://localhost:9093/ alert query
```

featurecontrol/featurecontrol.go
grobinson-grafana: What's the difference between `utf8-matchers-parsing` and `strict-matchers-parsing`?

grobinson-grafana:
> Looks good! I have a few suggestions, let me know what you think.
> 
> No matter what I tried, I couldn't get the message about the `Disabled new label matchers` to show up. Can you give me an example command that would make the show up?
> 
> I tried:
> 
> ```go
> ./amtool --enable-feature="disable-new-label-matchers" --alertmanager.url=http://localhost:9093/ alert query
> ```

Ah I was using `log.NewNopLogger()`! 😄

featurecontrol/featurecontrol.go
gotjosh: `utf8-matchers-parsing` is new parser and `strict-matchers-parsing` is the old parser - not sure if this is what you're asking though so feel free to clarify.

grobinson-grafana: Hmm. But the old parser is less strict than the new the parser? In either case, I was thinking of just having one feature flag rather than two, as with two we need to add code to detect when both are enabled.

cli/root.go
```diff
@@ -137,6 +142,17 @@ func Execute() {
 	app.Flag("timeout", "Timeout for the executed command").Default("30s").DurationVar(&timeout)
 	app.Flag("http.config.file", "HTTP client configuration file for amtool to connect to Alertmanager.").PlaceHolder("<filename>").ExistingFileVar(&httpConfigFile)
 	app.Flag("version-check", "Check alertmanager version. Use --no-version-check to disable.").Default("true").BoolVar(&versionCheck)
+	app.Flag("enable-feature", fmt.Sprintf("Experimental features to enable. The flag can be repeated to enable multiple features. Valid options: %s", strings.Join(featurecontrol.AllowedFlags, ", "))).Default("").StringVar(&featureFlags)
```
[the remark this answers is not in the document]
grobinson-grafana: I'm using `StringVar` here just to be consistent with all the other flags in `root.go` as all flags are declared [here](https://github.com/prometheus/alertmanager/blob/6ea06cd11ce37ac6df8b494ae75db14571e6c070/cli/root.go#L40-L51).

featurecontrol/featurecontrol.go
gotjosh: Perhaps we're talking about different restrictions - the new parser is more strict in terms of how it accepts inputs but less restrictive in terms of what kind of matchers it accepts (we accept any UTF-8 valid character as label name after all).

gotjosh: >In either case, I was thinking of just having one feature flag that disables the new parser as that's what we discussed together with Simon in the AM Working Group. If we have two feature flags I don't understand what we're supposed to do when either both feature flags are enabled or no feature flags are enabled?

For completeness we basically have 3 modes:

- just old parser
- just new parser
- both new and old parser

In my eyes, 2 new flags + a default doesn't seem too out of this world.

grobinson-grafana: OK! I think in such case we should find another name instead of `strict-matchers-parsing` because we are both extending and restricting the grammar with the new parser. Perhaps `stable-matchers-parsing`?

Yes, but in the Working Group I thought we agreed on just two modes. How come we're changing it now?

grobinson-grafana: OK! I think in such case we should find another name instead of strict-matchers-parsing because we are both extending and restricting the grammar with the new parser. Perhaps stable-matchers-parsing?

>F or completeness we basically have 3 modes:

Yes, but in the Working Group I thought we agreed on just two modes. How come we're changing it?

gotjosh: @simonpasquier do you have any concerns with the above approach?

My thinking of having 3 modes (default + 2 flags) is that:

1.  Early adopters and innovators can run with the flag of "only new parser" enabled (this future proofing them for when/if we make the new parser permanent)
2. The flag of "only old parser" it's there in case we have any sort of bugs on the new parser where we only want to enable old parsing to make sure users are not "stuck".
3. The default mode of having both parsers working together to ensure most people are able to upgrade as seamlessly as possible w/o any sort of breaking change.

The approach I'd take is that as part of the changelog, we recommend people to run with the flag of `--new-parser-enabled` if they have the time to fix things between upgrades as that would guarantee they are future proof.

cli/root.go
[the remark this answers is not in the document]
gotjosh: Gotcha, I missed that! I'm OK with this.

cli/root.go
gotjosh: I was thinking here in `root.go` but perhaps there is a problem with that?

grobinson-grafana: Yes `root.go` is what I meant 😅 oops!

But if `compatibleParser` is a local variable in `root.go`, how do we propagate it to [`utils.go`](https://github.com/prometheus/alertmanager/blob/main/cli/utils.go#L41) for example?

featurecontrol/featurecontrol.go
grobinson-grafana: I pushed a commit for the two-option flag (+ default) so we can see how it looks.

featurecontrol/featurecontrol.go
```diff
@@ -70,9 +107,17 @@ func NewFlags(logger log.Logger, features string) (Flagger, error) {
 		opt(fc)
 	}
 
+	if fc.stableMatchersParsing && fc.utf8MatchersParsing {
```
grobinson-grafana, line 110: @gotjosh this seemed like the most sensible thing to do here. I wasn't keen on having an implicit link between both `"stable-matchers-parsing"` and `"utf8-matchers-parsing"` being enabled and using the fallback parser.

grobinson-grafana, line 110: @gotjosh this seemed like the most sensible thing to do here. I wasn't keen on having the fallback parser be equivalent to both `"stable-matchers-parsing"` and `"utf8-matchers-parsing"` being enabled. It seems non-obvious and surprising, and I think the better option here is to fail and ask them to remove the flag for fallback instead.

grobinson-grafana:
@gotjosh I forgot to mention that #3507 must be merged before this PR, as the test I'm introducing in #3507 is supposed to fail on this branch! I will make this clear in both PRs.

grobinson-grafana:
Just rebased `main` to include #3507 and `make test` fails as expected 🥳 :

```
--- FAIL: TestQuerySilence (0.14s)
    cli_test.go:187:
        	Error Trace:	/Users/grobinson/go/src/github.com/prometheus/alertmanager/test/cli/acceptance/cli_test.go:187
        	Error:      	Not equal:
        	            	expected: []string{"alertname=\"{alertname=test3}\"", "severity=\"warn\""}
        	            	actual  : []string{"alertname=\"test3\"", "severity=\"warn\""}

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,3 @@
        	            	 ([]string) (len=2) {
        	            	- (string) (len=29) "alertname=\"{alertname=test3}\"",
        	            	+ (string) (len=17) "alertname=\"test3\"",
        	            	  (string) (len=15) "severity=\"warn\""
        	Test:       	TestQuerySilence
FAIL
```

grobinson-grafana:
Just rebased `main` to include #3507 and `make test` fails as expected 🥳 :

```
--- FAIL: TestQuerySilence (0.14s)
    cli_test.go:187:
        	Error Trace:	/Users/grobinson/go/src/github.com/prometheus/alertmanager/test/cli/acceptance/cli_test.go:187
        	Error:      	Not equal:
        	            	expected: []string{"alertname=\"{alertname=test3}\"", "severity=\"warn\""}
        	            	actual  : []string{"alertname=\"test3\"", "severity=\"warn\""}

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1,3 +1,3 @@
        	            	 ([]string) (len=2) {
        	            	- (string) (len=29) "alertname=\"{alertname=test3}\"",
        	            	+ (string) (len=17) "alertname=\"test3\"",
        	            	  (string) (len=15) "severity=\"warn\""
        	Test:       	TestQuerySilence
FAIL
```

I'll now work on making this test pass, which means adding a check for `{` and `}` to the relevant commands.

matchers/adapter/parse.go
grobinson-grafana: I think exceptions would be `foo=""` where `"` is interpolated in bash/zsh as `foo=`. It would need to be escaped as `foo=\"\"` or `'foo=""'`.

cli/alert_add.go
```diff
@@ -74,7 +75,7 @@ func (a *alertAddCmd) addAlert(ctx context.Context, _ *kingpin.ParseContext) err
 		// Allow the alertname label to be defined implicitly as the first argument rather
 		// than explicitly as a key=value pair.
 		if _, err := parseLabels([]string{a.labels[0]}); err != nil {
-			a.labels[0] = fmt.Sprintf("alertname=%s", a.labels[0])
+			a.labels[0] = fmt.Sprintf("alertname=%s", strconv.Quote(a.labels[0]))
```
grobinson-grafana, line 78: This is quoted to make sure commands such as `amtool alert add "foo with spaces"` is parsed as `alertname="foo with spaces"` and not `alertname=foo with spaces` - the latter being invalid in the UTF-8 parser.

cli/alert_query.go
```diff
@@ -80,7 +80,7 @@ func (a *alertQueryCmd) queryAlerts(ctx context.Context, _ *kingpin.ParseContext
 		// the user wants alertname=<arg> and prepend `alertname=` to
 		// the front.
 		m := a.matcherGroups[0]
-		_, err := labels.ParseMatcher(m)
+		_, err := compat.Matcher(m)
```
grobinson-grafana, line 83: Just replaced all occurences of `labels.ParseMatcher` with `compat.Matcher`. The switching of the feature happens in `cli/root.go`.

cli/root.go
```diff
@@ -40,11 +44,27 @@ var (
 	timeout         time.Duration
 	httpConfigFile  string
 	versionCheck    bool
+	featureFlags    string
 
 	configFiles = []string{os.ExpandEnv("$HOME/.config/amtool/config.yml"), "/etc/amtool/config.yml"}
 	legacyFlags = map[string]string{"comment_required": "require-comment"}
 )
 
+func initMatchersCompat(_ *kingpin.ParseContext) error {
```
grobinson-grafana, line 53: This is where the switching of the feature happens. I needed to use `app.Action` so it would run after Kingpin had computed the flags but before running any commands.

featurecontrol/featurecontrol.go
```diff
@@ -23,23 +24,41 @@ import (
 
 const (
 	fcReceiverNameInMetrics = "receiver-name-in-metrics"
+	fcStableMatchersParsing = "stable-matchers-parsing"
```
grobinson-grafana, line 27: `stable-matchers-parsing` refers to the old `pkg/labels` parser.

featurecontrol/featurecontrol.go
```diff
@@ -23,23 +24,41 @@ import (
...
+	fcUTF8MatchersParsing   = "utf8-matchers-parsing"
```
grobinson-grafana, line 28: And `utf-8-matchers-parsing` refers to the new UTF-8 parser.

featurecontrol/featurecontrol.go
```diff
@@ -70,9 +107,17 @@ func NewFlags(logger log.Logger, features string) (Flagger, error) {
 		opt(fc)
 	}
 
+	if fc.stableMatchersParsing && fc.utf8MatchersParsing {
+		return nil, errors.New("Both stable and UTF-8 matchers parsing is enabled, please choose one or remove the flag for both")
```
grobinson-grafana, line 111: As discussed, this is an error.

matchers/compat/parse.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/go-kit/log"
+	"github.com/go-kit/log/level"
+
+	"github.com/prometheus/alertmanager/featurecontrol"
+	"github.com/prometheus/alertmanager/matchers/parse"
+	"github.com/prometheus/alertmanager/pkg/labels"
+)
+
+var (
+	parseMatcher  = stableMatcherParser(log.NewNopLogger())
```
grobinson-grafana, line 29: The defaults are initialized with a `NewNopLogger` for tests. However, the initialization in `cli/root.go` will always change this to a `NewLogfmtLogger`.

grobinson-grafana:
Examples of amtool:

**Stable matchers parsing**

Here we are creating two alerts, and their respective silences, in amtool using `foo` and `alertname=bar baz` using the stable matchers parsing feature flag (`pkg/labels`):

```sh
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="stable-matchers-parsing" alert add foo
level=warn msg="Enabled stable matchers parsing"
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="stable-matchers-parsing" alert query
level=warn msg="Enabled stable matchers parsing"
Alertname  Starts At                  Summary  State
foo        2023-09-21 10:05:49 +0100           active

./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="stable-matchers-parsing" alert add "alertname=bar baz"
level=warn msg="Enabled stable matchers parsing"
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="stable-matchers-parsing" alert query
level=warn msg="Enabled stable matchers parsing"
Alertname  Starts At                  Summary  State
bar baz    2023-09-21 10:08:54 +0100           active
```

```sh
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="stable-matchers-parsing" silence add foo --comment="silence alertname=foo"
level=warn msg="Enabled stable matchers parsing"
93ab0290-cd7a-44f7-8f22-6fcf1cd7c58e
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="stable-matchers-parsing" silence query
level=warn msg="Enabled stable matchers parsing"
ID                                    Matchers         Ends At                  Created By  Comment
93ab0290-cd7a-44f7-8f22-6fcf1cd7c58e  alertname="foo"  2023-09-21 10:07:02 UTC  grobinson   silence alertname=foo

./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="stable-matchers-parsing" silence add "alertname=bar baz" -c="silence alertname=bar baz"
level=warn msg="Enabled stable matchers parsing"
58ae7810-6f5a-428d-be72-d4e3c6da5722
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="stable-matchers-parsing" silence query
level=warn msg="Enabled stable matchers parsing"
ID                                    Matchers             Ends At                  Created By  Comment
58ae7810-6f5a-428d-be72-d4e3c6da5722  alertname="bar baz"  2023-09-21 10:10:09 UTC  grobinson   =silence alertname=bar baz
```

**UTF-8 matchers parsing**

Here we are creating the same alerts as above but this time using the UTF-8 matchers parsing feature flag. You'll notice that when we use `"alertname=bar baz"` the alerts and silences come out as `alertname="alertname=bar baz"` because of the heuristic that attempts to detect the first label as the alertname. This is different for the strict and fallback parsers which create the alert and silence as `alertname="bar baz"` instead of `alertname="alertname=bar baz"`.

```sh
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="utf8-matchers-parsing" alert add foo
level=warn msg="Enabled UTF-8 matchers parsing"
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="utf8-matchers-parsing" alert query
level=warn msg="Enabled UTF-8 matchers parsing"
Alertname  Starts At                  Summary  State
foo        2023-09-21 10:11:34 +0100           active

./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="utf8-matchers-parsing" alert add "alertname=bar baz"
level=warn msg="Enabled UTF-8 matchers parsing"
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="utf8-matchers-parsing" alert query
level=warn msg="Enabled UTF-8 matchers parsing"
Alertname          Starts At                  Summary  State
alertname=bar baz  2023-09-21 10:19:14 +0100           active
```

```sh
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="utf8-matchers-parsing" silence add foo -c="silence alertname=foo"
level=warn msg="Enabled UTF-8 matchers parsing"
7b5a593d-416c-4561-a5d7-a50ca71f0864
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="utf8-matchers-parsing" silence query
level=warn msg="Enabled UTF-8 matchers parsing"
ID                                    Matchers         Ends At                  Created By  Comment
7b5a593d-416c-4561-a5d7-a50ca71f0864  alertname="foo"  2023-09-21 10:20:33 UTC  grobinson   =silence alertname=foo

./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="utf8-matchers-parsing" silence add "alertname=bar baz" -c="silence alertname=bar baz"
level=warn msg="Enabled UTF-8 matchers parsing"
534cc784-913e-4f5c-90ac-9d05bd127034
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="utf8-matchers-parsing" silence query
level=warn msg="Enabled UTF-8 matchers parsing"
ID                                    Matchers                       Ends At                  Created By  Comment
534cc784-913e-4f5c-90ac-9d05bd127034  alertname="alertname=bar baz"  2023-09-21 10:21:11 UTC  grobinson   =silence alertname=foo
```

**Fallback**

Here we are creating the same alerts as above but this time using the fallback parser (i.e. the default). The error message comes out twice because amtool parses the invalid input twice. I'm not sure if this is an issue or not.

```sh
./amtool --alertmanager.url=http://127.0.0.1:9093 alert add foo
./amtool --alertmanager.url=http://127.0.0.1:9093 alert query
Alertname  Starts At                  Summary  State
foo        2023-09-21 10:27:11 +0100           active

./amtool --alertmanager.url=http://127.0.0.1:9093 alert add "alertname=bar baz"
level=warn msg="Alertmanager is moving to a new parser for label matchers, and this input is incompatible. Alertmanager has instead parsed the input using the old matchers parser as a fallback. To make this input compatible with the new parser please make sure all regular expressions and values are double-quoted. If you are still seeing this message please open an issue." input="alertname=bar baz" err="14:17: unexpected bar: expected a comma or close brace" suggestion="alertname=\"bar baz\""
level=warn msg="Alertmanager is moving to a new parser for label matchers, and this input is incompatible. Alertmanager has instead parsed the input using the old matchers parser as a fallback. To make this input compatible with the new parser please make sure all regular expressions and values are double-quoted. If you are still seeing this message please open an issue." input="alertname=bar baz" err="14:17: unexpected bar: expected a comma or close brace" suggestion="alertname=\"bar baz\""
./amtool --alertmanager.url=http://127.0.0.1:9093 alert query
Alertname  Starts At                  Summary  State
bar baz    2023-09-21 10:27:37 +0100           active
```

```sh
./amtool --alertmanager.url=http://127.0.0.1:9093 silence add foo -c="silence alertname=foo"
cbe74cb0-0c57-4317-9955-a1130b45fb5b
./amtool --alertmanager.url=http://127.0.0.1:9093 silence query
ID                                    Matchers         Ends At                  Created By  Comment
cbe74cb0-0c57-4317-9955-a1130b45fb5b  alertname="foo"  2023-09-21 10:32:28 UTC  grobinson   =silence alertname=foo

./amtool --alertmanager.url=http://127.0.0.1:9093 silence add "alertname=bar baz" -c="silence alertname=bar baz"
level=warn msg="Alertmanager is moving to a new parser for label matchers, and this input is incompatible. Alertmanager has instead parsed the input using the old matchers parser as a fallback. To make this input compatible with the new parser please make sure all regular expressions and values are double-quoted. If you are still seeing this message please open an issue." input="alertname=bar baz" err="14:17: unexpected baz: expected a comma or close brace" suggestion="alertname=\"bar baz\""
level=warn msg="Alertmanager is moving to a new parser for label matchers, and this input is incompatible. Alertmanager has instead parsed the input using the old matchers parser as a fallback. To make this input compatible with the new parser please make sure all regular expressions and values are double-quoted. If you are still seeing this message please open an issue." input="alertname=bar baz" err="14:17: unexpected baz: expected a comma or close brace" suggestion="alertname=\"bar baz\""
74f79844-e087-4667-b1f0-8128df37f283
./amtool --alertmanager.url=http://127.0.0.1:9093 silence query
ID                                    Matchers             Ends At                  Created By  Comment
74f79844-e087-4667-b1f0-8128df37f283  alertname="bar baz"  2023-09-21 10:33:30 UTC  grobinson   =silence alertname=bar baz
```

grobinson-grafana:
Examples of amtool:

**Stable matchers parsing**

Here we are creating two alerts, and their respective silences, in amtool using `foo` and `alertname=bar baz` using the stable matchers parsing feature flag (`pkg/labels`):

```sh
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="stable-matchers-parsing" alert add foo
level=warn msg="Enabled stable matchers parsing"
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="stable-matchers-parsing" alert query
level=warn msg="Enabled stable matchers parsing"
Alertname  Starts At                  Summary  State
foo        2023-09-21 10:05:49 +0100           active

./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="stable-matchers-parsing" alert add "alertname=bar baz"
level=warn msg="Enabled stable matchers parsing"
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="stable-matchers-parsing" alert query
level=warn msg="Enabled stable matchers parsing"
Alertname  Starts At                  Summary  State
bar baz    2023-09-21 10:08:54 +0100           active
```

```sh
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="stable-matchers-parsing" silence add foo --comment="silence alertname=foo"
level=warn msg="Enabled stable matchers parsing"
93ab0290-cd7a-44f7-8f22-6fcf1cd7c58e
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="stable-matchers-parsing" silence query
level=warn msg="Enabled stable matchers parsing"
ID                                    Matchers         Ends At                  Created By  Comment
93ab0290-cd7a-44f7-8f22-6fcf1cd7c58e  alertname="foo"  2023-09-21 10:07:02 UTC  grobinson   silence alertname=foo

./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="stable-matchers-parsing" silence add "alertname=bar baz" -c="silence alertname=bar baz"
level=warn msg="Enabled stable matchers parsing"
58ae7810-6f5a-428d-be72-d4e3c6da5722
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="stable-matchers-parsing" silence query
level=warn msg="Enabled stable matchers parsing"
ID                                    Matchers             Ends At                  Created By  Comment
58ae7810-6f5a-428d-be72-d4e3c6da5722  alertname="bar baz"  2023-09-21 10:10:09 UTC  grobinson   =silence alertname=bar baz
```

**UTF-8 matchers parsing**

Here we are creating the same alerts as above but this time using the UTF-8 matchers parsing feature flag. You'll notice that when we use `"alertname=bar baz"` the alerts and silences come out as `alertname="alertname=bar baz"` because of the heuristic that attempts to detect the first label as the alertname. This is different for the strict and fallback parsers which create the alert and silence as `alertname="bar baz"` instead of `alertname="alertname=bar baz"`.

```sh
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="utf8-matchers-parsing" alert add foo
level=warn msg="Enabled UTF-8 matchers parsing"
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="utf8-matchers-parsing" alert query
level=warn msg="Enabled UTF-8 matchers parsing"
Alertname  Starts At                  Summary  State
foo        2023-09-21 10:11:34 +0100           active

./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="utf8-matchers-parsing" alert add "alertname=bar baz"
level=warn msg="Enabled UTF-8 matchers parsing"
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="utf8-matchers-parsing" alert query
level=warn msg="Enabled UTF-8 matchers parsing"
Alertname          Starts At                  Summary  State
alertname=bar baz  2023-09-21 10:19:14 +0100           active
```

```sh
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="utf8-matchers-parsing" silence add foo -c="silence alertname=foo"
level=warn msg="Enabled UTF-8 matchers parsing"
7b5a593d-416c-4561-a5d7-a50ca71f0864
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="utf8-matchers-parsing" silence query
level=warn msg="Enabled UTF-8 matchers parsing"
ID                                    Matchers         Ends At                  Created By  Comment
7b5a593d-416c-4561-a5d7-a50ca71f0864  alertname="foo"  2023-09-21 10:20:33 UTC  grobinson   =silence alertname=foo

./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="utf8-matchers-parsing" silence add "alertname=bar baz" -c="silence alertname=bar baz"
level=warn msg="Enabled UTF-8 matchers parsing"
534cc784-913e-4f5c-90ac-9d05bd127034
./amtool --alertmanager.url=http://127.0.0.1:9093 --enable-feature="utf8-matchers-parsing" silence query
level=warn msg="Enabled UTF-8 matchers parsing"
ID                                    Matchers                       Ends At                  Created By  Comment
534cc784-913e-4f5c-90ac-9d05bd127034  alertname="alertname=bar baz"  2023-09-21 10:21:11 UTC  grobinson   =silence alertname=foo
```

**Fallback**

Here we are creating the same alerts as above but this time using the fallback parser (i.e. the default). The error message comes out twice because amtool parses the invalid input twice. I'm not sure if this is an issue or not.

```sh
./amtool --alertmanager.url=http://127.0.0.1:9093 alert add foo
./amtool --alertmanager.url=http://127.0.0.1:9093 alert query
Alertname  Starts At                  Summary  State
foo        2023-09-21 10:27:11 +0100           active

./amtool --alertmanager.url=http://127.0.0.1:9093 alert add "alertname=bar baz"
level=warn msg="Alertmanager is moving to a new parser for label matchers, and this input is incompatible. Alertmanager has instead parsed the input using the old matchers parser as a fallback. To make this input compatible with the new parser please make sure all regular expressions and values are double-quoted. If you are still seeing this message please open an issue." input="alertname=bar baz" err="14:17: unexpected bar: expected a comma or close brace" suggestion="alertname=\"bar baz\""
level=warn msg="Alertmanager is moving to a new parser for label matchers, and this input is incompatible. Alertmanager has instead parsed the input using the old matchers parser as a fallback. To make this input compatible with the new parser please make sure all regular expressions and values are double-quoted. If you are still seeing this message please open an issue." input="alertname=bar baz" err="14:17: unexpected bar: expected a comma or close brace" suggestion="alertname=\"bar baz\""
./amtool --alertmanager.url=http://127.0.0.1:9093 alert query
Alertname  Starts At                  Summary  State
bar baz    2023-09-21 10:27:37 +0100           active
```

```sh
./amtool --alertmanager.url=http://127.0.0.1:9093 silence add foo -c="silence alertname=foo"
cbe74cb0-0c57-4317-9955-a1130b45fb5b
./amtool --alertmanager.url=http://127.0.0.1:9093 silence query
ID                                    Matchers         Ends At                  Created By  Comment
cbe74cb0-0c57-4317-9955-a1130b45fb5b  alertname="foo"  2023-09-21 10:32:28 UTC  grobinson   =silence alertname=foo

./amtool --alertmanager.url=http://127.0.0.1:9093 silence add "alertname=bar baz" -c="silence alertname=bar baz"
level=warn msg="Alertmanager is moving to a new parser for label matchers, and this input is incompatible. Alertmanager has instead parsed the input using the old matchers parser as a fallback. To make this input compatible with the new parser please make sure all regular expressions and values are double-quoted. If you are still seeing this message please open an issue." input="alertname=bar baz" err="14:17: unexpected baz: expected a comma or close brace" suggestion="alertname=\"bar baz\""
level=warn msg="Alertmanager is moving to a new parser for label matchers, and this input is incompatible. Alertmanager has instead parsed the input using the old matchers parser as a fallback. To make this input compatible with the new parser please make sure all regular expressions and values are double-quoted. If you are still seeing this message please open an issue." input="alertname=bar baz" err="14:17: unexpected baz: expected a comma or close brace" suggestion="alertname=\"bar baz\""
74f79844-e087-4667-b1f0-8128df37f283
./amtool --alertmanager.url=http://127.0.0.1:9093 silence query
ID                                    Matchers             Ends At                  Created By  Comment
74f79844-e087-4667-b1f0-8128df37f283  alertname="bar baz"  2023-09-21 10:33:30 UTC  grobinson   =silence alertname=bar baz
```

featurecontrol/featurecontrol.go
```diff
@@ -61,6 +92,12 @@ func NewFlags(logger log.Logger, features string) (Flagger, error) {
 		case fcReceiverNameInMetrics:
 			opts = append(opts, enableReceiverNameInMetrics())
 			level.Warn(logger).Log("msg", "Experimental receiver name in metrics enabled")
+		case fcStableMatchersParsing:
+			opts = append(opts, enableStableMatchersParsing())
+			level.Warn(logger).Log("msg", "Enabled stable matchers parsing")
```
grobinson-grafana, line 97: Put `enabled` at the end of the sentence to be consistent with Line 94

featurecontrol/featurecontrol.go
grobinson-grafana: `are` instead of `is`

cli/utils.go
```diff
@@ -107,7 +105,7 @@ func parseLabels(inputLabels []string) (models.LabelSet, error) {
 	labelSet := make(models.LabelSet, len(inputLabels))
 
 	for _, l := range inputLabels {
-		matcher, err := labels.ParseMatcher(l)
+		matcher, err := compat.Matcher(l)
```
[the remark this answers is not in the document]
gotjosh: Ok, after further inspection, I think it's worth stopping here for a moment to address these inconsistencies.

I see that:

1. `alert_add.go` uses `parseLabels` which under the hood uses `compat.Matcher`
2. `alert_query.go` uses `compat.Matcher` directly.
3. `silence_add.go` uses `parseMatchers` which under the hood uses `compat.Matcher`
4. `silence_query.go` uses `parseMatchers` which under the hood uses `compat.Matcher`

My opinion is that we should get rid of `parseMatchers` and `parseLabels` in `utils.go` and instead have everything in the CLI use `compat.*` directly. 

With that, we can leave the introduction of `compat.Label` for later but at least we're certain that we don't have this weird split of a direct use of `compat.*` and the util functions in `parse*`.

matchers/compat/parse.go
```diff
@@ -0,0 +1,218 @@
...
+
+// Matcher parses the matcher in the input string. It returns an error
+// if the input is invalid or contains two or more matchers.
+func Matcher(s string) (*labels.Matcher, error) {
+	return parseMatcher(s)
+}
+
+// Matchers parses one or more matchers in the input string. It returns
+// an error if the input is invalid.
+func Matchers(s string) (labels.Matchers, error) {
```
gotjosh, line 45: I went on a hunt to try and understand why we weren't using this function. On that mission, I concluded that we should address the inconsistency I see between the direct use of the functions in the `compat` package  and `utils.go` functions

gotjosh: If we're still not using this function after that refactor - I believe we should remove it.

cli/utils.go
```diff
@@ -40,16 +41,13 @@ func GetAlertmanagerURL(p string) url.URL {
 // parseMatchers parses a list of matchers (cli arguments).
 func parseMatchers(inputMatchers []string) ([]labels.Matcher, error) {
 	matchers := make([]labels.Matcher, 0, len(inputMatchers))
-
```
gotjosh, line 43: drive-by/nit: There's a function in this file `GetAlertmanagerURL` that is also not used, can you remove it as part of this change? If not, I'll create a separate PR for it.

matchers/compat/parse.go
```diff
@@ -0,0 +1,218 @@
...
+	if f.StableMatchersParsing() {
+		parseMatcher = stableMatcherParser(l)
+		parseMatchers = stableMatchersParser(l)
+	} else if f.UTF8MatchersParsing() {
+		parseMatcher = utf8MatcherParser(l)
+		parseMatchers = utf8MatchersParser(l)
+	} else {
+		parseMatcher = fallbackMatcherParser(l)
+		parseMatchers = fallbackMatchersParser(l)
+	}
```
gotjosh, line 60:
```suggestion
	if f.StableMatchersParsing() {
	    parseMatcher = stableMatcherParser(l)
	    parseMatchers = stableMatchersParser(l)
	}
	
        if f.UTF8MatchersParsing() {
	  parseMatcher = utf8MatcherParser(l)
	  parseMatchers = utf8MatchersParser(l)
	}
	
	parseMatcher = fallbackMatcherParser(l)
	parseMatchers = fallbackMatchersParser(l)
```

nit: I find this to be a bit easier to read - there's no way `UTF8MatchersParsing` and `StableMatchersParsing` can both we true.

gotjosh, line 60:
```suggestion
	if f.StableMatchersParsing() {
	    parseMatcher = stableMatcherParser(l)
	    parseMatchers = stableMatchersParser(l)
	}
	
        if f.UTF8MatchersParsing() {
	  parseMatcher = utf8MatcherParser(l)
	  parseMatchers = utf8MatchersParser(l)
	}
	
	parseMatcher = fallbackMatcherParser(l)
	parseMatchers = fallbackMatchersParser(l)
```

nit: I find this to be a bit easier to read - there's no way `UTF8MatchersParsing` and `StableMatchersParsing` can both be true.

matchers/compat/parse.go
```diff
@@ -0,0 +1,218 @@
...
+		if strings.HasPrefix(s, "{") || strings.HasSuffix(s, "}") {
+			return nil, fmt.Errorf("unexpected open or close brace: %s", s)
+		}
+		m, err = parse.Matcher(s)
+		if err != nil {
+			// The input is not valid in the old pkg/labels parser either,
+			// it cannot be valid input.
+			m, invalidErr = labels.ParseMatcher(s)
+			if invalidErr != nil {
+				return nil, invalidErr
```
gotjosh, line 149:
```suggestion

			m, invalidErr = labels.ParseMatcher(s)
			if invalidErr != nil {
			         // The input is not valid in the old pkg/labels parser either, it cannot be valid input.
				return nil, invalidErr
```
I got confused by this comment how do you feel about moving it inside the check?

matchers/compat/parse.go
```diff
@@ -0,0 +1,218 @@
...
+			}
+			// The input is valid in the old pkg/labels parser, but not the
+			// new matchers/parse parser.
+			suggestion := m.String()
+			level.Warn(l).Log(
+				"msg",
+				"Alertmanager is moving to a new parser for label matchers, and this input is incompatible. Alertmanager has instead parsed the input using the old matchers parser as a fallback. To make this input compatible with the new parser please make sure all regular expressions and values are double-quoted. If you are still seeing this message please open an issue.",
```
gotjosh, line 156:
```suggestion
				"Alertmanager is moving to a new parser for labels and matchers, and this input is incompatible. Alertmanager has instead parsed the input using the old matchers parser as a fallback. To make this input compatible with the new parser, please make sure all regular expressions and values are double-quoted. If you are still seeing this message, please open an issue.",
```

gotjosh: We use this for both labels (e.g. alert queries) and matchers.

gotjosh (commented):
Looks good so far, but please take a look at my comments.

cli/utils.go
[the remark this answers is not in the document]
grobinson-grafana: > I also don't love the fact that we execute this check as part of parseLabels:

Yeah! The issue here is `amtool` is misusing the label matchers parser to parse labels, despite labels and label matchers being quite different. I wasn't planning to address that in this work though as the root issue that needs to be addressed is finding a way to properly parse labels, instead of misuing the label matchers parser to parse input that isn't really a label matcher.

> My opinion is that we should get rid of parseMatchers and parseLabels in utils.go and instead have everything in the CLI use compat.* directly.

So would we inline the code (copy it) from `utils.go` in `alert_add.go`, `silence_add.go` and `silence_query.go`?

matchers/compat/parse.go
grobinson-grafana: This will be used in Alertmanager server, both in `api/v1/api.go` and `config/config.go`. I'm surprised it's not used in `api/v2/` though, I need to look into that.

matchers/compat/parse.go
grobinson-grafana: It's not correct though? This code will forever use the fallback parser?

cli/utils.go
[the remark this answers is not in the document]
gotjosh: > So would we inline the code (copy it) from utils.go in alert_add.go, silence_add.go and silence_query.go?

I believe so, I think it's more "clean" in a way to execute the refactor later as it is more obvious. At least everything now depends on `compat`.

cli/utils.go
```diff
@@ -107,7 +105,7 @@ func parseLabels(inputLabels []string) (models.LabelSet, error) {
 	labelSet := make(models.LabelSet, len(inputLabels))
 
 	for _, l := range inputLabels {
-		matcher, err := labels.ParseMatcher(l)
```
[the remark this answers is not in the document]
grobinson-grafana: Sure! Fwiw, I consider this behaviour broken:

> We use the same parser for labels as for matchers because they follow the same grammar

I wrote in another comment about fixing it in future. Labels cannot have `{` or trailing commas, so the grammar is not the same.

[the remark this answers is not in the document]
grobinson-grafana: Sure! Fwiw, I consider this behaviour broken:

> We use the same parser for labels as for matchers because they follow the same grammar

I wrote in another comment about fixing it in future. Labels cannot start with `{`, end with `}` or having trailing commas, so the grammar is not the same.

matchers/compat/parse.go
```diff
@@ -0,0 +1,218 @@
...
+// stableMatcherParser uses the old pkg/labels parser to parse the matcher in
+// the input string.
+func stableMatcherParser(l log.Logger) matcherParser {
+	return func(s string) (*labels.Matcher, error) {
+		level.Debug(l).Log(
+			"msg",
+			"Parsing with stable matchers parser",
+			"input",
+			s,
+		)
```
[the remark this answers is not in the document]
grobinson-grafana: It's just preference, will remove! Should I do the same for the large comment on Line 154 and 207 too?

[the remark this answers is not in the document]
gotjosh: Yes please.

cli/root.go
```diff
@@ -154,6 +175,8 @@ func Execute() {
 	configureConfigCmd(app)
 	configureTemplateCmd(app)
 
+	app.Action(initMatchersCompat)
```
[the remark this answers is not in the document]
grobinson-grafana: This is more or less documented in the doc comments for `app.Action`:

> Action callback to call when all values are populated and parsing is complete, but before any command, flag or argument actions.

If you still want it to be documented here I'm happy to make that change, but just wanted to share this information in case it changes your opinion.

cli/utils.go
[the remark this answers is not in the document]
grobinson-grafana: I think this comment is irrelevant now as we are going to remove the `parseLabels` as per https://github.com/prometheus/alertmanager/pull/3483#discussion_r1334205799. I can inline the comment where the function used to be though, if that's helpful?

cli/utils.go
grobinson-grafana: [Remove unused function GetAlertmanagerURL #3535](https://github.com/prometheus/alertmanager/pull/3535)

grobinson-grafana:
I had to force push to fix the conflict in `cli/utils.go`.

gotjosh (approved):
LGTM, any thoughts on my comment? I'll let this bake for a day or two in case @simonpasquier has any comments.

PR #3487, opened 2023-08-25
Add a `[SECURITY]` denomination to the changelog and update CVE-2023-…

gotjosh:
…40577

Some users have been vocal about the security fix not visible enough in the changelog, it seems like `prometheus` uses `[SECURITY]` to disclose these and I think it's a good practice.


simonpasquier:
Merging #3489 would bring the same thing with a cleaner Git history ;)

PR #3491, opened 2023-08-28

alexweav:
This PR is an extension of @OktarianTB's 9-month old PR https://github.com/prometheus/alertmanager/pull/3139 that's updated to work against the latest code in main, to hopefully increase its likelihood of being merged.

Credit for this work goes to @OktarianTB - I believe I've marked the commit authorship correctly, please let me know if I made any mistakes here.

> This PR aims to make progress towards #2845 by adding a new command to `amtool` to send test notifications to every receiver in an alertmanager config file. The command takes in an alertmanager config file and optionally an alert file to mock values for labels and annotations. This is the MVP for the command, potentially a few other options can be added later if deemed of use. This largely re-uses the equivalent implementation in [grafana/grafana](https://github.com/grafana/grafana/blob/main/pkg/services/ngalert/notifier/receivers.go).
> 
> I have only tested this with Slack so far.


OktarianTB:
Thanks for addressing the issues directly - I've lost most of the context since making that PR last year but hopefully it comes in useful 😄

gotjosh:
We'll put it to good use; thanks very much @OktarianTB.

alexweav:
I opened https://github.com/prometheus/alertmanager/pull/3553 to reduce some of the code duplication here.

SuperQ:
Hello from the Alertmanager bug scrub. This change needs a rebase.

PR #3494, opened 2023-08-31
[WIP] Validate notifier configurations statically.

### LINKED ISSUE #3490: Notifications are not resilient to bad templates
**What did you do?**

Mis-configured a notification such that a template string was invalid and could not be rendered.

* I expected `amtool check-config` to detect template errors in notifier configuration, and/or
* I expected notifiers to make a best-effort to attempt to notify despite configuration bugs, since failing to notify is a very bad outcome. A notification with incomplete data would be much more useful than no notification.

Validation with `amtool` didn't show any errors, and notifications were dropped.

**Environment**

We're running alertmanager 0.25, but I believe this issue affects all versions.

Logs look like this:

alertmanager ts=2023-08-25T20:43:03.251Z caller=dispatch.go:352 level=error component=dispatcher msg="Notify for alerts failed" num_alerts=1 err="opsgenie-receiver/opsgenie[0]: notify retry canceled due to unrecoverable error after 1 attempts: templating error: template: :1: bad character U+002D '-'"

I would be happy to write a pr to either cause amtool to check for invalid templates or to enable notifiers to send partial notifications on template errors—for example, dropping fields that fail to render. Or if there's a better approach, let me know!

jmcarp:
Resolves #3490.

As discussed in #3490, notifiers may have invalid configuration, including but not limited to invalid inline templates, that isn't detected until alertmanager attempts to send a notification. At that point, alertmanager drops alerts. This patch introduces a pattern, implemented so far for a few notifiers, of validating notifier configuration in the notifier constructor.

We discussed in #3490 whether we should add validation for notifiers one by one or all at once. As of this writing, I've drafted validation for the opsgenie and slack receivers, and you can see that both were pretty simple to implement, since we can mostly reuse the code that actually sends the notifications—specifically, we can build a request using an empty alert and check for errors without actually sending a notification. If we want to implement validation for all notifiers before merging any changes, I'm happy to continue working through the list of notifiers.

cc @grobinson-grafana @simonpasquier @gotjosh

### DIFF
--- a/notify/opsgenie/opsgenie_test.go
+++ b/notify/opsgenie/opsgenie_test.go
@@ -114,15 +114,13 @@ func TestOpsGenieValidation(t *testing.T) {
 			APIKey:     config.Secret(key),
 			HTTPConfig: &commoncfg.HTTPClientConfig{},
 			Message:    "{{ invalid",
-			// Details: map[string]string{
-			// 	"label": "{{ invalid",
-			// },
 		},
 		test.CreateTmpl(t),
 		log.NewNopLogger(),
 	)
 	require.ErrorContains(t, err, `function "invalid" not defined`)
 }
+
 func TestOpsGenie(t *testing.T) {
 	u, err := url.Parse("https://opsgenie/api")
 	if err != nil {
--- a/notify/slack/slack.go
+++ b/notify/slack/slack.go
@@ -55,14 +55,28 @@ func New(c *config.SlackConfig, t *template.Template, l log.Logger, httpOpts ...
 		return nil, err
 	}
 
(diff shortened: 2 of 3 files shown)


jmcarp:
I noticed that this might overlap with #3491 from @alexweav and @OktarianTB. Let me know if this is redundant! Although I think I still like the idea of being able to validate receiver/notifier configs statically—for example, I would probably prefer static validation over sending a real notification in a CI environment, which I don't want to depend on third-party APIs.

SuperQ:
Hello from the Alertmanager bug scrub. This change needs a rebase.

PR #3502, opened 2023-09-02
Cleanup golangci-lint errcheck

SuperQ:
Move the errcheck excludes list from an external file to inline in the golangci-lint config file.


gotjosh (approved):
LGTM

PR #3507, opened 2023-09-05
Add test for PromQL braces when parsing lists of matchers

grobinson-grafana:
This pull request has been added to address the discussion https://github.com/prometheus/alertmanager/pull/3453#discussion_r1311431141.


grobinson-grafana:
@gotjosh could you restart `test_frontend`?

grobinson-grafana:
As discussed offline, we are going to replace the unit test with acceptance tests. I've just pushed that commit this morning, and so a second review would be great whenever convenient! :slightly_smiling_face:

test/cli/acceptance/cli_test.go
```diff
@@ -153,18 +167,20 @@ receivers:
 
 	am := amc.Members()[0]
 
-	silence1 := Silence(0, 4).Match("alertname=test1", "severity=warn").Comment("test1")
+	silence1 := Silence(0, 4).Match("test1", "severity=warn").Comment("test1")
 	silence2 := Silence(0, 4).Match("foo").Comment("test foo")
```
grobinson-grafana, line 172: @gotjosh here I plan to also create a silence like:

```go
silence1 := Silence(0, 4).Match("{alertname=test1}", "severity=warn")
```

This will create a silence with the following labels:

```
alertname="{alertname=test3}"
severity=warn
```

We will be able to then see this behaviour break when we use the new fallback parser, as without the additional check, it will create a silence with the following labels instead:

```
alertname=test3"
severity=warn
```

I think this is the right test to write, what do you think? If we're both in agreement I'll add it to the pull request too!

test/cli/acceptance/cli_test.go
```diff
@@ -153,18 +167,28 @@ receivers:
 
 	am := amc.Members()[0]
 
-	silence1 := Silence(0, 4).Match("alertname=test1", "severity=warn").Comment("test1")
-	silence2 := Silence(0, 4).Match("foo").Comment("test foo")
+	silence1 := Silence(0, 4).Match("test1", "severity=warn").Comment("test1")
+	silence2 := Silence(0, 4).Match("alertname=test2", "severity=warn").Comment("test2")
+	silence3 := Silence(0, 4).Match("{alertname=test3}", "severity=warn").Comment("test3")
```
grobinson-grafana, line 172: @gotjosh here I create a silence with the following labels:

```
alertname="{alertname=test3}"
severity=warn
```

We will be able to then see this behaviour break when we use the new fallback parser as without the additional check we talked about in https://github.com/prometheus/alertmanager/pull/3453#discussion_r1311431141, it will create a silence with the following labels instead:

```
alertname=test3
severity=warn
```

gotjosh (approved):
LGTM

Issue #3505, opened 2023-09-04
Regression in 0.26.0 when using `amtool config rules` with `webhook_configs`


czenker:
**What did you do?**

Update from version `0.25.0` to `0.26.0`.

**What did you expect to see?**

We use `amtool config routes test alertname=Foobar severity=panic ` in an automated test to ensure there is no misconfiguration that would not route alerts into on-call. We expected the command to work as before in the sense, that it would tell us which receivers would receive that alert.

**What did you see instead? Under which circumstances?**

With amtool `0.26.0`:
````
error: scheme required for webhook url
````

With amtool `0.25.0` (error message is different, but the problem is the same):
````
Warning: amtool version (0.25.0) and alertmanager version (0.26.0) are different.
amtool: error: unsupported scheme "" for URL
````
amtool exits with status code `1` regardless and does not print the matched routes.

This is very likely a regression from https://github.com/prometheus/alertmanager/pull/3228 as the API does not output the url anymore, but amtool seems to be unaware of it and detects an error.

It also affects the `amtool config routes show` command.

**Alertmanager configuration file**

This probably only happens when using a `webhook_configs` with a `url`:

```
global: { } # ...
route: { } # ...
receivers:
- name: oncall-alerts
  webhook_configs:
  - url: http://example.com/webhook
templates: [] # ...
```

As a workaround the new `url_file` could be used as this does not seem to validate any scheme.

**Comment by grobinson-grafana** [comment]
Looks like this is an issue when using `amtool` with `--alertmanager.url`. It does not affect amtool when using `--config.file.`

**Comment by grobinson-grafana** [comment]
I think the issue is as @czenker explained.

For example, the command `amtool config routes show` fails on [Line 65](https://github.com/prometheus/alertmanager/blob/v0.26.0/cli/routing.go#L65) of `cli/routing.go`. This command seems to use the `/api/v2/status` API which returns the Alertmanager configuration but with the URL field for each `webhook_configs` as `<secret>`. However, if I revert the [URL](https://github.com/prometheus/alertmanager/blob/main/config/notifiers.go#L484) field for `WebhookConfig` from `*SecretURL` back to `*URL` then it works as the URL is now returned in the  `/api/v2/status` API:

```
./amtool config routes show --alertmanager.url=http://127.0.0.1:9093
Routing tree:
.
└── default-route  receiver: example

```

[...]

PR #3509, opened 2023-09-05
Fix scheme required for webhook url in amtool

grobinson-grafana:
This commit fixes issue #3505 where amtool would fail with `error: scheme required for webhook url` when using amtool with `--alertmanager.url`.

The issue here is that `UnmarshalYaml` for `WebhookConfig` checks if the scheme is present when `c.URL` is non-nil. However, `UnmarshalYam`l for `SecretURL` returns a non-nil, default value `url.URL{}` if the response from `api/v2/status` contains `<secret>` as the webhook URL.

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -503,11 +503,6 @@ func (c *WebhookConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if c.URL != nil && c.URLFile != "" {
 		return fmt.Errorf("at most one of url & url_file must be configured")
 	}
-	if c.URL != nil {
-		if c.URL.Scheme != "https" && c.URL.Scheme != "http" {
-			return fmt.Errorf("scheme required for webhook url")
-		}
-	}
 	return nil
 }
 
--- a/test/cli/acceptance.go
+++ b/test/cli/acceptance.go
@@ -658,6 +658,17 @@ func (am *Alertmanager) UpdateConfig(conf string) {
 	}
 }
 
+func (am *Alertmanager) ShowRoute() ([]byte, error) {
+	return am.showRouteCommand()
+}
+
+func (am *Alertmanager) showRouteCommand() ([]byte, error) {
+	amURLFlag := "--alertmanager.url=" + am.getURL("/")
+	args := []string{amURLFlag, "config", "routes", "show"}
+	cmd := exec.Command(amtool, args...)
+	return cmd.CombinedOutput()
+}
+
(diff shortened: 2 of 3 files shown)


test/cli/acceptance/cli_test.go
```diff
@@ -168,3 +168,36 @@ receivers:
 		t.Errorf("Incorrect number of silences queried, expected: %v, actual: %v", expectedSils, len(sils))
 	}
 }
+
+func TestRoutesShow(t *testing.T) {
```
grobinson-grafana, line 172: Revert the fix and this test fails:

```
--- FAIL: TestRoutesShow (1.09s)
    cli_test.go:202: amtool: error: scheme required for webhook url

    cli_test.go:203:
        	Error Trace:	/Users/grobinson/go/src/github.com/prometheus/alertmanager/test/cli/acceptance/cli_test.go:203
        	Error:      	Received unexpected error:
        	            	exit status 1
        	Test:       	TestRoutesShow
FAIL
```

config/notifiers.go
```diff
@@ -503,11 +503,6 @@ func (c *WebhookConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if c.URL != nil && c.URLFile != "" {
 		return fmt.Errorf("at most one of url & url_file must be configured")
 	}
-	if c.URL != nil {
```
grobinson-grafana, line 506: We think this is duplicate code as `SecretURL` is just a `URL`, whose `UnmarshalYaml` function also calls `parseURL` which contains the same check.

simonpasquier:
Thanks! I'd be inclined to have the fix in the release-0.26 branch.

grobinson-grafana:
Apologies for the force push, there was a spelling mistake in the commit message.

test/cli/acceptance.go
```diff
@@ -658,6 +658,17 @@ func (am *Alertmanager) UpdateConfig(conf string) {
...
 }
 
+func (am *Alertmanager) ShowRoute() ([]byte, error) {
+	return am.showRouteCommand()
+}
+
+func (am *Alertmanager) showRouteCommand() ([]byte, error) {
+	amURLFlag := "--alertmanager.url=" + am.getURL("/")
+	args := []string{amURLFlag, "config", "routes", "show"}
+	cmd := exec.Command(amtool, args...)
```
gotjosh, line 668: Can we also add one for `config routes test`?

gotjosh (approved):
LGTM!

PR #3518, opened 2023-09-11
jsm-prometheus integration

ozlemmcokuk:


jkroepke:
@zekihan @ozlemmcokuk is the JSM API compatible with OpsGenie?

ozlemmcokuk:
Hi @jkroepke 
Actually the endpoints for Opsgenie and Jira Service Management are not the same. We are bringing alert&on-call capabilities to Jira Service Management natively, so we will have a new integration on Jira Service Management product which will use this plugin as in Opsgenie. The functionality can be seen the same at first but the capabilities of Jira Service Management is much wider than Opsgenie as a product, so having a dedicated plugin will serve for visibility for both products.

ozlemmcokuk:
Hi @jkroepke
Is there anything that we need to do in this PR?

jkroepke:
I don't know. I not a maintainer, I just had a questions, since we also having Jira and OpsGenie and we may play to merge them.

ozlemmcokuk:
Hi @simonpasquier, @w0rm and @gotjosh 
Could you please help us with our new integration? We are planning to release it in December. Please let us know if you have any comments about the PR. Thanks 🙏

jkroepke:
Whats happens, if you enter the JSM endpoints on the OpsGenie integration?

jkroepke:
@ozlemmcokuk 

We ask the Jira support and they confirm that the API is equal to OpsGenie

> Hello,
> 
> I am glad that the Incoming call integration is working as expected :slight_smile:
> Regarding the Prometheus and Heartbeats, I would like to inform you that there are no changes to the API keys these integration will continue working as it is.
> 
> As a side note, I would like to inform you that the Incoming call routing and Heartbeat feature is available only with the Jira Service Management Premium plan.
> 
> I hope I was able to answer your queries, in case you need further assistance on this ticket, please revert would be happy to help!!
> 
> Best Regards,
> Arjun Ganapathy
> Atlassian Cloud Support

Efficiently, the existing OpsGenie Integration could be used for JSM which make the PR pointless for me. Maintain duplicate code make no sense.

simonpasquier:
Indeed if OpsGenie and JSM are just 2 sides of the same coin, I wouldn't create/maintain another integration.

james00002:
Even though the JSM and OpsGenie integrations are currently the same, Atlassian has stated their intention that the JSM features are going to diverge from OpsGenie in the future, and they are encouraging customers to switch over to the JSM Ops API.

https://support.atlassian.com/jira-service-management-cloud/docs/start-shifting-from-opsgenie-to-jira-service-management/

So even if there is not currently a need for a separate JSM receiver in prometheus, there probably will be a need for it in the semi-near future.

jkroepke:
> So even if there is not currently a need for a separate JSM receiver in prometheus

At the moment, die OpsGenie can be used with JSM endpoints, too

pharaujo:
We got a notice from Atlassian that they're deprecating the Opsgenie API in mid Febrary 2025 by making it read-only, and the migration documentation suggests [moving to the JSM API as soon as possible](https://support.atlassian.com/jira-service-management-cloud/docs/set-up-operations-and-complete-your-shift-to-jira-service-management/#Deprecated-features-and-API-changes). The [JSM REST API](https://developer.atlassian.com/cloud/jira/service-desk-ops/rest/v2/api-group-alerts/#api-group-alerts) doesn't seem compatible with the Opsgenie API ([example](https://github.com/prometheus/alertmanager/blob/main/notify/opsgenie/opsgenie.go#L160)), so it won't be possible to use the Opsgenie receiver with JSM.

jkroepke:
@pharaujo we are using the JSM endpoints with the opsgenie receiver since 3 months in production.

In general, it's correct that the JSM API for alerts is different. However, if you are using the Prometheus integration in JSM, the API including API Key remains the same. You can seamless switch the configuration to 

```
api_url: "https://api.atlassian.com/jsm/ops/integration/"
```

The heartbeat URL is completely different, but that not in scope of the notifier.

SuperQ:

PR #3520, opened 2023-09-13
Muted alerts are supressed and add "mutedBy" to APIv2

### LINKED ISSUE #3513: Muted alerts are not suppressed in API
### What

When an alert is silenced in Alertmanager its status is changed from active to suppressed, and the JSON responses for `/api/v2/alerts` contain the the IDs(s) of all silences that suppressed it:

```yml
[
  {
    "annotations": {},
    "endsAt": "2023-09-07T10:43:50.291+01:00",
    "fingerprint": "3fff2c2d7595e046",
    "receivers": [
      {
        "name": "test"
      }
    ],
    "startsAt": "2023-09-07T10:38:50.291+01:00",
    "status": {
      "inhibitedBy": [],
      "silencedBy": [
        "ed411cae-78ea-47bc-90f5-1e474d6f526e"
      ],
      "state": "suppressed"
    },
    "updatedAt": "2023-09-07T10:38:50.291+01:00",
    "labels": {
      "foo": "bar"
    }
  }
]
```

However, the same is not said for active and mute time intervals:

```
ts=2023-09-07T09:54:39.161Z caller=notify.go:902 level=debug component=dispatcher msg="Notifications not sent, route is within mute time"
```

```yml
  {
    "annotations": {},
    "endsAt": "2023-09-07T10:59:09.159+01:00",
    "fingerprint": "3fff2c2d7595e046",
    "receivers": [
      {
        "name": "test"
      }
    ],
    "startsAt": "2023-09-07T10:54:09.159+01:00",
    "status": {
      "inhibitedBy": [],
      "silencedBy": [],
      "state": "active"
    },
    "updatedAt": "2023-09-07T10:54:09.159+01:00",
    "labels": {
      "foo": "bar"
    }
  }
]
```

I expected to see `"state": "suppressed"` and `"mutedBy"`, just like we have `"inhibitedBy"` and `"silencedBy"`.

### Background

Alertmanager supports a number of features to suppress alerts. These are:

1. Silences
2. Inhibition rules
3. Active time intervals
4. Mute time intervals

However, while silences and inhibition rules target alerts, active and mute time intervals target routes. This means that a silenced or inhibited alert is suppressed irrespective of how it's routed, but for active and mute time intervals an alert can be active when matched to one route and suppressed when matched to another. If `continue` matching is also set to `true`, then an alert can be both active and suppressed at the same time.

### Problems

This presents a number of problems for the Alertmanager API:

1. `/api/v2/alerts` contains a list of Silence IDs and alert rule fingerprints that are silencing or inhibiting the alert. It does not show which groups the alert is silenced or inhibited in as it does not need to. Both silences and inhibition rules work irrespective of routes, so the alert is silenced and inhibited in all groups, even when `continue` matching is set to `true`.
2. Active and mute time intervals are per route, which means an alert can be active when matched to one route and suppressed when matched to another. However, `/api/v2/alerts` does not work with groups. Instead, there is `/api/v2/alerts/groups`. Should `/api/v2/alerts` show which groups an alert is suppressed in when the rest of the endpoint does not return information about groups?
3. What should the state be of an alert that is active in one group but suppressed in another? Should it be active, suppressed, or something else?

### Proposal

1. I propose that we **DO NOT** update `/api/v2/alerts` to show alerts as suppressed from active and mute time intervals. There are two reasons for this: a) it does not feel right to add information about groups when `/api/v2/alerts` does not work with groups b) there is an existing endpoint available that has this data called `/api/v2/alerts/groups`.
2. We update `/api/v2/alerts/groups` to show if an alert is suppressed from only the active and mute time intervals in the route that created the group, and no other routes.
2. In the long term we should remove `status` from `/api/v2/alerts`, which includes `state`, `inhibitedBy` and `suppressedBy`. The state of an alert cannot be fully known until it has been routed and put into a group, and so `status` should be removed from this endpoint. This is a breaking change.

### PRs

The following PRs should be reviewed and merged in the order specified:

1. https://github.com/prometheus/alertmanager/pull/3795
2. https://github.com/prometheus/alertmanager/pull/3794
3. https://github.com/prometheus/alertmanager/pull/3791
4. https://github.com/prometheus/alertmanager/pull/3792
5. https://github.com/prometheus/alertmanager/pull/3793
6. https://github.com/prometheus/alertmanager/pull/3797

grobinson-grafana:
This commit updates TimeMuteStage to mark muted alerts as suppressed, and adds the names of all mutes that mute the alert to a field called mutedBy. This field is also returned in APIv2.

Fixes #3513

### DIFF
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -262,7 +262,7 @@ func (api *API) getAlertsHandler(params alert_ops.GetAlertsParams) middleware.Re
 	alerts := api.alerts.GetPending()
 	defer alerts.Close()
 
-	alertFilter := api.alertFilter(matchers, *params.Silenced, *params.Inhibited, *params.Active)
+	alertFilter := api.alertFilter(matchers, *params.Silenced, *params.Inhibited, *params.Muted, *params.Active)
 	now := time.Now()
 
 	api.mtx.RLock()
@@ -399,7 +399,7 @@ func (api *API) getAlertGroupsHandler(params alertgroup_ops.GetAlertGroupsParams
 		}
 	}(receiverFilter)
 
-	af := api.alertFilter(matchers, *params.Silenced, *params.Inhibited, *params.Active)
+	af := api.alertFilter(matchers, *params.Silenced, *params.Inhibited, *params.Muted, *params.Active)
 	alertGroups, allReceivers := api.alertGroups(rf, af)
 
 	res := make(open_api_models.AlertGroups, 0, len(alertGroups))
@@ -424,7 +424,7 @@ func (api *API) getAlertGroupsHandler(params alertgroup_ops.GetAlertGroupsParams
 	return alertgroup_ops.NewGetAlertGroupsOK().WithPayload(res)
 }
 
-func (api *API) alertFilter(matchers []*labels.Matcher, silenced, inhibited, active bool) func(a *types.Alert, now time.Time) bool {
+func (api *API) alertFilter(matchers []*labels.Matcher, silenced, inhibited, muted, active bool) func(a *types.Alert, now time.Time) bool {
 	return func(a *types.Alert, now time.Time) bool {
 		if !a.EndsAt.IsZero() && a.EndsAt.Before(now) {
 			return false
@@ -445,6 +445,10 @@ func (api *API) alertFilter(matchers []*labels.Matcher, silenced, inhibited, act
 		}
 
 		if !inhibited && len(status.InhibitedBy) != 0 {
+			return false
+		}
+
+		if !muted && len(status.MutedBy) != 0 {
 			return false
 		}
 
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -288,7 +288,7 @@ func (api *API) getAlertsHandler(params alert_ops.GetAlertsParams) middleware.Re
 			continue
 		}
 
-		alert := AlertToOpenAPIAlert(a, api.getAlertStatus(a.Fingerprint()), receivers)
+		alert := AlertToOpenAPIAlert(a, api.getAlertStatus(a.Fingerprint()), receivers, nil)
 
 		res = append(res, alert)
 	}
@@ -415,7 +415,7 @@ func (api *API) getAlertGroupsHandler(params alertgroup_ops.GetAlertGroupsParams
 			fp := alert.Fingerprint()
 			receivers := allReceivers[fp]
 			status := api.getAlertStatus(fp)
-			apiAlert := AlertToOpenAPIAlert(alert, status, receivers)
+			apiAlert := AlertToOpenAPIAlert(alert, status, receivers, alertGroup)
 			ag.Alerts = append(ag.Alerts, apiAlert)
 		}
 		res = append(res, ag)
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -899,7 +899,11 @@ func (tms TimeMuteStage) Exec(ctx context.Context, l log.Logger, alerts ...*type
 	if err != nil {
 		return ctx, alerts, err
 	}
-
+	//for _, a := range alerts {
+	// If the alert is not muted then mutedBy is a nil slice. This will set the
+	// alerts back to active.
+	//tms.marker.SetMuted(a.Fingerprint(), mutedBy...)
+	//}
 	// If the current time is inside a mute time, all alerts are removed from the pipeline.
 	if muted {
 		level.Debug(l).Log("msg", "Notifications not sent, route is within mute time")
@@ -944,4 +948,21 @@ func (tas TimeActiveStage) Exec(ctx context.Context, l log.Logger, alerts ...*ty
 	}
 
 	return ctx, alerts, nil
+}
+
+// inTimeIntervals returns true if the current time is contained in one of the given time intervals.
+func inTimeIntervals(now time.Time, intervals map[string][]timeinterval.TimeInterval, intervalNames []string) (bool, []string, error) {
+	var in []string
+	for _, name := range intervalNames {
+		interval, ok := intervals[name]
+		if !ok {
+			return false, nil, errors.Errorf("time interval %s doesn't exist in config", name)
+		}
+		for _, ti := range interval {
+			if ti.ContainsTime(now.UTC()) {
+				in = append(in, name)
+			}
+		}
+	}
+	return len(in) > 0, in, nil
 }
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -947,20 +947,3 @@ func (tas TimeActiveStage) Exec(ctx context.Context, l log.Logger, alerts ...*ty
 
 	return ctx, alerts, nil
 }
-
-// inTimeIntervals returns true if the current time is contained in one of the given time intervals.
-func inTimeIntervals(now time.Time, intervals map[string][]timeinterval.TimeInterval, intervalNames []string) (bool, []string, error) {
-	var in []string
-	for _, name := range intervalNames {
-		interval, ok := intervals[name]
-		if !ok {
-			return false, nil, errors.Errorf("time interval %s doesn't exist in config", name)
-		}
-		for _, ti := range interval {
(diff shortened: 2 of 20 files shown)


grobinson-grafana:
Should I also update apiv1? I'm not sure if that's in maintenance mode or still receiving features?

notify/notify.go
```diff
@@ -375,6 +375,7 @@ func (pb *PipelineBuilder) New(
 	inhibitor *inhibit.Inhibitor,
 	silencer *silence.Silencer,
 	times map[string][]timeinterval.TimeInterval,
+	marker types.Marker,
```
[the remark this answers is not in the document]
gotjosh: I initially tried to suggest the changes I thought were necessary but instead decided it'll be simpler to just do it - I've created the abstraction here: https://github.com/prometheus/alertmanager/pull/3556

This leaves the functionality intact so that you can then rebase on top of that to add the marker usage directly on the `intervener` (for the lack of a better name).

gotjosh (commented):
Implementation-wise, this LGTM - I skimmed through the API changes and they look fine but I decided to put an emphasis in what we're doing as part of the notifier package.

Let's try to address that first.

types/types.go
```diff
@@ -53,22 +54,30 @@ type AlertStatus struct {
...
 	// AlertStateActive. Otherwise, it sets the provided alert to
 	// AlertStateSuppressed.
 	SetInhibited(alert model.Fingerprint, alertIDs ...string)
+	// SetMuted replaces the previous MutedBy by the provided time intervals.
+	// Unlike SetInhibited, the set of provided time intervals is expected
+	// to represent the complete set of relevant time intervals. If no time
+	// intervals are provided and both InhibitedBy and SilencedBy are empty,
+	// it sets the provided alert to AlertStateActive. Otherwise, it sets the
+	// provided alert to AlertStateSuppressed.
+	SetMuted(alert model.Fingerprint, alertIDs ...string)
```
gotjosh, line 80:
```suggestion
	// SetMuted replaces the previous MutedBy by the provided time intervals.
	// Similarly to SetInhibited and in contrast to SetActiveOrSilenced, the set of provided time intervals is expected
	// to not represent the complete set of relevant time intervals. If no time
	// intervals are provided and both InhibitedBy and SilencedBy are empty,
	// it sets the provided alert to AlertStateActive. Otherwise, it sets the
	// provided alert to AlertStateSuppressed.
	SetMuted(alert model.Fingerprint, alertIDs ...string)
```

api/v2/api.go
```diff
@@ -424,7 +424,7 @@ func (api *API) getAlertGroupsHandler(params alertgroup_ops.GetAlertGroupsParams
 	return alertgroup_ops.NewGetAlertGroupsOK().WithPayload(res)
 }
 
-func (api *API) alertFilter(matchers []*labels.Matcher, silenced, inhibited, active bool) func(a *types.Alert, now time.Time) bool {
+func (api *API) alertFilter(matchers []*labels.Matcher, silenced, inhibited, muted, active bool) func(a *types.Alert, now time.Time) bool {
```
gotjosh, line 427: nit: You've got it right on the places that this used but I feel that is a lot less risky if you add the new parameter at the end as opposed to before the last one.

types/types.go
gotjosh: My understanding of both `SetMuted` and `SetInhibited` is that if you have an alert that was previously muted by a rule or interval if that alert gets muted again it'll override the previous one - this can't happen in `SetActiveOrSilenced` because we'll query that's affecting.

notify/notify.go
```diff
@@ -889,11 +891,15 @@ func (tms TimeMuteStage) Exec(ctx context.Context, l log.Logger, alerts ...*type
...
+	muted, mutedBy, err := inTimeIntervals(now, tms.Times, muteTimeIntervalNames)
 	if err != nil {
 		return ctx, alerts, err
 	}
-
+	for _, a := range alerts {
+		// If the alert is not muted then mutedBy is a nil slice. This will set the
+		// alerts back to active.
+		tms.marker.SetMuted(a.Fingerprint(), mutedBy...)
+	}
```
gotjosh, line 902: with #3556 now merged we can remove all of the changes done to `notify.go` - we'll have to change the interface slightly because it now needs to accept `a.Fingerprint()` but that's ok given it is only used here.

gotjosh (commented):
We're almost there!

notify/notify.go
grobinson-grafana: I'm not sure if `a.Fingerprint()` is sufficient as time mutes are applied against the route, not the alert. We can pass a list of fingerprints, or a list of alerts. I started doing this:

```go
type TimeMuter interface {
	Mutes(timeIntervalName []string, now time.Time, alerts ...Alert) (bool, error)
}
```

types/types.go
grobinson-grafana: Ah no, that's not how it works in the PR. I'm recording the names of **all** time intervals that muted it, not just the first time interval. I understand that silences work the same?

notify/notify.go
gotjosh: Yeah, either can work! I chose a single one, because at least it's consistent with the other similar interface https://github.com/prometheus/alertmanager/blob/acb58400fd2b324dd4cc56110c8e891e8b1c0420/types/types.go#L380-L382

However, I just realised that what we want is `a.Labels` as opposed to `a.Fingerprint()` so that would be my suggestion.

types/types.go
grobinson-grafana: Ah. But if there the alert matches two routes, and both those routes have different mute time intervals, just one of them will be recorded. :frowning_face:

gotjosh:
I chatted with George offline for a bit, and it turns out this doesn't do what he thought it did. It suffers from a "race" where an alert from one route can override the status from the same alert going through another way.

Because time intervals are at a route level, the filtering must show on which route was the time interval active so that underlying data structure needs to change to accommodate that requirement.

api/v2/api.go
grobinson-grafana: I can change it 🙂 It sounds like we don't have sufficient test coverage to have confidence in this?

grobinson-grafana:
I've decided to make this a draft for the time being as there is a lot of work still to do and there are still some decisions to make about how to best add this feature.

types/types.go
```diff
@@ -381,9 +424,18 @@ type Muter interface {
 	Mutes(model.LabelSet) bool
 }
 
-// TimeMuter determines if alerts should be muted based on the specified current time and active time interval on the route.
+// TimeMuter determines if alerts should be muted or not for active and mute
+// time intervals based on the specified current time.
 type TimeMuter interface {
```
grobinson-grafana, line 429: This is where I made changes to the interface based on the discussion in #3562. I'm not super happy with how this interface has turned out though as there are now two methods `Active` and `Mutes` which look like inverses of each other when they are not.

For example, it looks like if I pass a series of mute time intervals to `Active` and it returns false, then  passing the same mute time intervals to `Mute` should return true. That is, `Active` and `Mute` are negations of each other. However, in this case both will instead return true, as `Active` is only supposed to be used with active time intervals and `Mute` is only supposed to be used with mute time intervals. I have attempted to document their behavior as such, but I'm still not happy with it.

timeinterval/timeinterval.go
```diff
@@ -25,34 +25,60 @@ import (
...
 	intervals map[string][]TimeInterval
+	marker    types.Marker
+}
+
+func (i *Intervener) Active(groupKey string, alerts []*types.Alert, names []string, now time.Time) (bool, error) {
+	matches, err := i.inIntervals(names, now)
+	if err != nil {
+		return false, err
+	}
+	for _, a := range alerts {
```
grobinson-grafana, line 44: The marking here is incorrect and will need to be fixed if we agree to continue having both `Active` and `Mute`.

grobinson-grafana:
I'm closing this to start over, as it's been almost 6 months since we last worked on it. There are some conflicts with main that I need to fix, and I am considering doing something a little different this time.

PR #3529, opened 2023-09-16
Set spellcheck=false attribute on silence filter <input>.

dsymonds:
This hints to iOS Safari and other browsers that they shouldn't use spellcheck, but, more importantly, shouldn't use smart quotes (like “this” instead of "this"). This matters here because Alertmanager requires quotes for a filter but won't accept smart quotes, making it tedious since the default input keyboard will normally use them automatically.


dsymonds:
I know I'm supposed to include an updated `asset/*_vfsdata.go` in the PR, but I can't seem to get all the Elm dev stuff working on my machine, so would be happy for someone else to take over this PR and complete it. I've tested it manually.

SoloJacobs:
I like this change, and have done the follow-up you requested here: https://github.com/prometheus/alertmanager/pull/4811

dsymonds:
Thanks @SoloJacobs!

PR #3530, opened 2023-09-18
Add s390x Docker image

Schmashlafu:
To be honest, I have no idea how circleci works.
I hope this does it.
Everything else seems to already build for s390x.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -127,7 +127,7 @@ workflows:
       - prometheus/build:
           name: build
           parallelism: 3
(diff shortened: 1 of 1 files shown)


jan--f:
alermanager also publishes s390x images looking at e.g. the [prometheus quay org](https://quay.io/repository/prometheus/alertmanager/manifest/sha256:11c3c3e1275e0a8d1fc4600d9480c08f98fe7f4141a62d9808a79513024b66de).
This change would add an s390x build to the PR pre-merge check iiuc.

Schmashlafu:
Ah, thanks for the pointer.
I didn't check Quay.

In that case my PR is redundant anyway.

PR #3532, opened 2023-09-21
Fix Github Actions publish

gotjosh:


jan--f:
Since we don't use the promci build action we need to call the save_artifacts action. I think after [thsi step](https://github.com/prometheus/alertmanager/blob/b517645d951da230fbb078da417c5ac46928dbd6/.github/workflows/ci.yml#L30) should be the right place.

Apologies, I thought I had added that, I might have rebased it away at some point.

jan--f:
@gotjosh see #3533

PR #3533, opened 2023-09-22
actions: cross build in ci and fix publish

jan--f:
This add

- a job to build on 3 architectures in the CI action
- Fixes the publish action to cross build all architectures first

~TODO: add a release action and discuss and add appropriate trigger filters for release and publish.~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,21 @@ jobs:
       - run: make apiv2
       - run: git diff --exit-code
 
+  build:
+    name: Build Alertmanager for common architectures
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        thread: [ 0, 1, 2 ]
+    steps:
+      - uses: actions/checkout@3df4ab11eba7bda6032a0b82a6bb43b11571feac # v4.0.0
+      - uses: prometheus/promci@3cb0c3871f223bd5ce1226995bd52ffb314798b6 # v0.1.0
+      - uses: ./.github/promci/actions/build
+        with:
+          promu_opts: "-p linux/amd64 -p windows/amd64 -p linux/arm64 -p darwin/amd64 -p darwin/arm64 -p linux/386"
+          parallelism: 3
+          thread: ${{ matrix.thread }}
+
   test:
     name: Test
     runs-on: ubuntu-latest
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -9,13 +9,27 @@ jobs:
     name: Run ci
     uses: ./.github/workflows/ci.yml
 
-  publish_main:
-    name: Publish main branch artefacts
+  build:
+    name: Build Alertmanager for all architectures
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        thread: [ 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11 ]
     needs: ci
     steps:
       - uses: actions/checkout@v3
-      - uses: prometheus/promci@v0.0.2
+      - uses: prometheus/promci@3cb0c3871f223bd5ce1226995bd52ffb314798b6 # v0.1.0
+      - uses: ./.github/promci/actions/build
+        with:
+          parallelism: 12
+          thread: ${{ matrix.thread }}
+  publish_main:
+    name: Publish main branch artefacts
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - uses: actions/checkout@v3
+      - uses: prometheus/promci@3cb0c3871f223bd5ce1226995bd52ffb314798b6 # v0.1.0
       - uses: ./.github/promci/actions/publish_main
         with:
           docker_hub_login: ${{ secrets.docker_hub_login }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,17 +9,30 @@ jobs:
     name: Run ci
     uses: ./.github/workflows/ci.yml
 
-  publish_release:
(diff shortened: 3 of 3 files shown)


.github/workflows/publish.yml
```diff
@@ -2,23 +2,38 @@
...
+          thread: ${{ matrix.thread }}
   publish_main:
     name: Publish main branch artefacts
     runs-on: ubuntu-latest
-    needs: ci
+    needs: build
     steps:
       - uses: actions/checkout@v3
-      - uses: prometheus/promci@v0.0.2
+      - uses: prometheus/promci@3cb0c3871f223bd5ce1226995bd52ffb314798b6 # v0.1.0
```
gotjosh, line 32: Why are we downgrading here? That's wrong with the latest version?

.github/workflows/publish.yml
```diff
@@ -2,23 +2,38 @@
...
       - uses: ./.github/promci/actions/publish_main
         with:
           docker_hub_login: ${{ secrets.docker_hub_login }}
           docker_hub_password: ${{ secrets.docker_hub_password }}
+          quay_io_organization: ${{ secrets.quay_io_organization }}
           quay_io_login: ${{ secrets.quay_io_login }}
           quay_io_password: ${{ secrets.quay_io_password }}
```
gotjosh, line 39: None of these secrets are defined. At which point would be ready to test them? How did you approach this in Prometheus?

gotjosh:
Looking at the current failures, I also see:

> The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

.github/workflows/publish.yml
jan--f: This is actually an upgrade from `v0.0.2` to `v0.1.0` :) The pinning to a sha I copied from Prometheus, seems like good practice and I intend to do that for the other instances too.

.github/workflows/publish.yml
jan--f: Yeah I'm looking into that, not sure myself at this point. In my clone I defined those secrets but it still fails, there is something I don't understand with multi arch builds.

[If those secrets don't exist the publish action is just a no-op](https://github.com/prometheus/promci/blob/3cb0c3871f223bd5ce1226995bd52ffb314798b6/actions/publish_main/action.yml#L31-L38). so we can leave this for now and once we understand what's going on, defining the secrets is all that is needed to publish from the action.

gotjosh: Great to know!

.github/workflows/publish.yml
gotjosh: 🤦 I need to learn to read a bit better - agreed, pinning of versions saves us from unexpected failures.

.github/workflows/publish.yml
roidelapluie: they should be defined at the prometheus org level

.github/workflows/publish.yml
```diff
@@ -2,23 +2,38 @@
 name: Publish
 on:  # yamllint disable-line rule:truthy
   push:
-    branches:
-      - main
+    # branches:
+    #   - main
```
roidelapluie, line 6: why?

.github/workflows/ci.yml
```diff
@@ -16,6 +16,21 @@ jobs:
       - run: make apiv2
       - run: git diff --exit-code
 
+  build:
+    name: Build Alertmanager for common architectures
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        thread: [ 0, 1, 2 ]
```
roidelapluie, line 24: This complexity is not needed for alertmanager

.github/workflows/publish.yml
jan--f: This is just temporary to tigger the action on a PR push. I'll drop all `DEBUG` commits before this merges.

.github/workflows/ci.yml
jan--f: I took this from the[ circleci build job](https://github.com/prometheus/alertmanager/blob/c6be0bcabf37e837d24786e3d5c10565ac499531/.circleci/config.yml#L127-L135) thinking its a reasonable check before merge. Happy to drop if it isn't needed.

jan--f: Also https://github.com/prometheus/alertmanager/pull/3530 wants to extend this.

.github/workflows/publish.yml
```diff
@@ -2,23 +2,38 @@
...
+    needs: build
     steps:
       - uses: actions/checkout@v3
-      - uses: prometheus/promci@v0.0.2
+      - uses: prometheus/promci@3cb0c3871f223bd5ce1226995bd52ffb314798b6 # v0.1.0
       - uses: ./.github/promci/actions/publish_main
         with:
           docker_hub_login: ${{ secrets.docker_hub_login }}
           docker_hub_password: ${{ secrets.docker_hub_password }}
+          quay_io_organization: ${{ secrets.quay_io_organization }}
```
gotjosh, line 37: @roidelapluie is right these all are defined at an org-level except this one - why do we need this one again?

jan--f: Again a debug commit to change the org to my quay org.

jan--f:
@gotjosh sorry, let me drop the debug commits to reduce noise.

jan--f:
Ok I added the build step before the release action as well. 
@roidelapluie @gotjosh Is there anything else needed here?

gotjosh (approved):

PR #3534, opened 2023-09-23
Add support to enforce label on alert

alileza:
# Problem Statement
In an organization where we manage a lot of different services, we all like to share common alerts, but the ownership of the alert might be a different team since the alert label is quite powerful, so I thought it would be a pretty nice feature to have to be able to set a simple rule to automatically add label whenever alert match certain label.

This use case is useful for alerts like OOMKilled, in which we want to be able to route the alert to every team, that is responsible for taking care of that, but we don't want to have duplicated alerts. With these enforcement label rules, it will automatically labelled by the rules being set by the orgs

```
enforce_label_rules:
  - key: name
    match: argocd
    then:
      team: world
      something: else
  ```
  
 this problem was resolved from https://github.com/alileza/alert-relabeller, that repo feels more like a hack, and I think it would be nice to be part of alert manager like this, I wonder what do you think? I put it as draft since I'm not sure if this kind of code would be able to get merged.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1397,7 +1397,7 @@ The fields are documented in the [Rocketchat API api models](https://github.com/
 [ msg: <tmpl_string> ]
 ```
 
-#### `<slack_config>`
+### `<slack_config>`
 
 Slack notifications can be sent via [Incoming webhooks](https://api.slack.com/messaging/webhooks) or [Bot tokens](https://api.slack.com/authentication/token-types).
 


alileza:
Actually, I just realized that there seems to be a big flaw in this implementation since it enforces the label on the alert manager, but the actual `ALERTS` metrics weren't being patched, which is what I need for the analysis

roidelapluie:
You will not be able to modify the ALERTS metrics. But you can achieve what you do here with alerts_relabel_configs, which will still not change ALERTS but will add the labels before sending to alertmanager.

alileza:
`alerts_relabel_configs`? do you mind giving pointer where see this specific configuration, I can't seem to find `alerts_relabel_configs` in documentation, or in the codebase

grobinson-grafana:
Hi! That's because it is a Prometheus feature, rather than Alertmanager. It's recommended to relabel alerts before being sent to the Alertmanager. You can find it [here](https://prometheus.io/docs/prometheus/latest/configuration/configuration/#alert_relabel_configs).

PR #3535, opened 2023-09-23
Remove unused function GetAlertmanagerURL

grobinson-grafana:
As discussed with @gotjosh [here](https://github.com/prometheus/alertmanager/pull/3483#discussion_r1334210090), we agreed to remove this function as it's un-used.


gotjosh (approved):
LGTM

PR #3538, opened 2023-09-29
Fix child routes not inheriting time interval configuration

### LINKED ISSUE #2833: mute_time_intervals should be inherited to sub routes
**What did you do?**
I set up a route to mute all warning alert outside work days, and several sub routes to route these alerts to different receivers.

I was expecting this alert could be silenced in all sub routes

the alert was sent regardless of the mute intervel setting
**Environment**

* System information:

* Alertmanager version:

	insert output of `alertmanager --version` here (repeat for each alertmanager
	version in your cluster, if relevant to the issue)
v0.23.0
* Prometheus version:

	insert output of `prometheus --version` here (repeat for each prometheus
	version in your cluster, if relevant to the issue)
```
insert configuration here
```

```
insert configuration here (if relevant to the issue)
```

* Logs:
```
insert Prometheus and Alertmanager logs relevant to the issue here
```

I'm aware the document said:

```
# Times when the route should be muted. These must match the name of a
# mute time interval defined in the mute_time_intervals section. 
# Additionally, the root node cannot have any mute times.
# When a route is muted it will not send any notifications, but
# otherwise acts normally (including ending the route-matching process
# if the `continue` option is not set.)
mute_time_intervals:
  [ - [<string>](https://www.prometheus.io/docs/alerting/latest/configuration/#string) ...]
```

but I think these rules should be inherited, this can make the config shorter and easier to read, otherwise I have to set mute time for every single route.

benridley:
Fixes #2833.

Fixes a bug in the creation of routes which would mistakenly override parent route config with empty time interval configuration if none were specified on the child.


juliusv:
Thanks! I haven't checked in detail, but @simonpasquier mentioned in https://github.com/prometheus/alertmanager/issues/2833#issuecomment-1698824341 that overriding time intervals in children could lead to issues because empty time intervals are also valid. Would setting an explicit empty list of intervals (vs. omitting the field entirely) in a child allow overriding the time intervals of a parent with this PR?

benridley:
> Thanks! I haven't checked in detail, but @simonpasquier mentioned in [#2833 (comment)](https://github.com/prometheus/alertmanager/issues/2833#issuecomment-1698824341) that overriding time intervals in children could lead to issues because empty time intervals are also valid. Would setting an explicit empty list of intervals (vs. omitting the field entirely) in a child allow overriding the time intervals of a parent with this PR?

Maybe I'm misunderstanding the implications of this, but I think this shouldn't have any impact. Previously, every new route would receive an explicit empty list of intervals (through the `DefaultRouteOpts` struct) regardless of whether the parent had a set of time intervals or not. Now, child route will receive an empty list, or the parent's list *if its not empty*.

So empty time intervals on the parent will result in the same behaviour before this change. The change will be if the parent has defined intervals which will now be inherited.

costela:
Friendly ping: the review seems to have fallen through the cracks? :innocent:

SoloJacobs:
Julius is correct there: Empty list `[]` is indeed overridden by the parent, despite the docs promising, that this mute nothing.

SoloJacobs:
We talked about this at Alertmanager Bug Scrub, decided that this PR cannot be merged, but the original [issue](https://github.com/prometheus/alertmanager/issues/2833) should remain open. The reason is that changing the meaning of the existing fields will silently mute a lot of alerts. Having a migration path for existing configuration is thus a prerequisite, something we will work towards in the future.

Sorry, to the users, which are still waiting for this change. It will have to wait for even longer.

Kind regards, Solomon Jacobs

PR #3547, opened 2023-10-05
Accept RocketChat success code

### LINKED ISSUE #3546: #3121 breaks Slack integration targeting RocketChat servers
**What did you do?**

We use the Slack integration to send notifications to a RocketChat server, which is reasonably compatible with Slack for this purpose. This has worked well for 4+ years, however after our last update we keep getting an ever increasing list of "Resolved" alerts every 5 minutes.

We expect to see an alert posted, then a resolved posted, then silence.

We see the alert posted, then the resolved posted over and over again (every 5 minutes) until the AlertManager instance is restarted.

**Environment**

We believe that change #3121 triggered this condition. That change checks a 200 status code to ensure if contains `"ok": true` (previously 200 meant OK without further checking).

Our RocketChat instances returns a different message on success:

```bash
$ curl ...
< HTTP/1.1 200 OK
< Content-Type: application/json
...
{"success":true}
```

We believe that this new code identifies this (incorrectly) as failed POST and thus doesn't remove the now resolved alert from its list, and dutifully tries again every 5 minutes for eternity.

(Credit to @bg-govau for pin-pointing the issue. PR forthcoming.)

ae-govau:
Fixes #3546

### DIFF
--- a/notify/slack/slack.go
+++ b/notify/slack/slack.go
@@ -261,13 +261,15 @@ func checkJSONResponseError(body []byte) (bool, error) {
 	type response struct {
 		OK    bool   `json:"ok"`
 		Error string `json:"error"`
+
+		Success bool `json:"success"` // RocketChat, which is mostly Slack compatible, returns {"success": true}
 	}
 
 	var data response
 	if err := json.Unmarshal(body, &data); err != nil {
 		return true, errors.Wrapf(err, "could not unmarshal JSON response %q", string(body))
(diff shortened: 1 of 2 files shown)


roidelapluie:
This should be fixed in rocketchat, not alertmanager

gotjosh:
Let's keep the conversation focused on the issue - we can always reopen this PR with what we decide.

PR #3324, opened 2023-04-12
Add msteams

zhan9san:
Related to #3076

As the `new line` is important to Markdown format, a `nested template definition` `__text_alert_list_markdown` is added.

For simplicity, only `text` field is used to send message in [Message card](https://learn.microsoft.com/en-us/outlook/actionable-messages/message-card-reference).

The message format is similar to Discord's.

Card design can be checked in [Card Playground](https://messagecardplayground.azurewebsites.net/) with following json text.

```json
{
	"@type": "MessageCard",
	"@context": "https://schema.org/extensions",
	"themeColor": "8C1A1A",
	"title": "[FIRING:1] InstanceDown (v critical core)",
	"text": "\n\nAlerts Firing:\n\nLabels:\n - alertname = InstanceDown\n - env = build-service\n - environment = v\n - instance = https://example.com\n - job = blackbox\n - severity = critical\n - type = core\n\nAnnotations:\n - description = https://example.com of job blackbox has been down for more than 5 minutes.\n - summary = Instance https://example.com down\n\nSource: http://v:9090/graph?g0.expr=probe_success%7Btype%3D%22core%22%7D+%3D%3D+0&g0.tab=1\n"
}
```

PR #3550, opened 2023-10-09
Fix Docs: default for msteams config

luka5:
now fits to https://github.com/prometheus/alertmanager/blob/v0.26.0/config/notifiers.go#L168-L175

similar to https://github.com/prometheus/alertmanager/pull/3495; introduced in https://github.com/prometheus/alertmanager/pull/3324


gotjosh (approved):
LGTM

PR #3552, opened 2023-10-09
Update configuration docs for Repeat interval

grobinson-grafana:
Although it is true that Repeat interval should be greater than or equal to the Group interval, it should also be a multiple of it too. If the Repeat interval is not a multiple, then because of how aggregation groups are flushed, it will be made into one implicitly. This commit documents this behavior.


gotjosh (approved):

gotjosh:

PR #3556, opened 2023-10-12
Refactor: Move `inTimeIntervals` from `notify` to `timeinterval`

gotjosh:
This should make implementing #3520 a bit cleaner given that we shouldn't need to introduce a direct dependency on marker in `notify`.

There's absolutely no change of functionality here and I've expanded coverage for similar logic in both places.

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -478,6 +478,8 @@ func run() int {
 			timeIntervals[ti.Name] = ti.TimeIntervals
 		}
 
+		intervener := timeinterval.NewIntervener(timeIntervals)
+
 		inhibitor.Stop()
 		disp.Stop()
 
@@ -497,7 +499,7 @@ func run() int {
 			waitFunc,
 			inhibitor,
 			silencer,
-			timeIntervals,
+			intervener,
 			notificationLog,
 			pipelinePeer,
 		)
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -894,7 +894,6 @@ func (tms TimeMuteStage) Exec(ctx context.Context, l log.Logger, alerts ...*type
 		return ctx, alerts, nil
 	}
 
-	//muted, err := inTimeIntervals(now, tms.Times, muteTimeIntervalNames)
 	muted, err := tms.muter.Mutes(muteTimeIntervalNames, now)
 	if err != nil {
 		return ctx, alerts, err
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -869,7 +869,6 @@ func (n SetNotifiesStage) Exec(ctx context.Context, l log.Logger, alerts ...*typ
 
 type timeStage struct {
 	muter types.TimeMuter
-	//Times map[string][]timeinterval.TimeInterval
 }
 
 type TimeMuteStage timeStage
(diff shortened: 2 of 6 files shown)


cmd/alertmanager/main.go
```diff
@@ -478,6 +478,8 @@ func run() int {
 			timeIntervals[ti.Name] = ti.TimeIntervals
 		}
 
+		intervener := timeinterval.NewIntervener(timeIntervals)
```
gotjosh, line 481: I started moving the loop above into `timeinterval` as well but quickly realised that we have an import problem - this is part of a larger problem and I'll address it in a separate PR.

I don't understand why the marshalling of config -> struct happens with structs that live in `timeinterval` when this is generally done as part of the`config` package.

notify/notify.go
```diff
@@ -889,7 +889,13 @@ func (tms TimeMuteStage) Exec(ctx context.Context, l log.Logger, alerts ...*type
 		return ctx, alerts, errors.New("missing now timestamp")
 	}
 
-	muted, err := inTimeIntervals(now, tms.Times, muteTimeIntervalNames)
+	// Skip this stage if there are no mute timings.
+	if len(muteTimeIntervalNames) == 0 {
+		return ctx, alerts, nil
+	}
```
gotjosh, line 895: This guard clause was missing from this path.

timeinterval/timeinterval_test.go
```diff
@@ -659,3 +660,97 @@ func mustLoadLocation(name string) *time.Location {
 	}
 	return loc
 }
+
+func TestIntervener_Mutes(t *testing.T) {
```
gotjosh, line 664: These tests are technically a duplicate of what we already have in `notify` - I think they have some merit in being here as there might be more we want to test as part of this interface in the future (e.g. when we introduce the marker) and it gave me more confidence in the refactor.

timeinterval/timeinterval_test.go
```diff
@@ -659,3 +660,97 @@ func mustLoadLocation(name string) *time.Location {
...
+			expected: false,
+		},
+		{
+			name:     "Should not mute on a Tuesday before 5pm",
+			firedAt:  "16 Nov 21 16:59 +1100",
+			expected: false,
+		},
+		{
+			name:     "Should mute on a Saturday",
+			firedAt:  "20 Nov 21 10:00 +1100",
```
grobinson-grafana, line 709: I expected this not to be muted? :confused:

timeinterval/timeinterval_test.go
```diff
@@ -659,3 +660,97 @@ func mustLoadLocation(name string) *time.Location {
...
+    - monday:friday
+  location: Australia/Sydney
+  months:
+    - November
+  times:
+    - start_time: 00:00
+      end_time: 09:00
+    - start_time: 17:00
+      end_time: 24:00
+- weekdays:
```
grobinson-grafana, line 678: Looking at the docs (https://prometheus.io/docs/alerting/latest/configuration/#time_interval), I expected this to mute on Saturdays and Sundays between 00:00 and 09:00, and between 17:00 and 24:00. Is that correct?

grobinson-grafana (commented):
LGTM!

grobinson-grafana (approved):

PR #3438, opened 2023-08-03
Log the GroupKey and alerts in retry

grobinson-grafana:
### What this PR does

This pull request updates `notify.go` to log the GroupKey and fingerprints of an alert at the debug level, and just the GroupKey at the warning level should the notify attempt fail.

Here is an example of it at debug level:

```
ts=2023-08-03T09:28:36.829Z caller=dispatch.go:163 level=debug component=dispatcher msg="Received alert" alert=foo[c5a7cb1][active]
ts=2023-08-03T09:28:38.725Z caller=dispatch.go:163 level=debug component=dispatcher msg="Received alert" alert=bar[4656093][active]
ts=2023-08-03T09:28:51.830Z caller=dispatch.go:515 level=debug component=dispatcher aggrGroup="{}:{alertname=\"foo\"}" msg=flushing alerts=[foo[c5a7cb1][active]]
ts=2023-08-03T09:28:51.984Z caller=notify.go:757 level=debug component=dispatcher receiver=email integration=email[0] aggrGroup="{}:{alertname=\"foo\"}" alerts=[foo[c5a7cb1][active]] msg="Notify success" attempts=1
ts=2023-08-03T09:28:53.726Z caller=dispatch.go:515 level=debug component=dispatcher aggrGroup="{}:{alertname=\"bar\"}" msg=flushing alerts=[bar[4656093][active]]
ts=2023-08-03T09:28:53.855Z caller=notify.go:757 level=debug component=dispatcher receiver=email integration=email[0] aggrGroup="{}:{alertname=\"bar\"}" alerts=[bar[4656093][active]] msg="Notify success" attempts=1
```

and when notify attempt fails:

```
ts=2023-08-03T09:34:42.505Z caller=notify.go:745 level=warn component=dispatcher receiver=email integration=email[0] aggrGroup="{}:{alertname=\"foo\"}" msg="Notify attempt failed, will retry later" attempts=1 err="establish connection to server: dial tcp 127.0.0.1:1025: connect: connection refused"
```

### Motivation

Without this change it's almost impossible to correlate a `Notify success` log line with the aggregation group's flush log line, and vice versa. This change will make it easier to debug issues around missing notifications or unexpected notifications for alerts when running large Alertmanager installations with 10,000s of alerts.

For example:

```
ts=2023-08-03T09:45:57.934Z caller=dispatch.go:515 level=debug component=dispatcher aggrGroup="{}:{alertname=\"foo\"}" msg=flushing alerts=[foo[c5a7cb1][active]]
[...]

PR #3558, opened 2023-10-15
Add debug logs for muted alerts

grobinson-grafana:
This commit adds debug logs to MuteStage that logs when an alert is muted. This can help operators root cause missing notifications when alerts are silenced by mistake or purpose but then forgotten about.

```
ts=2023-10-15T21:10:47.858Z caller=notify.go:537 level=debug component=dispatcher msg="Notifications will not be sent for muted alerts" alerts="[[26cdc50][active] [d7f5862][active]]"
```

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -518,15 +518,23 @@ func NewMuteStage(m types.Muter) *MuteStage {
 }
 
 // Exec implements the Stage interface.
-func (n *MuteStage) Exec(ctx context.Context, _ log.Logger, alerts ...*types.Alert) (context.Context, []*types.Alert, error) {
-	var filtered []*types.Alert
+func (n *MuteStage) Exec(ctx context.Context, logger log.Logger, alerts ...*types.Alert) (context.Context, []*types.Alert, error) {
+	var (
+		filtered []*types.Alert
+		muted    []*types.Alert
+	)
 	for _, a := range alerts {
 		// TODO(fabxc): increment total alerts counter.
 		// Do not send the alert if muted.
-		if !n.muter.Mutes(a.Labels) {
+		if n.muter.Mutes(a.Labels) {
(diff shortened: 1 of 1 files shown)


gotjosh (approved):
LGTM

notify/notify.go
```diff
@@ -518,16 +518,24 @@ func NewMuteStage(m types.Muter) *MuteStage {
...
-		if !n.muter.Mutes(a.Labels) {
+		if n.muter.Mutes(a.Labels) {
+			muted = append(muted, a)
+		} else {
 			filtered = append(filtered, a)
 		}
 		// TODO(fabxc): increment muted alerts counter if muted.
 	}
+	if len(muted) > 0 {
+		level.Debug(logger).Log("msg", "Notifications will not be sent for muted alerts", "alerts", fmt.Sprintf("%v", muted))
```
gotjosh, line 537: Sorry, I can't dismiss my approval, but this needs a change.

I think we should truncate the alert list to a maximum - I could have swore there are others places where we do this but I couldn't find them.

grobinson-grafana: Can you share what your main concerns are?

This would be the first place where we truncate the alert slice when logging in debug mode. The main place this happens untruncated is [here](https://github.com/prometheus/alertmanager/blob/280a742b766d0b4079b02be100cf9541e29f0a20/dispatch/dispatch.go#L515) and [here](https://github.com/prometheus/alertmanager/pull/3438).

If we do truncate it for silences, we should also truncate it elsewhere. However if we do that then debug logs become rather useless for debugging missing notifications as there is no other record of which alerts are in which aggregation groups.

gotjosh: You're right - I'm probably misremembering. 

My main concern was debug being too spammy but I this is a very valid use case.

PR #3561, opened 2023-10-17
fix: Bump golang.org/x/net to v0.17.0

slashpai:
cc: @simonpasquier 

Also wondering if we should consider a go bump here its on 1.18 currently


simonpasquier (approved):
Thanks!

PR #3562, opened 2023-10-17
Replace Intervener with ActiveIntervener and MuteIntervener

grobinson-grafana:
This commit replaces the Intervener in timeinterval package with an ActiveIntervener and a MuteIntervener for active time intervals and mute time intervals. It avoids having a double-negative in TimeActiveStage where Mute returns true if its not muted rather than if it is muted, and will also enable us to mark alerts as muted or not within ActiveIntervener and MuteIntervener.

### DIFF
--- a/timeinterval/timeinterval.go
+++ b/timeinterval/timeinterval.go
@@ -27,31 +27,56 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-// Intervener determines whether a given time and active route time interval should mute outgoing notifications.
+// MuteIntervener determines whether the mute time interval should mute outgoing notifications.
 // It implements the TimeMuter interface.
-type Intervener struct {
+type MuteIntervener struct {
 	intervals map[string][]TimeInterval
 }
 
-func (i *Intervener) Mutes(names []string, now time.Time) (bool, error) {
+func (i *MuteIntervener) Mutes(names []string, now time.Time) (bool, error) {
 	for _, name := range names {
 		interval, ok := i.intervals[name]
 		if !ok {
-			return false, fmt.Errorf("time interval %s doesn't exist in config", name)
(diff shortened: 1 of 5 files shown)


timeinterval/timeinterval.go
```diff
@@ -27,31 +27,56 @@ import (
...
+	}
+}
+
+// ActiveIntervener determines whether the active time interval should mute outgoing notifications.
+// It implements the TimeMuter interface.
+type ActiveIntervener struct {
+	intervals map[string][]TimeInterval
+}
+
+func (i *ActiveIntervener) Mutes(names []string, now time.Time) (bool, error) {
```
gotjosh, line 63: I'm not a fan of this - this is pretty much the same logic as the other but with a flipped conditional. I'd rather keep it as is and continue to have the `!muted` on the stage itself.

This seems like a lot of code to change/add for the sake of a flipping a conditional.

grobinson-grafana: That's OK. We'll then need to put the marker back into `TimeMuteStage` and `TimeActiveStage` though as otherwise `Intervener` doesn't know if it's activing or muting.

gotjosh: Maybe we should introduce it then? It's hard for me to see the benefit of this as is. 😭

grobinson-grafana: I do think making time active intervals adhere to the contract of `Mute` is a good win in of itself. Mute no longer returns false when it should be muted, and true when it shouldn't be muted.

grobinson-grafana: That still doesn't work I'm afraid, because `Intervener` doesn't know if the intervals in `names` are active time intervals or mute time intervals.

Issue #3512, opened 2023-09-06
Proposal: Add counter metric to track alerts dropped for being outside of active/within muted time bounds


tjhop:
When using [time intervals](https://prometheus.io/docs/alerting/latest/configuration/#time_interval) for routes, alerts received outside of the correct time window are dropped with a log message:

https://github.com/prometheus/alertmanager/blob/6ce841ca22b4724b5ac40e501d4198a40a262ecc/notify/notify.go#L841-L846
https://github.com/prometheus/alertmanager/blob/6ce841ca22b4724b5ac40e501d4198a40a262ecc/notify/notify.go#L878-L884

Imagine a scenario where:
1. there is >=1 routes with time intervals set (eg, engineering teams wanting alerts during business hours)
2. there is a NOC or similar team available to triage alerts 24/7

Since alerts received outside of valid time intervals aren't retried and are outright dropped, it could be helpful to the NOC to have a count of how many alerts have been dropped for tracking or meta-alerting purposes (for example, a high rate of dropped alerts could indicate a more severe issue and justify the NOC reaching out to subject matter experts, even if outside of business hours). 

Using the existing log lines, it is possible to, for example, create a recording rule within loki and generate a metric off of it. However, the log entries don't inform the user _how many_ alerts were dropped. 

I'm proposing a new counter metric `alertmanager_alerts_dropped_total` should be created, and when alerts are dropped, it should be incremented by how many alerts were dropped (ie, `counterVar.Add(len(alerts))`)

gotjosh:
I'm favour of this -- but I think the name should be `alertmanager_alerts_supressed_total` as this is the terminology we use when an alert is muted.

PR #3565, opened 2023-10-20
feat: add counter to track alerts dropped outside of time_intervals

tjhop:
Addresses: #3512

This adds a new counter metric `alertmanager_alerts_supressed_total` that is incremented by `len(alerts)` when an alert is suppressed for being outside of a time_interval, ie inside of a mute_time_intervals or outside of an active_time_intervals.

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -251,6 +251,7 @@ type Metrics struct {
 	numTotalFailedNotifications        *prometheus.CounterVec
 	numNotificationRequestsTotal       *prometheus.CounterVec
 	numNotificationRequestsFailedTotal *prometheus.CounterVec
+	numAlertsSuppressedTotal           prometheus.Counter
 	notificationLatencySeconds         *prometheus.HistogramVec
 
 	ff featurecontrol.Flagger
@@ -284,6 +285,11 @@ func NewMetrics(r prometheus.Registerer, ff featurecontrol.Flagger) *Metrics {
 			Name:      "notification_requests_failed_total",
 			Help:      "The total number of failed notification requests.",
 		}, labels),
+		numAlertsSuppressedTotal: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: "alertmanager",
+			Name:      "alerts_supressed_total",
+			Help:      "The total number of alerts suppressed for being outside of active time intervals or within muted time intervals.",
+		}),
 		notificationLatencySeconds: prometheus.NewHistogramVec(prometheus.HistogramOpts{
 			Namespace: "alertmanager",
 			Name:      "notification_latency_seconds",
@@ -296,7 +302,7 @@ func NewMetrics(r prometheus.Registerer, ff featurecontrol.Flagger) *Metrics {
 	r.MustRegister(
 		m.numNotifications, m.numTotalFailedNotifications,
 		m.numNotificationRequestsTotal, m.numNotificationRequestsFailedTotal,
-		m.notificationLatencySeconds,
+		m.numAlertsSuppressedTotal, m.notificationLatencySeconds,
 	)
 
 	return m
@@ -382,8 +388,8 @@ func (pb *PipelineBuilder) New(
 
 	ms := NewGossipSettleStage(peer)
 	is := NewMuteStage(inhibitor)
-	tas := NewTimeActiveStage(intervener)
-	tms := NewTimeMuteStage(intervener)
+	tas := NewTimeActiveStage(intervener, pb.metrics)
+	tms := NewTimeMuteStage(intervener, pb.metrics)
 	ss := NewMuteStage(silencer)
 
 	for name := range receivers {
@@ -877,13 +883,14 @@ func (n SetNotifiesStage) Exec(ctx context.Context, l log.Logger, alerts ...*typ
 }
 
 type timeStage struct {
-	muter types.TimeMuter
+	muter   types.TimeMuter
+	metrics *Metrics
 }
 
 type TimeMuteStage timeStage
 
-func NewTimeMuteStage(m types.TimeMuter) *TimeMuteStage {
-	return &TimeMuteStage{m}
+func NewTimeMuteStage(m types.TimeMuter, metrics *Metrics) *TimeMuteStage {
+	return &TimeMuteStage{m, metrics}
 }
 
 // Exec implements the stage interface for TimeMuteStage.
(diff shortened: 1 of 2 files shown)


tjhop:
@gotjosh this is an attempt to concisely shim in the new counter discussed in #3512, please lemme know your thoughts. 

Of note, I wonder if the naming of this new metric (`alertmanager_alerts_suppressed_total`) will cause confusion with the existing `alertmanager_alerts{state="suppressed"}` metric that gets exposed. Reading through the source indicates the existing metric only reflects alerts suppressed through silences or inhibitions, and this new metric cares only about alerts suppressed for violating a given time interval. It would seem they're complimentary, but I think the metric names used are close enough that people may not understand the distinction

notify/notify.go
```diff
@@ -251,6 +251,7 @@ type Metrics struct {
 	numTotalFailedNotifications        *prometheus.CounterVec
 	numNotificationRequestsTotal       *prometheus.CounterVec
 	numNotificationRequestsFailedTotal *prometheus.CounterVec
+	numAlertsSuppressedTotal           prometheus.Counter
```
gotjosh, line 254: Apologies for not catching this earlier on, but yes, what we want here is `numNotificationSuppressedTotal` it's a bit confusing but we have strict nomenclature at a metric level and `notification` is not the same thing as an `alert`.

Alerts are what the Alertmanager receives and as you correctly figured out we already reflect those with the suppressed label on `alertmanager_alerts`.

What we want here is `alertmanager_notifications_supressed_total`

notify/notify.go
```diff
@@ -284,6 +285,11 @@ func NewMetrics(r prometheus.Registerer, ff featurecontrol.Flagger) *Metrics {
 			Name:      "notification_requests_failed_total",
 			Help:      "The total number of failed notification requests.",
 		}, labels),
+		numAlertsSuppressedTotal: prometheus.NewCounter(prometheus.CounterOpts{
+			Namespace: "alertmanager",
+			Name:      "alerts_supressed_total",
```
gotjosh, line 290:
```suggestion
			Name:      "notifications_supressed_total",
```

notify/notify.go
gotjosh: There's also an argument on wether we should use a label to specify if this was suppressed by a an `active_interval` or a `muted_interval` but let's leave this out for now and see wether this is useful on its own as is.

notify/notify.go
```diff
@@ -902,6 +909,7 @@ func (tms TimeMuteStage) Exec(ctx context.Context, l log.Logger, alerts ...*type
 
 	// If the current time is inside a mute time, all alerts are removed from the pipeline.
 	if muted {
+		tms.metrics.numAlertsSuppressedTotal.Add(float64(len(alerts)))
 		level.Debug(l).Log("msg", "Notifications not sent, route is within mute time")
```
gotjosh, line 913:
```suggestion
		level.Debug(l).Log("msg", "Notifications not sent, route is within mute time", "alerts", len(alerts))
```

notify/notify.go
```diff
@@ -939,6 +947,7 @@ func (tas TimeActiveStage) Exec(ctx context.Context, l log.Logger, alerts ...*ty
 
 	// If the current time is not inside an active time, all alerts are removed from the pipeline
 	if !muted {
+		tas.metrics.numAlertsSuppressedTotal.Add(float64(len(alerts)))
 		level.Debug(l).Log("msg", "Notifications not sent, route is not within active time")
```
gotjosh, line 951:
```suggestion
		level.Debug(l).Log("msg", "Notifications not sent, route is not within active time", "alerts", len(alerts))
```

notify/notify.go
gotjosh: Now that we're here let's try and log the number of alerts we suppressed for traceability.

gotjosh (commented):
Thank you very much for your contribution, the logic overall looks great but please look at my comments!

notify/notify.go
tjhop: Ok, I like that distinction much better, I'll move towards this phrasing. 

> There's also an argument on wether we should use a label to specify if this was suppressed by a an active_interval or a muted_interval but let's leave this out for now and see wether this is useful on its own as is.

Agreed, and sounds good to me :+1:

notify/notify.go
tjhop: I'll fix this with the rest of the feedback (and I'll fix my typo there too :wink: )

notify/notify.go
tjhop: added in e15292f

notify/notify.go
tjhop: added in e15292f

notify/notify.go
tjhop: addressed in e15292f

tjhop:
@gotjosh ready for round 2 when time permits

tjhop:
Accidentally forgot to sign the last commit. Amended and force pushed to address

notify/notify.go
```diff
@@ -284,6 +285,11 @@ func NewMetrics(r prometheus.Registerer, ff featurecontrol.Flagger) *Metrics {
 			Name:      "notification_requests_failed_total",
 			Help:      "The total number of failed notification requests.",
 		}, labels),
+		numNotificationSuppressedTotal: prometheus.NewCounter(prometheus.CounterOpts{
```
alexweav, line 288: If we decide to follow the other comment, we could add a `reason` label here. Then, users can split up the individual rates based on what did the suppression (was it silenced, or part of a mute time? etc).

It may be worth adding this even if we don't follow my other comment - one reason for active times, and another for mute times, in case users are using both simultaneously.

notify/notify.go
```diff
@@ -382,8 +388,8 @@ func (pb *PipelineBuilder) New(
 
 	ms := NewGossipSettleStage(peer)
 	is := NewMuteStage(inhibitor)
-	tas := NewTimeActiveStage(intervener)
-	tms := NewTimeMuteStage(intervener)
+	tas := NewTimeActiveStage(intervener, pb.metrics)
+	tms := NewTimeMuteStage(intervener, pb.metrics)
 	ss := NewMuteStage(silencer)
```
alexweav, line 393: Have we considered adding silenced and inhibited notifications as part of this metric?  I.e. we could also pass the metrics into `NewMuteStage` here for silences and on line 390 for inhibitions, and increment based on `muted` in that stage as well.

I ask because we didn't specifically tie the metric to mutes (it's not called `num_notifications_muted_total`) and then users could track suppressions across all different methods.

tjhop: I'm not opposed to that :+1:

notify/notify.go
tjhop: Like I said in response to the other comment, I can add the metrics to the other stages as well. And agreed, I think a `reason` label is much more obviously valuable now

tjhop:
Rebased to handle merge conflict :+1:

tjhop:
@alexweav does the current change set more resemble what you had in mind? Lemme know your thoughts

gotjosh (approved):
LGTM

PR #3567, opened 2023-10-20
Support UTF-8 label matchers: Use compat package in Alertmanager server

grobinson-grafana:
This pull request adds use of the compat package in Alertmanager server that will allow users to switch between the new matchers/parse parser and the old pkg/labels parser. The new matchers/parse parser uses a fallback mechanism where if the input cannot be parsed in the new parser it then attempts to use the old parser. If an input is parsed in the old parser but not the new parser then a warning log is emitted.


api/v2/api.go
```diff
@@ -508,7 +509,7 @@ func (api *API) getSilencesHandler(params silence_ops.GetSilencesParams) middlew
 	matchers := []*labels.Matcher{}
 	if params.Filter != nil {
 		for _, matcherString := range params.Filter {
-			matcher, err := labels.ParseMatcher(matcherString)
+			matcher, err := compat.Matcher(matcherString)
```
grobinson-grafana, line 512: Thinking about Mimir here, we might want to add the function pointer to `API` instead of using the global compat here. This would allow us to switch between parsing modes for individual tenants inside of having to do it server-wide.

silence/silence.go
```diff
@@ -469,6 +471,20 @@ func (s *Silences) GC() (int, error) {
 
 // ValidateMatcher runs validation on the matcher name, type, and pattern.
 var ValidateMatcher = func(m *pb.Matcher) error {
+	return validateClassicMatcher(m)
+}
+
+// InitFromFlags initializes the validation function from the flagger.
+func InitFromFlags(l log.Logger, f featurecontrol.Flagger) {
```
grobinson-grafana, line 478: I found an interesting issue here! If you create a silence with UTF-8 matchers and then restart Alertmanager with the "classic-matchers-parsing" feature flag, you can no longer edit or expire the silence because the validation function returns an error. However, the silence will still expire once it's expiration time has elapsed.

grobinson-grafana: I've attempted to fix that with this commit [here](https://github.com/prometheus/alertmanager/pull/3567/commits/c0dedde5e3360f55bf4a7d557e5ea9f1a9269a24).

grobinson-grafana:
@gotjosh let me know when you have some time and we can review this together?

types/types.go
```diff
@@ -302,6 +306,58 @@ type Alert struct {
 	Timeout   bool
 }
 
+var validateLs = func(ls model.LabelSet) error {
```
gotjosh, line 309: I think we can do a bit better here - I'd hope we can do dependency injection for these changes.

1. It's future-proofing ourselves because as we add more feature flags, it's very likely we'd need to control these components.
2. It'd make things easier to follow (e.g. you wouldn't need to override the call to `Validate` or have a need to use global variables.

silence/silence.go
gotjosh: Great Find! I'll take a look once we have the other changes in.

silence/silence.go
```diff
@@ -469,6 +471,20 @@ func (s *Silences) GC() (int, error) {
 
 // ValidateMatcher runs validation on the matcher name, type, and pattern.
 var ValidateMatcher = func(m *pb.Matcher) error {
+	return validateClassicMatcher(m)
```
gotjosh, line 474: Please see my comment on types/types.go - I think we shoudl get rid of this global variable and use depency injection of the feature flagger to control how we validate these matchers.

cmd/alertmanager/main.go
gotjosh, line 1: To execute my comment on passing down feature control to each component, you'd just need to pass the feature control struct to both `silence.Options` and `api.Options`.

gotjosh (commented):
Great job so far, please take a look at my comments.

api/v2/api.go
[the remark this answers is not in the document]
gotjosh: Can you please give me a review when you get the chance so that you can rebase this PR?

cli/alert_query.go
```diff
@@ -82,7 +83,7 @@ func (a *alertQueryCmd) queryAlerts(ctx context.Context, _ *kingpin.ParseContext
 		m := a.matcherGroups[0]
 		_, err := compat.Matcher(m)
 		if err != nil {
-			a.matcherGroups[0] = fmt.Sprintf("alertname=%s", m)
+			a.matcherGroups[0] = fmt.Sprintf("alertname=%s", strconv.Quote(m))
```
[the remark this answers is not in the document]
grobinson-grafana: Yes, correct! I noticed it as the changes I made in this PR caused a test to fail.

test/with_api_v2/acceptance/silence_test.go
```diff
@@ -15,12 +15,147 @@ package test
...
+
+receivers:
+- name: "default"
+  webhook_configs:
+  - url: 'http://%s'
+`
+
+	at := NewAcceptanceTest(t, &AcceptanceOpts{
+		Tolerance: 150 * time.Millisecond,
+	})
```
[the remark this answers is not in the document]
grobinson-grafana: Remember the default mode of Alertmanager (unless you use `classic-mode`) is UTF-8, with fallback to classic mode. Makes me think I should rename `utf8-mode` to `utf8-strict-mode`.

test/with_api_v2/acceptance/silence_test.go
```diff
@@ -15,12 +15,147 @@ package test
...
+			StartsAt: dateTimePtr(strfmt.DateTime(now)),
+			EndsAt:   dateTimePtr(strfmt.DateTime(now.Add(24 * time.Hour))),
+		},
+	}
+	silenceParams := silence.NewPostSilencesParams()
+	silenceParams.Silence = &ps
+
+	_, err := am.Client().Silence.PostSilences(silenceParams)
+	require.NoError(t, err)
+}
```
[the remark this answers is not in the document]
grobinson-grafana: I'll take a look. I might need to make some larger changes to `acceptance.go` to support restarting clusters, as it seems at present stopping a cluster deletes all it's data.

gotjosh (commented):
Great job! I only have 2 files left to review - silence and silce_test.

api/v2/api.go
```diff
@@ -505,17 +513,10 @@ func matchFilterLabels(matchers []*labels.Matcher, sms map[string]string) bool {
...
-				level.Debug(logger).Log("msg", "Failed to parse matchers", "err", err)
-				return silence_ops.NewGetSilencesBadRequest().WithPayload(err.Error())
-			}
-
-			matchers = append(matchers, matcher)
-		}
+	matchers, err := parseFilter(params.Filter)
+	if err != nil {
+		level.Debug(logger).Log("msg", "Failed to parse matchers", "err", err)
+		return alert_ops.NewGetAlertsBadRequest().WithPayload(err.Error())
```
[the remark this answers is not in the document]
grobinson-grafana: Oops!

test/with_api_v2/acceptance/silence_test.go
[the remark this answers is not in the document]
gotjosh: Thanks! I completely forgot about this haha.

silence/silence.go
```diff
@@ -662,7 +701,28 @@ func (s *Silences) expire(id string) error {
 		sil.EndsAt = now
 	}
 
-	return s.setSilence(sil, now)
+	// This code has been copied from setSilence to avoid an issue where
```
gotjosh, line 704: There's a lot of risk with drift with copying this code, and I would rather avoid it if possible. Can you elaborate on your decision to outright remove the validation? What kind of risks does removing the validation pose?

I suggest you add a boolean to the arguments of `validateSilence` to skip the validation should you wish to continue with this approach.

silence/silence.go
```diff
@@ -487,6 +497,29 @@ var ValidateMatcher = func(m *pb.Matcher) error {
...
+		}
+		if _, err := regexp.Compile(m.Pattern); err != nil {
+			return fmt.Errorf("invalid regular expression %q: %s", m.Pattern, err)
+		}
+	default:
+		return fmt.Errorf("unknown matcher type %q", m.Type)
+	}
+	return nil
+}
+
```
gotjosh, line 522: I have my reservations when it comes to this code - to me, this feels wrong. We're accepting matches as user input from the silences API in a structured format:

```
 "matchers": [
    {
      "name": "string",
      "value": "string",
      "isRegex": true,
      "isEqual": true
    }
  ],
```

In my opinion, this validation should happen at an API level and ideally make the input matchers go through the parser, by taking each parser and constructing a string that would you give you the struct directly. 

My main worry is that right now validation for entities is spreadout in multiple layers which makes it difficult to track who validates what.

Not something that we need to solve right now, but it's worth thinking about.

gotjosh (commented):
LGTM but please take a look at my comment wrt the silences validation.

I think we're missing three things at this point:

1. Make the CI pass
2. The test to catch the removal of validation in `expire`
3. The avoidance of copy paste code.

silence/silence.go
gotjosh: > The validation was removed to fix this issue https://github.com/prometheus/alertmanager/pull/3567#discussion_r1392961898

Sorry, I don't think that was my question, I read your comment and your decision-making is still not clear - It was more around the decision to remove the validation as part of solving the issue, which I guess your second comment touches a bit on.

I think I follow your line of thought, it seems strange that expiring a silence validates the silence itself. I guess it's a byproduct (assuming original author's intention) of re-using the `setSilence` function and not because the actual validation is needed. 

If we go with that assumption, please make sure is documented at the call site when you add the boolean.

grobinson-grafana: > I think I follow your line of thought, it seems strange that expiring a silence validates the silence itself. I guess it's a byproduct (assuming original author's intention) of re-using the setSilence function and not because the actual validation is needed.

Yes! I think the reason `expire` uses `setSilence` is a.) it was convenient to re-use `setSilence` for expiration and b.) once a silence has been validated in the past then, unless the silence is updated, it should always be valid in the future. Not an unreasonable assumption if validation rules never change.

However, in my opinion deletion should not do validation as otherwise we risk locking users out of specific operations (i.e. what happens if a user somehow has an invalid silence that needs to be deleted? How can the user recover from an inconsistent state to a consistent state?).

silence/silence.go
```diff
@@ -662,7 +703,7 @@ func (s *Silences) expire(id string) error {
 		sil.EndsAt = now
 	}
 
-	return s.setSilence(sil, now)
+	return s.setSilence(sil, now, true)
```
gotjosh, line 706:
```suggestion
        // We must skip validation as we expire the silence to avoid not being able to expire a silence that was previously created using Valid UTF-8 characters. 
        // This validation was removed as we allowed the Alertmanager labels names and matchers to be any valid UTF-8 string and can be introduced if necessary once the feature is made permanent.
	return s.setSilence(sil, now, true)
```

gotjosh (approved):
LGTM - please take a look at my comments but overall I have no further blocking questions.

Great work and thank you very much for your contribution.

gotjosh:
I've merged #2469, can you please rebase so that I can also merge this one?

grobinson-grafana:
> I've merged #2469, can you please rebase so that I can also merge this one?

Done! 🙂

gotjosh (approved):
LGTM

PR #3568, opened 2023-10-24
Support UTF-8 label matchers: Remove braces from suggestion

grobinson-grafana:
This commit removes the open and close braces from the suggestion as braces do not make sense in the configuration file. This does not change the behavior of the suggestion whatsoever as these are optional.

### DIFF
--- a/matchers/compat/parse.go
+++ b/matchers/compat/parse.go
@@ -152,14 +152,12 @@ func fallbackMatchersParser(l log.Logger) matchersParser {
 				return nil, invalidErr
 			}
 			var sb strings.Builder
-			sb.WriteRune('{')
 			for i, n := range m {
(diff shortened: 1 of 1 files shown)


grobinson-grafana:
For example, if the configuration file contained the following:

```
receivers:
  - name: test
route:
  receiver: test
  group_wait: 15s
  group_interval: 30s
  routes:
    - receiver: test
      matchers:
        - foo=!!bar
 ```

The suggestion would have been to change `foo=!!bar` with `{foo="!!bar"}`. However 1. it's not required and 2. it looks bad in the configuration file. This pull request changes the suggestion to just `foo="!!bar"` instead.

gotjosh (approved):

PR #3571, opened 2023-10-25
Support UTF-8 label matchers: Do not allow unquoted escape sequences

grobinson-grafana:
This commit updates the matchers parser to reject unquoted openmetrics escape sequences. As an example, `foo=bar\n` will no longer parse, and must instead be written as `foo="bar\n"`. This avoids an issue where the input is valid in both the matchers and classic parsers, but results in two different parsings.


grobinson-grafana:
Please refer to the comment I left here https://github.com/prometheus/alertmanager/pull/3453#issuecomment-1779052198

gotjosh (approved):
LGTM

Discussed offline that this technically a bug as parsing in both but producing different results is not part of our strategy.

Issue #3486, opened 2023-08-24
Support UTF-8 names and values in label matchers


grobinson-grafana:
### What

[The Prometheus maintainers have agreed to support UTF-8 label matchers in Alertmanager](https://docs.google.com/document/d/11LC3wJcVk00l8w5P3oLQ-m3Y37iom6INAMEu2ZAGIIE/edit#bookmark=id.wm24eue9w2eg). This issue tracks the work being done to deliver this change in Alertmanager.

### Background

To support UTF-8 label matchers in Alertmanager we have restricted the grammar for label matchers and written a new LL(1) parser to parse this grammar. Our motivation for doing this, instead of extending the existing regular expression parser, is that we found a number of ambiguities in the existing grammar and parser that would have been made worse if we added support for UTF-8 label matchers. You can find a presentation that explains these ambiguities [here](https://docs.google.com/presentation/d/1WOnWFttxY8ApUB79RSt0-v9UBzxCGoID4ryqAGibD-E/edit#slide=id.p).

### Examples of ambiguities

```
{foo=}} equivalent to {foo="}"}, but I think this should be an error
{{foo=} is an error because of two {{, unlike the above
{foo=~} could be either {foo=~""} or {foo="~"}, it's interpreted in current versions as {foo=~""}
{foo=,} is equivalent to {foo=""}, but I think should be an error as a comma with no value has a high likelihood of being human error
{foo=,,} is an error, unlike {foo=,} or {foo=}}
{foo= } is equivalent to {foo=""}
{foo= }b is equivalent to {foo="}b"}
{foo= b} and {foo=b } are equivalent to {foo="b"}, but {foo=b b} is equivalent to {foo="b b"}
```

| Example       | Can parse?|
| ------------- | ----------------- |
| {foo=bar😊}   | ✅ |
| {foo="bar😊"} | ✅ |
| {foo😊=bar}   | ❌ |
| {"foo😊"=bar} | ✅ |

Here `bar😊` is allowed on the right hand side of the expression without quotes but not the left.

### Tasks

- [x] https://github.com/prometheus/alertmanager/pull/3453
- [x] https://github.com/prometheus/alertmanager/pull/3507 and https://github.com/prometheus/alertmanager/pull/3523
- [x] https://github.com/prometheus/alertmanager/pull/3483
- [x] https://github.com/prometheus/alertmanager/pull/3567

PR #3580, opened 2023-11-01
Support UTF-8 label matchers: Update Stringer for non-Prometheus compliant label names

grobinson-grafana:
This commit updates the String method to print non-Prometheus compliant label names in a format that can be parsed in the UTF-8 parser. Such inputs are never valid in the classic parser. If the label name is Prometheus compliant, it is still printed unquoted.

### DIFF
--- a/pkg/labels/matcher.go
+++ b/pkg/labels/matcher.go
@@ -18,7 +18,9 @@ import (
 	"encoding/json"
 	"fmt"
 	"regexp"
+	"strconv"
 	"strings"
+	"unicode"
 
 	"github.com/prometheus/common/model"
 )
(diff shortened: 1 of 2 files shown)


pkg/labels/matcher.go
```diff
@@ -199,3 +204,13 @@ func (ms Matchers) String() string {
...
 	return buf.String()
 }
+
+// This is copied from matchers/parse/lexer.go. It will be removed when
+// the transition window from classic matchers to UTF-8 matchers is complete,
+// as then we can use double quotes when printing the label name for all
+// matchers. Until then, the classic parser does not understand double quotes
+// around the label name, so we use this function as a heuristic to tell if
+// the matcher was parsed with the UTF-8 parser or the classic parser.
+func isReserved(r rune) bool {
```
gotjosh, line 214: I'm assuming making the function public and re-using it here was a no-go for some reason?

I'd love if we could please keep track of these tidbits that we need to remove once we're done.

gotjosh (approved):

pkg/labels/matcher.go
gotjosh: I added a task to https://github.com/prometheus/alertmanager/issues/3486

grobinson-grafana: It's a bit of an odd one I think. Remember how we made the lexer and all it's related functions package-private? It felt weird to me that we would expose this **one** function that checks for reserved characters when the rest of the lexer and it's related functions are inaccessible.

Besides, I'm thinking we'll delete this once we removed the old parser so I felt it was cleaner to duplicate it and add a comment.

gotjosh (approved):
LGTM

PR #3581, opened 2023-11-01
Update go.mod to require Go 1.21

grobinson-grafana:
I would like to propose we update the minimum Go version to 1.21. Go [1.20](https://tip.golang.org/doc/go1.20) and [1.21](https://tip.golang.org/doc/go1.21) have a number of useful additions to the built-in packages. For example, I would like to use `strings.ContainsFunc` [here](https://github.com/prometheus/alertmanager/pull/3580).

If we don't want to remove Go 1.18 and Go 1.19 support for the time being then that's fine too! I just thought I would see if this is an option before inlining the function to make it work on Go versions 1.19 and older.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ jobs:
           command: sudo rm -rf /usr/local/go
       # Whenever the Go version is updated here, .promu.yml should also be updated.
       - go/install:
-          version: "1.20"
+          version: "1.21"
       - run:
           name: Remove generated code
           command: make clean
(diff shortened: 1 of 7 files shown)


simonpasquier (approved):
LGTM

simonpasquier (commented):
Oh I guess you need to upgrade the circleci/go [orb](https://circleci.com/developer/orbs/orb/circleci/go) to 1.9.0.

simonpasquier (approved):

Issue #3577, opened 2023-10-31
Add jira notifier / integration of jiralert


jkroepke:
**What did you do?**

I would like to integrate a JIRA notifier into AlertManager.

**What did you expect to see?**

A native Atlassian Jira integration

**What did you see instead? Under which circumstances?**

Currenty, adapter is needed to create JIRA issues from AM. (https://github.com/prometheus-community/jiralert)

---

Before starting to develop such an integration into AM, I would like to ask if the maintainers are open for such an integration.

Currently, I able to find 5 year old comments (https://github.com/prometheus/alertmanager/pull/1012#pullrequestreview-67829943) which seems like an NO

An actual statement would helpfull to get the directory, if jiralert should be the focus or the integration into AM.

bwplotka:
We discussed this internally and indeed it's an option, but we were looking for maintainers:  https://github.com/prometheus-community/jiralert/issues/145

PR #3590, opened 2023-11-01
jira integration

jkroepke:
This PR track the Atlassian JIRA integration into Alertmanger.

See: #3577 

Replaces: https://github.com/prometheus-community/jiralert

---

What I test:

* Jira Cloud only
* Create issues
* Auto-Close issue with `resolve_transition`
* Reopen issue with `reopen_transition`
* Reopen duration
* Custom Fields
* `api_token_file`


k0ste:
Can this integration cover another of jiralert issue - is custom field for alert hash?
https://github.com/prometheus-community/jiralert/issues/164

Current code search hash in hardcoded place - in `labels`

```go
jql.WriteString(fmt.Sprintf(`project = "%s" and
labels=%q order by status ASC,
resolutiondate DESC`, n.conf.Project, key.Hash()))
```

Thanks

jkroepke:
Sure. After merge. Everyone is invited to improve.

I would not like to add more feature here. Even I understand you request.

The current PR covery already a lot of cases to test. Adding more new features results into more testing. I would prefer to add new features in separate PRs.

holger-waschke:
Is this implementation only for [Jira Cloud ](https://developer.atlassian.com/cloud/jira/platform/rest/v3/intro/#about) or can I still use the on-premise API with a PAT Authentication?

jkroepke:
Hi @holger-waschke 

I'm not able to test on-premise. But you can enter an Server API Path to an on premise instance and the authentication is still auth basic which should work on on-premise, too.

The integration does not use any cloud specific features. If https://github.com/prometheus-community/jiralert works on-premise, this integration should work, too.

I remember that Issue handling, the Jira Software Server API v2 and Cloud API v2 is the same.

* https://docs.atlassian.com/software/jira/docs/api/REST/9.11.0/#api/2/issue
* https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issues/#api-rest-api-2-issue-post

But please keep in mind that this integration will always have best-effort since most users are not able to setup a on premise instance.

docs/configuration.md
```diff
@@ -743,6 +750,92 @@ Microsoft Teams notifications are sent via the [Incoming Webhooks](https://learn
 [ http_config: <http_config> | default = global.http_config ]
 ```
 
+### `<jira_config>`
+
+JIRA notification are sent via [JIRA Rest API v2](https://developer.atlassian.com/cloud/jira/platform/rest/v2/intro/)
```
simonpasquier, line 755:
```suggestion
JIRA notifications are sent via [JIRA Rest API v2](https://developer.atlassian.com/cloud/jira/platform/rest/v2/intro/)
```

docs/configuration.md
```diff
@@ -743,6 +750,92 @@ Microsoft Teams notifications are sent via the [Incoming Webhooks](https://learn
...
+for format the issue description and V3 uses [Atlassian Document Format (ADF)](https://developer.atlassian.com/cloud/jira/platform/apis/document/structure/).
+The default `jira.default.description` template only works with V2.
+
+```yaml
+# Whether to notify about resolved alerts.
+[ send_resolved: <boolean> | default = true ]
+
+# The incoming webhook URL.
+[ webhook_url: <secret> ]
+
```
simonpasquier, line 768: I don't see it declared in pkg/config?
```suggestion
```

notify/jira/jira.go
```diff
@@ -0,0 +1,343 @@
...
+// Notifier implements a Notifier for JIRA notifications.
+type Notifier struct {
+	conf    *config.JiraConfig
+	tmpl    *template.Template
+	logger  log.Logger
+	client  *http.Client
+	retrier *notify.Retrier
+}
+
+// New returns a new Webhook.
```
simonpasquier, line 49: (nit)
```suggestion
```

simonpasquier (commented):
Thanks for starting this! I haven't looked yet in details into the notifier but I suppose that it's almost a 1:1 copy from the original project?

Not a blocker but I think that it's the first notifier which needs to query the external system before actually sending the notification. It should be highlighted in the documentation IMHO.

jkroepke:
> Thanks for starting this! I haven't looked yet in details into the notifier but I suppose that it's almost a 1:1 copy from the original project?

No. while jiralert uses a unofficial outdated REST SDK for JIRA, I used the standard http client to setup the few http calls.

config/config.go
```diff
@@ -741,6 +771,10 @@ type GlobalConfig struct {
 
 	HTTPConfig *commoncfg.HTTPClientConfig `yaml:"http_config,omitempty" json:"http_config,omitempty"`
 
+	JiraAPIURL           *URL       `yaml:"jira_api_url,omitempty" json:"jira_api_url,omitempty"`
+	JiraAPIUsername      string     `yaml:"jira_api_username,omitempty" json:"jira_api_username,omitempty"`
+	JiraAPIToken         Secret     `yaml:"jira_api_token,omitempty" json:"jira_api_token,omitempty"`
+	JiraAPITokenFile     string     `yaml:"jira_api_token_file,omitempty" json:"jira_api_token_file,omitempty"`
```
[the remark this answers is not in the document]
jkroepke: Done!

docs/configuration.md
```diff
@@ -743,6 +750,92 @@ Microsoft Teams notifications are sent via the [Incoming Webhooks](https://learn
 [ http_config: <http_config> | default = global.http_config ]
 ```
 
+### `<jira_config>`
+
+JIRA notification are sent via [JIRA Rest API v2](https://developer.atlassian.com/cloud/jira/platform/rest/v2/intro/)
+or [JIRA REST API v3](https://developer.atlassian.com/cloud/jira/platform/rest/v3/intro/#version).
+
+Both APIs have the same feature set. The difference is that V2 uses [Wiki Markup](https://jira.atlassian.com/secure/WikiRendererHelpAction.jspa?section=all)
+for format the issue description and V3 uses [Atlassian Document Format (ADF)](https://developer.atlassian.com/cloud/jira/platform/apis/document/structure/).
```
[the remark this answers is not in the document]
jkroepke: I have no idea. v3 is beta for years.

But I would keep v2 until it gets deprecated/removed:

* Jira On Premise support only v2, because v3 is cloud exclusive
* The body of jira issue in Rest V3 must be in ADF format which no one understand (each paragraph orlink is a nested json property). In v2, Jira Markup is supported.

config/config.go
```diff
@@ -539,6 +542,33 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+			}
+			if jira.APIURL == nil {
+				if c.Global.JiraAPIURL == nil {
+					return fmt.Errorf("no global Jira Cloud URL set")
+				}
+				jira.APIURL = c.Global.JiraAPIURL
+			}
+			if !strings.HasSuffix(jira.APIURL.Path, "/") {
+				jira.APIURL.Path += "/"
+			}
```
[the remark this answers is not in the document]
jkroepke: I will put the PR in draft until I do some tests with `url.JoinPath`.

jkroepke:
@simonpasquier I tested the PR after review and code changes again.

notify/jira/jira.go
```diff
@@ -0,0 +1,339 @@
...
+
+	url := n.conf.APIURL.JoinPath(path)
+	req, err := http.NewRequest(method, url.String(), body)
+	if err != nil {
+		return nil, false, err
+	}
+
+	var token string
+	if n.conf.APIToken != "" {
+		token, err = tmplTextFunc(string(n.conf.APIToken))
```
simonpasquier, line 301: why have a template here?

notify/jira/jira.go
```diff
@@ -0,0 +1,339 @@
...
+		if err != nil {
+			return nil, false, fmt.Errorf("template error: %w", err)
+		}
+	} else {
+		content, err := os.ReadFile(n.conf.APITokenFile)
+		if err != nil {
+			return nil, false, fmt.Errorf("read key_file error: %w", err)
+		}
+		token, err = tmplTextFunc(string(content))
```
simonpasquier, line 310: same question here

notify/jira/jira.go
```diff
@@ -0,0 +1,339 @@
...
+		if err != nil {
+			return nil, false, fmt.Errorf("template error: %w", err)
+		}
+	}
+
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Accept-Language", "en")
+	req.Header.Set("X-Force-Accept-Language", "true")
+	req.SetBasicAuth(n.conf.APIUsername, strings.TrimSpace(token))
```
simonpasquier, line 319: It wasn't obvious to me that the token would be used for BasicAuth. Also it seems to be deprecated/discouraged from Atlassian documentation (https://developer.atlassian.com/server/jira/platform/basic-authentication/).

notify/jira/types.go
```diff
@@ -0,0 +1,149 @@
...
+	Name           string `json:"name"`
+	StatusCategory struct {
+		Key string `json:"key"`
+	} `json:"statusCategory"`
+}
+
+type issueSearch struct {
+	Expand       []string `json:"expand"`
+	Fields       []string `json:"fields"`
+	FieldsByKeys bool     `json:"fieldsByKeys"`
```
simonpasquier, line 63: The presence of this field fails with my JIRA instance:

```
ts=2024-01-05T15:25:57.986Z caller=dispatch.go:353 level=error component=dispatcher msg="Notify for alerts failed" num_alerts=1 err="jira/jira[0]: notify retry canceled due to unrecoverable 
error after 1 attempts: unexpected status code 400: {\"errorMessages\":[\"Unrecognized field \\\"fieldsByKeys\\\" (Class com.atlassian.jira.rest.v2.search.SearchRequestBean), not marked as i
gnorable\\n at [Source: org.apache.catalina.connector.CoyoteInputStream@148a2304; line: 1, column: 54] (through reference chain: com.atlassian.jira.rest.v2.search.SearchRequestBean[\\\"field
sByKeys\\\"])\"]}"                                                                                                                                                                            

```

It succeeds if I remove/comment this line.

simonpasquier (commented):
I started testing this PR and have few more questions. Great work anyway!

notify/jira/jira.go
jkroepke: Since I'm familiar with OpsGenie, I start the integration based on OpsGenie integration first. The OpsGenie integration also does tmpl on the api key.

https://github.com/prometheus/alertmanager/blob/ab2248afa7e53b78c63e577a10649e6ae8b9db2a/notify/opsgenie/opsgenie.go#L277-L286

I'm just replicate the existing logic without having a deeper knowledge, if it would be necessary.

notify/jira/jira.go
jkroepke: see above

notify/jira/jira.go
jkroepke: The documentation you reference goes for Jira Server (on premise). 

The documentation for Cloud REST API does not deprecate the Auth Basic authentication flow: https://developer.atlassian.com/cloud/jira/platform/basic-auth-for-rest-apis/ and its recommend method for manual REST calls.

No words about deprecation here. Since Jira REST API is https exclusively, Auth Basic is sufficient enough.

notify/jira/types.go
jkroepke: Are you testing this against an on premise instance?

jkroepke: Removed `fieldsByKeys` in 3aa5a4c and tested against Cloud API

jkroepke:
@simonpasquier could you please take a look here? Thanks you.

christianjedroCDT:
i´ve tested the PR with our Jira Cloud Setup and everything looks good so far. Would be really interested in using this, is there an update @simonpasquier?

Matz13:
Showing some additional support here! The tests I ran all look good and I can't wait to get the official release.

jkroepke:
I fixed all the pipeline issues now

notify/jira/jira.go
```diff
@@ -0,0 +1,339 @@
...
+			jql.WriteString(`statusCategory != Done and `)
+		}
+	} else {
+		reopenDuration := int64(time.Duration(n.conf.ReopenDuration).Minutes())
+		if reopenDuration != 0 {
+			jql.WriteString(fmt.Sprintf(`(resolutiondate is EMPTY OR resolutiondate >= -%dm) and `, reopenDuration))
+		}
+	}
+
+	jql.WriteString(fmt.Sprintf(`project = "%s" and labels=%q order by status ASC,resolutiondate DESC`, n.conf.Project, key.Hash()))
```
simonpasquier, line 208:
```suggestion
	jql.WriteString(fmt.Sprintf(`project=%q and labels=%q order by status ASC,resolutiondate DESC`, n.conf.Project, key.Hash()))
```

notify/jira/jira.go
```diff
@@ -0,0 +1,339 @@
...
+	case 0:
+		level.Debug(n.logger).Log("msg", "found no existing issue", "alert", key.String())
+		return nil, false, nil
+	default:
+		level.Warn(n.logger).Log("msg", "more than one issue matched, picking most recently resolved", "alert", key.String(), "picked", issueSearchResult.Issues[0].Key)
+
+		fallthrough
+	case 1:
+		return &issueSearchResult.Issues[0], false, nil
+	}
```
simonpasquier, line 239: I've found the `fallthrough` trick a bit hard to follow. I'd prefer a simpler version.

```suggestion
	if issueSearchResult.Total == 0 {
		level.Debug(n.logger).Log("msg", "found no existing issue", "alert", key.String())
		return nil, false, nil
	}
	
	if issueSearchResult.Total > 1 {
		level.Warn(n.logger).Log("msg", "more than one issue matched, selecting the most recently resolved", "alert", key.String(), "selected", issueSearchResult.Issues[0].Key)
	}

	&issueSearchResult.Issues[0], false, nil
```

notify/jira/jira.go
```diff
@@ -0,0 +1,339 @@
...
+	sort.Strings(requestBody.Fields.Labels)
+
+	return requestBody, nil
+}
+
+func (n *Notifier) searchExistingIssue(tmplTextFunc templateFunc, key notify.Key, status model.AlertStatus) (*issue, bool, error) {
+	jql := strings.Builder{}
+
+	if n.conf.WontFixResolution != "" {
+		jql.WriteString(fmt.Sprintf(`resolution != "%s" and `, n.conf.WontFixResolution))
```
simonpasquier, line 193:
```suggestion
		jql.WriteString(fmt.Sprintf(`resolution != %q and `, n.conf.WontFixResolution))
```

notify/jira/jira.go
simonpasquier: I find personally that templating for credentials is a mistake and brings more problems than value. I would abstain from replicating this "functionality" across receivers.
See also my other comment about leveraging http_config and the fact that dedicated token fields aren't really needed.

notify/jira/jira.go
simonpasquier: Right I mixed up things. Having said that the Server docs encourage people to use Personal Access Tokens:

image

simonpasquier (commented):
Apologies it took so long for me to review! I'll keep on the top of my list if you have cycles to address my comments.
My main feedback is that I would leverage the `http_config` fields to configure authentication rather than adding new fields.

notify/jira/jira.go
jkroepke: fixed suggestion with return at the final line

```suggestion
	if issueSearchResult.Total == 0 {
		level.Debug(n.logger).Log("msg", "found no existing issue", "alert", key.String())
		return nil, false, nil
	}
	
	if issueSearchResult.Total > 1 {
		level.Warn(n.logger).Log("msg", "more than one issue matched, selecting the most recently resolved", "alert", key.String(), "selected", issueSearchResult.Issues[0].Key)
	}
	
	return &issueSearchResult.Issues[0], false, nil
```

config/notifiers.go
```diff
@@ -825,3 +833,53 @@ func (c *MSTeamsConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	CustomFields map[string]any `yaml:"custom_fields,omitempty" json:"custom_fields,omitempty"`
+}
+
+func (c *JiraConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	*c = DefaultJiraConfig
+	type plain JiraConfig
+	if err := unmarshal((*plain)(c)); err != nil {
+		return err
+	}
+	if c.APIToken == "" && c.APITokenFile == "" {
```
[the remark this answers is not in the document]
jkroepke: removing api_token anyways.

docs/configuration.md
```diff
@@ -84,6 +84,12 @@ global:
   # Note that Go does not support unencrypted connections to remote SMTP endpoints.
   [ smtp_require_tls: <bool> | default = true ]
 
+  # Default settings for the JIRA integration. 
+  [ jira_api_url: <string> ]
+  [ jira_api_username: <string> ]
+  [ jira_api_token: <secret> ]
+  [ jira_api_token_file: <filepath> ]
```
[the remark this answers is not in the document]
jkroepke: This make sense to me. I just looked at the OpsGenie notifier here, how it's solved there.

jkroepke:
@simonpasquier thanks for taking an other look here.

Since I open the PR 7 months ago, I learned bit more about go. In conclusion, I added some units tests and I was able to identity some bugs.

Sadly, adding the units tests here requires additional review cycles.

I will continue some integrations tests and once I'm done, I will re-request an review.

notify/jira/types.go
```diff
@@ -0,0 +1,110 @@
...
+	Total  int     `json:"total"`
+	Issues []issue `json:"issues"`
+}
+
+type issueTransitions struct {
+	Transitions []idNameValue `json:"transitions"`
+}
+
+// MarshalJSON merges the struct issueFields and issueFields.CustomField together.
+func (i issueFields) MarshalJSON() ([]byte, error) {
```
jkroepke, line 75: MarshalJSON is mandatory, because `encoding/json` does not support `json:",inline"`

jkroepke (commented):
I run some integration test against our complexest project (tons of custom fields) and we figure out that my custom implemention of marshal custom fields fails. I pulled a new depedency in which cast a nested map[any]any to an map with string keys. This is mandatory because stlib json encoder wont support map[any]any.

Additionally, I added some unit tests to cover the basics. They revival some bugs which are fixed now. I hope they are not too complex.

docs/configuration.md
```diff
@@ -931,6 +936,89 @@ Microsoft Teams notifications are sent via the [Incoming Webhooks](https://learn
...
+# Add labels to issues
+static_labels: 
+  [ - <string> ... ]
+  
+# Add specific group labels to issue
+group_labels: 
+  [ - <string> ... ]
+
+# Priority of issue
+[ priority: <tmpl_string> ]
```
jkroepke, line 976: Thats a bit sad. My hope was that is possible to set the priority based on the highest `severity` of all alerts in that group.  But it seems not possible yet. Not sure, if it should be covered.

config/notifiers.go
```diff
@@ -825,3 +834,42 @@ func (c *MSTeamsConfig) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
+	Description  string   `yaml:"description,omitempty" json:"description,omitempty"`
+	StaticLabels []string `yaml:"static_labels,omitempty" json:"static_labels,omitempty"`
+	GroupLabels  []string `yaml:"group_labels,omitempty" json:"group_labels,omitempty"`
+	Priority     string   `yaml:"priority,omitempty" json:"priority,omitempty"`
+	IssueType    string   `yaml:"issue_type,omitempty" json:"issue_type,omitempty"`
+
+	ReopenTransition  string   `yaml:"reopen_transition,omitempty" json:"reopen_transition,omitempty"`
+	ResolveTransition string   `yaml:"resolve_transition,omitempty" json:"resolve_transition,omitempty"`
+	WontFixResolution string   `yaml:"wont_fix_resolution,omitempty" json:"wont_fix_resolution,omitempty"`
+	ReopenDuration    Duration `yaml:"reopen_duration,omitempty" json:"reopen_duration,omitempty"`
```
simonpasquier, line 855: why not model.Duration?

config/notifiers.go
```diff
@@ -670,17 +678,17 @@ func (c *VictorOpsConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
 	return nil
 }
 
-type duration time.Duration
+type Duration time.Duration
```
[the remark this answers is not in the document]
simonpasquier: If we need this, let's have it a separate PR (see also my other comment below).

go.mod
```diff
@@ -37,6 +37,7 @@ require (
 	github.com/shurcooL/httpfs v0.0.0-20190707220628-8d4bc4ba7749
 	github.com/shurcooL/vfsgen v0.0.0-20200824052919-0d455de96546
 	github.com/stretchr/testify v1.9.0
+	github.com/trivago/tgo v1.0.7
```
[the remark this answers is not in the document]
simonpasquier: Not terribly excited to depend on a project that isn't very active (e.g. it doesn't have a go.mod). While I understand that it is stable and proven, I'll look for other options.

Can you also provide an example of something that breaks encoding/json?

docs/configuration.md
```diff
@@ -931,6 +936,89 @@ Microsoft Teams notifications are sent via the [Incoming Webhooks](https://learn
...
+[ reopen_transition: <string> ]
+
+# If reopen_transition is defined, ignore issues with that resolution
+[ wont_fix_resolution: <string> ]
+
+# If reopen_transition is defined, reopen issue not older than ... Note: The resolutiondate field is used to determine the age of the issue.
+[ reopen_duration: <duration> ]
+
+# Other issue and custom fields
+fields:
```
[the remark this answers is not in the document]
simonpasquier: From https://developer.atlassian.com/server/jira/platform/jira-rest-api-examples/#setting-custom-field-data-for-other-field-types

> A custom field is actually referenced by customfield\_ + the field ID, rather than just the field ID.

I'd rather keep a dedicated "custom fields" entry and have users providing the field IDs:

````
custom_fields:
  "10000": "foo"
  "10001": ["bar", "fred"]
``

File: go.mod (hunk 20 above)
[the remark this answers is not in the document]
Comment (jkroepke) [reply go.mod:40]:
```
custom_fields:
  customfield_10010: { value: "Monitoring" }
```

Alertmanager parses the nested map as `map[interface{}]interface{}` and `encoding/json` does not support this kind of map. The library ensures that each nested map is casted to `map[string]interface{}`

File: docs/configuration.md (hunk 21 above)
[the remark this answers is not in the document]
Comment (jkroepke) [reply docs/configuration.md:994]: In conclusion, 2 separate configs for fields?

```
fields:
  component: { name: "Monitoring" }
custom_fields:
  "10000": "foo"
  "10001": ["bar", "fred"]
```

I don't feel thats worth to have 2 here. I renamed it from `custom_fields` to `fields` because it would support jira native fields, too.

I personally would prefer a 1:1 mapping between the field configuration and Jira REST API

File: go.mod (hunk 20 above)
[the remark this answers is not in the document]
Comment (jkroepke) [reply go.mod:40]: >  I'll look for other options.

https://github.com/json-iterator/go natively supports encoding of `map[interface{}]interface{}`, but I feel there would much less acceptance for a different JSON library.

File: config/notifiers.go (hunk 18 above)
Comment (jkroepke) [reply config/notifiers.go]: I used Duration for constancy reasons. It's used inside the PushoverConfig. But I guess PushoverConfig can model.Duration as well.

File: config/notifiers.go (hunk 19 above)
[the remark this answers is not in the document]
Comment (jkroepke) [reply config/notifiers.go]: replace with model.Duration

**Comment by jkroepke** [comment]
I have tested the priority template together with `inhibit_rules` which works really great together.

```yaml
inhibit_rules:
  - equal:
      - namespace
      - alertname
    source_matchers:
      - severity = critical
    target_matchers:
      - severity =~ warning|info
  - equal:
      - namespace
      - alertname
    source_matchers:
      - severity = warning
    target_matchers:
      - severity = info
```

Manual testing against Jira Atlassian Cloud works fine.

In conclusion, we have 2 open topic:

- `fields` vs `custom_fields` inside jira_config
- `github.com/trivago/tgo` for cast nested map from `map[any]any` to `map[string]any`

notify/jira/jira.go
```diff
@@ -0,0 +1,319 @@
...
+	for _, issueTransition := range issueTransitions.Transitions {
+		if issueTransition.Name == transitionName {
+			return issueTransition.ID, false, nil
+		}
+	}
+
+	return "", false, fmt.Errorf("can't find transition %s for issue %s", transitionName, issueKey)
+}
+
+func (n *Notifier) TransitionIssue(key notify.Key, issueKey, transitionName string) (bool, error) {
```
cropalato, line 264: This comment is more of a curiosity. I was wondering why this function starts with a capital letter. In my understanding, the functions (Golang) with the first capital letter are normally used when they're referenced from outside the package, which makes sense for the New() function, but TransitionIssue() seemed out of the norm.

notify/jira/jira.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+
+	// Do not create new issues for resolved alerts
+	if existingIssue == nil && alerts.Status() == model.AlertResolved {
+		return false, nil
+	}
+
+	path := "issue"
+	method := http.MethodPost
+
```
cropalato, line 95: To make it clear to read, why do not move 
```
	path := "issue"
	method := http.MethodPost
```
into `if existingIssue == nil {`
you are redefining in `else` condition

notify/jira/jira.go
jkroepke: Good catch! I guess its an leftover since I do some copy paste from jiralert

cropalato (commented):
The code looks good to me. Unfortunately, I cannot test in my setup, because I have a problem with the field Labels. I'll way the code be merged to include https://github.com/prometheus-community/jiralert/pull/189

jkroepke:
> The code looks good to me. Unfortunately, I cannot test in my setup, because I have a problem with the field Labels. I'll way the code be merged to include https://github.com/prometheus-community/jiralert/

I‘m open to extend the feature set after this PR got merged.

Background: it seems like AM has some review capacity issues and increase the scope involves an additional review each time. The current scope is well tested (including manual test against prod systems).

SuperQ:
Yes, I would like to see this merged so we can iterate on additional featres.

docs/configuration.md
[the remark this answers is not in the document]
SuperQ: I agree, mapping with the Jira API makes sense. If they use a string convention rather than a separate API, we should follow the same. Perhaps document/link to this convention here.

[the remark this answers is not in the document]
simonpasquier: After more thinking, I agree with both of you.

docs/configuration.md
```diff
@@ -944,6 +949,92 @@ Microsoft Teams notifications are sent via the [Incoming Webhooks](https://learn
...
+
+# Name of the workflow transition to reopen an issue. The target status should not have the category "done".
+# NOTE: The name of the transition can be localized and depends on the language setting of the service account.
+[ reopen_transition: <string> ]
+
+# If reopen_transition is defined, ignore issues with that resolution.
+[ wont_fix_resolution: <string> ]
+
+# If reopen_transition is defined, reopen the issue when it is not older than this value (rounded down to the nearest minute).
+The resolutiondate field is used to determine the age of the issue.
```
SuperQ, line 1006: Oops, I think this lost a needed comment:

```suggestion
# The resolutiondate field is used to determine the age of the issue.
```

docs/configuration.md
```diff
@@ -944,6 +949,92 @@ Microsoft Teams notifications are sent via the [Incoming Webhooks](https://learn
...
+# The project key where issues are created.
+project: <string>
+
+# Issue summary template.
+[ summary: <tmpl_string> | default = '{{ template "jira.default.summary" . }}' ]
+
+# Issue description template.
+[ description: <tmpl_string> | default = '{{ template "jira.default.description" . }}' ]
+
+# Add labels to issues
```
SuperQ, line 980:
```suggestion
# Add labels to issues.
```

docs/configuration.md
```diff
@@ -944,6 +949,92 @@ Microsoft Teams notifications are sent via the [Incoming Webhooks](https://learn
...
+static_labels: 
+  [ - <string> ... ]
+  
+# Add specific group labels to issue
```
SuperQ, line 984:
```suggestion
# Add specific group labels to the issue.
```

docs/configuration.md
```diff
@@ -944,6 +949,92 @@ Microsoft Teams notifications are sent via the [Incoming Webhooks](https://learn
...
+[ reopen_transition: <string> ]
+
+# If reopen_transition is defined, ignore issues with that resolution.
+[ wont_fix_resolution: <string> ]
+
+# If reopen_transition is defined, reopen the issue when it is not older than this value (rounded down to the nearest minute).
+The resolutiondate field is used to determine the age of the issue.
+[ reopen_duration: <duration> ]
+
+# Other issue and custom fields
```
SuperQ, line 1009:
```suggestion
# Other issue and custom fields.
```

SuperQ (commented):
Some minor doc consistency nits.

docs/configuration.md
jkroepke: Thanks!

go.mod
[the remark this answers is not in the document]
jkroepke: I have no clue, why json-iterator works fine on the lined go test above, but while integrate it inside AM, tests in AM are failed with 

```
error marshalling request body: jira.issue.Fields: json: unsupported type: map[interface {}]interface {}
```

I may give up here, since I don't know why it works on a synthetic environment, but not on AM.

For reference, here is my code (https://github.com/prometheus/alertmanager/commit/cf85bbec731b1530e49daf1ee8a2a58974588b5c)

docs/configuration.md
```diff
@@ -944,6 +949,91 @@ Microsoft Teams notifications are sent via the [Incoming Webhooks](https://learn
...
+# Issue description template.
+[ description: <tmpl_string> | default = '{{ template "jira.default.description" . }}' ]
+
+# Add labels to issues
+static_labels: 
+  [ - <string> ... ]
+  
+# Add specific group labels to issue
+group_labels: 
+  [ - <string> ... ]
```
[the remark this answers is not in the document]
jkroepke: I just copy/paste the logic from jiralert, but I like the suggestion. 

Added.

SuperQ (approved):

cropalato (approved):

jkroepke:
Hi all, after 2 approvals. how we should proceed?

SuperQ:
Ping @gotjosh @simonpasquier

simonpasquier (commented):
I'll review by the end of the week, sorry for the delay.

simonpasquier (approved):
I'd have minor comments on making the code a bit more compact but nothing that should prevent merging this PR. Thanks again for your patience!
I'll also look into building a new RC for Alertmanager.

notify/jira/jira.go
```diff
@@ -0,0 +1,320 @@
...
+func (n *Notifier) prepareIssueRequestBody(tmplTextFunc templateFunc) (issue, error) {
+	summary, err := tmplTextFunc(n.conf.Summary)
+	if err != nil {
+		return issue{}, fmt.Errorf("template error: %w", err)
+	}
+
+	// Recursively convert any maps to map[string]interface{}, filtering out all non-string keys, so the json encoder
+	// doesn't blow up when marshaling JIRA requests.
+	fieldsWithStringKeys, err := tcontainer.ConvertToMarshalMap(n.conf.Fields, func(v string) string { return v })
```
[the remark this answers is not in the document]
dswarbrick: I'm having a bit of trouble understanding the need for this. [`ConvertToMarshalMap`](https://pkg.go.dev/github.com/trivago/tgo@v1.0.7/tcontainer#ConvertToMarshalMap) is being passed a `map[string]any` and returns a [`MarshalMap`](https://pkg.go.dev/github.com/trivago/tgo@v1.0.7/tcontainer#MarshalMap), which is a type alias for `map[string]interface{}`, aka `map[string]any`. Am I missing something?

If this `ConvertToMarshalMap` function is _really_ needed, it seems like quite an overkill to bring in a whole new (and maybe abandoned?) third-party package, just to call a [~30-line function](https://github.com/trivago/tgo/blob/v1.0.7/tcontainer/marshalmap.go#L52). Can't that function just be "vendored" (with copyright attribution)? It's essentially just a recursive function which loops over map values and does a bit of `reflect` introspection - not exactly rocket science.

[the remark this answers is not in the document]
jkroepke: @dswarbrick I agree with you. I was to blinded from the current jiralert. If you ask me, I would use plain strings here. I may propose an PR with would change that behavior.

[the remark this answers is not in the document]
dswarbrick: I'm a bit confused, because `n.conf.Fields` is defined as type `map[string]any`, and this is presumably populated by the config parser, reading YAML (or JSON?) config file. So if there is something that `encoding/json` does not like, presumably it's going to throw an error at _that_ stage.

I'm not too sure where the nervousness about filtering out non-string keys comes from either, since Go handles numeric keys fine:
```go
	foo := map[int]string{123: "abc", 456: "def"}
	b, _ := json.Marshal(foo)
	fmt.Println(string(b))
```
```
{"123":"abc","456":"def"}
```
Go map keys can be any type that is "comparable" - "in short, comparable types are boolean, numeric, string, pointer, channel, and interface types, and structs or arrays that contain only those types." ([ref](https://go.dev/blog/maps#key-types)). I haven't exhaustively tested all those types with JSON marshaling, but one also has to ask, how would such "exotic" types get into `n.conf.Fields` in the first place?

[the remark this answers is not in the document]
jkroepke: nested maps will be parsed as map[any]any which is not supported in stdlib.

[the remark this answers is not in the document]
jkroepke: the Alertmanager configuration is YAML. `yaml.Unmarshal` from `gopkg.in/yaml.v2` will be used to unmarshal to config. 

https://go.dev/play/p/yGRo4OgqXrY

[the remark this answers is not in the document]
dswarbrick: Right, I was beginning to suspect that the YAML unmarshal behaved a little differently to JSON unmarshal.

Interestingly, bumping the `yaml.v2` to `yaml.v3` would make things a lot simpler:
```
yaml.v2 unmarshal: map[string]interface {}{"field":[]interface {}{map[interface {}]interface {}{"value":0}}}
yaml.v3 unmarshal: map[string]interface {}{"field":[]interface {}{map[string]interface {}{"value":0}}}
```

Issue #3593, opened 2023-11-02
Flaky test: TestClusterJoinAndReconnect/TestTLSConnection


alexweav:
**What did you do?**

This test has failed on the CI intermittently a couple of times in roughly 10 builds, on changes that seem unrelated.

Example of an affected build:
https://github.com/prometheus/alertmanager/actions/runs/6737380881/job/18314694645?pr=3589

**What did you expect to see?**

I expected the test to pass on the CI when changes don't affect it.

**What did you see instead? Under which circumstances?**

https://github.com/prometheus/alertmanager/actions/runs/6737380881/job/18314694645?pr=3589

**Environment**

n/a - CI

gotjosh:
All the assertions as part of these tests are super flaky

We already have https://github.com/prometheus/alertmanager/issues/3287

grobinson-grafana:
Looking at this test, I'm not sure how this is expected to work? Peer 2 joins Peer 1 in a different goroutine:

```
go p2.Settle(context.Background(), 0*time.Second)
```

The test goroutine then calls:

```
require.Equal(t, 2, p1.ClusterSize())
```

which can happen before the Go runtime has even scheduled the goroutine that calls `Settle` - let alone the memberlist having settled. I think settle should be passed a context with a timeout, and then if it succeeds the test checks the cluster size, otherwise it fails.

grobinson-grafana:
Ah! I think this test is missing `require.NoError(t, p2.WaitReady(context.Background()))` as per the other tests.

gotjosh:
Thank you both for the effort! And in particular @alexweav for the fix.

PR #3594, opened 2023-11-03
Fix TestClusterJoinAndReconnect/TestTLSConnection

grobinson-grafana:
This commit attempts to fix the TestTLSConnection test from running the assertion before the memberlist is settled. It copies WaitReady from the other tests in this file.

#3593

### DIFF
--- a/cluster/cluster_test.go
+++ b/cluster/cluster_test.go
@@ -334,6 +334,7 @@ func testTLSConnection(t *testing.T) {
 	)
 	require.NoError(t, err)
 	go p2.Settle(context.Background(), 0*time.Second)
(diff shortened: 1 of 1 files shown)


cluster/cluster_test.go
```diff
@@ -334,6 +334,7 @@ func testTLSConnection(t *testing.T) {
 	)
 	require.NoError(t, err)
 	go p2.Settle(context.Background(), 0*time.Second)
+	require.NoError(t, p2.WaitReady(context.Background()))
```
gotjosh, line 337: https://github.com/prometheus/alertmanager/pull/3589 already included this so closing this.

gotjosh: Thanks though!

PR #3595, opened 2023-11-07
Support UTF-8 label matchers: unquote should check for invalid UTF-8 code points

grobinson-grafana:
Quoted tokens can contain both UTF-8 byte and code point literals that should be interpreted when quoted. However, we need to check that the interpreted literals are valid UTF-8 code points or not. This now happens in unquote.


gotjosh (approved):
LGTM

PR #3596, opened 2023-11-08
Support UTF-8 label matchers: Use DI for matchers compat instead of package-level variables

grobinson-grafana:
This commit changes amtool and the matchers/compat package to support DI instead of package-level variables. The purpose of this change to allow multi-tenant Alertmanagers, such as those used in Cortex and Mimir to configure the feature flags per tenant and not just process-wide.

### DIFF
--- a/cli/alert_add.go
+++ b/cli/alert_add.go
@@ -25,7 +25,6 @@ import (
 
 	"github.com/prometheus/alertmanager/api/v2/client/alert"
 	"github.com/prometheus/alertmanager/api/v2/models"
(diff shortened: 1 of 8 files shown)


grobinson-grafana:
Supporting this is going to require a lot of changes to Alertmanager server (for example [here](https://github.com/prometheus/alertmanager/pull/3567/commits/69606eca5243106a4fd3b7dde01e0c8d95b05664)). It adds a lot of tech debt that will need to be later removed, and we still don't know if we really need per-tenant feature flags in Mimir. Going to close this for the time being and we can come back to it if needed.

PR #3599, opened 2023-11-10
fix: label mismatch for alertmanager_notifications_failed_total

cr7258:
Since [Alertmanager 0.26.0 version](https://github.com/prometheus/alertmanager/blob/ce6efba023b0397cb522d64e910684e48d12455f/CHANGELOG.md#0260--2023-08-23), a new label `reason` was added in the `alertmanager_notifications_failed_total` metric to indicate the type of error of the alert delivery. 

As a result, the original alert rules are broken because labels are mismatched between `alertmanager_notifications_failed_total` and `alertmanager_notifications_total metrics`.
Prometheus requires samples with exactly the same labels to get matched together when performing calculations. [docs](https://prometheus.io/docs/prometheus/latest/querying/operators/#one-to-one-vector-matches)

Use the `on` vector matching keyword to match only the `integration` label to allow for matching between series with different labels.

### DIFF
--- a/doc/alertmanager-mixin/alerts.libsonnet
+++ b/doc/alertmanager-mixin/alerts.libsonnet
@@ -44,7 +44,7 @@
               (
                 rate(alertmanager_notifications_failed_total{%(alertmanagerSelector)s}[5m])
               /
-                rate(alertmanager_notifications_total{%(alertmanagerSelector)s}[5m])
+                ignoring (reason) group_left rate(alertmanager_notifications_total{%(alertmanagerSelector)s}[5m])
               )
               > 0.01
             ||| % $._config,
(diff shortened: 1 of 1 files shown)


doc/alertmanager-mixin/alerts.libsonnet
```diff
@@ -44,7 +44,7 @@
               (
                 rate(alertmanager_notifications_failed_total{%(alertmanagerSelector)s}[5m])
               /
-                rate(alertmanager_notifications_total{%(alertmanagerSelector)s}[5m])
+                on (integration) group_left rate(alertmanager_notifications_total{%(alertmanagerSelector)s}[5m])
```
simonpasquier, line 47: I'd suggest to ignore the `reason` label since it's the only one that's different between the 2 metrics. It will change a bit the alert condition in case of multiple reasons (e.g. the sum for all reasons might be above the threshold while the individual reasons are below). But the current threshold value is low enough that it shouldn't be an issue in practice.

```suggestion
                ignoring (reason) group_left rate(alertmanager_notifications_total{%(alertmanagerSelector)s}[5m])
```

simonpasquier (commented):
Thanks for fixing it!

doc/alertmanager-mixin/alerts.libsonnet
cr7258: Thanks for your suggestion. Fixed.

gotjosh (commented):
LGTM

PR #3604, opened 2023-11-15
Support UTF-8 label matchers: Rename feature flags and make package public

grobinson-grafana:
This commit renames the feature flags "classic-matchers-parsing" and "utf8-matchers-parsing" to "classic-matchers" and "utf8-matchers". This better represents their use, for example, when validating both alerts and silences created via the API.

It also makes the feature flags package public for two reasons:

1. AllowedFlags is a public variable
2. We would rather use these consts in Mimir then have to hardcode the strings ourselves, and be subject to breaking changes in future
3. We can use the feature flags in acceptance tests without having to hardcode them


featurecontrol/featurecontrol.go
```diff
@@ -23,40 +23,40 @@ import (
 )
 
 const (
-	fcReceiverNameInMetrics  = "receiver-name-in-metrics"
-	fcClassicMatchersParsing = "classic-matchers-parsing"
-	fcUTF8MatchersParsing    = "utf8-matchers-parsing"
+	FeatureReceiverNameInMetrics = "receiver-name-in-metrics"
+	FeatureClassicMatchers       = "classic-matchers"
```
grobinson-grafana, line 27: Perhaps even something like this:

```suggestion
	FeatureClassicMode  = "classic-mode"
	FeatureUTF8Mode     = "utf8-mode"
```

alexweav, line 27: Is changing the actual feature _value_ that appears in configuration considered a breaking change? Won't any configurations that had the old toggle start to crash with `Unknown option '%s' for --enable-feature`?

grobinson-grafana: Shouldn't happen, we haven't made any public releases of Alertmanager with the old feature flags.

matchers/repl/repl
gotjosh, line 1: Can you please remove this file?

gotjosh (commented):
LGTM

I'm happy with classic mode and uft8 mode.

matchers/repl/repl
grobinson-grafana: How did that get there :D

gotjosh (approved):

PR #3610, opened 2023-11-21
API: Allow the Silences API to use their own 400 response

gotjosh:
There's no change to the logic here - it just struck me as odd that we were reusing the response from alerting groups in the silences call when it's pretty simple to add a new one in the API description.

This is more accurate as well, as this will be used to power the documentation.

### DIFF
--- a/api/v2/openapi.yaml
+++ b/api/v2/openapi.yaml
@@ -53,6 +53,8 @@ paths:
           description: Get silences response
           schema:
             $ref: '#/definitions/gettableSilences'
+        '400':
+          $ref: '#/responses/BadRequest'
         '500':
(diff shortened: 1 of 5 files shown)


api/v2/openapi.yaml
```diff
@@ -53,6 +53,8 @@ paths:
           description: Get silences response
           schema:
             $ref: '#/definitions/gettableSilences'
+        '400':
+          $ref: '#/responses/BadRequest'
```
gotjosh, line 57: this and the change in api/v2/api.go are the only changes as part of this PR - everything else is generated.

grobinson-grafana (approved):

PR #3616, opened 2023-11-23
Add summary to msteams notification

romankubu:
Hi,
currently msteams notification does not support summary atribute from Message card. Result is that in Activity (inside MS Teams) you don't know what kind of alert you received and you need to klik on it. It is same for notification on phone etc.

Example (bottom one is without pull request, top one is with pull request):
image


romankubu:
If i run `go run -tags=dev asset_generate.go` in `/assets` then those alerts are working well. How can i fix it?

docs/configuration.md
```diff
@@ -736,6 +736,9 @@ Microsoft Teams notifications are sent via the [Incoming Webhooks](https://learn
 # Message title template.
 [ title: <tmpl_string> | default = '{{ template "msteams.default.title" . }}' ]
 
+# Message title template.
```
zhan9san, line 739: I guess it's a typo

```suggestion
# Message summary template.
```

zhan9san (changes requested):

docs/configuration.md
romankubu: Thanks i missed that

gotjosh (approved):
LGTM

rgarcia89:
Someone knows if a new release will be created close by? This is a very unfavorable bug

PR #3619, opened 2023-11-24
Fix UTF-8 not supported in group_by

grobinson-grafana:
This commit fixes missing UTF-8 support in the group_by for routes.


gotjosh (approved):
LGTM

PR #3620, opened 2023-11-25
use Go standard errors

mmorel-35:
* Since Dec 1, 2021, github.com/pkg/errors is an archived project.
* Since go 1.13, there is a standard errors and fmt modules that provides equivalent ways to handle errors.

Signed-off-by: Matthieu MOREL <matthieu.morel35@gmail.com>


SuperQ (changes requested):
This needs a rebase.

SuperQ (approved):
:tada: Awesome

PR #3622, opened 2023-11-29
If the promu tool already exists, there is no need to re-download it

linuxgcc:
Optimize the compilation process , reduce re- downloads of promu


linuxgcc:
#3621

PR #3631, opened 2023-12-05
Add context reasons to notifications failed counter

wallee94:
The alertmanager [cancels the Dispatcher context](https://github.com/prometheus/alertmanager/blob/main/cmd/alertmanager/main.go#L415) when it receives a reload request, which cancels in-flight notifications and counts them as failed with reason "other".

This change adds two new reasons for context cancellation and deadline exceeding. It sets the reason in the counter if no other error was raised in previous attempts. It is helpful to monitor for failed requests, but at the same time, be able to discard failures from contexts canceled due to reloads or sigterms.


SuperQ (approved):
I wish we didn't increment the failed counter if the notification is "failed" but rescheduled. But this should help with managing alerts in case of more reload cancels/reschedules on busy Alertmanagers.

SuperQ:
Odd, this doesn't seem right to me.

```
notify/notify.go:838:6: ineffectual assignment to iErr (ineffassign)
					iErr = NewErrorWithReason(ContextCanceledReason, iErr)
					^
notify/notify.go:840:6: ineffectual assignment to iErr (ineffassign)
					iErr = NewErrorWithReason(ContextDeadlineExceededReason, iErr)
					^
```

wallee94:
@SuperQ yep, typo, had added a colon by accident and created a new var 🤦

PR #3635, opened 2023-12-08
README: Use new matchers syntax in config examples

hoffie:
#1023 (0.22+?) introduced a new, now preferred matchers syntax for routes and inhibitions. The README still uses the old, deprecated syntax. This PR updates those examples (and fixes a very minor example config comment typo).

:heavy_check_mark: I've checked that the new config still parses with alertmanager v0.26.
:heavy_check_mark: The quoting is consistent with the style used in the already updated inhibitions example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ route:
 
   # The child route trees.
   routes:
-  # This routes performs a regular expression match on alert labels to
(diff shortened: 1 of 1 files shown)


simonpasquier (approved):
Thanks!

PR #3648, opened 2023-12-28
Fix race conditions in the memory alerts store

damnever:
The main branch will easily fail the newly added test case:
```
=== RUN   TestAlertsConcurrently
    mem_test.go:565:
                Error Trace:    /prometheus-io/alertmanager/provider/mem/mem_test.go:565
                Error:          Not equal:
                                expected: 0
                                actual  : -171
                Test:           TestAlertsConcurrently
```

There are multiple race conditions in the provider/mem:
1. Any Put or GC operation that occurs concurrently with this code block will introduce a race condition https://github.com/prometheus/alertmanager/blob/c920b605b647f2548e23a3dc2fea014b9c9dbf43/provider/mem/mem.go#L204-L228
2. A race condition between Put() and Subscribe() can cause some newly added Alerts to be missed.

@simonpasquier @w0rm @gotjosh please take a look

### DIFF
--- a/provider/mem/mem.go
+++ b/provider/mem/mem.go
@@ -34,10 +34,11 @@ const alertChannelLength = 200
 type Alerts struct {
 	cancel context.CancelFunc
 
+	mtx sync.Mutex
+
 	alerts *store.Alerts
 	marker types.AlertMarker
 
-	mtx       sync.Mutex
 	listeners map[int]listeningAlerts
 	next      int
 
@@ -100,35 +101,58 @@ func NewAlerts(ctx context.Context, m types.AlertMarker, intervalGC time.Duratio
 		logger:    log.With(l, "component", "provider"),
 		callback:  alertCallback,
 	}
-	a.alerts.SetGCCallback(func(alerts []types.Alert) {
-		for _, alert := range alerts {
-			// As we don't persist alerts, we no longer consider them after
-			// they are resolved. Alerts waiting for resolved notifications are
-			// held in memory in aggregation groups redundantly.
-			m.Delete(alert.Fingerprint())
-			a.callback.PostDelete(&alert)
-		}
-
-		a.mtx.Lock()
-		for i, l := range a.listeners {
-			select {
-			case <-l.done:
-				delete(a.listeners, i)
-				close(l.alerts)
-			default:
-				// listener is not closed yet, hence proceed.
-			}
-		}
-		a.mtx.Unlock()
-	})
 
 	if r != nil {
 		a.registerMetrics(r)
 	}
 
-	go a.alerts.Run(ctx, intervalGC)
+	go a.gcLoop(ctx, intervalGC)
 
 	return a, nil
+}
+
+func (a *Alerts) gcLoop(ctx context.Context, interval time.Duration) {
+	t := time.NewTicker(interval)
+	defer t.Stop()
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-t.C:
+			a.gc()
+		}
+	}
+}
+
+func (a *Alerts) gc() {
+	a.mtx.Lock()
+	defer a.mtx.Unlock()
+
+	for _, alert := range a.alerts.List() {
+		if alert.Resolved() {
+			fp := alert.Fingerprint()
+			if err := a.alerts.Delete(fp); err != nil {
+				level.Error(a.logger).Log("msg", "error on delete alert", "err", err)
+				continue
+			}
+
+			// As we don't persist alerts, we no longer consider them after
+			// they are resolved. Alerts waiting for resolved notifications are
+			// held in memory in aggregation groups redundantly.
+			a.marker.Delete(fp)
+			a.callback.PostDelete(alert)
+		}
+	}
+
+	for i, l := range a.listeners {
+		select {
+		case <-l.done:
+			delete(a.listeners, i)
+			close(l.alerts)
+		default:
+			// listener is not closed yet, hence proceed.
+		}
+	}
 }
 
 // Close the alert provider.
@@ -177,8 +201,11 @@ func (a *Alerts) GetPending() provider.AlertIterator {
 
 	go func() {
 		defer close(ch)
+		a.mtx.Lock()
+		alerts := a.alerts.List()
+		a.mtx.Unlock()
 
-		for _, a := range a.alerts.List() {
+		for _, a := range alerts {
 			select {
 			case ch <- a:
 			case <-done:
@@ -192,6 +219,8 @@ func (a *Alerts) GetPending() provider.AlertIterator {
 
 // Get returns the alert for a given fingerprint.
 func (a *Alerts) Get(fp model.Fingerprint) (*types.Alert, error) {
+	a.mtx.Lock()
+	defer a.mtx.Unlock()
 	return a.alerts.Get(fp)
 }
 
@@ -200,6 +229,7 @@ func (a *Alerts) Put(alerts ...*types.Alert) error {
 	for _, alert := range alerts {
 		fp := alert.Fingerprint()
 
+		a.mtx.Lock()
 		existing := false
 
 		// Check that there's an alert existing within the store before
@@ -215,18 +245,19 @@ func (a *Alerts) Put(alerts ...*types.Alert) error {
 		}
 
 		if err := a.callback.PreStore(alert, existing); err != nil {
+			a.mtx.Unlock()
 			level.Error(a.logger).Log("msg", "pre-store callback returned error on set alert", "err", err)
 			continue
 		}
 
 		if err := a.alerts.Set(alert); err != nil {
+			a.mtx.Unlock()
 			level.Error(a.logger).Log("msg", "error on set alert", "err", err)
 			continue
 		}
 
 		a.callback.PostStore(alert, existing)
 
-		a.mtx.Lock()
 		for _, l := range a.listeners {
 			select {
 			case l.alerts <- alert:
@@ -241,6 +272,9 @@ func (a *Alerts) Put(alerts ...*types.Alert) error {
 
 // count returns the number of non-resolved alerts we currently have stored filtered by the provided state.
 func (a *Alerts) count(state types.AlertState) int {
+	a.mtx.Lock()
+	defer a.mtx.Unlock()
+
 	var count int
 	for _, alert := range a.alerts.List() {
 		if alert.Resolved() {
--- a/provider/mem/mem.go
+++ b/provider/mem/mem.go
@@ -191,13 +191,12 @@ func (a *Alerts) GetPending() provider.AlertIterator {
 		ch   = make(chan *types.Alert, alertChannelLength)
 		done = make(chan struct{})
 	)
+	a.mtx.RLock()
+	defer a.mtx.RUnlock()
+	alerts := a.alerts.List()
 
 	go func() {
 		defer close(ch)
-		a.mtx.RLock()
-		alerts := a.alerts.List()
-		a.mtx.RUnlock()
-
 		for _, a := range alerts {
 			select {
 			case ch <- a:
@@ -219,10 +218,12 @@ func (a *Alerts) Get(fp model.Fingerprint) (*types.Alert, error) {
 
 // Put adds the given alert to the set.
 func (a *Alerts) Put(alerts ...*types.Alert) error {
+	a.mtx.Lock()
+	defer a.mtx.Unlock()
+
 	for _, alert := range alerts {
 		fp := alert.Fingerprint()
 
-		a.mtx.Lock()
 		existing := false
 
 		// Check that there's an alert existing within the store before
@@ -238,13 +239,11 @@ func (a *Alerts) Put(alerts ...*types.Alert) error {
 		}
 
 		if err := a.callback.PreStore(alert, existing); err != nil {
-			a.mtx.Unlock()
 			level.Error(a.logger).Log("msg", "pre-store callback returned error on set alert", "err", err)
 			continue
 		}
 
 		if err := a.alerts.Set(alert); err != nil {
-			a.mtx.Unlock()
 			level.Error(a.logger).Log("msg", "error on set alert", "err", err)
 			continue
 		}
@@ -257,7 +256,6 @@ func (a *Alerts) Put(alerts ...*types.Alert) error {
 			case <-l.done:
 			}
 		}
-		a.mtx.Unlock()
 	}
 
 	return nil
--- a/provider/mem/mem.go
+++ b/provider/mem/mem.go
@@ -134,7 +134,7 @@ func (a *Alerts) gc() {
 		// they are resolved. Alerts waiting for resolved notifications are
 		// held in memory in aggregation groups redundantly.
 		a.marker.Delete(alert.Fingerprint())
-		a.callback.PostDelete(alert)
+		a.callback.PostDelete(&alert)
 	}
 
 	for i, l := range a.listeners {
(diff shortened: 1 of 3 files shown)


provider/mem/mem.go
```diff
@@ -100,35 +101,57 @@ func NewAlerts(ctx context.Context, m types.Marker, intervalGC time.Duration, al
...
+		select {
+		case <-ctx.Done():
+			return
+		case <-t.C:
+			a.gc()
+		}
+	}
+}
+
+func (a *Alerts) gc() {
```
qinxx108, line 126: nit: just a style question, should we put this gc function not belong to Alerts type as this implementation is specific to mem.go class

damnever: originally, mem.go used a separate goroutine for this task. However, this is not appropriate in this context because we need the lock protection from Alerts. 

I agree that we can create a non-async GC function in mem.go.

provider/mem/mem.go
```diff
@@ -178,8 +194,11 @@ func (a *Alerts) GetPending() provider.AlertIterator {
 
 	go func() {
 		defer close(ch)
+		a.mtx.Lock()
```
emanlodovice, line 197: Since we are locking a lot more now, should we consider using `RWMutex` for the lock? concurrent `GetPending` and `count` should be safe right?

provider/mem/mem.go
```diff
@@ -100,37 +101,52 @@ func NewAlerts(ctx context.Context, m types.Marker, intervalGC time.Duration, al
 		logger:    log.With(l, "component", "provider"),
 		callback:  alertCallback,
 	}
-	a.alerts.SetGCCallback(func(alerts []*types.Alert) {
```
emanlodovice, line 103: `SetGCCallback` and `Run` method on Alerts store are now unused. Should those be deleted?

rajagopalanand: I think `Run` is called by Inhibitor but `SetGCCallback` is unused in AM but since it's a public method, deleting it might break other things

provider/mem/mem.go
```diff
@@ -193,6 +212,8 @@ func (a *Alerts) GetPending() provider.AlertIterator {
 
 // Get returns the alert for a given fingerprint.
 func (a *Alerts) Get(fp model.Fingerprint) (*types.Alert, error) {
+	a.mtx.Lock()
```
rajagopalanand, line 215: Same comment as @emanlodovice here. Would `RWMutex` help? If the intent is to prevent `gc` from running, acquiring read lock here would work, right? That way, concurrent `Get` can run, right?

provider/mem/mem.go
```diff
@@ -242,6 +265,9 @@ func (a *Alerts) Put(alerts ...*types.Alert) error {
 
 // count returns the number of non-resolved alerts we currently have stored filtered by the provided state.
 func (a *Alerts) count(state types.AlertState) int {
+	a.mtx.Lock()
```
rajagopalanand, line 268: same comment as https://github.com/prometheus/alertmanager/pull/3648/files#r1490221562

emanlodovice (approved):

damnever:
@simonpasquier @w0rm @gotjosh please take a look

beorn7:
@simonpasquier @gotjosh is this on your radar?

damnever:
@grobinson-grafana would you also mind taking a look at this?

provider/mem/mem.go
```diff
@@ -100,37 +101,53 @@ func NewAlerts(ctx context.Context, m types.Marker, intervalGC time.Duration, al
...
-		}
-		a.mtx.Unlock()
-	})
 
 	if r != nil {
 		a.registerMetrics(r)
 	}
 
-	go a.alerts.Run(ctx, intervalGC)
+	go a.gcLoop(ctx, intervalGC)
```
grobinson-grafana, line 109: I see we stopped using `a.alerts.SetGCCallback` and `a.alerts.Run`, but I don't understand how that helps fix the race condition? Can we not use the mutex at the start of the callback function?

damnever: The race condition exists because both the AlertStoreCallback and mem.Alerts.alerts are not protected by the same lock, as I have stated in this PR comment.

> Can we not use the mutex at the start of the callback function?

I do not quite understand this.

damnever: and the mutex from store.Alerts is actually useless in this context

damnever: The gccallback occurs after the item is deleted, as shown here: https://github.com/prometheus/alertmanager/blob/c4a763c401742f410580df35a4eedb8daef2a219/store/store.go#L72-L92. The action to delete this item from store.Alerts is not protected by mem.Alerts.mtx. This inconsistency can cause a race condition. We need to ensure that both the callback and the actual delete action are protected by the same lock for the AlertStoreCallback to function properly. Without this, there would be a gap between the delete(store.alerts, xkey) and mem.Alerts.PostDelete actions. This small gap opens up the possibility for a race condition to occur.

grobinson-grafana: Ah! I understand the issue! Thanks for the additional explanation, it helped a lot! 🙂

grobinson-grafana: I think we can even delete the callback from the `provider/mem/mem.go` as it's not used?

grobinson-grafana:
> A race condition between Put() and Subscribe() can cause some newly added Alerts to be missed.

Apologies for asking lots of questions, but I would like to understand this case. I can see there is a race condition between the `Get` and `Set` operations in `Put` because 1.) two or more goroutines can call `Put` at the same time and 2.) the calls to `Get` and `Set` can be interleaved with a gc operation on the store, causing the alert to be deleted and the added back. I agree this needs to be fixed.

What I would still like to understand is how would the alert be missed? The mutex is acquired in `Subscribe` and before the alert is sent to the listeners, so it should not happen that a new listener misses an alert? Would it be possible to show how this happens with a test?

provider/mem/mem.go
grobinson-grafana: Ah it looks like its used in Grafana Mimir.

provider/mem/mem.go
```diff
@@ -200,6 +222,7 @@ func (a *Alerts) Put(alerts ...*types.Alert) error {
 	for _, alert := range alerts {
 		fp := alert.Fingerprint()
 
+		a.mtx.Lock()
```
grobinson-grafana, line 225: I think to make the `Put` operation atomic we should acquire the lock at the start of the method and then release it with `defer`?

provider/mem/mem.go
```diff
@@ -177,8 +194,11 @@ func (a *Alerts) GetPending() provider.AlertIterator {
 
 	go func() {
 		defer close(ch)
+		a.mtx.RLock()
```
grobinson-grafana, line 197: What if we move this lock and the `alerts` variable outside the goroutine so we can use `defer` for the unlock?

grobinson-grafana:
I also opened a draft PR https://github.com/prometheus/alertmanager/pull/3840 that builds on this fix. It removes `gc` and the callback from `store.go`.

provider/mem/mem.go
damnever: moving to the start seems the same to me, I believe the original intent is to make the lock scope as small as possible

provider/mem/mem.go
damnever: We can, but I don't see why we should. I think I would prefer the other method, and we are unable to determine the start time of the goroutine.

provider/mem/mem.go
damnever: however, moving to the start is better as this simplifies the logic.

damnever:
> > A race condition between Put() and Subscribe() can cause some newly added Alerts to be missed.
> 
> Apologies for asking lots of questions, but I would like to understand this case. I can see there is a race condition between the `Get` and `Set` operations in `Put` because 1.) two or more goroutines can call `Put` at the same time and 2.) the calls to `Get` and `Set` can be interleaved with a gc operation on the store, causing the alert to be deleted and the added back. I agree this needs to be fixed.
> 
> What I would still like to understand is how would the alert be missed? The mutex is acquired in `Subscribe` and before the alert is sent to the listeners, so it should not happen that a new listener misses an alert? Would it be possible to show how this happens with a test?

This all comes together, not as an independent case. The reason is that we can delete alerts solely in store.Alerts without the lock from mem.Alerts.mtx. Imagine if the gc happens right after the Put(before the fix), then the listener might miss some alerts due to the race condition.

provider/mem/mem.go
grobinson-grafana: > moving to the start seems the same to me

It's not quite the same. The difference is if we move it to the start then the entire Put operation is atomic. This stops two goroutines from interleaving their Put operations.

For example, goroutines 1 puts two alerts:

```
labels={foo=bar},status=firing
labels={bar=baz},status=resolved
```

And goroutine 2 puts the same alerts, but with swapped statuses:

```
labels={foo=bar},status=resolved
labels={bar=baz},status=firing
```

The order of insertion between the two goroutines is random and unpredictable. But if we acquire the lock at the start we make the entire operation serializable.

grobinson-grafana:
> The reason is that we can delete alerts solely in store.Alerts without the lock from mem.Alerts.mtx.

I agree!

> Imagine if the gc happens right after the Put(before the fix), then the listener might miss some alerts due to the race condition.

In this case the listener will receive two messages: one for the alert before GC and one for the alert after GC. Those messages might also be out of order. But because the alert sent to the listeners comes from `Put` and not from the store, I don't think it's possible for a `Put` operation to not send to a listener, even in the presence of a race condition?

grobinson-grafana:
> > > A race condition between Put() and Subscribe() can cause some newly added Alerts to be missed.
> > 
> > 
> > Apologies for asking lots of questions, but I would like to understand this case. I can see there is a race condition between the `Get` and `Set` operations in `Put` because 1.) two or more goroutines can call `Put` at the same time and 2.) the calls to `Get` and `Set` can be interleaved with a gc operation on the store, causing the alert to be deleted and the added back. I agree this needs to be fixed.
> > What I would still like to understand is how would the alert be missed? The mutex is acquired in `Subscribe` and before the alert is sent to the listeners, so it should not happen that a new listener misses an alert? Would it be possible to show how this happens with a test?
> 
> This all comes together, not as an independent case. The reason is that we can delete alerts solely in store.Alerts without the lock from mem.Alerts.mtx. Imagine if the gc happens right after the Put(before the fix), then the listener might miss some alerts due to the race condition.

> The reason is that we can delete alerts solely in store.Alerts without the lock from mem.Alerts.mtx.

I agree!

> Imagine if the gc happens right after the Put(before the fix), then the listener might miss some alerts due to the race condition.

In this case the listener will receive two events: an event for the Put and a second event for the GC? It can receive them out of order, but I don't think events can be lost?

The reason I think that is the alerts that are sent to `a.listeners` in `Put` comes from the argument `alerts` and not from the alerts in the mem store:

```go
		...
		a.mtx.Lock()
		for _, l := range a.listeners {
			select {
			case l.alerts <- alert:
			case <-l.done:
			}
		}
		a.mtx.Unlock()
	}
```

That means even when there is a gc between `Set` and `range a.listeners`, the gc operation cannot stop these alerts from being sent to the listeners. The listeners might receive the events out of order, but I still don't see how an event can be missed?

provider/mem/mem.go
grobinson-grafana: Should be unaffected as the `go` statement doesn't block.

damnever:
okay, they cannot be stopped, but this is a race.

provider/mem/mem.go
damnever: I mean, the goroutine is scheduled by the runtime, so there will be some differences under high load.

provider/mem/mem.go
damnever: I believe this makes sense, but the difference in a batch isn't that significant either.

damnever: this lock may contains CPU-sensitive operations such as fingerprint calculations

damnever:
@grobinson-grafana I have rebased with the main branch and changed some locks to use defer. Please take another look. If there are no major change requirements, I believe we should merge this first and make further improvements as needed.

grobinson-grafana (commented):
LGTM. Just one comment!

provider/mem/mem.go
```diff
@@ -174,11 +191,13 @@ func (a *Alerts) GetPending() provider.AlertIterator {
 		ch   = make(chan *types.Alert, alertChannelLength)
 		done = make(chan struct{})
 	)
+	a.mtx.RLock()
+	defer a.mtx.RUnlock()
+	alerts := a.alerts.List()
```
grobinson-grafana, line 196: I think the important change here is that moving the alerts out of the goroutine means `alerts` contains the set of alerts at the time `Pending()` was called, not the set of alerts at the time the goroutine was scheduled. This can be different if a `Put` or GC operation interleaves a call to `Pending()` and the start of the goroutine.

provider/mem/mem.go
```diff
@@ -34,10 +34,11 @@ const alertChannelLength = 200
 type Alerts struct {
 	cancel context.CancelFunc
 
+	mtx sync.RWMutex
```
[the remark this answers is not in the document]
damnever: It's not strictly the same, as these two locks have different scopes, and I think I will keep this change, as this reason is not convincing to me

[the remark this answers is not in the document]
damnever: further improvements can continue from here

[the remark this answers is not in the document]
grobinson-grafana: I think there might have been some confusion here. I think changing the mutex from `sync.Mutex` to a `sync.RWMutex` is a nice performance improvement, I was just suggesting to do it in a separate PR. There are a couple of reasons for this:

1. You can make the performance improvement by changing `sync.Mutex` to `sync.RWMutex` in both the memstore and memprovider at the same time. You can also provide some benchmarks to show how much faster it is without it being lost in this discussion which is related to another issue (fixing a race condition).
2. By keeping the fix for the race condition and performance improvements in separate PRs, it makes it so much easier to see where changes came from and why they were added when running `git blame`. Otherwise it looks like the `sync.RWMutex` is somehow related to the race condition, when it's actually just a nice performance improvement. You have to instead look at the discussion on GitHub to find all the information and understand why it was added.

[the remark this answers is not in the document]
damnever: I think we've already made a few changes like this. However, I can change it back if you prefer.

store/store.go
```diff
@@ -90,6 +91,7 @@ func (a *Alerts) gc() {
 	}
 	a.Unlock()
 	a.cb(resolved)
```
grobinson-grafana, line 93: @gotjosh I want to remove the callback in a future PR, so I'm not too worried about both returning `resolved` and passing it to the callback.

grobinson-grafana (approved):
LGTM!

provider/mem/mem.go
```diff
@@ -100,37 +101,53 @@ func NewAlerts(ctx context.Context, m types.AlertMarker, intervalGC time.Duratio
...
+		a.callback.PostDelete(&alert)
+	}
+
+	for i, l := range a.listeners {
+		select {
+		case <-l.done:
+			delete(a.listeners, i)
+			close(l.alerts)
+		default:
+			// listener is not closed yet, hence proceed.
```
gotjosh, line 146:
```suggestion
			// Listener is not closed yet, hence proceed.
```
You can address it in the next PR.

gotjosh (approved):
LGTM

gotjosh:

PR #3658, opened 2024-01-03
Support UTF-8 label matchers: Add metrics to matchers compat package

grobinson-grafana:
This commit adds the following metrics to the compat package:

- `alertmanager_matchers_parse`
- `alertmanager_matchers_disagree`
- `alertmanager_matchers_incompatible`
- `alertmanager_matchers_invalid`

With a label called origin to differentiate the different sources of inputs: the configuration file, the API, and amtool.

The disagree_total metric is incremented when an input is invalid in both parsers, but results in different parsed representations, then there is disagreement. This should not happen, and suggests their is either a bug in one of the parsers or a mistake in the backwards compatible guarantees of the matchers/parse parser.

The incompatible_total metric is incremented when an input is valid in pkg/labels, but not the UTF-8 parser in matchers/parse. In such case, the matcher should be updated to be compatible. This often means adding double quotes around the right hand side of the matcher. For example, foo="bar".

The invalid_total metric is incremented when an input is invalid in both parsers. This was never a valid input.

The tests have been updated to check the metrics are incremented as expected.

### DIFF
--- a/matchers/compat/metrics.go
+++ b/matchers/compat/metrics.go
@@ -40,7 +40,7 @@ type Metrics struct {
 func NewMetrics(r prometheus.Registerer) *Metrics {
 	m := &Metrics{
 		Total: promauto.With(r).NewGaugeVec(prometheus.GaugeOpts{
-			Name: "alertmanager_matchers_parsed",
+			Name: "alertmanager_matchers_parse",
 			Help: "Total number of matcher inputs parsed, including invalid inputs.",
 		}, []string{"origin"}),
 		DisagreeTotal: promauto.With(r).NewGaugeVec(prometheus.GaugeOpts{
--- a/matchers/compat/metrics.go
+++ b/matchers/compat/metrics.go
@@ -1,3 +1,16 @@
+// Copyright 2023 The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 package compat
 
 import (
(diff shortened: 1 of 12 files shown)


gotjosh (approved):

matchers/compat/metrics.go
```diff
@@ -0,0 +1,60 @@
...
+			Help: "Total number of matcher inputs which produce different parsings (disagreement).",
+		}, []string{"origin"}),
+		IncompatibleTotal: promauto.With(r).NewGaugeVec(prometheus.GaugeOpts{
+			Name: "alertmanager_matchers_incompatible_total",
+			Help: "Total number of matcher inputs that are incompatible with the UTF-8 parser.",
+		}, []string{"origin"}),
+		InvalidTotal: promauto.With(r).NewGaugeVec(prometheus.GaugeOpts{
+			Name: "alertmanager_matchers_invalid_total",
+			Help: "Total number of matcher inputs that could not be parsed.",
+		}, []string{"origin"}),
```
gotjosh, line 57:
```suggestion
	m := &Metrics{
		Total: promauto.With(r).NewGaugeVec(prometheus.GaugeOpts{
			Name: "alertmanager_matchers_parsed",
			Help: "Total number of matcher inputs parsed, including invalid inputs.",
		}, []string{"origin"}),
		DisagreeTotal: promauto.With(r).NewGaugeVec(prometheus.GaugeOpts{
			Name: "alertmanager_matchers_disagreed",
			Help: "Total number of matcher inputs which produce different parsings (disagreement).",
		}, []string{"origin"}),
		IncompatibleTotal: promauto.With(r).NewGaugeVec(prometheus.GaugeOpts{
			Name: "alertmanager_matchers_incompatible",
			Help: "Total number of matcher inputs that are incompatible with the UTF-8 parser.",
		}, []string{"origin"}),
		InvalidTotal: promauto.With(r).NewGaugeVec(prometheus.GaugeOpts{
			Name: "alertmanager_matchers_invalid",
			Help: "Total number of matcher inputs that could not be parsed.",
		}, []string{"origin"}),
```

gotjosh: These are no longer counters, so the `total` suffix should be removed to avoid confusion.

matchers/compat/metrics.go
```diff
@@ -0,0 +1,60 @@
...
+	Total             *prometheus.GaugeVec
+	DisagreeTotal     *prometheus.GaugeVec
+	IncompatibleTotal *prometheus.GaugeVec
+	InvalidTotal      *prometheus.GaugeVec
+}
+
+func NewMetrics(r prometheus.Registerer) *Metrics {
+	m := &Metrics{
+		Total: promauto.With(r).NewGaugeVec(prometheus.GaugeOpts{
+			Name: "alertmanager_matchers_parse",
```
grobinson-grafana, line 43: Used present tense here to be consistent with the other metrics which are all present tense.

gotjosh: Sgtm!

Issue #3639, opened 2023-12-15
OpsGenie cross-team deduplication - alias prefix
 
Labels: component/notify


kbudde:
OpsGenie has built in deduplication of alerts.
This is a problem for us because we have multiple teams using the same OpsGenie account (different teams).
Deduplication is based on the opsGenie alias field. This field is set to the alertmanager alert key.Hash.
This means that if two teams have the same alert key, they will be deduplicated even if they are different alerts.

As a result, we have alerts which are only sent to one team, but are deduplicated on another team. The same can happen for resolving alerts. If the alert is resolved in the team that did not receive the alert, it will not be resolved in the team that did receive the alert and stay open until it is resolved manually.

One workaround is to configure a prefix for the Promtheus Integration in OpsGenie. This will add the prefix to the alias field. This will prevent deduplication between teams. 
Unfortunately, this does not support updating summary and description fields (config: update_alerts).

Another workaround is to change the "group by" rules in alertmanager and add another label, e.g. team. But this configuration is error prone.

Proposed solution:
Add a new field to the OpsGenie integration configuration in alertmanager. This field will be added to the alias field. This will allow us to configure a prefix for the alias field and prevent deduplication between teams.

It will also allow use cases where the alerts from different monitoring clusters should not be deduplicated. https://github.com/prometheus/alertmanager/issues/1598

kbudde:
Hi, 
The issue can happen in several ways, but the easiest one is probably, 
Two teams share the same opsgenie Organisation, they have both access to the same cluster and monitor the cluster health (e.g. using default set of alerts).

Only one of both teams will see the alerts in opsgenie due to opsgenie dedup.

grobinson-grafana:
> Two teams share the same opsgenie Organisation, they have both access to the same cluster and monitor the cluster health (e.g. using default set of alerts).

So just to make sure I understand 100%. There is an alert (or set of alerts) that are routed to both teams, either via a single route that has a receiver with multiple integrations, or two different routes. For example, something like this:

![issue-3639-example](https://github.com/prometheus/alertmanager/assets/85952834/bd4df0ad-5d48-41d9-972b-aaa025733057)

However, in this case, the alert delivered to both teams in Opsgenie has the same key because they are the same alert. This means for example, that Team 1 will receive the alert and Team 2 will not?

[...]

PR #3661, opened 2024-01-05
Expose opsgenie alias configuration

antonio-tolentino:
This change will help Opsgenie to deduplicate Alerts when it is necessary to send the same alert to multiple teams and each team is using its own integration instead of a global integration.

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -555,6 +555,7 @@ type OpsGenieConfig struct {
 	APIKey       Secret                    `yaml:"api_key,omitempty" json:"api_key,omitempty"`
 	APIKeyFile   string                    `yaml:"api_key_file,omitempty" json:"api_key_file,omitempty"`
 	APIURL       *URL                      `yaml:"api_url,omitempty" json:"api_url,omitempty"`
+	Alias        string                    `yaml:"alias,omitempty" json:"alias,omitempty"`
 	Message      string                    `yaml:"message,omitempty" json:"message,omitempty"`
 	Description  string                    `yaml:"description,omitempty" json:"description,omitempty"`
 	Source       string                    `yaml:"source,omitempty" json:"source,omitempty"`
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -767,6 +767,11 @@ OpsGenie notifications are sent via the [OpsGenie API](https://docs.opsgenie.com
 # The host to send OpsGenie API requests to.
 [ api_url: <string> | default = global.opsgenie_api_url ]
 
+# Alias used by Opsgenie to deduplicate alerts. Limit 512 characters.
+# Custom alias is particularly usefull in situations you have multiple teams to receive the same alert and
+# each one has its own integration instead of global integration.
+[ alias: <string> | default = route.group_by ]
+
 # Alert text limited to 130 characters.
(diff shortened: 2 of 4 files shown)


simonpasquier (commented):
Thanks for the PR. I'm not sure to understand exactly the gap here, do you mind providing more details about your use case?

antonio-tolentino:
Hi @simonpasquier,
Thank you for comment here.

I have a specific situation here that  I need to send the same alert to 2 different teams in Opsgenie. When Alertmanager send alerts to Opsgenie, it uses "group_by" config field to compose the alias. This way Opsgenie will receive 2 occurrences for the same alert because the alias generated is the same and just 1 of the teams will receive this alert. To deduplicate this in Opsgenie would be necessary to include "team name" for example which is not current possible.

If you need more information please let me know.

kbudde:
Hey 🖖,

nice to see more people working with opsgenie and facing the same issues.
I opened  #3640 to address the same issue, this MR is trying to solve :) 

@antonio-tolentino Maybe you can have a look at @3640. I think only allowing the user to add a prefix the alias is the better approach. Alertmanager is already grouping alerts. The alias is following the alertmanager grouping -> You will have the same alerts in Opsgenie. If you allow chaning the labels for grouping for opsgenie, you will get a different grouping of alerts there. 

@simonpasquier I tried to explain the shenanigans of opsgenie in #3639. The alias in opsgenie is used to dedpulicate incoming alerts. Unfortunately, if an alert is coming in it is even deduplicated across different teams. Which leads to undefined behaviour (open alert for team a, but it close for team b). Therefore each team should ensure, that they have unique "alias" values.

Note there is also an open MR to fix missing updates, if the alert is resolved: #3430.

Hope this helps

antonio-tolentino:
Hi @kbudde,

I do believe that if we combine group_by labels and alias_prefix we can definitely deduplicate alerts in Opsgenie.
Please, let's go for this and if you need anything from my side let me know.

Thanks a lot!

kbudde:
Hey, this is what I tried to tell you. You can already do this with the prefix itself. No need to change the grouping labels in the alias. This will lead to surprises.

You have one integration per team as we have it as well. Setting an alias as prefix (e.g. teamA) will Ensure unique alias between teams (no dedup cross team👍) AND keep the same grouping as in alertmanager. 
If you have a different grouping (alias) in opsgenie than in alertmanager, you will run into different amount of open alerts in both systems. So the only allowed change to the alias should be to extend it (fixed prefix or suffix) but not change the grouping labels IMHO.

You can set up a prefix in opsgenie per integration ( needs some license). This works quite nice already. We have it running since a long time like this. It's only an issue if alertmanager tries to update an alert as the prefix is not applied on server side in for update operations.

antonio-tolentino:
If you are thinking in only setup alias per team in Opsgenie, this won't work for us. In our case we currently have more than 150 K8s clusters and many teams per cluster and these figures will increase. The alias(group_by labels + alias_prefix) has to be generated by Alertmanager, in order to get them deduplicated, opened and closed automatically. This will guarantee team A does not close alert from team B and vice-versa and not less important I don't have to setup alias team by team. We do need [#3640](https://github.com/prometheus/alertmanager/pull/3640) code merged.

PR #3668, opened 2024-01-12
Fix inconsistent defaults in UTF-8 behavior

grobinson-grafana:
This commit fixes inconsistent UTF-8 behavior if the compat package is not initialized and feature flags are not passed to the API. This can happen when Alertmanager is used as a package in software such as Cortex or Mimir.

The inconsistent behavior is that Alertmanager will accept UTF-8 alerts but reject UTF-8 configurations.

Since feature flags are optional via api.Options, we cannot force them to be passed to api.New at compile time. Instead, it's better to defer back to the compat package which is consistent even when not initialized.


gotjosh (approved):
LGTM

PR #3673, opened 2024-01-16
Instrument with OTel tracing

### LINKED ISSUE #3670: Feature: Instrument Alertmanager for distributed tracing
It'd be ideal to have Alertmanager instrumented for tracing (with OpenTelemetry), to help when operating it at scale. Prometheus is already instrumented, and so it would make sense to instrument Alertmanager in a similar way.

Once Alertmanager is instrumented, I would also like to add trace propagation over SMTP for e-mail notifications. See [this PR in grafana/grafana](https://github.com/grafana/grafana/pull/80481) for a related approach.

/cc @gotjosh

hairyhenderson:
Fixes #3670

I've pulled in `tracing/tracing.go` from [`prometheus/prometheus`](https://github.com/prometheus/prometheus/tree/main/tracing) to kickstart this, and to provide a familiar (to Prometheus operators) configuration mechanism.

It's possible that it'd be worth pushing some of this up into `prometheus/common`, but I'll defer on that one.

So far, all incoming requests are instrumented, and notifications are instrumented, with webhooks and e-mails getting slightly more detailed instrumentation, both with downward trace propagation.

The decoupled nature of how alerts are handled within Alertmanager means that there's a bunch of disjointed spans, but I've attempted to rectify some of that by using span links.


hairyhenderson:
unclear why CI's failing - I wonder if that test's unrelatedly flaky?

els0r:
I've been wondering about the support of this for months and it's awesome to see movement on this front. Much appreciated!

hairyhenderson:
@els0r you're welcome! Let me know if there are any particular things you want traced - this is really just a start.

tracing/tracing.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+// ApplyConfig takes care of refreshing the tracing configuration by shutting down
+// the current tracer provider (if any is registered) and installing a new one.
+func (m *Manager) ApplyConfig(cfg *config.Config) error {
+	// Update only if a config change is detected. If TLS configuration is
+	// set, we have to restart the manager to make sure that new TLS
+	// certificates are picked up.
+	var blankTLSConfig config_util.TLSConfig
+	if reflect.DeepEqual(m.config, cfg.TracingConfig) && m.config.TLSConfig == blankTLSConfig {
```
qinxx108, line 77: should this be OR?  As if the configuration has been changed, even though tls config is empty, we should restart the manager as well?

hairyhenderson: TBH I'm not sure - this was copied from the Prometheus code - https://github.com/prometheus/prometheus/blob/main/tracing/tracing.go

dispatch/dispatch.go
```diff
@@ -161,20 +167,44 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
 				return
 			}
 
-			level.Debug(d.logger).Log("msg", "Received alert", "alert", alert)
+			// this block is wrapped in a function to make sure that the span
+			// is ended before the next alert is processed
+			func() {
+				traceCtx, span := tracer.Start(d.ctx, "dispatch.Dispatcher.dispatch",
+					trace.WithAttributes(
+						attribute.String("alert.name", alert.Name()),
```
grobinson-grafana, line 175: Compared to the fingerprint, `alert.Name()` is not used much in Alertmanager. We might be able to omit it here as it's a lot of data, and I think most operators have been fine using the fingerprint in the past.

dispatch/dispatch.go
```diff
@@ -161,20 +167,44 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
...
+						attribute.String("alert.fingerprint", alert.Fingerprint().String()),
+						attribute.String("alert.status", string(alert.Status())),
```
grobinson-grafana, line 177: I think it should be mentioned that the status of an alert can change between the dispatcher and when the aggregation group is flushed. If that will affect continuation of the trace, we should remove it.

tracing/tracing.go
```diff
@@ -0,0 +1,233 @@
...
+	otel.SetTextMapPropagator(propagation.TraceContext{})
+	otel.SetErrorHandler(otelErrHandler(func(err error) {
+		level.Error(m.logger).Log("msg", "OpenTelemetry handler returned an error", "err", err)
+	}))
+	<-m.done
+}
+
+// ApplyConfig takes care of refreshing the tracing configuration by shutting down
+// the current tracer provider (if any is registered) and installing a new one.
+func (m *Manager) ApplyConfig(cfg *config.Config) error {
```
grobinson-grafana, line 72: This is not thread-safe with `Stop`. Will it ever happen that two goroutines can call `ApplyConfig` and `Stop` concurrently?

grobinson-grafana (commented):
I took a quick look at the code and left some comments. @hairyhenderson, can you show me how to test this next week so I can better understand how it all works?

config/config.go
```diff
@@ -1053,3 +1056,111 @@ func (m Matchers) MarshalJSON() ([]byte, error) {
...
+
+	if t.Compression != "" && t.Compression != GzipCompression {
+		return fmt.Errorf("invalid compression type %s provided, valid options: %s",
+			t.Compression, GzipCompression)
+	}
+
+	return nil
+}
+
+var reservedHeaders = map[string]struct{}{
```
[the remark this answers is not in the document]
hairyhenderson: Yes - https://github.com/prometheus/prometheus/blob/main/config/config.go#L43

config/config.go
```diff
@@ -1053,3 +1056,111 @@ func (m Matchers) MarshalJSON() ([]byte, error) {
...
+
+	// Added by SigV4.
+	"x-amz-date":           {},
+	"x-amz-security-token": {},
+	"x-amz-content-sha256": {},
+}
+
+func validateHeadersForTracing(headers map[string]string) error {
+	for header := range headers {
+		if strings.ToLower(header) == "authorization" {
```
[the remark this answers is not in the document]
hairyhenderson: This is also copied from `prometheus/prometheus`: https://github.com/prometheus/prometheus/blob/main/config/config.go#L1083

The second quoted sentence refers to the field value, not the name. It's not permissible to have whitespace in names at all (they're just `token`s - i.e. one or more non-control/non-separator character). So even if someone had a configuration with a header that included whitespace, it would likely be ignored or discarded later on.

tracing/tracing.go
hairyhenderson: Apparently not - `Stop` isn't called at all (but I see your other comment suggesting I add a call)...

This was copied from `prometheus/prometheus`, where `ApplyConfig` and `Run` are called concurrently in a run.Group, and `Stop` is called as an interrupt (sequentially). I'll see about safely calling `Stop` in `main.go`...

cmd/alertmanager/main.go
```diff
@@ -474,6 +477,13 @@ func run() int {
 		go disp.Run()
 		go inhibitor.Run()
 
+		err = tracingManager.ApplyConfig(conf)
+		if err != nil {
+			return fmt.Errorf("failed to apply tracing config: %w", err)
+		}
+
+		go tracingManager.Run()
```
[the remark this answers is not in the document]
hairyhenderson: Yep - this was a miss on my part. I'll add that in.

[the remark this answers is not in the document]
hairyhenderson: Added in 3c5cf873b291fe2f97d79c97a3c734822bfbf95c

config/config.go
```diff
@@ -308,6 +309,8 @@ type Config struct {
 	MuteTimeIntervals []MuteTimeInterval `yaml:"mute_time_intervals,omitempty" json:"mute_time_intervals,omitempty"`
 	TimeIntervals     []TimeInterval     `yaml:"time_intervals,omitempty" json:"time_intervals,omitempty"`
 
+	TracingConfig TracingConfig `yaml:"tracing,omitempty" json:"tracing,omitempty"`
```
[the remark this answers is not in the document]
hairyhenderson: That was my first instinct actually, but in `prometheus/prometheus` it's not under `global`. Should we prefer to keep things under `global`, or follow the pattern established in `prometheus/prometheus`?

dispatch/dispatch.go
hairyhenderson: > We might be able to omit it here as it's a lot of data, and I think most operators have been fine using the fingerprint in the past.

isn't it what gets sent as the `alertname` label? surely that's a lot more human-readable than the fingerprint?

dispatch/dispatch.go
hairyhenderson: > If that will affect continuation of the trace, we should remove it.

Not at all - it's just metadata. But if it can change maybe that's misleading, and if we set it at all we should maybe set it at the end when we have the final status 🤔

dispatch/dispatch.go
```diff
@@ -161,20 +167,44 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
 				return
 			}
 
-			level.Debug(d.logger).Log("msg", "Received alert", "alert", alert)
+			// this block is wrapped in a function to make sure that the span
+			// is ended before the next alert is processed
+			func() {
+				traceCtx, span := tracer.Start(d.ctx, "dispatch.Dispatcher.dispatch",
```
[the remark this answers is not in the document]
hairyhenderson: (I thought I already responded to this but I guess my comment got lost somewhere 🤦‍♂️)

It might read nicer, but it'd imply that it's the span for the `run` method, which it isn't really as there may be many of these spans for a single call... maybe something like `handleAlert` would work?

hairyhenderson:
@grobinson-grafana 
> @hairyhenderson, can you show me how to test this next week so I can better understand how it all works?

sorry for the delay... laptop refresh caused me to lose my test configs 🤦‍♂️

I tested this by setting up a local Grafana, Tempo, and Agent (configured to receive OTLP).

Then I fired up the webhook echo service (`go run ./examples/webhook/echo.go` - I actually modified that code first to print headers too so I could see the trace headers)

I then ran `alertmanager` like this:

```console
$ ./alertmanager --web.listen-address=127.0.0.1:9093 --log.level=debug --config.file=am-trace-test.yml
```

Here's `am-trace-test.yml`:

```yaml
global:
  smtp_smarthost: 'localhost:2525'
route:
  group_by: ['alertname']
  group_wait: 30s
  group_interval: 5m
  repeat_interval: 1h
  receiver: 'web.hook'
  routes:
    - matchers:
        - severity="page"
      receiver: web.hook
    - matchers:
        - severity="email"
      receiver: smtp
receivers:
  - name: 'web.hook'
    webhook_configs:
  - name: 'smtp'
    email_configs:
      - to: 'nobody@nowhere.org'
inhibit_rules:
  - source_match:
      severity: 'critical'
    target_match:
      severity: 'warning'

tracing:
  client_type: "grpc"
  endpoint: 'localhost:4317'
  sampling_fraction: 1.0
  insecure: true
```

Then I ran `prometheus` like this:

```console
$ ./prometheus --config.file=prom-alert-tracing-test.yml --web.listen-address=127.0.0.1:9990
```

Here's `prom-alert-tracing-test.yml`:

```yaml
global:
  scrape_interval: 15s
  evaluation_interval: 15s

# Alertmanager configuration
alerting:
  alertmanagers:
    - static_configs:
        - targets:
          - 127.0.0.1:9093

rule_files:
  - prom-alert-tracing-rules.yml

scrape_configs:
  - job_name: "prometheus"
    static_configs:
      - targets: ["localhost:9990"]

tracing:
  client_type: "grpc"
  endpoint: 'localhost:4317'
  sampling_fraction: 1.0
  insecure: true
```

And here's the referenced `prom-alert-tracing-rules.yml`:

```yaml
groups:
  - name: alerts
    rules:
      - alert: AlwaysFiring
        expr: '1'
        labels:
          severity: page
      - alert: AlsoAlwaysFiring
        expr: '1'
        labels:
          severity: email
```

For the `smtp` receiver, I also set up an instance of [`smtprelay`](https://github.com/grafana/smtprelay) and a simple SMTP echo service. I've lost the configs for those, and can whip something up if you need...

Let me know if this makes sense, otherwise let's set up some time tomorrow or next week to pair on this 😉

dispatch/dispatch.go
grobinson-grafana: You could have thousands of alerts all with the same alertname. That's where fingerprint is useful, as it is a unique identifier for an alert.

hairyhenderson: From an operational perspective I think the alert name is still quite valuable... Here's a sample from my test setup:

```
alert.fingerprint | "2a7a97fbf890646a"
alert.name | "AlwaysFiring"
```

I wouldn't be able to relate this trace back to my alert configuration without seeing `AlwaysFiring`...

TheMeier:
IIRC this
```
--- FAIL: TestResolved (5.38s)
    acceptance.go:182: failed to start alertmanager cluster: unable to get a successful response from the Alertmanager: Get "http://127.0.0.1:40783/api/v2/status": dial tcp 127.0.0.1:40783: connect: connection refused
```
is a flaky test, re-running shoud fix it

cmd/alertmanager/main.go
```diff
@@ -474,6 +477,13 @@ func run() int {
 		go disp.Run()
 		go inhibitor.Run()
 
+		err = tracingManager.ApplyConfig(conf)
```
grobinson-grafana, line 480: I'm just documenting this here as I'm not so familiar with this code, this is not feedback or a request for change. This code here is run when the `Coordinator`, which is responsible for reloading configurations at runtime, calls the event handlers registered via the `Subscribe` method. This specific function is responsible for reading the configuration and setting up the receivers, routes, inhibition rules, time intervals, inhibitor, silences, etc.

grobinson-grafana, line 480:
```suggestion
		if err = tracingManager.ApplyConfig(conf); err != nil {
```

cmd/alertmanager/main.go
grobinson-grafana, line 485: This has a goroutine leak because when we reload the configuration, we don't stop the current tracing manager. Instead, we start to accumulate goroutines blocked on `<-m.done`. I was able to see this in `/debug/pprof`.

cmd/alertmanager/main.go
```diff
@@ -534,6 +547,11 @@ func run() int {
 			errc <- configCoordinator.Reload()
 		case <-term:
 			level.Info(logger).Log("msg", "Received SIGTERM, exiting gracefully...")
+
+			// shut down the tracing manager to flush any remaining spans.
```
grobinson-grafana, line 551: I think we also need to shut down on reload?

config/config.go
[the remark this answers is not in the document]
grobinson-grafana: Hmm. It looks like the Prometheus configuration file is structured a little different from Alertmanager as the Prometheus configuration file also has options to change the behavior of the Prometheus server, rather than just contain data like `rules`. However, Alertmanager configuration just contains data, and all behavioral configuration is instead done via command line flags. @gotjosh what should we do here?

config/config.go
```diff
@@ -1053,3 +1056,111 @@ func (m Matchers) MarshalJSON() ([]byte, error) {
 	}
 	return json.Marshal(result)
 }
+
+// TODO: probably move these into prometheus/common since they're copied from
```
grobinson-grafana, line 1060: Seems reasonable to me. Can we do this?

config/config.go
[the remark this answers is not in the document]
grobinson-grafana: Ah – I misread! 🙂

notify/email/email.go
```diff
@@ -123,11 +131,21 @@ func (n *Email) auth(mechs string) (smtp.Auth, error) {
 }
 
 // Notify implements the Notifier interface.
-func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
+func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (ok bool, err error) {
+	ctx, span := tracer.Start(ctx, "email.Email.Notify", trace.WithAttributes(
+		attribute.Int("alerts", len(as)),
```
grobinson-grafana, line 136: Let's move this up attribute up into `dispatch.go` instead?

tracing/http.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptrace"
+
+	"go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
+)
+
+// TODO: maybe move these into prometheus/common?
```
grobinson-grafana, line 26: This looks like good common code to me. Does `prometheus/prometheus` use this too?

dispatch/dispatch.go
```diff
@@ -161,20 +167,44 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
 				return
 			}
 
-			level.Debug(d.logger).Log("msg", "Received alert", "alert", alert)
+			// this block is wrapped in a function to make sure that the span
```
grobinson-grafana, line 170: Let's extract all this logic into a method like `dispatch` or something? This is what I was thinking:

```go
	for {
		select {
		case alert, ok := <-it.Next():
			if !ok {
				// Iterator exhausted for some reason.
				if err := it.Err(); err != nil {
					level.Error(d.logger).Log("msg", "Error on alert update", "err", err)
				}
				return
			}
			d.dispatch(alert)
```

dispatch/dispatch.go
grobinson-grafana: I think as long as we have both then 👍

dispatch/dispatch.go
grobinson-grafana: It can change because the `Status` is calculated using `time.Now()`.

dispatch/dispatch.go
```diff
@@ -161,20 +167,44 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
...
+						attribute.String("alert.status", string(alert.Status())),
+						attribute.String("receiver", d.route.RouteOpts.Receiver),
+					),
+					// we'll use producer here since the alert is not processed
+					// synchronously
+					trace.WithSpanKind(trace.SpanKindProducer),
+				)
+				defer span.End()
+
+				// make a link to this span - we can't make the processAlert
```
grobinson-grafana, line 186: Do we need to be aware of constraints using links here? For example, the aggregation group `run` method can run for minutes/hours/days/weeks/years. What will happen if the linked trace is older than the retention period?

cmd/alertmanager/main.go
grobinson-grafana: OK so the issue is that when the configuration doesn't change, `ApplyConfig` returns `nil`, but we still start another goroutine for `Run`.

tracing/tracing.go
```diff
@@ -0,0 +1,233 @@
...
+	m.shutdownFunc = shutdownFunc
+	m.config = cfg.TracingConfig
+	otel.SetTracerProvider(tp)
+
+	level.Info(m.logger).Log("msg", "Successfully installed a new tracer provider.")
+	return nil
+}
+
+// Stop gracefully shuts down the tracer provider and stops the tracing manager.
+func (m *Manager) Stop() {
```
grobinson-grafana, line 110: There is a race condition between `ApplyConfig` and `Stop` around `m.shutdownFunc` as one goroutine can be updating the variable while another goroutine can be calling the variable. I think we either need a mutex or some docs to mention that these methods are not thread-safe.

config/config.go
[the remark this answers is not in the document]
grobinson-grafana: I read on prometheus-users that if the configuration is expected to be reloadable at runtime it should go in the configuration file, otherwise it should go in command line args.

https://groups.google.com/g/prometheus-users/c/qYKpBiuN8y4/m/9z062yU4CQAJ

notify/email/email.go
hairyhenderson: it's already here actually: https://github.com/prometheus/alertmanager/pull/3673/files#diff-71b3813922f4c2abc260ff5be5e31cf64eeb31eb3d7272f19f4fa889fc3b27d5R93 so it could just be dropped...

stevesg (commented):
I've been testing this change (via vendoring into Mimir) and it looks good to me, definitely something we'd be interested in having. I grabbed a trace of a webhook (an initial connection):

Screenshot from 2024-09-24 13-36-50

jmichalek132:
> I've been testing this change (via vendoring into Mimir) and it looks good to me, definitely something we'd be interested in having. I grabbed a trace of a webhook (an initial connection):
> 
> Screenshot from 2024-09-24 13-36-50

Might be worth considering putting things such as `http.dns`, `http.connect` and `http.tls` as spans events instead of individual spans.

siavashs:
@hairyhenderson can you revive this PR and address all the comments?
If not please let me know so we can look into it.

siavashs:
I assigned this to myself, will rebase and build on top of this in a new PR.

siavashs:
Closing this in favour of #4745 
I tried to address all your suggestions there.
Credits given to @hairyhenderson for their contribution.
Thank you

PR #3680, opened 2024-01-22
Newlines in smtp password file are now ignored

### LINKED ISSUE #3549: Ignore trailing newline in `smtp_auth_password_file`
For what I've seen, most systems ignore a trailing newline in password files, but alertmanger does not.

Since there is no way an smtp password ends in \n, the failure mode is a cryptic smtp auth failure, and alertmanager's behaviour is unique in this, it is a frustrating issue to debug.

I think it would be an improvement if alertmanager ignored trailing newlines when reading `smtp_auth_password_file`.

giorgiga:
This should ignore any newlines in the smtp password file (actually, the file contents are trimmed).

fixed #3549 

**BEWARE** I did not test or even build the project after this single-line change (sorry: I don't know how to do it and honestly can't be bothered to learn how to for a drive-by contribution)

### DIFF
--- a/notify/email/email.go
+++ b/notify/email/email.go
@@ -368,7 +368,7 @@ func (n *Email) getPassword() (string, error) {
 		if err != nil {
 			return "", fmt.Errorf("could not read %s: %w", n.conf.AuthPasswordFile, err)
 		}
-		return string(content), nil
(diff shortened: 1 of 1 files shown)


simonpasquier (approved):
thanks!

PR #3711, opened 2024-02-06
Make metrics in compat package optional

grobinson-grafana:
This commit updates the compat package to make the metrics optional. This allows other projects that use this package, such as Cortex and Mimir, to opt-in and out of the metrics.


gotjosh:
There are simpler ways to do this, depending on how you're planning on using the "no metrics" mode.

If your thinking is that you do something like the following in the importing project:

```golang
var RegisteredMetrics = nil
```

You can replace it with:

```golang
var RegisteredMetrics = NewMetrics(nil)
```

and get rid of all the `nil` checks, `promauto.With(r)` will take care of the nil checks for you.

Issue #3682, opened 2024-01-24
Alert manager possible dead lock 


qinxx108:
### Description ###
Recently we noticed a dead lock issue happened in our running alert managers.
Reference:  
* dispather: https://github.com/prometheus/alertmanager/blob/main/dispatch/dispatch.go#L143
* mem.Alerts.Subscribe: https://github.com/prometheus/alertmanager/blob/1da134aa30c81e656c5156df1499a77d5df92269/provider/mem/mem.go#L151
* claim store.Alerts.mtx: https://github.com/prometheus/alertmanager/blob/1da134aa30c81e656c5156df1499a77d5df92269/provider/mem/mem.go#L157
https://github.com/prometheus/alertmanager/blob/1da134aa30c81e656c5156df1499a77d5df92269/store/store.go#L119
* gc: https://github.com/prometheus/alertmanager/blob/1da134aa30c81e656c5156df1499a77d5df92269/store/store.go#L72
* mem.gc.callback: https://github.com/prometheus/alertmanager/blob/1da134aa30c81e656c5156df1499a77d5df92269/provider/mem/mem.go#L112

![image](https://github.com/prometheus/alertmanager/assets/63399121/9012c552-4f58-4593-bfe4-5f08937fd4b7)

![image](https://github.com/prometheus/alertmanager/assets/63399121/9cbdfca4-bdab-4dc9-8b6c-169fd3704ebb)

### Proposal fix ###
* Have subscribe function here https://github.com/prometheus/alertmanager/blob/1da134aa30c81e656c5156df1499a77d5df92269/provider/mem/mem.go#L151 to claim store.Alerts.mux first before claim mem.Alerts.mux

yeya24:
Is this something we can close now with #3715 getting merged?

gotjosh:
Indeed, thanks @yeya24

PR #3715, opened 2024-02-08
Fix for deadlock issue #3682

rajagopalanand:
### What does this PR do?

This PR is a fix for #3682. In some instances, `mem.Alerts.Subscribe()` and `store.gc()` can get deadlocked

1. Lock acquisition in `store.Alerts.gc()`:
      - The method `store.Alerts.gc()` acquires a lock on its internal mutex `store.Alerts.mtx`
      - While holding the lock, it then calls the callback function which will try to acquire a lock on `mem.Alerts.mtx`
2. Concurrent Execution:
      -  `mem.Alerts.Subscribe()` acquires a lock on its internal mutex `mem.Alerts.mtx` and calls `store.Alerts.List()`
3. Deadlock situation:
      - Callback function tries to acquire a lock on `mem.Alerts.mtx`. However this lock is already being held by `mem.Subscribe()`
      - Similarly `mem.Subscribe()` cannot proceed because `store.List()` cannot acquire lock (`store.Alerts.mtx`) because it is being held by `store.gc()`

Another way of summarizing this is `store.Alerts.gc()` was holding the lock until callback function completed which in turn was waiting to acquire the lock. Callback function could not acquire the lock because `Subscribe()` was holding the lock. `Subscribe()` cannot progress because it calls `store.Alerts.List()` which was waiting for lock acquisition which was being held by `store.Alerts.gc()`. This fix releases the lock held by `store.Alerts.gc()` prior to calling the callback function

## 

AM_Deadlock

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -71,8 +71,6 @@ func (a *Alerts) Run(ctx context.Context, interval time.Duration) {
 
 func (a *Alerts) gc() {
 	a.Lock()
-	defer a.Unlock()
-
 	var resolved []*types.Alert
 	for fp, alert := range a.c {
 		if alert.Resolved() {
@@ -80,6 +78,7 @@ func (a *Alerts) gc() {
 			resolved = append(resolved, alert)
 		}
 	}
+	a.Unlock()
 	a.cb(resolved)
 }
 
--- a/provider/mem/mem.go
+++ b/provider/mem/mem.go
@@ -151,7 +151,6 @@ func max(a, b int) int {
 func (a *Alerts) Subscribe() provider.AlertIterator {
 	a.mtx.Lock()
 	defer a.mtx.Unlock()
-
 	var (
 		done   = make(chan struct{})
 		alerts = a.alerts.List()
--- a/provider/mem/mem_test.go
+++ b/provider/mem/mem_test.go
@@ -135,6 +135,63 @@ func TestAlertsSubscribePutStarvation(t *testing.T) {
 	}
 }
 
+func TestDeadLock(t *testing.T) {
+	t0 := time.Now()
+	t1 := t0.Add(5 * time.Second)
+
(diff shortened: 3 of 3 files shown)


rapphil (approved):

emanlodovice (approved):

gotjosh:
@rajagopalanand thank you very much for your contribution - can you please fix the linter?

gotjosh:
This is a duplciate of https://github.com/prometheus/alertmanager/pull/3648/files, right?

rajagopalanand:
> This is a duplicate of https://github.com/prometheus/alertmanager/pull/3648/files, right?

Was not aware of this but I will take a look

rajagopalanand:
> > This is a duplicate of https://github.com/prometheus/alertmanager/pull/3648/files, right?
> 
> Was not aware of this but I will take a look

The other [PR](https://github.com/prometheus/alertmanager/pull/3648) addresses issues more than just the deadlock issue mentioned in #3682. If that gets reviewed and merged, then this PR can be discarded

store/store.go
```diff
@@ -71,15 +71,14 @@ func (a *Alerts) Run(ctx context.Context, interval time.Duration) {
 
 func (a *Alerts) gc() {
 	a.Lock()
-	defer a.Unlock()
-
 	var resolved []*types.Alert
```
gotjosh, line 74: We need to stop passing `[]*types.Alert` and instead now send a slice of values to the callback.

```suggestion
	var resolved []types.Alert
```

gotjosh (commented):
Thank you so much for the detailed response and PR description.

This LGTM, but there's one more thing we need to do. Please take a look at my comment.

store/store.go
rajagopalanand: I can submit a separate PR for this

PR #3718, opened 2024-02-13
Fix a small number of inconsistencies in compat package logging

grobinson-grafana:
This commit fixes a small number of inconsistencies in the compat package logging. It now has consistent use of classic matchers parser and UTF-8 matchers parser, instead of old matchers parser and new matchers parser.


gotjosh (approved):
LGTM

PR #3723, opened 2024-02-15
Notify: Several improvements around the `suppressed_total` metric and…

gotjosh:
… logs for muting.

- Fix the description of the metric to include the fact that the metric is also incremented when the alert is silenced or inhibited.
- Make the reason constants public so that they can be-reused in projects that consume the Alertmanager as a library
- Add the reason to the debug line for muting.


grobinson-grafana (approved):
LGTM!

Issue #3735, opened 2024-02-25
Alertmanager should be cgroup aware


uhthomas:
**What did you do?**

I recently changed the CPU in my server from an Intel 13600k with 20 threads to an EPYC 7763 with 128 threads. I run alertmanager in Kubernetes with a CPU limit of [100m](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-cpu).

**What did you expect to see?**

No change in performance.

**What did you see instead? Under which circumstances?**

https://github.com/golang/go/issues/33803. This means that Prometheus is being throttled in containerised environments with CPU limits.

The first half of the time series is with the 13600k, and the second half is with the EPYC 7763.

![image](https://github.com/prometheus/alertmanager/assets/9749173/060b2b09-3002-48bf-ab46-465ad378ece4)

**Environment**

* System information:

```
Linux 6.1.69-talos x86_64
```

* Alertmanager version:

```
alertmanager, version 0.26.0 (branch: main, revision: 7a3c189315da7b8ed2d5e05f122c4cf4873c1379)
  build user:       root@9178099891ea
  build date:       20240123-14:29:39
  go version:       go1.21.6
  platform:         linux/amd64
  tags:             netgo
```

* Prometheus version:

N/A

* Alertmanager configuration file:

N/A

* Prometheus configuration file:

N/A

* Logs:

N/A

siavashs:
Fixed in #3837

uhthomas:
Ironically, automaxprocs should probably be removed now that Go 1.25 does this natively.

siavashs:
Alertmanager is not on Go 1.25 yet, we'll revisit this when we move to 1.25

PR #3736, opened 2024-02-25
chore: use Linux container CPU quota

uhthomas:
Go is not cgroup aware and by default will set GOMAXPROCS to the number of available threads, regardless of whether it is within the allocated quota. This behaviour causes high amount of CPU throttling and degraded application performance.

Fixes: #3735

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -41,6 +41,7 @@ import (
 	"github.com/prometheus/common/version"
 	"github.com/prometheus/exporter-toolkit/web"
 	webflag "github.com/prometheus/exporter-toolkit/web/kingpinflag"
+	_ "go.uber.org/automaxprocs"
 
 	"github.com/prometheus/alertmanager/api"
 	"github.com/prometheus/alertmanager/cluster"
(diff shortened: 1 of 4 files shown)


TheMeier:
Can you make the output `maxprocs: Updating GOMAXPROCS=2: determined from CPU quota` getting put in the logs instead of stdout?
https://github.com/uber-go/automaxprocs/blob/v1.5.3/automaxprocs.go#L32

roidelapluie:
LGTM, but I would prefer to see this behind a flag

TheMeier:
> Can you make the output `maxprocs: Updating GOMAXPROCS=2: determined from CPU quota` getting put in the logs instead of stdout? https://github.com/uber-go/automaxprocs/blob/v1.5.3/automaxprocs.go#L32

D'oh stdout is the log, right? So basically just adapt the tests?

gotjosh:
Duplicate of https://github.com/prometheus/alertmanager/pull/3736

uhthomas:
> Duplicate of https://github.com/prometheus/alertmanager/pull/3736

Wrong link? You've just marked it as a duplicate of itself.

PR #3737, opened 2024-02-26
feature: add optional native histogram support to latency metrics

krajorama:
Note that this does not stop showing classic metrics, for now it is up to the scrape config to decide whether to keep those.


krajorama:
Tests fail on main as well.

gotjosh (approved):
LGTM

PR #3770, opened 2024-03-20
add sprig template functions

### LINKED ISSUE #603: Accessing current date / other functions in the Alertmanager Go templates
We're using AlertManager to send out emails. 

**Question 1 :** 
Emails apps like Inbox group emails to create conversations based on the email title. Our current email title is something like "X services failing" so that makes for horrible grouping / conversations.

The requirement would now be to group them per day by including the date in the subject. 

However, I'm not able to access the current date in the Go Templates. 

Is there another way of doing this ? 

- Can it be passed on as a label by prometheus (using the alert.rules) ? 
- Can something be done in the alertmanager configuration ?

**Question 2 :** 
As a general remark , we often get requirements to change the email templates. For example : 

- Change the sorting of the alerts
- Add how long an alert was active (current date - startsAt)
- ...

These seem to be very difficult to implement using the Go templates as a lot of these functions aren't available (no sorting, no access to date functons / current date, ....)

How should we go about this ?

### LINKED ISSUE #3726: Add extensive templating helpers based on Masterminds/sprig
When writing custom templates, the current set of functions is fairly limited.

We could register the functions from https://github.com/Masterminds/sprig to offer the template authors an extensive set of helper functions. Docs of the functions is available here: https://masterminds.github.io/sprig/

If you are interested, I would be happy to send a patch.

### LINKED ISSUE #3762: Pass Go's `time` function into the alertmanager templates to be able to do time calculations
### Proposal

Currently, it's not possible to do time calculations when sending alerts to Slack for example. Explicitly converting to another time zone or calculating a duration between `startsAt` and now would be a tremendous help. There is nothing on Slack/Mattermost's side to do any conversions or calculations there.

This would be easily solved by passing the default `time` functions into the template at this location.
https://github.com/prometheus/alertmanager/blob/680568b518ddd728f46101952e006e7b13910901/template/template.go#L170

TheMeier:
fixes #3726
fixes #3762
fixes #603

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -21,11 +21,13 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"sort"
 	"strings"
 	tmpltext "text/template"
 	"time"
 
+	"github.com/Masterminds/sprig/v3"
 	commonTemplates "github.com/prometheus/common/helpers/templates"
 	"github.com/prometheus/common/model"
 	"golang.org/x/text/cases"
@@ -59,8 +61,8 @@ func New(options ...Option) (*Template, error) {
 		o(t.text, t.html)
 	}
 
-	t.text.Funcs(tmpltext.FuncMap(DefaultFuncs))
-	t.html.Funcs(tmplhtml.FuncMap(DefaultFuncs))
+	t.text.Funcs(tmpltext.FuncMap(DefaultFuncs)).Funcs(tmpltext.FuncMap(filteredSprigFuncMap()))
+	t.html.Funcs(tmplhtml.FuncMap(DefaultFuncs)).Funcs(tmplhtml.FuncMap(filteredSprigFuncMap()))
 
 	return t, nil
 }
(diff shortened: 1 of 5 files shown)


TheMeier:
Looks like circleci has a problem :(

TheMeier:
can I get a review @gotjosh

template/template.go
```diff
@@ -58,8 +59,8 @@ func New(options ...Option) (*Template, error) {
 		o(t.text, t.html)
 	}
 
-	t.text.Funcs(tmpltext.FuncMap(DefaultFuncs))
-	t.html.Funcs(tmplhtml.FuncMap(DefaultFuncs))
+	t.text.Funcs(tmpltext.FuncMap(DefaultFuncs)).Funcs(tmpltext.FuncMap(sprig.FuncMap()))
```
grobinson-grafana, line 62: You need to block functions like [`os`](https://masterminds.github.io/sprig/os.html) so people can't read environment variables.

grobinson-grafana (changes requested):

template/template.go
```diff
@@ -58,8 +59,11 @@ func New(options ...Option) (*Template, error) {
 		o(t.text, t.html)
 	}
 
-	t.text.Funcs(tmpltext.FuncMap(DefaultFuncs))
-	t.html.Funcs(tmplhtml.FuncMap(DefaultFuncs))
+	sprigFuncMap := sprig.FuncMap()
+	delete(sprigFuncMap, "env")
```
grobinson-grafana, line 63: I had another thought about how we can make this even safer. Instead of deleting functions from the map, let's declare our own map of Sprig functions that we know are safe.

The benefit here is if new, un-safe functions are added to Sprig, then we cannot introduce them accidentally by just updating the package in go.mod. Instead, we will need to add them to the allowlist of functions.

What do you think?

grobinson-grafana: I think having some kind of safe func map will be critical to merging this PR. I don't think it matters whether that func map is in Alertmanager or in Sprig though.

The reason is that Alertmanager is also used in multi-tenanted systems such as Mimir, in Cloud Managed environments. Bad actors could use Alertmanager templates to attempt to read secrets from environment variables, or traverse internal services using `getHostByName`.

TheMeier: @grobinson-grafana Thanks for this insightful feedback. I was not aware of thoses aspects. I will rethink this and rework the MR. 
Also, in the meantime, I had a look at the sprig project. The chances of getting a change like this merged there are slim to none.
So either I will still include sprig but just select dedicated functions or I consider duplicating stuff. There is `HermeticHtmlFuncMap()` which looks promising.

TheMeier: Changed to only allow functions explicitly listed.

TheMeier:
Also I noticed the code is actually prepared to allow for users to pass custom functions to the templating. 
https://github.com/prometheus/alertmanager/blob/main/template/template.go#L68C23-L70
https://github.com/prometheus/alertmanager/blob/main/template/template_test.go#L406-L468

But that is not accessible from what I have seen in code and docs. 
If it is accessible is even a much bigger issue if you consider users might use an alertmanager provided by someone else, since the user could execute harmful code. So I wonder if `option` for `template.New` and `template.FromGlobs` should be removed.

grobinson-grafana:
> Also I noticed the code is actually prepared to allow for users to pass custom functions to the templating. https://github.com/prometheus/alertmanager/blob/main/template/template.go#L68C23-L70 https://github.com/prometheus/alertmanager/blob/main/template/template_test.go#L406-L468
> 
> But that is not accessible from what I have seen in code and docs. If it is accessible it is a bigger issue if you consider users might use an alertmanager provided by someone else, since the user could execute harmful code. So I wonder if `option` for `template.New` and `template.FromGlobs` should be removed.

This is all internal code. Users cannot add their own functions without changing and re-compiling the code.

verdel:
@grobinson-grafana, @TheMeier, how can we speed up the review of this PR?

The functionality to be added, if this PR is accepted, will eliminate a large number of existing limitations. For example, time-related functions will allow for creating links to Grafana dashboards with a time range tied to the time the alert was triggered.

TheMeier:
> @grobinson-grafana, @TheMeier, how can we speed up the review of this PR?
> 
> The functionality to be added, if this PR is accepted, will eliminate a large number of existing limitations. For example, time-related functions will allow for creating links to Grafana dashboards with a time range tied to the time the alert was triggered.

The only ones who can help here are the maintainers. I guess they are busy with other stuff, so you just have to wait

verdel:
> The only ones who can help here are the maintainers. I guess they are busy with other stuff, so you just have to wait

Yep, I'll be hoping that they can find the time. This feature would indeed be very useful.

grobinson-grafana:
While we figure out whether or not to accept sprig, I've opened a PR that adds much requested date and timezone functions [here](https://github.com/prometheus/alertmanager/pull/3812).

grobinson-grafana:
I've spent some time thinking about this and my preference is that we don't add Sprig. Instead, we implement template functions as needed. We can also implement the most common functions in `prometheus/common` so they can be used in both Prometheus and Alertmanager.

The reasons I don't think we should add Sprig are the following:

1. We need to maintain an allow list of functions that we consider `safe`, and disable functions that do filesystem operations, network calls, etc.
2. Sprig adds 11 new dependencies to Alertmanager. I would like to reduce the number of dependencies we have over time, not increase them.
3. We need to track Sprig for updates and vet changes to existing functions.
4. If Sprig is ever deprecated or archived we need to implement every single function ourselves anyway, because users now depend on them.

TheMeier:
@grobinson-grafana your points are all valid, but it is missing a bit of context IMHO:
* sprig releases 1-2 times per year (actually no release at all since November 2022)
* sprig is **extremely** conservative with any changes 
* the logic about deprecation or archiving does apply to each and every dependency you will ever use. so by that logic, you never would use any prior art ;)

The 11 extra dependencies are a very good reason not to go this route though.
Given many, if not most, functions are trivial it should be not a lot of work.

PR #3773, opened 2024-03-21
Add more benchmarks for inhibition rules

grobinson-grafana:
This commit adds more benchmarks for inhibition rules where just the last rule in the benchmark inhibits the labels. Here is an example of these benchmarks:

```
BenchmarkMutes/10_inhibition_rules,_last_rule_matches-8        	 1094041	      1020 ns/op
BenchmarkMutes/100_inhibition_rules,_last_rule_matches-8       	  347079	      3306 ns/op
BenchmarkMutes/1000_inhibition_rules,_last_rule_matches-8      	   39068	     26769 ns/op
BenchmarkMutes/10000_inhibition_rules,_last_rule_matches-8     	    4387	    260687 ns/op
```

### DIFF
--- a/inhibit/inhibit_bench_test.go
+++ b/inhibit/inhibit_bench_test.go
@@ -197,6 +197,7 @@ func benchmarkMutes(b *testing.B, opts benchmarkOptions) {
 			b.Fatal(err)
 		}
 	}
+
 	ih := NewInhibitor(s, rules, m, log.NewNopLogger())
 	defer ih.Stop()
 	go ih.Run()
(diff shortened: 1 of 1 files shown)


inhibit/inhibit_bench_test.go
```diff
@@ -142,17 +197,12 @@ func benchmarkMutes(b *testing.B, opts benchmarkOptions) {
 			b.Fatal(err)
 		}
 	}
-
 	ih := NewInhibitor(s, rules, m, log.NewNopLogger())
 	defer ih.Stop()
 	go ih.Run()
 
 	// Wait some time for the inhibitor to seed its cache.
-	waitDuration := time.Millisecond * time.Duration(len(alerts))
```
grobinson-grafana, line 151: This didn't work when there was just 1 alert rule, `1ms` was too short.

gotjosh (approved):
LGTM

PR #3776, opened 2024-03-26
Support http.Headers in util functions

grobinson-grafana:
This commit updates the util functions such as PostJSON to support additional headers using http.Header. It will mean integrations such as RocketChat can use the notify functions instead of http.NewRequest.


notify/pushover/pushover.go
```diff
@@ -164,7 +164,7 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 	u.RawQuery = parameters.Encode()
 	// Don't log the URL as it contains secret data (see #1825).
 	level.Debug(n.logger).Log("msg", "Sending message", "incident", key)
-	resp, err := notify.PostText(ctx, n.client, u.String(), nil)
+	resp, err := notify.PostText(ctx, n.client, u.String(), nil, nil)
```
zecke, line 167: That is a lot of "nil" for a caller. What if one just passes the *http.Request to a function like notify.Do?

PR #3777, opened 2024-03-27
Rename matchers package to matcher singular

grobinson-grafana:
I realized that we had named the package plural `matchers` when its idiomatic in Go to use singular package names.


simonpasquier (approved):
Thanks!

PR #3791, opened 2024-04-02
#3513: TimeMuter returns the names of time intervals

grobinson-grafana:
This commit updates the TimeMuter interface to also return the names of the time intervals that muted the alerts.

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -949,7 +949,7 @@ func (tms TimeMuteStage) Exec(ctx context.Context, l log.Logger, alerts ...*type
 		return ctx, alerts, nil
 	}
 
-	muted, err := tms.muter.Mutes(muteTimeIntervalNames, now)
+	muted, _, err := tms.muter.Mutes(muteTimeIntervalNames, now)
 	if err != nil {
 		return ctx, alerts, err
 	}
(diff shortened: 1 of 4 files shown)


notify/notify.go
```diff
@@ -949,7 +949,7 @@ func (tms TimeMuteStage) Exec(ctx context.Context, l log.Logger, alerts ...*type
 		return ctx, alerts, nil
 	}
 
-	muted, err := tms.muter.Mutes(muteTimeIntervalNames, now)
+	muted, _, err := tms.muter.Mutes(muteTimeIntervalNames, now)
```
grobinson-grafana, line 952: We do not have `GroupMarker` (added [here](https://github.com/prometheus/alertmanager/pull/3792)) so there is nothing to be marked. Use the blank identifier here instead.

timeinterval/timeinterval.go
```diff
@@ -33,21 +33,22 @@ type Intervener struct {
 	intervals map[string][]TimeInterval
 }
 
-func (i *Intervener) Mutes(names []string, now time.Time) (bool, error) {
+func (i *Intervener) Mutes(names []string, now time.Time) (bool, []string, error) {
```
gotjosh, line 36:
```suggestion
// Mutes implements the TimeMuter interface.
func (i *Intervener) Mutes(names []string, now time.Time) (bool, []string, error) {
```

gotjosh (approved):
LGTM

PR #3792, opened 2024-04-02
#3513: Add GroupMarker interface

grobinson-grafana:
This commit adds a new GroupMarker interface that marks the status of groups. For example, whether a group is muted because or one or more active or mute time intervals. It renames the existing Marker interface to AlertMarker to avoid confusion.

It is based on https://github.com/prometheus/alertmanager/pull/3791.

### DIFF
--- a/types/types.go
+++ b/types/types.go
@@ -52,12 +52,12 @@ type AlertStatus struct {
 	silencesVersion int
 }
 
-// groupStatus stores the state of the group, and, as applicable, the fingerprints
-// of all muted alerts.
+// groupStatus stores the state of the group, and, as applicable, the names
+// of all active and mute time intervals that are muting it.
 type groupStatus struct {
-	// mutedAlerts contains the fingerprints of all muted alerts, and for each
-	// muted alert the active or mute time intervals that muted it.
-	mutedAlerts map[model.Fingerprint][]string
+	// mutedBy contains the names of all active and mute time intervals that
+	// are muting it.
+	mutedBy []string
 }
 
 // AlertMarker helps to mark alerts as silenced and/or inhibited.
@@ -103,7 +103,7 @@ type AlertMarker interface {
 type GroupMarker interface {
 	// Muted returns true if the alert is muted, otherwise false. If the alert
 	// is muted then it also returns the names of the time intervals that muted
-	// it. If the alert is not muted then the list is nil.
+	// it.
 	Muted(groupKey string, fingerprint model.Fingerprint) ([]string, bool)
 
 	// SetMuted marks the alert as muted, and sets the names of the time
@@ -130,44 +130,26 @@ type MemMarker struct {
 }
 
 // Muted implements GroupMarker.
-func (m *MemMarker) Muted(groupKey string, fingerprint model.Fingerprint) ([]string, bool) {
+func (m *MemMarker) Muted(groupKey string) ([]string, bool) {
 	m.mtx.Lock()
 	defer m.mtx.Unlock()
-	groupStatus, ok := m.groups[groupKey]
+	status, ok := m.groups[groupKey]
 	if !ok {
 		return nil, false
 	}
-	timeIntervalsNames, ok := groupStatus.mutedAlerts[fingerprint]
-	if !ok {
-		return nil, false
-	}
-	return timeIntervalsNames, ok
+	return status.mutedBy, len(status.mutedBy) > 0
 }
 
 // SetMuted implements GroupMarker.
-func (m *MemMarker) SetMuted(groupKey string, fingerprint model.Fingerprint, timeIntervalNames []string) {
+func (m *MemMarker) SetMuted(groupKey string, timeIntervalNames []string) {
 	m.mtx.Lock()
 	defer m.mtx.Unlock()
-	if len(timeIntervalNames) > 0 {
-		m.setMuteMarker(groupKey, fingerprint, timeIntervalNames)
-	} else {
-		m.removeMuteMarker(groupKey, fingerprint)
-	}
-}
-
-func (m *MemMarker) setMuteMarker(groupKey string, fingerprint model.Fingerprint, timeIntervalNames []string) {
 	status, ok := m.groups[groupKey]
 	if !ok {
-		status = &groupStatus{mutedAlerts: make(map[model.Fingerprint][]string)}
+		status = &groupStatus{}
 		m.groups[groupKey] = status
 	}
-	status.mutedAlerts[fingerprint] = timeIntervalNames
-}
-
-func (m *MemMarker) removeMuteMarker(groupKey string, fingerprint model.Fingerprint) {
-	if status, ok := m.groups[groupKey]; ok {
-		delete(status.mutedAlerts, fingerprint)
-	}
+	status.mutedBy = timeIntervalNames
 }
 
 func (m *MemMarker) registerMetrics(r prometheus.Registerer) {
--- a/types/types.go
+++ b/types/types.go
@@ -104,12 +104,12 @@ type GroupMarker interface {
 	// Muted returns true if the alert is muted, otherwise false. If the alert
 	// is muted then it also returns the names of the time intervals that muted
 	// it.
-	Muted(groupKey string, fingerprint model.Fingerprint) ([]string, bool)
+	Muted(groupKey string) ([]string, bool)
 
 	// SetMuted marks the alert as muted, and sets the names of the time
 	// intervals that mute it. If the list of names is nil or the empty slice
 	// then the muted marker is removed.
-	SetMuted(groupKey string, fingerprint model.Fingerprint, timeIntervalNames []string)
+	SetMuted(groupKey string, timeIntervalNames []string)
 }
 
 // NewMarker returns an instance of a AlertMarker implementation.
--- a/types/types.go
+++ b/types/types.go
@@ -101,12 +101,12 @@ type AlertMarker interface {
 }
 
 type GroupMarker interface {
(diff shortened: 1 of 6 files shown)


types/types.go
```diff
@@ -92,24 +100,59 @@ type Marker interface {
...
-// NewMarker returns an instance of a Marker implementation.
-func NewMarker(r prometheus.Registerer) Marker {
-	m := &memMarker{
-		m: map[model.Fingerprint]*AlertStatus{},
-	}
+type GroupMarker interface {
+	// Muted returns true if the group is muted, otherwise false. If the group
+	// is muted then it also returns the names of the time intervals that muted
+	// it.
+	Muted(groupKey string) ([]string, bool)
```
grobinson-grafana, line 107: In the original version of `GroupMarker` I had:

```go
type GroupMarker interface {
	// Muted returns true if the group is muted, otherwise false. If the group
	// is muted then it also returns the names of the time intervals that muted
	// it.
	Muted(groupKey string, fingerprint model.Fingerprint) ([]string, bool)
	...
}
```

but then I realized we didn't need to store the fingerprint at all.

The reason here is that active and mute timings work against routes. That means either all alerts in an aggregation group are suppressed because of time intervals, or none of them are. However, what is not possible is having two alerts in the same aggregation group A and B, where A is muted from a mute time interval and B is not.

grobinson-grafana, line 107: In the original version of `GroupMarker` I had:

```go
type GroupMarker interface {
	// Muted returns true if the group is muted, otherwise false. If the group
	// is muted then it also returns the names of the time intervals that muted
	// it.
	Muted(groupKey string, fingerprint model.Fingerprint) ([]string, bool)
	...
}
```

but then I realized we didn't need to store the fingerprint at all.

The reason here is that active and mute timings work against routes. That means either all alerts in an aggregation group are suppressed because of time intervals, or none of them are. It is not possible to have two alerts in the same aggregation group A and B, where A is muted from a mute time interval and B is not.

grobinson-grafana, line 107: In the original version of `GroupMarker` I had:

```go
type GroupMarker interface {
	// Muted returns true if the group is muted, otherwise false. If the group
	// is muted then it also returns the names of the time intervals that muted
	// it.
	Muted(groupKey string, fingerprint model.Fingerprint) ([]string, bool)
	...
}
```

but then I realized we didn't need to store the fingerprint at all.

The reason here is that active and mute timings work against routes. That means either all alerts in an aggregation group are suppressed because of time intervals, or none of them are. It is not possible to have two alerts A and B in the same aggregation group, where A is muted from a mute time interval and B is not.

grobinson-grafana, line 107: In the original version of `GroupMarker` I had:

```go
type GroupMarker interface {
	// Muted returns true if the group is muted, otherwise false. If the group
	// is muted then it also returns the names of the time intervals that muted
	// it.
	Muted(groupKey string, fingerprint model.Fingerprint) ([]string, bool)
	...
}
```

but then I realized we didn't need to store the fingerprint at all.

The reason here is that active and mute timings work against routes. That means either all alerts in an aggregation group are suppressed because of time intervals, or none of them are. It is not possible to have two alerts A and B in the same aggregation group, where A is muted from a mute time interval and B is not.

grobinson-grafana: We have found an issue where because `groupKey` is not guaranteed to be unique, it's possible for two (or more) different groups to have the same `groupKey`.  In such cases, where one group is muted and another group is not, both will be marked as muted.

grobinson-grafana: The following YAML shows a configuration containing two routes with the same `groupKey`. This happens because the `groupKey` is calculated using:

1. The matchers in the route
2. The `group_by` labels (default is empty)

And both routes have the same matchers and `group_by` labels.

```yml
receivers:
  - name: test1
  - name: test2
route:
  receiver: test1
  routes:
    - receiver: test1
      matchers:
        - foo=bar
    - receiver: test2
      matchers:
        - foo=bar
      mute_time_intervals:
        - name: weekends
```

The reason a user might have such a configuration is to mute notifications on the weekends, but still send webhooks to an issue tracker.

types/types.go
```diff
@@ -92,24 +100,65 @@ type Marker interface {
...
 
-// NewMarker returns an instance of a Marker implementation.
-func NewMarker(r prometheus.Registerer) Marker {
-	m := &memMarker{
-		m: map[model.Fingerprint]*AlertStatus{},
-	}
+// GroupMarker helps to mark groups as active or muted.
+// All methods are goroutine-safe.
+//
+// TODO(grobinson): routeID is used in Muted and SetMuted because groupKey
```
grobinson-grafana, line 106: I added a TODO in the source code to make it clear that this is not how I want the interface to look, but fixing this is blocked due to #3817 that we will fix in another PR.

types/types.go
```diff
@@ -52,9 +52,17 @@ type AlertStatus struct {
...
+type groupStatus struct {
+	// mutedBy contains the names of all active and mute time intervals that
+	// are muting it.
+	mutedBy []string
+}
+
+// AlertMarker helps to mark alerts as silenced and/or inhibited.
 // All methods are goroutine-safe.
-type Marker interface {
+type AlertMarker interface {
```
grobinson-grafana, line 65: I renamed this interface to `AlertMarker` so it could be better differentiated from `GroupMarker`.

types/types.go
```diff
@@ -92,24 +100,65 @@ type Marker interface {
...
 
 	mtx sync.RWMutex
 }
 
-func (m *memMarker) registerMetrics(r prometheus.Registerer) {
+// Muted implements GroupMarker.
+func (m *MemMarker) Muted(routeID, groupKey string) ([]string, bool) {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+	status, ok := m.groups[routeID+groupKey]
```
grobinson-grafana, line 142: Since `routeID` will be removed in the future I have chosen to just concatenate the two strings together,

grobinson-grafana, line 142: Since `routeID` will be removed in the future I have chosen to just concatenate the two strings together.

types/types.go
```diff
@@ -92,24 +100,65 @@ type Marker interface {
...
+// All methods are goroutine-safe.
+//
+// TODO(grobinson): routeID is used in Muted and SetMuted because groupKey
+// is not unique (see #3817). Once groupKey uniqueness is fixed routeID can
+// be removed from the GroupMarker interface.
+type GroupMarker interface {
+	// Muted returns true if the group is muted, otherwise false. If the group
+	// is muted then it also returns the names of the time intervals that muted
+	// it.
+	Muted(routeID, groupKey string) ([]string, bool)
```
grobinson-grafana, line 113: I added `routeID` to avoid the issue of non unique group keys from causing groups to be incorrectly marked as muted. See https://github.com/prometheus/alertmanager/issues/3817 for more information.

types/types.go
```diff
@@ -52,9 +52,17 @@ type AlertStatus struct {
 	silencesVersion int
 }
 
-// Marker helps to mark alerts as silenced and/or inhibited.
+// groupStatus stores the state of the group, and, as applicable, the names
+// of all active and mute time intervals that are muting it.
```
gotjosh, line 56: This claims to store the state of the group but its only attribute is `mutedBy` is this correct?

types/types.go
```diff
@@ -92,24 +100,65 @@ type Marker interface {
...
+// SetMuted implements GroupMarker.
+func (m *MemMarker) SetMuted(routeID, groupKey string, timeIntervalNames []string) {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+	status, ok := m.groups[routeID+groupKey]
+	if !ok {
+		status = &groupStatus{}
+		m.groups[routeID+groupKey] = status
+	}
+	status.mutedBy = timeIntervalNames
```
gotjosh, line 158: Don't we need to aggregate all the time-intervals that this has been muted by?

types/types.go
```diff
@@ -92,24 +100,65 @@ type Marker interface {
...
+	// intervals that mute it. If the list of names is nil or the empty slice
+	// then the muted marker is removed.
+	SetMuted(routeID, groupKey string, timeIntervalNames []string)
+}
 
+// NewMarker returns an instance of a AlertMarker implementation.
+func NewMarker(r prometheus.Registerer) *MemMarker {
+	m := &MemMarker{
+		alerts: map[model.Fingerprint]*AlertStatus{},
+		groups: map[string]*groupStatus{},
```
gotjosh, line 125: How is this gargage collected? As in, how does the marker now that it no longer needs to store the status of an alert.

gotjosh (commented):
Overall looks good, but please see my comment.

types/types.go
grobinson-grafana: See https://github.com/prometheus/alertmanager/pull/3793#discussion_r1549638644.

types/types.go
grobinson-grafana: We don't – the reason is that the `Mutes` method we just merged returns them all https://github.com/prometheus/alertmanager/blob/dacbf0050b72dc68952fcbb5ce934f287e4e1090/timeinterval/timeinterval.go#L37-L53.

types/types.go
grobinson-grafana: Yes! I copied it from here:

```go
// AlertStatus stores the state of an alert and, as applicable, the IDs of
// silences silencing the alert and of other alerts inhibiting the alert.
```

types/types.go
gotjosh: Thanks.

gotjosh (approved):
LGTM

Thanks for clarifying my doubts!

PR #3793, opened 2024-04-02
#3513: Mark muted alerts

grobinson-grafana:
This pull request updates TimeMuteStage and TimeActiveStage to mark groups as muted when its alerts are muted by an active or mute time interval, and remove any existing markers when outside all active and mute time intervals.

It is based on https://github.com/prometheus/alertmanager/pull/3792, https://github.com/prometheus/alertmanager/pull/3794 and https://github.com/prometheus/alertmanager/pull/3795.

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -78,6 +78,7 @@ type Dispatcher struct {
 	route   *Route
 	alerts  provider.Alerts
 	stage   notify.Stage
+	marker  types.GroupMarker
 	metrics *DispatcherMetrics
 	limits  Limits
 
@@ -107,7 +108,7 @@ func NewDispatcher(
 	ap provider.Alerts,
 	r *Route,
 	s notify.Stage,
-	mk types.AlertMarker,
+	mk types.GroupMarker,
 	to func(time.Duration) time.Duration,
 	lim Limits,
 	l log.Logger,
@@ -121,6 +122,7 @@ func NewDispatcher(
 		alerts:  ap,
 		stage:   s,
 		route:   r,
+		marker:  mk,
 		timeout: to,
 		logger:  log.With(l, "component", "dispatcher"),
 		metrics: m,
@@ -145,8 +147,8 @@ func (d *Dispatcher) Run() {
 }
 
 func (d *Dispatcher) run(it provider.AlertIterator) {
-	cleanup := time.NewTicker(30 * time.Second)
-	defer cleanup.Stop()
+	maintenance := time.NewTicker(30 * time.Second)
+	defer maintenance.Stop()
 
 	defer it.Close()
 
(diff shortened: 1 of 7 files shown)


dispatch/dispatch.go
```diff
@@ -182,6 +184,7 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
 				for _, ag := range groups {
 					if ag.empty() {
 						ag.stop()
+						d.marker.DeleteByGroupKey(ag.GroupKey())
```
grobinson-grafana, line 187: We need to delete the marker when the aggregation group is deleted. There is a similar mechanism in https://github.com/prometheus/alertmanager/blob/14cbe6301c732658d6fe877ec55ad5b738abcf06/provider/mem/mem.go#L108.

grobinson-grafana, line 187: We need to delete the marker when the aggregation group is deleted. There is a similar mechanism in `provider/mem/mem.go`: https://github.com/prometheus/alertmanager/blob/14cbe6301c732658d6fe877ec55ad5b738abcf06/provider/mem/mem.go#L108

dispatch/dispatch.go
```diff
@@ -175,28 +177,30 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
...
 	}
 }
 
+func (d *Dispatcher) doMaintenance() {
+	d.mtx.Lock()
+	for _, groups := range d.aggrGroupsPerRoute {
+		for _, ag := range groups {
+			if ag.empty() {
+				ag.stop()
+				d.marker.DeleteByGroupKey(ag.routeID, ag.GroupKey())
```
grobinson-grafana, line 194: Here is where we garbage collect markers.

notify/notify_test.go
```diff
@@ -981,35 +1006,59 @@ func TestTimeActiveStage(t *testing.T) {
...
+				mutedBy, isMuted := marker.Muted("route1", "group1")
+				require.False(t, isMuted)
+				require.Empty(t, mutedBy)
 				// The metric for total suppressed notifications should not
 				// have been incremented, which means it will not be collected.
-				require.NoError(t, prom_testutil.GatherAndCompare(r, strings.NewReader("")))
+				require.NoError(t, prom_testutil.GatherAndCompare(r, strings.NewReader(`
+# HELP alertmanager_marked_alerts How many alerts by state are currently marked in the Alertmanager regardless of their expiry.
+# TYPE alertmanager_marked_alerts gauge
+alertmanager_marked_alerts{state="active"} 0
```
grobinson-grafana, line 1044: Do I need to add metrics for marked groups? @gotjosh

dispatch/dispatch.go
```diff
@@ -175,28 +177,30 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
...
+		case <-maintenance.C:
+			d.doMaintenance()
 		case <-d.ctx.Done():
 			return
 		}
 	}
 }
 
+func (d *Dispatcher) doMaintenance() {
+	d.mtx.Lock()
```
gotjosh, line 189: nit: we tend to have this prefernece in the AM - it made sense before to have them split but not now.

```suggestion
	d.mtx.Lock()
	defer d.mtx.Unlock()
```

notify/notify.go
```diff
@@ -228,6 +233,13 @@ func ActiveTimeIntervalNames(ctx context.Context) ([]string, bool) {
 	return v, ok
 }
 
+// RouteID extracts a RouteID from the context. Iff none exists, the
+// // second argument is false.
```
gotjosh, line 237:
```suggestion
// RouteID extracts a RouteID from the context. If none exists, the
// second argument is false.
```

The other comments have the same typo of `iff`

gotjosh (approved):
LGTM

notify/notify_test.go
gotjosh: I don't think so - I wouldn't really understand their historically use-case.

notify/notify.go
grobinson-grafana: Oh iff is not a spelling mistake, it refers to if and only if.

dispatch/dispatch.go
grobinson-grafana: Will fix! 👍

notify/notify.go
gotjosh: Today I learned 🤯

PR #3794, opened 2024-04-02
#3513: Rewrite TestTimeMuteStage tests

grobinson-grafana:
This commit rewrites the existing TestTimeMuteStage unit tests to have complete isolation between test cases. Before this change, each test case affected the state of its subsequent tests.

The motiviation behind this change is to make it easier to assert that alerts have been marked as muted.


gotjosh (approved):
LGTM

PR #3797, opened 2024-04-03
#3513: Show muted alerts in the Alert Groups API

grobinson-grafana:
This commit updates `/api/v2/alerts/groups` to show if an alert is suppressed from one or more active or mute time intervals. While the muted by field can be found in `/api/v2/alerts`, it is not used here because `/api/v2/alerts` does not take aggregation or routing into consideration.

It also updates the UI to support filtering muted alerts via the Muted checkbox.


ui/app/src/Data/AlertGroupAlertStatus.elm
```diff
@@ -0,0 +1,51 @@
...
+
+type alias AlertGroupAlertStatus =
+    { state : State
+    , silencedBy : List String
+    , inhibitedBy : List String
+    , mutedBy : List String
+    }
+
+
+type State
```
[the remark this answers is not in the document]
grobinson-grafana: It still fails in CI though because I think it rebuilds the files using the same command.

api/v2/openapi.yaml
```diff
@@ -505,6 +517,29 @@ definitions:
       - state
       - silencedBy
       - inhibitedBy
+  alertGroupAlertStatus:
```
grobinson-grafana, line 520: This was duplicated to fix https://github.com/prometheus/alertmanager/pull/3797#discussion_r1550045174.

ui/app/src/Data/GettableAlert.elm
```diff
@@ -22,42 +22,42 @@ import Json.Encode as Encode
...
+    { annotations : Dict String String
     , receivers : List Receiver
     , fingerprint : String
     , startsAt : DateTime
     , updatedAt : DateTime
     , endsAt : DateTime
-    , status : AlertStatus
+    , labels : Dict String String
+    , generatorURL : Maybe String
+    , status : Maybe AlertStatus
```
grobinson-grafana, line 33: What's happened here is somehow this became a `Maybe`.

grobinson-grafana:
There is one issue that I still need to fix. Since both `/api/v2/alerts` and `/api/v2/alerts/groups` share the same OpenAPI schema `GettableAlert`, responses for `/api/v2/alerts` also have the `mutedBy` field even though it is not used. I have tried creating a separate schema for `/api/v2/alerts/groups`, but I keep running into issues with Elm.

grobinson-grafana:
I added a checkbox for filtering Muted alerts and a label to show if the alert is muted, like we have for silenced and inhibited alerts.

Screenshot 2024-04-05 at 5 20 14 PM

ui/app/src/Alerts/Api.elm
```diff
@@ -27,10 +27,10 @@ fetchAlertGroups apiUrl filter =
...
 
-fetchAlerts : String -> Filter -> Cmd (ApiData (List GettableAlert))
+fetchAlerts : String -> Filter -> Cmd (ApiData (List ExtendedGettableAlert))
 fetchAlerts apiUrl filter =
     let
         url =
             String.join "/" [ apiUrl, "alerts" ++ generateAPIQueryString filter ]
     in
-    Utils.Api.send (Utils.Api.get url (Json.Decode.list Data.GettableAlert.decoder))
+    Utils.Api.send (Utils.Api.get url (Json.Decode.list Data.ExtendedGettableAlert.decoder))
```
grobinson-grafana, line 36: This is the most important change to the Elm code. I'm decoding the response for `/api/v2/alerts` as `ExtendedGettableAlert` when in the OpenAPI schema it is `GettableAlert`. The difference between the two is that `ExtendedGettableAlert` has a `mutedBy` column in `status`.

api/v2/api.go
```diff
@@ -86,16 +88,18 @@ type (
 func NewAPI(
 	alerts provider.Alerts,
 	gf groupsFn,
-	sf getAlertStatusFn,
+	asf getAlertStatusFn,
```
grobinson-grafana, line 91: Not the best variable name, option to suggestions! 😄

api/v2/api.go
```diff
@@ -407,17 +411,22 @@ func (api *API) getAlertGroupsHandler(params alertgroup_ops.GetAlertGroupsParams
 	res := make(open_api_models.AlertGroups, 0, len(alertGroups))
 
 	for _, alertGroup := range alertGroups {
+		mutedBy, isMuted := api.groupMutedFunc(alertGroup.RouteID, alertGroup.GroupKey)
+		if !*params.Muted && isMuted {
```
grobinson-grafana, line 415: This skips muted groups if `muted` is false.

api/v2/openapi.yaml
```diff
@@ -451,6 +456,38 @@ definitions:
           - annotations
           - status
       - $ref: '#/definitions/alert'
+  extendedGettableAlert:
+    allOf:
+      - type: object
+        properties:
```
grobinson-grafana, line 462: I've had to duplicate all of these properties due to a bug in swagger generation for Elm that occurs when using inheritance.

api/v2/openapi.yaml
```diff
@@ -505,6 +542,29 @@ definitions:
       - state
       - silencedBy
       - inhibitedBy
+  extendedAlertStatus:
+    type: object
+    properties:
```
grobinson-grafana, line 547: I've also had to duplicate this otherwise the `state` enum does not get inherited and the Elm code fails.

dispatch/dispatch.go
```diff
@@ -206,6 +206,8 @@ type AlertGroup struct {
 	Alerts   types.AlertSlice
 	Labels   model.LabelSet
 	Receiver string
+	GroupKey string
+	RouteID  string
```
grobinson-grafana, line 210: Needed to add `RouteID` here because we added it to the marker. Without it, we cannot see if a group is muted or not.

api/v2/compat.go
```diff
@@ -117,7 +117,7 @@ func PostableSilenceToProto(s *open_api_models.PostableSilence) (*silencepb.Sile
 }
 
 // AlertToOpenAPIAlert converts internal alerts, alert types, and receivers to *open_api_models.GettableAlert.
-func AlertToOpenAPIAlert(alert *types.Alert, status types.AlertStatus, receivers []string) *open_api_models.GettableAlert {
+func AlertToOpenAPIAlert(alert *types.Alert, status types.AlertStatus, receivers, mutedBy []string) *open_api_models.GettableAlert {
```
grobinson-grafana, line 120: `mutedBy` is separate from `types.AlertStatus` because it is related to groups, and does not come from the `AlertMarker` interface.

api/v2/api.go
```diff
@@ -86,16 +88,18 @@ type (
 func NewAPI(
 	alerts provider.Alerts,
 	gf groupsFn,
-	sf getAlertStatusFn,
+	asf getAlertStatusFn,
+	gmf groupMutedFunc,
```
SuperQ, line 92: How about this for consistent naming:

```suggestion
	asf AlertStatusFn,
	gmf groupMutedFn,
```

SuperQ (approved):
Minor suggestion on the function naming, otherwise LGTM.

api/v2/api.go
grobinson-grafana: I don't mind the types as these two are consistent:

```go
getAlertStatusFn func(prometheus_model.Fingerprint) types.AlertStatus
setAlertStatusFn func(prometheus_model.LabelSet)
```

It was more the variable names `asf` and `gmf` that I didn't like.

PR #3801, opened 2024-04-10
Allow webex roomID from template

bakito:
I my environment I'd like to be able to define different webex roomIDs for different namespaces.
This PR adds the possibility to use templates in the webex roomID field, which allow to get the roomID from GroupVariables.


grobinson-grafana:
I took a look at some of the other integrations supported in Alertmanager (Discord, Slack, Microsoft Teams, Telegram, VictorOps) and it seems like templates are not supported for the channel/room. I think the general advice is to have a separate receiver for each Room ID, and then route to the correct receiver using label matchers. What do you think @gotjosh?

bakito:
We have a setup of prometheus and alertmanager using the `kube-prometheus-stack`.
We have several Team wit n namespaces, where each team has n namespaces.
Each Team has 1 - n Rooms for their notifications.

The main issue with webex is not the roomID, but the Token used for the Bot to authenticate.

Teams could create their own AlertmanagerConfig custom resource in each namespace. But then they would always need to know the Bot's auth token for this config. 

Managing all this recievers on the default config is not dynamic enough as teams can dynamically create and destroy namespaces via self service.

If the roomId could be fetched from GroupVariables provided by Prometheus Rules (Each namespace has already their own set of rules), one central receiver config could be used and teams would not have to bother about the auth token.

We would like to use the same Webex Bot for all notifications, but the token should not be spread over several namespaces.

grobinson-grafana:
> Teams could create their own AlertmanagerConfig custom resource in each namespace. But then they would always need to know the Bot's auth token for this config.

Would [`credentials_file`](https://prometheus.io/docs/alerting/latest/configuration/#http_config) work for that use case? You would mount a volume with the credentials file in it.

bakito:
> Would [`credentials_file`](https://prometheus.io/docs/alerting/latest/configuration/#http_config) work for that use case? You would mount a volume with the credentials file in it.

Mabe with the token, but not with the main issue I'm trying to address here. Which is the boilerplate configuration of n notifiers where only the roomId changes.

bakito:
@grobinson-grafana Is this going to be merged soon?

grobinson-grafana:
Hi! 👋 I don't have merge permissions, but I'll ping one of the maintainers.

gotjosh (approved):

Issue #3802, opened 2024-04-11
`route.ID()` returns conflicting IDs


grobinson-grafana:
**What did you do?**

It is possible to write an Alertmanager configuration file where `route.ID()` returns conflicting IDs. For example, the configuration file:

```yml
receiver: test
routes:
- matchers:
    - foo=bar
  continue: true
  routes:
  - matchers:
    - bar=baz
- matchers:
    - foo=bar
  continue: true
  routes:
  - matchers:
    - bar=baz
```

gives the following Route IDs:

```
{}
{}/{foo="bar"}/0
{}/{foo="bar"}/{bar="baz"}/0
{}/{foo="bar"}/1
{}/{foo="bar"}/{bar="baz"}/0
```

**What did you expect to see?**

It should give these IDs instead:

```
{}
{}/{foo="bar"}/0
{}/{foo="bar"}/0/{bar="baz"}/0
{}/{foo="bar"}/1
{}/{foo="bar"}/1/{bar="baz"}/0
```

PR #3803, opened 2024-04-11
Fix `Route.ID()` returns conflicting IDs

grobinson-grafana:
This pull request fixes a bug (#3802) where Route.ID() returns conflicting IDs.
For example, the configuration:

    receiver: test
    routes:
    - matchers:
        - foo=bar
      continue: true
      routes:
      - matchers:
        - bar=baz
    - matchers:
        - foo=bar
      continue: true
      routes:
      - matchers:
        - bar=baz

Gives the following Route IDs:

    {}
    {}/{foo="bar"}/0
    {}/{foo="bar"}/{bar="baz"}/0
    {}/{foo="bar"}/1
    {}/{foo="bar"}/{bar="baz"}/0

When it should give these Route IDs:

    {}
    {}/{foo="bar"}/0
    {}/{foo="bar"}/0/{bar="baz"}/0
    {}/{foo="bar"}/1
    {}/{foo="bar"}/1/{bar="baz"}/0

### DIFF
--- a/dispatch/route.go
+++ b/dispatch/route.go
@@ -197,6 +197,7 @@ func (r *Route) ID() string {
 			if r == r.parent.Routes[i] {
 				b.WriteRune('/')
 				b.WriteString(strconv.Itoa(i))
+				break
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


gotjosh (approved):
LGTM

Issue #3351, opened 2023-05-03
Webhooks for firing alerts contain zeroed EndsAt


grobinson-grafana:
**What did you do?**

It appears that webhooks for firing alerts contain zeroed EndsAt timestamps `0001-01-01T00:00:00Z` instead of the actual EndsAt timestamp of the alert as can be seen in `/v2/alerts`.

Having looked into this further the code that causes this can be found on [Line 507 of `dispatch.go`](https://github.com/prometheus/alertmanager/blob/v0.25.0/dispatch/dispatch.go#L507):

```go
for _, alert := range alerts {
	a := *alert
	// Ensure that alerts don't resolve as time move forwards.
	if !a.ResolvedAt(now) {
		a.EndsAt = time.Time{}
	}
	alertsSlice = append(alertsSlice, &a)
}
```

I think the intention of this code is to prevent alerts from becoming resolved during subsequent stages of the receiver such as WaitStage and DedupStage. This can happen when a firing alert enters the stages at time **t1** but doesn't finish the stages until time **t2 > EndsAt**.

**What did you expect to see?**

I expect to see either `EndsAt` contain a non-zero timestamp or for the `EndsAt` column to be omitted from the JSON for firing alerts.

**What did you see instead? Under which circumstances?**

Related: https://github.com/prometheus/alertmanager/issues/3341

grobinson-grafana:
I found the original commit by Brian [here](https://github.com/prometheus/alertmanager/commit/7078333202f37577d2549aa7e223f6eb98a2d9d1).

I think the need to do this is a consequence of how Alertmanager always uses  `EndsAt - now` to determine if an alert is firing or resolved, so as time moves forward the state of the alert can change between successive retries. This shouldn't be possible, but would require refactoring how the state of the alert is propagated to notifiers.

An immediate fix would be to instead remove the `EndsAt` field from the JSON for firing alerts, instead only including it for resolved alerts. This would at least remove some confusion from the user's perspective even if it doesn't fix the bigger issue.

gotjosh:
The more I think about it, the more I feel like the semantics of flushing and time are broken.

If we assume that once we've determined that an alert needs to flush and it enters the subsequent stages, it should not be time-constrained? Instead, it should just be forced flush. I'd argue that modifying endsAt is not something we wanna do.

grobinson-grafana:
[...]

PR #3805, opened 2024-04-12
Fix zeroed EndsAt timestamp

grobinson-grafana:
This commit fixes a bug where firing alerts had zeroed `EndsAt` timestamps in notifications. This bug was introduced in another PR (#1686) to fix issues where alerts are resolved during a flush.

This commit takes a different approach where instead of removing the `EndsAt` timestamp from the alert, it uses the `ResolvedAt` and `StatusAt` methods with the timestamp in the context, not the current time.

```json
{
  "receiver": "default",
  "status": "firing",
  "alerts": [
    {
      "status": "firing",
      "labels": {
        "bar": "baz"
      },
      "annotations": {},
      "startsAt": "2024-04-12T14:34:46.324903+02:00",
      "endsAt": "2024-04-12T14:39:46.324903+02:00",
      "generatorURL": "",
      "fingerprint": "6ad8c192a44006c5"
    }
  ],
  "groupLabels": {},
  "commonLabels": {
    "bar": "baz"
  },
  "commonAnnotations": {},
  "externalURL": "http://127.0.0.1:9093",
  "version": "4",
  "groupKey": "{}:{}",
  "truncatedAlerts": 0
}
```


dispatch/dispatch.go
```diff
@@ -493,29 +491,30 @@ func (ag *aggrGroup) empty() bool {
...
+		return
+	}
+
 	var (
 		alerts      = ag.alerts.List()
 		alertsSlice = make(types.AlertSlice, 0, len(alerts))
-		now         = time.Now()
 	)
 	for _, alert := range alerts {
 		a := *alert
```
grobinson-grafana, line 510: I suspect we can remove this, but will do so in another PR.

grobinson-grafana:
Build fails as we need https://github.com/prometheus/alertmanager/pull/3806.

dispatch/dispatch.go
zecke: If I understand the intent correctly this is passing the reference time through the context and every notifier should use the `.*At(now time.Time)` variant. As we don't (plan) to modify the alert we don't need to take a deep copy (some notifiers call types.Alerts).

Dropping this copy might be hard. On line 528 (529) the UpdatedAt time is compared between the copy and the storage of the aggregation group. Without a copy these numbers will always be the same.

I am wondering if there are ways to make sure that integrations never call the `Resolved()`, `Status()`? Given that alertmanager is likely to see an increase in number of integrations that appears desirable to me to ease writing and reviewing these.

Given that the model.Alert is copied and one needs the reference time. What is your thought on creating something like a model.AlertsSnapshot that holds an immutable snapshot of the alert and the status/reference time? This would ease writing and reviewing the notifiers, protect the core from accidental modificiations?

dispatch/dispatch.go
```diff
@@ -526,7 +525,7 @@ func (ag *aggrGroup) flush(notify func(...*types.Alert) bool) {
 				level.Error(ag.logger).Log("msg", "failed to get alert", "err", err, "alert", a.String())
 				continue
 			}
-			if a.Resolved() && got.UpdatedAt == a.UpdatedAt {
+			if a.ResolvedAt(now) && got.UpdatedAt == a.UpdatedAt {
```
zecke, line 528: That seems fine. If `got` has any modification reflected in UpdatedAt then we do not remove the alert.

dispatch/dispatch.go
grobinson-grafana: > Dropping this copy might be hard. On line 528 (529) the UpdatedAt time is compared between the copy and the storage of the aggregation group. Without a copy these numbers will always be the same.

Oh well spotted!

> What is your thought on creating something like a model.AlertsSnapshot that holds an immutable snapshot of the alert and the status/reference time?

This is what I tried at first, but I found it to be quite an intrusive change (https://github.com/prometheus/alertmanager/issues/3351#issuecomment-1567365253). I haven't ruled it out though - do you think that would be a better way to fix this?

zecke: I think longterm we are better off with a genuine snapshot. It eases writing integrations, it eases code review of them. I think some of the type changes we can probably automate with gofmt?

grobinson-grafana: Something like this?

```go
type AlertSnapshot struct {
	Alert
	// status of the alert at the time it was snapshot.
	status model.AlertStatus
}

func (a *AlertSnapshot) Resolved() bool {
	return a.status == model.AlertResolved
}

func (a *AlertSnapshot) Status() model.AlertStatus {
	return a.status
}

// Need to override String() as otherwise it will call Resolved()
// of model.Alert not AlertSnapshot.
func (a *AlertSnapshot) String() string {
	s := fmt.Sprintf("%s[%s]", a.Name(), a.Fingerprint().String()[:7])
	if a.Resolved() {
		return s + "[resolved]"
	}
	return s + "[active]"
}
```

grobinson-grafana: > UpdatedAt, Name, Status, Resolved, Labels, Annotations, StartsAt, EndsAt, GeneratorURL, Fingerprint, ResolvedAt

We would all inherit these from `model.Alert` embedded in the struct?

zecke: Yeah but from my point of view the trade-off is whether we want all of these fields to be mutable.

grobinson-grafana: Oh – I think I misunderstood. By mutable/immutable are you talking about the fields being package-public (capitalized) or package-private (starting with lowercase)?

zecke: Yes. Starting with lower case. I wanted to see how an implementation of this looks like and got the below to compile (and pass most tests but it requires some clean ups).

For me the biggest benefit is that it creates an immutable snapshot and simplifies API. The notification handlers are unable to modify the underlying Alert and the API removes the `*At(time.Time)` functions.


Please see:
https://github.com/prometheus/alertmanager/compare/main...zecke:alertmanager:zecke-alternative-ends-at

zecke: Sure. Let me fix the formatting and remove some of the repeated code in the tests.

grobinson-grafana:
I'm closing this PR in favor of #3827.

PR #3806, opened 2024-04-12
Bump prometheus/common to v0.52.3

grobinson-grafana:
This commit bumps prometheus/common to `v0.52.3`. It has a breaking change where the metric `alertmanager_build_info` has been renamed to `go_build_info` as the metric has been moved from `prometheus/common` to `prometheus/client_golang` and the `namspace` argument has been removed.


mortenmj:
This is stopping us from updating prometheus/common in our repo; would be great to see it merged.

Issue #3808, opened 2024-04-15
Different aggregation groups can share the same nflog


grobinson-grafana:
**What did you do?**

In certain cases, it is possible for two (or more) different aggregation groups to share an nflog. The conditions for this to happen are the following:

1. There must be at least two routes with the same matchers and group labels (`group_by`).
2. At least one of the routes must have `continue: true`.
3. Both routes must use the same receivers.

For example, take the following configuration:

```yml
receivers:
  - name: test1
    webhook_configs:
      - url: http://127.0.0.1:8080/test1
  - name: test2
    webhook_configs:
      - url: http://127.0.0.1:8080/test2
route:
  receiver: test1
  group_wait: 15s
  group_interval: 30s
  routes:
    - continue: true
      receiver: test1
      matchers:
      - foo=bar
    - continue: true
      receiver: test1
      matchers:
      - foo=bar
```

This configuration meets all of the requirements to share an nflog. 

When an Alertmanager is run with this configuration, you can see from the `flushing alerts` and `Notify success` lines that two aggregation groups are created and then flushed:

```
ts=2024-04-15T08:33:35.749Z caller=dispatch.go:164 level=debug component=dispatcher msg="Received alert" alert=[3fff2c2][active]
ts=2024-04-15T08:33:50.750Z caller=dispatch.go:516 level=debug component=dispatcher aggrGroup="{}/{foo=\"bar\"}:{}" msg=flushing alerts=[[3fff2c2][active]]
ts=2024-04-15T08:33:50.749Z caller=dispatch.go:516 level=debug component=dispatcher aggrGroup="{}/{foo=\"bar\"}:{}" msg=flushing alerts=[[3fff2c2][active]]
ts=2024-04-15T08:33:50.753Z caller=notify.go:863 level=debug component=dispatcher receiver=test1 integration=webhook[0] aggrGroup="{}/{foo=\"bar\"}:{}" alerts=[[3fff2c2][active]] msg="Notify success" attempts=1 duration=2.743458ms
ts=2024-04-15T08:33:50.753Z caller=notify.go:863 level=debug component=dispatcher receiver=test1 integration=webhook[0] aggrGroup="{}/{foo=\"bar\"}:{}" alerts=[[3fff2c2][active]] msg="Notify success" attempts=1 duration=2.85275ms
```

However, reading the nflog file (after shutdown) shows just one nflog entry on disk:

```
N
>
{}/{foo="bar"}:{}
test1webhook*
[...]

PR #3809, opened 2024-04-15
Make `route.Key()` unique and remove `route.ID()`

grobinson-grafana:
This pull request replaces the code in `route.Key()` with that of `route.ID()`, and removes `route.ID()`. The motivation behind this change is to fix a number of bugs caused by conflicting group keys such as "Different aggregation groups can share the same nflog" (#3808) and also prevent an issue where groups are incorrectly marked as muted when they are not.

### DIFF
--- a/dispatch/route.go
+++ b/dispatch/route.go
@@ -169,24 +169,12 @@ func (r *Route) Match(lset model.LabelSet) []*Route {
 	return all
 }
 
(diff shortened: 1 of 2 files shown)


grobinson-grafana:
I'm closing this as @gotjosh found this can have unintended side effects when a configuration is changed. For example, if a route is moved in the configuration file, but not edited, then it's index will change and cause its nflogs to be expired.

PR #3812, opened 2024-04-19
Add date and tz functions to templates

grobinson-grafana:
This commit adds the date and tz functions to templates. This means users can now format time in a specified format and also change the timezone to their specific locale.

An example of how these functions work, and can be composed together, can be seen here:

	{{ .StartsAt | tz "Europe/Paris" | date "15:04:05 MST" }}

### DIFF
--- a/template/template_test.go
+++ b/template/template_test.go
@@ -519,9 +519,9 @@ func TestTemplateFuncs(t *testing.T) {
 		exp:   "2024-01-01 09:15:30 +0100 CET",
 	}, {
 		title:  "Template using invalid tz",
-		in:     `{{ . | tz "Bad/Timezone" }}`,
+		in:     `{{ . | tz "Invalid/Timezone" }}`,
 		data:   time.Date(2024, 1, 1, 8, 15, 30, 0, time.UTC),
-		expErr: "template: :1:7: executing \"\" at <tz \"Bad/Timezone\">: error calling tz: unknown time zone Bad/Timezone",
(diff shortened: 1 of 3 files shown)


grobinson-grafana:
@verdel can you open a separate issue for that?

template/template_test.go
```diff
@@ -506,6 +506,16 @@ func TestTemplateFuncs(t *testing.T) {
 		title: "Template using reReplaceAll",
 		in:    `{{ reReplaceAll "ab" "AB" "abc" }}`,
 		exp:   "ABc",
+	}, {
```
gotjosh, line 509: Can you please add unhappy cases? I'm not sure what happens when you pass a non-valid format to `date` or when empty value `time` and an error.

grobinson-grafana: Yeah sure! Template panics like other functions. As far as I can tell, there is no discernible difference between returning an error and panicking in the function, both result in a panic:

```go
"reReplaceAll": func(pattern, repl, text string) string {
	re := regexp.MustCompile(pattern) // panics if regex is invalid
	return re.ReplaceAllString(text, repl)
},
```

```go
// tz returns the time in the timezone.
"tz": func(name string, t time.Time) (time.Time, error) {
	loc, err := time.LoadLocation(name)
	if err != nil {
		return time.Time{}, err // panics with err
	}
	return t.In(loc), nil
},
```

gotjosh (approved):
LGTM

PR #3820, opened 2024-04-25
Add (optional) names to routes

### LINKED ISSUE #3817: How to fix non unique GroupKeys?
I've been thinking on the issue of the GroupKey being non-unique for a little while, and if its possible to make the GroupKey both stable and unique without adding new fields to the configuration file.

What do I mean by stable? I mean that the GroupKey should not depend on the position of the route relative to its siblings. This is the reason why we cannot use the RouteID when calculating the GroupKey, because as soon as the route is moved higher or lower in the configuration file the group's notification log is invalidated causing repeat notifications.

One option we have is to use a non cryptographic hash function (for example fnv) to calculate a hash using extra metadata from the route that is not included in the GroupKey at present. For example:

1. Receiver name
2. Mute time intervals
3. Active time intervals
4. Child routes

This would change how GroupKey is calculated from being a top-down path of routes from parent to child (i.e. `route0/route1/route...N/groupLabels`) to a bottom-up Merkle tree where the parents are derived from their hash of their children.

The main issue I see with this solution is observability. Without adequate tools, it will be very hard to see which group is flushed in the logs just from its hash.

So perhaps it's better to just add an (optional) name to routes? If the name is absent, then Alertmanager uses the current mechanism (which can be non unique in some cases). If a unique GroupKey is required, then a name can be set on the route to discriminate between them.

grobinson-grafana:
This pull request adds (optional) names to routes to fix #3817. In the case where a user has two routes with the same receiver, matchers and group by, a name can be used to ensure their groups have unique group keys and avoid using the same nflog.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -549,6 +549,9 @@ func (c *Config) UnmarshalYAML(unmarshal func(interface{}) error) error {
 	if c.Route == nil {
 		return fmt.Errorf("no routes provided")
 	}
+	if len(c.Route.Name) > 0 {
+		return errors.New("root route cannot have a name")
+	}
 	if len(c.Route.Receiver) == 0 {
 		return fmt.Errorf("root route must specify a default receiver")
 	}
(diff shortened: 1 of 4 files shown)


simonpasquier:
It implies that the user knows in advance that 2 routes are going to create identical group keys? It doesn't seem to me a nice user experience.

What if Alertmanager adds the route index when it suspects a possible conflict? Would it solve the problem?

grobinson-grafana:
> What if Alertmanager adds the route index when it suspects a possible conflict? Would it solve the problem?

That's what we tried [here](https://github.com/prometheus/alertmanager/pull/3809#issuecomment-2061772387), but it has unintended side effects. For example, if a new route is added to the start of the `routes:` then it invalidates the nflogs for all other routes.

Issue #3667, opened 2024-01-11
Discord integration: no possibility to configure the webhook content object


aZ2394:
Hi there,
According to the code of the Discord integration, there is no chance to specify Web hook content field from the message notification template. Just a embed title&message can be configured. 
It's useful thing to have it when you want to tag person/role in alert. It can be doable in title&message of the embed message but user won't get a notification due to Discord API limitation.
So, it limit the tagging in Discord integration at all, and there is no chance with the current implementation to notify specific user about the alert.

**What did you do?**
Sent an alert to discord with the tagged user/role

**What did you expect to see?**
Tagged user should be notified in Discord client

**What did you see instead? Under which circumstances?**
Just an mention in alert without any notification for user in Discord client

* Alertmanager version:

alertmanager, version 0.26.0 (branch: HEAD, revision: d7b4f0c7322e7151d6e3b1e31cbc15361e295d8d)
  build user:       root@df8d7debeef4
  build date:       20230824-11:11:58
  go version:       go1.20.7
  platform:         linux/amd64
  tags:             netgo

* Prometheus version:

prometheus, version 2.48.1 (branch: HEAD, revision: 63894216648f0d6be310c9d16fb48293c45c9310)
  build user:       root@71f108ff5632
  build date:       20231208-23:33:22
  go version:       go1.21.5
  platform:         linux/amd64
  tags:             netgo,builtinassets,stringlabels

* Alertmanager configuration file:
```
route:
  group_by: ['alertname', 'job']

  group_wait: 30s
  group_interval: 5m
  repeat_interval: 6h

  receiver: discord2

  routes:
      - matchers:
          - service="discord1"
        receiver: discord1
        group_interval: 12h
        repeat_interval: 24h


receivers:
- name: discord1
  discord_configs:
  - webhook_url: ''
    title: '{{ template "discord.default.title" . }}'
    message: '{{ template "discord.default.message" . }}'

- name: discord2
  discord_configs:
  - webhook_url: ''
    title: '{{ template "discord.default2.title" . }}'
    message: '{{ template "discord.default2.message" . }}'
    send_resolved: false


templates:
- '/etc/alertmanager/templates/default.tmpl'
```

* Prometheus configuration file:
```
insert configuration here (if relevant to the issue)
```

* Logs:
```
insert Prometheus and Alertmanager logs relevant to the issue here
```

PR #3761, opened 2024-03-12
Extend Discord Notifier

LucaDev:
This PR extends the Discord notifier to improve the readability of alerts.
It changes 5 key elements:
1. A configureable icon as bot Logo (I'm currently using the GitHub Avatar. Is that alright? Is there a better way for this?)
2. A configurable name for the bot
3. Alert-ID in the footer
4. Start & end times in the footer (Even if the message is sent with a delay, the correct time will be displayed)
6. It uses Discords fields-feature to display labels
7. It sends one message per alert
Discord Screenshot

Before:
Screenshot old

PR #3821, opened 2024-04-26
Add discord message fields

yaskinny:
This PR adds some fields to Discord message like Content which is already mentioned in issue #3667.

Being able to set _Content_ is really important to have since at this moment, there is no way to tag users in alerts. Default Discord 
template embeds alerts and in Discord, ATM, tagging someone in embedded fields, does not tag them. It will just include user/role name without sending notification like when you get tagged in normal message.

Discord code [already](https://github.com/prometheus/alertmanager/blob/main/notify/discord/discord.go#L77) had `content` field but it's not used in the code nor config.

I added other fields like avatar url and username.

Below content are example configs and result:
Alertmanager receiver config:
```yaml

receivers:
  - name: 'discord'
    discord_configs:
    - webhook_url: 'https://discord.com/api/webhooks/1153054938155450368/kkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkkEwc3KouoUWE4'
      content: '{{ template "my.content" . }}'
      avatar_url: 'https://www.tbstat.com/wp/uploads/2023/05/Fvz9hOIXwAEaIR8-669x675.jpeg'
      username: GOFW
```

Template:
```
{{- define "my.content" }}
tags: {{- range .Alerts }}
{{- if .Annotations.tags }}
{{- reReplaceAll "," " " .Annotations.tags }}
{{- end }}
{{- end }}
{{- end }}
```

Prometheus alert:
```yaml
---
groups:
- name: watchdogs
  rules:
  - alert: WatchDog-1
    expr: vector(1)
    labels:
      severity: info
    annotations:
      summary: "WatchDog1"
      description: "FAKE"
      tags: '<@5132333333333336177>,<@5132333333333336177>'
        
  - alert: WatchDog-2
    expr: vector(2)
    labels:
      severity: critical
    annotations:
      summary: "WatchDog2"
      description: "FAKE"
      tags: '<@5132333333333336177>,<@5132333333333336177>'

```

Screenshot from 2024-04-26 20-04-24

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -220,8 +220,10 @@ type DiscordConfig struct {
 	WebhookURL     *SecretURL                  `yaml:"webhook_url,omitempty" json:"webhook_url,omitempty"`
 	WebhookURLFile string                      `yaml:"webhook_url_file,omitempty" json:"webhook_url_file,omitempty"`
 
-	Title   string `yaml:"title,omitempty" json:"title,omitempty"`
-	Message string `yaml:"message,omitempty" json:"message,omitempty"`
+	Title     string `yaml:"title,omitempty" json:"title,omitempty"`
+	Message   string `yaml:"message,omitempty" json:"message,omitempty"`
+	Content   string `yaml:"content,omitempty" json:"content,omitempty"`
+	AvatarURL string `yaml:"avatar_url,omitempty" json:"avatar_url,omitempty"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -224,6 +224,7 @@ type DiscordConfig struct {
 	Message   string `yaml:"message,omitempty" json:"message,omitempty"`
 	Content   string `yaml:"content,omitempty" json:"content,omitempty"`
 	AvatarURL string `yaml:"avatar_url,omitempty" json:"avatar_url,omitempty"`
+	Username  string `yaml:"username,omitempty" json:"username,omitempty"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
--- a/notify/discord/discord.go
+++ b/notify/discord/discord.go
@@ -159,6 +159,10 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 		w.Content = content
 	}
 
+	if len(n.conf.Username) != 0 {
+		w.Username = n.conf.Username
+	}
+
 	if len(n.conf.AvatarURL) != 0 {
 		if _, err := url.Parse(n.conf.AvatarURL); err == nil {
 			w.AvatarURL = n.conf.AvatarURL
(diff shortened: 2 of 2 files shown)


grobinson-grafana:
@yaskinny can you signoff your commits? That will fix the DCO check 🙂

yaskinny:
also it might be beneficial to edit the default discord template in a way that older config version(without username/content) works without any issue after updating to newer alertmanager version.

like making username/content optional in the template. since at this point, users need to rewrite their own template if they want to use username/content fields. with updating default template, it will has better UX since by default it will be included in the alerts.

@grobinson-grafana let me know if you want me to work on that or not.

grobinson-grafana:
> also it might be beneficial to edit the default discord template in a way that older config version(without username/content) works without any issue after updating to newer alertmanager version.
> 
> like making username/content optional in the template. since at this point, users need to rewrite their own template if they want to use username/content fields. with updating default template, it will has better UX since by default it will be included in the alerts.

I'm not sure I understand. What are you suggesting?

yaskinny:
@grobinson-grafana 

I mean the default Discord template does not include the _content_ field.  
[https://github.com/prometheus/alertmanager/blob/e5821b78917591a5603e3e1788b6a94d5b2972e0/template/default.tmpl#L126-L136](https://github.com/prometheus/alertmanager/blob/e5821b78917591a5603e3e1788b6a94d5b2972e0/template/default.tmpl#L126-L136)

It would be better if the default template used the _content_ field to allow anyone who uses the _content_ option to utilize it without having to write their own template from scratch. However, since older versions of Alertmanager do not include the _content_ field, we need to make it optional (use a condition in the template to check if _content_ is set before using it). By making it optional, users won't face errors after updating to the new Alertmanager version.

grobinson-grafana:
> @grobinson-grafana
> 
> I mean the default Discord template does not include the _content_ field.
> 
> https://github.com/prometheus/alertmanager/blob/e5821b78917591a5603e3e1788b6a94d5b2972e0/template/default.tmpl#L126-L136
> 
> It would be better if the default template used the _content_ field to allow anyone who uses the _content_ option to utilize it without having to write their own template from scratch. However, since older versions of Alertmanager do not include the _content_ field, we need to make it optional (use a condition in the template to check if _content_ is set before using it). By making it optional, users won't face errors after updating to the new Alertmanager version.

Is the suggestion to also replace the embeds with content? If not, what would the default template for `content` be and what would happen to the embeds?

grobinson-grafana:
You might also want to check out https://github.com/prometheus/alertmanager/pull/3761 which extends the use of embeds.

mogoll92:
> You might also want to check out #3761 which extends the use of embeds.

I think he meant don't touch embedded at all just to make content field configurable, atm we have it just in webhook struct but does use nowhere. 
```
type webhook struct {
	Content string         `json:"content"`
	Embeds  []webhookEmbed `json:"embeds"`
}
```
It would be really useful as for some custom solutions you would like to mention users over user_id or role_id and make sure they got notification. In current implementation of discord embedded everything that is tagged inside embedded won't trigger discord notification.

gotjosh:
Duplicate of #4007 and #4081

PR #3823, opened 2024-04-29
Add receiver name to route key

grobinson-grafana:
This commit adds the receiver name to the route key to reduce the chances of having non unique group keys (#3817). Like the previous version, it does not guarantee the group key is unique, however it does make it collisions less likely to occur.

### DIFF
--- a/dispatch/route.go
+++ b/dispatch/route.go
@@ -177,7 +177,11 @@ func (r *Route) Key() string {
 		b.WriteString(r.parent.Key())
 		b.WriteRune('/')
 	}
+	b.WriteRune('(')
+	b.WriteString(r.RouteOpts.Receiver)
+	b.WriteRune(',')
 	b.WriteString(r.Matchers.String())
+	b.WriteRune(')')
 	return b.String()
 }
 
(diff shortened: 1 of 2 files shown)


gotjosh:
I think this would invalidate the nflog upon a restart 😭 Perhaps we need a way to execute this change a bit more safely if we could migrate the nflog at the same time?

grobinson-grafana:
How would it be invalidated? I'm not sure I am following. The state key is currently derived as follows: 

```go
func receiverKey(r *pb.Receiver) string {
	return fmt.Sprintf("%s/%s/%d", r.GroupName, r.Integration, r.Idx)
}

// stateKey returns a string key for a log entry consisting of the group key
// and receiver.
func stateKey(k string, r *pb.Receiver) string {
	return fmt.Sprintf("%s:%s", k, receiverKey(r))
}
```

so adding the receiver name to the group key would only invalid the nflog if it was changed, much the same as if you changed the matchers?

grobinson-grafana:
I spoke to @gotjosh and realized that if we want to do this, we need to migrate existing nflogs so we don't break them on first boot. We will come back to this problem at a later date.

PR #3826, opened 2024-05-01
Fix race condition in dispatch.go

grobinson-grafana:
This commit fixes a race condition in dispatch.go that would cause a firing alert to be deleted from the aggregation group when instead it should have been flushed.

The root cause is a race condition that can occur when dispatch.go deletes resolved alerts from the aggregation group following a successful notification. If a firing alert with the same fingerprint is added back to the aggregation group at the same time then the firing alert can be deleted.

The issue was first reported in #3006.

#### How to observe the race condition

1. `git checkout v0.27.0`
2. Add the following sleep to `dispatch.go`. This makes it easier to observe the race condition because the scheduler can resume the test between `ag.alerts.Get(fp)` and `ag.alerts.Delete(fp)`, during which it should replace the resolved alert with a firing alert.

```go
index 640b22ab..1eafa054 100644
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -521,6 +521,7 @@ func (ag *aggrGroup) flush(notify func(...*types.Alert) bool) {
                        // again since we notified about it.
                        fp := a.Fingerprint()
                        got, err := ag.alerts.Get(fp)
+                       <-time.After(time.Millisecond)
                        if err != nil {
                                // This should never happen.
```

3. Add and run the following test to `dispatch_test.go`. It should fail:

```go
type bufferedStage struct {
	c chan struct{}
}

func (s *bufferedStage) Exec(ctx context.Context, _ log.Logger, _ ...*types.Alert) (context.Context, []*types.Alert, error) {
	s.c <- struct{}{}
	return ctx, nil, nil
}

func TestDispatcherRaceFiringAlertDeleted(t *testing.T) {
	r := Route{RouteOpts: RouteOpts{GroupWait: 0, GroupInterval: time.Second}}
	s := bufferedStage{c: make(chan struct{}, 1)}
	logger := log.NewLogfmtLogger(os.Stdout)
	marker := types.NewMarker(prometheus.NewRegistry())
	alerts, err := mem.NewAlerts(context.Background(), marker, time.Hour, nil, logger, nil)
	require.NoError(t, err)
	defer alerts.Close()
	timeout := func(d time.Duration) time.Duration { return time.Duration(0) }
	dispatcher := NewDispatcher(alerts, &r, &s, marker, timeout, nil, logger, NewDispatcherMetrics(false, prometheus.NewRegistry()))
	go dispatcher.Run()
	defer dispatcher.Stop()

	// Add a new firing alert and wait for it to be flushed.
	a1 := newAlert(model.LabelSet{"foo": "bar"})
	a1.StartsAt = time.Now().Add(-time.Second)
	a1.UpdatedAt = a1.StartsAt
	require.NoError(t, alerts.Put(a1))
	<-s.c

	// Change the firing alert to resolved and wait for it to be flushed too.
	a2 := newAlert(model.LabelSet{"foo": "bar"})
	a2.StartsAt = a1.StartsAt
	a2.UpdatedAt = a1.UpdatedAt
	a2.EndsAt = time.Now()
	require.NoError(t, alerts.Put(a2))
	<-s.c

	// Change the resolved alert back to firing. If the race condition is still
	// present, it will be deleted instead of flushed.
	a3 := newAlert(model.LabelSet{"foo": "bar"})
	a3.StartsAt = time.Now()
	a3.UpdatedAt = time.Now()
	require.NoError(t, alerts.Put(a3))
	select {
	case <-s.c:
	case <-time.After(2 * time.Second):
		t.Fatal()
	}
}
```

#### How to observe the fix

1. Check out this branch, add the sleep and re-run the same test. It should pass.

```go
index d78fbf68..efb0c246 100644
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -517,6 +517,7 @@ func (ag *aggrGroup) flush(notify func(...*types.Alert) bool) {

        if notify(alertsSlice...) {
                for _, a := range alertsSlice {
+                       <-time.After(time.Millisecond)
                        // Only delete if the fingerprint has not been inserted again since we notified about it.
                        if err := ag.alerts.DeleteIf(a.Fingerprint(), func(b *types.Alert) bool {
                                return a.Resolved() && a.UpdatedAt == b.UpdatedAt
```

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -517,19 +517,11 @@ func (ag *aggrGroup) flush(notify func(...*types.Alert) bool) {
 
 	if notify(alertsSlice...) {
 		for _, a := range alertsSlice {
-			// Only delete if the fingerprint has not been inserted
-			// again since we notified about it.
-			fp := a.Fingerprint()
-			got, err := ag.alerts.Get(fp)
-			if err != nil {
-				// This should never happen.
-				level.Error(ag.logger).Log("msg", "failed to get alert", "err", err, "alert", a.String())
-				continue
-			}
-			if a.Resolved() && got.UpdatedAt == a.UpdatedAt {
-				if err := ag.alerts.Delete(fp); err != nil {
-					level.Error(ag.logger).Log("msg", "error on delete alert", "err", err, "alert", a.String())
-				}
+			// Only delete if the fingerprint has not been inserted again since we notified about it.
+			if err := ag.alerts.DeleteIf(a.Fingerprint(), func(b *types.Alert) bool {
+				return a.Resolved() && a.UpdatedAt == b.UpdatedAt
+			}); err != nil {
+				level.Error(ag.logger).Log("msg", "error on delete alert", "err", err, "alert", a.String())
 			}
 		}
 	}
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -499,14 +499,17 @@ func (ag *aggrGroup) flush(notify func(...*types.Alert) bool) {
 	}
 
 	var (
-		alerts      = ag.alerts.List()
-		alertsSlice = make(types.AlertSlice, 0, len(alerts))
-		now         = time.Now()
+		alerts        = ag.alerts.List()
+		alertsSlice   = make(types.AlertSlice, 0, len(alerts))
+		resolvedSlice = make(types.AlertSlice, 0, len(alerts))
+		now           = time.Now()
 	)
 	for _, alert := range alerts {
 		a := *alert
 		// Ensure that alerts don't resolve as time move forwards.
-		if !a.ResolvedAt(now) {
+		if a.ResolvedAt(now) {
+			resolvedSlice = append(resolvedSlice, &a)
+		} else {
 			a.EndsAt = time.Time{}
 		}
 		alertsSlice = append(alertsSlice, &a)
@@ -516,13 +519,12 @@ func (ag *aggrGroup) flush(notify func(...*types.Alert) bool) {
 	level.Debug(ag.logger).Log("msg", "flushing", "alerts", fmt.Sprintf("%v", alertsSlice))
 
 	if notify(alertsSlice...) {
-		for _, a := range alertsSlice {
-			// Only delete if the fingerprint has not been inserted again since we notified about it.
-			if err := ag.alerts.DeleteIf(a.Fingerprint(), func(b *types.Alert) bool {
-				return a.Resolved() && a.UpdatedAt == b.UpdatedAt
-			}); err != nil {
-				level.Error(ag.logger).Log("msg", "error on delete alert", "err", err, "alert", a.String())
-			}
+		// Delete all resolved alerts as we just sent a notification for them,
+		// and we don't want to send another one. However, we need to make sure
+		// that each resolved alert has not fired again during the flush as then
+		// we would delete an active alert thinking it was resolved.
+		if err := ag.alerts.DeleteIfNotModified(resolvedSlice); err != nil {
+			level.Error(ag.logger).Log("msg", "error on delete alerts", "err", err)
 		}
 	}
 }
--- a/dispatch/dispatch_test.go
+++ b/dispatch/dispatch_test.go
@@ -236,9 +236,33 @@ func TestAggrGroup(t *testing.T) {
 		}
 	}
 
-	// Resolve all alerts, they should be removed after the next batch was sent.
-	a1r, a2r, a3r := *a1, *a2, *a3
-	resolved := types.AlertSlice{&a1r, &a2r, &a3r}
+	// Resolve an alert, and it should be removed after the next batch was sent.
+	a1r := *a1
+	a1r.EndsAt = time.Now()
+	ag.insert(&a1r)
+	exp := append(types.AlertSlice{&a1r}, removeEndsAt(types.AlertSlice{a2, a3})...)
+
+	select {
+	case <-time.After(2 * opts.GroupInterval):
+		t.Fatalf("expected new batch after group interval but received none")
+	case batch := <-alertsCh:
+		lastCurMtx.Lock()
+		s := time.Since(last)
+		lastCurMtx.Unlock()
+		if s < opts.GroupInterval {
+			t.Fatalf("received batch too early after %v", s)
+		}
+		sort.Sort(batch)
+
+		if !reflect.DeepEqual(batch, exp) {
+			t.Fatalf("expected alerts %v but got %v", exp, batch)
+		}
+	}
+
+	// Resolve all remaining alerts, they should be removed after the next batch was sent.
+	// Do not add a1r as it should have been deleted following the previous batch.
+	a2r, a3r := *a2, *a3
+	resolved := types.AlertSlice{&a2r, &a3r}
 	for _, a := range resolved {
 		a.EndsAt = time.Now()
 		ag.insert(a)
--- a/store/store.go
+++ b/store/store.go
@@ -123,6 +123,17 @@ func (a *Alerts) Delete(fp model.Fingerprint) error {
 	return nil
 }
 
+// DeleteIf removes the Alert with hte matching fingerprint from the store
+// if fn returns true.
+func (a *Alerts) DeleteIf(fp model.Fingerprint, fn func(*types.Alert) bool) error {
+	a.Lock()
+	defer a.Unlock()
+	if alert, ok := a.c[fp]; ok && fn(alert) {
+		delete(a.c, fp)
+	}
+	return nil
+}
+
 // List returns a slice of Alerts currently held in memory.
 func (a *Alerts) List() []*types.Alert {
 	a.Lock()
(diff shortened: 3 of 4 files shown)


store/store.go
```diff
@@ -123,6 +123,17 @@ func (a *Alerts) Delete(fp model.Fingerprint) error {
...
 
+// DeleteIf removes the Alert with the matching fingerprint from the store
+// if fn returns true.
+func (a *Alerts) DeleteIf(fp model.Fingerprint, fn func(*types.Alert) bool) error {
+	a.Lock()
+	defer a.Unlock()
+	if alert, ok := a.c[fp]; ok && fn(alert) {
+		delete(a.c, fp)
+	}
+	return nil
```
zecke, line 134: Do you have plans on returning an error in the near future?

store/store_test.go
```diff
@@ -54,6 +54,32 @@ func TestDelete(t *testing.T) {
 	require.Equal(t, ErrNotFound, err)
 }
 
+func TestDeleteIf(t *testing.T) {
+	a := NewAlerts()
+	alert := &types.Alert{
+		UpdatedAt: time.Now(),
+	}
+	require.NoError(t, a.Set(alert))
```
zecke, line 62: Should we add another alert and verify it isn't deleted (by accident)?

zecke:
LGTM. Definitely closes a time of check/time of use kind of problem. Now Set/DeleteIf/gc are protected by the same mutex.

store/store.go
grobinson-grafana: Nope! TBH I just added it because `Delete` has the same signature, even though it never returns an error:

```go
func (a *Alerts) Delete(fp model.Fingerprint) error {
```

dispatch/dispatch.go
```diff
@@ -517,19 +517,11 @@ func (ag *aggrGroup) flush(notify func(...*types.Alert) bool) {
 
 	if notify(alertsSlice...) {
 		for _, a := range alertsSlice {
-			// Only delete if the fingerprint has not been inserted
```
gotjosh, line 520: [nit] How do you feel about making the whole operation atomic? You'd only take the lock once guaranteed that all the alerts are cleared in one go. I feel this makes it easier to reason about it as there's no use case where an alert could get updated once we start checking. 

```git
index d78fbf68..6e44fbc2 100644
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -516,14 +516,7 @@ func (ag *aggrGroup) flush(notify func(...*types.Alert) bool) {
 	level.Debug(ag.logger).Log("msg", "flushing", "alerts", fmt.Sprintf("%v", alertsSlice))
 
 	if notify(alertsSlice...) {
-		for _, a := range alertsSlice {
-			// Only delete if the fingerprint has not been inserted again since we notified about it.
-			if err := ag.alerts.DeleteIf(a.Fingerprint(), func(b *types.Alert) bool {
-				return a.Resolved() && a.UpdatedAt == b.UpdatedAt
-			}); err != nil {
-				level.Error(ag.logger).Log("msg", "error on delete alert", "err", err, "alert", a.String())
-			}
-		}
+		ag.alerts.DeleteResolved(alertsSlice)
 	}
 }
 
diff --git a/store/store.go b/store/store.go
index bffa4eb6..4110a971 100644
--- a/store/store.go
+++ b/store/store.go
@@ -154,3 +154,19 @@ func (a *Alerts) Empty() bool {
 
 	return len(a.c) == 0
 }
+
+func (a *Alerts) DeleteResolved(slice types.AlertSlice) {
+	a.Lock()
+	defer a.Unlock()
+
+	for _, as := range slice {
+		fp := as.Fingerprint()
+		alert, ok := a.c[as.Fingerprint()]
+		if !ok {
+		}
+
+		if as.Resolved() && as.UpdatedAt == alert.UpdatedAt {
+			delete(a.c, fp)
+		}
+	}
+}

```

store/store.go
```diff
@@ -123,6 +123,17 @@ func (a *Alerts) Delete(fp model.Fingerprint) error {
 	return nil
 }
```
gotjosh, line 125: If we chose to go with my approach or your approach - we'd need to remove the `Delete` method as it is no longer used.

gotjosh: I think the same holds true for `Get`, which is also used in `store` but I don't see it used anywhere else (but tests).

gotjosh (approved):
LGTM, but please see my comments.

Great job @grobinson-grafana !

store/store.go
grobinson-grafana: I don't have a strong opinion here, and I'm happy to remove `Get` and `Delete`, but I did just want to raise a counter-argument for this to see if we still think it makes sense.

I understand that `Get` and `Delete` will be un-used once we merge this PR, but does that mean we should delete them? These operations seem kind of important for a store interface, even if we don't make use of them today. It's more of a question of semantics rather than pragmatics, but I feel it's important to at least discuss it.

dispatch/dispatch.go
grobinson-grafana: I quite like the condition being a closure, so what if we had something like:

```go
func (a *Alerts) DeleteFingerprints(fp []model.Fingerprint, fn func(int, *types.Alert) bool) error {
	...
}
```

I did think about adding something like `DeleteResolved` but with fingerprints:

```go
func (a *Alerts) DeleteResolved(fp []model.Fingerprint) error {
	...
}
```

but I decided I didn't like the semantics of it because it brought business logic from the dispatcher [1] into the store, and I felt the store should be agnostic of that kind of business logic.

[1] For example:

1. The name `DeleteResolved` felt confusing as you would expect a store with a method called `DeleteResolved` to delete all resolved alerts it knows about, not the resolved alerts the caller knows about (it doesn't delete all resolved alerts, just those in the argument).
2. If the caller thinks the alert is firing, but the store thinks the alert is resolved, then `DeleteResolved` will not delete the alert. This feels weird given the method is called `DeleteResolved`.
3. A resolved alert is deleted iff UpdatedAt hasn't changed since the last flush (but why does a generic store care about flushes?).

What do you think?

gotjosh: That's a fair point. 

I'm OK with:

> func (a *Alerts) DeleteFingerprints(fp []model.Fingerprint, fn func(int, *types.Alert) bool) error {

As what I care about is the atomicity of the operation. 

I think your point about logic is valid but they both share the same types anyways 🤷 so I kind feel that where the logic goes doesn't matter to me as much.

store/store.go
grobinson-grafana: OK! 🙂 `Get` and `Delete` have been deleted! 👍

grobinson-grafana: Oops! It turns out I cannot remove `Get` because it's used in `provider/mem/mem.go`.

gotjosh: Yeah, but if you look at the places where that Get is used you'll realise it's only in tests.

grobinson-grafana: It's also used on Line 207 of `provider/mem/mem.go` which isn't a test?

dispatch/dispatch_test.go
```diff
@@ -236,9 +236,33 @@ func TestAggrGroup(t *testing.T) {
 		}
 	}
 
-	// Resolve all alerts, they should be removed after the next batch was sent.
-	a1r, a2r, a3r := *a1, *a2, *a3
-	resolved := types.AlertSlice{&a1r, &a2r, &a3r}
+	// Resolve an alert, and it should be removed after the next batch was sent.
```
grobinson-grafana, line 239: I wanted to add this additional test for completeness.

store/store.go
gotjosh: Sorry, I could have been clearer. I meant the `Get` on this file. 

```diff
diff --git a/provider/mem/mem.go b/provider/mem/mem.go
index cfc3bfc3..b84133e4 100644
--- a/provider/mem/mem.go
+++ b/provider/mem/mem.go
@@ -20,12 +20,10 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
-	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/common/model"
-
 	"github.com/prometheus/alertmanager/provider"
 	"github.com/prometheus/alertmanager/store"
 	"github.com/prometheus/alertmanager/types"
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 const alertChannelLength = 200
@@ -190,11 +188,6 @@ func (a *Alerts) GetPending() provider.AlertIterator {
 	return provider.NewAlertIterator(ch, done, nil)
 }
 
-// Get returns the alert for a given fingerprint.
-func (a *Alerts) Get(fp model.Fingerprint) (*types.Alert, error) {
-	return a.alerts.Get(fp)
-}
-
 // Put adds the given alert to the set.
 func (a *Alerts) Put(alerts ...*types.Alert) error {
 	for _, alert := range alerts {

```

gotjosh (approved):
LGTM

store/store.go
gotjosh: Discussed this offline, we can't remove it because due to the interface this struct implements.

Issue #3834, opened 2024-05-02
FR: sort matchers on silences page


rc5hack:
Currently matchers are not sorted on "silences" page. I ask to add sorting.

**What did you do?**
Create 2 silences with matchers: 

1. instance=foo job=myjob alertname=InstanceDown
2. alertname=InstanceDown instance=bar job=myjob

Go to alertmanager silences page.

**What did you expect to see?**
Sorted matchers, like this:
1. alertname=InstanceDown instance=foo job=myjob
2. alertname=InstanceDown instance=bar job=myjob

**What did you see instead? Under which circumstances?**
Unsorted matchers:
1. instance=foo job=myjob alertname=InstanceDown
2. alertname=InstanceDown instance=bar job=myjob

**Environment**
* Alertmanager version:
0.27.0 (0aa3c2aad14cff039931923ab16b26b7481783b5)

* Prometheus version:
2.51.2 (b4c0ab52c3e9b940ab803581ddae9b3d9a452337)

grobinson-grafana:
@rc5hack sounds fine to me! Can you open a PR for this?

SoloJacobs:
Not opposed to this change in general, but could one of those interested clarify the intent here? @rc5hack @TheMeier @hagen1778

PR #3842, opened 2024-05-15
Add silence subcommand output sorting

linuxgcc:
Sort command:`amtool silence query`output column `Matchers` by Name #3834


gotjosh:
This looks good to me, can you please add a test?

PR #3843, opened 2024-05-15
Sorting API interface output data

linuxgcc:
When the browser accesses `silence`, the output column `Matchers` is not sorted.
When the browser accesses `alert`, the output column `Labels` is not sorted.
This pr completes the above sorting 
#3834


api/v2/compat.go
```diff
@@ -180,8 +181,15 @@ func OpenAPIAlertsToAlerts(apiAlerts open_api_models.PostableAlerts) []*types.Al
...
+
+	keys := make([]string, 0, len(modelLabelSet))
 	for key, value := range modelLabelSet {
-		apiLabelSet[string(key)] = string(value)
+		keys = append(keys, string(key))
+		_ = value
+	}
+	sort.Strings(keys)
+	for _, key := range keys {
+		apiLabelSet[string(key)] = string(modelLabelSet[prometheus_model.LabelName(key)])
```
grobinson-grafana, line 192: Hi! 👋 Thanks for opening a PR, however I'm not sure this works. `apiLabelSet` is a map which is not sorted, so while you have sorted the keys, when you put them back into `apiLabelSet` they become unsorted again.

grobinson-grafana: Yes but when `apiLabelSet` is JSON encoded it will be done so in random order because in Go maps are not sorted.

linuxgcc: tks for your prompt reply, that's great, but I don't quite understand your question.
If the `apiLabelSet`  map is  sorted, the JSON encoded results are also sorted, in the same order as the map。
As for the randomed `modelLabelSet`, we don't need to pay attention.

grobinson-grafana: > If the apiLabelSet map is sorted, the JSON encoded results are also sorted, in the same order as the map

That's the issue 🙂 The `apiLabelSet` map is not sorted. Even though you are inserting keys in a sorted order, they get unsorted by the internal hash function within the map. You cannot sort maps in Go.

grobinson-grafana: My suggestion is if you want them to appear sorted in the UI the best place to do it is in the Elm code. You can find it in `ui/app`.

linuxgcc: Go's map is not sorted, but if the key of the map is of string type, it will be sorted by key when using json.Marshal encoding, so there is no need to sort the map

grobinson-grafana: `json.Marshal` sorts maps – I learned something! 😄

cli/format/sort.go
```diff
@@ -52,3 +53,11 @@ func (s ByAddress) Less(i, j int) bool {
 	}
 	return bytes.Compare(net.ParseIP(ip1), net.ParseIP(ip2)) < 0
 }
+
+type SMatcher []*silencepb.Matcher
+
+func (m SMatcher) Len() int      { return len(m) }
+func (m SMatcher) Swap(i, j int) { m[i], m[j] = m[j], m[i] }
+func (m SMatcher) Less(i, j int) bool {
+	return m[i].Name < m[j].Name
```
grobinson-grafana, line 62: You need to make this sort stable as Alertmanager does not force name to be unique in silences. For example, `foo="bar"` and `foo=~[a-zA-Z]+` can be used in the same silence. I don't think a lot of users do this, but the sort is non-deterministic when two names are the same.

cli/format/sort.go
```diff
@@ -52,3 +53,11 @@ func (s ByAddress) Less(i, j int) bool {
 	}
 	return bytes.Compare(net.ParseIP(ip1), net.ParseIP(ip2)) < 0
 }
+
+type SMatcher []*silencepb.Matcher
```
grobinson-grafana, line 57: I would move this to the `api` package since it's unrelated to the CLI?

cli/format/sort.go
linuxgcc: For example `foo="bar"`, in order to stabilize the sort, we need to construct a string such as` "foo=bar"`, and then sort the result, not just `"foo"`

PR #3851, opened 2024-05-31
Fix incorrect use of fmt.Errorf with no args

grobinson-grafana:
This pull request fixes incorrect use of `fmt.Errorf` when no args are passed. It should be `errors.New`.


provider/mem/mem.patch
```diff
@@ -0,0 +1,53 @@
+diff --git a/provider/mem/mem.go b/provider/mem/mem.go
```
zecke, line 1: That seems to be committed in error.

PR #3852, opened 2024-05-31
Add limits for silences

grobinson-grafana:
This commit adds limits for silences including the maximum number of active and pending silences, and the maximum size per silence (in bytes).

PR #3862, opened 2024-06-03
Limits should include expired silences

grobinson-grafana:
The limits we added in #3852 should include expired silences.

### DIFF
--- a/silence/silence_test.go
+++ b/silence/silence_test.go
@@ -465,8 +465,12 @@ func TestSilenceLimits(t *testing.T) {
 			MaxSilences:        1,
 			MaxPerSilenceBytes: 2 << 11, // 4KB
 		},
+		Retention: 100 * time.Millisecond,
 	})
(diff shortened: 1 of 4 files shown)


silence/silence_test.go
```diff
@@ -465,8 +465,12 @@ func TestSilenceLimits(t *testing.T) {
 			MaxSilences:        1,
 			MaxPerSilenceBytes: 2 << 11, // 4KB
 		},
+		Retention: 100 * time.Millisecond,
 	})
 	require.NoError(t, err)
+	stopCh := make(chan struct{})
+	defer close(stopCh)
+	go s.Maintenance(100*time.Millisecond, "", stopCh, nil)
```
grobinson-grafana, line 473: We cannot force a maintenance without running the goroutine, hence the calls to `time.Sleep`. I think this is fine, and I can improve the tests later.

gotjosh (approved):
LGTM

PR #3866, opened 2024-06-05
Fix TestSilenceLimits tests

grobinson-grafana:
This commit fixes silence tests that relied on the maintenance function running at a fixed 100ms interval. If the go runtime that runs the maintenance is not scheduled with 150ms then the test will fail.

### DIFF
--- a/silence/silence_test.go
+++ b/silence/silence_test.go
@@ -465,12 +465,8 @@ func TestSilenceLimits(t *testing.T) {
 			MaxSilences:        1,
 			MaxPerSilenceBytes: 2 << 11, // 4KB
 		},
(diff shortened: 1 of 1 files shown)


silence/silence_test.go
```diff
@@ -465,12 +465,7 @@ func TestSilenceLimits(t *testing.T) {
 			MaxSilences:        1,
 			MaxPerSilenceBytes: 2 << 11, // 4KB
 		},
-		Retention: 100 * time.Millisecond,
```
grobinson-grafana, line 468: I disabled the retention period so I don't need `time.Sleep` in the tests.

PR #3868, opened 2024-06-06
Fix inconsistent Id returned for silences.Set

grobinson-grafana:
This pull request fixes inconsistent behavior of `silences.Set()` where it can return an empty string and an error but also set `sil.Id` to a UUID. This can happen when creating a silence that is invalid or exceeds limits. This bug was introduced in #3852.

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -621,13 +621,6 @@ func (s *Silences) Set(sil *pb.Silence) (string, error) {
 		}
 	}
 
-	// If we got here it's either a new silence or a replacing one.
-	if s.limits.MaxSilences > 0 {
-		if len(s.st)+1 > s.limits.MaxSilences {
-			return "", fmt.Errorf("exceeded maximum number of silences: %d (limit: %d)", len(s.st), s.limits.MaxSilences)
-		}
-	}
-
 	uid, err := uuid.NewV4()
 	if err != nil {
 		return "", fmt.Errorf("generate uuid: %w", err)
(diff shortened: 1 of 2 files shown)


silence/silence.go
```diff
@@ -638,11 +631,14 @@ func (s *Silences) Set(sil *pb.Silence) (string, error) {
 		sil.StartsAt = now
 	}
 
-	if err = s.setSilence(sil, now, false); err != nil {
-		return "", err
+	// If we got here it's either a new silence or a replacing one.
+	if s.limits.MaxSilences > 0 {
+		if len(s.st)+1 > s.limits.MaxSilences {
+			return sil.Id, fmt.Errorf("exceeded maximum number of silences: %d (limit: %d)", len(s.st), s.limits.MaxSilences)
```
grobinson-grafana, line 637: I return `sil.Id` here as that's what happens if `s.setSilence` returns an error too. `s.setSilence` can return an error if the silence is invalid or maximum size has been exceeded.

grobinson-grafana:
I'm going to close this PR as I think the correct behavior should be `sil.Id` is not set when creating a silence fails.

grobinson-grafana:
I closed and then re-opened this PR as what I want to do is make sure `sil.Id` is `""` when creating a silence fails, but making this consistent is much more complicated than I first thought because of how updates (expire then re-create) works.

silence/silence.go
```diff
@@ -638,11 +631,14 @@ func (s *Silences) Set(sil *pb.Silence) (string, error) {
...
+		}
 	}
 
-	return sil.Id, nil
+	return sil.Id, s.setSilence(sil, now, false)
```
grobinson-grafana, line 641: This is how it was before https://github.com/prometheus/alertmanager/pull/3866.

grobinson-grafana:
I'm going to close this in favor of doing the bigger changes, as mentioned [here](https://github.com/prometheus/alertmanager/pull/3868#issuecomment-2153012396).

PR #3870, opened 2024-06-06
feat: add opsgenie priority template based on severity label

ncrmro:
P1 in my opinion is reserved for actual outage, in our case if cloudflare fails the external domain health check though I'm open to updating this _default_.

### DIFF
--- a/template/default.tmpl
+++ b/template/default.tmpl
@@ -52,7 +52,9 @@ Alerts Resolved:
 {{- end }}
 {{- end }}
 {{ define "opsgenie.default.source" }}{{ template "__alertmanagerURL" . }}{{ end }}
-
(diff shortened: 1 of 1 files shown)


grobinson-grafana:
Hi! 👋 Thanks for opening this PR. As mentioned [here](https://github.com/prometheus/alertmanager/pull/3847), I think the same applies to `priority` label:

> Alertmanager doesn't have severity, and Prometheus users are not required to set use a severity label, so I'm not sure if this is something that will be accepted. Is it not possible to template the severity for Opsgenie using the existing templating infrastructure?

Why not use a custom template to set the priority field in Opsgenie for your use case?

ncrmro:
@grobinson-grafana we did just that!

PR #3877, opened 2024-06-14
Fix MaxSilences limit causes incomplete updates of existing silences

grobinson-grafana:
This commit fixes a bug where the MaxSilences limit can cause an incomplete update of existing silences, where the old silence can be expired but the new silence cannot be created because it would exceeded the maximum number of silences.

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -564,6 +564,13 @@ func (s *Silences) getSilence(id string) (*pb.Silence, bool) {
 	return msil.Silence, true
 }
 
+func (s *Silences) toMeshSilence(sil *pb.Silence) *pb.MeshSilence {
+	return &pb.MeshSilence{
+		Silence:   sil,
+		ExpiresAt: sil.EndsAt.Add(s.retention),
+	}
+}
+
 func (s *Silences) setSilence(sil *pb.Silence, now time.Time, skipValidate bool) error {
 	sil.UpdatedAt = now
 
@@ -573,10 +580,7 @@ func (s *Silences) setSilence(sil *pb.Silence, now time.Time, skipValidate bool)
 		}
 	}
 
-	msil := &pb.MeshSilence{
-		Silence:   sil,
-		ExpiresAt: sil.EndsAt.Add(s.retention),
-	}
+	msil := s.toMeshSilence(sil)
 	b, err := marshalMeshSilence(msil)
 	if err != nil {
 		return err
@@ -612,22 +616,21 @@ func (s *Silences) Set(sil *pb.Silence) error {
 		return ErrNotFound
 	}
 
-	if ok {
-		if canUpdate(prev, sil, now) {
-			return s.setSilence(sil, now, false)
-		}
-		if getState(prev, s.nowUTC()) != types.SilenceStateExpired {
-			// We cannot update the silence, expire the old one.
-			if err := s.expire(prev.Id); err != nil {
-				return fmt.Errorf("expire previous silence: %w", err)
-			}
-		}
+	if ok && canUpdate(prev, sil, now) {
+		return s.setSilence(sil, now, false)
 	}
 
 	// If we got here it's either a new silence or a replacing one.
 	if s.limits.MaxSilences != nil {
 		if m := s.limits.MaxSilences(); m > 0 && len(s.st)+1 > m {
 			return fmt.Errorf("exceeded maximum number of silences: %d (limit: %d)", len(s.st), m)
+		}
+	}
+
+	if ok && getState(prev, s.nowUTC()) != types.SilenceStateExpired {
+		// We cannot update the silence, expire the old one.
+		if err := s.expire(prev.Id); err != nil {
+			return fmt.Errorf("expire previous silence: %w", err)
 		}
 	}
 
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -620,7 +620,9 @@ func (s *Silences) Set(sil *pb.Silence) error {
 		return s.setSilence(sil, now, false)
 	}
 
-	// If we got here it's either a new silence or a replacing one.
+	// If we got here it's either a new silence or a replacing one (which would
+	// also create a new silence) so we need to make sure we have capacity for
+	// the new silence.
 	if s.limits.MaxSilences != nil {
 		if m := s.limits.MaxSilences(); m > 0 && len(s.st)+1 > m {
 			return fmt.Errorf("exceeded maximum number of silences: %d (limit: %d)", len(s.st), m)
@@ -628,7 +630,8 @@ func (s *Silences) Set(sil *pb.Silence) error {
 	}
 
 	if ok && getState(prev, s.nowUTC()) != types.SilenceStateExpired {
-		// We cannot update the silence, expire the old one.
+		// We cannot update the silence, expire the old one to leave a history of
+		// the silence before modification.
 		if err := s.expire(prev.Id); err != nil {
 			return fmt.Errorf("expire previous silence: %w", err)
 		}
--- a/silence/silence_test.go
+++ b/silence/silence_test.go
@@ -15,6 +15,7 @@ package silence
 
 import (
 	"bytes"
+	"fmt"
 	"os"
 	"runtime"
 	"sort"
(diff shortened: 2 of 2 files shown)


silence/silence_test.go
```diff
@@ -492,10 +492,7 @@ func TestSilenceLimits(t *testing.T) {
 	require.NoError(t, s.Set(sil2))
 	require.NotEqual(t, "", sil2.Id)
 
-	// Should be able to update sil2 without hitting the limit.
```
grobinson-grafana, line 495: Moved this check to Line 534.

silence/silence_test.go
```diff
@@ -523,8 +520,44 @@ func TestSilenceLimits(t *testing.T) {
 	// Do not check the exact size as it can change between consecutive runs
 	// due to padding.
 	require.Contains(t, err.Error(), "silence exceeded maximum size")
-	// TODO: Once we fix https://github.com/prometheus/alertmanager/issues/3878 this should be require.Equal.
```
grobinson-grafana, line 526: I removed this comment as making this transactional is proving more difficult then I expected.

silence/silence.go
```diff
@@ -612,16 +612,8 @@ func (s *Silences) Set(sil *pb.Silence) error {
 		return ErrNotFound
 	}
 
-	if ok {
-		if canUpdate(prev, sil, now) {
-			return s.setSilence(sil, now, false)
-		}
-		if getState(prev, s.nowUTC()) != types.SilenceStateExpired {
```
grobinson-grafana, line 619: The entire change here is moving the expiration of the existing silence to run after checking the `MaxSilences()` limit. This ensures that an existing silence is not expired if we are at the limit, as what would happen is the existing silence is expired and then the new silence cannot be created.

silence/silence.go
```diff
@@ -612,16 +612,8 @@ func (s *Silences) Set(sil *pb.Silence) error {
...
-			// We cannot update the silence, expire the old one.
-			if err := s.expire(prev.Id); err != nil {
-				return fmt.Errorf("expire previous silence: %w", err)
-			}
-		}
+	if ok && canUpdate(prev, sil, now) {
```
grobinson-grafana, line 615: This code should run before checking the `MaxSilences()` limit because we don't want to prevent updating an existing silence if we are at the limit and the update does not require expiring and creating a new silence.

silence/silence_test.go
```diff
@@ -523,8 +520,44 @@ func TestSilenceLimits(t *testing.T) {
...
+	// expired silences.
+	sil6 := cloneSilence(sil5)
+	sil6.StartsAt = time.Now().Add(5 * time.Minute)
+	sil6.EndsAt = time.Now().Add(10 * time.Minute)
+	err = s.Set(sil6)
+	require.Error(t, err)
+	require.EqualError(t, err, "exceeded maximum number of silences: 1 (limit: 1)")
+	require.NotEqual(t, "", sil6.Id)
+
+	// sil5 should not be expired because the update failed.
```
grobinson-grafana, line 557: This here shows that the existing silence is not expired if we are at the limit.

grobinson-grafana:
This fix is included in https://github.com/prometheus/alertmanager/pull/3892.

silence/silence.go
```diff
@@ -564,6 +564,13 @@ func (s *Silences) getSilence(id string) (*pb.Silence, bool) {
 	return msil.Silence, true
 }
 
+func (s *Silences) toMeshSilence(sil *pb.Silence) *pb.MeshSilence {
```
grobinson-grafana, line 567: I've moved this to a function for two reasons:

1. It allows us to check the full error in unit tests. For example:

    ```go
    require.EqualError(t, s.Set(sil3), fmt.Sprintf("silence exceeded maximum size: %d bytes (limit: 4096 bytes)", s.toMeshSilence(sil3).Size()))
    ```

2. We will use it in the next PR where we fix the same bug but for maximum size limits and validation errors.

silence/silence.go
```diff
@@ -612,16 +616,8 @@ func (s *Silences) Set(sil *pb.Silence) error {
...
-		if canUpdate(prev, sil, now) {
-			return s.setSilence(sil, now, false)
-		}
-		if getState(prev, s.nowUTC()) != types.SilenceStateExpired {
-			// We cannot update the silence, expire the old one.
-			if err := s.expire(prev.Id); err != nil {
-				return fmt.Errorf("expire previous silence: %w", err)
-			}
-		}
+	if ok && canUpdate(prev, sil, now) {
```
grobinson-grafana, line 619: This code should run before checking the `MaxSilences()` limit because we don't want to prevent updating an existing silence if we are at the limit and the update does not require expiring and creating a new silence.

silence/silence.go
```diff
@@ -631,6 +627,13 @@ func (s *Silences) Set(sil *pb.Silence) error {
 		}
 	}
 
+	if ok && getState(prev, s.nowUTC()) != types.SilenceStateExpired {
```
grobinson-grafana, line 630: The entire change here is moving the expiration of the existing silence to run after checking the `MaxSilences()` limit. This ensures that an existing silence is not expired if we are at the limit, as what would happen is the existing silence is expired and then the new silence cannot be created.

silence/silence_test.go
```diff
@@ -470,32 +471,24 @@ func TestSilenceLimits(t *testing.T) {
...
+	// Insert sil2 should fail because maximum number of silences has been
+	// exceeded.
 	sil2 := &pb.Silence{
-		Matchers: []*pb.Matcher{{Name: "a", Pattern: "b"}},
+		Matchers: []*pb.Matcher{{Name: "c", Pattern: "d"}},
 		StartsAt: time.Now(),
 		EndsAt:   time.Now().Add(5 * time.Minute),
 	}
 	require.EqualError(t, s.Set(sil2), "exceeded maximum number of silences: 1 (limit: 1)")
-	require.Equal(t, "", sil2.Id)
```
grobinson-grafana, line 483: I have removed all checks for `require.Equal(t, "", id)` in this test because it's unrelated to limits.

silence/silence_test.go
```diff
@@ -505,26 +498,55 @@ func TestSilenceLimits(t *testing.T) {
...
-		CreatedBy: strings.Repeat("e", 2<<9),
-		Comment:   strings.Repeat("f", 2<<9),
+		CreatedBy: strings.Repeat("i", 2<<9),
+		Comment:   strings.Repeat("j", 2<<9),
 		StartsAt:  time.Now(),
 		EndsAt:    time.Now().Add(5 * time.Minute),
 	}
-	err = s.Set(sil3)
-	require.Error(t, err)
-	// Do not check the exact size as it can change between consecutive runs
```
grobinson-grafana, line 523: Fixed this.

silence/silence_test.go
```diff
@@ -505,26 +498,55 @@ func TestSilenceLimits(t *testing.T) {
...
-	// due to padding.
-	require.Contains(t, err.Error(), "silence exceeded maximum size")
-	// TODO: Once we fix https://github.com/prometheus/alertmanager/issues/3878 this should be require.Equal.
-	require.NotEqual(t, "", sil3.Id)
+	require.EqualError(t, s.Set(sil3), fmt.Sprintf("silence exceeded maximum size: %d bytes (limit: 4096 bytes)", s.toMeshSilence(sil3).Size()))
+
+	// Should be able to insert sil4.
```
grobinson-grafana, line 516: I added additional tests here to cover the following cases:

1. Updating an existing silence without modifications.
2. Adding a comment to an existing silence.
3. Failing to update an existing silence because the maximum number of silences has been exceeded should not expire the existing silence.

silence/silence.go
```diff
@@ -612,16 +616,8 @@ func (s *Silences) Set(sil *pb.Silence) error {
...
-			// We cannot update the silence, expire the old one.
-			if err := s.expire(prev.Id); err != nil {
-				return fmt.Errorf("expire previous silence: %w", err)
-			}
-		}
+	if ok && canUpdate(prev, sil, now) {
+		return s.setSilence(sil, now, false)
 	}
 
 	// If we got here it's either a new silence or a replacing one.
```
gotjosh, line 623:
```suggestion
	// If we got here it's either a new silence or a replacing an existing one (which would also create a new silence) so we need to make sure we have capacity for the new silence.
```

silence/silence.go
```diff
@@ -631,6 +627,13 @@ func (s *Silences) Set(sil *pb.Silence) error {
 		}
 	}
 
+	if ok && getState(prev, s.nowUTC()) != types.SilenceStateExpired {
+		// We cannot update the silence, expire the old one.
```
gotjosh, line 631:
```suggestion
		// We cannot update the silence, expire the old one to leave a history of the silence before modification. 
```

gotjosh (approved):
LGTM

PR #3881, opened 2024-06-17
Remove Id return from silences.Set(*pb.Silence)

grobinson-grafana:
This commit removes the Id from the method `silences.Set(*pb.Silence)` as it is redundant. The Id is still set even when creating a silence fails. This will be fixed in a later change.


gotjosh (approved):
LGTM

simonpasquier (approved):

PR #3882, opened 2024-06-17
Introduce AlertSnapshot and AlertsSnapshot to create a snapshot of the alert

zecke:
Introduce types.AlertSnapshot and types.AlertsSnapshot that keeps track of the original dispatch time and uses it for the Status() and Resolved() method. This makes sure the notification is consistent across retries and/or other delays.

This is in preparation to not zero out the EndsAt time for active alerts.


zecke:
Any chance we can make progress here?

PR #3885, opened 2024-06-20
Silence limits as functions

grobinson-grafana:
This commit changes silence limits from a struct of ints to a struct of functions that return individual limits. This allows limits to be lazy-loaded and updated without having to call `silences.New()`.


gotjosh (approved):
LGTM

PR #3896, opened 2024-06-24
Improve test coverage for silences

grobinson-grafana:
This commit improves the existing test coverage for silences to cover a number of additional cases, and also improve the comments of existing cases.


gotjosh (approved):
LGTM

PR #3892, opened 2024-06-22
Issue #3878: Fix bugs in silences

grobinson-grafana:
This commit fixes a number of bugs that can occur when creating and updating silences:

1. When creating a silence fails because the silence is invalid or would exceed limits, subsequent calls for the same silence fail even when the validation or limits are fixed. This happens because the silence is assigned an ID that does not exist. This commit fixes that issue as it assigns a new ID to silences that do not exist.

2. When updating an existing silence, and the update changes how alerts are matched or the time window in which the silence is active (excluding time extensions), then the existing silence must be expired and a new silence created. However, if the new silence is invalid or would exceed limits, the existing silence is expired, but the new silence is not created. This commit fixes that issue such that existing silences are not expired unless the replacing silence is both valid and within limits.

Fixes #3877

PR #3897, opened 2024-06-24
Fix MaxSilenceSizeBytes limit causes incomplete updates of existing silences

grobinson-grafana:
This commit fixes a bug where the `MaxSilenceSizeBytes` limit can cause an incomplete update of existing silences, where the old silence can be expired but the new silence cannot be created because it would exceed the maximum size limit.

I moved `sil.UpdatedAt = now` out of `setSilence` and into `Set` and `Expire` because we want to make sure the `UpdatedAt` timestamp is calculated in the size when we call `checkSizeLimits`. Forget to set `sil.UpdatedAt` in either of these places and the following tests fail:

```
--- FAIL: TestSilenceSet (0.00s)
--- FAIL: TestSilenceLimits (0.00s)
--- FAIL: TestSetActiveSilence (0.00s)
--- FAIL: TestSilenceExpire (0.00s)
--- FAIL: TestSilenceExpireWithZeroRetention (0.00s)
--- FAIL: TestSilenceExpireInvalid (0.00s)
```

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -550,6 +550,16 @@ func cloneSilence(sil *pb.Silence) *pb.Silence {
 	return &s
 }
 
+func (s *Silences) checkSizeLimits(msil *pb.MeshSilence) error {
+	if s.limits.MaxSilenceSizeBytes != nil {
+		n := msil.Size()
+		if m := s.limits.MaxSilenceSizeBytes(); m > 0 && n > m {
+			return fmt.Errorf("silence exceeded maximum size: %d bytes (limit: %d bytes)", n, m)
+		}
+	}
+	return nil
+}
+
 func (s *Silences) getSilence(id string) (*pb.Silence, bool) {
 	msil, ok := s.st[id]
 	if !ok {
@@ -565,30 +575,15 @@ func (s *Silences) toMeshSilence(sil *pb.Silence) *pb.MeshSilence {
 	}
 }
 
-func (s *Silences) setSilence(sil *pb.Silence, now time.Time) error {
-	sil.UpdatedAt = now
-
-	msil := s.toMeshSilence(sil)
+func (s *Silences) setSilence(msil *pb.MeshSilence, now time.Time) error {
 	b, err := marshalMeshSilence(msil)
 	if err != nil {
 		return err
 	}
-
-	// Check the limit unless the silence has been expired. This is to avoid
-	// situations where silences cannot be expired after the limit has been
-	// reduced.
-	if s.limits.MaxSilenceSizeBytes != nil {
-		n := msil.Size()
-		if m := s.limits.MaxSilenceSizeBytes(); m > 0 && n > m && sil.EndsAt.After(now) {
-			return fmt.Errorf("silence exceeded maximum size: %d bytes (limit: %d bytes)", n, m)
-		}
-	}
-
 	if s.st.merge(msil, now) {
 		s.version++
 	}
 	s.broadcast(b)
-
 	return nil
 }
 
@@ -613,7 +608,12 @@ func (s *Silences) Set(sil *pb.Silence) error {
 	}
 
 	if ok && canUpdate(prev, sil, now) {
-		return s.setSilence(sil, now)
+		sil.UpdatedAt = now
+		msil := s.toMeshSilence(sil)
+		if err := s.checkSizeLimits(msil); err != nil {
+			return err
+		}
+		return s.setSilence(msil, now)
 	}
 
 	// If we got here it's either a new silence or a replacing one (which would
@@ -622,14 +622,6 @@ func (s *Silences) Set(sil *pb.Silence) error {
 	if s.limits.MaxSilences != nil {
 		if m := s.limits.MaxSilences(); m > 0 && len(s.st)+1 > m {
 			return fmt.Errorf("exceeded maximum number of silences: %d (limit: %d)", len(s.st), m)
-		}
-	}
-
-	if ok && getState(prev, s.nowUTC()) != types.SilenceStateExpired {
-		// We cannot update the silence, expire the old one to leave a history of
-		// the silence before modification.
-		if err := s.expire(prev.Id); err != nil {
-			return fmt.Errorf("expire previous silence: %w", err)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


silence/silence.go
```diff
@@ -518,9 +518,6 @@ func matchesEmpty(m *pb.Matcher) bool {
 }
 
 func validateSilence(s *pb.Silence) error {
-	if s.Id == "" {
```
grobinson-grafana, line 521: I have removed checks for the `Id` and `UpdatedAt` fields from `validateSilence`. There are two reasons for this:

1. I have moved validation of silences to the start of the function before the `Id` and `UpdatedAt` fields are assigned. This is to fix the issue where an existing silence is expired but the new silence is not created.
2. These fields are supposed to be set in `silences.Set(*pb.Silence)`, not by the calling code. We assert that these are present within tests, not validation code that covers both internal and external fields.

silence/silence.go
```diff
@@ -611,43 +602,55 @@ func (s *Silences) Set(sil *pb.Silence) error {
 	defer s.mtx.Unlock()
 
 	now := s.nowUTC()
-	prev, ok := s.getSilence(sil.Id)
+	if sil.StartsAt.IsZero() {
```
grobinson-grafana, line 605: This is to cover the following test case https://github.com/prometheus/alertmanager/blob/v0.27.0/silence/silence_test.go#L342.

silence/silence.go
```diff
@@ -611,43 +602,55 @@ func (s *Silences) Set(sil *pb.Silence) error {
...
+	}
+
+	existing, ok := s.getSilence(sil.Id)
 	if sil.Id != "" && !ok {
 		return ErrNotFound
 	}
 
-	if ok && canUpdate(prev, sil, now) {
-		return s.setSilence(sil, now, false)
+	// updateExisting is true iff this is an existing silence, and the update does
```
grobinson-grafana, line 618: In the end I kept all this code. Here is the reasoning:

1. I've done a lot of reasoning and manual testing, in additional to the unit tests, and I'm more confident about it then other implements I tried in the last week. This implementation is the outcome of about 5 hours of thinking and experimenting how I can solve this problem as cleanly as possible.
2. I think now that we've made the fixes more incremental it's easier to review then it was in #3892.

Please let me know if this is a problem and I'll see how I can rewrite the code to be more like the original.

silence/silence.go
```diff
@@ -611,43 +602,55 @@ func (s *Silences) Set(sil *pb.Silence) error {
...
+	// partial updates when limits are exceeded. Limits are checked against
+	// the mesh silence as this is the silence that is gossiped and stored
+	// on disk.
+	msil := s.toMeshSilence(sil)
+	if err := s.checkLimits(msil, updateExisting); err != nil {
+		return err
 	}
 
-	if ok && getState(prev, s.nowUTC()) != types.SilenceStateExpired {
+	if ok && !updateExisting && getState(existing, s.nowUTC()) != types.SilenceStateExpired {
```
grobinson-grafana, line 645: As we mentioned this morning, I don't think we need the check for `getState(existing, s.nowUTC()) != types.SilenceStateExpired`. But I don't want to introduce a possible regression here, so I will look into that another time.

silence/silence_test.go
```diff
@@ -486,6 +486,8 @@ func TestSilenceLimits(t *testing.T) {
 	n, err := s.GC()
 	require.NoError(t, err)
 	require.Equal(t, 1, n)
+	// Need to remove the Id from the previous attempt.
+	sil2.Id = ""
```
grobinson-grafana, line 490: This is required because prior to this commit the maximum silences limit was checked before an `Id` was assigned to the silence. Now limits are checked afterwards.

silence/silence_test.go
```diff
@@ -547,6 +549,33 @@ func TestSilenceLimits(t *testing.T) {
 	sil6, err = s.QueryOne(QIDs(sil6.Id))
 	require.NoError(t, err)
 	require.Equal(t, types.SilenceStateActive, getState(sil6, s.nowUTC()))
+
+	// Should not be able to update with a comment that exceeds maximum size.
```
grobinson-grafana, line 553: These are the additional test cases that make sure an existing silence is not expired when the `MaxSilenceSizeBytes` limit is exceeded.

silence/silence.go
```diff
@@ -571,36 +583,15 @@ func (s *Silences) toMeshSilence(sil *pb.Silence) *pb.MeshSilence {
 	}
 }
 
-func (s *Silences) setSilence(sil *pb.Silence, now time.Time, skipValidate bool) error {
-	sil.UpdatedAt = now
```
grobinson-grafana, line 575: Moved `sil.UpdatedAt` out of `setSilence` so when we check the limits in `CheckLimits` the full size of the `*pb.MeshSilence` is taken into account, not just the size of `*pb.Silence`.

silence/silence.go
grobinson-grafana: Here is an alternative implementation which you might prefer:

```diff
@@ -611,43 +602,58 @@ func (s *Silences) Set(sil *pb.Silence) error {
        defer s.mtx.Unlock()

        now := s.nowUTC()
-       prev, ok := s.getSilence(sil.Id)
-       if sil.Id != "" && !ok {
-               return ErrNotFound
+       if sil.StartsAt.IsZero() {
+               sil.StartsAt = now
        }

-       if ok && canUpdate(prev, sil, now) {
-               return s.setSilence(sil, now, false)
+       if err := validateSilence(sil); err != nil {
+               return fmt.Errorf("invalid silence: %w", err)
        }

-       // If we got here it's either a new silence or a replacing one (which would
-       // also create a new silence) so we need to make sure we have capacity for
-       // the new silence.
-       if s.limits.MaxSilences != nil {
-               if m := s.limits.MaxSilences(); m > 0 && len(s.st)+1 > m {
-                       return fmt.Errorf("exceeded maximum number of silences: %d (limit: %d)", len(s.st), m)
-               }
+       existing, ok := s.getSilence(sil.Id)
+       if sil.Id != "" && !ok {
+               return ErrNotFound
        }

-       if ok && getState(prev, s.nowUTC()) != types.SilenceStateExpired {
-               // We cannot update the silence, expire the old one to leave a history of
-               // the silence before modification.
-               if err := s.expire(prev.Id); err != nil {
-                       return fmt.Errorf("expire previous silence: %w", err)
+       if ok && canUpdate(existing, sil, now) {
+               sil.UpdatedAt = now
+               msil := s.toMeshSilence(sil)
+               if err := s.checkLimits(msil, true); err != nil {
+                       return err
                }
+               return s.setSilence(msil, now)
        }

+       // If this is a new silence, or it is replacing an existing silence,
+       // then give it an ID and make sure the start time is not in the past.
        uid, err := uuid.NewV4()
        if err != nil {
-               return fmt.Errorf("generate uuid: %w", err)
+               return fmt.Errorf("failed to create UUID: %w", err)
        }
        sil.Id = uid.String()
-
        if sil.StartsAt.Before(now) {
                sil.StartsAt = now
        }
+       sil.UpdatedAt = now

-       return s.setSilence(sil, now, false)
+       // Check the limits before expiring the existing silence to prevent
+       // partial updates when limits are exceeded. Limits are checked against
+       // the mesh silence as this is the silence that is gossiped and stored
+       // on disk.
+       msil := s.toMeshSilence(sil)
+       if err := s.checkLimits(msil, false); err != nil {
+               return err
+       }
+
+       if ok && getState(existing, s.nowUTC()) != types.SilenceStateExpired {
+               // We cannot update the silence, expire the old one to leave a history of
+               // the silence before modification.
+               if err := s.expire(existing.Id); err != nil {
+                       return fmt.Errorf("expire previous silence: %w", err)
+               }
+       }
+
+       return s.setSilence(msil, now)
 }
```

gotjosh (approved):

PR #3898, opened 2024-06-24
Fix invalid silence causes incomplete updates

grobinson-grafana:
This commit fixes a bug where an invalid silence causes incomplete updates of existing silences. What happens here is in some cases where the new silence is invalid, it can expire the old silence without creating the new silence. What should happen instead is the entire operation is aborted, and the original silence is left in place. This has been fixed by moving validation out of the `setSilence` method and putting it at the start of the `Set` method instead.

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -518,9 +518,6 @@ func matchesEmpty(m *pb.Matcher) bool {
 }
 
 func validateSilence(s *pb.Silence) error {
-	if s.Id == "" {
-		return errors.New("ID missing")
-	}
 	if len(s.Matchers) == 0 {
 		return errors.New("at least one matcher required")
 	}
@@ -543,9 +540,6 @@ func validateSilence(s *pb.Silence) error {
 	}
 	if s.EndsAt.Before(s.StartsAt) {
 		return errors.New("end time must not be before start time")
-	}
-	if s.UpdatedAt.IsZero() {
-		return errors.New("invalid zero update timestamp")
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


silence/silence.go
```diff
@@ -518,9 +518,6 @@ func matchesEmpty(m *pb.Matcher) bool {
 }
 
 func validateSilence(s *pb.Silence) error {
-	if s.Id == "" {
```
grobinson-grafana, line 521: I have removed checks for the `Id` and `UpdatedAt` fields from `validateSilence`. There are two reasons for this:

1. I have moved validation of silences to the start of the function before the `Id` and `UpdatedAt` fields are assigned. This is to fix the issue where an existing silence is expired but the new silence is not created.
2. These fields are supposed to be set in `silences.Set(*pb.Silence)`, not by the calling code. We assert that these are present within tests, not validation code that covers both internal and external fields.

silence/silence.go
```diff
@@ -611,13 +599,21 @@ func (s *Silences) Set(sil *pb.Silence) error {
 	defer s.mtx.Unlock()
 
 	now := s.nowUTC()
+	if sil.StartsAt.IsZero() {
```
grobinson-grafana, line 602: This is to cover the following test case https://github.com/prometheus/alertmanager/blob/v0.27.0/silence/silence_test.go#L342.

silence/silence.go
```diff
@@ -544,9 +541,6 @@ func validateSilence(s *pb.Silence) error {
 	if s.EndsAt.Before(s.StartsAt) {
 		return errors.New("end time must not be before start time")
 	}
-	if s.UpdatedAt.IsZero() {
```
grobinson-grafana, line 547: We have unit tests that check `UpdatedAt` is not zero.

silence/silence_test.go
```diff
@@ -468,6 +468,19 @@ func TestSilenceSet(t *testing.T) {
 		},
 	}
 	require.Equal(t, want, s.st, "unexpected state after silence creation")
+
+	// Updating an existing silence with an invalid silence should not expire
```
grobinson-grafana, line 472: This test fails in `main` because the original silence is expired.

gotjosh (approved):
LGTM

grobinson-grafana:
Will rebase https://github.com/prometheus/alertmanager/pull/3899 to show that no behavior has changed.

grobinson-grafana:
I just rebased main to include #3899.

PR #3899, opened 2024-06-25
Add tests for create and then update silences

grobinson-grafana:
This commit adds a unit test for the `postSilencesHandler` to create and then update a silence. It shows that changing the ID of an existing silence returns 404 Not Found, and removing the ID of an existing silence re-creates that silence with a different ID.


gotjosh (approved):
LGTM

PR #3900, opened 2024-06-25
Support Noop limits for silences

grobinson-grafana:
This commit changes how limits work for silences. Instead of checking if a limit is non-nil, it instead adds support for a set of no-op limits called `NoopLimits`.

It changes the limits in `Options` and `Silences` from `Limits` to `*Limits` because Go cannot do comparisons on structs containing function values, so we must use a nil check to see if limits are set in the options or not.


silence/silence.go
```diff
@@ -619,10 +627,8 @@ func (s *Silences) Set(sil *pb.Silence) error {
 	// If we got here it's either a new silence or a replacing one (which would
 	// also create a new silence) so we need to make sure we have capacity for
 	// the new silence.
-	if s.limits.MaxSilences != nil {
-		if m := s.limits.MaxSilences(); m > 0 && len(s.st)+1 > m {
-			return fmt.Errorf("exceeded maximum number of silences: %d (limit: %d)", len(s.st), m)
-		}
+	if m := s.limits.MaxSilences(); m > 0 && len(s.st)+1 > m {
```
zecke, line 630: What if we make this two functions? Then the standalone checkSizeLimits can be mostly removed as well. It also avoids having to replace the nil check with a `m > 0`

grobinson-grafana: I'm not 100% sure I can see what those functions would look like, can you perhaps share an example?

PR #3924, opened 2024-07-26
chore: use HumanizeTimestamp from prometheus/common

freak12techno:
Adding HumanizeDuration from prometheus/common (moved it there in this PR: https://github.com/prometheus/common/pull/654).
Also copy-pasted some test cases from Prometheus repo.


freak12techno:
@gotjosh since it was you who reviewed my last PR here, can you check or ask someone to do so?

freak12techno:
Seems like the test failing is not mine and should've been caused by my PR right?

simonpasquier:
Thanks for PR! I believe that the PR should be retitled "chore: use HumanizeTimestamp from prometheus/common". Also please add the new `humanizeTimestamp` function to docs/notifications.md.

I'm not quite sure where someone would have the opportunity to use the function in Alertmanager templates?

freak12techno:
@simonpasquier thanks for noticing the PR naming, I updated it. On docs, will also add it a bit later.
For usage, I personally have some alerts that expose unix timestamp as a value, would be nice to be able to format it in a nice way.

freak12techno:
@simonpasquier fixed the docs as well

simonpasquier:
> For usage, I personally have some alerts that expose unix timestamp as a value, would be nice to be able to format it in a nice way.

Got it but the value shouldn't be present in the data accessible to the Alertmanager templates. Or am I missing something?

freak12techno:
@simonpasquier actually yes, haven't thought of that. Do you think it'll be still useful or if I should close it?

simonpasquier:
I wouldn't add the function until we have a use case (otherwise it creates unnecessary load for users and maintainers).

freak12techno:
Gotcha, I'll close it then. Will re-add if there's a usecase for it later.

Issue #3925, opened 2024-07-28
Feature Request: Enable Sending AdaptiveCards to Webex for Enhanced Alert Notifications


florianspk:
**What did you do?**

I noticed that Alertmanager currently does not support sending AdaptiveCards to Webex, which can be very useful for enhancing the clarity and interactivity of alert notifications. Webex supports AdaptiveCards via its API with a specific format, which allows for richer message content including buttons and dropdown menus.

**What did you expect to see?**

I expected to see an option to configure Alertmanager to send alert notifications as AdaptiveCards to Webex, leveraging the Webex API format to enhance the alert details and provide interactive elements directly in the notification.

**What did you see instead? Under which circumstances?**

Currently, Alertmanager can send basic messages to Webex, but it lacks support for the richer formatting and interactivity provided by AdaptiveCards.

**Environment**

* System information:

	Linux 6.5.0-44-generic x86_64

* Alertmanager version:

alertmanager, version 0.27.0 (branch: main, revision: cad5fa580108431e6ed209f2a23a373aa50c098f)
  build date:       20240728-09:39:12
  go version:       go1.21.1
  platform:         linux/amd64
  tags:             netgo

<details>
<summary> Alertmanager configuration file:  </summary>

```yaml
global:
  webex_api_url: https://webexapis.com/v1/messages

route:
  receiver: 'webex'
  group_by: ['alertname', 'cluster']
  group_wait: 15s
  group_interval: 15s
  repeat_interval: 30s

receivers:
- name: 'webex'
  webex_configs:
  - send_resolved: true
    room_id: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
    adaptive_card: |-
      {
        "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
        "version": "1.2",
        "type": "AdaptiveCard",
        "body": [
          {
            "type": "TextBlock",
            "text": "{{ .Status | title }}",
            "size": "large",
            "color": "light",
            "weight": "bolder"
          },
          {{- range $index, $alert := .Alerts -}}
          {{- range $alert.Annotations.SortedPairs }}
            {{- if eq .Name "message" -}}
              {
              "type": "TextBlock",
              "text": "{{- .Value -}}",
              "color": "Accent",
              "size": "Medium",
              "wrap": true
              },
            {{- end }}
          {{- end -}}
          {
            "type": "ColumnSet",
            "columns": [
              {
                "type": "Column",
[...]

PR #3926, opened 2024-07-28
#3925 Enable Sending AdaptiveCards to Webex for Enhanced Alert Notifi…

florianspk:
Issue: #3925 

I've implemented the feature to enable sending AdaptiveCards to Webex from Alertmanager. This enhancement aims to improve the clarity and interactivity of alert notifications.

I have conducted some tests, and you can find the attached results directly on the test Webex channel. If you have any questions or need further clarification, please feel free to reach out. I'm available for any discussions or additional testing as needed.

Thank you for considering this improvement!

@SimonPasquier

image


florianspk:
Add I rename the type of the struc Webex and not Webhook

grobinson-grafana:
We have the problem with other integrations like supporting AdaptiveCards in Microsoft Teams. However, I'm not a fan of giving users a big JSON blob to template as I have had to template JSON and YAML in Jinja2 templates in the past and it was a terrible experience. @gotjosh and @simonpasquier might have different opinions from mine, so I'll wait for them to give their thoughts on this.

grobinson-grafana:
We have the same problem in other integrations. For example, supporting AdaptiveCards in Microsoft Teams. However, I'm not a fan of giving users a big JSON blob to template. I have past experience of templating JSON and YAML in Jinja2 and it was awful. I don't want to push that onto users. @gotjosh and @simonpasquier might have different opinions from mine, so I'll wait for them to give their thoughts on this.

florianspk:
I understand your point. However, at the moment, we have tools that allow us to generate that large JSON blob, and this also helps us customize AdaptiveCards according to the alerts, which is what I’m doing in my case.
Is there a Slack channel, Discord server, or another place where we can discuss this further? Right now, the markdown alerts are not very appealing.

florianspk:
@gotjosh and @simonpasquier, do you have any thoughts on this topic?

grobinson-grafana:
> I understand your point. However, at the moment, we have tools that allow us to generate that large JSON blob, and this also helps us customize AdaptiveCards according to the alerts, which is what I’m doing in my case.

That's great, but what about other Alertmanager users? That is the main problem I'm getting at here. You need tools to make generating JSON templates usable, and most users won't have such tools. Is the expectation on them to build their own?

florianspk:
I understand, and I think we can try to find a compromise between our two ideas. Why not offer a default template where we could simply fill in the title, body, buttons, etc., as you suggest in your PR? At the same time, we could also allow the integration of a JSON blob for those who want to customize their AdaptiveCard. There are tools like [Adaptive Cards Designer](https://adaptivecards.io/designer/) or [Webex Cards Designer](https://developer.webex.com/buttons-and-cards-designer) that make generating these blobs easier.

grobinson-grafana:
I don't think either of those suggestions address the fundamental issue I'm afraid. Here is an example of what I'm getting at here. Here is an example of a template that it outputs invalid JSON if there are more than two alerts in the notification. This is just a simple example, what about debugging a template that's 10x the size? It's these kind of issues that make me hesitant to allow templating of JSON blobs.

```json
{
    "type": "AdaptiveCard",
    "body": [
        {{ range .Alerts.Firing -}}
        {
            "type": "TextBlock",
            "text": "{{ .Labels }}",
            "wrap": true
        },
        {
            "type": "ActionSet",
            "actions": [{
                "type": "Action.OpenUrl",
                "title": "Silence",
                "url": "{{ .SilenceURL }}"
            }]
        }
        {{ end -}}
    ],
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
    "version": "1.6"
}
```

grobinson-grafana:
I don't think either of those suggestions address the fundamental issue I'm afraid. Here is an example of a template that outputs invalid JSON if there are more than two alerts in the notification. This is quite a basic example, but what about debugging a template that is 10x the size with nested loops and if statements? It's these kind of issues that make me hesitant to allow templating of JSON blobs.

```json
{
    "type": "AdaptiveCard",
    "body": [
        {{ range .Alerts.Firing -}}
        {
            "type": "TextBlock",
            "text": "{{ .Labels }}",
            "wrap": true
        },
        {
            "type": "ActionSet",
            "actions": [{
                "type": "Action.OpenUrl",
                "title": "Silence",
                "url": "{{ .SilenceURL }}"
            }]
        }
        {{ end -}}
    ],
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
    "version": "1.6"
}
```

florianspk:
I understand your concern, and you're probably right that we need to find a solution to facilitate the integration of Adaptive Cards. I noticed your PR regarding Adaptive Cards; have you received any feedback on it?

I've also seen open PRs for integrating this feature into Slack, Teams, and Webex. I think it would be beneficial for us to discuss and come up with a common solution, especially since this feature is being requested across multiple platforms. What do you think?

florianspk:
UP ?

Issue #2659, opened 2021-07-30
Memory leak when repeating silences


ngc104:
**What did you do?**

I'm using Kthxbye. When an alert fires and I add a silence with Kthxbye, the memory usage of Alertmanager increases.

You can reproduce this without Kthxbye :

1/ generate an alert (or use any alert sent by Prometheus), for example `PrometheusNotIngestingSamples`.

2/ With Alertmanager, generate silences like this :

```
while true; do
  date # useless; just for tracing...
  amtool --alertmanager.url http://localhost:9093 silence add alertname=PrometheusNotIngestingSamples -a "MemoryLeakLover" -c "Test memory leak in Alertmanager" -d "1m"
  sleep 50
done
```

Note : The behaviour of Kthxbye is similar, but default config is 15 min instead of 1 min. However, with amtool you can see that Kthxbye has nothing to do with this bug.

**What did you expect to see?**

Nothing interesting (no abnormal memory increase)

**What did you see instead? Under which circumstances?**

Follow the metric `container_memory_working_set_bytes` for Alertmanager. After some hours you can see it slightly grow up.

Here is a screenshot of the above test, for a little more than 12 hours : test started at 12h20 and finished at 9h the day after.

![image](https://user-images.githubusercontent.com/67380650/127658457-270837b9-f52f-4caa-9d42-5dd0d620695d.png)

My Alertmanager is running with the default `--data.retention=120h`. I guessed that after 5 days it would stop increasing. Wrong guess : it stops increasing only at OOM and automatic kill.

![image](https://user-images.githubusercontent.com/67380650/127658713-a1301d8f-6aff-477b-9fa1-170e475315da.png)
The above graph was made with Kthxbye running. The pod restarts after an OOM (left side) or after a `kubectl delete pod` (right side).

**Environment**

* System information:

	Kubernetes (deployed with https://github.com/prometheus-community/helm-charts/tree/main/charts/alertmanager)

* Alertmanager version:

```
/alertmanager $ alertmanager --version
alertmanager, version 0.21.0 (branch: HEAD, revision: 4c6c03ebfe21009c546e4d1e9b92c371d67c021d)
  build user:       root@dee35927357f
  build date:       20200617-08:54:02
  go version:       go1.14.4
```

* Alertmanager configuration file:
```
/alertmanager $ cat /etc/alertmanager/alertmanager.yml
global:
  resolve_timeout: 5m
receivers:
- name: rocketchat
  webhook_configs:
[...]

PR #3930, opened 2024-07-31
bugfix: fix leaking of Silences matcherCache entries

Spaceman1701:
There's a small memory leak from the `matcherCache` when a silence is updated in place by this branch of `Silences.Set`:
```
if ok && canUpdate(prev, sil, now) {
	sil.UpdatedAt = now
	msil := s.toMeshSilence(sil)
	if err := s.checkSizeLimits(msil); err != nil {
		return err
	}
	return s.setSilence(msil, now)
}
```

`Silences.Set` will always create a new silence instance. If `canUpdate` is true, the new instance will replace the old one in the silences state. However, the `matcherCache` is keyed by the pointer to the instance so the entry in the matcherCache is ends up dangling. This means that the both compiled matchers _and_ the silence itself are leaked. The `Silences.GC` run doesn't take care of this because it never searches for dangling references in the matcherCache.

We've observed this issue in the real world running a slightly modified version of 0.26.0. In this PR, I've added a new test (`TestSilenceGCOverTime`) which fails when the matcherCache leaks entries. This test still fails when run against the latest code on main:
```
--- FAIL: TestSilenceGCOverTime (0.01s)
    --- FAIL: TestSilenceGCOverTime/silence_update_does_not_leak_state (0.00s)
        silence_test.go:218:
            	Error Trace:	/home/ehunter/nfs-de/oss/forks/alertmanager/silence/silence_test.go:218
            	Error:      	Not equal:
            	            	expected: 1
            	            	actual  : 2
            	Test:       	TestSilenceGCOverTime/silence_update_does_not_leak_state
            	Messages:   	there are extra entries in the matcher cache
FAIL
FAIL	github.com/prometheus/alertmanager/silence	0.040s
FAIL
```

There are a few ways to fix this, but I've chosen to modify `matcherCache` to use the silence UUID as the cache key instead of the pointer to the silence instance. I think this is the best fix because it removes the fragile assumption that the `pb.Silence` pointer will never change over the lifecycle of a silence and replaces it with the existing assumption that the silence's matchers will not change over the lifecycle of a silence. To state this a different way: this fix removes a required invariant and does not add any new ones.

This fix causes the new test cases to pass and has been running in our environment for a while without any problems.

I'm not 100% sure, but I suspect this is the root cause of #2659


grobinson-grafana:
Hello! 👋 Thank you for opening this PR. I haven't had time to do an in-depth review, but my initial impressions are fantastic work! Thank you for tracking this down and also creating a fix.

I agree with the decision to change the key from a pointer to the UUID, and it is my understanding that you cannot change the matchers of a silence without creating a new UUID, therefore it should never be possible to read stale matchers from the cache.

I'll take an in-depth look later this week.

silence/silence_test.go
```diff
@@ -109,6 +109,119 @@ func TestSilencesGC(t *testing.T) {
...
+		return &pb.Silence{
+			Id:       id,
+			StartsAt: now.Add(-time.Second),
+			EndsAt:   exp,
+			Matchers: []*pb.Matcher{
+				{Name: "foo", Type: pb.Matcher_REGEXP, Pattern: "bar"},
+			}}
+	}
+
+	cases := map[string]testCase{
```
grobinson-grafana, line 136: It's more idiomatic to write test cases like this 

```diff
diff --git a/silence/silence_test.go b/silence/silence_test.go
index c290522a..7a1bf335 100644
--- a/silence/silence_test.go
+++ b/silence/silence_test.go
@@ -114,11 +114,6 @@ func TestSilenceGCOverTime(t *testing.T) {
                s                    *pb.Silence
                expectPresentAfterGc bool
        }
-       type testCase struct {
-               initialState    []silenceEntry
-               updates         []silenceEntry
-               expectedGCCount int
-       }

        c := clock.NewMock()
        now := c.Now().UTC()
@@ -133,35 +128,39 @@ func TestSilenceGCOverTime(t *testing.T) {
                        }}
        }

-       cases := map[string]testCase{
-               "gc does not clean active silences": {
-                       initialState: []silenceEntry{
-                               {s: newSilence("1", now), expectPresentAfterGc: false},
-                               {s: newSilence("2", now.Add(-time.Second)), expectPresentAfterGc: false},
-                               {s: newSilence("3", now.Add(time.Second)), expectPresentAfterGc: true},
-                       },
+       cases := []struct {
+               name            string
+               initialState    []silenceEntry
+               updates         []silenceEntry
+               expectedGCCount int
+       }{{
+               name: "gc does not clean active silences",
+               initialState: []silenceEntry{
+                       {s: newSilence("1", now), expectPresentAfterGc: false},
+                       {s: newSilence("2", now.Add(-time.Second)), expectPresentAfterGc: false},
+                       {s: newSilence("3", now.Add(time.Second)), expectPresentAfterGc: true},
                },
-               "silences added with Set are handled correctly": {
-                       initialState: []silenceEntry{
-                               {s: newSilence("1", now), expectPresentAfterGc: false},
-                       },
-                       updates: []silenceEntry{
-                               {s: newSilence("", now.Add(time.Second)), expectPresentAfterGc: true},
-                               {s: newSilence("", now.Add(-time.Second)), expectPresentAfterGc: false},
-                       },
+       }, {
+               name: "silences added with Set are handled correctly",
+               initialState: []silenceEntry{
+                       {s: newSilence("1", now), expectPresentAfterGc: false},
                },
-               "silence update does not leak state": {
-                       initialState: []silenceEntry{
-                               {s: newSilence("1", now), expectPresentAfterGc: false},
-                       },
-                       updates: []silenceEntry{
-                               {s: newSilence("1", now.Add(time.Second)), expectPresentAfterGc: true},
-                       },
+               updates: []silenceEntry{
+                       {s: newSilence("", now.Add(time.Second)), expectPresentAfterGc: true},
+                       {s: newSilence("", now.Add(-time.Second)), expectPresentAfterGc: false},
                },
-       }
+       }, {
+               name: "silence update does not leak state",
+               initialState: []silenceEntry{
+                       {s: newSilence("1", now), expectPresentAfterGc: false},
+               },
+               updates: []silenceEntry{
+                       {s: newSilence("1", now.Add(time.Second)), expectPresentAfterGc: true},
+               },
+       }}

-       for name, tc := range cases {
-               t.Run(name, func(t *testing.T) {
+       for _, tc := range cases {
+               t.Run(tc.name, func(t *testing.T) {
                        silences, err := New(Options{})
                        silClock := clock.NewMock()
                        silences.clock = silClock
```

silence/silence_test.go
```diff
@@ -109,6 +109,119 @@ func TestSilencesGC(t *testing.T) {
...
+			},
+			updates: []silenceEntry{
+				{s: newSilence("1", now.Add(time.Second)), expectPresentAfterGc: true},
+			},
+		},
+	}
+
+	for name, tc := range cases {
+		t.Run(name, func(t *testing.T) {
+			silences, err := New(Options{})
```
grobinson-grafana, line 165: Please just move the `require` after the `New` call:

```suggestion
			silences, err := New(Options{})
			require.NoError(t, err)
```

silence/silence_test.go
```diff
@@ -109,6 +109,119 @@ func TestSilencesGC(t *testing.T) {
...
+				}
+				if sil.expectPresentAfterGc {
+					expectedRemaining = append(expectedRemaining, sil.s.Id)
+				} else {
+					expectedGCCount += 1
+				}
+				// simulate this silences being seen in a query
+				silences.mc.Get(silences.st[sil.s.Id].Silence)
+			}
+			silClock.Add(-time.Second)
```
grobinson-grafana, line 185: I'm not sure I understand the significance of rewinding the clock, although I can see that if I comment this out the test fails. How does rewinding the clock help us test GC behavior?

silence/silence_test.go
```diff
@@ -109,6 +109,119 @@ func TestSilencesGC(t *testing.T) {
 	require.Equal(t, want, s.st)
 }
 
+func TestSilenceGCOverTime(t *testing.T) {
```
grobinson-grafana, line 112: It seems to me that we should fold this test into `TestSilenceGC`?

Spaceman1701: That seems reasonable to me as well - I left them both in just to avoid adding additional mental overhead when looking at the new test. The first test case should be exactly the same as `TestSilenceGC`

silence/silence_test.go
Spaceman1701: Thanks for point this out! I did this because adding a silence with `Set` that expires before the clock's `now` is a no-op. Typically, this case is handled before we reach `Set`. However, rewinding the clock is a really unclear way to handle this.

Instead, I should've started with the clock 2 seconds behind and then incremented the clock forward by one second after the `initialState` is applied and then again after `updates` are applied. That has the exact same behavior, but is much more expressive to the reader.

Spaceman1701 (commented):
Thanks for the review! I will make all the requested stylistic changes and the change to how the test clock is handled.

silence/silence_test.go
```diff
@@ -84,29 +84,130 @@ func TestOptionsValidate(t *testing.T) {
...
+		{
+			name: "gc does not clean active silences",
+			initialState: []silenceEntry{
+				{s: newSilence("1", now), expectPresentAfterGc: false},
+				{s: newSilence("2", now.Add(-time.Second)), expectPresentAfterGc: false},
+				{s: newSilence("3", now.Add(time.Second)), expectPresentAfterGc: true},
+			},
+		},
+		{
+			name: "silences added with Set are handled correctly",
```
grobinson-grafana, line 121: > handled correctly

What does this mean?

grobinson-grafana (commented):
I think the code is fine, but the tests are still very hard to follow and understand. I'm going to make an attempt at refactoring them further in a local branch.

grobinson-grafana:
This is how I would make the tests simpler. Let me know if I'm missing a test case.

```go
func TestSilenceGCOverTime(t *testing.T) {
	t.Run("GC does not remove active silences", func(t *testing.T) {
		s, err := New(Options{})
		require.NoError(t, err)
		s.clock = clock.NewMock()
		now := s.nowUTC()
		s.st = state{
			"1": &pb.MeshSilence{Silence: &pb.Silence{Id: "1"}, ExpiresAt: now},
			"2": &pb.MeshSilence{Silence: &pb.Silence{Id: "2"}, ExpiresAt: now.Add(-time.Second)},
			"3": &pb.MeshSilence{Silence: &pb.Silence{Id: "3"}, ExpiresAt: now.Add(time.Second)},
		}
		want := state{
			"3": &pb.MeshSilence{Silence: &pb.Silence{Id: "3"}, ExpiresAt: now.Add(time.Second)},
		}
		n, err := s.GC()
		require.NoError(t, err)
		require.Equal(t, 2, n)
		require.Equal(t, want, s.st)
	})

	// This test checks for a memory leak that occurred in the matcher cache when
	// updating an existing silence.
	t.Run("Updating an existing silences does not leak cache entries", func(t *testing.T) {
		s, err := New(Options{})
		require.NoError(t, err)
		clock := clock.NewMock()
		s.clock = clock
		sil1 := &pb.Silence{
			Id: "1",
			Matchers: []*pb.Matcher{{
				Type:    pb.Matcher_EQUAL,
				Name:    "foo",
				Pattern: "bar",
			}},
			StartsAt: clock.Now(),
			EndsAt:   clock.Now().Add(time.Minute),
		}
		s.st["1"] = &pb.MeshSilence{Silence: sil1, ExpiresAt: clock.Now().Add(time.Minute)}
		// Need to query the silence to populate the matcher cache.
		s.Query(QMatches(model.LabelSet{"foo": "bar"}))
		require.Len(t, s.mc, 1)
		// must clone sil1 before updating it.
		sil2 := cloneSilence(sil1)
		require.NoError(t, s.Set(sil2))
		// The memory leak occurred because updating a silence would add a new
		// entry in the matcher cache even though no new silence was created.
		// This check asserts that this no longer happens.
		require.Len(t, s.st, 1)
		require.Len(t, s.mc, 1)
		// Move time forward and both silence and cache entry should be garbage
		// collected.
		clock.Add(time.Minute)
		n, err := s.GC()
		require.NoError(t, err)
		require.Equal(t, 1, n)
		require.Len(t, s.st, 0)
		require.Len(t, s.mc, 0)
	})
}
```

silence/silence_test.go
```diff
@@ -84,29 +84,130 @@ func TestOptionsValidate(t *testing.T) {
...
 	}
-	want := state{
-		"3": newSilence(now.Add(time.Second)),
+
+	// The GC will run with it's clock equal to 1 second after now
+	cases := []struct {
+		name            string
+		initialState    []silenceEntry
+		updates         []silenceEntry
+		expectedGCCount int
```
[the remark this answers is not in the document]
Spaceman1701: it is used on line 194

silence/silence_test.go
Spaceman1701: Ah this case name isn't great - in this case "handled correctly" just means "all the invariants we're testing remain satisfied"

Spaceman1701:
> This is how I would make the tests simpler. Let me know if I'm missing a test case.

This looks good to me, but it does make it a bit harder to add new cases in the future. There's no test here which actually validates that the GC runs as expected with silences added via the normal `Set` method, but this might not be important enough to worry about.

The `GC does not remove active silences` case is also missing a validation that the matcher cache length is correct after all the operations.

silence/silence_test.go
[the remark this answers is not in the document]
grobinson-grafana: No that's a different variable with the same name.

silence/silence_test.go
grobinson-grafana: > "all the invariants we're testing remain satisfied"

Yeah but what are the invariants being tested? I'm sort of coming at this from the perspective of someone who hasn't reviewed this PR but needs to understand the tests. It's **so hard** to understand **what** is being tested here from looking at the test case:

```go
{
	name: "silences added with Set are handled correctly",
	initialState: []silenceEntry{
		{s: newSilence("1", now), expectPresentAfterGc: false},
	},
	updates: []silenceEntry{
		{s: newSilence("", now.Add(time.Second)), expectPresentAfterGc: true},
		{s: newSilence("", now.Add(-time.Second)), expectPresentAfterGc: false},
	},
},
```

grobinson-grafana:
> This looks good to me, but it does make it a bit harder to add new cases in the future.

I think for making these specific tests table-driven make them more difficult to understand. There are lots of cases where table-driven tests do make a lot of sense, but I don't think this is one of those. For example, my comment [here](https://github.com/prometheus/alertmanager/pull/3930#discussion_r1722159246) highlights what I mean.

> There's no test here which actually validates that the GC runs as expected with silences added via the normal `Set` method, but this might not be important enough to worry about.

We can fix that 👍 

> The `GC does not remove active silences` case is also missing a validation that the matcher cache length is correct after all the operations.

👍 

> Would you like me to replace my test in this PR with this new implementation?

Yes please! Let me first add a new comment that has the updated tests including your feedback 👍

grobinson-grafana:
> This looks good to me, but it does make it a bit harder to add new cases in the future.

I think writing these specific tests as table-driven tests make them more difficult to understand. There are lots of cases where table-driven tests do make a lot of sense, but I don't think this is one of those. For example, my comment [here](https://github.com/prometheus/alertmanager/pull/3930#discussion_r1722159246) highlights what I mean.

> There's no test here which actually validates that the GC runs as expected with silences added via the normal `Set` method, but this might not be important enough to worry about.

We can fix that 👍 

> The `GC does not remove active silences` case is also missing a validation that the matcher cache length is correct after all the operations.

👍 

> Would you like me to replace my test in this PR with this new implementation?

Yes please! Let me first add a new comment that has the updated tests including your feedback 👍

Spaceman1701:
> I think writing these specific tests as table-driven tests make them more difficult to understand. There are lots of cases where table-driven tests do make a lot of sense, but I don't think this is one of those. For example, my comment https://github.com/prometheus/alertmanager/pull/3930#discussion_r1722159246 highlights what I mean.

Sure, I think that's fair enough. The way you've rewritten does seem more readable to me as well. Regardless, I'm very happy to conform to whatever is conventional for Alertmanager.

> Yes please! Let me first add a new comment that has the updated tests including your feedback 👍
Alright, great. Thanks!

grobinson-grafana:
```go
func TestSilenceGCOverTime(t *testing.T) {
	t.Run("GC does not remove active silences", func(t *testing.T) {
		s, err := New(Options{})
		require.NoError(t, err)
		s.clock = clock.NewMock()
		now := s.nowUTC()
		s.st = state{
			"1": &pb.MeshSilence{Silence: &pb.Silence{Id: "1"}, ExpiresAt: now},
			"2": &pb.MeshSilence{Silence: &pb.Silence{Id: "2"}, ExpiresAt: now.Add(-time.Second)},
			"3": &pb.MeshSilence{Silence: &pb.Silence{Id: "3"}, ExpiresAt: now.Add(time.Second)},
		}
		want := state{
			"3": &pb.MeshSilence{Silence: &pb.Silence{Id: "3"}, ExpiresAt: now.Add(time.Second)},
		}
		n, err := s.GC()
		require.NoError(t, err)
		require.Equal(t, 2, n)
		require.Equal(t, want, s.st)
	})

	t.Run("GC does not leak cache entries", func(t *testing.T) {
		s, err := New(Options{})
		require.NoError(t, err)
		clock := clock.NewMock()
		s.clock = clock
		sil1 := &pb.Silence{
			Matchers: []*pb.Matcher{{
				Type:    pb.Matcher_EQUAL,
				Name:    "foo",
				Pattern: "bar",
			}},
			StartsAt: clock.Now(),
			EndsAt:   clock.Now().Add(time.Minute),
		}
		require.NoError(t, s.Set(sil1))
		// Need to query the silence to populate the matcher cache.
		s.Query(QMatches(model.LabelSet{"foo": "bar"}))
		require.Len(t, s.st, 1)
		require.Len(t, s.mc, 1)
		// Move time forward and both silence and cache entry should be garbage
		// collected.
		clock.Add(time.Minute)
		n, err := s.GC()
		require.NoError(t, err)
		require.Equal(t, 1, n)
		require.Len(t, s.st, 0)
		require.Len(t, s.mc, 0)
	})

	t.Run("replacing a silences does not leak cache entries", func(t *testing.T) {
		s, err := New(Options{})
		require.NoError(t, err)
		clock := clock.NewMock()
		s.clock = clock
		sil1 := &pb.Silence{
			Matchers: []*pb.Matcher{{
				Type:    pb.Matcher_EQUAL,
				Name:    "foo",
				Pattern: "bar",
			}},
			StartsAt: clock.Now(),
			EndsAt:   clock.Now().Add(time.Minute),
		}
		require.NoError(t, s.Set(sil1))
		// Need to query the silence to populate the matcher cache.
		s.Query(QMatches(model.LabelSet{"foo": "bar"}))
		require.Len(t, s.st, 1)
		require.Len(t, s.mc, 1)
		// must clone sil1 before replacing it.
		sil2 := cloneSilence(sil1)
		sil2.Matchers = []*pb.Matcher{{
			Type:    pb.Matcher_EQUAL,
			Name:    "bar",
			Pattern: "baz",
		}}
		require.NoError(t, s.Set(sil2))
		// Need to query the silence to populate the matcher cache.
		s.Query(QMatches(model.LabelSet{"bar": "baz"}))
		require.Len(t, s.st, 2)
		require.Len(t, s.mc, 2)
		// Move time forward and both silence and cache entry should be garbage
		// collected.
		clock.Add(time.Minute)
		n, err := s.GC()
		require.NoError(t, err)
		require.Equal(t, 2, n)
		require.Len(t, s.st, 0)
		require.Len(t, s.mc, 0)
	})

	// This test checks for a memory leak that occurred in the matcher cache when
	// updating an existing silence.
	t.Run("updating a silences does not leak cache entries", func(t *testing.T) {
		s, err := New(Options{})
		require.NoError(t, err)
		clock := clock.NewMock()
		s.clock = clock
		sil1 := &pb.Silence{
			Id: "1",
			Matchers: []*pb.Matcher{{
				Type:    pb.Matcher_EQUAL,
				Name:    "foo",
				Pattern: "bar",
			}},
			StartsAt: clock.Now(),
			EndsAt:   clock.Now().Add(time.Minute),
		}
		s.st["1"] = &pb.MeshSilence{Silence: sil1, ExpiresAt: clock.Now().Add(time.Minute)}
		// Need to query the silence to populate the matcher cache.
		s.Query(QMatches(model.LabelSet{"foo": "bar"}))
		require.Len(t, s.mc, 1)
		// must clone sil1 before updating it.
		sil2 := cloneSilence(sil1)
		require.NoError(t, s.Set(sil2))
		// The memory leak occurred because updating a silence would add a new
		// entry in the matcher cache even though no new silence was created.
		// This check asserts that this no longer happens.
		require.Len(t, s.st, 1)
		require.Len(t, s.mc, 1)
		// Move time forward and both silence and cache entry should be garbage
		// collected.
		clock.Add(time.Minute)
		n, err := s.GC()
		require.NoError(t, err)
		require.Equal(t, 1, n)
		require.Len(t, s.st, 0)
		require.Len(t, s.mc, 0)
	})
}
```

silence/silence_test.go
```diff
@@ -84,29 +84,135 @@ func TestOptionsValidate(t *testing.T) {
...
+		// Need to query the silence to populate the matcher cache.
+		s.Query(QMatches(model.LabelSet{"foo": "bar"}))
+		require.Len(t, s.mc, 1)
+		// must clone sil1 before updating it.
+		sil2 := cloneSilence(sil1)
+		require.NoError(t, s.Set(sil2))
+		// The memory leak occurred because updating a silence would add a new
+		// entry in the matcher cache even though no new silence was created.
+		// This check asserts that this no longer happens.
+		s.Query(QMatches(model.LabelSet{"foo": "bar"}))
```
Spaceman1701, line 204: I added this line because the memory leak only occurs if the matcher cache is populated because of a query.

grobinson-grafana (approved):
@gotjosh @simonpasquier I approve this fix. Please take a look so we can get it into the next release! Thanks 💯

grobinson-grafana:
Sorry, there are a couple of lint failures where we need to use `require.Empty`. Can you make those fixes so lint passes?

grobinson-grafana:
Can you also rebase `main` branch? It looks like the frontend tests are failing, which are unrelated to your changes.

Spaceman1701:
> Can you also rebase main branch? It looks like the frontend tests are failing, which are unrelated to your changes.

It looks like that test is still broken - is it possible that it's broken on main? As far as I can tell, the diff between this branch and main doesn't include any frontend files.

grobinson-grafana:
> > Can you also rebase main branch? It looks like the frontend tests are failing, which are unrelated to your changes.
> 
> It looks like that test is still broken - is it possible that it's broken on main? As far as I can tell, the diff between this branch and main doesn't include any frontend files.

Looks like all PRs are broken due to frontend tests. We will fix it 👍 You'll need to rebase a second time once it's fixed.

gotjosh:
Thank you very much for your contribution and @grobinson-grafana for reviewing.

PR #3944, opened 2024-08-05
fix: pageduty custom_details in payload

### LINKED ISSUE #2477: alertmanager-config.yaml file for pagerduty_configs.details does not respect EventApi v2 spec
**What did you do?**
Generate a nested object for pagerduty_configs.details field inside alertmanager-config.yaml

amtool should not fail and alertmanager must allow nested objects as it's compatible with pagerduty event api v2

Error when running amtool and alertmanager 

**Environment**

* System information:

	Linux 5.3.0-46-generic x86_64

* Alertmanager version:

	alertmanager, version 0.21.0 (branch: HEAD, revision: 4c6c03ebfe21009c546e4d1e9b92c371d67c021d)
          build user:       root@dee35927357f
          build date:       20200617-08:54:02
          go version:       go1.14.4

* Prometheus version:

	insert output of `prometheus --version` here (repeat for each prometheus
	version in your cluster, if relevant to the issue)

```
global:
  resolve_timeout: "10m"

templates:
  - '/etc/alertmanager/secrets/pagerduty.integration.key.tmpl'

route:
  group_by: [ '...' ]
  group_wait: 30s
  group_interval: 5m
  repeat_interval: 12h
  receiver: 'black-hole'

  routes:
    - match_re:
        severity: "warning|critical"
      receiver: "all"

inhibit_rules:
  - source_match:
      severity: 'critical'
    target_match:
      severity: 'warning'
    equal: [ 'alertname' ]

receivers:
  - name: "all"
    pagerduty_configs:
      - routing_key: '{{ template "pagerduty.integration.key" . }}'
        details:
          firing:       '{{ template "pagerduty.default.instances" .Alerts.Firing }}'
          resolved:     '{{ template "pagerduty.default.instances" .Alerts.Resolved }}'
          num_firing:   '{{ .Alerts.Firing | len }}'
          num_resolved: '{{ .Alerts.Resolved | len }}'
          toto: [ "hello", "hoho" ] #this line will fail
```

* Logs:
```
level=error ts=2021-02-05T10:57:07.101Z caller=coordinator.go:124 component=configuration msg="Loading configuration file failed" file=/etc/alertmanager/config/
alertmanager.yaml err="yaml: unmarshal errors:\n  line 91: cannot unmarshal !!seq into string"
```

This is of course expected as alertmanager is not able to handle the `toto` list we added, but if I craft the same with a curl and contact directly `events.pagerduty.com/v2/enqueue` it get what I wanted.

```
curl -X POST -d '{"routing_key": "XXXX", "event_action": "trigger", "payload": {"summary": "one alert", "source": "curl", "severity": "critical", "custom_details": { "toto": ["hello", "hoho"]}}}' https://events.pagerduty.com/v2/enqueue
```

### LINKED ISSUE #3218: PagerDuty and structured alerts data
I feel there is a need for a new PagerDuty integration in the Alertmanager which would keep alerts data in a structured form when transmitted to PagerDuty as opposed to the currently templated form.

I believe we need this because PagerDuty offers Automation features which allow to take certain actions based on field values. The fact that Alertmanager marshalls annotations and labels values in big chunks of text in the `firing` or `resolved` custom details fields limit greatly the possibility of using these data in automation rules.

PagerDuty V2 Event API support has a `custom_details` field which is currently used like this:

```
"custom_details": {
  "firing":       `{{ template "pagerduty.default.instances" .Alerts.Firing }}`,
  "resolved":     `{{ template "pagerduty.default.instances" .Alerts.Resolved }}`,
  "num_firing":   `{{ .Alerts.Firing | len }}`,
  "num_resolved": `{{ .Alerts.Resolved | len }}`
}
```

Unless I'm mistaken, the keys of the `custom_details` object must be hardcoded in the receiver configuration. There is no possibility to define them dynamically using `.Alerts.(Resolved|Firing).Labels` or `.Alerts.(Resolved|Firing).Annotations`.

After some testing it seems that PagerDuty V2 Event API supports at least 1 level of objects in `custom_details` allowing to do something like that.

PSEUDO CODE that describes valid JSON map[string]map[string]string for `custom_details`:
```go
"custom_details": {
  "annotations": {
{{ range .Alerts.(Resolved|Firing)[0].Annotations.SortedPairs }}
    "{{ .Name }}": "{{ .Value }}",
{{ end }}
  },
  "labels": {
{{ range .Alerts.(Resolved|Firing)[0].Labels.SortedPairs }}
    "{{ .Name }}": "{{ .Value }}",
{{ end }}
  },  
}
```

In that form, I think it will offer Alertmanager users the possibility to use advanced PagerDuty features.

Regards.

siavashs:
This change fixes the `custom_details` value in PagerDuty V2 API payloads.

Fixes #2477
Fixes #3218

Alertmanager PagerDuty configuration allows users to define `details` under `pagerduty_configs`:
```yaml
[ details: { <string>: <tmpl_string>, ... } | default = {
  firing:       '{{ template "pagerduty.default.instances" .Alerts.Firing }}'
  resolved:     '{{ template "pagerduty.default.instances" .Alerts.Resolved }}'
  num_firing:   '{{ .Alerts.Firing | len }}'
  num_resolved: '{{ .Alerts.Resolved | len }}'
} ]
```

The internal Alertmanager configuration structure is defined as:
```go
// PagerdutyConfig configures notifications via PagerDuty.
type PagerdutyConfig struct {
...
	Details        map[string]string `yaml:"details,omitempty" json:"details,omitempty"`
}
```
And the PagerDuty payload is defined as:
```go
type pagerDutyPayload struct {
...
	CustomDetails map[string]string `json:"custom_details,omitempty"`
```

The current flow of configuration parsing and encoding is:
1. `pagerduty_configs[].details` are read from configuration as `map[string]string`
2. The `details` map values are parsed as text templates
3. The result is passed as `custom_details`, part of the JSON payload sent to PagerDuty API V2

Here is an example payload using `.Alerts.Firing`, which returns a list of currently firing alert objects in this group. Documented here: https://prometheus.io/docs/alerting/latest/notifications/#data This results in a payload like below:
```json
{
  "client": "Alertmanager",
...
  "custom_details": {
    "firing": "Labels:
 - alertname = Server_Down
...
Annotations:
 - summary = Server is down
...
"
    }
}
```
Using `map[string]string` in payloads cause the rendered YAML templates to be encoded as JSON strings. This is undesirable in most cases as the end-user might expect a nested JSON object which is machine readable/parsable by it's fields.

This change:
- adds logic to unmarshall `details` values as YAML with fallback to string
- uses `map[string]interface{}` to encode `custom_details` to JSON
- requires gopkg.in/yaml.v3, see https://github.com/go-yaml/yaml/issues/591
- does not change existing yaml.v2 dependency, see https://github.com/prometheus/alertmanager/pull/3322


siavashs:
> I don't think it is intentional that the default template is valid YAML, it's just meant to be plain text. I'd even argue it's meant to be Markdown. If I understand the use case, you would like to be able to add structured objects (with nesting) to the `custom_details` field?

Correct, our teams prefer that so they can add logic on PagerDuty side depending on the field values.
I tried to keep the logic generic so it would resolve the 2 referenced issues as well.

grobinson-grafana:
Have you considered changing `Details` in the configuration file from `map[string]string` to `map[string]interface{}`?

```go
// PagerdutyConfig configures notifications via PagerDuty.
type PagerdutyConfig struct {
...
	Details        map[string]interface{} `yaml:"details,omitempty" json:"details,omitempty"`
}
```

Would this avoid having to embed YAML inside existing YAML? I haven't checked if this will affect existing configurations, but  it would be good to understand if this is a better option.

siavashs:
> Have you considered changing `Details` in the configuration file from `map[string]string` to `map[string]interface{}`?

That was the initial approach I though about. It will create a free style format and would require checking the interfaces one by one as they can be either maps or templated strings. I'm open to experiment with it.

Daniel-Vaz:
Curious to know... Is this going forward ?
Seems a real blessing having PD Custom Details properly rendered from the Alert Payload.

siavashs:
I'll update this PR with the suggested `map[string]interface{}` to support nested fields.

siavashs:
Thinking about this more, using a template like `alerts: {{ .Alerts }}` will translate into `alert` key with value of string as templates are only rendered into strings.

Parsing template results into maps or slices is possible but could be error-prone and we'd need a format for it.
Alternative option could be toggles to include specific details in the payload:
```yaml
details: ...

details_include:
  alerts: true
  ...
```

siavashs:
I updated the PR to use the new `toJson` template function.
Also we can now support this in both Events API v1 and v2.
image

NargiT:
@siavashs you are a true hero

abhijith-db:
@siavashs For v1 integration, this seems to have changed the urls in label/annotations to not render as links in PD. PD used to make them clickable hyperlinks when `firing` value was string, but now it's all jsonified and render as plain text.

siavashs:
@abhijith-db this sounds like a UI issue on PD side.
As a workaround you can always use a "non-json" template on Alertmanager side and it should restore the old format.

abhijith-db:
@siavashs Yeah that's a workaround. Only problem is there's no global PD default template to configure.  Each receiver has to use the non-json template to linkify urls in annotations

siavashs:
I see, we'll add a global config for PagerDuty.

abhijith-db:
@siavashs Also the deep copy with template was affecting this too. Had to fork and fix on our end to exclude string type. As:
```
		if strVal, ok := v.(string); ok {
			// For string values, evaluate the template directly without
			// YAML-parsing the result. This preserves plain text output
			// (e.g. from pagerduty.default.instances) as strings.
			rendered[k], err = tmplTextFunc(strVal)
		} else {
			rendered[k], err = template.DeepCopyWithTemplate(v, tmplTextFunc)
		}
```

Failing testcase that would have passed before this change:
```
		{
			name: "default text template",
			args: args{
				details: map[string]any{
					"firing":       `{{ template "pagerduty.default.instances" .Alerts.Firing }}`,
					"resolved":     `{{ template "pagerduty.default.instances" .Alerts.Resolved }}`,
					"num_firing":   `{{ .Alerts.Firing | len }}`,
					"num_resolved": `{{ .Alerts.Resolved | len }}`,
				},
				data: &template.Data{
					Alerts: template.Alerts{
						{
							Status: "firing",
							Labels: template.KV{
								"alertname": "TestAlert",
							},
							Annotations: template.KV{
								"description": "Something is wrong",
							},
							GeneratorURL: "http://localhost:9090/graph",
						},
					},
				},
			},
			want: map[string]any{
				"firing":       "Labels:\n - alertname = TestAlert\nAnnotations:\n - description = Something is wrong\nSource: http://localhost:9090/graph\n",
				"resolved":     "",
				"num_firing":   "1",
				"num_resolved": "0",
			},
			wantErr: false,
		},
```
It's a regression only because PD would render addresses in annotations etc as links which broke after this.

Rohlik:
This PR should be marked as potentially breaking changes in the changelog 😞.  
In our case, it completely broke up our incident routing on the PagerDuty side via the Event Orchestration feature, where we have routes configured like:  
```
event.custom_details matches part 'owner = security' and event.custom_details matches part 'cluster = worker01'
```
  
So, we were heavily reliant on the legacy non-JSON format, and we found that the hard way, unfortunately 😢.

PR #3945, opened 2024-08-06
ui: Decode "+" in the query parameters as "%20"

jmleddy:
I just copied the solution from here since it's been a while since
using functional languages and I didn't know the order of the parser,
but I suspect there are much better places to put this, including
repacing in `rest` above

https://github.com/elm/url/issues/32#issuecomment-803595052


jmleddy:
https://github.com/elm/url/issues/32

TheMeier:
@jmleddy can you rebase this?

SoloJacobs:
Hi @jmleddy ,

I'm afraid this isn't the correct fix. We can't fix this in the Elm UI:

The `+` in your URL is a space that got encoded by something following `application/x-www-form-urlencoded` rules, where a space becomes `+`. The UI decodes URLs the ordinary way (RFC 3986), where a space is `%20` and `+` is just the plus character. Hence the mismatch.

The problem with translating `+` to a space here is that the UI can't tell your case apart from a label value that genuinely contains a plus. `alertname="Disk+usage"` links work today. With this change they would silently start filtering for `Disk usage` and match nothing.

Only whatever builds the URL knows whether a `+` was a space or a plus, so that's where the encoding needs fixing. In this case, `urlquery` is using `application/x-www-form-urlencoded` rules, so that is where the problem is. This gives you a correct URL:
```
{{ .Value | urlquery | reReplaceAll "[+]" "%20" }}
```

PR #3948, opened 2024-08-07
chore: update golangci-lint-action version

dongjiang1989:
update golangci-lint-action version to `v6.1.0`


simonpasquier (approved):
Thanks!

PR #3949, opened 2024-08-08
opsgenie: Add compatibility with JSM Operations

jkroepke:
At the moment, the OpsGenie API and JSM Operations (Jira Service Management Operations) API are equal

* OpsGenie https://docs.opsgenie.com/docs/alert-api#create-alert
* JSM https://developer.atlassian.com/cloud/jira/service-desk-ops/rest/v2/api-group-alerts/#api-v1-alerts-get

However, the OpsGenie v2 Alert API is re-introduced as v1 API at Atlassian side.

Background: https://support.atlassian.com/jira-service-management-cloud/docs/start-shifting-from-opsgenie-to-jira-service-management/

### DIFF
--- a/notify/opsgenie/opsgenie.go
+++ b/notify/opsgenie/opsgenie.go
@@ -38,11 +38,12 @@ const maxMessageLenRunes = 130
 
 // Notifier implements a Notifier for OpsGenie notifications.
 type Notifier struct {
-	conf    *config.OpsGenieConfig
-	tmpl    *template.Template
-	logger  log.Logger
-	client  *http.Client
-	retrier *notify.Retrier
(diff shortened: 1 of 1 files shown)


z0rc:
I don't think domain assumption is a valid solution here, at least as it's implemented now. 

We're using Jira SM "Prometheus" integration as documented in https://support.atlassian.com/jira-service-management-cloud/docs/integrate-with-prometheus/, It gives us API URL `https://api.atlassian.com/jsm/ops/integration/`, which impements `v2/alerts` endpoints that work just fine with current opsgenie notifier.

jkroepke:
Ah, good to know that undocumented stuff works. Then I will close the issue.

PR #3958, opened 2024-08-15
opsgenie: Add visible_to

### LINKED ISSUE #3953: [Feature Request] Add `visibleTo` field to OpsGenie integration
**What did you do?**
I checked the [documentation](https://prometheus.io/docs/alerting/latest/configuration/#opsgenie_config) and [code](https://github.com/prometheus/alertmanager/blob/main/config/notifiers.go#L102-L111) and didn't find any option to map the field `visibleTo` for the OpsGenie integration. The field `visibleTo` would be very useful to give a team access to see the alert, but not trigger the normal responder workflow.

**What do you want to solve?**
*Usecase*: Several teams of developers have ownership of several applications each. All these applications have global alerts that the developers want to receive, so they are the responders. The infrastructure team want insight, but doesn't need to respond to the developers alerts and this is where `visibleTo` comes in.

**Suggestion**
I suggest a new field called `visible_to` (following the naming scheme that's already in place) that's mapped to OpsGenie's Alert API with the field name `visibleTo`. Here's the relevant [OpsGenie documentation](https://docs.opsgenie.com/docs/alert-api#create-alert). I've added the relevant row from the table in the documentation below for convenience.

| Field | Mandatory | Description | Limit |
|--|--|--|--|
| visibleTo | false | Teams and users that the alert will become visible to without sending any notification. `type` field is mandatory for each item, where possible values are **team** and **user**. In addition to the `type` field, either **id** or **name** should be given for teams and either id or username should be given for users. Please note: that alert will be visible to the teams that are specified within `responders` field by default, so there is no need to re-specify them within `visibleTo` field. You can refer below for example values. | 50 teams or users in total |

chlunde:
Add visible_to with the same support for templating and fields as the
existing responders field.

Opsgenie documentation:

>   Teams and users that the alert will become visible to without sending
>   any notification. type field is mandatory for each item, where possible
>   values are team and user. In addition to the type field, either id or
>   name should be given for teams and either id or username should be given
>   for users. Please note: that alert will be visible to the teams that are
>   specified within responders field by default, so there is no need to
>   re-specify them within visibleTo field. You can refer below for example
>   values.

The functionality has been tested with unit tests, and with the following
configuration:

```yaml
  route:
    receiver: og
    group_wait: 30s
    group_interval: 5m
    repeat_interval: 12h
  receivers:
  - name: og
    opsgenie_configs:
    - send_resolved: true
      api_key: ...top secret...
      api_url: https://api.eu.opsgenie.com/
      message: 'Lab message'
      description: 'Static description'
      source: lab
      details:
        foo: bar
      priority: P4
      visible_to:
      - type: user
        username: foo@example.com
```

Fixes #3953

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -584,6 +584,7 @@ type OpsGenieConfig struct {
 	Details      map[string]string         `yaml:"details,omitempty" json:"details,omitempty"`
 	Entity       string                    `yaml:"entity,omitempty" json:"entity,omitempty"`
 	Responders   []OpsGenieConfigResponder `yaml:"responders,omitempty" json:"responders,omitempty"`
+	VisibleTo    []OpsGenieConfigVisibleTo `yaml:"visible_to,omitempty" json:"visible_to,omitempty"`
 	Actions      string                    `yaml:"actions,omitempty" json:"actions,omitempty"`
 	Tags         string                    `yaml:"tags,omitempty" json:"tags,omitempty"`
 	Note         string                    `yaml:"note,omitempty" json:"note,omitempty"`
@@ -625,6 +626,24 @@ func (c *OpsGenieConfig) UnmarshalYAML(unmarshal func(interface{}) error) error 
 		}
 	}
 
+	for _, v := range c.VisibleTo {
+		if v.ID == "" && v.Username == "" && v.Name == "" {
+			return fmt.Errorf("opsGenieConfig visible_to %v has to have at least one of id, username or name specified", v)
+		}
+
+		if strings.Contains(v.Type, "{{") {
+			_, err := template.New("").Parse(v.Type)
+			if err != nil {
+				return fmt.Errorf("opsGenieConfig visible_to %v type is not a valid template: %w", v, err)
+			}
+		} else {
+			v.Type = strings.ToLower(v.Type)
+			if !opsgenieTypeMatcher.MatchString(v.Type) {
+				return fmt.Errorf("opsGenieConfig visible_to %v type does not match valid options %s", v, opsgenieValidTypesRe)
+			}
+		}
+	}
+
 	return nil
 }
 
@@ -635,6 +654,16 @@ type OpsGenieConfigResponder struct {
 	Username string `yaml:"username,omitempty" json:"username,omitempty"`
 
 	// team, user, escalation, schedule etc.
+	Type string `yaml:"type,omitempty" json:"type,omitempty"`
+}
+
+type OpsGenieConfigVisibleTo struct {
+	// One of those 3 should be filled.
+	ID       string `yaml:"id,omitempty" json:"id,omitempty"`
+	Name     string `yaml:"name,omitempty" json:"name,omitempty"`
+	Username string `yaml:"username,omitempty" json:"username,omitempty"`
+
+	// team, user
 	Type string `yaml:"type,omitempty" json:"type,omitempty"`
 }
 
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -592,9 +592,15 @@ type OpsGenieConfig struct {
 	UpdateAlerts bool                      `yaml:"update_alerts,omitempty" json:"update_alerts,omitempty"`
 }
 
-const opsgenieValidTypesRe = `^(team|teams|user|escalation|schedule)$`
+const (
+	opsgenieValidResponderTypesRe = `^(team|teams|user|escalation|schedule)$`
+	opsgenieValidVisibleToTypesRe = `^(team|teams|user)$`
+)
 
-var opsgenieTypeMatcher = regexp.MustCompile(opsgenieValidTypesRe)
+var (
+	opsgenieResponderTypeMatcher = regexp.MustCompile(opsgenieValidResponderTypesRe)
+	opsgenieVisibleToTypeMatcher = regexp.MustCompile(opsgenieValidVisibleToTypesRe)
+)
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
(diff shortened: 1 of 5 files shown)


config/notifiers.go
```diff
@@ -625,6 +626,24 @@ func (c *OpsGenieConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
 		}
 	}
 
+	for _, v := range c.VisibleTo {
+		if v.ID == "" && v.Username == "" && v.Name == "" {
+			return fmt.Errorf("opsGenieConfig visible_to %v has to have at least one of id, username or name specified", v)
+		}
+
+		if strings.Contains(v.Type, "{{") {
```
grobinson-grafana, line 634: I'm afraid this check doesn't work. Consider the following example:

```
this is an invalid template }}
```

I would just remove this if statement and check all VisibleTo's using `template.New("").Parse()`?

docs/configuration.md
grobinson-grafana, line 1: Excellent! Thanks for updating the docs 👍

config/notifiers.go
```diff
@@ -638,6 +657,16 @@ type OpsGenieConfigResponder struct {
 	Type string `yaml:"type,omitempty" json:"type,omitempty"`
 }
 
+type OpsGenieConfigVisibleTo struct {
+	// One of those 3 should be filled.
```
grobinson-grafana, line 661: Comment doesn't match the behavior in the code. The code allows for **at least one** to be filled, not **at most one**?

docs/configuration.md
```diff
@@ -1114,6 +1118,22 @@ responders:
 type: <tmpl_string>
 ```
 
+#### `<visible_to>`
+
+```yaml
+# Exactly one of these fields should be defined.
```
grobinson-grafana, line 1124: Same here.

notify/opsgenie/opsgenie.go
```diff
@@ -211,13 +219,44 @@ func (n *Notifier) createRequests(ctx context.Context, as ...*types.Alert) ([]*h
...
+		var visibleTos []opsGenieCreateMessageVisibleTo
+		for _, v := range n.conf.VisibleTo {
+			visibleTo := opsGenieCreateMessageVisibleTo{
+				ID:       tmpl(v.ID),
+				Name:     tmpl(v.Name),
+				Username: tmpl(v.Username),
+				Type:     tmpl(v.Type),
+			}
+
+			if visibleTo == (opsGenieCreateMessageVisibleTo{}) {
```
grobinson-grafana, line 231: Why would someone want to do this? Is it to catch badly written templates that produce no output?

notify/opsgenie/opsgenie.go
```diff
@@ -211,13 +219,44 @@ func (n *Notifier) createRequests(ctx context.Context, as ...*types.Alert) ([]*h
...
+				// Filter out empty responders. This is useful if you want to fill
+				// responders dynamically from alert's common labels.
+				continue
+			}
+
+			if visibleTo.Type == "teams" {
+				teams := safeSplit(visibleTo.Name, ",")
+				for _, team := range teams {
+					newVisibleTo := opsGenieCreateMessageVisibleTo{
+						Name: tmpl(team),
```
grobinson-grafana, line 241: Why template the name a second time?

notify/opsgenie/opsgenie.go
```diff
@@ -211,13 +219,44 @@ func (n *Notifier) createRequests(ctx context.Context, as ...*types.Alert) ([]*h
...
+						Type: tmpl("team"),
```
grobinson-grafana, line 242: This doesn't make sense to me. You could template it back to `user` after we've checked that it is `teams`.

notify/opsgenie/opsgenie.go
```diff
@@ -211,13 +219,44 @@ func (n *Notifier) createRequests(ctx context.Context, as ...*types.Alert) ([]*h
...
+				Type:     tmpl(v.Type),
+			}
+
+			if visibleTo == (opsGenieCreateMessageVisibleTo{}) {
+				// Filter out empty responders. This is useful if you want to fill
+				// responders dynamically from alert's common labels.
+				continue
+			}
+
+			if visibleTo.Type == "teams" {
```
grobinson-grafana, line 237: Where does `teams` plural come from btw? In the documentation all I can see is `team` and `user`?

> type field is mandatory for each item, where possible values are team and user

https://docs.opsgenie.com/docs/alert-api#create-alert

config/notifiers.go
grobinson-grafana: I suppose you could also check for `strings.Contains(v.Type, `}}`)?

notify/opsgenie/opsgenie.go
chlunde: Copy and paste from receivers, I did not catch that, thanks! Would you like me to remove the double templating in the code above too? Or have a dedicated PR for that? It does not change the results of the unit tests.

Fixed in new commit.

notify/opsgenie/opsgenie.go
chlunde: I think it should just be `Type: "team"`. Fixed in new commit.

Not sure if I understand the reference to `user` here?

chlunde (commented):
Thanks for the review, sorry for the delay. I added a new commit that addresses some of the code review comments.

But a lot of the comments apply to the existing code, so I would like to get your input again on them before making further changes. Could you mark the ones you think are OK as resolved, and comment on the others?

chlunde:
@grobinson-grafana Could you take another look? Or should we involve the authors of the original code? Alternatively I could make a stripped down version of the PR but that would make this field behave differently from receivers

PR #3965, opened 2024-08-21
Fix up linting issue

SuperQ:
Fix `govet` linting issue: `printf: non-constant format string`.


gotjosh (approved):

SuperQ:
It seems like we didn't completely fix the stable asset generation.

simonpasquier:
Yeah, some issue with a newer version of juice from what I can tell. See https://github.com/prometheus/alertmanager/pull/3972

Issue #3999, opened 2024-08-26
Add Metric Buckets for Notification Latency


kennytrytek-wf:
**Description**

We wanted to use `notification_latency_seconds` to indicate whether our alert manager was overly-delayed in sending notifications. The highest configured bucket boundary is 20 seconds, which is lower than we would like. A latency of up to 60 seconds is acceptable to us, but the histogram does not allow for alerting based on that latency threshold.

https://github.com/prometheus/alertmanager/blob/c7097ad76c07c7fc325292718115e3de9d0a125f/notify/notify.go#L309

**Want**

Add a 60s bucket to the metric.

* Alertmanager version: 0.27.0

grobinson-grafana:
I think the reason the largest bucket is `20s` is because the default peer timeout is `15s`. If it takes more than `15s` to send a notification then HA starts failing over. The other issue is depending on the Group interval, notifications might never be delivered as each time the Group interval timer elapses any in-flight notifications from the previous timer are canceled. Here is an example of that:

```
ts=2024-08-27T09:25:00.016Z caller=dispatch.go:166 level=debug component=dispatcher msg="Received alert" alert=[3fff2c2][active]
ts=2024-08-27T09:25:15.017Z caller=dispatch.go:526 level=debug component=dispatcher aggrGroup={}:{} msg=flushing alerts=[[3fff2c2][active]]
ts=2024-08-27T09:25:45.017Z caller=dispatch.go:357 level=error component=dispatcher msg="Notify for alerts failed" num_alerts=1 err="webhook/webhook[0]: notify retry canceled after 2 attempts: context deadline exceeded"
ts=2024-08-27T09:25:45.017Z caller=dispatch.go:526 level=debug component=dispatcher aggrGroup={}:{} msg=flushing alerts=[[3fff2c2][active]]
ts=2024-08-27T09:26:15.018Z caller=dispatch.go:357 level=error component=dispatcher msg="Notify for alerts failed" num_alerts=1 err="webhook/webhook[0]: notify retry canceled after 2 attempts: context deadline exceeded"
ts=2024-08-27T09:26:15.018Z caller=dispatch.go:526 level=debug component=dispatcher aggrGroup={}:{} msg=flushing alerts=[[3fff2c2][active]]
```

Given what I've written above, are you 100% sure that this is the case?

> A latency of up to 60 seconds is acceptable to us, but the histogram does not allow for alerting based on that latency threshold.

grobinson-grafana:
[...]

PR #4000, opened 2024-08-26
Add bucket to notification latency metric

kennytrytek-wf:
Fixes https://github.com/prometheus/alertmanager/issues/3999


kennytrytek-wf:
The test failure doesn't look related? I don't have permissions to rerun it.

Issue #4026, opened 2024-09-09
Telegram notifier ignores HTML tags


ntk148v:
**What did you do?**
Setup Prometheus and Alertmanager to send message to Telegram. Message is sent as raw text, no HTML tags are parsed.

![image](https://github.com/user-attachments/assets/091b16e4-017f-4b86-a3fa-ae8c1b21de0d)

**What did you expect to see?**

Message is formatted.

![image](https://github.com/user-attachments/assets/0a3fde08-cf65-4267-803d-8b8ec60ea982)

**What did you see instead? Under which circumstances?**

**Environment**

* System information:

* Alertmanager version:

Alertmanager image tag 'main':

```shell
docker pull quay.io/prometheus/alertmanager@sha256:1e5f1f22c0875ebffad1c39ffb5f831c94ba6dcf0be4d5e5d910cfaade11b1b7
```

* Prometheus version:

* Alertmanager configuration file:

```yaml
route:
  group_wait: 5s
  group_interval: 30s
  receiver: 'telegram'

templates:
  - iaas-alertmanager.tmpl

receivers:
  - name: 'telegram'
    telegram_configs:
      - bot_token: 'xxxx'
        chat_id: -xxxx
        message: '{{ template "telegram.iaas.message" . }}'
        parse_mode: 'HTML'
```

`iaas-alertmanager.tmpl` file (simple notification template with HTML tags)

```tmpl
{{ define "__iaas_alert_list" }}
{{ range . }}
---
🪪 <b>{{ .Labels.alertname }}</b>
<b></b>
{{- if .Annotations.summary }}
📝 Summary: {{ .Annotations.summary }}{{ end }}{{- if .Annotations.title }}
📕 Title: {{ .Annotations.title }}{{ end }}{{- if .Annotations.description }}
📖 Description:
<code>{{ .Annotations.description }}</code>{{ end }}{{ if .Annotations.assignees }}
🏷️ Labels:{{ if .Labels.severity }}
    <i>Instance</i>: <code>{{ .Labels.instance  }}</code>{{ end }}{{ if .Labels.job }}
    <i>Job</i>: <code>{{ .Labels.job }}</code>{{ end }}
Owner: <code>Iaas team</code>
Since: <code>{{ .StartsAt.Local.Format "02/01/2006 15:04:05"  }}</code>
{{ if not .EndsAt.IsZero }}
[...]

PR #4027, opened 2024-09-09
fix(telegram): passing ParseMode when sending message

ntk148v:
Before commit [1], the message parse mode value is the same as bot parse mode. Therefore, we don't need to pass ParseMode whenever sending message. But after that commit, if we don't pass SendOpts's ParseMode, the library will use the default mode which is empty string "", and HTML tags aren't parsed.

[1] https://github.com/tucnak/telebot/commit/864bef4e4d3a60b8079db4fdb51cac2e779349cd

Close-Issue: https://github.com/prometheus/alertmanager/issues/4026


AzureLeMoon:
@grobinson-grafana @simonpasquier  hey there, would you be so kind to review this pr and merge it if everything is in order? would be much appreciated.

PR #4029, opened 2024-09-17
jira integration template customfields

holger-waschke:
We using templates to fill the jira customfields, e.g.

```
{{- define "asp.priority" -}}
  {{- with index .Alerts 0 -}}
    {{- if .Labels.severity -}}
      {{- if eq .Labels.severity "disaster" -}}
        Sehr Hoch
        {{- else if eq .Labels.severity "critical" -}}
          Mittel
        {{- else if eq .Labels.severity "warning" -}}
          Mittel
        {{- else if eq .Labels.severity "standard" -}}
          Mittel
        {{- else -}}
          Mittel
      {{- end -}}
      {{- else -}}
        Mittel
    {{- end -}}
  {{- end -}}
{{- end -}}
```

This code changes templates the custom fields before creating the issue.


jkroepke:
Damn, seems like I forgot that.

asset/assets_vfsdata.go
jkroepke, line 1: To solve the CI failure from circle ci, I would recommend to revert this file. I have similar issue that the files always gets changed, but it should not commit.

holger-waschke: Missed the frontend part, after doing make all and make assets it looks good now

holger-waschke:
> Damn, seems like I forgot that. I would recommend to extend the tests.

No problem, I added a new template jira.host for the unit tests, please tell me if youre fine with this. otherwise we could use some existing template to test the customfields but I think with the new template it´s cleaner

jkroepke (approved):
LGTM - but I'm not an official maintainer.

jkroepke:
@simonpasquier is there any chance to get this in 0.28?

holger-waschke:
I´ll fix the merge conflicts If the maintainer confirm this to be merged.

holger-waschke:
@simonpasquier 
I fixed the merge conflicts, would be highly appreciated to get this into main as we need this feature in our production environment.

k0ste:
Closes #4110

holger-waschke:
@simonpasquier could you have a look into this PR? 

As we´re migration to Jira Cloud we really need to have the ability to use templated custom jira fields.

heliapb:
Hi any prediction on this getting merged? thanks

holger-waschke:
Hey @sysadmind any chance of this getting merged? We´re running this code in a production environment without any problems.
Would be great to use the official version again and not a fork. I have one more improvement to the Jira Integration coming and then we could switch to the official version

SuperQ:
Looks like there are some Go linting issues to fix.

notify/jira/jira.go
```diff
@@ -387,3 +395,67 @@ func (n *Notifier) doAPIRequestFullPath(ctx context.Context, method, path string
...
+			var inlineType interface{}
+			err := yaml.Unmarshal([]byte(parsed), &inlineType)
+			if err != nil || (inlineType != nil && reflect.TypeOf(inlineType).Kind() == reflect.String) {
+				// ignore error, thus the string is not an interface
+				return parsed, ok
+			}
+			return deepCopyWithTemplate(inlineType, tmplTextFunc)
+		} else {
+			return parsed, ok
+		}
```
SuperQ:
```suggestion
		}
		return parsed, ok
```

template/default.tmpl
```diff
@@ -217,3 +217,11 @@ Alerts Resolved:
 {{ define "rocketchat.default.emoji" }}{{ end }}
 {{ define "rocketchat.default.iconurl" }}{{ end }}
 {{ define "rocketchat.default.text" }}{{ end }}
+
+{{ define "jira.host" }}
```
Spaceman1701: I don't think there are any other places in the code which establish a convention around a `hostname` label. I'd prefer to avoid making assumptions about what labels alerts will have if we can.

I also think this is only used in a test, so it seems like it's leaking testing logic out into the API. Would you be open to moving this template definition into the test itself?

notify/jira/jira_test.go
```diff
@@ -347,6 +347,9 @@ func TestJiraTemplating(t *testing.T) {
 			cfg: &config.JiraConfig{
 				Summary:     `{{ template "jira.default.summary" . }}`,
 				Description: `{{ template "jira.default.description" . }}`,
+				Fields: map[string]any{
```
Spaceman1701, line 354: I think we should add test cases for each of the possible types that can appear in a field. As it is, I'm not sure that the recursive copy is actually being exercised by the tests.

notify/jira/jira.go
```diff
@@ -387,3 +395,67 @@ func (n *Notifier) doAPIRequestFullPath(ctx context.Context, method, path string
...
 	return responseBody, false, nil
 }
+
+// deepCopyWithTemplate returns a deep copy of a map/slice/array/string/int/bool or combination thereof, executing the
+// provided template (with the provided data) on all string keys or values. All maps are connverted to
+// map[string]interface{}, with all non-string keys discarded.
+// deepCopyWithTemplateFunc returns a deep copy of a map/slice/array/string/int/bool or combination thereof,
+// executing tmplTextFunc on all string keys or values. All maps are converted to map[string]interface{},
+// with all non-string keys discarded.
+func deepCopyWithTemplate(value interface{}, tmplTextFunc templateFunc) (interface{}, error) {
```
siavashs: This seems generic enough to be moved to a shared package so more integrations can benefit from it.
For example it can be used for PagerDuty in #3944

holger-waschke: Good point — I’ve moved it to template.go

template/default.tmpl
holger-waschke: You’re right — I’ve moved it into the test logic.
Currently, we’re only using two custom fields in our JIRA integration (hostname and environment), and both don’t exist in the current default.tmpl.
So I created them temporarily within the unit test without modifying default.tmpl.

notify/jira/jira_test.go
holger-waschke: done

holger-waschke:
Moved the DeepCopyWithTemplate function to template.go and improved its unit tests.

Spaceman1701 (approved):
With the small refactor and new tests, this LGTM. 

I think the behavior of `DeepCopyWithTemplate` is a little strange, but it's practical, so let's go for it 👍

heliapb:
Thank you guys, hope this can come in a new release soon 🤞

PR #4035, opened 2024-09-24
Add parseJSON template func

swchand:
This PR introduces a new parseJSON function to the template functions. The parseJSON function allows users to parse JSON strings within templates.

**Changes:**

- Added parseJSON function to the template functions map.
- The parseJSON function takes a JSON string as input and returns the parsed JSON object or an error if the JSON is invalid.
- Updated TestTemplateFuncs in template/template_test.go to include test cases for the parseJSON function:
- Valid JSON string test case.
- Invalid JSON string test case.

### DIFF
--- a/template/template.go
+++ b/template/template.go
@@ -15,6 +15,7 @@ package template
 
 import (
 	"bytes"
+	"encoding/json"
 	tmplhtml "html/template"
 	"io"
 	"net/url"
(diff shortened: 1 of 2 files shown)


SoloJacobs:
Hi @swchand ,

could you say a few words on the motivation for this change?

SoloJacobs:
Hi @swchand ,

we haven't heard back here in a while. Since we can't judge why this feature is useful, I will close this PR. Feel free to open a new one.

Kind regards

PR #4039, opened 2024-09-30
Fix panic during metrics registering

titolins:
Should fix https://github.com/grafana/grafana/issues/92707

We have fixed this before for the grafana metrics (https://github.com/grafana/grafana/pull/92725/files), but missed prometheus AM metrics.


titolins:
My bad, meant to open it in https://github.com/grafana/prometheus-alertmanager

PR #4007, opened 2024-08-30
Extend discord messages with content

mogoll92:
These changes are aimed to support discord content parameter besides with embedded message and allow to overcome limitation when you need to ping someone or group of people. (Ping over embedded messages doesn't trigger discord notifications)

Alertmanager already has struct that allows to send content.
```
type webhook struct {
	Content string         `json:"content"`
	Embeds  []webhookEmbed `json:"embeds"`
}
```

These changes make content optional and allow to send any text that doesn't break discord limits with embedded message.
To enable this user should apply the following changes:
```
receivers:
  - name: nothing
  - name: discord-success
    discord_configs:
      - webhook_url: ""
        content: '{{ template "discord.content" . }}'
        title: '{{ template "discord.title" . }}'
        message: '{{ template "discord.message" . }}'
 ```
 
 and override text in custom template like the following.
 ```
 {{ define "discord.content" }}
Hey you! <@6532081833221312329>
{{ end }}
 ```
 
 Here is the result.
Screenshot 2024-08-30 at 15 34 41

PR #4081, opened 2024-10-23
Discord notifier: Add username and avatar_url

featheredtoast:
Allow for custom username and avatar URLs to be set in discord notifications.

Add `username` and `avatar_url` to discord configuration, default empty string.

Re-introduce the rest of what was proposed in #3821 now that #4007 is merged


gotjosh (approved):
LGTM - I've added a test for it, tested it and fixed the documentation as these are not templetables. 

image

PR #4083, opened 2024-10-25
alertmanager: Replace typed fields with plain string values

jkroepke:
Fixes #4029 

The Jira API requests different fields value types depends on the custom field type itself.

Ref. https://developer.atlassian.com/server/jira/platform/jira-rest-api-examples/#setting-custom-field-data-for-other-field-types, for example:

```
fields:
    # Components
    components: { name: "Monitoring" }
    # Custom Field TextField
    customfield_10001: "Random text"
    # Custom Field SelectList
    customfield_10002: {"value": "red"}
    # Custom Field MultiSelect
    customfield_10003: [{"value": "red"}, {"value": "blue"}, {"value": "green"}]
```

While jiralert allowed field values with different types as well, it might be problematic in context of the Alertmanager.

* A new dependency `github.com/trivago/tgo` was necessary to cast all keys to strings. Ref. https://github.com/prometheus/alertmanager/pull/3590#discussion_r1816521316
* Looking at Prometheus Operator, Kubernetes CRDs doesn't [support such flexible types](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema), because type is not allowed inside `anyOf`. From Kubernetes point of view, all field values must be declared as string and an additional transformation logic is necessary to convert string values to complex values.

I feel that re-implement the logic from jiralert was design flaw while implement the jira notifier at Alertmanager.

WDYT: @simonpasquier @dswarbrick


dswarbrick:
As I commented in https://github.com/prometheus/alertmanager/pull/3590#discussion_r1817166452, the need for elaborate custom YAML unmarshaling code would be avoided if Alertmanager updated to yaml.v3. I see there was an earlier PR (https://github.com/prometheus/alertmanager/pull/3322) that attempted to update this, but it was not merged because:
> We have good reasons not to upgrade to v3, as common and prometheus are stuck with v2.

I see at least one other open PR (#3944) which seems to be encountering a similar scenario.

dswarbrick:
Where does this PR stand with respect to the soon-to-be-released v0.28.0?

jkroepke:
Had a conversation with in Slack about this and it feel not worth.

dswarbrick:
That's unfortunate. If Alertmanager v0.28.0 goes ahead with depending on the large and apparently no longer maintained trivago library, I will most likely patch out that functionality in Debian. I don't have the inclination to package the trivago library just because one application depends on one small function from it.

k0ste:
Closes #4110

PR #4084, opened 2024-10-27
Updated wechat api link

koeliga:
Outdated api link redirected to homepage, so replaced it with new api url


gotjosh:

Issue #4088, opened 2024-10-29
TextBlock used for `text` in msteamsv2 receiver must have `wrap: true` to display the content correctly


Lddeiva:
TextBlock used for `title` in msteamsv2 receiver needs `wrap: true` to display the content correctly.
```go
{
	Type:   "TextBlock",
	Text:   title,
	Weight: "Bolder",
	Size:   "Medium",
	Wrap:   true,
	Style:  "heading",
	Color:  color,
},
{
	Type: "TextBlock",
	Text: text,
},
```

Note: - The TextBlock used for `title` is using `wrap: true` already. https://github.com/prometheus/alertmanager/blob/ff8c09d60174f27850830ceed6be38b3cf86974b/notify/msteamsv2/msteamsv2.go#L162

When there is a lengthy line, the `title` is displayed correctly but the `text` is not. So, please add `wrap: true` in `text` as well.

```go
{
	Type:   "TextBlock",
	Text:   title,
	Weight: "Bolder",
	Size:   "Medium",
	Wrap:   true,
	Style:  "heading",
	Color:  color,
},
{
	Type: "TextBlock",
	Text: text,
	Wrap: true,
},

**Comment by grobinson-grafana** [comment]
Hi! 👋 Do you think you could open a PR with the fix (and some screenshots to show broken/fixed behavior)?

**Comment by Lddeiva** [comment]
I couldn't paste any screenshots but I built the `alertmanager` locally (with `wrap: true`) and tested the Microsoft Teams messages.

The existing code without wrap when used will display messages like this. The lines which are lengthy will be shown like this.
```yaml
Status: firing
Severity: warning
Description: Job ns-prod-app1/metric-exporter-2882016...
Instance: 10.79.0.233:8080
```

The new code (with `wrap: true`) wraps the text and shows the lines within the viewable area.
```yaml
Status: firing
Severity: warning
Description: Job ns-prod-app1/metric-exporter-2882016
has failed to complete
Instance: 10.79.0.233:8080
```

Note: - This might be due to the `AdaptiveCard` format in which not entire screen area is used to display the message (we might need to use `ColumnSet` with `Width` but that requires additional changes and testing). As you can see in the screenshot below, only half of the viewable area is used to display the card.
![image](https://github.com/user-attachments/assets/63be3a7f-b39c-449b-964a-8c17c154eac8)

The fix is something like this. I'll submit an MR by referring this issue.

```go
{
	Type:   "TextBlock",
	Text:   title,
[...]

PR #4106, opened 2024-11-04
Add `Wrap: true` to `TextBlock` named `text` in msteamsv2 receiver

Lddeiva:
`TextBlock` used for `text` in msteamsv2 receiver must have `wrap: true` to display the content correctly #4088


EricHorst:
It would be good if this made it into the 0.28.0 release. msteamsv2 notifications are nearly useless without.

Issue #3962, opened 2024-08-19
 Test certificates expiring in <2y


NightTsarina:
Hi,

A very minor issue: the CA certificates used for testing are due to expire in May 2026, and so it would be good to re-generate them in advance, to avoid build failures in frozen releases.

```
cluster/testdata/certs/ca.pem: 
        Validity
            Not Before: May  5 16:16:00 2021 GMT
            Not After : May  4 16:16:00 2026 GMT
```

This is mostly relevant to distros like Debian, where we ensure that all packages can be rebuilt from source during a release's support period, but it would be nice not to have to patch this.

Original bug report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1078634

PR #4112, opened 2024-11-06
Set testdata / example CA expiry to 20 years from issue date

dswarbrick:
Also add human-readable x509 text format (as requested by @SuperQ in related PR https://github.com/prometheus/prometheus/pull/14696)

Fixes: #3962

### DIFF
--- a/cluster/testdata/certs/ca.pem
+++ b/cluster/testdata/certs/ca.pem
@@ -1,8 +1,66 @@
+Certificate:
+    Data:
+        Version: 3 (0x2)
+        Serial Number:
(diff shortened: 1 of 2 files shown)


dswarbrick:
Incidentally, the expiry dates of the certs (both CA and server/client) seem to be a bit wild and inconsistent. The original CA had a lifetime of only 5 years, whilst the `node{1,2}.pem` certs had a 100 year lifetime (!). Usually the CA lifetime would be longer than host/node certs (although 100 years is a bit excessive).

It might be worth regenerating all the test / example certs with more sane lifetimes - and possibly consider bumping the key length to 4096 bits.

TheMeier:
@dswarbrick Thank you for the contribution could you rebase that?

PR #4113, opened 2024-11-08
Logging improvements

tjhop:
Some logging fixes to address PR feedback from #4089; correct error levels in a few spots and improve consistency.

### DIFF
--- a/notify/webhook/webhook.go
+++ b/notify/webhook/webhook.go
@@ -82,12 +82,10 @@ func (n *Notifier) Notify(ctx context.Context, alerts ...*types.Alert) (bool, er
 
 	groupKey, err := notify.ExtractGroupKey(ctx)
 	if err != nil {
-		// @tjhop: should we `return false, err` here as we do in most
-		// other Notify() implementations?
-		n.logger.Error("error extracting group key", "err", err)
+		return false, err
 	}
 
-	// @tjhop: should we debug log the key here like most other Notify() implementations?
+	n.logger.Debug("extracted group key", "incident", groupKey)
 
 	msg := &Message{
 		Version:         "4",
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -446,8 +446,7 @@ Loop:
 			break Loop
 		case <-t.C:
 			if err := runMaintenance(doMaintenance); err != nil {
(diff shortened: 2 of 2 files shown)


tjhop:
@grobinson-grafana for your comment [here](https://github.com/prometheus/alertmanager/pull/4089#discussion_r1824474224), do you have a preference what key name we align on?

grobinson-grafana:
> @grobinson-grafana for your comment [here](https://github.com/prometheus/alertmanager/pull/4089#discussion_r1824474224), do you have a preference what key name we align on?

Group key is the correct name.

SuperQ:
Ping, any additional work needed here?

tjhop:
Sorry for the noise here -- I needed to resolve a conflict and accidentally rebased on my copy of main instead of upstream. This is all cleaned up and ready for review now :+1:

notify/discord/discord.go
```diff
@@ -95,10 +95,11 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 		return false, err
 	}
 
-	n.logger.Debug("extracted group key", "key", key)
+	logger := n.logger.With("group_key", key)
+	logger.Debug("extracted group key")
```
siavashs, line 99: Shall we move this inside `notify.ExtractGroupKey()` instead of repeating the same line of code in all integrations?

tjhop: Sounds reasonable, but I'm trying to think through the most sane way to do that. Currently, `notify.ExtractGroupKey()` has the following signature:

https://github.com/prometheus/alertmanager/blob/29d491e7cdfe4fdb2b143e967d1f46c15b35387e/notify/util.go#L159-L166

Do we want to update the function signature to accept a logger as well? 
Alternatively, I could use the top level `slog.Debug()` function inside of `notify.ExtractGroupKey()` if we want to shim a call in to main somewhere after we initialize the logger to `slog.SetDefault(logger)`. 

Thoughts?

siavashs: I'd say skip it, not worth the effort.

tjhop: Agreed. Still a good point, maybe a better option will present in the future

PR #4119, opened 2024-11-13
replace Resolved with ResolvedAt to reduce amount call to time.Now()

evkuzin:
with big number of inhibitions I see pretty bad picture when we spend plenty of time in time.Now()
This can be fixed by calling it once per `hasEqual`
image

### DIFF
--- a/inhibit/inhibit.go
+++ b/inhibit/inhibit.go
@@ -230,10 +230,11 @@ func NewInhibitRule(cr config.InhibitRule) *InhibitRule {
 // is returned. If excludeTwoSidedMatch is true, alerts that match both the
 // source and the target side of the rule are disregarded.
 func (r *InhibitRule) hasEqual(lset model.LabelSet, excludeTwoSidedMatch bool) (model.Fingerprint, bool) {
+	now := time.Now()
 Outer:
 	for _, a := range r.scache.List() {
 		// The cache might be stale and contain resolved alerts.
-		if a.Resolved() {
+		if a.ResolvedAt(now) {
 			continue
 		}
 		for n := range r.Equal {


evkuzin:
@w0rm WDYT?

grobinson-grafana:
Seems reasonable to me. What does the profile look like with the change?

evkuzin:
@grobinson-grafana thx for taking a look!
attaching a picture from pyroscope. On the left is a binary with a patch and on the right second node from the same cluster
image

evkuzin:
I suppose we can go further and replace this comparison https://github.com/prometheus/alertmanager/pull/4119/files#diff-279dfe3df5d26e996ad18fe4a97c9dea0c7c4d5079c11c96d4a4644e942602dfL240 with something better. We already using FNV-1A fingerprint and comparing uint64 instead of strings would dramatically improve this other bit where we waste tonn of time (runtime.mapaccess1_faststr ). But for now I just want this quick win.

evkuzin:
Regarding the failed test. it seems like [this](https://github.com/prometheus/alertmanager/blob/main/test/with_api_v2/acceptance.go#L342) is not enough for alertmannger to start on CI? Do I need to fix it?

grobinson-grafana:
I would benchmark the difference between `mapaccess1_faststr ` and `mapaccess1_fast64` to be sure there is a considerable speed up, as from what I can tell `fast64` also calculates hashes on the int64.

grobinson-grafana:
I would benchmark the difference between `mapaccess1_faststr` and `mapaccess1_fast64` to be sure there is a considerable speed up, as from what I can tell `fast64` also calculates hashes on the int64. It might be that `fnv1` + `fast64` is slower than `faststr`.

evkuzin:
fair enough. Anyway, I didnt plan to include this in the current PR. Does this look like it can be accepted by any chance?

grobinson-grafana (approved):

SuperQ (approved):

SuperQ:
I would like one of the Alertmanager maintainers to look at this before merging.

evkuzin:
I suppose santa won't come this year, shame. Merry christmas!

PR #4158, opened 2024-12-06
Update config to allow showing secret values when marshaled

wallee94:
I have a deployment where a sidecar between the config-reloader and alertmanager loads and unmarshals a base configuration, extends it with more routes and receivers, and then marshals it back for the alertmanager. 

image

However, the default Marshal implementations in the receivers hide all Secret and SecretURLs when marshaling the config back to a YAML. We have to duplicate all the receiver structs in our code to be able to switch them to plain strings and URLs.

I'd be helpful to be able to re-use the structs already here in the alertmanager. 

This is already supported by some of the Prometheus common configs by using a global `MarshalSecretValue` [here](https://github.com/prometheus/common/blob/aeb616a795915a559fe308885ab6de6101180ed6/config/config.go#L33).

For now, I'm adding support for a `MarshalSecretValue` here as well. I think the Secret in this repo could also be replaced for the Secret in Prometheus common configs, but I'd like to try this first, as switching to the common Secret would be a longer PR.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -39,6 +39,11 @@ const secretToken = "<secret>"
 
 var secretTokenJSON string
 
+// MarshalSecretValue if set to true will expose Secret type
+// through the marshal interfaces. Useful for outside projects
+// that load and marshal the Alertmanager config.
+var MarshalSecretValue bool = commoncfg.MarshalSecretValue
+
 func init() {
 	b, err := json.Marshal(secretToken)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


wallee94:
I didn't notice that this was already discussed here: https://github.com/prometheus/alertmanager/issues/1985

I understand that the main blocker at the time was having consistency across all projects, but looks like this is something that can already be enabled in Prometheus configs

SuperQ (approved):
I think this is fine, org level consistency shouldn't be blocked on this kind of improvement.

PR #4159, opened 2024-12-06
Enable templating for Jira project and issue_type

wallee94:
Update Jira notifications to support templating in `project` and `issue_type`.

I'd like to be able to open issues in multiple Jira projects with different priorities and issue types depending on the alert labels or annotations.

Some Jira fields, like `summary` and `priority`, already allow templating, but it's missing in these two.


SuperQ (approved):
LGTM, thanks!

PR #4162, opened 2024-12-16
feat: add api http metrics

siavashs:
- add api http metrics for requests and latency
- set `version="v2"` label on all api metrics
- update existing metric help strings for distinction (api vs. web)


api/api.go
```diff
@@ -132,19 +136,39 @@ func New(opts Options) (*API, error) {
 		return nil, err
 	}
 
+	latency := prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:        "alertmanager_api_http_request_duration_seconds",
+			Help:        "Histogram of latencies for api HTTP requests.",
+			ConstLabels: prometheus.Labels{"version": "v2"},
+			Buckets:     []float64{.05, 0.1, .25, .5, .75, 1, 2, 5, 10, 20},
```
SuperQ: Should we use `prometheus.DefBuckets` here?

api/api.go
```diff
@@ -132,19 +136,39 @@ func New(opts Options) (*API, error) {
 		return nil, err
 	}
 
+	latency := prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:        "alertmanager_api_http_request_duration_seconds",
+			Help:        "Histogram of latencies for api HTTP requests.",
+			ConstLabels: prometheus.Labels{"version": "v2"},
```
SuperQ: What is this for?

api/api.go
```diff
@@ -132,19 +136,39 @@ func New(opts Options) (*API, error) {
...
+			Buckets:     []float64{.05, 0.1, .25, .5, .75, 1, 2, 5, 10, 20},
+		},
+		[]string{"code", "handler", "method"},
+	)
+	receivedRequests := prometheus.NewCounterVec(prometheus.CounterOpts{
+		Name:        "alertmanager_api_http_requests_received_total",
+		Help:        "The total number of received api HTTP requests.",
+		ConstLabels: prometheus.Labels{"version": "v2"},
+	}, []string{"code", "handler", "method"})
```
SuperQ: This metric duplicates the latency histogram's built-in counter. We don't need it.

SuperQ (changes requested):

api/api.go
siavashs: Alertmanager has both API v1 and v2, this just indicates that the metric only measures v2 requests.
But we can drop it since v1 is deprecated and will eventually be removed.

SuperQ: Ahh, that makes sense. If you want to keep the label, let's use `api_version` to disambiguate it from other version related labels.

Typically when we do this kind of thing, we use an existing label metric, but add the label in the top-level creation and do something like `metric.With(prometheus.Labels{"api_version":"v2"})`. This way we don't have to use `ConstLabels`.

siavashs: I removed the const label, hopefully we will drop API v1 soon :)

siavashs:
CircleCI is flaky as always, i think we should drop it and just use Github actions.

SuperQ:
We've been slowly migrating things to GitHub Actions. While I'm not a huge fan, it's the most supported free pipeline system right now.

I have a backlog TODO to refactor the promci actions into separate repos for different steps. We started it as a monorepo, which really doesn't work for actions.

api/api.go
```diff
@@ -132,8 +135,14 @@ func New(opts Options) (*API, error) {
 		return nil, err
 	}
 
-	// TODO(beorn7): For now, this hardcodes the method="get" label. Other
-	// methods should get the same instrumentation.
+	latency := prometheus.NewHistogramVec(
```
SuperQ: Thinking about this a bit more. What if we pass the `alertmanager_http_request_duration_seconds` metric var from `main.go` to `api.New()`? We'd need to add `code` to the metric, but I think it would make more sense to have a single metric since it's all on the same HTTP server/mux.

siavashs: Good idea, done!

siavashs:
Lint issues are unrelated!
We run lint in a dedicated job and also in the CI job.

SuperQ:
Fix for test failure: https://github.com/prometheus/alertmanager/pull/4613

siavashs:
thanks, rebased!

PR #4163, opened 2024-12-17
Add silence logging option

anarcat:
Past silences are really important to postmortem our alert management operations.
We need to get "when" and "what kind of" create/edit/delete silences.
So this PR added optional function that logging silence change.

Reroll of #2536 to address review item.


silence/silence.go
```diff
@@ -1053,3 +1067,35 @@ func openReplace(filename string) (*replaceFile, error) {
...
 	return rf, nil
 }
+
+// logging silence status changes.
+//
+// XXX: This rewrites some code present in cli/format/format.go,
+// labelsMatcher() and cli/format/format_extended.go,
+// FormatSilences(). Refactoring this seems a little out of scope for
+// now.
+func (s *Silences) logginSilences(msg string, sil *pb.Silence) {
```
SuperQ: Typo?

```suggestion
func (s *Silences) loggingSilences(msg string, sil *pb.Silence) {
```

SuperQ (commented):
Overall LGTM.

SuperQ:
Needs a rebase.

silence/silence.go
anarcat: i renamed this to logSilence, and rebased everything.

anarcat:
i rebased, but the history is not super clean, should i squash?

TheMeier:
Squashing the commit sounds like a good idea to me. Just a question would it make sense to add a minimal test for the new function?

anarcat:
probably! not sure where to start with that though

TheMeier:
How about something like this https://github.com/TheMeier/alertmanager/commit/ada4ea8dc92d4464810ab23d80b09ab69f29ccfd

anarcat:
> How about something like this [TheMeier@ada4ea8](https://github.com/TheMeier/alertmanager/commit/ada4ea8dc92d4464810ab23d80b09ab69f29ccfd)

that looks fantastic.

TheMeier:
So how about you steal it into your branch and then squash everything ;)

anarcat:
> So how about you steal it into your branch and then squash everything ;)

done!

... and rebased.

anarcat:
so this has conflicted again, but before i start fixing this, perhaps it would be best to get a confirmation it's worth my time... ;) is this go or not? especially in view of log restructuring that's happening in parallel in https://github.com/prometheus/alertmanager/pull/5072

anarcat:
conflicts resolved, rebased.

anarcat:
whoohoo! thanks! :)

PR #4166, opened 2024-12-18
Fix build info metric

SuperQ:
Use the new version collector to expose the `alertmanager_build_info` metric. This fixes an incorrect deprecation migration done in #3806.


machine424 (approved):
lgtm, thanks.

PR #4196, opened 2025-01-10
Chore: Update automemlimit to v0.7.0

dongjiang1989:


SuperQ (approved):
Thanks, usually we use dependabot for these updates.

dongjiang1989:
> Thanks, usually we use dependabot for these updates.

Thanks @SuperQ 
Is it necessary to set dependabot time interval to daily? Discover dependency version updates faster

PR #4204, opened 2025-01-15
chore(test): support race detector for arm64

magiskboy:
According to https://go.dev/doc/articles/race_detector, go supported race detector for arm64

Signed-off-by: Nguyen Khac Thanh <nguyenkhacthanh244@gmail.com>


SuperQ (changes requested):
This file is synced from `prometheus/prometheus`. All changes must be made there.

PR #4206, opened 2025-01-16
mixin: longer rate interval for Alertmanager[Cluster]FailedToSendAlerts

jan--f:
5m is an unfortunate value since its the same as the default group_interval. If only a single alert fires and a single notifier fails consistently the alert will never fire, since the rate[5m] will be 0 after less then the group_interval.


marioferh (commented):
/lgtm

gotjosh (approved):
LGTM

PR #4214, opened 2025-01-20
Fix descriptions for --enable-feature

am97:
Currently, passing multiple `--enable-feature` doesn't work:

```console
$ docker run --rm -it --entrypoint sh prom/alertmanager:v0.28.0
/alertmanager $ alertmanager --config.file=/etc/alertmanager/alertmanager.yml --enable-feature=auto-gomemlimit --enable-feature=auto-gomaxprocs
alertmanager: error: flag 'enable-feature' cannot be repeated, try --help
```

The intended syntax seems to be a single `--enable-feature` flag, with a comma-separated list of features as value. This is what is tested here:

https://github.com/prometheus/alertmanager/blob/3b06b97af4d146e141af92885a185891eb79a5b0/featurecontrol/featurecontrol_test.go#L40-L44

### DIFF
--- a/cli/root.go
+++ b/cli/root.go
@@ -155,7 +155,7 @@ func Execute() {
 	app.Flag("timeout", "Timeout for the executed command").Default("30s").DurationVar(&timeout)
 	app.Flag("http.config.file", "HTTP client configuration file for amtool to connect to Alertmanager.").PlaceHolder("<filename>").ExistingFileVar(&httpConfigFile)
 	app.Flag("version-check", "Check alertmanager version. Use --no-version-check to disable.").Default("true").BoolVar(&versionCheck)
-	app.Flag("enable-feature", fmt.Sprintf("Experimental features to enable. The flag can be repeated to enable multiple features. Valid options: %s", strings.Join(featurecontrol.AllowedFlags, ", "))).Default("").StringVar(&featureFlags)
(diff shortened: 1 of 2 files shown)


cli/root.go
```diff
@@ -155,7 +155,7 @@ func Execute() {
 	app.Flag("timeout", "Timeout for the executed command").Default("30s").DurationVar(&timeout)
 	app.Flag("http.config.file", "HTTP client configuration file for amtool to connect to Alertmanager.").PlaceHolder("<filename>").ExistingFileVar(&httpConfigFile)
 	app.Flag("version-check", "Check alertmanager version. Use --no-version-check to disable.").Default("true").BoolVar(&versionCheck)
-	app.Flag("enable-feature", fmt.Sprintf("Experimental features to enable. The flag can be repeated to enable multiple features. Valid options: %s", strings.Join(featurecontrol.AllowedFlags, ", "))).Default("").StringVar(&featureFlags)
+	app.Flag("enable-feature", fmt.Sprintf("Comma-separated experimental features to enable. Valid options: %s", strings.Join(featurecontrol.AllowedFlags, ", "))).Default("").StringVar(&featureFlags)
```
grobinson-grafana, line 158:
```suggestion
	app.Flag("enable-feature", fmt.Sprintf("Experimental features to enable, comma separated. Valid options: %s", strings.Join(featurecontrol.AllowedFlags, ", "))).Default("").StringVar(&featureFlags)
```

grobinson-grafana (commented):
I would prefer we put what it does first and the expected format second. Otherwise LGTM!

gotjosh (approved):

PR #3799, opened 2024-04-03
Adaptive Cards Support

6fears7:
To continue #3503. 

Credit to @aw1cks provided the original code, I slightly tweaked it. 

Notables from that feature request:

- MessageCards is deprecated, MSTeams integration should support AdaptiveCards
- The original proposal suggested that by switching the payload format to adaptive web cards, the responsibility for creating valid adaptive cards is entirely upon the user's template. Additionally, perhaps this could be a config option.
- I rest the case with the Alertmanager team for the option of a config option versus hard cutover.
- I initially tried to keep the new template proposal below simple with raw string, but receiving multiple grouped alerts was so ugly that I created an alternative default for historical purposes

### Alertname: "Example1"

image

## Resolved
image

## Label / Annotations View:
image

### Uses .CommonLabels
image

<details>
<summary>Click to view Template</summary>

```
{{ define "new.text" }}
{
    "$schema": "http://adaptivecards.io/schemas/adaptive-card.json",
    "type": "AdaptiveCard",
    "version": "1.2",
    "padding": "None",
    "msteams": {
        "width": "Full"
    },
    
    "body": [
        {
            "type": "Container",
            "id": "alert-msg",
            "padding": "Default",
            "items": [
                {
                    "type": "TextBlock",
                    "id": "alert-summary-title",
                    "text": "[{{- if gt (len .Alerts.Firing) 0 }}FIRING: {{ .Alerts.Firing | len }}] 🔥 {{end}} {{- if gt (len .Alerts.Resolved) 0 }}RESOLVED: {{ .Alerts.Resolved | len }}] ✅ {{end}} {{ with index .Alerts 0 -}}{{ .Labels.alertname }}{{ end }}",
                    "weight": "Bolder",
                    "color": "{{- if gt (len .Alerts.Firing) 0 }}Attention{{end}}{{- if gt (len .Alerts.Resolved) 0 }}Good{{end}}",
                    "size": "ExtraLarge",
                    "horizontalAlignment": "Left"
                },
                {
                    "type": "Container",
                    "id": "alert-summary-container",
                    "padding": "None",
                    "items": [
                        {
                            "type": "TextBlock",
                            "id": "alert-summary-description",
                            "text": "{{ .Alerts.Firing | len }} alerts are firing",
                            "wrap": true
                        },
                         {
                            "type": "ActionSet",
                            "id": "alert-silence-action",
                            "actions": [
                                {
                                    "type": "Action.OpenUrl",
                                    "id": "silence",
[...]

PR #4215, opened 2025-01-21
feat: webhook template

baerwang:


grobinson-grafana:
I appreciate this a popular request, but I am reluctant to support templating of JSON in Alertmanager. The main reason for this is that usability is very poor, and writing correct JSON in templating languages is very difficult once you start using conditionals and loops (i.e. trailing commas). I raised the same concerns on https://github.com/prometheus/alertmanager/pull/3799#issuecomment-2040767899. I'll leave it up to the other maintainers to make a decision on whether this is something we want in Alertmanager.

TheMeier:
related https://github.com/prometheus/alertmanager/pull/4243

Amejonah1200:
The reason to wanting that feature would be to forward alerts to another alertmanager.

Instead of using go templating, why no use something like gojq instead?

kunlun122:
We also need this feature too.

TheMeier:
Even if templating the webhook payload would be accepted, I think it should be done in the existing webhook notifier and not in a new one

SoloJacobs (commented):
Hi,

thanks for this contribution. As already pointed out, we should update the existing webhook receiver. We are trying to implement this here: https://github.com/prometheus/alertmanager/pull/5011

In case you have any use cases and specific requirements, feel free to comment on the PR.

Kind regards

PR #4250, opened 2025-02-12
bump go dependency

yurnov:
Security-related bump of go dependency, several high vulnerabilities of dependencies, i.e. CVE-2024-45338


SuperQ:
We use dependabot to manage dependencies.

PR #4290, opened 2025-03-03
chore: log when there is no route matched

eyazici90:
This is to get some insights in case no route is matched for the alert. This is something we almost don't want at all. Any alert should have a matcher under normal circumstances. Would be good to log to ease troubleshooting.

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -171,7 +171,11 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
 			}
 
 			now := time.Now()
-			for _, r := range d.route.Match(alert.Labels) {
+			routes := d.route.Match(alert.Labels)
+			if len(routes) == 0 {
+				d.logger.Info("No routes matched the alert", "alert", alert)
+			}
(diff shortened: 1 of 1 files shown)


eyazici90:
May I get a review @grobinson-grafana ? 🙏

dispatch/dispatch.go
```diff
@@ -171,7 +171,11 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
 			}
 
 			now := time.Now()
-			for _, r := range d.route.Match(alert.Labels) {
+			routes := d.route.Match(alert.Labels)
+			if len(routes) == 0 {
```
grobinson-grafana, line 175: I don't think this is possible? I.e. `len(routes) == 0` can never be true. Every alert will always match at least one route, even if it is the top-level route. And if you don't have a top-level route, the Alertmanager will fail to start?

eyazici90: ahh right. I just realized https://github.com/prometheus/alertmanager/blob/main/config/config_test.go#L50. 🤦  Closing the PR

PR #4297, opened 2025-03-07
Jira: Add option to disable issue updates

waltherlee:
Currently, every Jira notification [sends a request](https://github.com/prometheus/alertmanager/blob/main/notify/jira/jira.go#L109-L114) that overwrites user changes to issue fields (title, description, priority, etc). 

Losing changes may be unexpected for many Jira users, and this PR adds a receiver config to be able to skip the update request for existing issues without affecting resolve and reopen transitions.


grobinson-grafana:
I don't know about the flag `DisableFieldUpdates`. This feels like something that should be consistent. We should look at how it works for other integrations like Pagerduty and do the same here.

wallee94:
Sure, I can look into that. I'm pretty sure they don't overwrite, so that works too. On the other hand, without a flag, it will be a breaking change for people expecting the current implementation, which propagates changes in the receiver config to existing open issues.

Spaceman1701:
Just tagging this as related/similar to #4621

SuperQ:
I think this is solved by https://github.com/prometheus/alertmanager/pull/4621

SoloJacobs:
Hey,

thanks for this PR, there hasn't been an update here in a while, and the issue seems mostly addressed.

Kind regards

PR #4304, opened 2025-03-10
Improve documentation around group timers

grobinson-grafana:
This commit improves the documentation around the group timers such as group_wait, group_interval and repeat_interval. It offers a more in-depth explanation of how these timers work without going into detail about how they are implemented in the code.


SuperQ (approved):
Nice

PR #4308, opened 2025-03-11
Update README information about api/v2 and api/v1

nikodemas:
README still references `/api/v1` and provides a message that discourages to use `/api/v2`, so I think it would be nice to change this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -195,12 +195,10 @@ of the HTTP handlers themselves. The API specification can be found in
 accessed [here](http://petstore.swagger.io/?url=https://raw.githubusercontent.com/prometheus/alertmanager/main/api/v2/openapi.yaml).
 Clients can be easily generated via any OpenAPI generator for all major languages.
 
-With the default config, endpoints are accessed under a `/api/v1` or `/api/v2` prefix.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -195,13 +195,11 @@ of the HTTP handlers themselves. The API specification can be found in
 accessed [here](http://petstore.swagger.io/?url=https://raw.githubusercontent.com/prometheus/alertmanager/main/api/v2/openapi.yaml).
 Clients can be easily generated via any OpenAPI generator for all major languages.
 
-With the default config, endpoints are accessed under a `/api/v1` or `/api/v2` prefix.
+APIv2 is accessed via the `/api/v2` prefix. APIv1 was deprecated `0.16.0` and is removed as of version `0.27.0`.
```
grobinson-grafana, line 198:
```suggestion
APIv2 is accessed via the `/api/v2` prefix. APIv1 was deprecated in `0.16.0` and is removed as of version `0.27.0`.
```

grobinson-grafana (commented):
Apologies, one small mistake I made with the previous suggestion.

PR #4317, opened 2025-03-27
Update Go

SuperQ:
* Update Go build to 1.24.
* Update minimum Go version to 1.23.
* Fixup various testifylint issues.
* Update swagger generation.


.github/workflows/mixin.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: install Go
         uses: actions/setup-go@0a12ed9d6a96ab950c8f026ed9f722fe0da7ef32 # v5.0.2
         with:
-          go-version: 1.23.x
+          go-version: 1.24.x
```
jkroepke, line 17:
```suggestion
          go-version-file: 'go.mod'
```

Thats what I favorite at least. It uses the go directive from go.mod

jkroepke (approved):
I have no clue about the CircieCI part and no idea what the reason for using it.

Otherwise LGTM. There is an 2.0 release for golangci-lint - maybe a different PR.

Do you want to implement low hanging fruits from go 1.24 as well? For example, they are some benefits related to bench tests. Instead the for increment, just use `for b.Loop()`

https://www.bytesizego.com/blog/go-124-new-benchmark-function

.github/workflows/mixin.yml
SuperQ: But won't that make it use 1.23? I don't think that's a good idea.

SuperQ:
> There is an 2.0 release for golangci-lint

We sync the golangci-lint version from prometheus/prometheus. We will update it there first.

SuperQ:
> Do you want to implement low hanging fruits from go 1.24 as 

No, I'm just trying to get the minimum done here to make the upgrade happen.

PR #4323, opened 2025-03-31
Fix logged notification count

rajagopalanand:
A simple fix to correct the log message. In cases where context cancelation occurs, the logs show notification was attempted twice when in fact it was only attempted once

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -828,7 +828,7 @@ func (r RetryStage) exec(ctx context.Context, l *slog.Logger, alerts ...*types.A
 	}
 
 	for {
-		i++
+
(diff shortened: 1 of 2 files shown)


notify/notify.go
```diff
@@ -848,6 +848,7 @@ func (r RetryStage) exec(ctx context.Context, l *slog.Logger, alerts ...*types.A
 		default:
 		}
 
+		i++
```
grobinson-grafana, line 851: Doesn't this mean that if Line 844 is true, the error message contains `0 attempts`, which is incorrect?

grobinson-grafana (approved):

PR #4341, opened 2025-04-02
chore: rephrase classic mode enabled log message

joobisb:
Fixes https://github.com/grafana/mimir/issues/10449

### DIFF
--- a/featurecontrol/featurecontrol.go
+++ b/featurecontrol/featurecontrol.go
@@ -120,7 +120,7 @@ func NewFlags(logger *slog.Logger, features string) (Flagger, error) {
 			logger.Warn("Experimental receiver name in metrics enabled")
 		case FeatureClassicMode:
 			opts = append(opts, enableClassicMode())
-			logger.Warn("Classic mode enabled")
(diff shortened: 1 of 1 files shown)


featurecontrol/featurecontrol.go
```diff
@@ -120,7 +120,7 @@ func NewFlags(logger *slog.Logger, features string) (Flagger, error) {
 			logger.Warn("Experimental receiver name in metrics enabled")
 		case FeatureClassicMode:
 			opts = append(opts, enableClassicMode())
-			logger.Warn("Classic mode enabled")
+			logger.Warn("UTF-8 strict mode is disabled; non-UTF-8 characters will be accepted")
```
grobinson-grafana, line 123: This doesn't quite explain what classic mode does, as it's a little more then UTF-8 strict mode being disabled. Let me take a look at the Mimir issue to see how we can help here because "classic mode" is documented in the Alertmanager UTF-8 migration guide for over 12 months now, and I don't want to create additional confusion renaming bits of it.

PR #4343, opened 2025-04-03
fix: set context timeout for resolvePeers

eyazici90:
# Context

Given resolvePeers is context aware, yet, uses background that may lead to blocking call. This is due to the fact that `LookupIPAddr` is called under the hood that does dns look up which may be blocked in some platforms.

In case of dns lookups are blocked, it may lead to hanging goroutine due to the fact that main goroutine is being used for the call that leads to application hanging & being unresponsive as well as keep getting restarted by kube with no info in the app.

Below logs are taken from one of the alertmanager kube/pods.

> │ ts=2025-04-03T11:12:38.668Z caller=main.go:181 level=info msg="Starting Alertmanager" version="(version=0.27.0, branch=HEAD, revision=0aa3c2aad14cff039931923ab16b26b7481783b5)"                                                                                                                                                                                       │
│ ts=2025-04-03T11:12:38.668Z caller=main.go:182 level=info build_context="(go=go1.21.7, platform=linux/amd64, user=root@22cd11f671e9, date=20240228-11:51:20, tags=netgo)"

 Cillumpolicy was not applied to allowe to dns resolve yet. This led to app hangs with no logs & keep getting restarted due to /healthy is not registered. Which was pretty painful to troubleshoot.

This was due to [LookupIPAddr](https://github.com/prometheus/alertmanager/blob/main/cluster/cluster.go#L736) hanging on main goroutine.

```go
select {
	case <-ctx.Done():
		// Our context was canceled. If we are the only
		// goroutine looking up this key, then drop the key
		// from the lookupGroup and cancel the lookup.
		// If there are other goroutines looking up this key,
		// let the lookup continue uncanceled, and let later
		// lookups with the same key share the result.
		// See issues 8602, 20703, 22724.
		if r.getLookupGroup().ForgetUnshared(lookupKey) {
			lookupGroupCancel()
			go dnsWaitGroupDone(ch, func() {})
		} else {
			go dnsWaitGroupDone(ch, lookupGroupCancel)
		}
		err := newDNSError(mapErr(ctx.Err()), host, "")
		if trace != nil && trace.DNSDone != nil {
			trace.DNSDone(nil, false, err)
		}
		return nil, err
	case r := <-ch:
		dnsWaitGroup.Done()
		lookupGroupCancel()
		err := r.Err
		if err != nil {
			if _, ok := err.(*DNSError); !ok {
				err = newDNSError(mapErr(err), host, "")
			}
		}
		if trace != nil && trace.DNSDone != nil {
			addrs, _ := r.Val.([]IPAddr)
			trace.DNSDone(ipAddrsEface(addrs), r.Shared, err)
		}
		return lookupIPReturn(r.Val, err, r.Shared)
	}

```

Letting the control flow continues will lead to [l.Debug("resolved peers to following addresses", "peers", strings.Join(resolvedPeers, ","))](https://github.com/prometheus/alertmanager/blob/main/cluster/cluster.go#L173) that will not block the app to log as well as registering liveness probes(`/healthy`, `/ready`).


eyazici90:
Hey @grobinson-grafana 👋 
Could u take a quick glance into that, lmk what u think? Based on ur response, we may push it further or not.

cluster/cluster.go
```diff
@@ -166,7 +169,9 @@ func Create(
 		}
 	}
 
-	resolvedPeers, err := resolvePeers(context.Background(), knownPeers, advertiseAddr, &net.Resolver{}, waitIfEmpty)
+	ctx, cancel := context.WithTimeout(context.Background(), resolveTimeout)
+	defer cancel()
+	resolvedPeers, err := resolvePeers(ctx, knownPeers, advertiseAddr, &net.Resolver{}, waitIfEmpty)
```
grobinson-grafana, line 174: OK, I think I understand. If this timesout, Alertmanager will still fail to start because the peer cannot be resolved, but at least this time it emits an error message?

eyazici90: I don't think AM will fail to start. See https://github.com/prometheus/alertmanager/blob/main/cluster/cluster.go#L737-L740. It just takes the direct addr & moves on with the loop. 

```go
// inlining
ips, err := res.LookupIPAddr(ctx, host)
		if err != nil {
			// Assume direct address.
			resolvedPeers = append(resolvedPeers, peer)
			continue
		}

```

Afterwards we will see debug logs https://github.com/prometheus/alertmanager/blob/main/cluster/cluster.go#L173 resolved addr for peers.

```go
	l.Debug("resolved peers to following addresses", "peers", strings.Join(resolvedPeers, ","))

```

Far better than hanging app with no logs emitted, imho 😓

grobinson-grafana: So how does Alertmanager behave if it continues without having resolved peers? I haven't had time to look into it I'm afraid

eyazici90: Seems like there is [a test for initially failed peers](https://github.com/prometheus/alertmanager/blob/main/cluster/cluster_test.go#L229). I reckon, am(alertmngr) tries to reconnect for certain amount of time and eventually drops failed peers. Nevertheless, health probes and rest would work same. Only gossip settlings wont. That is my understanding at least :D

eyazici90:
Could u help me with review @SuperQ ?

cluster/cluster.go
Spaceman1701: The [refresh](https://github.com/prometheus/alertmanager/blob/main/cluster/cluster.go#L445) method is called by with [runPeriodicTask](https://github.com/prometheus/alertmanager/blob/main/cluster/cluster.go#L288-L291), so if alertmanager fails to resolve peers on start, it'll try again after [DefaultRefreshInterval](https://github.com/prometheus/alertmanager/blob/main/cluster/cluster.go#L127).

And I agree that alertmanager won't fail to start/crash - failing to resolve peers [only writes a debug message:](https://github.com/prometheus/alertmanager/blob/main/cluster/cluster.go#L451). Furthermore, [failing to join the cluster is not treated as a fatal error at startup](https://github.com/prometheus/alertmanager/blob/main/cmd/alertmanager/main.go#L324-L340).

Spaceman1701: > And I agree that alertmanager won't fail to start/crash - failing to resolve peers [only writes a debug message:](https://github.com/prometheus/alertmanager/blob/main/cluster/cluster.go#L451). Furthermore, [failing to join the cluster is not treated as a fatal error at startup](https://github.com/prometheus/alertmanager/blob/main/cmd/alertmanager/main.go#L324-L340).

Ah sorry, this is the wrong reasoning since `cluster.Create` calls `resolvePeers` and errors from that call are treated as fatal. However, as @eyazici90 already noted, resolution errors aren't bubbled up out of `resolvePeers` unless the initial lookup returns [no results and no error ](https://github.com/prometheus/alertmanager/blob/main/cluster/cluster.go#L746) and then future lookups raise an error

Spaceman1701 (approved):
Did a more detailed read through the code and didn't find that concerns me. 

I think this makes a lot of sense, and it's very bad behavior to hang on `resolvePeers` right now. 

I am a little nervous about this change since it essentially adds a new exit condition, but I think 15 seconds is more than long enough to do a few DNS resolves. I guess if other people are also concerned, we could bump it to 30s by default, but that seems extremely long.

SuperQ:
Looks like a couple tests need fixing.

eyazici90:
> I guess if other people are also concerned, we could bump it to 30s by default, but that seems extremely long.

I am fine to change it as `default: 30s`, however, imo 15s is already enough as default.

eyazici90:
> Looks like a couple tests need fixing.

Fixed @SuperQ

SuperQ (approved):

PR #4351, opened 2025-04-11
inhibit: compare value of different source and target labels

### LINKED ISSUE #2254: Inhibit Rules should be able to consider different labels in their equal statement
While the title basically says it all, I will try to back this up using a concrete example. Imagine a setup where in addition to all routers being a target for some type of `blackbox_exporter`-style metrics an additional source of data is used to generate JSON target lists for `file_sd`. 

In my example, this additional source could be a routers configuration backup, which gives us a definitive list of link addresses configured on any interface of any router, as well as any configured meta data for each interface. It is trivial to build a JSON file containing the targets (all link addresses configured locally) with the relevant labels:

 * the router this adress is configured on
 * the interface name/description
 * the remote hostname, for instance parsed from the description

These metrics could end up looking like this (`10.0.0.0/8` are link adresses, `192.168.0.0/16` are loopbacks):

```
# job: ping-router-loopback
probe_success{instance="192.168.0.1", hostname="r1"} 0
probe_success{instance="192.168.0.2", hostname="r2"} 1

# job: ping-router-interface
probe_success{instance="10.0.0.1", hostname="r2", interface="Te0/7/0/12", remote="r1"} 0
probe_success{instance="10.0.0.2", hostname="r1", interface="Te0/2/0/1", remote="r2"} 0
```

Alerting in the most obvious way would create alerts similar to the job names, for instance a `RouterDown` alert with the expression `probe_success{job="ping-router-loopback"} == 0`. I would obviously want the following inhibition rule:

```
- source_match:
    alertname: RouterDown
  target_match:
    alertname: InterfaceDown
  equal:
  - hostname
```

This would inhibit the alert informing me that an interface is down on a router which is already being alerted as Down itself. I would however like to go one step further using a inhibit rule such as the following one, as it does not come as a surprise that any interface adjacent to the downed router will go down in turn, even though it is on another router/in another region/whatever.

```
# option A
- source_match:
    alertname: RouterDown
  target_match:
    alertname: InterfaceDown
  equal:
  - source_label: hostname
    target_label: remote

# option B, which would make the original `equal` kind of unnecessary
# by using `hostname: $hostname` for instance
- source_match:
    alertname: RouterDown
  target_match:
    alertname: InterfaceDown
    remote: "$hostname"
```

While the proposed syntax variants are just a general idea, I feel that this should be possible in some way which does not involve hacking around with the underlying alerts expressions.

monkz:
Adds equal_pairs config option to inhibit rules, which allows to compare the value of different source and target labels.

Closes https://github.com/prometheus/alertmanager/issues/2254

Credit [@swoga](https://github.com/swoga) https://github.com/prometheus/alertmanager/pull/3525

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -966,9 +966,11 @@ type InhibitRule struct {
 	TargetMatchRE MatchRegexps `yaml:"target_match_re,omitempty" json:"target_match_re,omitempty"`
 	// TargetMatchers defines a set of label matchers that have to be fulfilled for target alerts.
 	TargetMatchers Matchers `yaml:"target_matchers,omitempty" json:"target_matchers,omitempty"`
-	// A set of labels that must be equal between the source and target alert
+	// A set of labels that must be equal between the source and target alert.
 	// for them to be a match.
 	Equal []string `yaml:"equal,omitempty" json:"equal,omitempty"`
+	// EqualPairs defines a set of pairs that have to be equal in the source and target alert.
+	EqualPairs []LabelPair `yaml:"equal_pairs,omitempty" json:"equal_pairs,omitempty"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface for InhibitRule.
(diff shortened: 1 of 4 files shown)


ddarby-lewis-oxe:
What would be required to get this merged? I could make use of it in our alerting.

Spaceman1701:
Hi! This is an interesting change. I can definitely see the value. It's also a pretty significant behavior change, so we'll want to make sure it's something we want to support in the long term.

However, I think the implementation is going to need to change pretty significantly because of the new inhibitor implementation on `main` (see #4607). The new implementation keeps an index on the fingerprint of `equal` labels. I think that the indexing scheme would need to change if in order to support this behavior.

Would you be interested in trying to rebase this onto the latest `main`?

Malte-Wagner:
This is not forgotten :)

PR #4353, opened 2025-04-15
Fix error handling in template rendering for Telegram

m-burst:
`notify.TmplText` and `notify.TmplHTML` use an unusual error handling pattern: the error is not explicitly returned but still must be checked by callers.

In the Telegram notifier the check is missing. Because of this, if template rendering fails, Alertmanager still attempts to send an empty message to Telegram and logs an error with text `Bad Request: message text is empty (400)`.

With this patch applied, Alertmanager will log the actual error that happens during template rendering.

### DIFF
--- a/notify/telegram/telegram.go
+++ b/notify/telegram/telegram.go
@@ -80,6 +80,9 @@ func (n *Notifier) Notify(ctx context.Context, alert ...*types.Alert) (bool, err
 	}
 
 	messageText, truncated := notify.TruncateInRunes(tmpl(n.conf.Message), maxMessageLenRunes)
+	if err != nil {
+		return true, err
+	}
 	if truncated {
 		n.logger.Warn("Truncated message", "alert", key, "max_runes", maxMessageLenRunes)
 	}


notify/telegram/telegram.go
```diff
@@ -80,6 +80,9 @@ func (n *Notifier) Notify(ctx context.Context, alert ...*types.Alert) (bool, err
 	}
 
 	messageText, truncated := notify.TruncateInRunes(tmpl(n.conf.Message), maxMessageLenRunes)
+	if err != nil {
+		return true, err
```
grobinson-grafana, line 84: I think this should be `false`, no? The message cannot be retried as it will fail again for the same reason?

m-burst: It will most probably fail again indeed.

However, I see that below in the code `true` is returned if `getBotToken()` returns an error. As far as I understand, `getBotToken()` will not probably recover on its own (only if someone reloads the config or changes the `bot_token_file` on the filesystem), similarly to the template rendering. This is why I chose `true` in this case, but I can change this to `false` if fits better here.

grobinson-grafana: I think returning `true` for `getBotToken()` is also a bug 👍

m-burst: Ok, changed this case to `false`

grobinson-grafana (approved):

PR #4354, opened 2025-04-17
feat: add timeout option for pagerduty notifier 

eyazici90:
# Context

This is to manager timeout for the context that is passed to pagerduty calls. Group level context is set to wider intervals such as 5mins, whereas we would like to set lower level threasholds for pagerduty integrations to have limit on retries and intermediate blockers in the network.


rajagopalanand (approved):
LGTM

eyazici90:
@grobinson-grafana  Could u help me with the review? 🙏

eyazici90:
Could u help me with review @SuperQ ?

PR #4356, opened 2025-04-18
feat(amtool): allow overriding config with AMTOOL_CONFIG_FILE env var

dronenb:
This PR adds the ability to override the `amtool` config file with an environment variable

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,7 +305,11 @@ amtool template render --template.glob='/foo/bar/*.tmpl' --template.text='{{ tem
 
 ### Configuration
 
-`amtool` allows a configuration file to specify some options for convenience. The default configuration file paths are `$HOME/.config/amtool/config.yml` or `/etc/amtool/config.yml`
+`amtool` allows a configuration file to specify some options for convenience. The default configuration file paths are, in order of precedence:
+
+1. The environment variable `AMTOOL_CONFIG_FILE`, if set.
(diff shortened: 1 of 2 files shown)


TheMeier (commented):
Well it probably works and I understand the idea. I just wanted to point out that currently using environment variables is not a  very common thing in the code base. See https://github.com/search?q=repo%3Aprometheus%2Falertmanager%20os.GetEnv&type=code 
2 times in test code, 1 for ZONEINFO, 1 for DEBUG in altrermanager
Also why not add it as cli flag? Or why not both?

Overall I think adding configurability here is a good idea.

dronenb:
The decision to make it an environment variable was intentional - similar to `KUBECONFIG` for `kubectl`. In my case, I work with many different AlertManager clusters - as such, it would be useful for me to be able to target different alertmanager instances from different shell environments (although I agree that adding the option to specify the config via a flag is also a good idea).

siavashs:
I think it would be better to add `--config.file=` flag to `amtool` and then one could use the flag directly or shell `alias`es to swap configs more easily.

SoloJacobs:
Hi @dronenb ,

thank you for putting the time into this PR. Based on the review comments here, we will not accept your PR.

Feel free to open a PR for adding `--config.file`, or an issue such that somebody else may pick up this feature.

Kind regards

PR #4362, opened 2025-04-22
Pushover: Add monospace message formatting, add documentation for existing HTML styling

bl1nk:
Hey all,

this is my first contibution to alertmanager. Please let me know if I missed something.

There are two changes in this PR, but hey seemed very related:

- Document the (already existing) feature of enabling HTML message formatting when using Pushover
- Add the possibility to use monospace formatting for messages in Pushover (+documentation, +test)

API documentation can be found [here](https://pushover.net/api#html).

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -741,7 +741,8 @@ type PushoverConfig struct {
 	Retry       duration `yaml:"retry,omitempty" json:"retry,omitempty"`
 	Expire      duration `yaml:"expire,omitempty" json:"expire,omitempty"`
 	TTL         duration `yaml:"ttl,omitempty" json:"ttl,omitempty"`
-	HTML        bool     `yaml:"html" json:"html,omitempty"`
+	HTML        bool     `yaml:"html,omitempty" json:"html,omitempty"`
+	Monospace   bool     `yaml:"monospace,omitempty" json:"monospace,omitempty"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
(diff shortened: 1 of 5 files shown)


config/notifiers.go
```diff
@@ -742,6 +742,7 @@ type PushoverConfig struct {
 	Expire      duration `yaml:"expire,omitempty" json:"expire,omitempty"`
 	TTL         duration `yaml:"ttl,omitempty" json:"ttl,omitempty"`
 	HTML        bool     `yaml:"html" json:"html,omitempty"`
+	Monospace   bool     `yaml:"monospace" json:"monospace,omitempty"`
```
grobinson-grafana, line 745: It looks like we are missing `omitempty` on the `yaml` tag.

config/notifiers.go
```diff
@@ -763,6 +764,9 @@ func (c *PushoverConfig) UnmarshalYAML(unmarshal func(interface{}) error) error
 	if c.Token != "" && c.TokenFile != "" {
 		return errors.New("at most one of token & token_file must be configured")
 	}
+	if c.HTML && c.Monospace {
+		return errors.New("monospace may not be used if html is used, and vice versa")
```
grobinson-grafana, line 768: Keeps it consistent will standard error message text:

```suggestion
		return errors.New("at most one of monospace & html must be configured")
```

grobinson-grafana (approved):

grobinson-grafana (changes requested):
A couple small changes, otherwise LGTM! 👍

config/notifiers.go
bl1nk: Should the HTML yaml tag have an `omitempty` as well? It currently does not have it, so I kept this in line with the other bool.

grobinson-grafana (approved):

PR #4364, opened 2025-04-23
feat: configurable provider channel buffer length

siavashs:
This change makes the memory provider channel buffer length configurable.
The previously hard-coded value of 200 is used as default.
This should result in better performance and less timeouts during sudden alert spikes, specially if the pipeline speed is affected by slow receivers.


grobinson-grafana:
Discussed [here](https://github.com/prometheus/alertmanager/issues/4392). I'm not against making this configurable, but I'm concerned that we don't understand enough about how and why this channel is becoming full and blocking. I've asked if we can get some profiles to better understand this.

siavashs:
Closing this, see the comment is linked issue.

PR #4368, opened 2025-04-25
chore: bump golangci-lint to v2.3.0

mmorel-35:
#### Description

Bump golangci-lint to v2.3.0 and fixes newly raised issues


matcher/compat/parse.go
```diff
@@ -138,7 +138,7 @@ func FallbackMatcherParser(l *slog.Logger) ParseMatcher {
 		}
 		// If the input is valid in both parsers, but produces different results,
 		// then there is disagreement.
-		if nErr == nil && cErr == nil && !reflect.DeepEqual(nMatcher, cMatcher) {
+		if cErr == nil && !reflect.DeepEqual(nMatcher, cMatcher) {
```
kakkoyun, line 141: What was the specific linter error here?

.golangci.yml
```diff
@@ -1,99 +1,96 @@
...
+    godot:
+      scope: toplevel
+      exclude:
+        - "^ ?This file is safe to edit"
+        - "^ ?scheme value"
+      period: true
+      capital: true
+    revive:
+      rules:
+        - name: blank-imports
```
sysadmind, line 44: This is somewhat different than what the prometheus project uses. I think it would be good to try to standardize.

.golangci.yml
```diff
@@ -1,99 +1,96 @@
...
-      - error-nil
-      - expected-actual
-      - len
-      - require-error
-      - suite-dont-use-pkg
-      - suite-extra-assert-call
+formatters:
+  enable:
+    - gofumpt
+    - goimports
```
sysadmind, line 90: Prometheus moved to gci for imports

sysadmind (approved):

.golangci.yml
mmorel-35: I can provide this in a following PR.
I suggest to narrow this one to the upgrade to v2

PR #4370, opened 2025-04-26
fix(jira): correctly handle issues with null resolution (Fixes #4295)

### LINKED ISSUE #4295: Native Jira Integration: JQL Query Excludes Unresolved Issues When Using wont_fix_resolution
# JQL Query Excludes Unresolved Issues When Using wont_fix_resolution

## Description

When defining a wont_fix_resolution, the JQL query used in Alertmanager 0.28 (as seen in the debug output) is:

`jql="resolution != \"Won't Do\" and (resolutiondate is EMPTY OR resolutiondate >= -1440m) and project=\"TEST\" and labels=\"ALERT{........}\" order by status ASC,resolutiondate DESC"`

However, this query does not return issues that have no resolution set. This is due to the behavior of JIRA's != operator, which excludes issues where the resolution field is empty (i.e., unresolved).

## Reference

See JIRA documentation for details on this behavior: https://confluence.atlassian.com/jirakb/using-not-equals-on-a-resolution-does-not-return-unresolved-issues-635897091.html

## Expected Behavior

The query should return both unresolved issues and issues with resolutions that do not match "Won't Do".
Possible Fix

To correctly include unresolved issues, the JQL should be modified to explicitly account for them, such as:

`jql="(resolution is EMPTY OR resolution != \"Won't Do\") and (resolutiondate is EMPTY OR resolutiondate >= -1440m) and project=\"TEST\" and labels=\"ALERT{........}\" order by status ASC,resolutiondate DESC"`

## Steps to Reproduce

- Define a wont_fix_resolution with a JQL query similar to the one above.
- Ensure there are unresolved issues in the project.
- Run the query and observe that unresolved issues are missing from the results.

## Environment

- Alertmanager version: 0.28

surinkim:
This pull request fixes a bug where Alertmanager would create a duplicate Jira issue if the existing issue's resolution field was `null` (for example, when the issue was still "To Do" or "In Progress").

Previously, the search query only checked for `resolution != "Won't Do"`, so issues with `null` resolution were not properly matched, causing Alertmanager to incorrectly create a new issue.

This fix updates the JQL condition to also check `resolution is EMPTY or resolution != "Won't Do"`, ensuring that open issues are correctly found and reused.

## Changes
- Modified the Jira JQL query condition to correctly handle issues with `null` resolution.
- Added corresponding unit tests to verify this behavior.

### Before
- Search JQL: `resolution != "Won't Do"` only
- Behavior: **create new issue** even if there is an open issue

Before Fix

### After
- Search JQL: `resolution is EMPTY or resolution != "Won't Do"`
- Behavior: **update existing issue** properly without creating a duplicate

After Fix

---

_Fixes #4295._


cicadadevops:
Could you please share the Jira integration and Alertmanager configuration?

surinkim:
Here is the Jira integration and Alertmanager configuration I used for testing:

---

**Jira Environment**  
- Instance: [https://surinkim10.atlassian.net](https://surinkim10.atlassian.net)  
- Note: This is a temporary trial Jira Cloud project created for testing purposes.  
  (External access is restricted.)
 
 ---
  **Alertmanager Configuration**  
- Configuration details are shared here:  
  👉 [Alertmanager Jira Test Configuration (Gist)](https://gist.github.com/surinkim/85fe961f65d4ec6fa34e2c924eb9f5d1)

cicadadevops:
Has anyone tried integrating Alertmanager with Jira Data Center?
I’m using Jira Data Center version 10.3.6, but unfortunately, I haven’t been able to get it to work correctly.
Sometimes it fails to reopen existing tickets, and sometimes it doesn’t create new ones at all.

PR #4393, opened 2025-05-06
feat(rocketchat): support plain text attribute

Patrick-Remy:
RocketChat doesn't set the attachment title as notification title. Instead with the current implementation of this notifier, you receive notifications like `"bot: "`. This is really unuseful.

The `PostMessage.Text` attribute was already typed and can be additionally set to set a "plain text".
I also fixed the previously 404 rocketchat developer documentation which has moved in the meantime and removed the unused (and undocumented) property`ParseUrls`.


Patrick-Remy:
Maybe @TheMeier as the original contributor for the rocketchat integration can review my contribution.
Unfortunately I have no idea why tests are failing, but as disclaimer I am not a Go developer ;)

TheMeier:
I just tested it and then remembered, if you send a `Text` field (even if empty) The attachments will be collapsed and truncated. I cant even un-collaps them:

image

TheMeier:
@Patrick-Remy are you still interested in working on this?

Issue #416, opened 2016-06-29
[HA AM] Retention set by writer


fabxc:
Per in-person discussion, having the retention/deletion time for state items shared via gossip directly attached to the object, and defined by the creating/modifying peer, will avoid inconsistencies.

fabxc:
Fixed in HA branch for notification information now.

simonpasquier:
I think this one can be closed. Gossiped nflogs and silence items contain their expiration date/time.

https://github.com/prometheus/alertmanager/blob/8c9e0cf50cba43476c2ee16d52d44f37df064425/nflog/nflogpb/nflog.proto#L44-L52

https://github.com/prometheus/alertmanager/blob/8c9e0cf50cba43476c2ee16d52d44f37df064425/silence/silencepb/silence.proto#L61-L66

PR #4425, opened 2025-06-04
Add documentation for custom headers

laserscout:
The option to configure custom http headers via the `http_headers` configuration option in `http_config` was added on [prometheus/common #416](https://github.com/prometheus/common/pull/416].

This was documented in prometheus [prometheus/prometheus #14660](https://github.com/prometheus/prometheus/pull/14660] but not in alertmanager.

I have migrated over the documentation after verifying that the `http_headers` configuration option indeed works starting with alertmanager 0.28.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -820,6 +820,18 @@ oauth2:
 # Configures the TLS settings.
 tls_config:
   [ <tls_config> ]
+
+# Custom HTTP headers to be sent along with each request.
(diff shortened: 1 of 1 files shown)


grobinson-grafana:
Hi! 👋 Could you please sign off your commit so the DCO check passes?

laserscout:
Sorry for the delay @grobinson-grafana 

I've signed off the commit and rebased onto main.

PR #4430, opened 2025-06-11
Use model.Duration for timeout option in webhook notifier

alexander-akhmetov:
Changed webhook timeout field from `time.Duration` to `model.Duration` to make JSON marshalling consistent with YAML.

`time.Duration` marshals to JSON as an integer (nanoseconds), while to YAML (with `gopkg.in/yaml.v3`) it's marshalled as a duration string. 

`time.Duration` example: https://go.dev/play/p/Ml2Mb_VFNyu:

```
JSON: {"timeout":90000000000}
YAML: timeout: 1m30s
```

`model.Duration` also marshals to human-readable duration strings in both JSON and YAML, and is consistent with other duration fields in Alertmanager.

`model.Duration` example: https://go.dev/play/p/rqts9yDSEQL

```
JSON: {"timeout":"1m30s"}
YAML: timeout: 1m30s
```

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -538,7 +538,7 @@ type WebhookConfig struct {
 
 	// Timeout is the maximum time allowed to invoke the webhook. Setting this to 0
 	// does not impose a timeout.
-	Timeout time.Duration `yaml:"timeout" json:"timeout"`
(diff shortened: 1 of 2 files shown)


alexander-akhmetov:
Closing this for now

Issue #4402, opened 2025-05-22
Alert marker fills up with deleted alerts' states


siavashs:
[Alert marker](https://github.com/prometheus/alertmanager/blob/0ce3cfb962db3cbb1649d3e816a49a13c4036cd1/types/types.go#L65) is shared between multiple components, and this results in the marker to fill up with states of alerts which are already deleted.

This probably happens in multiple scenarios but one confirmed case is:
1. Alert is marked `active`
2. Alert is dispatched
3. Alert is `resolved`
    - Alert is deleted from memory provider(s)
    - Alert is deleted from marker
4. Notifier fails to send the notification dispatched form step 2
     - Notifier retries
     - Alert is added with `active` status to the marker again and will never be deleted

This is visible on `alertmanager_marked_alerts` metric, the marker holds hundreds of thousands of states until Alertmanager is restarted.

<img width="1078" alt="Image" src="https://github.com/user-attachments/assets/90168751-9fdd-4653-a0b5-36c0cf41ba6c" />

<img width="1078" alt="Image" src="https://github.com/user-attachments/assets/d7510269-a6ac-475f-9f5b-2cee36129715" />

PR #4437, opened 2025-06-23
fix(inhibit): drop invalid inhibit marker

siavashs:
We must not mark active alerts as inhibited.
This should fix one of marker state leakage mentioned in #4402


siavashs:
Actually there is logic in types package to use this for marking alerts as active.

PR #4438, opened 2025-06-23
fix(marker): stop state leakage from aggregation groups

siavashs:
This change makes aggregation groups to delete resolved alerts from marker, therefore avoiding the leakage of ghost states mentioned in #4402.

Fixes: https://github.com/prometheus/alertmanager/issues/4402

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -340,7 +340,7 @@ func (d *Dispatcher) processAlert(alert *types.Alert, route *Route) {
 		return
 	}
 
-	ag = newAggrGroup(d.ctx, groupLabels, route, d.timeout, d.logger)
+	ag = newAggrGroup(d.ctx, groupLabels, route, d.timeout, d.marker.(types.AlertMarker), d.logger)
 	routeGroups[fp] = ag
 	d.aggrGroupsNum++
 	d.metrics.aggrGroups.Inc()
@@ -389,6 +389,7 @@ type aggrGroup struct {
 	routeKey string
 
 	alerts  *store.Alerts
+	marker  types.AlertMarker
 	ctx     context.Context
 	cancel  func()
 	done    chan struct{}
@@ -400,7 +401,7 @@ type aggrGroup struct {
 }
 
 // newAggrGroup returns a new aggregation group.
-func newAggrGroup(ctx context.Context, labels model.LabelSet, r *Route, to func(time.Duration) time.Duration, logger *slog.Logger) *aggrGroup {
+func newAggrGroup(ctx context.Context, labels model.LabelSet, r *Route, to func(time.Duration) time.Duration, marker types.AlertMarker, logger *slog.Logger) *aggrGroup {
 	if to == nil {
 		to = func(d time.Duration) time.Duration { return d }
 	}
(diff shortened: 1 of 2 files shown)


types/types.go
```diff
@@ -281,6 +287,20 @@ func (m *MemMarker) Delete(alert model.Fingerprint) {
 	delete(m.alerts, alert)
 }
 
+// GC implements AlertMarker.
+func (m *MemMarker) GC() {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+
+	// Expire states after 15 minutes.
+	expiry := time.Now().Add(time.Minute * -15)
```
OGKevin: Instead of hardcoding 15, is there a way to use the default time that AM uses when you don't specify endTime when sending an alert? 

Or make it configurable, since repeat interval is also configurable. What if prometheus repeats the same alert after 15m? Could this create a scenario where the marker no longer has the alert, but the Alerts struct still has it?

siavashs: > Instead of hardcoding 15, is there a way to use the default time that AM uses when you don't specify endTime when sending an alert?

Alertmanager needs to hold the state longer, because of the group wait, notification retries, etc. 15m is just a "safe" value.

> Or make it configurable, since repeat interval is also configurable. 

we can use the same value used for alert GC which is configurable.

> What if prometheus repeats the same alert after 15m?

that's fine, state is recreated and then cleared again after timeout.

siavashs: Updated the code to use `2 * alert GC interval` as a safe expiry value.

siavashs:
> Looking at the test changes, it doesn't seem like this exercises the problem behavior. Would you mind including some additional testing to make sure we don't break in the future?

Done, added 3 scenarios. The code is change to make sure that we don't accidentally delete a marker due to a race condition. Please have a look and let me know if we should add more.

But while adding these I noticed the whole Marker setup to be a week point in terms of consistency, we need to come up with a better solution at some point.

SuperQ (approved):
Nice, thanks!

dispatch/dispatch.go
```diff
@@ -539,6 +541,14 @@ func (ag *aggrGroup) flush(notify func(...*types.Alert) bool) {
 		// we would delete an active alert thinking it was resolved.
 		if err := ag.alerts.DeleteIfNotModified(resolvedSlice); err != nil {
 			ag.logger.Error("error on delete alerts", "err", err)
+		} else {
+			// Delete markers for resolved alerts that are not in the store.
+			for _, alert := range resolvedSlice {
```
grobinson-grafana, line 546: Just heads up this is a race condition for the same reason we had to implement `DeleteIfNotModified`. There is a missing check to make sure the deleted alert is the same alert returned from `ag.alerts.Get(alert.Fingerprint())`.

Without this check what can happen is we delete the alert in between `DeletedIfNotModified` and then before we call `Get` a new alert is received. What happens then is we delete the marker of the new alert by mistake.

siavashs: As we discussed on Slack, this is safe since even if we delete the marker the alert status would then be `unprocessed`: https://github.com/prometheus/alertmanager/blob/52eb1fc4aa84d24fab9cb61cc4af57aeb773b6f9/types/types.go#L261-L274

PR #4466, opened 2025-07-06
[4465] Fix CVE-2025-22871: Update Go version to 1.24.4

a4abhishek:
- Update go.mod: go directive to 1.23.10, toolchain to go1.24.4
- Update CircleCI go/install version to 1.24.4
- Docker images already use golang-builder:1.24-base (Go 1.24.4)

This addresses CVE-2025-22871 affecting Go 1.23.7 and below, and Go 1.24.1 and below, by ensuring all builds use Go 1.24.4 which contains the security fix for improper handling of bare LF in chunked data chunk-size lines in net/http package.

Fixes: CVE-2025-22871

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ jobs:
           command: sudo rm -rf /usr/local/go
       # Whenever the Go version is updated here, .promu.yml should also be updated.
       - go/install:
-          version: "1.24.1"
+          version: "1.24.4"
       - run:
           name: Remove generated code
           command: make clean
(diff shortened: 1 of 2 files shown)


jsto:
following up on this. Also, when is the new tag containing this fix will be released?

SuperQ:
Updating go.mod does not address CVEs.

Please do not report raw vulnerability scanner results. They are prone to false positives and cause the Prometheus team toil in verifying. Please verify vulnerability reports and include specific details as to which components are directly exploitable.

PR #4467, opened 2025-07-08
chore: bump go to 1.24.4 and update dependencies

slashpai:


cli/root.go
```diff
@@ -52,7 +52,7 @@ var (
 func initMatchersCompat(_ *kingpin.ParseContext) error {
 	promslogConfig := &promslog.Config{Writer: os.Stdout}
 	if verbose {
-		promslogConfig.Level = &promslog.AllowedLevel{}
+		promslogConfig.Level = promslog.NewLevel()
```
slashpai, line 55: Due to breaking change in https://github.com/prometheus/common/pull/754

slashpai:
will update the tests with utf-8 updates as latest prometheus 3 default is utf-8 .Some testcases are failing which checks for utf-8 as invalid labels

SuperQ:
Updated in #4589

PR #4468, opened 2025-07-08
fix(ui usability): Quote matcher values when viewing silences.

rudo-thomas:
Problem: When viewing a silence in the UI (in the list or individually) its matchers are rendered without quotes around the value, which makes it impossible to differentiate between matchers like `a="~b"` and `a=~"b"`. That makes silences difficult to debug using the web UI.

Fix: Quote the value of the matcher, just as it is quoted when editing/adding matchers in a silence.


grobinson-grafana:
I think this is might not be enough, you need to also escape quotes inside the quotes? Otherwise you will have something like:

```
foo=""bar"
```

instead of:

```
foo="\"bar"
```

rudo-thomas:
I tried that before sending this PR, and the escaping seemed to be happening already.

I'll try to find the code where it happens (in elm), or at least I'll post a screenshot. Tomorrow :)

rudo-thomas:
Update: You are right, the code I changed does indeed need to escape quotes (and other special characters).

My earlier attempts appeared to work, but it turns out there is a bug in the web UI: when a matcher like `a="b\"c"` is added in a silence, you'll get `a="b\\\"c"` (i.e. the backslash makes it into the matcher). This is visible via the API.

Matchers created via the API work as expected and uncover a bug in this PR.

TL;DR this needs a fix in multiple places. I'm closing this PR.

Issue #4436, opened 2025-06-23
Make /alertmanager group-writable in dockerfile


calliecameron:
https://github.com/prometheus/prometheus/pull/16073 made /prometheus group-writable, to support running prometheus in docker as an arbitrary user. The same change should be made in alertmanager's dockerfile.

PR #4469, opened 2025-07-10
Make /alertmanager group-writable in Dockerfile

VaibhavOza1997:
This change makes the `/alertmanager` directory group-writable in the Dockerfile.

It follows the precedent set in [prometheus/prometheus#16073](https://github.com/prometheus/prometheus/pull/16073) to support running the Alertmanager container as an arbitrary user. This aligns with best practices for container security and facilitates use cases where the container is run without root privileges.

Fixes: #4436


VaibhavOza1997:
Hi maintainers 👋,

This is my first contribution to the project. This PR makes the `/alertmanager` directory group-writable in the Dockerfile, following the precedent from prometheus/prometheus#16073. This change allows running Alertmanager containers as arbitrary non-root users.

All feedback is very welcome. Thank you for your time and for maintaining this awesome project!

Best regards,  
Vaibhav Oza

TheMeier:
hmm usually you use mounted volumes for that also best practice is to run production workload with read-only for the container fs. So doesn't seem to make much sense to me

VaibhavOza1997:
Thank you for the quick feedback!

That makes sense. I see your point about using mounted volumes and keeping the container filesystem read-only in production. The motivation for this change came from issue #4436, where making `/alertmanager` group-writable was suggested to support running the container as an arbitrary user without needing to adjust volume permissions externally. I’m happy to close this PR if you think this approach doesn’t align with the project’s best practices. Just wanted to try to help based on the open issue. 😊

PR #4473, opened 2025-07-13
add Saturday to the First day of the week options

haghighi:


haghighi:
I think it's ready

cc @sysadmind

PR #4477, opened 2025-07-15
CI: add maildev tests back to GA

sysadmind:
I think this is the only test missing from github actions. Adding this should allow us to remove the flaky circleci workflow.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,11 @@
 
 version: 2.1
 
+executors:
+  golang:
+    docker:
+      - image: busybox
+
 jobs:
   noopjob:
     executor: golang


SuperQ (approved):
Can you update the test comment at the top of `notify/email/email_test.go` to match the new docker image location/version?

SuperQ:
I think we should update a couple things about the test setup.
* Mention maildev in the README.md
* Add a go testing tag/env so that maildev can be explicitly enabled for testing.

But this is fine for now.

SuperQ:
Should we also remove the CircleCI test pipeline in this PR?

sysadmind:
I think your comments are good. I am not sure I want to keep maildev, it seems to have some limitations. I did pull CircleCI and update the comments in the test file.

SuperQ:
Nice, thanks!

Issue #4481, opened 2025-07-30
Migrate SNS notifier code to a new AWS SDK


pehlicd:
Currently, sns notifier depends on [aws-sdk-go](https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.55.6#section-readme), which is deprecated. It will not be supported after 31.07.2025. Users are advised to migrate to [AWS SDK for Go v2](https://aws.github.io/aws-sdk-go-v2/docs/).

There is already issue for it in prometheus about that topic, see: https://github.com/prometheus/prometheus/issues/15572

PR #4482, opened 2025-07-30
chore(deps): migrate aws sdk to v2

pehlicd:
Closes: #4481


notify/sns/sns.go
```diff
@@ -22,18 +22,22 @@ import (
...
+	"github.com/aws/smithy-go"
+	smithyhttp "github.com/aws/smithy-go/transport/http"
+
 	commoncfg "github.com/prometheus/common/config"
 
 	"github.com/prometheus/alertmanager/config"
 	"github.com/prometheus/alertmanager/notify"
 	"github.com/prometheus/alertmanager/template"
-	"github.com/prometheus/alertmanager/types"
+	amtypes "github.com/prometheus/alertmanager/types"
```
sysadmind: I think it makes more sense to leave this package as `types` and alias the SNS package to something like `snsTypes`

notify/sns/sns.go
```diff
@@ -81,69 +93,85 @@ func (n *Notifier) Notify(ctx context.Context, alert ...*types.Alert) (bool, err
...
-				Config: aws.Config{
-					Region:      aws.String(n.conf.Sigv4.Region),
-					Credentials: creds,
-				},
-				Profile: n.conf.Sigv4.Profile,
-			})
-			if err != nil {
-				return nil, err
-			}
+		stsCfg, err := awsconfig.LoadDefaultConfig(ctx, baseCfgOpts...)
```
sysadmind, line 144: This is needed later so it might make sense to call this once and only swap the credentials in this branch.

notify/sns/sns.go
```diff
@@ -212,24 +241,42 @@ func (n *Notifier) createPublishInput(ctx context.Context, tmpl func(string) str
...
 
-func validateAndTruncateMessage(message string, maxMessageSizeInBytes int) (string, bool, error) {
+func validateAndTruncateMessage(message string, maxMessageSize int) (string, bool, error) {
 	if !utf8.ValidString(message) {
 		return "", false, fmt.Errorf("non utf8 encoded message string")
 	}
-	if len(message) <= maxMessageSizeInBytes {
+
+	// Check if truncation is needed based on the appropriate metric (bytes or runes).
+	isSMS := maxMessageSize == 1600
```
sysadmind: I think this slightly changes the behavior right? The previous code didn't check for this. I don't think it's a bad change, but it might be better to make this a separate change.

notify/sns/sns.go
pehlicd: The two calls are required because they configure two different clients:

1. The first LoadDefaultConfig is used to build the STS client. Its specific purpose is to create the AssumeRole provider which will fetch the temporary credentials.

2. The second LoadDefaultConfig builds the final configuration for the SNS client. This step is essential to combine the base options with both the new role provider and the custom apiURL.

We can't just modify the first config (stsCfg) because it's targeted at the STS service and doesn't include the SNS-specific endpoint. This two-step process ensures all options are correctly applied for the final client. Please see line 141,142.

notify/sns/sns.go
pehlicd: definitely it makes sense will update it.

notify/sns/sns.go
pehlicd: Thank you for the feedback. You're right, it's better to separate the new behavior from this one. I will completely revert this function to main one.

pehlicd:
Hey @sysadmind , can I get another review 🙂

sysadmind:
@pehlicd are you a user of the SNS notifications? I don't have an AWS account to test these changes with and I want to make sure we have tested against the real AWS API if at all possible.

pehlicd:
@sysadmind Unfortunately I don't have an aws env either. However, I have manually validated these changes using localstack. If testing on a real aws env necessary, I can spin one up and test it out.

pehlicd:
Hey @sysadmind, I tried with actual aws environment as well and it is working. Can we continue and merge this one?

sysadmind:
I think it's good to merge at this point. Thanks for the legwork.

PR #4511, opened 2025-08-18
chore: improves logging around webhook dispatch failure

CameronJHall:
## description

Tying the alerts that are being processed to failures reported under the dispatcher can be tricky, especially if there are many possible alerts that go through the same integration with different origins, etc. 

There are a few different ways we could associate these like using the aggrGroup logger instead of the dispatcher logger, including the alerts list in the log directly. It felt cleanest to just attach the aggrGroup key to the log so that it can be associated with any logs that come from the aggrGroup logger. I'm open to alternatives though.

## changes

- adds aggrGroup key to logs reporting dispatcher execution errors
- adds matching debug log to webhook notifier as it was missing


CameronJHall:
It's not clear why the cli tests would have failed here. Debug logs show there was only one alert in the group. 😃

pehlicd (approved):

Issue #4513, opened 2025-08-25
Jira Integration deprecated Endpoints


holger-waschke:
Creating a Jira Ticket via the Jira Integration resulting in a HTTP 401 followed by a deprecation message

`https://xxx/rest/api/2/search?fields=summary%2Cstatus%2Cresolution%2Cresolutiondate%2Cdescription%2Ccomment&jql=project+in%28%27ASP%27%29+and+labels%3D%22JIRALERT%7Bf0f4c710df4c8cf21efd079230fd5caa10e1820e8849b3210c1b547a8186741d7dc0fc016780a6677cf718b91feafbd998f08f8a474d3fcf8f8ffb2aeb986f13%7D%22+order+by+resolutiondate+desc&maxResults=2 returned status 410 Gone, error \"Die angeforderte API wurde entfernt. Bitte verwenden Sie stattdessen die neuere und verbesserte suchbasierte API. Informationen zur Einstellung finden Sie unter https://developer.atlassian.com/changelog/#CHANGE-2046.: request failed. Please analyze the request body for more details. Status code: 410\", body \"\"" receiver=jira_asp_alarm_update_summary_sandbox groupLabels="unsupported value type"`

More informations can be found here
https://developer.atlassian.com/changelog/#CHANGE-2046

The API endpoint has to be migrated to this
https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issue-search/#api-rest-api-3-search-jql-get

jkroepke:
After a [chat](https://cloud-native.slack.com/archives/C01AUBA4PFE/p1757006154523549) with @sysadmind, I would like to share a proposal to solve this issue:

### Proposal

Introduce a new configuration toggle `api_type` to control which Jira API the notifier uses.

### Motivation

Currently, the notifier relies on a single code path that supports Jira Data Center APIs and Jira Cloud APIs which are deprecated. This causes problems for Jira Cloud users, since newer endpoints are required for issue search.
By introducing `api_type`, we can:

* Provide support for the latest Jira Cloud APIs while keeping compatibility with Data Center
* Make the API choice explicit while still offering an `auto` mode for convenience
* Reduce the risk of breakage caused by deprecated Cloud APIs

### Options

* `cloud`
  Use the latest Jira Cloud APIs for issue search. Always enforce API v3 for issue search, even if v2 is configured.

* `datacenter`
[...]

PR #4514, opened 2025-08-25
Adapt Jira Integration to new search endpoints

holger-waschke:
this resolves https://github.com/prometheus/alertmanager/issues/4513

### DIFF
--- a/notify/jira/jira.go
+++ b/notify/jira/jira.go
@@ -241,12 +241,13 @@ func (n *Notifier) searchExistingIssue(ctx context.Context, logger *slog.Logger,
 		return nil, false, err
 	}
 
-	if issueSearchResult.Total == 0 {
+	issuesCount := len(issueSearchResult.Issues)
+	if issuesCount == 0 {
 		logger.Debug("found no existing issue")
 		return nil, false, nil
 	}
 
-	if issueSearchResult.Total > 1 {
+	if issuesCount > 1 {
 		logger.Warn("more than one issue matched, selecting the most recently resolved", "selected_issue", issueSearchResult.Issues[0].Key)
 	}
 
--- a/notify/jira/jira.go
+++ b/notify/jira/jira.go
@@ -225,12 +225,12 @@ func (n *Notifier) searchExistingIssue(ctx context.Context, logger *slog.Logger,
 		JQL:        jql.String(),
 		MaxResults: 2,
 		Fields:     []string{"status"},
-		Expand:     []string{},
+		Expand:     "",
 	}
 
 	logger.Debug("search for recent issues", "jql", requestBody.JQL)
 
-	responseBody, shouldRetry, err := n.doAPIRequest(ctx, http.MethodPost, "search", requestBody)
+	responseBody, shouldRetry, err := n.doAPIRequest(ctx, http.MethodPost, "search/jql", requestBody)
 	if err != nil {
 		return nil, shouldRetry, fmt.Errorf("HTTP request to JIRA API: %w", err)
 	}
--- a/notify/jira/jira_test.go
+++ b/notify/jira/jira_test.go
@@ -65,7 +65,7 @@ func TestSearchExistingIssue(t *testing.T) {
 	expectedJQL := ""
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Path {
-		case "/search":
+		case "/search/jql":
 			body, err := io.ReadAll(r.Body)
 			if err != nil {
 				http.Error(w, "Error reading request body", http.StatusBadRequest)
@@ -188,7 +188,7 @@ func TestSearchExistingIssue(t *testing.T) {
 func TestJiraTemplating(t *testing.T) {
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		switch r.URL.Path {
-		case "/search":
+		case "/search/jql":
 			w.Write([]byte(`{"total": 0, "issues": []}`))
 			return
 		default:
@@ -627,7 +627,7 @@ func TestJiraNotify(t *testing.T) {
 		t.Run(tc.title, func(t *testing.T) {
 			srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 				switch r.URL.Path {
-				case "/search":
+				case "/search/jql":
 					enc := json.NewEncoder(w)
 					if err := enc.Encode(tc.searchResponse); err != nil {
 						panic(err)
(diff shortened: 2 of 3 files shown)


k0ste:
From what JIRA version this endpoints available? I'm interested because it might break compat Alertmanager with non super-recent JIRA versions

jkroepke (changes requested):
The new endpoint seems not part of the v2 API and is not available for Jira On Premise. 

We want to preserve Jira On Premise compatibility, so I guess this is not enough.

holger-waschke:
> The new endpoint seems not part of the v2 API and is not available for Jira On Premise.

No, it´s part of v2 too, see here.
https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issue-search/#api-rest-api-2-search-jql-get

API v2 and v3 are technically identical, the only difference is the support of the ADF format, see source [here](https://developer.atlassian.com/cloud/jira/platform/rest/v3/intro/#version).


> We want to preserve Jira On Premise compatibility, so I guess this is not enough.

Well, as the documentation says 

> _Note: This integration is only tested against a Jira Cloud instance. Jira Data Center (on premise instance) can work, but it's not guaranteed._

Source [here](https://github.com/prometheus/alertmanager/blob/main/docs/configuration.md#jira_config)

This doesn´t sound for me that it´s top priority to keep the combability. Imho it shouldn´t be priority as on-premise Jira Server is EOL anyway, see [here](https://www.atlassian.com/migration/assess/journey-to-cloud).

jkroepke:
> API v2 and v3 are technically identical

OK. Cloud API v2 and v3 are technically identical (expect for issue body, where v3 requires [ADF](https://developer.atlassian.com/cloud/jira/platform/apis/document/structure/)).

But I can't find the Rest Endpoint in the Jira Server Rest API.

https://docs.atlassian.com/software/jira/docs/api/REST/8.20.1/

> Server is EOL anyway

[Jira Data Center](https://www.atlassian.com/enterprise/data-center/jira) is not EOL.

k0ste:
> This doesn´t sound for me that it´s top priority to keep the combability. Imho it shouldn´t be priority as on-premise Jira Server is EOL anyway, see [here](https://www.atlassian.com/migration/assess/journey-to-cloud).

If you make a patch for software that will break current deployments, it is a regression. Let me give you an example: the integration of AlertManager and JIRA works in a company where version v8.8.1 is deployed. There are _no plans_ to update the version, as well as to retire JIRA. That is, most likely, this version will work for decades without changes. In this case, it seems much more convenient to add the field to the AM integration YAML, that will indicate which version of the API should be used. Or to select the version programmatically (some capabilities endpoint)

Screenshot 2025-09-04 at 17 00 29

Otherwise, we will have to roll back this patch during AM build, since the integration is already completed and the human is only solving the problem, the JIRAlert does everything else. Thanks

jkroepke:
> In this case, it seems much more convenient to add the field to the AM integration YAML, that will indicate which version of the API should be used.

Would you say v2 is Jira Data Center, v3 is Jira Cloud? Not easy. Keep in mind, the Create Issue Rest API between v2 and v3 is not equal. While V2 API supports plain text with Jira Markup, in API v3 you must use [ADF](https://developer.atlassian.com/cloud/jira/platform/apis/document/structure/) to create an issue. It might be result into a mix of using API Versions.

holger-waschke:
> > In this case, it seems much more convenient to add the field to the AM integration YAML, that will indicate which version of the API should be used.
> 
> Would you say v2 is Jira Data Center, v3 is Jira Cloud? Not easy. Keep in mind, the Create Issue Rest API between v2 and v3 is not equal. While V2 API supports plain text with Jira Markup, in API v3 you must use [ADF](https://developer.atlassian.com/cloud/jira/platform/apis/document/structure/) to create an issue. It might be result into a mix of using API Versions.

Yes, the endpoints are deprecated in **both** API versions, so this wouldn´t help. 

Let´s get some things sorted:
Jira is doing rolling updates on their cloud instances and enforcing the new endpoints. So without a change soon this integration will not work anymore on the platform it was initially designed for.

While you can run software versions for decades without updating you have no assurance that 3rd party tools which are using this API will not implement breaking changes.

I can only follow official API documentation from ATL and the new Endpoint is available in V2 and V3 API Version. It´s not available  in the REST API Documentation you posted, but there´s a big deprecation warning there.
_Support for Server products ended Feb. 15, 2024. [Learn what this means for you.](https://developer.atlassian.com/developer-guide/moving-from-server-to-cloud-for-developers/)_

So IMHO we should really focus on the official supported API Versions, not the deprecated ones.

k0ste:
> While you can run software versions for decades without updating you have no assurance that 3rd party tools which are using this API will not implement breaking changes.
> 
> I can only follow official API documentation from ATL and the new Endpoint is available in V2 and V3 API Version. It´s not available in the REST API Documentation you posted, but there´s a big deprecation warning there. _Support for Server products ended Feb. 15, 2024. [Learn what this means for you.](https://developer.atlassian.com/developer-guide/moving-from-server-to-cloud-for-developers/)_
> 
> So IMHO we should really focus on the official supported API Versions, not the deprecated ones.

I don't see that you added deprecation notes to this PR. About new version of Alertmanager is impossible to use on-prem, or you will be tagged several times a month when users try to integrate AM with on-prem JIRA

jkroepke:
> So IMHO we should really focus on the official supported API Versions, not the deprecated ones.

I agree! Jira Data Center is officially supported by Atlassian, so we have to cover it as well.

https://docs.atlassian.com/software/jira/docs/api/REST/9.14.0/ - thats Jira Data center 9.14 supported until 15 February 2026. Jira 11 will be supported until 13 August 2027 as well.

See https://confluence.atlassian.com/support/atlassian-support-end-of-life-policy-201851003.html 

@k0ste Whats your On Prem Version? I can only see Jira Software V8, but maybe your version is newer.

jkroepke:
I wrote an proposal to solve this issue: https://github.com/prometheus/alertmanager/issues/4513#issuecomment-3255745841

k0ste:
> @k0ste Whats your On Prem Version? I can only see Jira Software V8, but maybe your version is newer.

Version: 8.8.1
Build Number: 808001
Build Revision: e5cdcf9fa57df21d4b441572dda42ac58b826e01

lerminou:
Hi, any news here ? jira cloud now blocks the old endpoint, the integration is KO 

conf:
`api_url: https://company.atlassian.net/rest/api/3/`

error:
```failed to look up existing issues: HTTP request to JIRA API: unexpected status code 410: {\"errorMessages\":[\"L'API demandée a été supprimée. Veuillez migrer vers l'API /rest/api/3/search/jql```

lerminou:
Ok there is another PR merged instead of this one

holger-waschke:
> Ok there is another PR merged instead of this one

yes, superseded by https://github.com/prometheus/alertmanager/pull/4542

PR #4517, opened 2025-08-31
Replace example matcher from env=\"production\" to env=\"staging\

demkkka:
This PR updates the example `exampleMatcher` configuration to improve safety when users define their own environment-specific silences. Previously, users might mistakenly create silences targeting a production environment (e.g., with `env="production"`) when they intended to target staging. Such configuration errors can lead to silencing production alerts unintentionally, which poses significant risks.  

To address this, I changed the example to prioritize silencing staging (`env="staging"`) in case of misconfiguration. This approach minimizes potential impact if users make mistakes while adapting the example for their use case.  

I'd appreciate any feedback or suggestions on this approach! Thanks!


SuperQ:
No, I don't think this is going to prevent users from doing the wrong thing. Not all organizations even use the `env` label.

Issue #4540, opened 2025-09-04
AggregationGroup maintenance is a bottleneck in alert processing pipeline


siavashs:
As mentioned in #4392 the alerts channel which dispatcher subscribes to can fill up.
We measured this with the patch in #4364 and it indeed can hit 100% capacity:

<img width="1523" height="515" alt="Image" src="https://github.com/user-attachments/assets/a778e2d9-d955-4818-9c87-ce14cd12bf31" />

But root cause of this bottleneck is not the channel capacity itself, capacity only acts as a buffer.
The root cause is the maintenance done in Dispatcher: https://github.com/prometheus/alertmanager/blob/f6b942cf9b3a503d59192eada300d2ad97cba82f/dispatch/dispatch.go#L187-L201

which happens every 30s: https://github.com/prometheus/alertmanager/blob/f6b942cf9b3a503d59192eada300d2ad97cba82f/dispatch/dispatch.go#L149

When Alertmanager receives a high alert volume, Dispatcher will dynamically create a lot of `AggregationGroup`s depending on configuration:

<img width="958" height="914" alt="Image" src="https://github.com/user-attachments/assets/eabfbd1d-5233-4fd0-8b1a-fc46b6e39aa9" />

But we observe that the rate of alert processing by dispatcher is very spiky, since every 30s it paused to maintenance and stop processing alerts.

Also the more AGs there are the longer the maintenance will take as it has to loop over tens of thousands of AGs (AGs explosion is described in #4503).

siavashs:
We know now that it was inhibition and silences.
Closing.

PR #4541, opened 2025-09-04
feat(dispatcher): add maintenance interval config

siavashs:
- make dispatcher maintenance interval configurable

Related to #4540

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -143,13 +143,14 @@ func run() int {
 	}
 
 	var (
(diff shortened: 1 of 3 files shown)


sysadmind:
Can we leave the default to the original value? I'm not very familiar with the maintenance and don't know what other effects this could have. I would be inclined to merge with the original as the default so that we don't accidentally change the behavior for others in a negative way.

siavashs:
Made the default maintenance config `30s` to match the previous static value.
Ci is as always falky!

sysadmind (approved):

PR #4542, opened 2025-09-05
Add new Jira search endpoint with new api_type option and auto detect

### LINKED ISSUE #4513: Jira Integration deprecated Endpoints

cjedro:
closes #4513

i followed https://github.com/prometheus/alertmanager/issues/4513#issuecomment-3255745841

maybe api_type = datacenter should not override host *.atlassian.net but lets discuss

also, as in the comment, only overriding the search api version for now

### DIFF
--- a/notify/jira/jira.go
+++ b/notify/jira/jira.go
@@ -248,6 +248,18 @@ func (n *Notifier) searchExistingIssue(ctx context.Context, logger *slog.Logger,
 	return &issueSearchResult.Issues[0], false, nil
 }
 
+// prepareSearchRequest builds the request body and search path for Jira issue search.
+//
+// Atlassian announced (see https://developer.atlassian.com/changelog/#CHANGE-2046) that
+// the legacy /search endpoint is no longer available on Jira Cloud. The replacement
+// endpoint (/rest/api/3/search/jql) is currently not available in Jira Data Center.
+//
+// Selection logic:
+//   - If APIType is "datacenter", always use the v2 /search endpoint.
+//   - If APIType is "cloud", or if APIType is "auto" and the host ends with
+//     "atlassian.net", use the v3 /search/jql endpoint.
+//   - Otherwise (APIType is "auto" without an atlassian.net host),
+//     use the v2 /search endpoint.
 func (n *Notifier) prepareSearchRequest(jql string) (issueSearch, string) {
 	requestBody := issueSearch{
 		JQL:        jql,
@@ -255,20 +267,16 @@ func (n *Notifier) prepareSearchRequest(jql string) (issueSearch, string) {
 		Fields:     []string{"status"},
 	}
 
-	// if the API type is datacenter always go for v2
 	if n.conf.APIType == "datacenter" {
 		searchPath := n.conf.APIURL.JoinPath("/search").String()
 		return requestBody, searchPath
 	}
 
-	// if the API type is cloud always go for v3
-	// if the API type is auto and the URL has an atlassian.net suffix, go for v3
 	if n.conf.APIType == "cloud" || n.conf.APIType == "auto" && strings.HasSuffix(n.conf.APIURL.Host, "atlassian.net") {
 		searchPath := strings.Replace(n.conf.APIURL.JoinPath("/search/jql").String(), "/2", "/3", 1)
 		return requestBody, searchPath
 	}
 
-	// if the API type is auto or empty and the URL does not have an atlassian.net suffix, go for v2
 	searchPath := n.conf.APIURL.JoinPath("/search").String()
 	return requestBody, searchPath
 }


config/notifiers.go
```diff
@@ -900,7 +900,8 @@ type JiraConfig struct {
 	NotifierConfig `yaml:",inline" json:",inline"`
 	HTTPConfig     *commoncfg.HTTPClientConfig `yaml:"http_config,omitempty" json:"http_config,omitempty"`
 
-	APIURL *URL `yaml:"api_url,omitempty" json:"api_url,omitempty"`
+	APIURL  *URL   `yaml:"api_url,omitempty" json:"api_url,omitempty"`
+	APITYPE string `yaml:"api_type,omitempty" json:"api_type,omitempty"`
```
jkroepke:
```suggestion
	APIType string `yaml:"api_type,omitempty" json:"api_type,omitempty"`
```

URL is an abbreviation, while Type is just a single world. Please update. the references as well.

notify/jira/jira.go
```diff
@@ -253,6 +248,39 @@ func (n *Notifier) searchExistingIssue(ctx context.Context, logger *slog.Logger,
...
+	if n.conf.APITYPE == "datacenter" {
+		requestBody := issueSearchDatacenter{
+			JQL:        jql,
+			MaxResults: 2,
+			Fields:     []string{"status"},
+			Expand:     []string{},
+		}
+		searchPath := n.conf.APIURL.JoinPath("/search").String()
+		return requestBody, searchPath
+	}
```
jkroepke: is this if condition useful? It feel like the same code as in line 274-281.

notify/jira/types.go
```diff
@@ -54,7 +54,15 @@ type issueStatus struct {
 	} `json:"statusCategory"`
 }
 
-type issueSearch struct {
+type issueSearchCloud struct {
```
jkroepke:
```suggestion
type issueSearch struct {
	Fields     []string `json:"fields"`
	JQL        string   `json:"jql"`
	MaxResults int      `json:"maxResults"`
}
```

My question is, if the `Expand` and `StartAt` struct members are necessary. I go trough the code and `StartAt` is not used anywhere. and `Expand` is always empty and according API docs, not mandatory.

I guess both are left-overs from the jiralert code base, but I both are remove, requests for cloud and datacenter can share the same struct. Then, we revert the any return value at the prepareSearchRequest function.

cjedro:
@jkroepke thanks so much for the review, I will work on your recommendations later today or tomorrow

jkroepke:
Consider to add the default value to https://github.com/prometheus/alertmanager/blob/d5636601f8220f7e0306b38dc0f0ee907f9e34f9/config/notifiers.go#L198

Consider to add a validation for the values here
https://github.com/prometheus/alertmanager/blob/d5636601f8220f7e0306b38dc0f0ee907f9e34f9/config/notifiers.go#L920

similar to here: 
https://github.com/prometheus/alertmanager/blob/d5636601f8220f7e0306b38dc0f0ee907f9e34f9/config/notifiers.go#L833-L837

notify/jira/jira.go
cjedro: i changed some things and added comments for each path, hope this makes more sense to you but please let me know if i should change things here.

with this i basically wanted to have the case when api_type is empty or auto but the host does not have atlassian.net which whould mean we have to use datacenter

config/config.go
```diff
@@ -609,6 +609,12 @@ func (c *Config) UnmarshalYAML(unmarshal func(any) error) error {
 				}
 				jira.APIURL = c.Global.JiraAPIURL
 			}
+			if jira.APIType == "" {
+				if c.Global.JiraAPIType == "" {
+					jira.APIType = "auto"
+				}
+				jira.APIType = c.Global.JiraAPIType
```
chrko: Isn't this now always set to the global value and therefore never `auto` if not explicitly set?
```suggestion
				} else {
					jira.APIType = c.Global.JiraAPIType
				}
```

config/notifiers.go
```diff
@@ -930,6 +931,12 @@ func (c *JiraConfig) UnmarshalYAML(unmarshal func(any) error) error {
 	if c.IssueType == "" {
 		return errors.New("missing issue_type in jira_config")
 	}
+	if c.APIType != "" &&
```
chrko: I'd cancel that option here as the empty string should always have been resolved to `auto` already.

notify/jira/jira.go
```diff
@@ -253,6 +248,31 @@ func (n *Notifier) searchExistingIssue(ctx context.Context, logger *slog.Logger,
...
+
+	// if the API type is datacenter always go for v2
+	if n.conf.APIType == "datacenter" {
+		searchPath := n.conf.APIURL.JoinPath("/search").String()
+		return requestBody, searchPath
+	}
+
+	// if the API type is cloud always go for v3
+	// if the API type is auto or empty and the URL has an atlassian.net suffix, go for v3
+	if n.conf.APIType == "cloud" || (n.conf.APIType == "auto" || n.conf.APIType == "") && strings.HasSuffix(n.conf.APIURL.Host, "atlassian.net") {
```
chrko: Based on my other comments, I think `n.conf.APIType` can never be the empty string. Therefore I'd remove this case here.
```suggestion
	if n.conf.APIType == "cloud" || (n.conf.APIType == "auto" && strings.HasSuffix(n.conf.APIURL.Host, "atlassian.net")) {
```
I've added the braces for grouping clarity (not everybody is always directly remembering that `&&` has higher precedence than `||`), but that's my personal opinion.

chrko (commented):
The requested/recommended DefaultConfig part is missing (https://github.com/prometheus/alertmanager/pull/4542#issuecomment-3261746416)
> Consider to add the default value to
> 
> https://github.com/prometheus/alertmanager/blob/d5636601f8220f7e0306b38dc0f0ee907f9e34f9/config/notifiers.go#L198

notify/jira/jira.go
cjedro: thanks for your input, I will work on your recommendations

cjedro:
@jkroepke, @chrko i added the things discussed here. 
Besides this i talked to @jkroepke and he recommended to leave out the global jira api type so i removed it, if we want to keep it i'm happy to re-add for sure

is someone aware why the tests are failing? i found this in the logs
time=2025-09-11T11:46:46.762Z level=ERROR source=main.go:559 msg="Listen error" err="listen tcp 127.0.0.1:39549: bind: address already in use"

notify/jira/jira.go
```diff
@@ -253,6 +248,31 @@ func (n *Notifier) searchExistingIssue(ctx context.Context, logger *slog.Logger,
 	return &issueSearchResult.Issues[0], false, nil
 }
 
+func (n *Notifier) prepareSearchRequest(jql string) (any, string) {
```
jkroepke:
```suggestion
// prepareSearchRequest builds the request body and search path for Jira issue search.
//
// Atlassian announced (see https://developer.atlassian.com/changelog/#CHANGE-2046) that
// the legacy /search endpoint is no longer available on Jira Cloud. The replacement
// endpoint (/rest/api/3/search/jql) is currently not available in Jira Data Center.
//
// Selection logic:
//   • If APIType is "datacenter", always use the v2 /search endpoint.
//   • If APIType is "cloud", or if APIType is "auto" and the host ends with
//     "atlassian.net", use the v3 /search/jql endpoint.
//   • Otherwise (APIType is "auto" or empty without an atlassian.net host),
//     use the v2 /search endpoint.
func (n *Notifier) prepareSearchRequest(jql string) (issueSearch, string) {
```

cjedro: @jkroepke would you add this comment or use it as a replacement for the comments inline of the function?

jkroepke:
```suggestion
func (n *Notifier) prepareSearchRequest(jql string) (issueSearch, string) {
```

lets return `issueSearch` as type here, any is not longer nessesary.

jkroepke:
Hey @sysadmind could you do a second look?

- And no idea about circle ci. Seems like a flasky test.

notify/jira/jira.go
cjedro: you're right

cjedro:
> Hey @sysadmind could you do a second look?
> 
> * And no idea about circle ci. Seems like a flasky test.

it worked now..

dinwath:
Hi, is there any planned date to release this feature? we would like to integrate jira with our alerting workflow but we're currently struck due to Atlassian change of API in Jira Cloud

cjedro:
@dinwath at least the Release is getting prepared: https://github.com/prometheus/alertmanager/pull/4586

jkroepke:
Should be part of https://github.com/prometheus/alertmanager/releases/tag/v0.29.0-rc.1

PR #4552, opened 2025-09-08
fix(dispatch): reduce locking contention

siavashs:
Reduce the amount of time spent holding locks in the dispatcher by using
the new data structures:

- doMaintenance() snapshots empty groups and deletes them afterwards
- Groups() snapshots routes and queries groups per route on demand
- processAlert() only locks the group it is processing an alert for

None of the above 3 methods hold any locks directly.

This results in -68% maintenance overhead or +12991% alert processing rate improvement:
```
goos: darwin
goarch: arm64
pkg: github.com/prometheus/alertmanager/dispatch
cpu: Apple M3 Pro
                                                 │ bench-dispatch-main.txt │  bench-dispatch-nested-map-locks.txt   │
                                                 │         sec/op          │    sec/op      vs base                 │
Dispatch_100k_AggregationGroups_10k_Empty-12                  1.242µ ±  1%   1.193µ ±   5%    -3.91% (p=0.037 n=10)
Dispatch_100k_AggregationGroups_20k_Empty-12                  1.329µ ±  2%   1.188µ ±   3%   -10.65% (p=0.000 n=10)
Dispatch_100k_AggregationGroups_30k_Empty-12                  1.437µ ±  3%   1.288µ ±   7%   -10.37% (p=0.001 n=10)
Dispatch_100k_AggregationGroups_40k_Empty-12                  1.695µ ±  8%   1.236µ ±   6%   -27.06% (p=0.000 n=10)
Dispatch_100k_AggregationGroups_50k_Empty-12                  2.185µ ± 11%   1.286µ ±  11%   -41.17% (p=0.000 n=10)
Dispatch_20k_AggregationGroups_Groups_Impact-12          189007.708µ ±  6%   1.143µ ±  18%  -100.00% (p=0.000 n=10)
Dispatch_50k_AggregationGroups_Groups_Impact-12              602.17m ± 69%   28.09m ± 605%   -95.33% (p=0.000 n=10)
Dispatch_100k_AggregationGroups_Groups_Impact-12               1.543 ± 29%    1.272 ±  96%         ~ (p=0.247 n=10)
geomean                                                       187.7µ         24.22µ          -87.10%

                                                 │ bench-dispatch-main.txt │       bench-dispatch-nested-map-locks.txt        │
                                                 │       alerts/sec        │     alerts/sec      vs base                      │
Dispatch_100k_AggregationGroups_10k_Empty-12             1.616M ±       1%        1.767M ±   3%         +9.36% (p=0.000 n=10)
Dispatch_100k_AggregationGroups_20k_Empty-12             1.412M ±       2%        1.715M ±   1%        +21.45% (p=0.000 n=10)
Dispatch_100k_AggregationGroups_30k_Empty-12             1.235M ±       4%        1.637M ±   4%        +32.58% (p=0.002 n=10)
Dispatch_100k_AggregationGroups_40k_Empty-12             950.9k ±      14%       1614.0k ±   8%        +69.73% (p=0.000 n=10)
Dispatch_100k_AggregationGroups_50k_Empty-12             693.5k ±      18%       1370.7k ±  17%        +97.65% (p=0.000 n=10)
Dispatch_20k_AggregationGroups_Groups_Impact-12           5.586 ±      14%   2027690.000 ±   5%  +36299398.75% (p=0.000 n=10)
Dispatch_50k_AggregationGroups_Groups_Impact-12           3.277 ±      40%    742918.500 ± 167%  +22667131.12% (p=0.000 n=10)
Dispatch_100k_AggregationGroups_Groups_Impact-12          1.424 ± 4255939%    252756.500 ±  73%  +17749654.21% (p=0.000 n=10)
geomean                                                  9.134k                   1.196M            +12991.42%

                                             │ bench-dispatch-main.txt │      bench-dispatch-nested-map-locks.txt       │
                                             │ maintenance_overhead_%  │ maintenance_overhead_%  vs base                │
Dispatch_100k_AggregationGroups_10k_Empty-12              17.185 ±  7%              5.905 ±   ?  -65.64% (p=0.002 n=10)
Dispatch_100k_AggregationGroups_20k_Empty-12               36.50 ± 11%              13.66 ± 26%  -62.58% (p=0.000 n=10)
Dispatch_100k_AggregationGroups_30k_Empty-12               55.44 ± 13%              17.45 ± 26%  -68.52% (p=0.002 n=10)
Dispatch_100k_AggregationGroups_40k_Empty-12              125.65 ± 27%              27.03 ± 34%  -78.49% (p=0.000 n=10)
Dispatch_100k_AggregationGroups_50k_Empty-12              172.40 ± 36%              64.64 ± 60%  -62.51% (p=0.000 n=10)
geomean                                                    59.62                    18.97        -68.17%
```

### feat(dispatch): add nested map implementation
Add custom nested map implementation with locks at root and branches.
This allows to read/write operation to not always block each other.

### chore(dispatch): split dispatch into multiple files
- split dispatch.go into multiple files
- update copyright

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -21,6 +21,8 @@ import (
 	"sync"
 	"time"
 
+	"go.uber.org/atomic"
+
 	"github.com/prometheus/common/model"
 
 	"github.com/prometheus/alertmanager/notify"
@@ -41,8 +43,7 @@ type Dispatcher struct {
 	timeout func(time.Duration) time.Duration
 
 	mtx                sync.RWMutex
-	aggrGroupsPerRoute map[*Route]map[model.Fingerprint]*aggrGroup
-	aggrGroupsNum      int
+	aggrGroupsPerRoute routeGroups
 
 	done   chan struct{}
 	ctx    context.Context
@@ -84,8 +85,10 @@ func (d *Dispatcher) Run() {
 	d.done = make(chan struct{})
 
 	d.mtx.Lock()
-	d.aggrGroupsPerRoute = map[*Route]map[model.Fingerprint]*aggrGroup{}
-	d.aggrGroupsNum = 0
+	d.aggrGroupsPerRoute = routeGroups{
+		groupsNum: &atomic.Int64{},
+		limits:    d.limits,
+	}
 	d.metrics.aggrGroups.Set(0)
 	d.ctx, d.cancel = context.WithCancel(context.Background())
 	d.mtx.Unlock()
@@ -134,40 +137,58 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
 }
 
 func (d *Dispatcher) doMaintenance() {
-	d.mtx.Lock()
-	defer d.mtx.Unlock()
-	for _, groups := range d.aggrGroupsPerRoute {
-		for _, ag := range groups {
+	type groupToRemove struct {
+		route *Route
+		fp    model.Fingerprint
+		ag    *aggrGroup
+	}
+
+	var toRemove []groupToRemove
+
+	// First pass: collect groups to remove
+	d.aggrGroupsPerRoute.Range(func(route *Route, groups *fingerprintGroups) bool {
+		groups.Range(func(fp model.Fingerprint, ag *aggrGroup) bool {
 			if ag.empty() {
-				ag.stop()
-				d.marker.DeleteByGroupKey(ag.routeID, ag.GroupKey())
-				delete(groups, ag.fingerprint())
-				d.aggrGroupsNum--
-				d.metrics.aggrGroups.Dec()
+				toRemove = append(toRemove, groupToRemove{route, fp, ag})
 			}
+			return true
+		})
+		return true
+	})
+
+	// Second pass: remove collected groups
+	for _, item := range toRemove {
+		item.ag.stop()
+		d.marker.DeleteByGroupKey(item.ag.routeID, item.ag.GroupKey())
+		groupsMap := d.aggrGroupsPerRoute.GetRoute(item.route)
+		if groupsMap != nil {
+			groupsMap.RemoveGroup(item.fp)
+			d.metrics.aggrGroups.Dec()
 		}
 	}
 }
 
 // Groups returns a slice of AlertGroups from the dispatcher's internal state.
 func (d *Dispatcher) Groups(routeFilter func(*Route) bool, alertFilter func(*types.Alert, time.Time) bool) (AlertGroups, map[model.Fingerprint][]string) {
-	groups := AlertGroups{}
+	// Snapshot the outer map in routeGroups to
+	//  avoid holding the read lock when dispatcher has 1000s of aggregation groups.
+	routeGroups := routeGroups{}
+	d.aggrGroupsPerRoute.Range(func(route *Route, groups *fingerprintGroups) bool {
+		routeGroups.AddRoute(route)
+		return true
+	})
 
-	d.mtx.RLock()
-	defer d.mtx.RUnlock()
-
-	// Keep a list of receivers for an alert to prevent checking each alert
-	// again against all routes. The alert has already matched against this
-	// route on ingestion.
+	// TODO: move this processing out of Dispatcher, it does not belong here.
+	alertGroups := AlertGroups{}
 	receivers := map[model.Fingerprint][]string{}
-
 	now := time.Now()
-	for route, ags := range d.aggrGroupsPerRoute {
+	routeGroups.Range(func(route *Route, _ *fingerprintGroups) bool {
 		if !routeFilter(route) {
-			continue
+			return true
 		}
 
-		for _, ag := range ags {
+		// Read inner fingerprintGroups if necessary.
+		d.aggrGroupsPerRoute.GetRoute(route).Range(func(fp model.Fingerprint, ag *aggrGroup) bool {
 			receiver := route.RouteOpts.Receiver
 			alertGroup := &AlertGroup{
 				Labels:   ag.labels,
@@ -197,22 +218,24 @@ func (d *Dispatcher) Groups(routeFilter func(*Route) bool, alertFilter func(*typ
 				filteredAlerts = append(filteredAlerts, a)
 			}
 			if len(filteredAlerts) == 0 {
-				continue
+				return true
 			}
 			alertGroup.Alerts = filteredAlerts
 
-			groups = append(groups, alertGroup)
-		}
-	}
-	sort.Sort(groups)
-	for i := range groups {
-		sort.Sort(groups[i].Alerts)
+			alertGroups = append(alertGroups, alertGroup)
+			return true
+		})
+		return true
+	})
+	sort.Sort(alertGroups)
+	for i := range alertGroups {
+		sort.Sort(alertGroups[i].Alerts)
 	}
 	for i := range receivers {
 		sort.Strings(receivers[i])
 	}
 
-	return groups, receivers
+	return alertGroups, receivers
 }
 
 // Stop the dispatcher.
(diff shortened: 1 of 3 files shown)


OGKevin:
The thing I like about

```go
mtx.Lock()
defer mtx.Unlock()
```

is that it makes the management of the lock very clear and explicit.

Your changes make sense to me. However, I wonder if, for improved maintainability, it would be worth extracting the logic that needs to hold locks into smaller functions where we can still use `defer` for unlocking.

For example, instead of:

```go
func (d *Dispatcher) doMaintenance() {
    d.mtx.RLock()
    empty := make(map[*Route]map[model.Fingerprint]*aggrGroup)
    for route, groups := range d.aggrGroupsPerRoute {
        for fp, ag := range groups {
            if ag.empty() {
                if empty[route] == nil {
                    empty[route] = make(map[model.Fingerprint]*aggrGroup)
                }
                empty[route][fp] = ag
            }
        }
    }
    d.mtx.RUnlock()
}
```

it could become:

```go
func (d *Dispatcher) makeEmptyAggrMap() map[*Route]map[model.Fingerprint]*aggrGroup {
    d.mtx.RLock()
    defer d.mtx.RUnlock()

    empty := make(map[*Route]map[model.Fingerprint]*aggrGroup)
    for route, groups := range d.aggrGroupsPerRoute {
        for fp, ag := range groups {
            if ag.empty() {
                if empty[route] == nil {
                    empty[route] = make(map[model.Fingerprint]*aggrGroup)
                }
                empty[route][fp] = ag
            }
        }
    }

    return empty
}

func (d *Dispatcher) doMaintenance() {
    empty := d.makeEmptyAggrMap()
}
```

This approach could be applied to all the places where lock holding is involved, essentially making the code easier to maintain and reason about.

wdyt?

siavashs:
I reimplemented the change using custom nested structures.
This allows dispatcher to hold r/w locks at different branches of the data structure, so it is less blocking for parallel operations.
It also addresses your comment @OGKevin since we don't hold any locks in dispatcher methods.

I will update PR description and benchmarks tomorrow.

dispatch/group_test.go
```diff
@@ -0,0 +1,351 @@
...
+	removeEndsAt := func(as types.AlertSlice) types.AlertSlice {
+		for i, a := range as {
+			ac := *a
+			ac.EndsAt = time.Time{}
+			as[i] = &ac
+		}
+		return as
+	}
+
+	// Test regular situation where we wait for group_wait to send out alerts.
```
OGKevin, line 134: This test function covers several scenarios. Would it be better to wrap each scenario in a t.Run() block? That would allow a single failing test to be run repeatedly for easier debugging.

dispatch/group_test.go
```diff
@@ -0,0 +1,351 @@
...
+
+			if !reflect.DeepEqual(batch, exp) {
+				t.Fatalf("expected alerts %v but got %v", exp, batch)
+			}
+		}
+	}
+
+	ag.stop()
+
+	// Add an alert that started more than group_interval in the past. We expect
```
OGKevin, line 185: Same idea here

dispatch/group_test.go
```diff
@@ -0,0 +1,351 @@
...
+			exp := removeEndsAt(types.AlertSlice{a1, a2, a3})
+			sort.Sort(batch)
+
+			if !reflect.DeepEqual(batch, exp) {
+				t.Fatalf("expected alerts %v but got %v", exp, batch)
+			}
+		}
+	}
+
+	// Resolve an alert, and it should be removed after the next batch was sent.
```
OGKevin, line 233: and here

dispatch/group_test.go
```diff
@@ -0,0 +1,351 @@
...
+			t.Fatalf("received batch too early after %v", s)
+		}
+		sort.Sort(batch)
+
+		if !reflect.DeepEqual(batch, exp) {
+			t.Fatalf("expected alerts %v but got %v", exp, batch)
+		}
+	}
+
+	// Resolve all remaining alerts, they should be removed after the next batch was sent.
```
OGKevin, line 256: And this last scenario

OGKevin (commented):
Looks 👌🏾 Just a few comments on wrapping test sencarios.

Would be nice to make metrics an interface so that we can also test if metrics are being set correctly, but I would say that is for a diff mr.

dispatch/group_test.go
siavashs: I think we can do this in a separate MR.

The `group_test.go` was split from `dispatch_test.go` without modifications.
In `dispatch_test.go` there is `TestGroupsWithLimits` which checks the `alertmanager_dispatcher_aggregation_group_limit_reached_total` metric.

Note: Limits are currently not used in Alertmanager, it's a feature imported by Grafana Mimir.

dispatch/notify.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"context"
+
+	"github.com/prometheus/alertmanager/types"
+)
+
+// notifyFunc is a function that performs notification for the alert
+// with the given fingerprint. It aborts on context cancelation.
+// Returns false iff notifying failed.
```
sysadmind, line 24:
```suggestion
// Returns false if notifying failed.
```

dispatch/map.go
```diff
@@ -0,0 +1,135 @@
...
+// routeGroups is a map of routes to fingerprintGroups.
+// It is a nested map implementation which avoid lock contention.
+// The outer map is protected by a single mutex.
+// The inner maps are protected by dedicated mutexes.
+// The shared atomic counter is used to track the number of nested groups.
+// Limits are shared between all groups.
+// // Each branch of the map can hold its own R(W)Lock.
+type routeGroups struct {
+	mu          sync.RWMutex
+	routeGroups map[*Route]*fingerprintGroups
```
sysadmind, line 33: This seems like a weird name conflict with the struct name. The comment is also slightly misleading. This struct isn't the map, but a structure that holds the map. Should this be called something like fingerprints because that's what it is pointing to?

dispatch/map.go
```diff
@@ -0,0 +1,135 @@
...
+	groupsNum   *atomic.Int64
```
sysadmind, line 34: This seems like it would be better named in the reverse - `numGroups`

dispatch/map.go
```diff
@@ -0,0 +1,135 @@
...
+	rg.mu.RLock()
+	defer rg.mu.RUnlock()
+	for route, groups := range rg.routeGroups {
+		if !fn(route, groups) {
+			break
+		}
+	}
+}
+
+// fingerprintGroups is a map of fingerprints to aggregation groups.
```
sysadmind, line 74: This comment is also technically wrong

dispatch/map.go
```diff
@@ -0,0 +1,135 @@
...
+// It is protected by a dedicated RW mutex.
+// It inherits the shared atomic counter from the parent routeGroups to track the number of total groups.
+// It inherits the limits from the parent routeGroups.
+type fingerprintGroups struct {
+	mu         sync.RWMutex
+	aggrGroups map[model.Fingerprint]*aggrGroup
+	groupsNum  *atomic.Int64
```
sysadmind, line 81: I think this would also be better to rename as `numGroups`

dispatch/map.go
```diff
@@ -0,0 +1,135 @@
...
+// If the group already exists, it returns the existing aggregation group.
+func (fg *fingerprintGroups) AddGroup(fp model.Fingerprint, ag *aggrGroup) (group *aggrGroup, count int64, limit int) {
+	fg.mu.Lock()
+	defer fg.mu.Unlock()
+
+	if fg.aggrGroups == nil {
+		fg.aggrGroups = make(map[model.Fingerprint]*aggrGroup)
+	}
+	// Check if we've reached the rate limit before creating a new group.
+	if fg.LimitReached() {
```
sysadmind, line 103: Would this not be a good time to return an error?

dispatch/dispatch.go
```diff
@@ -185,60 +137,58 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
...
+			return true
+		})
+		return true
+	})
 
-func (ag AlertGroups) Swap(i, j int) { ag[i], ag[j] = ag[j], ag[i] }
-func (ag AlertGroups) Less(i, j int) bool {
-	if ag[i].Labels.Equal(ag[j].Labels) {
-		return ag[i].Receiver < ag[j].Receiver
+	// Second pass: remove collected groups
```
sysadmind, line 159: The trouble with not having the lock this whole time is that between defining what to remove above and actually removing it below, the state could have changed because we give up the read lock, then try to get a write lock. Who knows what happens in between. Is this maintenance function the one that take the significant amount of time? 

If so, we could try to range over it to find empty groups, and when we do, instead of putting it in a slice, we could decide to do the expensive operation of taking a write lock, again reading to make sure it's still empty, then deleting/removing it. That would keep write locks down to only when we have something to do, even if we pay that price twice in the less likely scenario that we have cleanup to do.

sysadmind (commented):
It took me a while to review, mostly because of how much was moved. At this point we should leave it like it is, because I've gone through the review. I would suggest in the future to move code in a separate PR because I spent a lot of time just reviewing lines that moved to a different file.

siavashs:
> It took me a while to review, mostly because of how much was moved. At this point we should leave it like it is, because I've gone through the review. I would suggest in the future to move code in a separate PR because I spent a lot of time just reviewing lines that moved to a different file.

There are 3 atomic commits which can be reviewed separately, if that helps.

dispatch/dispatch.go
siavashs: > Is this maintenance function the one that take the significant amount of time?

yes, if you check the benchmarks there is a huge perfomance penalty of locking the dispatcher every 30s.

> Who knows what happens in between. Is this maintenance function the one that take the significant amount of time?

I think the code can be improved, but we don't want to keep the lock at the root while looping over branches.
When trying to remove a group, we can double check if it's still empty or not.

sysadmind: I think for correctness we have to re-check for emptyness inside an RWLock.

sysadmind:
Looks like there's a merge conflict.

siavashs:
So #4607 turned out to be the major bottleneck in dispatch.
But I will simplify this PR and update it since there are still minor improvements we can add.

siavashs:
Closing in favour of #4670

Issue #4319, opened 2025-03-28
msteamsv2 Receiver does not support Summary field


cbryant42:
The old msteams receiver supported adding/templating a Summary field, which is the display text for the notification and the displayed text inside the activity center. Now the notifications just say "{webhook name} Card". This isn't the biggest issue, but it would be nice to be able to customize this text to allow better notifications. 

I've been looking into what changes this would require, and it does not look significant. However I do not have a pull request at the moment. I thought I would open this issue in advance to see if others also wanted this fixed. 

Thanks!

grobinson-grafana:
When we looked at the new Adaptive Cards/Power Automate API, it seems there was nowhere to use the summary, so we removed it from `msteamsv2`. If this is incorrect, please feel free to open a pull request with the fix.

cbryant42:
We have been using adaptive cards with a custom proxy for a while now is the only reason I knew this was possible. I'll work on getting a pull request in, but in the meantime, this is how you can structure the card to set the summary field. 

```// A message as referenced in https://learn.microsoft.com/en-us/connectors/teams/?tabs=text1%2Cdotnet#request-body-schema
	t := teamsMessage{
		Type:    "message",
		Summary: summary,
		Attachments: []Attachment{
			{
				ContentType: "application/vnd.microsoft.card.adaptive",
				ContentURL:  nil,
				Content: Content{
					Schema:  "http://adaptivecards.io/schemas/adaptive-card.json",
					Type:    "AdaptiveCard",
					Version: "1.2",
					Body: []Body{
						{
							Type:   "TextBlock",
							Text:   title,
							Weight: "Bolder",
							Size:   "Medium",
							Wrap:   true,
							Style:  "heading",
							Color:  color,
						},
						{
							Type: "TextBlock",
							Text: text,
							Wrap: true,
						},
					},
					Msteams: Msteams{
						Width: "full",
					},
				},
			},
		},
	}```

grobinson-grafana:
Where did you find this as it's not mentioned in https://learn.microsoft.com/en-us/connectors/teams/?tabs=text1%2Cdotnet#request-body-schema?

cbryant42:
Honestly... I'm not sure. I made that change to my custom proxy in 2023 and didn't document where I found the documentation. I think I just tested a few iterations based on various forum posts I found around the topic like: https://stackoverflow.com/questions/68297798/push-notifications-toasts-summary-for-adaptive-cards-on-microsoft-teams
[...]

Issue #4161, opened 2024-12-12
[FEATURE REQUEST] - MS Teams Adaptive Card Additional Configuration


mwasson74:
## Problem Statement
We are being **_forced_** to transition from Slack Alerts to Microsoft Teams for notifications. Currently, the Adaptive Cards supported by AlertManager are limited to just "Title" and "Text." This restricts formatting options and makes it challenging to replicate the functionality and customization available in Slack's receiver configuration. As a result, creating effective and well-structured alerts for Teams requires overly complex/creative workarounds in the formatting that the go template produces.

## Proposed Solution
Enhance the Adaptive Cards support in AlertManager with the following features:
1. **Clickable Titles:** Allow the "Title" to be a hyperlink that redirects to the alert, similar to Slack's receiver configuration.
2. **Action Support:** Add support for "actions," enabling users to include interactive buttons like those available in Slack's configuration.
3. **Flexible TextBlock Collections:** Introduce the ability to configure a collection of "TextBlocks" in the `alertmanager.yml` receiver configuration. This would provide greater control over card formatting by allowing multiple sections or fields instead of a single "Title" and "Text."
4. **Additional Adaptive Card Features:** Include a broader range of Adaptive Card capabilities, such as:
   - Custom layouts and dynamic fields (e.g., images, charts).
   - Advanced styling options (e.g., font sizes, colors, alignment).
   - Conditional formatting or visibility rules.
5. **Improved Configuration Support:** Provide a simplified way to leverage these features directly in the `alertmanager.yml` file, making the experience intuitive for users.

## Benefits
- **Streamlined Transition:** Makes it easier to switch from Slack to Teams without losing customization and interactivity.
- **Improved Formatting:** Enables better organization of information in alerts, enhancing readability and usability.
- **Enhanced User Experience:** Facilitates rich, interactive notifications that can be tailored to specific workflows.
- **Reduced Complexity:** Eliminates the need for overly creative or complex go templates to achieve desired layouts and functionality.

## Additional Context
Using AlertManager version `v0.28.0-rc.0`, I’ve had to get creative with what my go templates produce to approximate the functionality that Slack’s receiver configuration provides by default. Improved Adaptive Card configurability would simplify this process significantly and bring parity to the Teams integration.

1. **Slack Alert Example:**  
![image](https://github.com/user-attachments/assets/fadc6fdf-c94c-46fb-9f34-84b92cf6071f)

2. **Microsoft Teams Alert Example:**
[...]

PR #4553, opened 2025-09-08
Use markdown link for alert source URL

GaryGSC:
I realize there's been _much_ discussion (#3503, #3799, #3925, #4161, #4243, #4319, etc.) around using Adaptive Cards and generally making Microsoft Teams notifications prettier and more skimmable. Those issues have made slow progress because they typically tangent into [larger discussions about user-defined templates or concerns that templates haven't handled all possible edge cases](https://github.com/prometheus/alertmanager/pull/4243#issuecomment-2718061662).

In my org, we'd get huge usability wins out of just linkifying the source URL.

Right now, the source URL is a wall of text (bad for skimming) that has to be copied and pasted into a browser (currently impossible on the Teams mobile app):
image

This PR uses Markdown syntax to condense that all down to a clickable link. It's a 1-liner within a fragment that's already titled `__text_alert_list_markdown`.


GaryGSC:
It appears that some Jira tests are failing because there are some cases where `GeneratorURL` isn't provided and so the updated template is creating broken markdown links ( `[Source]()` ).

As someone unfamiliar with the Jira ecosystem, how crucial is it that we keep the literal text `Source:` even when the `GeneratorURL` is missing?

This feels silly to me:
```go
{{- if .GeneratorURL }}
[Source]({{ .GeneratorURL }})
{{- else }}
Source:
{{- end }}
```

I'd rather do this:
```go
{{- if .GeneratorURL }}
[Source]({{ .GeneratorURL }})
{{- end }}
```

rhettjay:
I wonder if it would even fix it with adding `:`
```
{{- if .GeneratorURL }}
[Source:]({{ .GeneratorURL }})
{{- end }}
```

or another alternative:

```
{{- if .GeneratorURL }}
Source: [{{ .GeneratorURL }}]({{ .GeneratorURL }})
{{- end }}
```

if that's too verbose:

```
{{- if .GeneratorURL }}
Source: [Alertmanager Link]({{ .GeneratorURL }})
{{- end }}
```

PR #4561, opened 2025-09-16
feat(notify): implement executable-based notifier

### LINKED ISSUE #4558: Add generic pluggable exec notifier
A generic, pluggable notifier would reduce the number of feature requests for new notifier implementations.

# Motivation

The current recommendation for integrating services ($SVC) into Prometheus Alertmanager (AM) is to use the webhook notifier. If the $SVC does not support AM natively, a bridge has to be used in order to translate the webhook DTO into whatever data packets the receiving end supports. When integrating AM into many such services, this results in many additional services being brought up, all of which need to be scaled, placed, monitored, ... . While i am not against microservices per-se, deploying AM-bridges seems like an overkill to me, given that these services "only" translate. While i could write a generic bridge that supports multiple target services, i was thinking about a pluggable solution for AM.

Providing a generic notifier implementation based on `os/exec` would allow operators to bring their own implementations without having to deal with scaling/monitoring/...

# Proposal

Notifier implementation based on `os/exec` which runs a command for each notification request. Data (similar to the webhook DTO) is provided via STDIN. The exit code is used to signal the result (0=success, 3=retry, else=failure)

I do have a PoC that i can clean up and provide as first draft via MR.

UiP9AV6Y:
invoke an executable for each notification request and stream the alert payload JSON-encoded via STDIN.

closes #4558


Spaceman1701:
For the time being, we've decided that we should not include an executable notifier integration in alertmanager. See https://github.com/prometheus/alertmanager/issues/4558#issuecomment-3647415598 on the associated issue.

Thank you for taking the time to contribute and open this PR!

Issue #4585, opened 2025-10-02
jira v3 description is not Atlassian ADF format


lerminou:
Migrating from the v2 to the v3 format due to the atlassian deprecation and deletion of the previous API.

I followed the documentation here to set the description

https://prometheus.io/docs/alerting/latest/configuration/#jira_config

> Both APIs have the same feature set. The difference is that V2 supports [Wiki Markup](https://jira.atlassian.com/secure/WikiRendererHelpAction.jspa?section=all) for the issue description and V3 supports [Atlassian Document Format (ADF)](https://developer.atlassian.com/cloud/jira/platform/apis/document/structure/). The default jira.default.description template only works with V2.

 So I set the descrpition to a new template (without any variables just like their example)

https://developer.atlassian.com/cloud/jira/platform/apis/document/structure/

And I got an error in the logs 

`level=ERROR source=dispatch.go:363 msg="Notify for alerts failed" component=dispatcher num_alerts=2 err="jira/jira[0]: notify retry canceled due to unrecoverable error after 1 attempts: failed to POST request to \"issue\": unexpected status code 400: {\"errorMessages\":[],\"errors\":{\"description\":\"La valeur de l'opération doit être un document Atlassian (voir le format de document Atlassian)\"}}"`

I see in the code you try to unmarshal the description: https://github.com/prometheus/alertmanager/blob/main/notify/jira/jira.go#L167-L172

lerminou:
According to my friend chatgpt, I tried a version directly in the `fields` instead of the description like this:

```
receivers:
- jira_configs:
  - api_url: https://company.atlassian.net/rest/api/3/
    fields:
      description:
        type: doc
        version: 1
        content:
          - type: paragraph
            content:
              - type: text
                text: "Alert triggered from Alertmanager"
      customfield_10002:
      - 1
```

And this way, the unmarshal is not performed and the alert is well created in jira But I haven"t the templating option.

bensalah-nabil:
i have the same issue to use jira api with alertmanager latest  version

[...]

PR #4621, opened 2025-10-21
Jira Integration: Allow configuring issue update via parameter

holger-waschke:
This change allows disabling the update of issue descriptions.

It´s a non breaking change the default is still to update everything. Only if you set the parameter disable_update_description explicit to true the description won´t get updated anymore. 

We need this feature as there a lot of jira automations which updates the description itself, jiralert would revert the changes back to the original templated description state once the repeat_interval kicks in and the alert is still active, so it´s handy to have the possibility to disable it.

I added a new unit test for this function to test.

Exempel config

```yaml
receivers:
- name: jira_alarm
  jira_configs:
  - project: MON
    issue_type: Bug
    resolve_transition: "Geschlossen"
    reopen_transition: "Erfasst"
    reopen_duration: "1h"
    disable_update_description: true
```

### DIFF
--- a/notify/jira/types.go
+++ b/notify/jira/types.go
@@ -69,7 +69,7 @@ type issueTransitions struct {
 
 // MarshalJSON merges the struct issueFields and issueFields.CustomField together.
 func (i issueFields) MarshalJSON() ([]byte, error) {
-	jsonFields := map[string]interface{}{}
+	jsonFields := map[string]any{}
 
 	if i.Summary != nil {
 		jsonFields["summary"] = *i.Summary


cjedro:
this is also important due to upcoming rate-limiting https://developer.atlassian.com/cloud/jira/platform/rate-limiting/

holger-waschke:
> this is also important due to upcoming rate-limiting https://developer.atlassian.com/cloud/jira/platform/rate-limiting/

tbh this sounds to me like it will be mandatory to have some kind of back off mechanism to avoid running into these enforced rate limits?

cjedro:
@holger-waschke i think this is already implemented
https://github.com/prometheus/alertmanager/blob/f7ff687529f86467e11fe1336e56756c0f275bdb/notify/jira/jira.go#L63

fluktuid:
I'm with you, but this handles every alarm separate.
So in the worst-case when many alarms are open and firing, the rate-limit could result in no new alarms opened.
This could happen because of the push-mechanism that re-sends old alarms for updates.

I think generally an optional circuit-breaker could solve this.
Spontaneously I would think about two queues:
1. more important (e.g. new alarms)
2. less important (e.g. update/close alarms)

Maybe we should discuss this in an issue.

Spaceman1701:
I like this idea, but I'm wondering if we should have a more generic implementation. I'm less familiar with the Jira integration than I should be, honestly. Are there any other fields which we might want to avoid updating on repeat notifications? I'm wondering if we should allow fields to have sub-keys that indicate if we should update them or not.

I also think the discussion about rate-limiting is important enough to move to a dedicated issue. It's really a problem for any receiver, so it's something we should ideally solve generically.

holger-waschke:
> I like this idea, but I'm wondering if we should have a more generic implementation. I'm less familiar with the Jira integration than I should be, honestly. Are there any other fields which we might want to avoid updating on repeat notifications? I'm wondering if we should allow fields to have sub-keys that indicate if we should update them or not.
> 
> I also think the discussion about rate-limiting is important enough to move to a dedicated issue. It's really a problem for any receiver, so it's something we should ideally solve generically.

we´re using this Jira API Endpoint to update existing issues
https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issues/#api-rest-api-2-issue-issueidorkey-put

In the request body we send the fields we would like to update. which are the same as creating a new issue so basically everything. 
from my experience what should be configurable (whats most relevant to users) is the summary and description. customfields, priority and other stuff is more static in most cases. (for example customfield hostname or environment will not change)

Spaceman1701:
> from my experience what should be configurable (whats most relevant to users) is the summary and description. customfields, priority and other stuff is more static in most cases. (for example customfield hostname or environment will not change)

That makes sense to me... The thing I'm worried about is that we'll end up having multiple `disable_update_*` keys floating around (which seems a little ugly to me). 

What do you think about introducing a new `JiraField` type in the config that we can either deserialize from string (to maintain backwards compatibility) or from an object with keys that control updating? It's a bit more complexity, but I think it's a bit cleaner in the long run.

This would result in config like:
```
receivers:
- name: jira_alarm
  jira_configs:
  - project: MON
    issue_type: Bug
    resolve_transition: "Geschlossen"
    reopen_transition: "Erfasst"
    reopen_duration: "1h"
    description:
      disable_update: true
```
We could then reuse the type for `summary` or any other field that might end up needing this behavior.

cc @siavashs and @grobinson-grafana since they might have a different opinion.

holger-waschke:
@Spaceman1701 
I implemented your suggestion. The unmarshalling remains backward-compatible, so nothing breaks — both formats are now supported:

```
# Old/legacy format (just a string)
summary: '{{ template "jira.default.summary" . }}'
description: '{{ template "jira.default.description" . }}'
# New format (object with fields)
summary:
  template: '{{ template "jira.default.summary" . }}'
  disable_update: true
description:
  template: '{{ template "jira.default.description" . }}'
  disable_update: true
```
The jira_test required a fair amount of refactoring to accommodate this, but it feels justified.

notify/jira/types.go
```diff
@@ -30,7 +30,7 @@ type issueFields struct {
 	Priority    *idNameValue  `json:"priority,omitempty"`
 	Project     *issueProject `json:"project,omitempty"`
 	Resolution  *idNameValue  `json:"resolution,omitempty"`
-	Summary     string        `json:"summary"`
+	Summary     any           `json:"summary"`
```
Spaceman1701: why did this change? I see that `Description` is also `any`, so I think this is fine. Just answering here in github is enough that someone will be able to learn the answer if they come looking for it in the future 😄

config/notifiers.go
```diff
@@ -991,6 +1002,27 @@ type JiraConfig struct {
 	Fields map[string]any `yaml:"fields,omitempty" json:"custom_fields,omitempty"`
 }
 
+// Supports both the legacy string and the new object form.
```
Spaceman1701, line 1012: This looks great! The documentation in `docs/configuration.md` should be updated to reflect this new behavior.

notify/jira/types.go
holger-waschke: `any` is an alias for `interface{}` and we´re using it in combination with the MarshalJSON function

```
// MarshalJSON merges the struct issueFields and issueFields.CustomField together.
func (i issueFields) MarshalJSON() ([]byte, error) {
	jsonFields := map[string]interface{}{}

	if i.Summary != nil {
		jsonFields["summary"] = i.Summary
	}

	if i.Description != nil {
		jsonFields["description"] = i.Description
	}
```

to ensure that the field is nil and not an empty string when updating existing issues with the fields not being included. parsing an empty string would just update the field to an empty string so we have to make sure it´s omitted from the JSON completely. I´m not quite sure if a string pointer could be used here as well I tested it and I think it would need quite some reworking.

grobinson-grafana: We try to avoid using `any` or `interface{}` in the configuration file. Instead, we tend to work on the basis that an empty string means no value. Why can't we just do `if i.Summary != ""`?

Spaceman1701: Yeah, I'm inclined to agree here. And if the intent is "a string or nil", I think think using a `*string` expresses that much better than `any`. Otherwise, we're opening ourselves up to future changes introducing mistyped fields in jira requests.

grobinson-grafana: Yep! Use `*string` if you want empty string to be a valid value instead of empty, otherwise use `string`.

holger-waschke: somehow anticipated this change request and I yes I agree with you :-) . Changed it to use *string. Added little helper to the unit test 

```
func stringPtr(v string) *string {
	return &v
}

```

holger-waschke:
Built in more improvements 
- improved log output when updating existing issues debug output will now print out which labels are being skipped, e.g.

> level=DEBUG source=jira.go:119 msg="updating existing issue without summary and description" integration=jira group_key="{}/{test=\"true\"}:{alertname=\"E2E Failed Application\", environment=\"Abnahme\"}" issue_key=ASPMIG-447810

- improved JiraFieldConfig unmarshaling to respect the template field default. Setting summary and description is fully optional. you can have it like this or leave it full out.
```
- name: jira_asp_alarm
  jira_configs:
  - project: ASPMIG
    issue_type: Alarm
    summary:
      template: '{{ template "jira.default.summary" . }}'
    description:
      disable_update: true
```

new logic for the atlassian v3 api endpoint. (which uses atlassian document format). I had to rewrite this because of the string pointer anyway so It´s optimized for only check for valid JSON. The actual encoding takes place in the doAPIRequestFullPath function anyway.
```
	descriptionCopy := issueDescriptionString
	if isAPIv3Path(n.conf.APIURL.Path) {
		if !json.Valid([]byte(descriptionCopy)) {
			return issue{}, fmt.Errorf("description template: invalid JSON for API v3")
		}
	}
	requestBody.Fields.Description = &descriptionCopy
```

However, v3 endpoint with atlassian document format is broken atm, but this is another issue I´ll reply to https://github.com/prometheus/alertmanager/issues/4585. This needs quite some refactoring, too.

notify/jira/jira.go
```diff
@@ -104,14 +104,35 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
...
+			disabled = append(disabled, "summary")
+		}
+		if n.conf.Description.DisableUpdate {
+			disabled = append(disabled, "description")
+		}
+		if len(disabled) > 0 {
+			msg += " without " + strings.Join(disabled, " and ")
+		}
+
+		logger.Debug(msg, "issue_key", existingIssue.Key)
```
Spaceman1701: small nitpick: It's not very idiomatic to build a dynamic `msg` for structured logging. I'd rather see 
```suggestion
		logger.Debug(msg, "issue_key", existingIssue.Key, "disabled", disabled)
```
This makes things a little better for those who are indexing their logs. Example: https://go.dev/play/p/0dlLuNhdO8e

holger-waschke: > small nitpick: It's not very idiomatic to build a dynamic `msg` for structured logging. I'd rather see
> 
> This makes things a little better for those who are indexing their logs. Example: https://go.dev/play/p/0dlLuNhdO8e

sure thing :smile:

ultrotter:
@SuperQ mentioned one last change he would like: could we have the flag be positive?
i.e. change disable_update_description: true to update_description: false ?

Thanks!

holger-waschke:
> @SuperQ mentioned one last change he would like: could we have the flag be positive? i.e. change disable_update_description: true to update_description: false ?
> 
> Thanks!

sure, however it´s a breaking change then because the default behavior will change from update everything to NOT update summary / description if not explicit set otherwise. is that ok for you?

grobinson-grafana:
> > @SuperQ mentioned one last change he would like: could we have the flag be positive? i.e. change disable_update_description: true to update_description: false ?
> > Thanks!
> 
> sure, however it´s a breaking change then because the default behavior will change from update everything to NOT update summary / description if not explicit set otherwise. is that ok for you?

What about `update_description: true` as the default?

SuperQ:
Yes, my suggestion was to change the naming / default from `disable_update_description: false` to `update_description: true`.

This avoids the `false: false` double negative.

holger-waschke:
> Yes, my suggestion was to change the naming / default from `disable_update_description: false` to `update_description: true`.
> 
> This avoids the `false: false` double negative.

I see. But since it´s a bool and defaults to false how exactly you want to make it default to true? 
Use a pointer an threat nil as true with a helper function
Or extend the UnmarshalYAML with a default?

SuperQ:
Take a look at the Prometheus `config/config.go`.

https://github.com/prometheus/prometheus/blob/main/config/config.go#L192-L200

holger-waschke:
> Take a look at the Prometheus `config/config.go`.
> 
> https://github.com/prometheus/prometheus/blob/main/config/config.go#L192-L200

Played around a bit and came up with this solution.
use a pointer on the bool and a little helper function
```
func (f *JiraFieldConfig) EnableUpdateValue() bool {
	if f.EnableUpdate == nil {
		return true
	}
	return *f.EnableUpdate
}
```

SuperQ (approved):

PR #4623, opened 2025-10-22
feat: implement phantom_threading to group email alerts into threads

stapelberg:
Some email clients such as Gmail apparently use their own heuristics for threading and already implement this behavior based on the subject.

But for users of other email clients that only implement threading based on the relevant headers (e.g. notmuch), those users currently get one email thread for each newly firing alert.

With phantom_threading enabled, all alert emails (of the same alert) on the same day are grouped into the same thread. Much nicer :)

---

I have tested this manually and you can see the effect start to work in this screenshot:

2025-10-22-alertmanager-threading

(Monday morning, I got one thread per alert email notification; in the evening, the threading change was effective and emails are grouped into the daily thread.)

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -315,6 +315,13 @@ type EmailConfig struct {
 	Text             string               `yaml:"text,omitempty" json:"text,omitempty"`
 	RequireTLS       *bool                `yaml:"require_tls,omitempty" json:"require_tls,omitempty"`
 	TLSConfig        *commoncfg.TLSConfig `yaml:"tls_config,omitempty" json:"tls_config,omitempty"`
+	Threading        ThreadingConfig      `yaml:"threading,omitempty" json:"threading,omitempty"`
+}
+
+// ThreadingConfig configures mail threading.
+type ThreadingConfig struct {
+	Enabled      bool   `yaml:"enabled,omitempty" json:"enabled,omitempty"`
+	ThreadByDate string `yaml:"thread_by_date,omitempty" json:"thread_by_date,omitempty"`
 }
 
 // UnmarshalYAML implements the yaml.Unmarshaler interface.
@@ -337,6 +344,15 @@ func (c *EmailConfig) UnmarshalYAML(unmarshal func(any) error) error {
 		normalizedHeaders[normalized] = v
 	}
 	c.Headers = normalizedHeaders
+
+	if c.Threading.Enabled {
+		if _, ok := normalizedHeaders["References"]; ok {
+			return errors.New("conflicting configuration: threading.enabled conflicts with custom References header")
+		}
+		if _, ok := normalizedHeaders["In-Reply-To"]; ok {
+			return errors.New("conflicting configuration: threading.enabled conflicts with custom In-Reply-To header")
+		}
+	}
 
 	return nil
 }
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -992,6 +992,15 @@ tls_config:
 # Further headers email header key/value pairs. Overrides any headers
 # previously set by the notification implementation.
 [ headers: { <string>: <tmpl_string>, ... } ]
+
+# Email threading configuration.
+threading:
+  # Whether to enable threading, which makes alert notifications in the same
+  # alert group show up in the same email thread.
+  [ enabled: <boolean> | default = false ]
+  # What granularity of current date to thread by. Accepted values: daily, none.
+  # (none means group by alert group key, no date).
+  [ thread_by_date: <string> | default = daily ]
 ```
 
 ### `<mattermost_config>`
--- a/notify/email/email.go
+++ b/notify/email/email.go
@@ -265,6 +265,31 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 		fmt.Fprintf(buffer, "Message-Id: %s\r\n", fmt.Sprintf("<%d.%d@%s>", time.Now().UnixNano(), rand.Uint64(), n.hostname))
 	}
 
+	if n.conf.Threading.Enabled {
+		key, err := notify.ExtractGroupKey(ctx)
+		if err != nil {
+			return false, err
+		}
+		// Add threading headers. All notifications for the same alert group
+		// (identified by key hash) are threaded together.
+		threadBy := ""
+		if n.conf.Threading.ThreadByDate != "none" {
+			// ThreadByDate is 'daily':
+			// Use current date so all mails for this alert today thread together.
+			threadBy = time.Now().Format("2006-01-02")
+		}
+		keyHash := key.Hash()
+		if len(keyHash) > 16 {
+			keyHash = keyHash[:16]
+		}
+		// The thread root ID is a Message-ID that doesn't correspond to
+		// any actual email. Email clients following the (commonly used) JWZ
+		// algorithm will create a dummy container to group these messages.
+		threadRootID := fmt.Sprintf("<alert-%s-%s@alertmanager>", keyHash, threadBy)
+		fmt.Fprintf(buffer, "References: %s\r\n", threadRootID)
+		fmt.Fprintf(buffer, "In-Reply-To: %s\r\n", threadRootID)
+	}
+
 	multipartBuffer := &bytes.Buffer{}
 	multipartWriter := multipart.NewWriter(multipartBuffer)
 
--- a/notify/email/email_test.go
+++ b/notify/email/email_test.go
@@ -52,6 +52,7 @@ import (
 	"gopkg.in/yaml.v2"
 
 	"github.com/prometheus/alertmanager/config"
+	"github.com/prometheus/alertmanager/notify"
 	"github.com/prometheus/alertmanager/template"
 	"github.com/prometheus/alertmanager/types"
 )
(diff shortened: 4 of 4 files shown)


notify/email/email.go
```diff
@@ -265,6 +265,19 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 		fmt.Fprintf(buffer, "Message-Id: %s\r\n", fmt.Sprintf("<%d.%d@%s>", time.Now().UnixNano(), rand.Uint64(), n.hostname))
 	}
 
+	if n.conf.PhantomThreading && len(as) > 0 {
+		// Add threading headers. All notifications for the same alert
+		// (identified by fingerprint) on the same day are threaded together.
+		// The thread root ID is a phantom Message-ID that doesn't correspond to
```
sysadmind: I assume this is where the name "Phantom Threading" comes from? I'm not sure people will understand that based on the name. Maybe "daily threading" would make more sense, although in that case what if we want to do a different threading strategy in the future?

notify/email/email.go
```diff
@@ -385,3 +398,18 @@ func (n *Email) getPassword() (string, error) {
...
 	return string(n.conf.AuthPassword), nil
 }
+
+func generateThreadRootID(alerts []*types.Alert, hostname string) string {
+	if len(alerts) == 0 {
+		return ""
+	}
+
+	// Use first alert as representative of the alert group.
+	alert := alerts[0]
```
sysadmind: Doesn't that mean even if we have 30 alerts, whatever the first one is, that's what will be used for fingerprinting? What if that alert resolves, but the other 29 remain?

stapelberg:
Thank you both for your review! Answers inline:

> 1. Did you consider using the `GroupKey` has to source for the thread root id? If I understand the rest of this change correct, this would result in one thread per group. It's also consistent with how we typically group/dedup notifications. For example, the PagerDuty integration uses the hash of the group key as the PagerDuty dedup key: https://github.com/prometheus/alertmanager/blob/main/notify/pagerduty/pagerduty.go#L241. This would also make threading configured directly by routing config.

Thanks, that’s a great tip! Done.

> 2. It looks like we end up just setting some headers on the email. Is this something that could be implemented by templated header? I'm also wondering if this has any strange interaction with the user's config if they set headers.

I prototyped this idea, and if we expose the shortened GroupKeyHash and n.hostname to the template, then users would be able to configure a templated header like so:

```
References: <alert-{{ .GroupKeyHash }}-{{ range $idx, $alert := .Alerts }}{{ if (eq $idx 0) }}{{ $alert.StartsAt.Format \"2006-01-02\" }}{{ end }}{{ end }}@{{ .Hostname }}>
In-Reply-To: <alert-{{ .GroupKeyHash }}-{{ range $idx, $alert := .Alerts }}{{ if (eq $idx 0) }}{{ $alert.StartsAt.Format \"2006-01-02\" }}{{ end }}{{ end }}@{{ .Hostname }}>
```

…but that’s pretty complicated for a user. 

My preference would be to stick with the high-level configuration option (phantom threading enabled/disabled). If we really think it’s required, we can make the date configurable, but realistically, I don’t see one-thread-per-month as an option that anyone would want. Maybe one-thread-per-calendar-week? But that seems rather unconventional, too.

> 3. The comment mentions email clients that use "`(commonly used) JWZ`" - do you know how this change interacts with email clients that don't behave that way?

The worst that can happen is that email clients keep threading the way they currently do (i.e. no threading for alerts). Email clients must cope with non-existing references (common case: somebody adds you to an email thread, so you don’t have the earlier messages).

notify/email/email.go
stapelberg: Yeah, this is now obsolete thanks to @Spaceman1701’s suggestion to use the (hashed) GroupKey.

notify/email/email.go
stapelberg: No, the term “phantom threading” is a creation of my own. I am open to a better suggestion. If we decide that “daily” is the only granularity we want, daily threading would be fine with me.

Spaceman1701: what if we just call it `threading`? Or if that's too concrete, maybe `attempt_threading`

stapelberg: Done.

docs/configuration.md
```diff
@@ -992,6 +992,13 @@ tls_config:
 # Further headers email header key/value pairs. Overrides any headers
 # previously set by the notification implementation.
 [ headers: { <string>: <tmpl_string>, ... } ]
+
+# Email threading configuration.
+threading:
+  # Whether to enable threading, which results in one thread per day
```
Spaceman1701: I think this text needs another pass since `thread_by_date` is nominally configurable.

I also think we should include support for `thread_by_date: none` in this PR. This would support users who want threading by-group, but not threading by day. Sorry if that wasn't clear in my last review!

notify/email/email.go
```diff
@@ -385,3 +402,12 @@ func (n *Email) getPassword() (string, error) {
 	}
 	return string(n.conf.AuthPassword), nil
 }
+
+func generateThreadRootID(keyHash, hostname string) string {
+	// Use current date so all mails for this alert today thread together.
+	date := time.Now().Format("2006-01-02")
+	if len(keyHash) > 16 {
+		keyHash = keyHash[:16]
```
Spaceman1701: is there a reason this needs to be limited to 16 runes?

notify/email/email.go
```diff
@@ -265,6 +265,23 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
...
+		if err != nil {
+			return false, err
+		}
+		// Add threading headers. All notifications for the same alert
+		// (identified by fingerprint) on the same day are threaded together.
+		// The thread root ID is a Message-ID that doesn't correspond to
+		// any actual email. Email clients following the (commonly used) JWZ
+		// algorithm will create a dummy container to group these messages.
+		threadRootID := generateThreadRootID(key.Hash(), n.hostname)
+		if threadRootID != "" {
```
Spaceman1701: I don't think this condition is possible - the `Sprintf` call cannot return an empty string.

stapelberg: Indeed! This was a left-over from an older version of the PR. Now removed.

notify/email/email.go
stapelberg: There’s no reason that the hash needs to be limited to 16 runes specifically, but I do think it is good style to limit it. https://datatracker.ietf.org/doc/html/rfc5322#section-2.1.1 recommends a line length limit of 78 characters, so with `In-Reply-To: <>` taking 15 characters, that leaves us with 64 characters for the hash, date, and hostname. 

The full key.Hash() is a SHA256 sum displayed in hex (%x), which is 64 characters in length. So that consumes the entire available space already 😅

With a limit of 16, the resulting header looks like:

```
In-Reply-To: <alert-d86a4f96117781b7-2025-11-24@ax52>
```

Which I think is a reasonable compromise between leaving enough digits of the hash for the ID to be unique, and keeping it short enough to keep the ID readable and below the 78 character line length limit.

docs/configuration.md
stapelberg: Oh, I see! Done.

notify/email/email.go
```diff
@@ -265,6 +265,31 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
...
+		}
+		keyHash := key.Hash()
+		if len(keyHash) > 16 {
+			keyHash = keyHash[:16]
+		}
+		// The thread root ID is a Message-ID that doesn't correspond to
+		// any actual email. Email clients following the (commonly used) JWZ
+		// algorithm will create a dummy container to group these messages.
+		threadRootID := fmt.Sprintf("<alert-%s-%s@%s>", keyHash, threadBy, n.hostname)
+		fmt.Fprintf(buffer, "References: %s\r\n", threadRootID)
```
Spaceman1701, line 289: Ah, sorry I didn't notice this before, but the docs for the email integration currently say:
```
# Further headers email header key/value pairs. Overrides any headers
# previously set by the notification implementation.
[ headers: { [<string>](https://prometheus.io/docs/alerting/latest/configuration/#string): [<tmpl_string>](https://prometheus.io/docs/alerting/latest/configuration/#tmpl_string), ... } ]
```

Specifically, we need to do something about:
> Overrides any headers previously set by the notification implementation.

I'm not a fan of how this is implemented elsewhere - It looks like the logic that sets the `Message-Id` header does this via checking if `n.conf.Headers` contains `Message-Id` before setting the value, but there's no check happening when we write user headers... This makes it easy to accidentally introduce bugs.

I also don't think it's appropriate to silently let a user configured value override this behavior - it is a configuration error to set both `threading.enabled` and a custom `References` or `In-Reply-To` header.

In my opinion, the right thing to do is to add a check to the email config's `UnmarshalYAML` implementation which raises an error if threading is enabled and either of these headers is set: https://github.com/prometheus/alertmanager/blob/e542c68a6bc1a2f3ad187cb6b4db8dd925816520/config/notifiers.go#L331-L339

What to do you think?

stapelberg: Done. Verified it works:

> Checking '/nix/store/smkp4z11mzcp1kjx4acqfjkq9y6cd4df-checked-config'  FAILED: conflicting configuration: threading.enabled conflicts with custom References header

notify/email/email.go
```diff
@@ -265,6 +265,31 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
...
+			threadBy = time.Now().Format("2006-01-02")
+		}
+		keyHash := key.Hash()
+		if len(keyHash) > 16 {
+			keyHash = keyHash[:16]
+		}
+		// The thread root ID is a Message-ID that doesn't correspond to
+		// any actual email. Email clients following the (commonly used) JWZ
+		// algorithm will create a dummy container to group these messages.
+		threadRootID := fmt.Sprintf("<alert-%s-%s@%s>", keyHash, threadBy, n.hostname)
```
Spaceman1701: ah, sorry again - caught on more thing! On re-read I noticed that `n.hostname` is included in the `threadRootID` - I don't think Alertmanager's hostname should be included in this. When in cluster mode, any Alertmanager node could send the notification. As written, we risk creating one-thread-per-node in the cluster. This shouldn't happen, as the selection of the node which sends the notification should be transparent to the user.

All this is to say that the `threadRootID` should be identical regardless of the node which computes it.

config/notifiers.go
```diff
@@ -338,6 +345,15 @@ func (c *EmailConfig) UnmarshalYAML(unmarshal func(any) error) error {
 	}
 	c.Headers = normalizedHeaders
 
+	if c.Threading.Enabled {
+		if _, ok := normalizedHeaders["References"]; ok {
+			return errors.New("conflicting configuration: threading.enabled conflicts with custom References header")
+		}
+		if _, ok := normalizedHeaders["In-Reply-To"]; ok {
+			return errors.New("conflicting configuration: threading.enabled conflicts with custom In-Reply-to header")
```
Spaceman1701: tiny nitpick
```suggestion
			return errors.New("conflicting configuration: threading.enabled conflicts with custom In-Reply-To header")
```

notify/email/email.go
```diff
@@ -265,6 +265,31 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 		fmt.Fprintf(buffer, "Message-Id: %s\r\n", fmt.Sprintf("<%d.%d@%s>", time.Now().UnixNano(), rand.Uint64(), n.hostname))
 	}
 
+	if n.conf.Threading.Enabled && len(as) > 0 {
```
Spaceman1701: small nitpick, but `len(as) > 0` is guaranteed to be true, I don't think we need to check it. The code in this block doesn't index into `as` anyway, so I think this condition should be removed.

notify/email/email.go
stapelberg: What else can we put there? We need a Message-ID-shaped value (something@something), and we need a value that is unique per deployment (per cluster, I suppose?).

Spaceman1701: Unfortunately, there's no mechanism to name alertmanager clusters right now, so there's no obvious value to pick. A couple options I can think of:
1. just set in config, defaulting to `alertmanager` (easy)
2. using the domain name from whatever the user put in `from` (a little harder, and makes `from` required for `threading`)
3. adding an optional flag, `--cluster.name` and wiring that through  (harder)

Since `threading` is creating a "fake" id, I like (1) and (3) the most. There are wider implications of adding a `--cluster.name`, so I lean towards (1). RFC 5322 [is pretty relaxed](https://www.rfc-editor.org/rfc/rfc5322#section-3.6.4) about what can go on the right side of the `@`, so hopefully (?) implementations will be tolerant of everything allowed there.

What do you think? I'm open to any of these options, or other options if you have ideas.

stapelberg: I was about to say that I don’t like defaulting to `alertmanager`, because then your alerts will be grouped together if you receive alerts from two different alertmanager instances (e.g. a work and a personal setup).

But then it occurred to me that the alert group key would *also* need to be identical, and that seems extremely unlikely because usually one has alert labels that identify the target job. 

So perhaps just hard-coding `alertmanager` is good enough after all. Done.

Spaceman1701 (approved):
Thanks for making those changes, this looks good to me!

TheMeier:
Mabye we want some kind of testing?

stapelberg:
> Mabye we want some kind of testing?

Added a test.

SuperQ (approved):

fauust:
:heart: this feature!! Thanks a lot @stapelberg !

notify/email/email.go
```diff
@@ -265,6 +265,31 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
...
 
+	if n.conf.Threading.Enabled {
+		key, err := notify.ExtractGroupKey(ctx)
+		if err != nil {
+			return false, err
+		}
+		// Add threading headers. All notifications for the same alert group
+		// (identified by key hash) are threaded together.
+		threadBy := ""
+		if n.conf.Threading.ThreadByDate != "none" {
```
GhostLyrics, line 276: @stapelberg @SuperQ @Spaceman1701 
I'm sorry to comment on a MR that's already been merged, but reading the added documentation here as well the implementation, it looks to me like it's quite easy to shoot yourself in the foot as a user.

Consider the following scenario:
- user sees in existing configuration 
  ```yaml
  threading:
    enabled: true
    thread_by_date: daily
  ```
- user assumes this is configurable (because in my opinion this strongly suggests it is) and sets it to "hourly" or "monthly"
- alertmanager will silently accept the parameter but will still deliver them **daily** because every non-None value triggers the daily behaviour. See the line from the MR below.
  ```go
  if n.conf.Threading.ThreadByDate != "none" {...}
  ```

Spaceman1701: Good catch, we should probably patch this so that the user sees a configuration error if it's set to anything other than `none` or `daily`.

PR #4627, opened 2025-10-23
Remove obsolete CircleCI config

SuperQ:
Builds have been stopped in CircleCI, we can remove the config.


sysadmind:
Oh can we just remove the whole file? I saw Prometheus still has it, so I did the same. https://github.com/prometheus/prometheus/blob/main/.circleci/config.yml

SuperQ:
You can, but you first have to manually shutdown the build pipeline in CircleCI's UI. So making a noop first is needed.

I think nobody every bothered to do the same for prometheus.

I did the same for Prom and opened a PR. https://github.com/prometheus/prometheus/pull/17392

sysadmind (approved):

PR #4628, opened 2025-10-23
feat: add names to inhibit rules

siavashs:
- add Name field to config.InhibitRule
- add Name field to inhibit.InhibitRule
- update docs for inhibit_rule

The name will be used in new metrics to be added in a separate change.

### DIFF
--- a/inhibit/inhibit.go
+++ b/inhibit/inhibit.go
@@ -153,6 +153,8 @@ func (ih *Inhibitor) Mutes(lset model.LabelSet) bool {
 // from sending notifications if their meaning is logically a subset of a
 // higher-level alert.
 type InhibitRule struct {
+	// Name is an optional name for the inhibition rule.
+	Name string
 	// The set of Filters which define the group of source alerts (which inhibit
 	// the target alerts).
 	SourceMatchers labels.Matchers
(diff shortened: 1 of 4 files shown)


inhibit/inhibit.go
```diff
@@ -227,17 +233,33 @@ func NewInhibitRule(cr config.InhibitRule) *InhibitRule {
...
+	}
+
 	rule.scache.SetGCCallback(rule.gcCallback)
 
 	return rule
 }
 
+// fingerprint returns the fingerprint of the inhibit rule.
+func (r *InhibitRule) fingerprint() model.Fingerprint {
+	sum := xxhash.New()
```
SuperQ: Since this is only loaded at config load time, do we need to use a third party library here?

~~What's wrong with using `model.FingerprintFromString()`?~~

Oops, that function is not what I was thinking.

Bah, it looks like the common/model hash function is not public.

siavashs: Decided to drop the generated names for now, instead the unnamed rules will have an empty label value when we add the metrics.

MichaHoffmann (approved):
LGTM! Empty name by default makes sense to me too. Especially if we want to use this for metrics eventually.

PR #4629, opened 2025-10-24
feat(inhibit): add inhibition metrics

siavashs:
Add metrics for inhibitor:
- alertmanager_inhibitor_source_alerts_cache_items
- alertmanager_inhibitor_source_alerts_index_items
- alertmanager_inhibitor_mutes_duration_seconds

Add metrics for inhibition rules:
- alertmanager_inhibit_rule_source_alerts_cache_items
- alertmanager_inhibit_rule_source_alerts_index_items
- alertmanager_inhibit_rule_matches_duration_seconds
- alertmanager_inhibit_rule_mutes_duration_seconds

Other changes:
- Add debug logs for duplicate inhibition rule names
- Add Len() method to store.Alerts struct
- Add Len() method to inhibit.index struct
- updated docs

### DIFF
--- a/inhibit/inhibit.go
+++ b/inhibit/inhibit.go
@@ -20,6 +20,7 @@ import (
 	"time"
 
 	"github.com/oklog/run"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/common/model"
 
 	"github.com/prometheus/alertmanager/config"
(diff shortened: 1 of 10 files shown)


inhibit/metric.go
```diff
@@ -0,0 +1,132 @@
...
+	m.sourceAlertsCacheSize.With(prometheus.Labels{"rule": "all"}).Set(float64(size))
+}
+
+func (m *InhibitorMetrics) ObserveSourceAlertsIndexSize(size int) {
+	m.sourceAlertsIndexSize.With(prometheus.Labels{"rule": "all"}).Set(float64(size))
+}
+
+func (m *InhibitorMetrics) ObserveRuleMutesDuration(rule, muted string, duration time.Duration) {
+	m.ruleMuteDuration.With(prometheus.Labels{"rule": rule, "muted": muted}).Observe(duration.Seconds())
+}
```
SuperQ: I dislike these kinds of one-line helper functions. The `prometheus.Labels{}` here is actually removing the struct safety by relying on ordered input to `ObserveRuleMutesDuration()`.

Please just call `.Observe()` where it's needed.

siavashs: Updated, also separate `inhibitor` and `rule` metrics.

inhibit/inhibit.go
```diff
@@ -33,25 +34,37 @@ import (
...
+		logger:  logger,
+		metrics: metrics,
 	}
-	for _, cr := range rs {
-		r := NewInhibitRule(cr)
+
+	ruleNames := make(map[string]struct{})
+	for i, cr := range rs {
+		if _, ok := ruleNames[cr.Name]; ok {
+			ih.logger.Warn("duplicate inhibition rule name", "index", i, "name", cr.Name)
```
MichaHoffmann: I wonder if we should reject the config in this case, or at least increment a metric so operators can alert on it maybe?

siavashs: Rule names are optional and duplicate names only affect the metrics.

SuperQ: I think `Debug()` is fine here. Names are optional hints, mostly for metrics and maybe UI. Side note, we should probably document this desired behavior. So duplicate names shouldn't affect operations.

siavashs: Added `alertmanager_inhibitor_duplicate_rule_names`

siavashs: Just saw the comment from Ben, I'll add debug logs and documentation.

siavashs: Added debug log and updated config docs.

inhibit/metric.go
```diff
@@ -0,0 +1,124 @@
...
+	ruleMatchesDuration       *prometheus.SummaryVec
+	ruleMutesDuration         *prometheus.SummaryVec
+}
+
+// NewInhibitorMetrics returns a new InhibitorMetrics.
+func NewInhibitorMetrics(reg prometheus.Registerer) *InhibitorMetrics {
+	metrics := &InhibitorMetrics{
+		sourceAlertsCacheSize: prometheus.NewGauge(
+			prometheus.GaugeOpts{
+				Name: "alertmanager_inhibitor_source_alerts_cache_size",
```
SuperQ: Size is a bit vague. Is this the number of entries in the cache?

```suggestion
				Name: "alertmanager_inhibitor_source_alerts_cache_entries",
```

siavashs: Updated.

SuperQ (approved):

PR #4633, opened 2025-10-27
fix(notify): use worker routines for notifications

siavashs:
This change reduces the number of waiting routines in notify pipeline.
It adds long running worker routines per receiver integrations.
Workers receive notification requests from a queue to process and send.


siavashs:
Closing this as we will try different approaches planned after v1.0

PR #4634, opened 2025-10-27
feat(notify): make mute stage honor send_resolved

siavashs:
Make mute stage honor `send_resolved` in receivers. This fixes cases where an alert was already sent to a receiver before being silenced.
Sending resolved notifications for a now silenced alert helps update the receiver's status.
For example an opened PagerDuty incident will be closed.

Related #226

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -541,8 +541,10 @@ const (
 
 // MuteStage filters alerts through a Muter.
 type MuteStage struct {
-	muter   types.Muter
-	metrics *Metrics
+	muter           types.Muter
+	metrics         *Metrics
+	notificationLog NotificationLog
+	receivers       map[string][]Integration
 }
 
 // NewMuteStage return a new MuteStage.
@@ -550,8 +552,26 @@ func NewMuteStage(m types.Muter, metrics *Metrics) *MuteStage {
 	return &MuteStage{muter: m, metrics: metrics}
 }
 
+// NewMuteStageWithSendResolved returns a new MuteStage that honors send_resolved
+// for silenced alerts. This should be used for silence.Silencer to allow resolved
+// notifications for previously notified alerts.
+func NewMuteStageWithSendResolved(m types.Muter, notificationLog NotificationLog, receivers map[string][]Integration, metrics *Metrics) *MuteStage {
+	return &MuteStage{
+		muter:           m,
+		metrics:         metrics,
+		notificationLog: notificationLog,
+		receivers:       receivers,
+	}
+}
+
 // Exec implements the Stage interface.
 func (n *MuteStage) Exec(ctx context.Context, logger *slog.Logger, alerts ...*types.Alert) (context.Context, []*types.Alert, error) {
+	// If this is a silencer with send_resolved support, use the enhanced logic
+	if _, isSilencer := n.muter.(*silence.Silencer); isSilencer && n.notificationLog != nil && n.receivers != nil {
+		return n.execWithSendResolved(ctx, logger, alerts...)
+	}
+
+	// Standard muting logic for inhibitors or silencers without send_resolved support
 	var (
 		filtered []*types.Alert
 		muted    []*types.Alert
@@ -578,6 +598,114 @@ func (n *MuteStage) Exec(ctx context.Context, logger *slog.Logger, alerts ...*ty
 		}
 		n.metrics.numNotificationSuppressedTotal.WithLabelValues(reason).Add(float64(len(muted)))
 		logger.Debug("Notifications will not be sent for muted alerts", "alerts", fmt.Sprintf("%v", muted), "reason", reason)
+	}
+
+	return ctx, filtered, nil
+}
+
+// execWithSendResolved implements enhanced silence filtering that honors send_resolved config.
+// It allows resolved notifications for silenced alerts if the alert was previously notified
+// and the receiver has send_resolved enabled.
+func (n *MuteStage) execWithSendResolved(ctx context.Context, logger *slog.Logger, alerts ...*types.Alert) (context.Context, []*types.Alert, error) {
+	var (
+		filtered []*types.Alert
+		muted    []*types.Alert
+	)
+
+	// Get the receiver name from context
+	receiverName, ok := ReceiverName(ctx)
+	if !ok {
+		return ctx, nil, errors.New("receiver name missing")
+	}
+
+	// Get the group key from context
+	groupKey, ok := GroupKey(ctx)
+	if !ok {
+		return ctx, nil, errors.New("group key missing")
+	}
+
+	// Check if any integration for this receiver has send_resolved enabled
+	receiverIntegrations, receiverExists := n.receivers[receiverName]
+	if !receiverExists {
+		// Receiver not found, fall back to standard behavior
+		for _, a := range alerts {
+			if n.muter.Mutes(a.Labels) {
+				muted = append(muted, a)
+			} else {
+				filtered = append(filtered, a)
+			}
+		}
+		if len(muted) > 0 {
+			n.metrics.numNotificationSuppressedTotal.WithLabelValues(SuppressedReasonSilence).Add(float64(len(muted)))
+			logger.Debug("Notifications will not be sent for silenced alerts", "alerts", fmt.Sprintf("%v", muted), "reason", SuppressedReasonSilence)
+		}
+		return ctx, filtered, nil
+	}
+
+	// Check if any integration has send_resolved enabled
+	hasSendResolved := false
+	for _, integration := range receiverIntegrations {
+		if integration.SendResolved() {
+			hasSendResolved = true
+			break
+		}
+	}
+
+	// Process each alert
+	for _, a := range alerts {
+		if !n.muter.Mutes(a.Labels) {
+			// Alert is not silenced, let it through
+			filtered = append(filtered, a)
+			continue
+		}
+
+		// Alert is silenced
+		if !a.Resolved() || !hasSendResolved {
+			// Alert is firing or receiver doesn't have send_resolved, filter it out
+			muted = append(muted, a)
+			continue
+		}
+
+		// Alert is resolved and receiver has send_resolved enabled
+		// Check if this alert was previously notified
+		wasNotified := false
+		for _, integration := range receiverIntegrations {
+			if !integration.SendResolved() {
+				continue
+			}
+
+			recv := &nflogpb.Receiver{
+				GroupName:   receiverName,
+				Integration: integration.Name(),
+				Idx:         uint32(integration.Index()),
+			}
+
+			entries, err := n.notificationLog.Query(nflog.QGroupKey(groupKey), nflog.QReceiver(recv))
+			if err != nil && !errors.Is(err, nflog.ErrNotFound) {
+				logger.Warn("Failed to query notification log for silenced resolved alert", "alert", a.Name(), "err", err)
+				continue
+			}
+
+			if len(entries) > 0 && len(entries[0].FiringAlerts) > 0 {
+				// This alert was previously notified as firing for this integration
+				wasNotified = true
+				break
+			}
+		}
+
+		if wasNotified {
+			// Alert was previously notified and is now resolved, let it through for resolved notification
+			filtered = append(filtered, a)
+			logger.Debug("Allowing resolved notification for silenced alert", "alert", a.Name())
+		} else {
+			// Alert was not previously notified, filter it out
+			muted = append(muted, a)
+		}
+	}
+
+	if len(muted) > 0 {
+		n.metrics.numNotificationSuppressedTotal.WithLabelValues(SuppressedReasonSilence).Add(float64(len(muted)))
+		logger.Debug("Notifications will not be sent for silenced alerts", "alerts", fmt.Sprintf("%v", muted), "reason", SuppressedReasonSilence)
 	}
 
 	return ctx, filtered, nil
--- a/notify/notify_test.go
+++ b/notify/notify_test.go
@@ -823,6 +823,204 @@ func TestMuteStageWithSilences(t *testing.T) {
 	}
 }
 
+func TestMuteStageWithSendResolved(t *testing.T) {
+	// Create silences
+	silences, err := silence.New(silence.Options{Retention: time.Hour})
+	require.NoError(t, err)
+
+	sil := &silencepb.Silence{
+		EndsAt:   utcNow().Add(time.Hour),
+		Matchers: []*silencepb.Matcher{{Name: "mute", Pattern: "me"}},
+	}
+	require.NoError(t, silences.Set(sil))
+
+	reg := prometheus.NewRegistry()
+	marker := types.NewMarker(reg)
+	silencer := silence.NewSilencer(silences, marker, promslog.NewNopLogger())
+
+	// Create notification log
+	nflog, err := nflog.New(nflog.Options{
+		Retention: time.Hour,
+		Logger:    promslog.NewNopLogger(),
+		Metrics:   reg,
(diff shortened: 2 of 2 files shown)


SuperQ:
Nice. I think the only real issue is that this is a reasonably impactful behavior change. So we should probably put it behind a flag as to not surprise users.

grobinson-grafana:
Thanks for doing all this work. If possible, I think we should explore making this behavior a first class citizen of silences instead of having silences inherit `send_resolved` from downstream integrations.

The main argument for this is that it keeps silences predictable.

When we inherit behavior from downstream integrations, the the same silence now behaves differently in different aggregation groups, as its behaviour changes based on which integration is used.

This is likely going to be really really confusing for operators as some silenced alerts will now send resolved notifications while others will not, or even the same silenced alert sends a resolved notification in one aggregation group but not another.

By making this a property of the silence instead we can make it an explicit choice of the silence author, and it becomes auditable via the history of expired silences.

What do you think?

siavashs:
Sounds good to me, let me try and add this.

notify/notify.go
```diff
@@ -541,17 +541,37 @@ const (
...
 
+// NewMuteStageWithSendResolved returns a new MuteStage that honors send_resolved
+// for silenced alerts. This should be used for silence.Silencer to allow resolved
+// notifications for previously notified alerts.
+func NewMuteStageWithSendResolved(m types.Muter, notificationLog NotificationLog, receivers map[string][]Integration, metrics *Metrics) *MuteStage {
+	return &MuteStage{
+		muter:           m,
+		metrics:         metrics,
+		notificationLog: notificationLog,
+		receivers:       receivers,
```
Spaceman1701, line 563: One problem I see here is that we need to decide if we should `send_resolved` for all integrations when we make the decision in the mute stage. What if we changed the position of the `MuteStage` in the pipeline so that we get one stage per-integration? I think this would be a very simple change (just move construction of this stage into `createReceiverStage`.

In my mind, that simplifies things a little bit by explicitly making the `MuteStage` per-integration.

Spaceman1701 (commented):
I think this is a great idea! We have some hacky behavior in our internal fork that approximates what this PR does. 

I actually think it's very useful for the Silence's `send_resolve` behavior to be tied to each integration. My main argument for this is that most integrations talk to stateful services, and it's really up to the integration whether resolved messages make sense. When they do make sense, I think it's much better if we make an effort to always deliver them.

The easiest example of this is pagerduty: when alertmanager sends a notification to pagerduty, it creates an incident. The incident isn't resolved until alertmanager sends a resolved notification. If a new alert is sent to pagerduty from the same group (e.g. with the same group_key), *pagerduty will not notify a human again*.

Today, the behavior of alertmanager silences interacts with this very poorly. If a firing alert gets silenced, it will never send a resolved notification to pagerduty and thus will never close the incident. This means that a future alert in the same group doesn't page until the incident is manually closed. This is such a problem that we run a separate service which periodically closes stale pagerduty incidents for our users. 

I think this same argument can apply to most integrations, even if it's not a machine keeping track of state like in pagerduty. If users configure their MSTeams/slack/email integrations with `send_resolved`, it makes sense to always send those resolved messages, regardless of the silence itself.

notify/notify.go
grobinson-grafana: This is one of the reasons I propose we make this a first class citizen of silences instead. It removes a lot of the ambiguity while still allowing integrations to make their own decisions for resolved notifications.

Spaceman1701: That's fair, though I'm still not sure I agree. Just to make sure I'm clear on the behavior, are you proposing that:
```
Silence.send_resolved: true
Integration.send_resolved: true
-> resolved notifications for silenced alerts (that already notified)
```
```
Silence.send_resolved: false
Integration.send_resolved: true
-> No resolved notifications for silenced alerts
```
```
Silence.send_resolved: true
Integration.send_resolved: false
-> No resolved notifications for silenced alerts
```

I'd supposed I'd be happy with this behavior, though I'd probably disallow non-send-resolve silences via tooling. That's all right, and it's good to give users the option both ways.

I guess the question here is whether it's more confusing for some silences to allow resolved notifications or for some integrations to always send resolved notifications, regardless of alert state. 

My biggest concern is still for integrations that maintain state, like pagerduty. I'm pretty firmly of the belief that it's always wrong to skip sending a resolved notification to pagerduty because it'll cause the incident to become stale. This prevents future alerts from notifying a human. So I wouldn't want a user of our alertmanager cluster to accidentally create a silence that blocks send_resolved to pagerduty. 

But really, I'd be happy enough with either version of this behavior... I really want this change 😄

Spaceman1701: Oh and addressing the technical comment: Whether or not this is a flag on the silence, we'd need to adjust the notification pipeline so that the `MuteStage` is applied per-integration (since `send_resolved` is a setting on the integration, not the receiver). 

That should actually be as simple as moving the stage further into the pipeline. There's some performance impact, but I think it's pretty minimal. The notification pipeline isn't a big resource user and integration pipelines can happen concurrently.

grobinson-grafana: > Just to make sure I'm clear on the behavior, are you proposing that...

Yep, exactly this!

> I guess the question here is whether it's more confusing for some silences to allow resolved notifications or for some integrations to always send resolved notifications, regardless of alert state.

Far more confusing for integrations to influence the behavior of silences (and if we extend this to inhibition rules, the behavior of inhibition rules too).

> Whether or not this is a flag on the silence, we'd need to adjust the notification pipeline so that the MuteStage is applied per-integration

If we make this a flag on the silence, I think we avoid having to execute the `MuteStage` for each integration. Instead we can execute it for each aggregation group which is a big difference, since silences and inhibitions are computationally expensive.

Spaceman1701: >Far more confusing for integrations to influence the behavior of silences (and if we extend this to inhibition rules, the behavior of inhibition rules too).

For me, honestly, it's the opposite that's confusing. However, this is just a matter of opinion at the end of the day, so I'm ok with either. As long as it's all clearly documented, I'll be happy. 

I do quite like that your proposed behavior will give operators a choice, rather than make a mandatory behavior change. 

> If we make this a flag on the silence, I think we avoid having to execute the MuteStage for each integration. Instead we can execute it for each aggregation group which is a big difference, since silences and inhibitions are computationally expensive.

I guess this is true, but I think it has some potential for subtle issues, unless I'm missing something (which is certainly possible 😄 ). Is the thinking here is that the `RetryStage` will handle `send_resolved` per-integration, so that the `MuteStage` can just return resolved alerts whenever any integration has `send_resolved: true`? And then let the `RetryStage`s for each integration handle them correctly per-integration? 

Since this change only sends resolved notifications for silenced alerts for alerts which already notified and the nflog is per-integration, this really has to be queried per-integration as well. It is true that we generally expect the nflog for each integration on an aggrGroup to contain the same messages, but this isn't necessarily always the case - for example, after a config reload the receiver could have a new integration that never sent the message originally. Or, if the integration couldn't reach it's target (e.g. slack is down), a notification may not have been sent to that integration. 

These are definitely edge cases. We could choose to allow misbehavior here in the name of simplicity or performance, but we should be clear that that's the decision we're making.

In terms of performance, I don't think we have to worry too much about the notification pipeline. Internally, we've split the `MuteStage` into a `CalculateMuteStage` and an `ApplyMuteStage` which separate calculating which alerts are muted and actually filtering them. The `CalculateMuteStage` is per-aggrGroup and the `ApplyMuteStage` is per-integration. However, we found that this was a premature optimization - more detailed profiling has shown that the performance of the notification pipeline has little impact of the rest of alertmanager. This makes some sense since the notification pipeline doesn't hold any of the important locks, spends most of its time waiting on IO, and runs concurrently. 

Regardless, if we're worried about performance impacts, we shouldn't trust my anecdotes. We should do some benchmarking and profiling. I'd be hesitant to let our concerns about performance guide the behavior of any feature without more data.

grobinson-grafana: > Is the thinking here is that the RetryStage will handle send_resolved per-integration, so that the MuteStage can just return resolved alerts whenever any integration has send_resolved: true? And then let the RetryStages for each integration handle them correctly per-integration?

Not quite.

What I'm proposing is each silence has its own bool `send_resolved`.

When a flush happens, the alerts are passed through the mute stage just like normal.

During the execution of the mute stage, it now takes this new field into consideration such that we have the following behaviour:

1. If an alert is not silenced it is passed to the next stage.
2. If an alert is silenced, and active, it is filtered out.
3. If an alert is silenced, and resolved:
  a. If `send_resolved` is false, it is filtered out
  b. If `send_resolved` is true, it is passed to the next stage

3b is the new behaviour, and is equivalent to the silence not existing.

What then happens is any resolved alerts pass through the subsequent stages until each integrations `send_resolved` stages are considered in the DedupStage.

But importantly this requires zero code changes other than to the MuteStage, and also importantly, it means the MuteStage does not need to be aware of receivers or hold a reference to the nflog.

Spaceman1701: That is much less complex! It's actually very similar to our internal fork's hack (except we control it via an `always_send_resolved` flag on the integration). 

However, I don't think `DedupStage` ever checks that a resolved alert actually sent a non-resolved notification. It only checks if the resolved alert already sent a resolved notification. This means that silences with `send_resolved` will send resolved notifications for alerts that have been silenced for their entire lifecycle, which is much less useful.

The other problem is that the `DedupStage` currently passes groups that only contain resolved alerts (because we need to write the nflog entry for these groups whether or not we notify). Since the algorithm here will filter active alerts but not muted alerts for `send_resolved` silences, that'll probably cause some problems as well.

I think we could probably change the logic of the `DebupStage` to accommodate the new behavior, but it's not immediately obvious to me that it's better than checking the nflog in the `MuteStage`.

SuperQ:
@Spaceman1701 Yup, PagerDuty incidents is exactly the reason I raised this issue in the past.

notify/notify.go
grobinson-grafana: > However, I don't think DedupStage ever checks that a resolved alert actually sent a non-resolved notification

Doesn't that happen [here](https://github.com/prometheus/alertmanager/blob/main/notify/notify.go#L680)?

There is an edge case if you have a non previously notified resolved alert grouped with other active alerts, in which case the non previously notified resolved alert will be included in the notification. 

My answer to this edge case is that it already happens in the general case when you don't even use silences. I think the fix for that needs to be done in the DedupStage for it to cover all edge cases, and we can do it separately from this effort?

> because we need to write the nflog entry for these groups whether or not we notify

I'm not following this edge case, can you share an example?

grobinson-grafana: Just to add, whatever we can do to keep the MuteStage clean is my preferred option. I really want to avoid two "DedupStages" where some nflog deduplication happens in the MuteStage and the rest of the nflog deduplication happens in the DedupStage. I would really like to keep the responsibilities separate.

Spaceman1701:
> @Spaceman1701 Yup, PagerDuty incidents is exactly the reason I raised this issue in the past.

@SuperQ 👍 this is why my preference is for alertmanager's contract to be "For integrations with `send_resolved: true`, If we sent a notification, we will send a resolved notification"

grobinson-grafana:
> > @Spaceman1701 Yup, PagerDuty incidents is exactly the reason I raised this issue in the past.
> 
> @SuperQ 👍 this is why my preference is for alertmanager's contract to be "For integrations with `send_resolved: true`, If we sent a notification, we will send a resolved notification"

I think if that's a goal it may be worth a wider discussion in an issue as we also would need to address resolved notifications for these additional cases:

1. Muted by an inhibition rule. Example: you have a warning alert that fired first, then a critical alert. The critical alert inhibits the warning. Both resolved at the same time, but only the critical sends a resolved notification.


3. Muted by an active/mute time window. The alert fires outside the window, and then resolves within the window.

4. Alertmanager crashes/restarts just before a long `group_interval` flushes (i.e. multi hour group intervals).

I can see that being a wider engineering effort that yourself and @siavashs could lead if you would like to!

notify/notify.go
Spaceman1701: >Doesn't that happen [here](https://github.com/prometheus/alertmanager/blob/main/notify/notify.go#L680)?
> There is an edge case if you have a non previously notified resolved alert grouped with other active alerts, in which case the non previously notified resolved alert will be included in the notification.

Ah yeah, kinda. But as you say, this only stops the notification if the group did not have any firing alerts in the last during the last notification attempt. If it did, we'd send resolutions for all the alerts, even the ones which were not in the original group. It also has an edge case going to opposite direction - if some alert in the group is not silenced, it'll cause us to pass forward all silenced resolved alerts even if we otherwise shouldn't have.

> I think the fix for that needs to be done in the DedupStage for it to cover all edge cases, and we can do it separately from this effort?

But perhaps you're right, this is already a problem in some cases today. We'll make it worse since it'll now happen for any group that has resolutions while it's silenced by a `send_resolved: true` silence. We could adjust the `DedupStage` so that it only passes resolved alerts that are actually a subset of the previous notification's firing alerts set. I think that'd solve the problem in the general case.

> I'm not following this edge case, can you share an example?

This is part of what's handled by the same condition here https://github.com/prometheus/alertmanager/blob/main/notify/notify.go#L679

In order to re-notify after all alerts in a group are resolved, we need to write an nflog entry indicating that alerts have resolved. Otherwise, we need to wait for the nflog entry to expire (which could take a very long time).

The other half of the logic is in `RetryStage`: https://github.com/prometheus/alertmanager/blob/main/notify/notify.go#L790-L805. The key part being https://github.com/prometheus/alertmanager/blob/main/notify/notify.go#L798. This causes the `SetNotifiesStage` to see the 'resolution' message as sent and record it in the nflog. So this is all part of the same set of edge cases we're discussing. 

Imagine this scenario:
1. Alert `alertname=x, instance=a` fires and sends a notification
3. User creates a silence with `send_resolved: true` for `alertname=x`
4. Alert `alertname=x, instance=b` fires. No notification is sent because of the silence
5. Alert `alertname=x, instance=b` resolves. Since the silence has `send_resolved: true`, the the user will get a resolved message. Since the message is filtered by the `MuteStage` to just resolved alerts, it will write an nflog entry that contains only resolved alerts. If the receiver is pagerduty, the incident will be closed by the resolution message.
6. The user lifts the silence
7. Since `alertname=x, instance=a` is still firing and the last nflog entry contains no firing alerts, it sends a duplicate notification on the next flush

This particular case would be avoided if the `DedupStage` checked that the resolved alerts its forwarding are actually a subset of the previous firing alerts. 

I actually think if we re-ordered the steps we could recreate this behavior for any of the proposed implementations... So that's not ideal. The problem here is that we don't record which alerts were muted in the nflog. The next flush isn't able to distinguish between "some alerts were resolved, but not all of them" and "all alerts were resolved". 

Maybe we could work around that by limiting the `send_resolved` silence behavior to groups that _only_ contain resolved alerts? This way we would only send a resolved notification for a silenced group if everything in the group is resolved. That's actually closer to the behavior for unsilenced groups since we only send resolved messages when the whole group is resolved.

Spaceman1701: There are versions of this sequence that can happen without `send_resolved` silences as well, but I think `send_resolved` makes it more likely to happen...

The `DedupStage` logic is actually pretty fragile when silences partially mute alerts...

Spaceman1701:
This topic came up again during the maintainers mini-meetup (@SuperQ this is the PR I was referring to). 

I think we need to take a step back, define a contract for notification behavior that includes all possible states, and go from there. I'm pretty interested in sorting this out, so I'll take a crack at a proposal in a new issue.

SoloJacobs:
@siavashs Seems like we can close the PR itself. Let's revisit the topic once we have a clear vision, what the correct approach is.

PR #4636, opened 2025-10-28
feat(cluster): persistant peer names

siavashs:
Currently the default behaviour is to generate a new, random peer name on every startup.
It might be desired to have a persistent peer name in consistent setups like running the cluster on bare-metal nodes.

This change adds an optional cli arg that allows specifying a custom peer name.
If no peer name is specified a random name is generated.

Duplicated peer names are detected, logged and increment the new metric `alertmanager_cluster_peer_name_conflicts_total`.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Prometheus Team
+// Copyright The Prometheus Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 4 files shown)


TheMeier:
Maybe add at least one testcase in cluster/cluster_test.go that does pass something else then "" as name arg into Create() ?

siavashs:
> Maybe add at least one testcase in cluster/cluster_test.go that does pass something else then "" as name arg into Create() ?

Updated, ~also implemented readable random names~ 🙂 
Removed the peer host resolution for now, will do it in a separate PR.

cluster/cluster.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2018 Prometheus Team
+// Copyright Prometheus Authors
```
SuperQ:
```suggestion
// Copyright The Prometheus Authors
```

PR #4670, opened 2025-10-30
reduce the time Dispatch.Group holds the mutex

Spaceman1701:
`Groups` calls can take a long time when there are many aggrGroups or when on of the filter functions is slow. Right now, `Groups` holds the `Dispatcher` lock for the entire duration of `Groups`. This is aggravated by the fact that the API passes filter functions which themselves call `Silences.Mutes` and `Inhibits.Mutes` which themselves hold locks. 

Since the `Dispatcher` needs to hold a write lock on `Dispatcher.mtx` in order to ingest alerts, `Groups` calls essentially block alert ingestion.  Since `Groups` depends on `Silences.Mutes`, this also means that calls to `Silences.Mutes` can block ingestion. Since that blocks all the various `Silences` API endpoints _and_ some of the gossip channels, this becomes. big knot of locks which causes the alertmanager to hang up if something is hammering `GET /alerts/groups`. Unfortunately, many dashboard services do just that.

This patch just copies the `aggrGroupsPerRoute` map out of the dispatcher and then releases the lock for the rest of the `Groups` call. This ensures that we never need to hold the dispatcher lock and the silencer or inhibitor locks at the same time. 

We've been running this patch in production for quite a while now. We've found that performance is substantially improved, especially around startup time (when the silencer/inhibitor are both extra slow). We've also measured much less mutex contention after adding this.

I don't have any synthetic benchmarks for this one unfortunately.

Here's some profiling comparisons of an alertmanager restart before and after the patch:
Screenshot 2024-12-23 at 4 09 06 PM
Screenshot 2024-12-23 at 4 02 40 PM

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"maps"
 	"sort"
 	"sync"
 	"time"
@@ -240,11 +241,7 @@ func (d *Dispatcher) Groups(routeFilter func(*Route) bool, alertFilter func(*typ
 		// have a mutex protecting their internal state.
 		// The aggrGroup methods use the internal lock. It is important to avoid
 		// accessing internal fields on the aggrGroup objects.
-		copiedMap := map[model.Fingerprint]*aggrGroup{}
-		for fp, ag := range ags {
-			copiedMap[fp] = ag
-		}
-		aggrGroupsPerRoute[route] = copiedMap
+		aggrGroupsPerRoute[route] = maps.Clone(ags)
 	}
 	d.mtx.RUnlock()
 


dispatch/dispatch.go
```diff
@@ -225,15 +228,23 @@ func (d *Dispatcher) Groups(routeFilter func(*Route) bool, alertFilter func(*typ
 	groups := AlertGroups{}
 
 	d.mtx.RLock()
-	defer d.mtx.RUnlock()
+	aggrGroupsPerRoute := map[*Route]map[model.Fingerprint]*aggrGroup{}
+	for route, ags := range d.aggrGroupsPerRoute {
+		copiedMap := map[model.Fingerprint]*aggrGroup{}
```
rajagopalanand: I think since aggrGroup{} is a pointer, if the underlying aggr changes after copy, could it potentially lead to unintended behavior? I am wondering if we need to add more concurrent tests to ensure correctness.

Spaceman1701: This is a good thing to point out, but I believe it's safe as written. `aggrGroup`'s internal state is protected by a mutex and it is meant to be safe for concurrent use.

This method already uses the mutex-protected methods (named, `alerts := ag.alerts.List()` on line 261), so it should be ok.

grobinson-grafana: I was looking at this too because I had the same thought and I believe you do not access any fields that are mutated by other gourtoines (since these fields are only mutated when the aggregation group is initialized). 

It would be easy though for someone to make a similar mistake in future with other fields, so perhaps a comment in the code explaining something similar to what I wrote would be valuable to future contributors.

Spaceman1701: Sure, how does the new comment look to you all? 

I could also break this out into a `Dispatch.snapshot` method, if you think that'd make things more clear.

siavashs: How about moving AGs to a sub-package and only expose specific read/write methods?
This way we block direct access to mutating fields and avoid potential issues in future.

siavashs: Also I dislike the fact that we do API payload related manipulations within dispatcher, we should just return a snapshot of the state and let API package do whatever it wants with it (with write access removed of coarse with special readonly Interfaces).

Spaceman1701: @siavashs, I think moving the aggrGroup into a subpackage is a good idea. Maybe we can do that in a follow up change? There's enough ambiguities we'd have to decide that it's probably warranted. For example: what do we do with the existing AlertGroups type? It's meant to be the "public" version of aggrGroup, but moving aggrGroup into a subpackage makes it public.

Same goes for changing the location for the logic, I think it's a good idea, but I'd rather try that in a follow up.

There's already plenty of logic is `dispatch.go` which works with aggrGroups, so it's not like this change introduces a new footgun around concurrent access.

siavashs: Agreed, let's move forward with the current setup.
We can look into the sub-package + read/write interfaces later (I can look into it this week or next).

grobinson-grafana: I am happy to expose read/write methods, I think it's a good idea. I don't think you need a separate sub-package as the abstraction level is dispatching. Using Go modules as a proxy for `private/public/protected` fields is an anti-pattern though, I would avoid going down this rabbit hole.

grobinson-grafana: I am happy to expose read/write methods, but I don't think we need a separate sub-package here as the abstraction level is dispatching. Using modules as a proxy for private/public/protected fields is bad Go because then every struct would live in its own sub-module.

siavashs: Here is a quick draft on top of this PR https://github.com/prometheus/alertmanager/pull/4679

dispatch/dispatch.go
```diff
@@ -227,16 +227,34 @@ func (ag AlertGroups) Len() int { return len(ag) }
...
+		// Since other goroutines could modify d.aggrGroupsPerRoute, we need to
+		// copy it. We DON'T need to copy the aggrGroup objects because they each
+		// have a mutex protecting their internal state.
+		// The aggrGroup methods use the internal lock. It is important to avoid
+		// accessing internal fields on the aggrGroup objects.
+		copiedMap := map[model.Fingerprint]*aggrGroup{}
+		for fp, ag := range ags {
+			copiedMap[fp] = ag
+		}
+		aggrGroupsPerRoute[route] = copiedMap
```
ultrotter: Could you do:
aggrGroupsPerRoute[route] = maps.Clone(ags)
?

PR #4679, opened 2025-11-03
feat(dispatch): expose readonly interfaces

siavashs:
Groups() method in dispatcher does extensive data manipulation to
staisfy the API. Instead it can return a snapshot of its internal state
and let the API handle the data processing.

- expose readonly interfaces for AggregationGroup and Route
- move Groups logic out of dispatcher and into the api handler

### DIFF
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Prometheus Team
+// Copyright The Prometheus Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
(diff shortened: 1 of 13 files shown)


api/v2/api.go
```diff
@@ -405,11 +405,68 @@ func (api *API) getAlertGroupsHandler(params alertgroup_ops.GetAlertGroupsParams
...
 	af := api.alertFilter(matchers, *params.Silenced, *params.Inhibited, *params.Active)
-	alertGroups, allReceivers := api.alertGroups(rf, af)
+	dispatchGroups := api.dispatchGroups(rf, af)
 
-	res := make(open_api_models.AlertGroups, 0, len(alertGroups))
+	groups := AlertGroups{}
+	// Keep a list of receivers for an alert to prevent checking each alert
+	// again against all routes. The alert has already matched against this
+	// route on ingestion.
+	receivers := map[prometheus_model.Fingerprint][]string{}
```
ultrotter: Should we preallocate the size?

PR #4693, opened 2025-11-05
Add parallel benchmarks to silence_bench_test

ultrotter:
Add benchmarks about parallel queries, parallel queries in concurrency with adds, and about running mutes.

These will be then used to compare improvements to the silence module.


SuperQ:
Looks like this is a noop now with #4694

PR #4695, opened 2025-11-05
compile silence matchers when the silence is added

ultrotter:
If QMatches doesn't update the map, multiple queries can execute
    concurrently, with just a read lock. We have had this in production for
    over a year, and it significantly improves performance for cases where
    many silences need to be processed against many alerts.

    This makes load time a bit slower, but a snapshot file with about
    120.000 silences can still load in about 2 seconds, so we can accept
    this, in exchange for the ability to run queries in parallel.

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -215,19 +215,20 @@ type Limits struct {
 type MaintenanceFunc func() (int64, error)
 
 type metrics struct {
-	gcDuration                prometheus.Summary
-	snapshotDuration          prometheus.Summary
-	snapshotSize              prometheus.Gauge
-	queriesTotal              prometheus.Counter
-	queryErrorsTotal          prometheus.Counter
-	queryDuration             prometheus.Histogram
-	silencesActive            prometheus.GaugeFunc
-	silencesPending           prometheus.GaugeFunc
-	silencesExpired           prometheus.GaugeFunc
-	propagatedMessagesTotal   prometheus.Counter
-	maintenanceTotal          prometheus.Counter
-	maintenanceErrorsTotal    prometheus.Counter
-	matcherCompileErrorsTotal *prometheus.CounterVec
+	gcDuration                            prometheus.Summary
+	snapshotDuration                      prometheus.Summary
+	snapshotSize                          prometheus.Gauge
+	queriesTotal                          prometheus.Counter
+	queryErrorsTotal                      prometheus.Counter
+	queryDuration                         prometheus.Histogram
+	silencesActive                        prometheus.GaugeFunc
+	silencesPending                       prometheus.GaugeFunc
+	silencesExpired                       prometheus.GaugeFunc
+	propagatedMessagesTotal               prometheus.Counter
+	maintenanceTotal                      prometheus.Counter
+	maintenanceErrorsTotal                prometheus.Counter
+	matcherCompileCacheSilenceErrorsTotal prometheus.Counter
+	matcherCompileLoadSnapshotErrorsTotal prometheus.Counter
 }
 
 func newSilenceMetricByState(s *Silences, st types.SilenceState) prometheus.GaugeFunc {
@@ -272,13 +273,15 @@ func newMetrics(r prometheus.Registerer, s *Silences) *metrics {
 		Name: "alertmanager_silences_maintenance_errors_total",
 		Help: "How many maintenances were executed for silences that failed.",
 	})
-	m.matcherCompileErrorsTotal = prometheus.NewCounterVec(
+	matcherCompileErrorsTotal := prometheus.NewCounterVec(
 		prometheus.CounterOpts{
 			Name: "alertmanager_silences_matcher_compile_errors_total",
 			Help: "How many silence matcher compilations failed.",
 		},
 		[]string{"stage"},
 	)
+	m.matcherCompileCacheSilenceErrorsTotal = matcherCompileErrorsTotal.WithLabelValues("cache_silence")
+	m.matcherCompileLoadSnapshotErrorsTotal = matcherCompileErrorsTotal.WithLabelValues("load_snapshot")
 	m.queriesTotal = prometheus.NewCounter(prometheus.CounterOpts{
 		Name: "alertmanager_silences_queries_total",
 		Help: "How many silence queries were received.",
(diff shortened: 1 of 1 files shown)


silence/silence.go
```diff
@@ -51,14 +51,12 @@ var ErrInvalidState = errors.New("invalid state")
 
 type matcherCache map[string]labels.Matchers
 
-// Get retrieves the matchers for a given silence. If it is a missed cache
-// access, it compiles and adds the matchers of the requested silence to the
-// cache.
+// Get retrieves the matchers for a given silence.
```
SuperQ: This seems like it's a breaking change to the behavior of the public function.

I think it would be better to have a new non-adding function like `Check` that had different behavior.

ultrotter: We can add a new function and then change all usage to that function if you prefer... 

But all in all it shouldn't change the behavior in any meaningful way, given that if the silence is correct, it is already in the matcher cache, and thus the matchers are returned, and if it was incorrect, it did not get added, and the user gets still an error.

Also is there any user of this method, besides QMatches in this module? Arguably the matcherCache is an internal of Silences, which is the actual interface, so we can also make this method private in case?

silence/silence.go
```diff
@@ -562,6 +560,11 @@ func (s *Silences) checkSizeLimits(msil *pb.MeshSilence) error {
 	return nil
 }
 
+func (s *Silences) silenceAdded(sil *pb.Silence) {
+	s.version++
+	s.mc.add(sil)
```
siavashs: No error checks here, also how should this recover on error?

silence/silence.go
SuperQ: Yea, making it private seems like the right thing to do.

SuperQ:
Looks like this needs a rebase.

silence/silence.go
```diff
@@ -562,6 +560,11 @@ func (s *Silences) checkSizeLimits(msil *pb.MeshSilence) error {
 	return nil
 }
 
+func (s *Silences) silenceAdded(sil *pb.Silence) {
```
siavashs: This reads weird, should we use a better name?
It can also become a callback.

silence/silence.go
ultrotter: done thanks, and addressed the error cases

silence/silence.go
ultrotter: Ah thanks! I actually fixed this already in the latest version!

silence/silence.go
ultrotter: How/why/when would you like to make it a callback? As in a callback to pass to merge?
And what should we call it? handleAddedSilence? .... Something else?

ultrotter:
> Looks like this needs a rebase.

Done!

silence/silence.go
siavashs: We could but not necessary, just throwing an idea out there.

But regarding naming, maybe we can call it `addCache()` or `appendCache()`? because that's what it does.

rajagopalanand: `cacheSilence()` or `cacheNewSilence()` might also be options to consider

rajagopalanand (approved):
Overall LGTM. Added just a nit comment

silence/silence.go
ultrotter: Fixed thanks!

ultrotter:
> Overall LGTM. Added just a nit comment

Thanks! What's the comment? I only saw the suggestion for the function name, which I apply, but I can't see anything else... Is there another one? Thanks!

rajagopalanand:
> > Overall LGTM. Added just a nit comment
> 
> Thanks! What's the comment? I only saw the suggestion for the function name, which I apply, but I can't see anything else... Is there another one? Thanks!

The function name suggestion was the nit comment

SuperQ (approved):

silence/silence.go
```diff
@@ -217,18 +215,19 @@ type Limits struct {
...
+	queriesTotal              prometheus.Counter
+	queryErrorsTotal          prometheus.Counter
+	queryDuration             prometheus.Histogram
+	silencesActive            prometheus.GaugeFunc
+	silencesPending           prometheus.GaugeFunc
+	silencesExpired           prometheus.GaugeFunc
+	propagatedMessagesTotal   prometheus.Counter
+	maintenanceTotal          prometheus.Counter
+	maintenanceErrorsTotal    prometheus.Counter
+	matcherCompileErrorsTotal *prometheus.CounterVec
```
SuperQ: Minor suggestion.

In the struct, declare two `prometheus.Counter()` values.

Then in the `newMetrics()` declare the `NewCounterVec()` and assign the label values.

silence/silence.go
```diff
@@ -273,6 +272,13 @@ func newMetrics(r prometheus.Registerer, s *Silences) *metrics {
 		Name: "alertmanager_silences_maintenance_errors_total",
 		Help: "How many maintenances were executed for silences that failed.",
 	})
+	m.matcherCompileErrorsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "alertmanager_silences_matcher_compile_errors_total",
+			Help: "How many silence matcher compilations failed.",
+		},
+		[]string{"stage"},
+	)
```
SuperQ:
```suggestion
	matcherCompileErrorsTotal := prometheus.NewCounterVec(
		prometheus.CounterOpts{
			Name: "alertmanager_silences_matcher_compile_errors_total",
			Help: "How many silence matcher compilations failed.",
		},
		[]string{"stage"},
	)
	m.matcherCompileCacheSilenceErrorsTotal = matcherCompileErrorsTotal.WithLabelValues("cache_silence")
	m.matcherCompileLoadSnapshotErrorsTotal = matcherCompileErrorsTotal.WithLabelValues("load_snapshot")
	
```

Doing it this way will cache the `WithLabelValues()` function. A small performance win.

It will also make sure we always init the 0 value when the metrics are created.

ultrotter: Thank you! Done! Sorry, I hadn't realized there was a button to commit the suggestion... Next time I guess!

SuperQ: Well, my suggested code was incomplete anyway. :)

PR #4696, opened 2025-11-05
Remove duplicate slice during Silences query

ultrotter:
If there are a lot of silences Silences.query, for queries without
    q.ids, first copies all their pointers to a slice, causing a lot of
    allocations, and then does it again to a new shorter slice filtered by
    the query. We avoid this by only adding to the result slice in case of a
    match. This creates savings for each queries, scaling with the number of silences.


silence/silence.go
```diff
@@ -833,40 +836,37 @@ func (s *Silences) query(q *query, now time.Time) ([]*pb.Silence, int, error) {
...
-				return nil, s.version, err
-			}
-			if !ok {
-				remove = true
-				break
-			}
+	return res, s.version, nil
+}
+
+func (s *Silences) appendIfFiltersMatch(res *[]*pb.Silence, sil *pb.Silence, q *query, now time.Time) error {
```
siavashs: What do you think about the following?

- define a new type for `[]*pb.Silence`
- add this method to that type
- pass `s` as an argument to that method

for example:
```suggestion
type result []*pb.Silence

func (r *result) appendOnMatch(s *Silences, sil *pb.Silence, q *query, now time.Time) error {
```

Spaceman1701: If the concern is the mutation of the `res` argument, I think we could just write

```suggestion
func (s *Silences) appendIfFiltersMatch(res []*pb.Silence, sil *pb.Silence, q *query, now time.Time) ([]*pb.Silence, error) {
    for _, f := range q.filters {
		ok, err := f(sil, s, now)
		if err != nil {
			return res, err
		}
		if !ok {
			return res, nil
		}
	}
	return append(res, cloneSilence(sil)), nil
}
```

Spaceman1701: And this would be called like

```go
var err error
res, err = appendIfFiltersMatch(res, sil, now)
if err != nil {
  // whatever
}
```

ultrotter: I tried going for @Spaceman1701's approach to avoid defining a new internal type with a one method just for this... Hopefully it's as clear!

silence/silence.go
```diff
@@ -834,45 +834,56 @@ func (s *Silences) CountState(states ...types.SilenceState) (int, error) {
...
 		}
 	}
 
-	return resf, s.version, nil
+	return res, s.version, nil
+}
+
+// appendIfFiltersMatch appends the given silence to the result set
+// if it matches all filters in the query. In case of a filter error, the error is returned.
+func (s *Silences) appendIfFiltersMatch(res []*pb.Silence, sil *pb.Silence, q *query, now time.Time) ([]*pb.Silence, error) {
```
rajagopalanand: Nit: Consider making this a standalone function as opposed to a method on Silences. This is a utility function to append to a slice and IMO is not a method on Silences

ultrotter: We can change this, but it would still need to take Silences in, be private, and be dependent on being called under sileces rlock, so perhaps it's better to keep it as a method of Silences to make this aspect clear?

ultrotter: On the flip side, I can try with making it a local function inside query, since it's just an internal deduplication helper

silence/silence.go
```diff
@@ -834,45 +834,56 @@ func (s *Silences) CountState(states ...types.SilenceState) (int, error) {
...
+// query executes the given query and returns the resulting silences.
 func (s *Silences) query(q *query, now time.Time) ([]*pb.Silence, int, error) {
-	// If we have no ID constraint, all silences are our base set.  This and
-	// the use of post-filter functions is the trivial solution for now.
 	var res []*pb.Silence
+	var err error
 
+	// appendIfFiltersMatch appends the given silence to the result set
+	// if it matches all filters in the query. In case of a filter error, the error is returned.
+	appendIfFiltersMatch := func(res []*pb.Silence, sil *pb.Silence) ([]*pb.Silence, error) {
```
SuperQ, line 844: Does this need to be an inline func?

ultrotter: Well, there are a lot of options, but all in all, does it matter?
It mostly exists to avoid duplicate code between the code branches below.
Originally it was a private method of Silences, but people didn't like that (see the discussion above) because it didn't do anything on silences, and it was mostly something happening on the result. We could make it a general module level helper function, which takes silences and everything else, but it's unclear whether it's a good thing to do so, as it's called just from one place and it's pretty specific code to this code path and we don't expect it to be needed outside of query.

Given it's private and inline we can always refactor it to the module, or a method, as we change silences, but so far it doesn't seem to hurt? Or what would you say are the counterindications of doing that?

Thanks!

SuperQ: Yea, I guess this is fine for now.

PR #4698, opened 2025-11-05
fix: pprof debug endpoints not working with --web.route-prefix

wbh1:
When using `--web.route-prefix` or `--web.external-url` with a path component, the pprof debug endpoints were not accessible. The handlers were directly passing requests to http.DefaultServeMux without adjusting the URL path to match what the automatic pprof handlers expect.

For example, with `--web.route-prefix=/prometheus/alertmanager`, a request to `/prometheus/alertmanager/debug/pprof/` was being forwarded with the full path intact, but http.DefaultServeMux only has handlers registered for paths starting with `/debug/pprof/`.

This fix reconstructs the request path by extracting the subpath parameter from the route and prepending /debug to it, ensuring the path matches what http.DefaultServeMux expects. The fix also preserves trailing slashes, which are required by some pprof handlers.

Added tests to verify pprof endpoints work correctly both with and without route prefix configuration.

### DIFF
--- a/ui/web_test.go
+++ b/ui/web_test.go
@@ -39,7 +39,7 @@ func TestDebugHandlersWithRoutePrefix(t *testing.T) {
 	router.ServeHTTP(w, req)
 
 	require.Equal(t, http.StatusOK, w.Code)
-	require.NotEqual(t, 0, w.Body.Len())
+	require.Contains(t, w.Body.String(), "/debug/pprof/", "pprof page did not load with expected content when using a route prefix")
 
 	// Test GET request to pprof heap endpoint
 	req = httptest.NewRequest("GET", routePrefix+"/debug/pprof/heap", nil)
@@ -57,4 +57,5 @@ func TestDebugHandlersWithRoutePrefix(t *testing.T) {
 	router2.ServeHTTP(w, req)
 
 	require.Equal(t, http.StatusOK, w.Code)
+	require.Contains(t, w.Body.String(), "/debug/pprof/", "pprof page did not load with expected content")
 }


ui/web.go
```diff
@@ -87,8 +88,24 @@ func Register(r *route.Router, reloadCh chan<- chan error, logger *slog.Logger)
...
+	})
+	r.Post("/debug/*subpath", func(w http.ResponseWriter, req *http.Request) {
+		subpath := route.Param(req.Context(), "subpath")
+		req.URL.Path = path.Join("/debug", subpath)
+		// path.Join removes trailing slashes, but some pprof handlers expect them.
+		if strings.HasSuffix(subpath, "/") && !strings.HasSuffix(req.URL.Path, "/") {
+			req.URL.Path += "/"
+		}
+		http.DefaultServeMux.ServeHTTP(w, req)
+	})
```
siavashs: Let's avoid duplicate code, we can define a function and then use it for both `get` and `post` handlers.

siavashs: @wbh1 can you update the PR to resolve my comment above?

wbh1: Yep! Sorry for the delay - I'm at kubecon this week but I'll try to update it today!

wbh1: Updated and RFR. Thanks @siavashs!

ui/web_test.go
```diff
@@ -0,0 +1,67 @@
+// Copyright 2015 Prometheus Team
```
siavashs: Let's use the new copyright header:
```suggestion
// Copyright The Prometheus Authors
```

Also update `web.go` header please.

ui/web.go
siavashs: > Yep! Sorry for the delay - I'm at kubecon this week but I'll try to update it today!

No rush, enjoy the conference.

ui/web_test.go
```diff
@@ -0,0 +1,67 @@
...
+	Register(router, reloadCh, logger)
+
+	// Test GET request to pprof index (note: pprof index returns text/html)
+	req := httptest.NewRequest("GET", routePrefix+"/debug/pprof/", nil)
+	w := httptest.NewRecorder()
+	router.ServeHTTP(w, req)
+
+	if w.Code != http.StatusOK {
+		t.Errorf("Expected status %d for %s, got %d. Body: %s", http.StatusOK, req.URL.Path, w.Code, w.Body.String())
+	}
```
siavashs: Let's use `github.com/stretchr/testify/require` instead of all the `if` statements in tests.
I expected the linter to catch these but it did not.

wbh1: Updated! Thanks @siavashs. RFR again 🤞🏻

ui/web_test.go
```diff
@@ -0,0 +1,60 @@
...
+	router := route.New().WithPrefix(routePrefix)
+	Register(router, reloadCh, logger)
+
+	// Test GET request to pprof index (note: pprof index returns text/html)
+	req := httptest.NewRequest("GET", routePrefix+"/debug/pprof/", nil)
+	w := httptest.NewRecorder()
+	router.ServeHTTP(w, req)
+
+	require.Equal(t, http.StatusOK, w.Code)
+	require.NotEqual(t, 0, w.Body.Len())
```
siavashs:
```suggestion
	require.Positive(t, w.Body.Len(), "body should not be empty")
```

Alternatively we could replace the checks with https://pkg.go.dev/github.com/stretchr/testify/require#Assertions.HTTPBodyContains
but feel free to skip it.

siavashs: But I think `HTTPBodyContains` might not be suitable here, it is best for custom handlers and not a router.

wbh1: Updated in 7317f2d to use `require.Contains()` for checking the body, since -- as you said -- `HTTPBodyContains` would've required having access to the handlerFunc to pass to the test case.

siavashs:
@SuperQ we can merge this one.

PR #4701, opened 2025-11-05
Fix more flaky tests in TestClusterJoinAndReconnect

SoloJacobs:
As explained in more detail in commit
`88b4e13d09975b61290493f8893bb4fb7796a93f`, the tests in `cluster_test.go` will occasionally fail. The underlying reason for the failure is that the test only waits for `p2` to be ready, but this does not reflect whether `p1` has updated its memberlist.

However, that commit only fixed the test `TestJoinLeave`. I have now ran the other tests in `cluster_test.go` 2400 times, and identified the following problematic tests: `TestSetPeerNames` and `TestTLSConnection`.

Since the underlying faulty code is mostly identical, so is the corresponding fix.

_Note_

For those, whom reviewed my previous pull request: Back then I simply fixed the first failure I found. This time I tried to find all the failures in `cluster_test.go`.

### DIFF
--- a/cluster/cluster_test.go
+++ b/cluster/cluster_test.go
@@ -348,9 +348,9 @@ func testTLSConnection(t *testing.T) {
 	p2.WaitReady(context.Background())
 	require.Equal(t, "ready", p2.Status())
 
-	require.Equal(t, 2, p1.ClusterSize())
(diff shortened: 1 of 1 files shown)


ultrotter:
Looks good, thanks!

SoloJacobs:
Kindly requesting another review @SuperQ

PR #4703, opened 2025-11-06
Rename matcherCache to matcherIndex

ultrotter:
The matcherCache is in fact not a cache.
We never used to evict from it, and now we need to update it when we update the state, as such call it a name that identifies its purpose.

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -281,7 +281,7 @@ func newMetrics(r prometheus.Registerer, s *Silences) *metrics {
 		},
 		[]string{"stage"},
 	)
-	m.matcherCompileIndexSilenceErrorsTotal = matcherCompileErrorsTotal.WithLabelValues("cache_silence")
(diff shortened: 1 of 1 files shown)


silence/silence.go
```diff
@@ -280,7 +280,7 @@ func newMetrics(r prometheus.Registerer, s *Silences) *metrics {
 		},
 		[]string{"stage"},
 	)
-	m.matcherCompileCacheSilenceErrorsTotal = matcherCompileErrorsTotal.WithLabelValues("cache_silence")
+	m.matcherCompileIndexSilenceErrorsTotal = matcherCompileErrorsTotal.WithLabelValues("cache_silence")
```
SuperQ:
```suggestion
	m.matcherCompileIndexSilenceErrorsTotal = matcherCompileErrorsTotal.WithLabelValues("index")
```

PR #4704, opened 2025-11-06
feat(dispatch): add start delay

siavashs:
This change adds a new cmd flag `--dispatch.start-delay` which corresponds to the `--rules.alert.resend-delay` flag in Prometheus.
This flag controls the minimum amount of time that Prometheus waits before resending an alert to Alertmanager.

By adding this value to the start time of Alertmanager, we delay the aggregation groups' first flush, until we are confident all alerts are resent by Prometheus instances.

This should help avoid race conditions in inhibitions after a (re)start.

Other improvements:
- remove hasFlushed flag from aggrGroup
- remove mutex locking from aggrGroup

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -143,6 +143,7 @@ func run() int {
 		maxSilenceSizeBytes         = kingpin.Flag("silences.max-silence-size-bytes", "Maximum silence size in bytes. If negative or zero, no limit is set.").Default("0").Int()
 		alertGCInterval             = kingpin.Flag("alerts.gc-interval", "Interval between alert GC.").Default("30m").Duration()
 		dispatchMaintenanceInterval = kingpin.Flag("dispatch.maintenance-interval", "Interval between maintenance of aggregation groups in the dispatcher.").Default("30s").Duration()
+		DispatchStartDelay          = kingpin.Flag("dispatch.start-delay", "Minimum amount of time to wait before dispatching alerts. This option should be synced with value of --rules.alert.resend-delay on Prometheus.").Default("0s").Duration()
 
 		webConfig      = webflag.AddFlags(kingpin.CommandLine, ":9093")
 		externalURL    = kingpin.Flag("web.external-url", "The URL under which Alertmanager is externally reachable (for example, if Alertmanager is served via a reverse proxy). Used for generating relative and absolute links back to Alertmanager itself. If the URL has a path portion, it will be used to prefix all HTTP endpoints served by Alertmanager. If omitted, relevant URL components will be derived automatically.").String()
@@ -186,6 +187,8 @@ func run() int {
 	logger := promslog.New(&promslogConfig)
 
 	logger.Info("Starting Alertmanager", "version", version.Info())
+	startTime := time.Now()
+
 	logger.Info("Build context", "build_context", version.BuildContext())
 
 	ff, err := featurecontrol.NewFlags(logger, *featureFlags)
@@ -493,7 +496,17 @@ func run() int {
 			silencer.Mutes(labels)
 		})
 
-		disp = dispatch.NewDispatcher(alerts, routes, pipeline, marker, timeoutFunc, *dispatchMaintenanceInterval, nil, logger, dispMetrics)
+		disp = dispatch.NewDispatcher(
+			alerts,
+			routes,
+			pipeline,
+			marker,
+			timeoutFunc,
+			*dispatchMaintenanceInterval,
+			nil,
+			logger,
+			dispMetrics,
+		)
 		routes.Walk(func(r *dispatch.Route) {
 			if r.RouteOpts.RepeatInterval > *retention {
 				configLogger.Warn(
@@ -520,7 +533,7 @@ func run() int {
 			}
 		})
 
-		go disp.Run()
+		go disp.Run(startTime.Add(*DispatchStartDelay))
 		go inhibitor.Run()
 
 		return nil
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -1,4 +1,4 @@
-// Copyright 2018 Prometheus Team
+// Copyright The Prometheus Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -20,6 +20,7 @@ import (
 	"log/slog"
 	"sort"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -30,6 +31,12 @@ import (
 	"github.com/prometheus/alertmanager/provider"
 	"github.com/prometheus/alertmanager/store"
 	"github.com/prometheus/alertmanager/types"
+)
+
+const (
+	DispatcherStateUnknown = iota
+	DispatcherStateWaitingToStart
+	DispatcherStateRunning
 )
 
 // DispatcherMetrics represents metrics associated to a dispatcher.
@@ -90,6 +97,9 @@ type Dispatcher struct {
 	cancel              func()
 
 	logger *slog.Logger
+
+	startTimer *time.Timer
+	state      int
 }
 
 // Limits describes limits used by Dispatcher.
@@ -102,39 +112,44 @@ type Limits interface {
 
 // NewDispatcher returns a new Dispatcher.
 func NewDispatcher(
-	ap provider.Alerts,
-	r *Route,
-	s notify.Stage,
-	mk types.GroupMarker,
-	to func(time.Duration) time.Duration,
-	mi time.Duration,
-	lim Limits,
-	l *slog.Logger,
-	m *DispatcherMetrics,
+	alerts provider.Alerts,
+	route *Route,
+	stage notify.Stage,
+	marker types.GroupMarker,
+	timeout func(time.Duration) time.Duration,
+	maintenanceInterval time.Duration,
+	limits Limits,
+	logger *slog.Logger,
+	metrics *DispatcherMetrics,
 ) *Dispatcher {
-	if lim == nil {
-		lim = nilLimits{}
+	if limits == nil {
+		limits = nilLimits{}
 	}
 
 	disp := &Dispatcher{
-		alerts:              ap,
-		stage:               s,
-		route:               r,
-		marker:              mk,
-		timeout:             to,
-		maintenanceInterval: mi,
-		logger:              l.With("component", "dispatcher"),
-		metrics:             m,
-		limits:              lim,
+		alerts:              alerts,
+		stage:               stage,
+		route:               route,
+		marker:              marker,
+		timeout:             timeout,
+		maintenanceInterval: maintenanceInterval,
+		logger:              logger.With("component", "dispatcher"),
+		metrics:             metrics,
+		limits:              limits,
+		state:               DispatcherStateUnknown,
 	}
 	return disp
 }
 
 // Run starts dispatching alerts incoming via the updates channel.
-func (d *Dispatcher) Run() {
+func (d *Dispatcher) Run(dispatchStartTime time.Time) {
 	d.done = make(chan struct{})
 
 	d.mtx.Lock()
+	d.logger.Debug("preparing to start", "startTime", dispatchStartTime)
+	d.startTimer = time.NewTimer(time.Until(dispatchStartTime))
+	d.state = DispatcherStateWaitingToStart
+	d.logger.Debug("setting state", "state", "waiting_to_start")
 	d.aggrGroupsPerRoute = map[*Route]map[model.Fingerprint]*aggrGroup{}
 	d.aggrGroupsNum = 0
 	d.metrics.aggrGroups.Set(0)
@@ -175,6 +190,18 @@ func (d *Dispatcher) run(it provider.AlertIterator) {
 				d.processAlert(alert, r)
 			}
 			d.metrics.processingDuration.Observe(time.Since(now).Seconds())
+
+		case <-d.startTimer.C:
+			if d.state == DispatcherStateWaitingToStart {
+				d.state = DispatcherStateRunning
+				d.logger.Debug("started", "state", "running")
+				d.logger.Debug("Starting all existing aggregation groups")
+				for _, groups := range d.aggrGroupsPerRoute {
+					for _, ag := range groups {
+						d.runAG(ag)
+					}
+				}
+			}
 
 		case <-maintenance.C:
 			d.doMaintenance()
@@ -311,6 +338,7 @@ type notifyFunc func(context.Context, ...*types.Alert) bool
 // processAlert determines in which aggregation group the alert falls
 // and inserts it.
 func (d *Dispatcher) processAlert(alert *types.Alert, route *Route) {
+	now := time.Now()
 	groupLabels := getGroupLabels(alert, route)
 
 	fp := groupLabels.Fingerprint()
@@ -347,6 +375,30 @@ func (d *Dispatcher) processAlert(alert *types.Alert, route *Route) {
 	// alert is already there.
 	ag.insert(alert)
 
+	if alert.StartsAt.Add(ag.opts.GroupWait).Before(now) {
+		ag.logger.Debug(
+			"Alert is old enough for immediate flush, resetting timer to zero",
+			"alert", alert.Name(),
+			"fingerprint", alert.Fingerprint(),
+			"startsAt", alert.StartsAt,
+		)
+		ag.resetTimer(0)
+	}
+	// Check dispatcher and alert state to determine if we should run the AG now.
+	switch d.state {
+	case DispatcherStateWaitingToStart:
+		d.logger.Debug("Dispatcher still waiting to start")
+	case DispatcherStateRunning:
+		d.runAG(ag)
+	default:
+		d.logger.Warn("unknown state detected", "state", "unknown")
+	}
+}
+
+func (d *Dispatcher) runAG(ag *aggrGroup) {
+	if ag.running.Load() {
+		return
+	}
 	go ag.run(func(ctx context.Context, alerts ...*types.Alert) bool {
 		_, _, err := d.stage.Exec(ctx, d.logger, alerts...)
 		if err != nil {
@@ -392,13 +444,18 @@ type aggrGroup struct {
 	done    chan struct{}
 	next    *time.Timer
 	timeout func(time.Duration) time.Duration
-
-	mtx        sync.RWMutex
-	hasFlushed bool
+	running atomic.Bool
 }
 
 // newAggrGroup returns a new aggregation group.
-func newAggrGroup(ctx context.Context, labels model.LabelSet, r *Route, to func(time.Duration) time.Duration, marker types.AlertMarker, logger *slog.Logger) *aggrGroup {
+func newAggrGroup(
+	ctx context.Context,
+	labels model.LabelSet,
+	r *Route,
+	to func(time.Duration) time.Duration,
+	marker types.AlertMarker,
+	logger *slog.Logger,
+) *aggrGroup {
 	if to == nil {
 		to = func(d time.Duration) time.Duration { return d }
 	}
@@ -436,6 +493,7 @@ func (ag *aggrGroup) String() string {
 }
 
 func (ag *aggrGroup) run(nf notifyFunc) {
+	ag.running.Store(true)
 	defer close(ag.done)
 	defer ag.next.Stop()
 
@@ -462,10 +520,7 @@ func (ag *aggrGroup) run(nf notifyFunc) {
 			ctx = notify.WithRouteID(ctx, ag.routeID)
 
 			// Wait the configured interval before calling flush again.
-			ag.mtx.Lock()
-			ag.next.Reset(ag.opts.GroupInterval)
-			ag.hasFlushed = true
-			ag.mtx.Unlock()
+			ag.resetTimer(ag.opts.GroupInterval)
 
 			ag.flush(func(alerts ...*types.Alert) bool {
 				return nf(ctx, alerts...)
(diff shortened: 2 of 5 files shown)


config/config.go
```diff
@@ -909,6 +909,7 @@ type Route struct {
 	GroupWait      *model.Duration `yaml:"group_wait,omitempty" json:"group_wait,omitempty"`
 	GroupInterval  *model.Duration `yaml:"group_interval,omitempty" json:"group_interval,omitempty"`
 	RepeatInterval *model.Duration `yaml:"repeat_interval,omitempty" json:"repeat_interval,omitempty"`
+	WaitOnStartup  bool            `yaml:"wait_on_startup" json:"wait_on_startup,omitempty"`
```
ultrotter: Shouldn't this be the default?

dispatch/dispatch.go
```diff
@@ -395,22 +397,32 @@ type aggrGroup struct {
...
-		done:     make(chan struct{}),
+		labels:    labels,
+		routeID:   r.ID(),
+		routeKey:  r.Key(),
+		opts:      &r.RouteOpts,
+		timeout:   to,
+		alerts:    store.NewAlerts(),
+		marker:    marker,
+		done:      make(chan struct{}),
+		startTime: startTime,
```
ultrotter: This is not an attribute of the alert group though, logically? Isn't there a better place we could get it from? Should we really add a pointer in all alertgroups to store a global alertmanager value?

ultrotter (changes requested):

config/config.go
siavashs: I think so too 
The condition in code uses an `or` so not setting this still makes the AG to wait, seems redundant.
cc @alxric

dispatch/dispatch.go
siavashs: Good point, we can add 2 global variables:
- `AlertmanagerStartTime`: for any component to access
- `DispatcherStartTime`: for `dispatcher` and `AG`s

dispatch/dispatch.go
```diff
@@ -486,17 +499,33 @@ func (ag *aggrGroup) stop() {
...
+}
+
+// shouldWaitForGroup checks if AG should wait for group before sending notification.
+func (ag *aggrGroup) shouldWaitForGroup(alert *types.Alert) bool {
+	now := time.Now()
+	return alert.StartsAt.Add(ag.opts.GroupWait).Before(now)
+}
+
+// shouldReset checks if AG timer should reset.
+func (ag *aggrGroup) shouldReset(alert *types.Alert) bool {
```
Spaceman1701: small nitpick: I think `shouldReset` should accept `now` as an argument. This makes things a little easier to test and ensures that the whole method uses the same value for `now`.

dispatch/dispatch.go
```diff
@@ -486,17 +499,33 @@ func (ag *aggrGroup) stop() {
...
 func (ag *aggrGroup) insert(alert *types.Alert) {
 	if err := ag.alerts.Set(alert); err != nil {
 		ag.logger.Error("error on set alert", "err", err)
 	}
-
 	// Immediately trigger a flush if the wait duration for this
 	// alert is already over.
 	ag.mtx.Lock()
 	defer ag.mtx.Unlock()
-	if !ag.hasFlushed && alert.StartsAt.Add(ag.opts.GroupWait).Before(time.Now()) {
```
Spaceman1701, line 499: Somewhat unrelated to this change, but I noticed it when reviewing the new code - I think there's a very minor logic bug here - if an alert's `StartsAt` is in the past, but not at least `ag.opts.GroupWait` in the past, I think we should check if the next flush is before or after it would be scheduled purely from the new alert. If it's after, we should reset the timer to that duration. I don't thin we're keeping track of the next flush time outside of the timer, so that'd need to change too 🤔 

E.g. 

```go
wantedFlush := time.Since(alert.StartsAt.Add(ag.opts.GroupWait))
if wantedFlush < time.Duration(0) {
    wantedFlush = time.Duration(0)
}
actualFlush := ag.durationToNextFlush()
if wantedFlush < actualFlush {
  timer.Reset(wantedFlush)
}
```

I don't think we should change the behavior in this PR though. Perhaps as a follow up.

dispatch/dispatch.go
```diff
@@ -486,17 +499,33 @@ func (ag *aggrGroup) stop() {
...
 }
 
+// shouldWaitOnStartup checks if AG should wait on initial startup before sending notification.
+func (ag *aggrGroup) shouldWaitOnStartup() bool {
+	now := time.Now()
+	return !ag.opts.WaitOnStartup || ag.startTime.Add(ag.opts.GroupWait).Before(now)
+}
+
+// shouldWaitForGroup checks if AG should wait for group before sending notification.
+func (ag *aggrGroup) shouldWaitForGroup(alert *types.Alert) bool {
```
Spaceman1701: One more small nitpick: maybe it's just me, but I think this method name is reversed - it returns `true` if we _should not_ wait for `group_wait`.

juliusv:
> there's no guarantee that `group_wait` is the right duration to wait after a restart.

That's what I was thinking as well: some people may even have a `group_wait: 1d` for low-prio grouped alerts. Then you would never get any alerts if you restarted Alertmanager once a day, right?

siavashs:
I'm dropping the `WaitOnStartup` as we never used it internally and based on the comments it can be tricky if user uses a long `group_wait` value.

config/config.go
siavashs: Dropped this as discussed below.

dispatch/dispatch.go
siavashs: Dropped and moved the logic to dispatcher.

dispatch/dispatch.go
siavashs: This logic is dropped from AG and moved to dispatcher.

dispatch/dispatch.go
siavashs: Good catch, we can add it here or as a follow up.

dispatch/dispatch.go
siavashs: This is removed now.

cmd/alertmanager/main.go
```diff
@@ -491,7 +493,18 @@ func run() int {
...
-		disp = dispatch.NewDispatcher(alerts, routes, pipeline, marker, timeoutFunc, *dispatchMaintenanceInterval, nil, logger, dispMetrics)
+		disp = dispatch.NewDispatcher(
+			alerts,
+			routes,
+			pipeline,
+			marker,
+			timeoutFunc,
+			startTime.Add(*prometheusAlertResendDelay),
+			*dispatchMaintenanceInterval,
+			nil,
```
ultrotter, line 506: Are we using limits, except in that one test? If not, is it worth cleaning it up?

dispatch/dispatch.go
```diff
@@ -102,30 +103,32 @@ type Limits interface {
...
-	mi time.Duration,
-	lim Limits,
-	l *slog.Logger,
-	m *DispatcherMetrics,
+	alerts provider.Alerts,
+	route *Route,
+	stage notify.Stage,
+	marker types.GroupMarker,
+	timeout func(time.Duration) time.Duration,
+	startTime time.Time,
```
ultrotter: This is confusing. One expects startTime to be the start time :) But it's actually the time at which the dispatcher should start dispatching... Maybe let's call it minDispatchTime or something?

dispatch/dispatch.go
```diff
@@ -347,6 +350,15 @@ func (d *Dispatcher) processAlert(alert *types.Alert, route *Route) {
 	// alert is already there.
 	ag.insert(alert)
 
+	// If the alert is old enough, reset the timer to send the notification
+	// immediately.
+	if alert.StartsAt.Add(ag.opts.GroupWait).Before(d.startTime) {
+		// Check if we can start dispatching the alert.
+		if time.Now().After(d.startTime) {
```
ultrotter: Isn't there still a very minimal race condition here, if prometheus sends us alerts after exactly the right time, but we are processing one while the other is waiting for the lock? At minimum we should suggest making the flag time longer than the resend delay in prometheus. If we do so we should probably call it something else? like dispatcher-start-notify-delay, and then explain that it should be set depending on your prometheus resend delay?

dispatch/route.go
```diff
@@ -236,6 +236,9 @@ type RouteOpts struct {
 
 	// A list of time intervals for which the route is active.
 	ActiveTimeIntervals []string
+
+	// Honor the group_wait on initial startup even if incoming alerts are old
+	WaitOnStartup bool
```
ultrotter: Are these still used?

dispatch/dispatch.go
```diff
@@ -347,6 +350,15 @@ func (d *Dispatcher) processAlert(alert *types.Alert, route *Route) {
 	// alert is already there.
 	ag.insert(alert)
 
+	// If the alert is old enough, reset the timer to send the notification
+	// immediately.
+	if alert.StartsAt.Add(ag.opts.GroupWait).Before(d.startTime) {
```
ultrotter: Here we don't check hasFlushed anymore... Is that on purpose? Note that with the previous codepath we avoid time comparisons altogether once that condition/boolean has happened, which is nice.

siavashs:
We are now failing this test which is vague and I remember debugging before but not documenting: https://github.com/prometheus/alertmanager/blob/aa879e10d1a254a51e8c6791c7fcec2b9fa530db/test/with_api_v2/acceptance/send_test.go#L422-L466

dispatch/dispatch.go
ultrotter: Also here, is it worth setting a boolean on the dispatcher like "sendingStarted" and going

if d.sendingStarted || time.Now().After(....)
and then d.sendingStarted = true to avoid keeping checking long before it's useful? Or it it unnecessary as the previous condition would prevent this path?

ultrotter:
> We are now failing this test which is vague and I remember debugging before but not documenting:
> 
> https://github.com/prometheus/alertmanager/blob/aa879e10d1a254a51e8c6791c7fcec2b9fa530db/test/with_api_v2/acceptance/send_test.go#L422-L466

Maybe checking the hasFlushed condition would help this test too? After all that should exactly prevent from notifying twice?

dispatch/dispatch.go
siavashs: Right, the alert still end up in the old dispatcher not the new one which is created after reload.

siavashs: Or am I mistaken?

siavashs:
> Maybe checking the hasFlushed condition would help this test too? After all that should exactly prevent from notifying twice?

So the problem is not duplicate notification but earlier than expected notification:
```
        interval [2,2.5]
        ---
        - &{map[] 0001-01-01T00:00:00.000Z <nil> [] 0001-01-01T00:00:01.000Z <nil> <nil> { map[alertname:test1]}}[-9.223372036854776e+09:]
          [ ✓ ]
        interval [9,9.5]
        ---
        - &{map[] 0001-01-01T00:00:00.000Z <nil> [] 0001-01-01T00:00:01.000Z <nil> <nil> { map[alertname:test1]}}[-9.223372036854776e+09:]
        - &{map[] 0001-01-01T00:00:00.000Z <nil> [] 0001-01-01T00:00:04.000Z <nil> <nil> { map[alertname:test2]}}[-9.223372036854776e+09:]
          [ ✗ ]

        received:
        @ 2.00549375
        - &{map[] 0001-01-01T00:00:00.000Z <nil> [] 2025-11-07T15:47:48.705+01:00 <nil> <nil> { map[alertname:test1]}}[1.002707:]
        @ 4.009307375
        - &{map[] 0001-01-01T00:00:00.000Z <nil> [] 2025-11-07T15:47:48.705+01:00 <nil> <nil> { map[alertname:test1]}}[1.002707:]
        - &{map[] 0001-01-01T00:00:00.000Z <nil> [] 2025-11-07T15:47:51.706+01:00 <nil> <nil> { map[alertname:test2]}}[4.003107:]
```

dispatch/dispatch.go
siavashs: It seems the failing test expects `group_interval` to be respected after reload, but I don't see where the old code does that when the dispatcher and AGs are recreated.

siavashs: OK, now I see what is happening.
previous implementation would flush immediately after reload but the `DedupStage` would prevent the alert from being sent and alert would flush again at `group_interval`
Now we have the default `1m` delay of startup, which means we don't hit `DedupStage` and instead flush at `group_wait` instead of `group_interval`.

cmd/alertmanager/main.go
siavashs: Mimir uses these limits.

siavashs:
`hasFlashed` is dropped now since Dispatcher controls that with status checks after creating an AG.
This also means that AGs are now lock-free.

dispatch/dispatch.go
siavashs: Updated the logic.
Closing this since the code is outdated.

dispatch/route.go
siavashs: No, I'll cleanup.

dispatch/dispatch.go
siavashs: It is now part of `Run()` logic and is called `minDispatchTime`.

dispatch/dispatch.go
siavashs: I added states to Dispacher so it know if it is an initial run (optionally delayed) or a re-run (no delays).

dispatch/dispatch.go
```diff
@@ -32,6 +32,12 @@ import (
 	"github.com/prometheus/alertmanager/types"
 )
 
+const (
+	DISPATCHER_STATE_UNKNOWN = iota
```
Spaceman1701: small style nitpick: constants should be `MixedCaps`
```suggestion
	DispatcherStateUnknown = iota
	DispatcherStateStarting
	DispatcherStateRunning
```

dispatch/dispatch.go
```diff
@@ -347,6 +364,31 @@ func (d *Dispatcher) processAlert(alert *types.Alert, route *Route) {
 	// alert is already there.
 	ag.insert(alert)
 
+	// Check dispatcher and alert state to determine if we should start dispatching the alert.
+	now := time.Now()
+	switch d.state {
+	case DISPATCHER_STATE_STARTING:
+		if now.After(d.minDispatchTime) {
+			d.logger.Debug("Setting dispatch state to: running")
+			d.state = DISPATCHER_STATE_RUNNING
```
Spaceman1701: This implementation feels a little awkward to me. I have two problems with it:
1. Logically, the dispatcher transitions into the running state as soon as the `minDispatchTime` has passed. As implemented, the transition only happens when the dispatcher sees an alert after `minDispatchTime`.
2. We influence notifications indirectly by setting the aggrGroups' timers externally, rather than leaving it managed internally to the aggrGroup. This seems like something that'll become confusing over time.

I wonder if we could make this a little cleaner by decoupling the logic for the dispatcher state transition from the logic for ingesting an alert.

Perhaps we can start a timer that expires at `minDispatchTime` and make the state change then? We could wait on the timer in the main `select` inside `run`. And perhaps we can control the aggr group behavior by avoiding calling `ag.run` until `minDispatchTime`? We'd need to start any aggrGroups that exist but haven't been started, but that should be possible.

siavashs: Good point, refactored the code.

siavashs:
It seems I broke acceptance tests again, I'll check and fix them.

cmd/alertmanager/main.go
```diff
@@ -143,6 +143,7 @@ func run() int {
 		maxSilenceSizeBytes         = kingpin.Flag("silences.max-silence-size-bytes", "Maximum silence size in bytes. If negative or zero, no limit is set.").Default("0").Int()
 		alertGCInterval             = kingpin.Flag("alerts.gc-interval", "Interval between alert GC.").Default("30m").Duration()
 		dispatchMaintenanceInterval = kingpin.Flag("dispatch.maintenance-interval", "Interval between maintenance of aggregation groups in the dispatcher.").Default("30s").Duration()
+		DispatchStartDelay          = kingpin.Flag("dispatch.start-delay", "Minimum amount of time to wait before dispatching alerts. This option should be synced with value of --rules.alert.resend-delay on Prometheus.").Default("0s").Duration()
```
TheMeier, line 146: Default in prometheus is `1m` so should the default in AM also be `1m`?

siavashs: I initially set this to `1m` but it changes the default behaviour and lots of acceptance tests fail. We can avoid breaking the existing tests by setting it to `0` for all tests.
Adjusting timings is not possible since we add `+1m` to each test.

But I thought same thing could happen to users unexpectedly if they don't pay attention to the changelog and this new cmd flag.

I'm open to setting the default to `1m` to sync it with Prometheus defaults.

TheMeier: I see, whatever you choose, you will choose wrong :D

cmd/alertmanager/main.go
```diff
@@ -415,7 +416,7 @@ func run() int {
 		prometheus.DefaultRegisterer,
 		configLogger,
 	)
-	configCoordinator.Subscribe(func(conf *config.Config) error {
+	configCoordinator.Subscribe(func(conf *config.Config, initial bool) error {
```
ultrotter: Would it make sense to rather pass a delay here, of type time.Duration, then we can pass 0 on reload, and *DispatchStartDelay on initial, rather?

config/coordinator.go
```diff
@@ -30,9 +30,10 @@ type Coordinator struct {
 	logger         *slog.Logger
 
 	// Protects config and subscribers
-	mutex       sync.Mutex
-	config      *Config
-	subscribers []func(*Config) error
+	mutex         sync.Mutex
+	config        *Config
+	subscribers   []func(*Config, bool) error
+	initialReload bool
```
ultrotter: Here we could have an initialDelay, and pass it into the function, plus set it to 0 after the initial notification.
This would b set explicitly at New and not implied...

dispatch/dispatch.go
```diff
@@ -392,13 +454,18 @@ type aggrGroup struct {
 	done    chan struct{}
 	next    *time.Timer
 	timeout func(time.Duration) time.Duration
-
-	mtx        sync.RWMutex
-	hasFlushed bool
+	running atomic.Bool
```
ultrotter, line 447: I know we use go.uber.org/atomic elsewhere, but does this buy us anything over sync/atomic Bool? https://pkg.go.dev/sync/atomic#pkg-types Should we use this and start a switch?

siavashs: There is currently a lint check which enforces this.
I guess we inherit this from Prometheus.
I need to check if this is still required.
cc @SuperQ

juliusv: FYI, there's an open effort in Prometheus to switch to the new stdlib atomic types (those only got added in Go 1.19):

- https://github.com/prometheus/prometheus/issues/14866
- https://github.com/prometheus/prometheus/pull/17358

SuperQ: Yea, I'm not sure. This was introduced in Prometheus in 2020. https://github.com/prometheus/prometheus/issues/7647

I guess we'll need to find out from the rest of the Prometheus devs if we still need this.

bwplotka: We can remove the check, std lib is what we want as per https://github.com/prometheus/prometheus/issues/14866

SuperQ: Looks like we can migrate to to the stdlib now: https://github.com/prometheus/prometheus/issues/14866

config/coordinator.go
siavashs: In that case we can bypass this and just do the reset to zero after the first time dispatch runs in `main`.

dispatch/dispatch.go
```diff
@@ -25,13 +25,26 @@ import (
...
 	"github.com/prometheus/alertmanager/types"
 )
 
+const (
+	DispatcherStateUnknown = iota
+	DispatcherStateWaitingToStart
+	DispatcherStateRunning
+)
+
+var state = map[int]string{
```
Spaceman1701: It's probably more idiomatic to implement `Stringer` if we need this.

e.g.
```
type DispatcherState int

const (
    DispatcherStateUnknown DispatcherState = iota
	DispatcherStateWaitingToStart DispatcherState
	DispatcherStateRunning DispatcherState
)

func (d DispatcherState) String() string {
    switch d {
        case DispatcherStateUnknown:
           return "unknown"
     ...
}
```

dispatch/dispatch.go
```diff
@@ -102,39 +118,44 @@ type Limits interface {
...
+		metrics:             metrics,
+		limits:              limits,
+		state:               DispatcherStateUnknown,
 	}
 	return disp
 }
 
 // Run starts dispatching alerts incoming via the updates channel.
-func (d *Dispatcher) Run() {
+func (d *Dispatcher) Run(dispatchDelay time.Duration) {
```
Spaceman1701: tiny nitpick, take it or leave it: I think it'd be _slightly_ better if this took at `startAt time.Time` since that's a little closer to the meaning and it's a little more flexible for the caller, in my opinion.

dispatch/dispatch.go
siavashs: Dropped it.

dispatch/dispatch.go
siavashs: That was the original approach, let me revert to that.

siavashs: Done.

config/coordinator.go
siavashs: Removed this boolean, we now calculate and use an absolute start time.

cmd/alertmanager/main.go
siavashs: Removed.

dispatch/dispatch.go
siavashs: I guess we wait for this to be fix in Prometheus, and Alertmanager will get it as part of the next sync of common ci/build setup.

TheMeier:
Should we add this one to the v0.30.0 project? Look pretty far along to me

siavashs:
> Should we add this one to the v0.30.0 project? Look pretty far along to me

I think it can make it for that release, all related comments are resolved.
I'll wait for a final review by @Spaceman1701 and @ultrotter

Spaceman1701 (approved):
LGTM. Thanks for making those changes!

dispatch/dispatch.go
ultrotter: Yeah, I was going to note that the new stdlib fixes the issue and provides that... I think we can target a migration for post 0.30, since it's more of a cleanup than something that is urgently needed?

dispatch/dispatch.go
```diff
@@ -102,39 +112,44 @@ type Limits interface {
...
+		metrics:             metrics,
+		limits:              limits,
+		state:               DispatcherStateUnknown,
 	}
 	return disp
 }
 
 // Run starts dispatching alerts incoming via the updates channel.
-func (d *Dispatcher) Run() {
+func (d *Dispatcher) Run(startTime time.Time) {
```
ultrotter: Can we rename this one? It'll be confusing that it's named like start time in the caller (time.Now()) but then passed a value which is increased.... maybe minDispatchTime?

cmd/alertmanager/main.go
SuperQ: I think we should set this to `1m` to match the Prometheus defaults. But we can do that in a followup PR.

dispatch/dispatch.go
siavashs: Renamed it `dispatchStartTime` for clarity.

PR #4705, opened 2025-11-06
Add new behavior to avoid races on config reload

Spaceman1701:
When the config is reloaded, alertmanager re-creates the `inhibit.Inhibitor` and the `dispatch.Dispatcher`. This is necessary because both the `Inhibitor` and `Dispatcher` have internal state which depends on the config. 

In theory, the re-build should be fast because the `provider.Alerts` still contains all the active alerts in memory. In practice, it causes some problems. Since the `Inhibitor` and `Dispatcher` both ingest alerts one at a time on separate goroutines, their state is built concurrently.

The `Inhibitor` works by building an internal cache of inhibiting alerts. If an active alert is missing from the `Inhibitor`'s cache, it does not cause inhibitions. This means that an inhibitor with a partially built cache can erroneously return `false` from `Inhibitor.Mutes`.  

The `Dispatcher` is responsible for building `aggrGroup`s which in turn flush alerts into the notification pipeline. The notification pipeline calls `Inhibitor.Mutes` to prevent notifying for inhibited alerts.

This all comes together to cause a problem: If the `Dispatcher` is able to build an `aggrGroup` that contains inhibited alerts before the `Inhibitor` is able to process those alerts, it may cause a incorrect notification to fire. In the worst case, if the `Dispatcher` is able to build _any_ `aggrGroup` before the `Inhibitor` has completed building its cache, we could see incorrect notifications. Essentially, config reloads cause a race condition between the `Dispatcher` and `Inhibitor` internal caches.

#4704 largely solves this problem for alertmanager restarts by causing the `Dispatcher` to delay sending alerts after startup. However, this isn't desirable during config reloads for a number of reasons. Most importantly, config reloads need to be applied to the entire alertmanager cluster all at once. Any artificial delay will delay any notifications from the cluster. In practice, we've seen this as a spike of notifications for inhibited alerts right after a config reload.

Another related problem is the API. If an API function calls `Dispatcher.Groups` right after a config reload, it might see fewer groups than the in-memory state of alerts would actually create. This is because the `disp` pointer that the API uses is swapped as soon as the new `DIspatcher` is constructed. In practice, we've seen this as the `/alerts/groups` endpoint returning nothing right after a config reload.

This PR adds new mechanisms to avoid all these race conditions. Since the `provider.Alerts` isn't reconstructed, we just need the `Dispatcher` to wait for the `Inhibitor` to process all the alerts which are already in the `provider.Alerts`. Unfortunately, there's no interface to do that. This PR adds
1. A new `provider.Alerts.SlurpAndSubscribe` method which allows implementations of `provider.Alerts` to return a batch of alerts to the caller immediately, rather than one at a time through a `provider.AlertIterator`. The implementation in the `mem.Alerts` is very simple - just return everything that's currently in memory as a batch and then construct the iterator as normal.
2. Add mechanisms to both the `Inhibitor` and `Dispatcher` to indicate whether they're done loading. This just uses a `sync.WaitGroup` internal to each struct. Both are updated to call `SlurpAndSubscribe` and indicate loading is finished when they've finished processing the initial batch.
3. Re-order the logic for reconstructing the `Dispatcher` and `Inhibitor` during a config reload. 
    1. Reload the `Inhibitor` first, and then reload the `Dispatcher` when that's finished
    2. Don't swap the `disp` pointer to the newly reconstructed `Dispatcher` until it's done loading. This prevents the API from seeing the wrong number of groups when it calls `Dispatcher.Groups`.

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -356,8 +356,8 @@ func run() int {
 		disp.Stop()
 	}()
 
-	groupFn := func(routeFilter func(*dispatch.Route) bool, alertFilter func(*types.Alert, time.Time) bool) (dispatch.AlertGroups, map[model.Fingerprint][]string) {
-		return disp.Groups(routeFilter, alertFilter)
+	groupFn := func(ctx context.Context, routeFilter func(*dispatch.Route) bool, alertFilter func(*types.Alert, time.Time) bool) (dispatch.AlertGroups, map[model.Fingerprint][]string, error) {
+		return disp.Groups(ctx, routeFilter, alertFilter)
 	}
 
 	// An interface value that holds a nil concrete value is non-nil.
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -243,6 +243,16 @@ func (d *Dispatcher) WaitForLoading() {
 	d.loadingFinished.Wait()
 }
 
+func (d *Dispatcher) LoadingDone() <-chan struct{} {
+	doneChan := make(chan struct{})
+	go func() {
+		d.WaitForLoading()
+		close(doneChan)
+	}()
+
+	return doneChan
+}
+
 // AlertGroup represents how alerts exist within an aggrGroup.
 type AlertGroup struct {
 	Alerts   types.AlertSlice
@@ -264,7 +274,12 @@ func (ag AlertGroups) Less(i, j int) bool {
 func (ag AlertGroups) Len() int { return len(ag) }
 
 // Groups returns a slice of AlertGroups from the dispatcher's internal state.
-func (d *Dispatcher) Groups(routeFilter func(*Route) bool, alertFilter func(*types.Alert, time.Time) bool) (AlertGroups, map[model.Fingerprint][]string) {
+func (d *Dispatcher) Groups(ctx context.Context, routeFilter func(*Route) bool, alertFilter func(*types.Alert, time.Time) bool) (AlertGroups, map[model.Fingerprint][]string, error) {
+	select {
+	case <-ctx.Done():
+		return nil, nil, ctx.Err()
+	case <-d.LoadingDone():
+	}
 	d.WaitForLoading()
 	groups := AlertGroups{}
 
@@ -327,7 +342,7 @@ func (d *Dispatcher) Groups(routeFilter func(*Route) bool, alertFilter func(*typ
 		sort.Strings(receivers[i])
 	}
 
-	return groups, receivers
+	return groups, receivers, nil
 }
 
 // Stop the dispatcher.
--- a/api/api.go
+++ b/api/api.go
@@ -14,6 +14,7 @@
 package api
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"log/slog"
@@ -83,7 +84,7 @@ type Options struct {
 	// GroupFunc returns a list of alert groups. The alerts are grouped
 	// according to the current active configuration. Alerts returned are
 	// filtered by the arguments provided to the function.
-	GroupFunc func(func(*dispatch.Route) bool, func(*types.Alert, time.Time) bool) (dispatch.AlertGroups, map[model.Fingerprint][]string)
+	GroupFunc func(context.Context, func(*dispatch.Route) bool, func(*types.Alert, time.Time) bool) (dispatch.AlertGroups, map[model.Fingerprint][]string, error)
 }
 
 func (o Options) validate() error {
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -14,6 +14,7 @@
 package v2
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"log/slog"
@@ -77,7 +78,7 @@ type API struct {
 }
 
 type (
-	groupsFn         func(func(*dispatch.Route) bool, func(*types.Alert, time.Time) bool) (dispatch.AlertGroups, map[prometheus_model.Fingerprint][]string)
+	groupsFn         func(context.Context, func(*dispatch.Route) bool, func(*types.Alert, time.Time) bool) (dispatch.AlertGroups, map[prometheus_model.Fingerprint][]string, error)
 	groupMutedFunc   func(routeID, groupKey string) ([]string, bool)
 	getAlertStatusFn func(prometheus_model.Fingerprint) types.AlertStatus
 	setAlertStatusFn func(prometheus_model.LabelSet)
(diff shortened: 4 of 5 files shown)


cmd/alertmanager/main.go
```diff
@@ -518,8 +518,18 @@ func run() int {
 			}
 		})
 
-		go disp.Run()
+		// first, start the inhibitor so the inhibition cache can populate
+		// wait for this to load alerts before starting the dispatcher so
+		// we don't accidentially notify for an alert that will be inhibited
 		go inhibitor.Run()
```
ultrotter, line 539: I am wondering if this is better, or we should instead load everything not in a goroutine, then call run?

cmd/alertmanager/main.go
```diff
@@ -518,8 +518,18 @@ func run() int {
...
+		inhibitor.WaitForLoading()
+
+		// next, start the dispatcher and wait for it to load before swapping the disp pointer.
+		// This ensures that the API doesn't see the new dispatcher before it finishes populating
+		// the aggrGroups
+		go newDisp.Run()
```
ultrotter: Same here...

provider/provider.go
```diff
@@ -76,6 +76,18 @@ type Alerts interface {
...
+	// SlurpAndSubcribe returns a list of all active alerts which are available
+	// in the provider before the call to SlurpAndSubcribe and an iterator
+	// of all alerts available after the call to SlurpAndSubcribe.
+	// SlurpAndSubcribe can be used by clients which need to build in memory state
+	// to know when they've processed the 'initial' batch of alerts in a provider
+	// after they reload their subscription.
+	// Implementation of SlurpAndSubcribe is optional - providers may choose to
+	// return an empty list for the first return value and the result of Subscribe
+	// for the second return value.
+	SlurpAndSubscribe(name string) ([]*types.Alert, AlertIterator)
```
ultrotter, line 89: Ok the name is funny, but maybe SnapshotAndSubscribe?

cmd/alertmanager/main.go
Spaceman1701: Ah yeah, that actually would simplify things a little. The reason I did it this way was to allow this all to happen concurrently (and without waiting) if we wanted. However, in practice it's all sequential. 

Maybe I could change this so that both the `Inhibitor` and `Dispatcher` have `Load()` methods? I think `Run` should still call `Load()`, but maybe we can make this work such that `Load` is a no-op after being called once.

provider/provider.go
siavashs: Alternative names `SubscribeWithHistory()` or `SubscribeWithReplay()`.
Also why not modify `Subscribe()`? On startup it has no alerts, so we can always send if anything is in memory.

Spaceman1701: Sure, happy to change the name. I think the origin of this name was "slurp up all the existing alerts" 😆 

>Also why not modify Subscribe()? On startup it has no alerts, so we can always send if anything is in memory.

I wanted to avoid every possible consumer needing to make a code change after this update. However, we already made a breaking change to `Subscribe` in the last few weeks, so maybe we can sneak this in there too.

I know there are a few projects out there which are linking alertmanager and depending on these interfaces...

cmd/alertmanager/main.go
Spaceman1701: > Maybe I could change this so that both the Inhibitor and Dispatcher have Load() methods? I think Run should still call Load(), but maybe we can make this work such that Load is a no-op after being called once.

I tried this, and it ended up pretty awkward. It might work, but not without a bigger refactoring. Basically, it becomes necessary to move the iterator into a struct field that's protected by the struct's global mutex. This is very awkward because we'd have to lock and unlock the mutex around reads of the iterator's channel. 

The only easy fix would be splitting `SlurpAndSubscribe` method into two, but that's not desirable because the entire benefit comes from being able to atomic separate alerts into "before the call" and "after the call" groups.

dispatch/dispatch.go
```diff
@@ -222,6 +238,7 @@ func (ag AlertGroups) Len() int { return len(ag) }
 
 // Groups returns a slice of AlertGroups from the dispatcher's internal state.
 func (d *Dispatcher) Groups(routeFilter func(*Route) bool, alertFilter func(*types.Alert, time.Time) bool) (AlertGroups, map[model.Fingerprint][]string) {
+	d.WaitForLoading()
```
grobinson-grafana, line 283: Can I suggest replacing this with a select so the caller can cancel? It will allow goroutines from API requests to clean themselves up if the client timesout waiting for the condition to become true.

Spaceman1701: yeah, that's a good idea. I'll make this select between the `Dispatcher` context and `WaitForLoading`. I'll add a `WaitForLoading` chan as well so that the API is a little cleaner

Spaceman1701: Well actually, sadly, `Groups` can't return an error right now, so the only option is to return empty. That's not ideal since we wouldn't want a client to get an empty response if they timed out while waiting for loading.

It'd be an API break to change this now... Thoughts?

grobinson-grafana: It needs to be like this:

```go
func (d *Dispatcher) Groups(ctx context.Context, routeFilter func(*Route) bool, alertFilter func(*types.Alert, time.Time) bool) (AlertGroups, map[model.Fingerprint][]string, error) {
```

so the API code can pass [`Request.Context`](https://pkg.go.dev/net/http#Request.Context) from the request.

Spaceman1701: Agreed - I'm just unsure about making a breaking change to the public API. It sounds like you'd prefer that?

grobinson-grafana: Hmm I'm not following. What breaking change to the public API? If the client has already timed out, the client won't wait for a response,  it will just close the connection?

siavashs: The API controls the context and `Group()` will return when that happens, so it should not break anything.

Spaceman1701: I mean the public API of `Groups` itself - any code which links Alertmanager's Go code may have to make changes. 

I'm not sure what we consider the "public" supported API for Alertmanager's Go code, but I do know there are several projects (like Mimir) which depend on it.

Anyway, it sounds like this isn't a concern for either of you, so I'll make the change.

siavashs: Mimir embeds Alertmanager, so I think this method will only be called if something hits the Alertmanager API exposed by Mimir (which will take the same path).

provider/provider.go
siavashs: I'm fine to keep the current name, we can reiterate on this later if needed.

siavashs (approved):
LGTM

Spaceman1701:
> Needs a rebase

Done, thanks for the heads up!

PR #4707, opened 2025-11-07
Don't allow calling QIDs with an empty IDs list

ultrotter:
Calling QIDs with an empty list of IDs is meaningless, and the result may be unexpected (at the moment it would return all silences, but one might expect it to never return any). Make it an error.

### DIFF
--- a/silence/silence.go
+++ b/silence/silence.go
@@ -736,7 +736,7 @@ type silenceFilter func(*pb.Silence, *Silences, time.Time) (bool, error)
 func QIDs(ids ...string) QueryParam {
 	return func(q *query) error {
 		if len(ids) == 0 {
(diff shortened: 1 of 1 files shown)


silence/silence.go
```diff
@@ -735,6 +735,9 @@ type silenceFilter func(*pb.Silence, *Silences, time.Time) (bool, error)
 // QIDs configures a query to select the given silence IDs.
 func QIDs(ids ...string) QueryParam {
 	return func(q *query) error {
+		if len(ids) == 0 {
+			return fmt.Errorf("QIDs filter must have at least one id")
```
SuperQ: Please use `errors` package here.

```suggestion
			return errors.New("QIDs filter must have at least one id")
```

TheMeier (approved):

PR #4709, opened 2025-11-07
Fix silence_bench_test FAIL

ultrotter:
In 92ecf8b93b6f8fa7cce0fb329de86a1133362765 silence_bench_test.go was left behind since it's not run automatically, and started failing.

Fix by passing a new registry when creating Silences.


ultrotter:
@SuperQ

PR #4713, opened 2025-11-07
use renameio for atomic file replacement

ultrotter:
Since there is a google provided little library for this, and it doesn't have any big dependencies, it's worth using it instead of our own, so we don't have to worry about the details.

This also allows us to remove code that we had duplicated across modules.


ajaychandrak2002-bit:
please test the code and run it

ultrotter:
> please test the code and run it

TBH I did run tests, I forgot lint, but now I fixed it

ultrotter:
Unsupported on windows, so I guess we can't use this as is for now.

PR #4714, opened 2025-11-07
loadSnapshot: update matcher index properly while not holding lock

ultrotter:
Since we're building the state separately, we will do the same with the matcher index, and then update it into the Silences with the lock held.


ajaychandrak2002-bit:
i think cms builder will be closed

ajaychandrak2002-bit:
nigreat work

ajaychandrak2002-bit:
I have done with this ,

PR #4719, opened 2025-11-09
Add chat_id_file configuration parameter for Telegram

### LINKED ISSUE #4672: Add `chat_id_file` configuration parameter for Telegram
**Description**
I want to suggest a feature to add `chat_id_file` Telegram configuration parameter to provide more securely pass `chat_id` like already existing `bot_token_file`.

**Problem**
Now I have to pass `chat_id` in cleartext inside config file and it's not safe.
This feature allow to pass file contains `chat_id` value using secrets or volumes sections in docker-compose.

**How to solve**
Add `receivers['telegram']/telegram_configs/chat_id_file` field with type `string`.
Field `chat_id_file` must take a path to file with chat_id.

I've already implemented it locally and I'm ready to contribute it.

tennisleng:
This PR adds support for reading `chat_id` from a file, similar to `bot_token_file`. This allows users to securely pass `chat_id` using secrets or volumes in docker-compose without exposing it in cleartext in the config file.

## Changes
- Added `ChatIDFile` field to `TelegramConfig` struct
- Updated validation to require either `chat_id` or `chat_id_file` (but not both)
- Added `getChatID()` method to read chat_id from file
- Updated `Notify()` method to use the new `getChatID()` method

## Testing
- Follows the same pattern as `bot_token_file`
- Validates that either `chat_id` or `chat_id_file` is provided
- Validates that both are not provided simultaneously
- Reads and parses chat_id from file as int64

Fixes #4672

### DIFF
--- a/notify/telegram/telegram.go
+++ b/notify/telegram/telegram.go
@@ -19,6 +19,7 @@ import (
 	"log/slog"
 	"net/http"
 	"os"
+	"strconv"
 	"strings"
 
 	commoncfg "github.com/prometheus/common/config"
@@ -92,7 +93,12 @@ func (n *Notifier) Notify(ctx context.Context, alert ...*types.Alert) (bool, err
 		return true, err
 	}
 
-	message, err := n.client.Send(telebot.ChatID(n.conf.ChatID), messageText, &telebot.SendOptions{
+	chatID, err := n.getChatID()
+	if err != nil {
+		return true, err
+	}
+
+	message, err := n.client.Send(telebot.ChatID(chatID), messageText, &telebot.SendOptions{
 		DisableNotification:   n.conf.DisableNotifications,
 		DisableWebPagePreview: true,
 		ThreadID:              n.conf.MessageThreadID,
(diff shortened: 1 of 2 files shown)


notify/telegram/telegram.go
```diff
@@ -130,3 +136,19 @@ func (n *Notifier) getBotToken() (string, error) {
 	}
 	return string(n.conf.BotToken), nil
 }
+
+func (n *Notifier) getChatID() (int64, error) {
+	if len(n.conf.ChatIDFile) > 0 {
+		content, err := os.ReadFile(n.conf.ChatIDFile)
```
ultrotter, line 142: Do we expect the contents to change as we go? Would we consider saving the chatID to n.conf.ChatID after we read it, and then in this function return that one first, and only read the file if ChatID is still 0?

ultrotter (approved):

notify/telegram/telegram.go
tennisleng: I've added the caching you suggested - getChatID() now returns the cached value first and only reads from file if ChatID is still 0, then caches it for future calls. This avoids the file I/O on every notification

ultrotter (changes requested):
Please can you remove the unrelated changes from this branch? Otherwise I don't think it can be merged, as is...

Thanks!

G

TheMeier:
If chat_id is considered a secret it also get mapped to a secret type. Right now it is mapeed to int and is visible unmasked in `/#/status`

TheMeier:
related #2498

notify/telegram/telegram.go
ultrotter: I think you removed this improvement when you removed the unrelated changes from the branch... You could cherry-pick just this one patch on top, or push to a version that has just those.

PR #4722, opened 2025-11-10
Improve silence benchmarks

ultrotter:
- Add some benchmarks for GC
- Add benchmarks that call Mutes while adding some silences, to better test how it would be called in the real world
- Change the existing Mutes test to have some matching and some non-matching silences, to be more realistic to how things would look in a real alertmanager
- Pass a NoOpLogger to quartz to make the tests less noisy
- Use Set() to add silences instead of pushing them onto the state

### DIFF
--- a/silence/silence_bench_test.go
+++ b/silence/silence_bench_test.go
@@ -14,7 +14,8 @@
 package silence
 
 import (
-	"bytes"
+	"fmt"
+	"math/rand"
 	"strconv"
 	"sync"
 	"testing"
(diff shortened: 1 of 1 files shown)


ultrotter:
GC Benchmarks are a bit slow. Options:

1) Move them to a separate file with a -slow tag so that one can run them less often/on demand more easily
2) Include the whole loop in the timer, which would measure also the snapshot load, other than the GC
3) Look into speeding up the setup we do?

Any preferences?

PR #4730, opened 2025-11-13
tests: Use t.TempDir()

ultrotter:
A lot of tests create temporary files. Only one cleans up.
Use t.TempDir() in the CreateTemp calls, to get automatic cleanup.

Signed-off-by: Guido Trotter <guido@hudson-trading.com>


ultrotter:
> Don't know if there was/is actually a reason to not use t.TempDir(). But looks like an improvement to me

Well it was added in 2020, and alertmanager is older. Maybe nobody thought about adding it?

PR #4731, opened 2025-11-13
Change URL type from secret to string in configuration

zhuqifu:
Just a small suggestion: perhaps using `url: <string>` would be clearer? I'm not entirely sure.

@metalmatze


TheMeier:
IMHO this is correct as is https://github.com/prometheus/alertmanager/blob/main/config/notifiers.go#L604

PR #4734, opened 2025-11-13
feat: add config directive to pass wechat api secret via file

TheMeier:
Ref: #2498

### DIFF
--- a/notify/wechat/wechat.go
+++ b/notify/wechat/wechat.go
@@ -23,6 +23,8 @@ import (
 	"log/slog"
 	"net/http"
 	"net/url"
+	"os"
+	"strings"
 	"time"
 
 	commoncfg "github.com/prometheus/common/config"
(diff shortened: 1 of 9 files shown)


notify/wechat/wechat.go
```diff
@@ -97,7 +99,11 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 	// Refresh AccessToken over 2 hours
 	if n.accessToken == "" || time.Since(n.accessTokenAt) > 2*time.Hour {
 		parameters := url.Values{}
-		parameters.Add("corpsecret", tmpl(string(n.conf.APISecret)))
+		apiSecret, err := n.getApiSecret()
```
ultrotter, line 104: Do we want to read the file on each notification/expect it to change?

TheMeier: I must admit I just copied the pattern in https://github.com/prometheus/alertmanager/blob/main/notify/pagerduty/pagerduty.go#L228

But here this only run if accesToken is empty or older then 2 hours (line 100), so it should be fine.

PR #4735, opened 2025-11-14
Fix silence import: also wait for the error collection goroutine to finish

ultrotter:
As noticed by George Robinson the error collection goroutine in silence
import is also not waited for, so we may get an incorrect count when we
exit. This adds a done channel for that goroutine, and checks that the
error count is correct with a new test.

Signed-off-by: Guido Trotter <guido@hudson-trading.com>

### DIFF
--- a/cli/silence_import.go
+++ b/cli/silence_import.go
@@ -102,18 +102,16 @@ func (c *silenceImportCmd) bulkImport(ctx context.Context, _ *kingpin.ParseConte
 	errDone := make(chan struct{})
 
 	var wg sync.WaitGroup
-	cleanupDone := false
+	var once sync.Once
 
 	closeChannels := func() {
-		if cleanupDone {
-			return
-		}
-		close(silencec)
-		wg.Wait()
-		close(errc)
-		<-errDone
-		close(errDone)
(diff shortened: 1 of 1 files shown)


TheMeier:
Looks ok, but I said that with the previous one and now I'm unsure :D 

I was wondering if the cleanupDone could be done with https://pkg.go.dev/sync#Once

eg
```
	var once sync.Once
	closeChannels := func() {
		once.Do(func() {
			close(silencec)
			wg.Wait()
			close(errc)
			<-errDone
			close(errDone)
		})
	}
```

ultrotter:
> Looks ok, but I said that with the previous one and now I'm unsure :D
> 
> I was wondering if the cleanupDone could be done with https://pkg.go.dev/sync#Once
> 
> eg
> 
> ```
> 	var once sync.Once
> 	closeChannels := func() {
> 		once.Do(func() {
> 			close(silencec)
> 			wg.Wait()
> 			close(errc)
> 			<-errDone
> 			close(errDone)
> 		})
> 	}
> ```

Done, thanks for the suggestion!

siavashs (approved):

Issue #1236, opened 2018-02-12
Alert manager unable to send unsecured emails


orlmar:
We just upgraded Alertmanager from version 0.9.1 to 0.13.0 and Prometheus from 1.8.1 to 2.1. The Alertmanager's email receiver doesn't work anymore with identical configuration.

In the log, we have an error:
msg="Notify attempt failed" attempt=10 integration=email err="*smtp.plainAuth failed: unencrypted connection" 

Could it be related to this issue in net/smtp? 
https://github.com/golang/go/issues/22166

Th smtp related configuration is:

global:
  smtp_smarthost: 'smtpserver:25'
  smtp_from: 'user@domain'
  smtp_auth_username: 'user'
  smtp_requiere_tls: false
... 
receivers:
- name: 'emailreceiver'
  email_configs:
  - to: 'user@domain'
    headers:
      subject: 'testing'
    html: 'some body'

Also we tested requiere_tls parameter in email_configs section for the case it doesn't read the global parameter.

stuartnelson3:
The issue you linked says the fix for that regression was added to go1.9.2, which was used to compile v0.13.0.

simonpasquier:
As stated in the [SMTP](https://godoc.org/net/smtp#PlainAuth) godoc, the library will allow plain authentication only if TLS is used or the server is local. So you'll have to fulfill one of the conditions to make it work. I'm afraid there's nothing much that can be done on the AlertManager's side.

note:
@simonpasquier That's understandable then but I would say that it would be nice to have it stated in the alertmanager's docs. If the application exposes `smtp_require_tls` and the documentation does not states that it's not effective (outside of localhost) then it's quite confusing for the user.

khavul:
Bump. Running into the same issue, need answers.

xkfen:
just change the `global.smtp_require_tls` from `false` to `true`

TheMeier:
As discussed this is a limitation of underlying libraries. Also it is not good practice to use plain-text credentials over unencrpted channels. If you still require this feature, pleas re-open this issue or even better, provide a PR implementing it.

PR #4741, opened 2025-11-14
chore: add comment about SMTP PLAIN authentication

TheMeier:
Ref: #1236

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -80,6 +80,7 @@ global:
   # The default hostname to identify to the SMTP server.
   [ smtp_hello: <string> | default = "localhost" ]
   # SMTP Auth using CRAM-MD5, LOGIN and PLAIN. If empty, Alertmanager doesn't authenticate to the SMTP server.
+  # PLAIN is only supported when using TLS.
   [ smtp_auth_username: <string> ]
   # SMTP Auth using LOGIN and PLAIN.
   [ smtp_auth_password: <secret> ]
@@ -1155,7 +1156,7 @@ The default `jira.default.description` template only works with V2.
 # The API Type to use for search requests, can be either auto, cloud or datacenter
 # Example: cloud
 [ api_type: <string> | default = auto ]
-   
+
 # The project key where issues are created.
 project: <string>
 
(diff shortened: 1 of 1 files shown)


SoloJacobs (approved):
LGTM.

I have been thinking whether we should also a validation step to `amtool check-config`. This should be possible, since the validation done by `PlainAuth` is straightforward.

TheMeier:
So I  have added a commit that implements validation if TLS is disabled and SMTP PLAIN auth config params are set. But I would consider this somewhat breaking since it will prevent AM starting with some (broken) configs which it previously did accept.
I can also break out that commit into another PR.

config/config.go
```diff
@@ -451,6 +451,9 @@ func (c *Config) UnmarshalYAML(unmarshal func(any) error) error {
 				ec.RequireTLS = new(bool)
 				*ec.RequireTLS = c.Global.SMTPRequireTLS
 			}
+			if (ec.AuthUsername != "" || ec.AuthPassword != "" || ec.AuthPasswordFile != "") && !*ec.RequireTLS {
```
SoloJacobs: Shouldn't we still allow `localhost`? This is how `PlainAuth` checks it:
```go
func isLocalhost(name string) bool {                                                                  
    return name == "localhost" || name == "127.0.0.1" || name == "::1"                                  
} 
```

TheMeier: I have removed the validation and will open another PR for it. Thanks for your feedback @SoloJacobs

TheMeier:
So about that validation. I think a more correct implementation would be to use something like this:

```
			if !*ec.RequireTLS && (ec.AuthUsername != "" || ec.AuthPassword != "" || ec.AuthPasswordFile != "") {
				if ip := net.ParseIP(ec.Smarthost.Host); ip != nil && !ip.IsLoopback() {
					return errors.New("PLAIN SMTP authentication without TLS can only be used with loopback (aka localhost) addresses")
				}
			}
```
But that of cause is different than https://cs.opensource.google/go/go/+/refs/tags/go1.25.4:src/net/smtp/auth.go;l=67
The issue I see here is, whatever validation we implement is prone to become different at some point _if_ the implementation in net/smtp changes in the future. So I also don't feel very happy to just copy their `isLocalhost` implementation.

SoloJacobs:
The validation also seems ok to me. Shame that there is no public function to check that SMTP is safe.

PR #4749, opened 2025-11-17
Use ulid v2

SuperQ:
Update the ulid generator library to v2.


SuperQ:
I noticed upstream Prometheus updated this library, mirror changes from https://github.com/prometheus/prometheus/pull/16168.

cluster/cluster.go
```diff
@@ -196,11 +195,7 @@ func Create(
 
 	// Generate a random name if none is provided.
 	if name == "" {
-		id, err := ulid.New(ulid.Now(), rand.New(rand.NewSource(time.Now().UnixNano())))
-		if err != nil {
-			return nil, err
-		}
-		name = id.String()
+		name = ulid.Make().String()
```
TheMeier: Is it ok to use the ulid.Make() which does panic on error here?
There is also 
```
func New(ms uint64, entropy io.Reader) (id ULID, err error) {
```

SuperQ: Sure, I can re-add the error check. I'm not sure what we do at runtime if `cluster.Create()` fails.

SuperQ:
No worries, I figure this is better after reading over the Prometheus use of the library. More consistent.

PR #4750, opened 2025-11-18
Enable modernize linter

SuperQ:
Enable the golangci-lint modernize linter.
* Add exception for the omitempty struct issue.
* Apply modernize fixes.

PR #4751, opened 2025-11-18
chore: interface{} -> any in comments

SoloJacobs:
Tiny left over from
https://github.com/prometheus/alertmanager/pull/4750

(I noticed after rebasing on @SuperQ changes)


SoloJacobs:
Hmm, I am a bit confused about this:
```
$ head api/v2/restapi/configure_alertmanager.go
// This file is safe to edit. Once it exists it will not be overwritten
```
I feel about not chaning `api/v2` anyway, so I dropped the change.

Issue #4742, opened 2025-11-14
Fix flakes in `TestMergeAlerts`


SoloJacobs:
Passes locally fine, but failed here: https://github.com/prometheus/alertmanager/pull/4739#issuecomment-3533668109

```
go test -race  ./...
?   	github.com/prometheus/alertmanager/api	[no test files]
?   	github.com/prometheus/alertmanager/api/metrics	[no test files]
ok  	github.com/prometheus/alertmanager/api/v2	1.068s
?   	github.com/prometheus/alertmanager/api/v2/client	[no test files]
?   	github.com/prometheus/alertmanager/api/v2/client/alert	[no test files]
?   	github.com/prometheus/alertmanager/api/v2/client/alertgroup	[no test files]
?   	github.com/prometheus/alertmanager/api/v2/client/general	[no test files]
?   	github.com/prometheus/alertmanager/api/v2/client/receiver	[no test files]
?   	github.com/prometheus/alertmanager/api/v2/client/silence	[no test files]
?   	github.com/prometheus/alertmanager/api/v2/models	[no test files]
?   	github.com/prometheus/alertmanager/api/v2/restapi	[no test files]
?   	github.com/prometheus/alertmanager/api/v2/restapi/operations	[no test files]
?   	github.com/prometheus/alertmanager/api/v2/restapi/operations/alert	[no test files]
?   	github.com/prometheus/alertmanager/api/v2/restapi/operations/alertgroup	[no test files]
?   	github.com/prometheus/alertmanager/api/v2/restapi/operations/general	[no test files]
?   	github.com/prometheus/alertmanager/api/v2/restapi/operations/receiver	[no test files]
?   	github.com/prometheus/alertmanager/api/v2/restapi/operations/silence	[no test files]
?   	github.com/prometheus/alertmanager/asset	[no test files]
ok  	github.com/prometheus/alertmanager/cli	1.056s
ok  	github.com/prometheus/alertmanager/cli/config	1.025s
?   	github.com/prometheus/alertmanager/cli/format	[no test files]
ok  	github.com/prometheus/alertmanager/cluster	3.304s
?   	github.com/prometheus/alertmanager/cluster/clusterpb	[no test files]
ok  	github.com/prometheus/alertmanager/cmd/alertmanager	1.051s
?   	github.com/prometheus/alertmanager/cmd/amtool	[no test files]
ok  	github.com/prometheus/alertmanager/config	1.141s
[...]

PR #4753, opened 2025-11-18
Fix flakes in tests using `freeAddress`

SoloJacobs:
This commit addresses the following error:
```
 time=2025-11-14T16:14:59.573Z level=ERROR source=main.go:559
msg="Listen error" err="listen tcp 127.0.0.1:36357: bind: address
already in use"
```

`freeAddress` is called multiple times within one test, but also in parallel in multiple tests. `freeAddress` immediately releases the port despite it being used later in a test. This causes ports to errorenously to be reused.

We replace `net.Listen` in order to avoid calling `syscall.Listen`. This way the port can be reused by AM, without the kernel assuming that port can be reused by other processes specifiying port `0`.

Fixes: #4742

### DIFF
--- a/test/cli/acceptance.go
+++ b/test/cli/acceptance.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"io"
 	"maps"
-	"net"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


TheMeier:
I tried this locally and it still produces the same error but now in `TestColdStart`

```
--- FAIL: TestColdStart (5.07s)
    acceptance.go:193: failed to start alertmanager cluster: unable to get a successful response from the Alertmanager: Get "http://127.0.0.1:37501/api/v2/status": dial tcp 127.0.0.1:37501: connect: connection refused
    send_test.go:481: 
        collector "webhook":
        
        interval [4,5]
        ---
        - &{map[] 0001-01-01T00:00:00.000Z <nil> [] 0001-01-01T00:00:01.000Z <nil> <nil> { map[alertname:test1]}}[1:]
```

SoloJacobs:
Ah, wow. That is definitely a problem.

PR #4765, opened 2025-11-22
chore: gofrs/uuid -> google/uuid

SoloJacobs:
We already depend indirectly on `google/uuid`, and removing `gofrs/uuid` from our dependencies is a win. This change does not affect behaviour.


SoloJacobs:
Changed the commit message from `chore` to `refactor`, otherwise just a simple rebase.

Issue #684, opened 2017-04-03
Allow templating in the webhook URL


prasenforu:
Alertmanager version: 0.5.1

My alert configure as follows ...

```
ALERT PoDHighCPUUsage
  IF sum(rate(container_cpu_usage_seconds_total{container_label_io_kubernetes_pod_namespace!=""}[2m])) BY (kubernetes_pod_name, kubernetes_io_hostname, container_label_io_kubernetes_pod_namespace) >= 0.2
  FOR 2m
  LABELS {severity="criticalup"}
  ANNOTATIONS {summary="CPU usage of POD {{ $labels.kubernetes_pod_name }} {{ $labels.container_label_io_kubernetes_pod_namespace }} utilization is  >= 0.4 cores (current value: {{ $value }}%)"}
```
webhooks setup as follows ..

```
- name: 'email-n-webhookup'
  email_configs:
  - to: 'mail-id'
    send_resolved: true

  webhook_configs:
  - url: http://<webhook_ip>:9000/hooks/ose-webhook?in1={{.Labels.container_label_io_kubernetes_pod_namespace}}&in2={{.Labels.kubernetes_pod_name}}&in3=plus

- name: 'email-n-webhookdn'
  email_configs:
  - to: 'mail-id'
    send_resolved: true

  webhook_configs:
  - url: http://<webhook_ip>:9000/hooks/ose-webhook?in1=tomi&in2=chat&in3=minus
```

In webhook server logs I am getting data as follows.

```
[webhook] 2017/04/03 10:03:17 2017-04-03T10:03:17Z | 200 |       194.078µs | 172.30.200.104:9000 | POST /hooks/ose-webhook
[webhook] 2017/04/03 10:03:17 executing /etc/webhook/ose.sh (/etc/webhook/ose.sh) with arguments ["/etc/webhook/ose.sh" "{{.labels.container_label_io_kubernetes_pod_namespace}}" "{{$labels.kubernetes_pod_name}}" "plus"] and environment [] using /etc/webhook as cwd
```
not as a data, its coming as a complete variable.

So I am not able to pass the value/data "labels.kubernetes_pod_name" & "labels.container_label_io_kubernetes_pod_namespace" in webhook_configs "email-n-webhookup"

Statically it worked as per "email-n-webhookdn"

I think its supported for Email as per https://www.robustperception.io/using-labels-to-direct-email-notifications/ 

not sure if I can use it in webhook as well ......

Please help ....

[...]

PR #4798, opened 2025-12-05
Support templating in webhook url fields

waltherlee:
It's an old thread, but there's been some discussion here about supporting templating in webhook url fields: https://github.com/prometheus/alertmanager/issues/684

The issue was closed in 2017 because the alertmanager didn't support templating for any http/url fields, but there're some examples where it does support it now, for example [SlackAction.URL](https://github.com/prometheus/alertmanager/blob/v0.29.0/config/notifiers.go#L406) and [PushoverConfig.URL](https://github.com/prometheus/alertmanager/blob/v0.29.0/config/notifiers.go#L796), so I think we can re-evaluate adding support for this.

There's some good use cases described by others in the thread since it was closed.

This change adds templating to the fields. Existing tests already check that secret URL is still not exposed.


SoloJacobs (commented):
From what I understand:

1) We don't expose `MarshallYAML`, so the new implementation is backwards compatible. I.e, all old configurations are accepted by the new `UnmarshallYAML` implementation. Here are two configurations, which are only accepted by the new implementation.

```yaml
- name: 'empty-url'
  webhook_configs:
  - url: ''
    url_file: 'url.txt'

- name: 'not-a-url'
  webhook_configs:
  - url: 'ssh://ab'
```

For the `not-a-url` config the new behaviour is rather unfortunate. Did you consider validating URLs, which don't have a template?

2) Could we maybe get some tests for the new templating in `notify/webhook/webhook.go`?
3) There are also some other configurations, which could be handled in a similar manor I think:

* `SlackConfig`
* `MSTeamsConfig`
* `MSTeamsV2Config`
* `MattermostConfig`
* `DiscordConfig`

I guess that is out-off-scope for this PR, though.

waltherlee:
Sure, I can add the change for those too and some tests. Thanks Solomon

waltherlee:
The change from `SecretURL` to `Secret` was breaking some tests and validations, so I created a new `SecretTemplURL` to keep all the existing checks at config-level for URLs without templating.

I added tests for templated urls in webhook but won't change the other URL fields from other configs in this PR to keep it small. I'll send a second one with just the switch to `SecretTemplURL`.

config/config.go
```diff
@@ -199,6 +200,85 @@ func (s *SecretURL) UnmarshalJSON(data []byte) error {
...
+	if err != nil {
+		return true, err
+	}
+	return true, nil
+}
+
+// SecretTemplURL is a Secret string that represents a URL which may contain
+// Go template syntax. Unlike SecretURL, it allows templated values and only
+// validates non-templated URLs at unmarshal time.
+type SecretTemplURL Secret
```
Spaceman1701: this doesn't need to block merging this change, but I'd prefer if we named this `SecretTemplateURL` - it's only a few more characters, and I think it's quite a bit more clear to a reader

waltherlee: Sure, renamed.

SuperQ (approved):

PR #2539, opened 2021-04-08
Fix grammar for Silence is not yet valid

jsoref:

PR #4800, opened 2025-12-05
Fix grammar for Silence is not yet valid

SoloJacobs:
I'm just adding the follow-up commit here, so we can merge the original contribution, see https://github.com/prometheus/alertmanager/pull/2539


SuperQ:
I think it's easier to merge this with the cherry-pick.

PR #4807, opened 2025-12-12

SoloJacobs:
This PR is an update of https://github.com/prometheus/alertmanager/pull/3491 that's updated to work against main, which in turn is an update of https://github.com/prometheus/alertmanager/pull/3139 .
So far, I have held off on making some changes myself, except for what was necessary to get CI green again. I also have a small laundry list of things I would fix here, but I would like a second pair of eyes on this request first.

> This PR is an extension of @OktarianTB's 9-month old PR https://github.com/prometheus/alertmanager/pull/3139 that's updated to work against the latest code in main, to hopefully increase its likelihood of being merged.
>
> Credit for this work goes to @OktarianTB - I believe I've marked the commit authorship correctly, please let me know if I made any mistakes here.
>
> > 
> > This PR aims to make progress towards https://github.com/prometheus/alertmanager/issues/2845 by adding a new command to amtool to send test notifications to every receiver in an alertmanager config file. The command takes in an alertmanager config file and optionally an alert file to mock values for labels and annotations. This is the MVP for the command, potentially a few other options can be added later if deemed of use. This largely re-uses the equivalent implementation in [grafana/grafana](https://github.com/grafana/grafana/blob/main/pkg/services/ngalert/notifier/receivers.go).
> > 
> > I have only tested this with Slack so far.


TheMeier:
I gave it a try, its a useful helper. I know you did not add additional code on purpose yet. But I think we should at least include all notifiers that are in the current main branch. I noticed at least teamsv2 and rocketchat missing.

TheMeier (approved):

TheMeier:
@SoloJacobs can you rebase that?

cli/receivers.go
```diff
@@ -0,0 +1,309 @@
+// Copyright 2022 Prometheus Team
```
siavashs, line 1:
```suggestion
// Copyright The Prometheus Authors
```

cli/receivers.go
```diff
@@ -0,0 +1,309 @@
...
+				integrations = append(integrations, ReceiverIntegration{
+					Integration: notify.NewIntegration(n, rs, name, i, nc.Name),
+				})
+			}
+		}
+	)
+
+	for i, c := range nc.WebhookConfigs {
+		add("webhook", i, c, func(l *slog.Logger) (notify.Notifier, error) { return webhook.New(c, tmpl, l) })
+	}
```
siavashs, line 270: Doing this per receiver seems excessive to me.
We should instead just have one call to a module to add all the configured receivers.
Otherwise this logic will be always missing one or more receivers in future.

cli/test_receivers.go
```diff
@@ -0,0 +1,119 @@
+// Copyright 2022 Prometheus Team
```
siavashs, line 1:
```suggestion
// Copyright The Prometheus Authors
```

cli/test_receivers_test.go
```diff
@@ -0,0 +1,58 @@
+// Copyright 2022 Prometheus Team
```
siavashs, line 1:
```suggestion
// Copyright The Prometheus Authors
```

PR #4809, opened 2025-12-12
fix(provider): reduce lock contention

siavashs:
The provider loops over all alerts per state, which results in 3 loops over all store alerts and one call per alert to the marker per loop.

Add a custom collector from counting alerts by state. This reduces the number of calls to store and marker to 1/3.


provider/mem/mem.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 Prometheus Team
+// Copyright Prometheus Authors
```
SuperQ:
```suggestion
// Copyright The Prometheus Authors
```

PR #4811, opened 2025-12-12
Set spellcheck=false attribute on silence filter <input>

SoloJacobs:
This change simply tidies up an existing change: https://github.com/prometheus/alertmanager/pull/3529

It fixes the red underlining of labels:
image

Original PR:

> This hints to iOS Safari and other browsers that they shouldn't use
> spellcheck, but, more importantly, shouldn't use smart quotes (like
> “this” instead of "this"). This matters here because Alertmanager
> requires quotes for a filter but won't accept smart quotes, making it
> tedious since the default input keyboard will normally use them
> automatically.


TheMeier:
LGTM, but I can't really verify the result due to lack of apple devices.

SoloJacobs:
The red squiggly lines also happen on Linux, which is why I deem this worth merging.

@SuperQ Can you merge this ?

SoloJacobs:
Reproduced on Mac as well now. From my test, smart quoting appears to be a user-defined setting, which was turned on my notebook. The fix those what it should.

SuperQ:
lol, oof, smartquotes.

Issue #4757, opened 2025-11-20
If I customize the SSL/TLS port, how should Alertmanager support?


Gintoms:
## Summary
The current email notification implementation has hardcoded logic that only enables direct TLS connections for port 465. This creates a limitation when using TLS on other ports, causing `create SMTP client: EOF` errors when the server expects TLS connections on non-465 ports.

## Description
In [notify/email/email.go:133](https://github.com/prometheus/alertmanager/blob/main/notify/email/email.go#L133), there's hardcoded logic that only uses `tls.Dial()` for port 465, while all other ports use plain TCP connections. This causes issues when SMTP servers require TLS connections on ports other than 465.

However, some SMTP servers may be configured to require TLS connections on non-standard ports but don't support STARTTLS, leading to EOF errors.

Spaceman1701:
Hi, thanks for looking into this issue! Allowing tls over non-standard ports seems like a very reasonable feature. 

I think the best way to do this would be a new field in `EmailConfig` (https://github.com/prometheus/alertmanager/blob/main/config/notifiers.go#L300-L317) that "forces" TLS. This way, we can keep the existing behavior where we assume a TLS connection is required if the port is `465` while allowing users to configure TLS for other ports. 

What do you think of that plan? Would you be interested in contributing this change?

Gintoms:
> Hi, thanks for looking into this issue! Allowing tls over non-standard ports seems like a very reasonable feature.
> 
> I think the best way to do this would be a new field in `EmailConfig` (https://github.com/prometheus/alertmanager/blob/main/config/notifiers.go#L300-L317) that "forces" TLS. This way, we can keep the existing behavior where we assume a TLS connection is required if the port is `465` while allowing users to configure TLS for other ports.
> 
> What do you think of that plan? Would you be interested in contributing this change?

Sounds good!

Gintoms:
https://github.com/prometheus/alertmanager/pull/4818

SoloJacobs:
@Gintoms I think this resolved? (Thank you very much for your contribution.)

Gintoms:
> [@Gintoms](https://github.com/Gintoms) I think this resolved? (Thank you very much for your contribution.)

yes

PR #4818, opened 2025-12-17
smtp: Customize the SSL/TLS port support (#4757)

Gintoms:
## Summary
The current email notification implementation has hardcoded logic that only enables direct TLS connections for port 465. This creates a limitation when using TLS on other ports, causing `create SMTP client: EOF` errors when the server expects TLS connections on non-465 ports.

## Description
In [notify/email/email.go:133](https://github.com/prometheus/alertmanager/blob/main/notify/email/email.go#L133), there's hardcoded logic that only uses `tls.Dial()` for port 465, while all other ports use plain TCP connections. This causes issues when SMTP servers require TLS connections on ports other than 465.

However, some SMTP servers may be configured to require TLS connections on non-standard ports but don't support STARTTLS, leading to EOF errors.

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -315,7 +315,12 @@ type EmailConfig struct {
 	Text             string               `yaml:"text,omitempty" json:"text,omitempty"`
 	RequireTLS       *bool                `yaml:"require_tls,omitempty" json:"require_tls,omitempty"`
 	TLSConfig        *commoncfg.TLSConfig `yaml:"tls_config,omitempty" json:"tls_config,omitempty"`
-	Threading        ThreadingConfig      `yaml:"threading,omitempty" json:"threading,omitempty"`
+	// ForceImplicitTLS controls whether to use implicit TLS (direct TLS connection).
+	// true: force use of implicit TLS (direct TLS connection)
+	// false: force disable implicit TLS (use explicit TLS/STARTTLS if required)
+	// nil (default): auto-detect based on port (465=implicit, other=explicit) for backward compatibility
+	ForceImplicitTLS *bool           `yaml:"force_implicit_tls,omitempty" json:"force_implicit_tls,omitempty"`
+	Threading        ThreadingConfig `yaml:"threading,omitempty" json:"threading,omitempty"`
 }
 
 // ThreadingConfig configures mail threading.
--- a/notify/email/email.go
+++ b/notify/email/email.go
@@ -134,7 +134,16 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 		err     error
 		success = false
 	)
-	if n.conf.Smarthost.Port == "465" {
+	// Determine whether to use Implicit TLS
+	var useImplicitTLS bool
+	if n.conf.ForceImplicitTLS != nil {
+		useImplicitTLS = *n.conf.ForceImplicitTLS
+	} else {
+		// Default logic: port 465 uses implicit TLS (backward compatibility)
+		useImplicitTLS = n.conf.Smarthost.Port == "465"
+	}
+
+	if useImplicitTLS {
 		tlsConfig, err := commoncfg.NewTLSConfig(n.conf.TLSConfig)
 		if err != nil {
 			return false, fmt.Errorf("parse TLS configuration: %w", err)
@@ -177,7 +186,7 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 	}
 
 	// Global Config guarantees RequireTLS is not nil.
-	if *n.conf.RequireTLS {
+	if *n.conf.RequireTLS && !useImplicitTLS {
 		if ok, _ := c.Extension("STARTTLS"); !ok {
 			return true, fmt.Errorf("'require_tls' is true (default) but %q does not advertise the STARTTLS extension", n.conf.Smarthost)
 		}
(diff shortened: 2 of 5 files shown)


notify/email/email.go
```diff
@@ -130,7 +130,17 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 		err     error
 		success = false
 	)
-	if n.conf.Smarthost.Port == "465" {
+	// Determine whether to use Implicit TLS
+	var useImplicitTLS bool
+	if n.conf.ImplicitTLS != nil && *n.conf.ImplicitTLS {
+		// User explicitly configured to use implicit TLS
+		useImplicitTLS = true
```
Spaceman1701: small nitpick, but this seems to control whether TLS is used, so I think it could just be called `useTLS`.

config/notifiers.go
```diff
@@ -315,6 +315,10 @@ type EmailConfig struct {
 	Text             string               `yaml:"text,omitempty" json:"text,omitempty"`
 	RequireTLS       *bool                `yaml:"require_tls,omitempty" json:"require_tls,omitempty"`
 	TLSConfig        *commoncfg.TLSConfig `yaml:"tls_config,omitempty" json:"tls_config,omitempty"`
+	// ImplicitTLS controls whether to use implicit TLS (direct TLS connection).
+	// true: force use of implicit TLS (direct TLS connection)
+	// nil (default): auto-detect based on port (465=implicit, other=explicit) for backward compatibility
+	ImplicitTLS      *bool                `yaml:"implicit_tls,omitempty" json:"implicit_tls,omitempty"`
```
Spaceman1701: I hate to bikeshed, but I think this name is a little confusing. This flag controls whether TLS is forced even when TLS is not required. I think that's the opposite of "implicit". My suggestion would be `ForceTLS` or `UseTLS`.

notify/email/email.go
```diff
@@ -173,7 +183,7 @@ func (n *Email) Notify(ctx context.Context, as ...*types.Alert) (bool, error) {
 	}
 
 	// Global Config guarantees RequireTLS is not nil.
-	if *n.conf.RequireTLS {
+	if *n.conf.RequireTLS && !useImplicitTLS {
```
Spaceman1701, line 189: this subtle change in behavior - It seems like the logic on main will enter this branch regardless of the port (and thus regardless of whether we're using implicit tls). Was this change intentional?

Spaceman1701 (commented):
Thanks for opening this PR!

One general comment: Would you mind adding a summary to the PR? I'm aware that this is for #4757, but it helps discoverability if the PR has a descriptive summary.

notify/email/email.go
Gintoms: yes

Spaceman1701: So on one hand, it's silly to use `STARTTLS` on an existing secure connection... On the other hand, I'm a little worried about changing this behavior out from under people.

Do you know what the spec says about issuing `STARTTLS` on a secure connection? If this is defined as a no-op, I'm ok with the behavior change.

Gintoms: > So on one hand, it's silly to use `STARTTLS` on an existing secure connection... On the other hand, I'm a little worried about changing this behavior out from under people.
> 
> Do you know what the spec says about issuing `STARTTLS` on a secure connection? If this is defined as a no-op, I'm ok with the behavior change.

Regarding the [spec](https://datatracker.ietf.org/doc/html/rfc3207), RFC 3207 Section 4 explicitly forbids this behavior:
"A client MUST NOT attempt to start a TLS session if a TLS session is already active."

Spaceman1701: I think RFC 3207 is referring to sending `STARTTLS` more than once per session, not issuing `STARTTLS` on an implicit TLS connection. 

That being said, [RFC 8314](https://datatracker.ietf.org/doc/html/rfc8314) doesn't even seem to require a client which provides `STARTTLS` over cleartext also implement it on the implicit TLS port, so I'm inclined to agree that the current implementation is wrong.

config/notifiers.go
Spaceman1701: Two small bikesheds: I think this should be called `ForceImplicitTLS` since implicit TLS is _always_ enabled for port 465. 

I also think this should be available as a global setting (like `smtp_force_implicit_tls`) since I think it's common to configure one smtp configuration for all email receivers. 

https://github.com/prometheus/alertmanager/blob/3a2c66e7acff848f5ae480120e64fcee29d9d366/config/config.go#L984-L994

Gintoms: > Two small bikesheds: I think this should be called `ForceImplicitTLS` since implicit TLS is _always_ enabled for port 465.
> 
> I also think this should be available as a global setting (like `smtp_force_implicit_tls`) since I think it's common to configure one smtp configuration for all email receivers.
> 
> https://github.com/prometheus/alertmanager/blob/3a2c66e7acff848f5ae480120e64fcee29d9d366/config/config.go#L984-L994

Renamed to ForceImplicitTLS and added global config support as suggested.

nutmos:
Hi, just wondering which field name should I use? I saw the doc uses `implicit_tls` while I saw from this commit, it should be `force_implicit_tls`.

Spaceman1701:
> Hi, just wondering which field name should I use? I saw the doc uses `implicit_tls` while I saw from this commit, it should be `force_implicit_tls`.

Hi! It looks like the docs are wrong - the field is `force_implicit_tls`. I'll open a PR to fix the docs later today. Sorry about that!

PR #4819, opened 2025-12-17
feat(provider): implement per-alert limits

siavashs:
## Add new limit package with bucket

Add a new limit package with generic bucket implementation.
This can be used for example to limit the number of alerts in memory.

Benchmarks:
```go
goos: darwin
goarch: arm64
pkg: github.com/prometheus/alertmanager/limit
cpu: Apple M3 Pro
BenchmarkBucketUpsert/EmptyBucket-12  	 8816954	       122.4 ns/op	      56 B/op	       2 allocs/op
BenchmarkBucketUpsert/AddToFullBucketWithExpiredItems-12         	 9861010	       123.0 ns/op	      56 B/op	       2 allocs/op
BenchmarkBucketUpsert/AddToFullBucketWithActiveItems-12          	 8343778	       143.6 ns/op	      56 B/op	       2 allocs/op
BenchmarkBucketUpsert/UpdateExistingItem-12                     	10107787	       118.9 ns/op	      56 B/op	       2 allocs/op
BenchmarkBucketUpsert/MixedWorkload-12                           	 9436174	       126.0 ns/op	      56 B/op	       2 allocs/op
BenchmarkBucketUpsertScaling/BucketSize_10-12                    	10255278	       115.4 ns/op	      56 B/op	       2 allocs/op
BenchmarkBucketUpsertScaling/BucketSize_50-12                    	10166518	       117.1 ns/op	      56 B/op	       2 allocs/op
BenchmarkBucketUpsertScaling/BucketSize_100-12                   	10457394	       115.0 ns/op	      56 B/op	       2 allocs/op
BenchmarkBucketUpsertScaling/BucketSize_500-12                   	 9644079	       115.2 ns/op	      56 B/op	       2 allocs/op
BenchmarkBucketUpsertScaling/BucketSize_1000-12                  	10426184	       116.6 ns/op	      56 B/op	       2 allocs/op
BenchmarkBucketUpsertConcurrent-12                               	 5796210	       216.3 ns/op	     406 B/op	       5 allocs/op
PASS
ok  	github.com/prometheus/alertmanager/limit	15.497s
```

## Implement per-alert limits

Use the new limit module to add optional per alert-name limits.
The metrics for limited alerts can be enabled using `alerts-limited-metric` feature flag.

---

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -1,4 +1,4 @@
-// Copyright 2015 Prometheus Team
+// Copyright The Prometheus Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
@@ -143,6 +143,7 @@ func run() int {
 		maxSilences                 = kingpin.Flag("silences.max-silences", "Maximum number of silences, including expired silences. If negative or zero, no limit is set.").Default("0").Int()
 		maxSilenceSizeBytes         = kingpin.Flag("silences.max-silence-size-bytes", "Maximum silence size in bytes. If negative or zero, no limit is set.").Default("0").Int()
 		alertGCInterval             = kingpin.Flag("alerts.gc-interval", "Interval between alert GC.").Default("30m").Duration()
+		perAlertNameLimit           = kingpin.Flag("alerts.per-alertname-limit", "Maximum number of alerts per alertname. If negative or zero, no limit is set.").Default("0").Int()
 		dispatchMaintenanceInterval = kingpin.Flag("dispatch.maintenance-interval", "Interval between maintenance of aggregation groups in the dispatcher.").Default("30s").Duration()
 		DispatchStartDelay          = kingpin.Flag("dispatch.start-delay", "Minimum amount of time to wait before dispatching alerts. This option should be synced with value of --rules.alert.resend-delay on Prometheus.").Default("0s").Duration()
 
@@ -345,7 +346,16 @@ func run() int {
 		go peer.Settle(ctx, *gossipInterval*10)
 	}
 
-	alerts, err := mem.NewAlerts(context.Background(), marker, *alertGCInterval, nil, logger, prometheus.DefaultRegisterer)
+	alerts, err := mem.NewAlerts(
+		context.Background(),
+		marker,
+		*alertGCInterval,
+		*perAlertNameLimit,
+		nil,
+		logger,
+		prometheus.DefaultRegisterer,
+		ff,
+	)
 	if err != nil {
 		logger.Error("error creating memory provider", "err", err)
 		return 1
--- a/docs/alertmanager.md
+++ b/docs/alertmanager.md
@@ -72,3 +72,17 @@ Alertmanager supports configuration to create a cluster for high availability.
 This can be configured using the [--cluster-*](https://github.com/prometheus/alertmanager#high-availability) flags.
 
 It's important not to load balance traffic between Prometheus and its Alertmanagers, but instead, point Prometheus to a list of all Alertmanagers.
+
+## Alert limits (optional)
+
+Alertmanager supports configuration to limit the number of active alerts per alertname.
+This can be configured using the [--alerts.per-alertname-limit] flag.
+
+When the limit is reached any new alerts are dropped, heartbeats from already know alerts are processed.
+The known alert (fingerprint) automatically expire to make room for new alerts.
+
+This feature is useful when an unexpected high number of instances of the same alert are sent to Alertmanager.
+Limiting the number of alerts per alertname can prevent reliability issues and avoid alert receivers from being flooded.
+
+The `alertmanager_alerts_limited_total` metric shows the total number of alerts that were dropped due to per alert name limit.
+Enabling the `alert-names-in-metrics` feature flag will add the `alertname` label to the metric.
--- a/provider/mem/mem.go
+++ b/provider/mem/mem.go
@@ -15,6 +15,7 @@ package mem
 
 import (
 	"context"
+	"errors"
 	"log/slog"
 	"sync"
 	"time"
@@ -27,6 +28,7 @@ import (
 	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/trace"
 
+	"github.com/prometheus/alertmanager/featurecontrol"
 	"github.com/prometheus/alertmanager/provider"
 	"github.com/prometheus/alertmanager/store"
 	"github.com/prometheus/alertmanager/types"
@@ -52,7 +54,10 @@ type Alerts struct {
 
 	logger     *slog.Logger
 	propagator propagation.TextMapPropagator
+	flagger    featurecontrol.Flagger
 
+	alertsLimit             prometheus.Gauge
+	alertsLimitedTotal      *prometheus.CounterVec
 	subscriberChannelWrites *prometheus.CounterVec
 }
 
@@ -79,6 +84,23 @@ type listeningAlerts struct {
 func (a *Alerts) registerMetrics(r prometheus.Registerer) {
 	r.MustRegister(&alertsCollector{alerts: a})
 
+	a.alertsLimit = promauto.With(r).NewGauge(prometheus.GaugeOpts{
+		Name: "alertmanager_alerts_per_alert_limit",
+		Help: "Current limit on number of alerts per alert name",
+	})
+
+	labels := []string{}
+	if a.flagger.EnableAlertNamesInMetrics() {
+		labels = append(labels, "alertname")
+	}
+	a.alertsLimitedTotal = promauto.With(r).NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "alertmanager_alerts_limited_total",
+			Help: "Total number of alerts that were dropped due to per alert name limit",
+		},
+		labels,
+	)
+
 	a.subscriberChannelWrites = promauto.With(r).NewCounterVec(
 		prometheus.CounterOpts{
 			Name: "alertmanager_alerts_subscriber_channel_writes_total",
@@ -89,25 +111,44 @@ func (a *Alerts) registerMetrics(r prometheus.Registerer) {
 }
 
 // NewAlerts returns a new alert provider.
-func NewAlerts(ctx context.Context, m types.AlertMarker, intervalGC time.Duration, alertCallback AlertStoreCallback, l *slog.Logger, r prometheus.Registerer) (*Alerts, error) {
+func NewAlerts(
+	ctx context.Context,
+	m types.AlertMarker,
+	intervalGC time.Duration,
+	perAlertNameLimit int,
+	alertCallback AlertStoreCallback,
+	l *slog.Logger,
+	r prometheus.Registerer,
+	flagger featurecontrol.Flagger,
+) (*Alerts, error) {
 	if alertCallback == nil {
 		alertCallback = noopCallback{}
+	}
+
+	if perAlertNameLimit > 0 {
+		l.Info("per alert name limit enabled", "limit", perAlertNameLimit)
+	}
+
+	if flagger == nil {
+		flagger = featurecontrol.NoopFlags{}
 	}
 
 	ctx, cancel := context.WithCancel(ctx)
 	a := &Alerts{
 		marker:     m,
-		alerts:     store.NewAlerts(),
+		alerts:     store.NewAlerts().WithPerAlertLimit(perAlertNameLimit),
 		cancel:     cancel,
 		listeners:  map[int]listeningAlerts{},
 		next:       0,
 		logger:     l.With("component", "provider"),
 		propagator: otel.GetTextMapPropagator(),
 		callback:   alertCallback,
+		flagger:    flagger,
 	}
 
 	if r != nil {
 		a.registerMetrics(r)
+		a.alertsLimit.Set(float64(perAlertNameLimit))
 	}
 
 	go a.gcLoop(ctx, intervalGC)
@@ -271,7 +312,14 @@ func (a *Alerts) Put(ctx context.Context, alerts ...*types.Alert) error {
 		}
 
 		if err := a.alerts.Set(alert); err != nil {
-			a.logger.Error("error on set alert", "err", err)
+			a.logger.Warn("error on set alert", "alertname", alert.Name(), "err", err)
+			if errors.Is(err, store.ErrLimited) {
+				labels := []string{}
+				if a.flagger.EnableAlertNamesInMetrics() {
+					labels = append(labels, alert.Name())
+				}
+				a.alertsLimitedTotal.WithLabelValues(labels...).Inc()
+			}
 			continue
 		}
 
--- a/store/store.go
+++ b/store/store.go
@@ -21,8 +21,12 @@ import (
 
 	"github.com/prometheus/common/model"
 
+	"github.com/prometheus/alertmanager/limit"
 	"github.com/prometheus/alertmanager/types"
 )
+
+// ErrLimited is returned if a Store has reached the per-alert limit.
+var ErrLimited = errors.New("alert limited")
 
 // ErrNotFound is returned if a Store cannot find the Alert.
 var ErrNotFound = errors.New("alert not found")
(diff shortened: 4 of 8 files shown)


limit/bucket.go
```diff
@@ -0,0 +1,148 @@
...
+
+// item represents a value and its priority based on time.
+type item[V any] struct {
+	value    V
+	priority time.Time
+	index    int
+}
+
+// expired returns true if the item is expired (priority is in the past).
+func (i *item[V]) expired() bool {
```
Spaceman1701: tiny nitpick: I think this ought to take an `at time.Time` rather than call `time.Now`. This makes things easier to test and is helpful if we ever want to call `expired` in a loop.

limit/bucket.go
```diff
@@ -0,0 +1,148 @@
...
+		capacity: capacity,
+	}
+}
+
+// Upsert tries to add a new value and its priority to the bucket.
+// If the value is already in the bucket, its priority is updated.
+// If the bucket is not full, the new value is added.
+// If the bucket is full, oldest expired item is evicted based on priority and the new value is added.
+// Otherwise the new value is ignored and the method returns false.
+func (b *Bucket[V]) Upsert(value V, priority time.Time) (ok bool) {
```
Spaceman1701, line 117: on the flip side, I think this method should _not_ take an `at` since it'd be an error to call it with any time other than now (so no change from how it's written).

siavashs: With the current use-case, the time argument is the Alert `EndsAt` value which is in future so we have to pass it to the bucket.

Spaceman1701: Understood, to be clear: my comment was just in reference to my other comment. I wanted to make sure to communicate that I don't think we should propagate an `at time.Time` up through the entire call stack. 

More generally, when it would be an error to pass a time other than `now`, I think functions should not accept an `at`. When it would not be an error, I think they should. My two comments were meant to ask for `expired` to gain an `at` and `Upsert` to stay as it is now.

docs/alertmanager.md
```diff
@@ -72,3 +72,11 @@ Alertmanager supports configuration to create a cluster for high availability.
 This can be configured using the [--cluster-*](https://github.com/prometheus/alertmanager#high-availability) flags.
 
 It's important not to load balance traffic between Prometheus and its Alertmanagers, but instead, point Prometheus to a list of all Alertmanagers.
+
+## Alert limits (optional)
+
+Alertmanager supports configuration to limit the number of alerts per alertname.
```
SuperQ: Minor wording suggestion to make it more clear.

```suggestion
Alertmanager supports configuration to limit the number of active alerts per alertname.
```

docs/alertmanager.md
```diff
@@ -72,3 +72,11 @@ Alertmanager supports configuration to create a cluster for high availability.
...
+This can be configured using the [--alerts.per-alert-limit] flag.
```
SuperQ: Can you describe a bit about what happens when the limit is reached?

cmd/alertmanager/main.go
```diff
@@ -143,6 +143,7 @@ func run() int {
 		maxSilences                 = kingpin.Flag("silences.max-silences", "Maximum number of silences, including expired silences. If negative or zero, no limit is set.").Default("0").Int()
 		maxSilenceSizeBytes         = kingpin.Flag("silences.max-silence-size-bytes", "Maximum silence size in bytes. If negative or zero, no limit is set.").Default("0").Int()
 		alertGCInterval             = kingpin.Flag("alerts.gc-interval", "Interval between alert GC.").Default("30m").Duration()
+		perAlertLimit               = kingpin.Flag("alerts.per-alert-limit", "Maximum number of alerts per alert name. If negative or zero, no limit is set.").Default("0").Int()
```
SuperQ: I think I figured out what my confusion was. This is a limit per alert name, not "alert per alert".

```suggestion
		perAlertNameLimit               = kingpin.Flag("alerts.per-alertname-limit", "Maximum number of alerts per alert name. If negative or zero, no limit is set.").Default("0").Int()
```

store/store.go
```diff
@@ -73,9 +91,9 @@ func (a *Alerts) Run(ctx context.Context, interval time.Duration) {
 func (a *Alerts) GC() []types.Alert {
```
ultrotter, line 91: Should we also at GC() clean buckets that might have become empty?

```
  for name, bucket := range a.limits {
      if bucket.items.Len() == 0 {
          delete(a.limits, name)
      }
  }
```

docs/alertmanager.md
```diff
@@ -72,3 +72,11 @@ Alertmanager supports configuration to create a cluster for high availability.
...
+
+This feature is useful when unexpected high number of instances of the same alert are sent to Alertmanager.
```
ultrotter: nit: 'an unexpected high number' ?

docs/alertmanager.md
```diff
@@ -72,3 +72,11 @@ Alertmanager supports configuration to create a cluster for high availability.
...
+Limiting the number of alerts per alertname can prevent reliability issues and avoid flooding alert receivers.
```
ultrotter: We can add: When the limit is reached new alerts are not sent, until some previous ones resolve, bringing the number back below the limit

store/store.go
ultrotter: Also for GC purposes, are we worried that potentially all buckets could be full of expired alerts? e.g. if the limit is 1000 the buckets can have 1000 entries, and every time a new alert comes we will evict one, and add the new one, but in the meantime the other 999 entries never go away till the next alert comes? Maybe here at GC we should also go through the buckets and aggressively remove all expired alerts?

Spaceman1701: > Should we also at GC() clean buckets that might have become empty?

This is a good point. Without it, I think the limits map will grow indefinitely. 

> Also for GC purposes, are we worried that potentially all buckets could be full of expired alerts?

I also thought about this, but my thinking is that it's ~fine to keep the long list of alerts in exchange for not needing the GC to lock the entire limits map. But maybe that doesn't matter? If the GC has locked the `Alerts`, we can't ingest things anyway. I guess if we have 10k alert names and 10k limits per alert, that's `100000000` alerts which is like 800 megabytes! That's probably not ideal.

siavashs: Our current bouncer setup uses ~500-700MB of memory, which is acceptable for our setup but I can imagine it can be high in specific setup specially if there are high variation of alert names.
We could have a simpler GC which would empty the buckets if their latest alert has expired in a configurable `<duration>`.
I'll update the PR to add this option an the GC.

siavashs: Actually we do not need a `duration` or `at` variable, if the `latest` item in the bucket is expired at `now` we can empty the bucket and an empty bucket can be removed.
I updated the code based on this.

docs/alertmanager.md
siavashs: Added another paragraph about this.

docs/alertmanager.md
siavashs: Added another paragraph about this.

provider/mem/mem.go
```diff
@@ -271,7 +300,10 @@ func (a *Alerts) Put(ctx context.Context, alerts ...*types.Alert) error {
 		}
 
 		if err := a.alerts.Set(alert); err != nil {
-			a.logger.Error("error on set alert", "err", err)
+			a.logger.Error("error on set alert", "alertname", alert.Name(), "err", err)
```
SuperQ: Should we make this a `Warn()`?

provider/mem/mem.go
```diff
@@ -79,6 +82,19 @@ type listeningAlerts struct {
...
+		Name: "alertmanager_alerts_per_alert_limit",
+		Help: "Current limit on number of alerts per alert name",
+	})
+
+	a.alertsLimitedTotal = promauto.With(r).NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "alertmanager_alerts_limited_total",
+			Help: "Total number of alerts that were dropped due to per alert name limit",
+		},
+		[]string{"alertname"},
```
SuperQ: I worry that this will be a cardinality risk. Do we really need to have the label given the alertname is now logged for every alert of this conditino?

limit/bucket.go
```diff
@@ -0,0 +1,166 @@
...
+	defer b.mtx.Unlock()
+	if b.items.Len() == 0 {
+		return true
+	}
+
+	latest := b.items[b.items.Len()-1]
+	if latest.expired(time.Now()) {
+		// Remove all items from the heap and index.
+		b.items = b.items[:0]
+		b.index = make(map[V]*item[V], b.capacity)
```
Spaceman1701: I see that later in the code we `delete` the bucket from the limits map if this method returns `true`, so I'm not sure we should be making an allocation here. I don't know if the assignment to `b.items` causes an allocation, but definitely `make(map[V]*item[V], b.capacity)` will cause an allocation.

provider/mem/mem.go
ultrotter: Well with some luck there would be only very few misbehaving alerts in this condition? This is more a safeguard and you want to know which alert is to blame, rather than something that should regularly happen for big number of alerts, isn't it?

siavashs: Exactly, here is the cardinality of a similar alert at Cloudflare over past year:
image
image

Note that depending on the configuration, setting a limit too low would drop many alerts and therefore cardinality can increase.

limit/bucket.go
siavashs: Good catch, I updated the code to `clear()` the index map instead of recreating it.
We also avoid deleting the buckets from `store`, since empty buckets don't use much memory and same alert names might reappear again during runtime.

provider/mem/mem.go
siavashs: Done.

provider/mem/mem.go
SuperQ: > Well with some luck

"Hope is not a strategy". :grin:

SuperQ (changes requested):
After thinking about this, I'm going to say that we should not have `alertname` as a metric label.

Please remove this, it's just too risky for end users. We have logs for debugging real prod issues.

It also solves the zero set issues with only one hit.

limit/bucket.go
```diff
@@ -0,0 +1,166 @@
...
+	defer b.mtx.Unlock()
+	if b.items.Len() == 0 {
+		return true
+	}
+
+	latest := b.items[b.items.Len()-1]
+	if latest.expired(time.Now()) {
+		// Remove all items from the heap and index.
+		b.items = b.items[:0]
+		clear(b.index)
```
ultrotter: I don't think we need either of these lines, because if the bucket itself has been deleted from the map and is unreachable, the gc will not count the pointers in it as active, and whatever they point to will be cleared anyway, so we don't need to clear the content of them... Should we check? Or remove trhe two lines?

Spaceman1701: I agree this isn't necessary - the bucket isn't valid after delete is called

Spaceman1701:
> After thinking about this, I'm going to say that we should not have `alertname` as a metric label.
> 
> Please remove this, it's just too risky for end users. We have logs for debugging real prod issues.
> 
> It also solves the zero set issues with only one hit.

I think it'd be useful to have this so we can know which alerts are being dropped. Maybe it could be behind a flag?

Since prometheus keeps the `ALERTS` metric, there's already at least one timeseries per alert in the TSDB.

siavashs:
> After thinking about this, I'm going to say that we should not have `alertname` as a metric label.
> 
> Please remove this, it's just too risky for end users. We have logs for debugging real prod issues.
> 
> It also solves the zero set issues with only one hit.

The use-case we have at Cloudflare is to notify about these dropped alerts caused by specific alert flooding the pipeline so the owner of the alert can fix the alert expression or improve it to avoid so many instances of the same alert, etc.

siavashs:
> > After thinking about this, I'm going to say that we should not have `alertname` as a metric label.
> > Please remove this, it's just too risky for end users. We have logs for debugging real prod issues.
> > It also solves the zero set issues with only one hit.
> 
> I think it'd be useful to have this so we can know which alerts are being dropped. Maybe it could be behind a flag?
> 
> Since prometheus keeps the `ALERTS` metric, there's already at least one timeseries per alert in the TSDB.

By default no limits are applied and therefore there are no metrics for dropped alerts, but I can put the metric behind a feature flag.

limit/bucket.go
siavashs: I removed the bucket deletion logic from `store`, so we will keep empty buckets to reuse them if necessary.

siavashs:
Added `alerts-limited-metric` feature flag.

provider/mem/mem.go
siavashs: Added `alerts-limited-metric feature` flag.

limit/bucket.go
ultrotter: That would still cause the store to potentially grow indefinitely. How about we delete the buckets, but increase the len of the condition for example to `if latest.expired(time.Now()+1h)`  instead if it expired now, if it expired more than 1h ago or so, then we delete the bucket altogether because we can definitely afford a reallocation for something firing so infrequently?

ultrotter: Also if you just clear the index but leave it there it still uses all the space anyway? GC doesn't run so frequently, so maybe just deleting it is fine, it will only happen to alerts that are not so infrequent and then we can afford the allocation...

Spaceman1701:
> > After thinking about this, I'm going to say that we should not have `alertname` as a metric label.
> > Please remove this, it's just too risky for end users. We have logs for debugging real prod issues.
> > It also solves the zero set issues with only one hit.
> 
> The use-case we have at Cloudflare is to notify about these dropped alerts caused by specific alert flooding the pipeline so the owner of the alert can fix the alert expression or improve it to avoid so many instances of the same alert, etc.

Yep, we do a very similar thing at HRT, but our limiter is a standalone service that ingests alerts first.

limit/bucket.go
siavashs: > That would still cause the store to potentially grow indefinitely.

Max store bucket usage would be equal to cardinality of `alertname` for dropped alerts.
We only store fingerprints which are `uint64`s and `time.Time` in the buckets so footprint is quite small.
Since alert limits are not enabled by default, we don't need to prematurely optimise this IMHO, we can get feedback from users enabling the feature and maybe then do further optimisations.

I'm still open to adding more logic for GC.

siavashs:
> > > After thinking about this, I'm going to say that we should not have `alertname` as a metric label.
> > > Please remove this, it's just too risky for end users. We have logs for debugging real prod issues.
> > > It also solves the zero set issues with only one hit.
> > 
> > 
> > The use-case we have at Cloudflare is to notify about these dropped alerts caused by specific alert flooding the pipeline so the owner of the alert can fix the alert expression or improve it to avoid so many instances of the same alert, etc.
> 
> Yep, we do a very similar thing at HRT, but our limiter is a standalone service that ingests alerts first.

Same at Cloudflare.

provider/mem/mem.go
SuperQ: That feature flag is a bit confusing.

How about `enable-alerts-limited-alertname-label`?

siavashs: I made the feature name and implementation similar with what we have for `receiver` in `notify` package.
We have a metric by default without any labels, enabling the feature `alert-names-in-metrics` will add the `alertname` dimention:

```
# HELP alertmanager_alerts_limited_total Total number of alerts that were dropped due to per alert name limit
# TYPE alertmanager_alerts_limited_total counter
alertmanager_alerts_limited_total 1
```
vs.
```
# HELP alertmanager_alerts_limited_total Total number of alerts that were dropped due to per alert name limit
# TYPE alertmanager_alerts_limited_total counter
alertmanager_alerts_limited_total{alertname="foo"} 1
```

SuperQ: Thanks! Last nit, can you add the feature flag to the docs?

siavashs: Added the docs, thanks for the suggestion!

SuperQ:
@ultrotter any more comments? Otherwise I think this is ready to merge.

limit/bucket.go
Spaceman1701: I don't think is a premature optimization - as it's written, there's a memory leak. Without restarting, the limiter will allocate but never free buckets. I really don't think it should do that. 

What is the motivation for trying to reuse buckets?

siavashs: Basically avoiding allocations.
But we can instead avoid doing GC within the bucket and just drop it.

siavashs: I updated the code again, we basically do this:

```go
	for alertName, bucket := range a.limits {
		if bucket.IsStale() {
			delete(a.limits, alertName)
		}
	}
```

ultrotter:
Let's merge it, I think we may want to still improve the GC a little. But I
believe it can be done as a separate commit, at this point. We'll chat
about it later today but I think in the meantime we can progress!

Thanks,

Guido

SuperQ:
Ok, going to merge this so we can get it into 0.31.

PR #4823, opened 2025-12-21
Support global Telegram bot token

guoard:
This PR adds global configuration support for Telegram bot tokens in Alertmanager. Until now, `telegram_bot_token` and `telegram_bot_token_file` could only be set per receiver, which could lead to redundant configuration when multiple receivers used the same bot.


guoard:
@SoloJacobs Thank you for your time and review.

Done.

config/config.go
```diff
@@ -699,6 +703,13 @@ func (c *Config) UnmarshalYAML(unmarshal func(any) error) error {
 			if telegram.APIUrl == nil {
 				telegram.APIUrl = c.Global.TelegramAPIUrl
 			}
+			if telegram.BotToken == "" && len(telegram.BotTokenFile) == 0 {
+				if c.Global.TelegramBotToken == "" && len(c.Global.TelegramBotTokenFile) == 0 {
+					return errors.New("missing bot_token or bot_token_file on telegram_config")
```
TheMeier, line 661: Others mostly us a message like "no global Telegram token set either inline or in a file" , there are already other variations like: "no msteams webhook URL or URLFile provided"

We should come up with a unified wording for these errors, maybe also including the receiver name.

SoloJacobs:
@guoard There are some merge conflicts. Could you resolve them? I can also do, but would have to pick your commits to a branch of my own.

guoard:
> @guoard There are some merge conflicts. Could you resolve them? I can also do, but would have to pick your commits to a branch of my own.

Done.

PR #4825, opened 2025-12-23
Fix: Allow empty group_by to override parent route

### LINKED ISSUE #4221: Overriding group_by to an empty list is ignored
If I specify group_by on the root route, it is inherited by child routes, as expected. If I specify group_by on a child route, it correctly overrides the inherited group_by. However, if a child attempts to override using `group_by: []`, this is ignored.

adity-a34:
Fixes #4221

Child routes with `group_by: []` were incorrectly inheriting parent route's 
group_by configuration instead of overriding it. This fix initializes `GroupBy` 
as an empty slice when `GroupByStr` is explicitly set to empty, allowing child 
routes to properly disable grouping.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -1720,6 +1720,8 @@ receivers:
 	require.Len(t, cfg.Route.GroupBy, 2)
 	require.NotNil(t, cfg.Route.Routes[0].GroupBy)
 	require.Empty(t, cfg.Route.Routes[0].GroupBy)
+}
+
 func TestWechatNoAPIURL(t *testing.T) {
 	_, err := LoadFile("testdata/conf.wechat-no-api-secret.yml")
 	if err == nil {


config/config_test.go
```diff
@@ -1593,3 +1593,26 @@ func TestInhibitRuleEqual(t *testing.T) {
...
+      group_by: []
+receivers:
+  - name: 'default'
+  - name: 'database-team'
+`
+	cfg, err := Load(in)
+	require.NoError(t, err)
+	// Root route should have group_by labels.
+	require.Len(t, cfg.Route.GroupBy, 2)
+	// Child route should have empty non-nil slice (explicit override, not inheritance).
```
SoloJacobs: I don't think these comments are necessary.

config/config_test.go
```diff
@@ -1593,3 +1593,26 @@ func TestInhibitRuleEqual(t *testing.T) {
...
 	require.Equal(t, []string{"qux🙂", "corge"}, r.Equal)
 }
+
+func TestGroupByEmptyOverride(t *testing.T) {
+	in := `
+route:
+  receiver: 'default'
+  group_by: ['alertname', 'cluster']
+  routes:
+    - match:
```
SoloJacobs: `match` and `receiver` can be dropped, since this is not a real configuration.

SoloJacobs (changes requested):
Did you check that this fix resolves the issue completely, e.g., by creating a small test setup?

adity-a34:
Hi @SoloJacobs 
> Did you check that this fix resolves the issue completely, e.g., by creating a small test setup?

I tested for the unit test and go tests for build and gofmt etc. However I did not create a dedicated test setup/script, but since you mentioned it, I did that too

**validation script confirms:**
- Parent route: `group_by = [alertname cluster]` (2 labels) ✅
- Child route: `group_by = []` (empty, **not nil**) ✅
- Key proof: `Is nil? false` - the fix distinguishes empty from inheritance ✅

The fix correctly resolves #4221 by initializing `GroupBy` as an empty slice when `group_by: []` is explicitly set, allowing child routes to override parent configuration.

I've also simplify the test config as suggested (removing unnecessary match/receiver fields).

SoloJacobs (approved):
Very nice, well done 🙂 Just a reminder, due to the current time constraints of maintainers, all changes take some time to be merged.

PR #4827, opened 2025-12-26
ci: add macOS test job alongside Linux CI

### LINKED ISSUE #4754: Adopt github/workflows to run tests on supported platforms
We build AM with
```
      - uses: ./.github/promci/actions/build
        with:
          promu_opts: "-p linux/amd64 -p windows/amd64 -p linux/arm64 -p darwin/amd64 -p darwin/arm64 -p linux/386"
```
but only test it with:
```
  test:
    name: Test
    runs-on: ubuntu-latest
```

mihir-dixit2k27:
### Summary
This PR extends Alertmanager CI by adding a macOS test job alongside the existing Linux-based tests.

Linux tests continue to run inside the containerized `golang-builder` image with required mail service dependencies, while macOS tests run natively without containers or service services, respecting GitHub Actions runner limitations.

### Motivation
Alertmanager binaries are built for multiple platforms, but CI tests were previously executed only on `ubuntu-latest`. Adding macOS test coverage improves confidence across supported platforms without changing existing Linux CI behavior.

### Notes
- No changes to existing Linux CI logic
- macOS job runs without containers or service dependencies
- Windows testing can be explored in a follow-up if desired

Fixes #4754

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -82,11 +82,32 @@ endif
 PREFIX                  ?= $(shell pwd)
 BIN_DIR                 ?= $(shell pwd)
 DOCKER_IMAGE_TAG        ?= $(subst /,-,$(shell git rev-parse --abbrev-ref HEAD))
-DOCKERFILE_PATH         ?= ./Dockerfile
 DOCKERBUILD_CONTEXT     ?= ./
 DOCKER_REPO             ?= prom
 
+# Check if deprecated DOCKERFILE_PATH is set
+ifdef DOCKERFILE_PATH
+$(error DOCKERFILE_PATH is deprecated. Use DOCKERFILE_VARIANTS ?= $(DOCKERFILE_PATH) in the Makefile)
+endif
+
 DOCKER_ARCHS            ?= amd64
+DOCKERFILE_VARIANTS     ?= Dockerfile $(wildcard Dockerfile.*)
+
+# Function to extract variant from Dockerfile label.
+# Returns the variant name from io.prometheus.image.variant label, or "default" if not found.
+define dockerfile_variant
+$(strip $(or $(shell sed -n 's/.*io\.prometheus\.image\.variant="\([^"]*\)".*/\1/p' $(1)),default))
+endef
+
+# Check for duplicate variant names (including default for Dockerfiles without labels).
+DOCKERFILE_VARIANT_NAMES := $(foreach df,$(DOCKERFILE_VARIANTS),$(call dockerfile_variant,$(df)))
+DOCKERFILE_VARIANT_NAMES_SORTED := $(sort $(DOCKERFILE_VARIANT_NAMES))
+ifneq ($(words $(DOCKERFILE_VARIANT_NAMES)),$(words $(DOCKERFILE_VARIANT_NAMES_SORTED)))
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,14 +1,15 @@
----
 name: CI
 on:  # yamllint disable-line rule:truthy
   pull_request:
   workflow_call:
+
```
ultrotter: Do we need an extra empty line?

.github/workflows/ci.yml
```diff
@@ -1,14 +1,15 @@
...
 jobs:
   test_frontend:
     name: Test alertmanager frontend
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      # Updated to official v4 hash
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```
ultrotter: Why from 5.0.0 to 4.1.1 ?

ultrotter:
approving to run CI, still. Comments stay valid.

.github/workflows/ci.yml
mihir-dixit2k27: Good catch! I removed the unnecessary blank line in the latest commit.

.github/workflows/ci.yml
mihir-dixit2k27: The comment # v5.0.0 appeared to be a typo or incorrect, as actions/checkout v5 has not been released yet. I updated it to use the official SHA for v4.1.1 to ensure we are using a valid, stable version

TheMeier: I am confused https://github.com/actions/checkout/commit/08c6903cd8c0fde910a37f88322edcfb5dd907a8 is the v5 prepare commit. 
And there certainly is a v5 https://github.com/actions/checkout/releases/tag/v5.0.0

mihir-dixit2k27:
@TheMeier I saw that main was updated to v6.0.1 yesterday (#4846).

I’ve resolved the conflicts by adopting the new promci setup for the Linux jobs and manually updating my new macOS job to use the matching checkout@v6.0.1 hash (8e8c483...).

All jobs should now be consistent with the latest main branch.

TheMeier (approved):

SuperQ:
I don't really see the point of running MacOS tests for Alertmanager. As far as I know there's no platform specific code in alertmanager that needs to be tested.

mihir-dixit2k27:
Thanks for the review @SuperQ!

I picked this up to address issue #4754, which noted that while we build binaries for macOS, we don't actually test them in CI.

Interestingly, the current run failed on macOS (while passing on Linux), which suggests there might be some platform-specific behaviors (likely in filesystem or networking tests) that are currently going unchecked.

I'm happy to debug this failure to get it passing, but if you feel #4754 is no longer relevant, I can close this PR

SuperQ:
Unless we find an actual platform-specific problem, I don't think it's worth the extra CPU cost to constantly run tests on every PR. Even if the Prometheus project gets free resources for CI, it's still a bunch of waste.

mihir-dixit2k27:
Understood @SuperQ . Thanks for the clarification regarding the CI resource costs. I'll focus my efforts on the Slack integration PR and other issues instead. Thanks for the review!

PR #4829, opened 2025-12-27
mantine-ui: Add API client to manage data fetching and caching

sysadmind:
- Add tanstack/react-query for data fetching and caching
- Add data hooks for silences, alerts, and status
- Add server proxy for local development API requests

### DIFF
--- a/ui/mantine-ui/src/App.tsx
+++ b/ui/mantine-ui/src/App.tsx
@@ -30,14 +30,14 @@ export default function App() {
                     </Box>
                   }
                 >
-              {/* Main content will be rendered here by the Router */}
-              <Routes>
-                {/* Redirect the root path to the alerts page */}
-                {/* TODO(@sysadmind): This should take the fact that previous UI used /#/routeName */}
-                <Route path="/" element={<Navigate to="/alerts" replace />} />
-                <Route path="/alerts" element={<AlertsPage />} />
-                <Route path="/silences" element={<SilencesPage />} />
-              </Routes>
+                  {/* Main content will be rendered here by the Router */}
+                  <Routes>
+                    {/* Redirect the root path to the alerts page */}
+                    {/* TODO(@sysadmind): This should take the fact that previous UI used /#/routeName */}
+                    <Route path="/" element={<Navigate to="/alerts" replace />} />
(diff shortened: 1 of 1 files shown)


ui/mantine-ui/src/data/api.ts
```diff
@@ -0,0 +1,121 @@
...
+      const startTime = Date.now();
+
+      const res = await fetch(`${pathPrefix}/${API_PATH}${path}${queryString}`, {
+        cache: 'no-store',
+        credentials: 'same-origin',
+        signal,
+      });
+
+      if (!res.ok && !res.headers.get('content-type')?.startsWith('application/json')) {
+        // For example, Prometheus may send a 503 Service Unavailable response
```
juliusv: This comment still mentions Prometheus (not Alertmanager), just checking that this is the same in Alertmanager?

ui/mantine-ui/src/data/api.ts
```diff
@@ -0,0 +1,121 @@
...
+type QueryOptions = {
+  key?: QueryKey;
+  path: string;
+  params?: Record<string, string>;
+  enabled?: boolean;
+  refetchInterval?: false | number;
+  recordResponseTime?: (time: number) => void;
+};
+
+export const useAPIQuery = <T>({
```
juliusv, line 104: Maybe you'll also want to copy over `useSuspenseAPIQuery` from Prometheus, at least it's been very handy there for any of the normal pages where you first have to load the page's data and then show the page (not for loading data interactively, like for PromQL queries).

juliusv (approved):
Looks generally fine to me, although I didn't review the Alertmanager API types in detail - I trust you on those :)

ui/mantine-ui/src/data/api.ts
sysadmind: I wasn't familiar with the Suspense versions. Thanks for pointing that out. I have copied that over and updated the hooks to use it as well as added the Suspense and ErrorBoundary components in the tree. Really helpful.

juliusv: For sure, it's super helpful to avoid all those loading/error conditionals within the page components :)

juliusv:
:+1:

PR #4863, opened 2026-01-03
config: honor empty group_by override

### LINKED ISSUE #4221: Overriding group_by to an empty list is ignored

KyungHwanKim-devs:
## Summary
- Honor explicit `group_by: []` overrides during config unmarshal.
- Add route test coverage for empty group_by overrides.
- Update changelog.

## Testing
- make test

Fixes #4221.


TheMeier:
There is already a PR for this https://github.com/prometheus/alertmanager/pull/4825

PR #4865, opened 2026-01-05
use promauto to register silence state metrics

SungJin1212:
While #4700 migrated most metrics to use `promauto` for automatic registration, the `alertmanager_silences` metrics (Active, Pending, Expired) generated by `newSilenceMetricByState` were missed.

This PR migrate missed metric to use `promauto`.


SungJin1212:
@SuperQ 
Could you take a look when you have a time?

ultrotter (approved):

PR #2405, opened 2020-10-28
Added receiver SIGNL4

rons4:
Added SIGNL4 as an alerting option.

PR #4870, opened 2026-01-07
SIGNL4 Receiver Added

rons4:
SIGNL4 receiver added as discussed here: https://github.com/prometheus/alertmanager/pull/2405

Signed-off-by: Ronald Czachara <ron@signl4.com>

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -339,6 +339,9 @@ func resolveFilepaths(baseDir string, cfg *Config) {
 		for _, cfg := range receiver.PagerdutyConfigs {
 			cfg.HTTPConfig.SetDirectory(baseDir)
 		}
+		for _, cfg := range receiver.SIGNL4Configs {
+			cfg.HTTPConfig.SetDirectory(baseDir)
+		}
 		for _, cfg := range receiver.PushoverConfigs {
 			cfg.HTTPConfig.SetDirectory(baseDir)
 		}
@@ -601,6 +604,11 @@ func (c *Config) UnmarshalYAML(unmarshal func(any) error) error {
 					return errors.New("no global PagerDuty URL set")
 				}
 				pdc.URL = c.Global.PagerdutyURL
+			}
+		}
+		for _, s4 := range rcv.SIGNL4Configs {
+			if s4.HTTPConfig == nil {
+				s4.HTTPConfig = c.Global.HTTPConfig
 			}
 		}
 		for _, iio := range rcv.IncidentioConfigs {
@@ -1153,6 +1161,7 @@ type Receiver struct {
 	EmailConfigs      []*EmailConfig      `yaml:"email_configs,omitempty" json:"email_configs,omitempty"`
 	IncidentioConfigs []*IncidentioConfig `yaml:"incidentio_configs,omitempty" json:"incidentio_configs,omitempty"`
 	PagerdutyConfigs  []*PagerdutyConfig  `yaml:"pagerduty_configs,omitempty" json:"pagerduty_configs,omitempty"`
+	SIGNL4Configs     []*SIGNL4Config     `yaml:"signl4_configs,omitempty" json:"signl4_configs,omitempty"`
 	SlackConfigs      []*SlackConfig      `yaml:"slack_configs,omitempty" json:"slack_configs,omitempty"`
 	WebhookConfigs    []*WebhookConfig    `yaml:"webhook_configs,omitempty" json:"webhook_configs,omitempty"`
 	OpsGenieConfigs   []*OpsGenieConfig   `yaml:"opsgenie_configs,omitempty" json:"opsgenie_configs,omitempty"`
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -78,6 +78,13 @@ var (
 		"num_resolved": `{{ .Alerts.Resolved | len }}`,
 	}
 
+	// DefaultSignl4Config defines default values for SIGNL4 configurations.
+	DefaultSIGNL4Config = SIGNL4Config{
+		NotifierConfig: NotifierConfig{
+			VSendResolved: true,
+		},
+	}
+
 	// DefaultPagerdutyConfig defines default values for PagerDuty configurations.
 	DefaultPagerdutyConfig = PagerdutyConfig{
 		NotifierConfig: NotifierConfig{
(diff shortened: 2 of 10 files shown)


SoloJacobs (commented):
Hi @rons4,

we discussed your PR in a recent bug-scrub, and it seems that we cannot take your PR :-( . I apologize for that, I value the time of contributors and I want to ensure that it is spent well.

The following concerns were raised:

- There does not appear any way to setup a local instance of `SIGNL4` for development. In particular, we can't add integration tests. From what we can tell `SIGNL4` is not source available either.
- It is unclear what this integration adds beyond what is already available via: https://docs.signl4.com/integrations/prometheus/prometheus.html#how-to-integrate 
It seems that the Alertmanager webhook already covers the functionality sufficiently.
- There seems little interest Alertmanager users in this feature. This point can't really be resolved.

Some review remarks, which don't seem so relevant anymore:
* I like the general shape of this PR: well-documented, includes tests, no unnecessary complexity.
* You should use `PostJSON` over creating your own request. This is needed for instance for tracing.

Kind regards

rons4:
Thank you for your message and explanations, @SoloJacobs. Please allow me to comment from my side.

- Indeed, SIGNL4 is closed source. However, it can of course be tested using a trial version or even the available free plan. The integration proposed in the PR itself is open source. I have tested it and it works just fine. From my perspective, it is quite similar to existing receivers such as PagerDuty, OpsGenie, and VictorOps (all of which are also closed source).
- The basic functionality is certainly available via the webhook integration. A dedicated SIGNL4 receiver would make things more convenient for our joint customers and would also offer additional functionality, such as service / category support (alert color, icon, behavior), location information, etc. More features may follow in future versions. In addition, this would be a solid foundation for a later two-way integration, for example to support acknowledgements.
- We certainly have customers using SIGNL4 in conjunction with Prometheus / Alertmanager. Unfortunately, we cannot track the number when using the webhook, and to be honest, I forgot to implement tracking in the current PR. I would be happy to add it (it should be just one line, I believe).

Of course, it is entirely your decision, and I fully respect that.

However, from my point of view (admittedly a bit biased 😉), I really see a win–win–win situation here:
- You win because you gain an additional alerting receiver, while we take care of maintenance and support.
- Customers win because they can choose what fits them best. For some European customers, having a European alerting provider does indeed matter.
- And yes, we win too, because (potential) customers can more easily see what is possible and integrate SIGNL4 with more supported features out of the box.

So, if you feel the PR makes sense, I will be happy to adapt it, switch to PostJSON, and add usage tracking.

Thanks a lot for all your help.

SoloJacobs:
Hi @rons4 ,

I spent some time discussing this with other maintainers in the working group, and the verdict is a pretty firm no. There is a lot of positive things I can say about the points you raised. However,

* it is still not clear what this feature does, that the webhook receiver cannot do. What am I missing that you can't simply modify the receiver in SIGN4L to build the features you mention?
* As for the interest of the community, I suggest you open an issue in Alertmanager. This way people can find and upvote the feature set, which you are proposing.

I feel somewhat torn, whether to close this PR right away. Whether something is technically feasible via the webhook receiver is worth a discussion and as engineers should be able to achieve a shared understanding of the issue. On the other hand, it is looking increasingly likely that we can't accept this PR and I don't want to take more time of you than we already have. What do you think?

Kind regards

rons4:
Thanks again, @SoloJacobs. I really appreciate your feedback and that you look into this.

If I understand correctly, the webhook receiver sends in a fixed format. Otherwise, I guess that the functionality can be achieved using the webhook received is true also for the existing receivers like PagerDuty, OpsGenie, VictorOps, Pushover, Telegram, etc.

SIGNL4 can cope with a variety of formats and tries to display things nicely but ideally, we can use our own format (https://docs.signl4.com/integrations/webhook/webhook.html) along with some custom parameters.

There are some "control parameters" that allow display and control the alert in the right way, e.g. the following:
- Alert subject
- Alert message / description
- Service / category to control color, icon, certain behavior of the alert
- Location (longitude / latitude to render a map)
- Information to trigger / close alerts
- Information to allow later acknowledgements (SIGNL4 -> Alertmanager)
- Tracking information to know how many users are using this integration

While some if these might be convenience features, from my experience with integrations in other monitoring tools I can say that this really provides value for customers.

Just let me know what you think. If you are in favor, I will be more than happy to proceed and adapt the PR (PostJSON, tracking), otherwise I will open an issue / feature request and hope to proceed at some time in the future ;-)

In any case I thank you for all your support and input ....

SoloJacobs:
Yes, the webhook format is indeed fixed. However, I think you misunderstood my question, so let me spell out my assumptions. 

You control the endpoint and the JSON parsing. The payload includes the receiver name (e.g., `{"receiver": "signl4-us-east", ...}`). Couldn't your users simply associate the location metadata with the receiver name inside SIGNL4's configuration?

For example:
* User defines receiver `signl4-us-east` in Alertmanager.
* User maps `signl4-us-east` to `"Location: US-East"` in the SIGNL4 dashboard.
* Your endpoint receives the payload, looks up the receiver name, and injects the location before processing.

This workflow should cover the features in this PR.

As for the other features, those are not clear to me. What is meant by allowing acknowledgments? What does a user get out of that?

rons4:
Sorry for the misunderstanding @SoloJacobs.
With location I do not mean the endpoint, e.g. us-east, europe, etc. I mean actual coordinates to render a map as part of the alert. In this case the user can click the map and directly navigate to the location of the incident. This is for example useful for field service workers.

Maybe this sample payload makes is more transparent:

{
    "Title": "Alertmanager", - Some alert title.
    "Message": "Hello world.", - Some alert message.
    "Source": "Alertmanager", - A tag to allow usage tracking.
    "X-S4-Service": "Security", - A category name that allows to define icon, color and other behavior of the alert.
    "X-S4-Location": "40.6413111,-73.7781391", - Location date to render a position on the map.
    "X-S4-AlertingScenario": "multi_ack", - One ore multiple users need to acknowledge the alert.
    "X-S4-ExternalID": "INC091210" - External ID to allow later status updates or closing the alert.
    "X-S4-Status": "new" - Status, e.g. new, resolved, acknowledged.
}

I hope this makes it clearer. Besides this you can also use any custom parameters. The X-S4 parameters allow controlling the alerting procedure and behavior. Any fixed standard JSON is just something SIGNL4 tries to display as good as possible ;-)

If you should have any further questions, please just let me know.

SoloJacobs:
I will go ahead and rip off the bandage here. We will not accept this feature request for the aforementioned reasons. While no decision in software engineering is arguably ever final, we don't plan to support this currently. We hope Alertmanager and SIGNL4 can continue to work well together.

Please go ahead and open the issue, which we discussed.

Kind regards

rons4:
Hi @SoloJacobs, thanks and understood. We surly have joint sustomers and I am sure and I hope our pathes will cross again in the future.
I will open the issue and let's see how it goes.
All the best and thanks a lot for all your support and patience.

Issue #4609, opened 2025-10-15
HTTP request to JIRA API: unexpected status code 404:


alex-deploys:
I updated to the 0.29.0-rc.1 version

My alertmanager.yml looks like this:

```
  - name: 'jira'
    jira_configs:
        api_url: 'https://company.atlassian.net/rest/api/2/'
        api_type: 'cloud'
```

and I get this error in the logs

```
time=2025-10-15T11:22:38.145Z level=ERROR source=dispatch.go:363 msg="Notify for alerts failed" component=dispatcher num_alerts=1 err="jira/jira[0]: notify retry canceled due to unrecoverable error after 1 attempts: failed to lo
ok up existing issues: HTTP request to JIRA API: unexpected status code 404: <!DOCTYPE html>\n<html lang=\"en\">\n    <head>\n        <meta charset=\"utf-8\">\n        <meta name=\"description\" content=\"\">\n        <meta name
=\"viewport\" content=\"width=device-width, initial-scale=1\">\n        <title>Atlassian Cloud Notifications - Page Unavailable</title>\n        <link rel=\"stylesheet\" href=\"/_edge/resources/aui-prototyping.css\">\n    </head
>\n    <body>\n        <div id=\"page\">\n            <section id=\"content\" role=\"main\" class=\"aui-page-notification aui-page-size-medium\">\n                <div id=\"notifications\" class=\"aui-page-panel\">\n
        <div class=\"aui-page-header-inner\">\n                        <section class=\"aui-page-panel-content\">\n                            <h1>Page unavailable</h1>\n                            <div class=\"aui-page-notifica
tion-description\">\n                                Your Atlassian Cloud site is currently unavailable.\n                                <br/><br/>\n                                Please check <a href=\"https://status.atlassia
n.com\">Atlassian Status</a> for known problems.\n                                <br />\n                                If there are no known problems and your page hasn't appeared again in 5-10 minutes then please contact our
 support team.\n                            </div>\n                            <br>\n                            <p>\n                                <a href=\"https://status.atlassian.com\" class=\"aui-button aui-button-primar
y\">\n                                    Atlassian Status\n                                </a>\n                                &nbsp;\n                                <a href=\"https://support.atlassian.com\" class=\"aui-butt
on\">\n                                    Contact Support\n                                </a>\n                            </p>\n                            <br>\n                            <p style=\"color:#B3B3B3; font:nor
[...]

PR #4892, opened 2026-01-13
fix(jira): prevent hostname corruption in Cloud API URL replacement

alex-deploys:
## Description
This PR fixes a bug in the Jira Cloud integration where hostnames containing the digit '2' were incorrectly modified during API version replacement, resulting in 404 errors when attempting to create or search for issues.

## Problem
The current implementation uses `strings.Replace(url, "/2", "/3", 1)` to upgrade the API version from v2 to v3. However, this replaces the **first** occurrence of `/2` in the entire URL string. 

When a Jira Cloud instance has a hostname containing the digit '2' (e.g., `example2.atlassian.net`), the replacement incorrectly modifies the hostname to `example3.atlassian.net`, causing 404 errors.

**Example:**
- Input: `https://example2.atlassian.net/rest/api/2/search/jql`
- Current behavior: `https://example3.atlassian.net/rest/api/2/search/jql` ❌
- Expected behavior: `https://example2.atlassian.net/rest/api/3/search/jql` ✅

## Solution
Changed the replacement pattern from `/2` to the more specific `/rest/api/2/`, ensuring only the API version in the path is modified and not characters in the hostname.

## Testing
- Tested with a Jira Cloud instance at `acme2corp.atlassian.net`
- Previously failed with 404 errors when searching for existing issues
- After fix: successfully creates and searches for Jira issues

## Related
This addresses the Jira Cloud endpoint migration mentioned in the code comments, where `/rest/api/2/search` was deprecated in favor of `/rest/api/3/search/jql` (see https://developer.atlassian.com/changelog/#CHANGE-2046).

### DIFF
--- a/notify/jira/jira.go
+++ b/notify/jira/jira.go
@@ -283,7 +283,7 @@ func (n *Notifier) prepareSearchRequest(jql string) (issueSearch, string) {
 	}
 
 	if n.conf.APIType == "cloud" || n.conf.APIType == "auto" && strings.HasSuffix(n.conf.APIURL.Host, "atlassian.net") {
-		searchPath := strings.Replace(n.conf.APIURL.JoinPath("/search/jql").String(), "/2", "/3", 1)
+		searchPath := strings.Replace(n.conf.APIURL.JoinPath("/search/jql").String(), "/rest/api/2/", "/rest/api/3/", 1)
 		return requestBody, searchPath
 	}
 


alex-deploys:
Pinging @SoloJacobs @holger-waschke @tjhop because they were active on this file. Can you please check this?

holger-waschke:
seems like a reasonable fix to me

alex-deploys:
@holger-waschke thank you

I’m also linking this issue that I had when receiving the 404 errors and didn’t know why it was

https://github.com/prometheus/alertmanager/issues/4609

TheMeier:
I think this warrants some tests.

ultrotter (approved):

SuperQ:
Still seems fragile that it doesn't act only on the non-hostname part. But it's an improvement.

PR #4896, opened 2026-01-16
chore(api): update swagger

TheMeier:


TheMeier:
goswagger [removed](https://github.com/go-swagger/go-swagger/commit/50d4897a41d3cf2c6a394acd283041fd0473f0e8) `go` from their image. I guess one "solution" here is run this job with the golang image and install the swagger version inside the action :(

Makefile
```diff
@@ -52,13 +52,14 @@ template/email.tmpl: template/email.html
...
-	--rm \
-	-v $(shell pwd):/go/src/github.com/prometheus/alertmanager \
-	-w /go/src/github.com/prometheus/alertmanager quay.io/goswagger/swagger:v0.31.0
+SWAGGER_VERSION ?= v0.33.1
+SWAGGER := $(FIRST_GOPATH)/bin/swagger
 
-api/v2/models api/v2/restapi api/v2/client: api/v2/openapi.yaml
+$(SWAGGER):
+	mkdir -p $(FIRST_GOPATH)/bin
+	GOBIN=$(FIRST_GOPATH)/bin go install github.com/go-swagger/go-swagger/cmd/swagger@$(SWAGGER_VERSION)
```
siavashs: Let's use https://github.com/prometheus/alertmanager/blob/main/internal/tools/tools.go 
Then we can update the API generation the same way proto generation tools are installed and invoked.

TheMeier: I will have a look, I modled this approach after golangci-lint handling in Makefile.common

TheMeier: I did adapt it @siavashs

api/v2/client/general/get_status_responses.go
```diff
@@ -112,7 +113,7 @@ func (o *GetStatusOK) readResponse(response runtime.ClientResponse, consumer run
 	o.Payload = new(models.AlertmanagerStatus)
 
 	// response payload
-	if err := consumer.Consume(response.Body(), o.Payload); err != nil && err != io.EOF {
+	if err := consumer.Consume(response.Body(), o.Payload); err != nil && !stderrors.Is(err, io.EOF) {
```
SoloJacobs, line 116: Interesting, this does seem like it could fix some bugs. However, upstream just considers this a linting issue: https://github.com/go-swagger/go-swagger/pull/3202

I guess, we can ignore this for now.

SoloJacobs (approved):

PR #4898, opened 2026-01-18
refactor: replace config.Secret with commoncfg.Secret

TheMeier:
To be able to move notifier specific code out of the config and notify packages,
some shared code needs to be moved to other/new packages. Since commoncfg.Secret
already does what config.Secret does, this replaces all usages of config.Secret.
    
This requires external users of the config package to change their code if they used
config.MarshalSecretValue. This exported variable was added in
https://github.com/prometheus/alertmanager/pull/4158

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -41,48 +41,12 @@ const secretToken = "<secret>"
 
 var secretTokenJSON string
 
-// MarshalSecretValue if set to true will expose Secret type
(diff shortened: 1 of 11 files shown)


TheMeier:
Sadly there is no  SecretURL type in the common/config package. Since prometheus itself doesn't need it I think it is unlikely we cam move that type there.

PR #4899, opened 2026-01-18
Add arbitrary key/value data store to the nflog

Spaceman1701:
This change adds a new `receiver_data` field to the nflog entry protobuf message. `receiver_data` is a `string -> int|double|string` map. 

Integrations can interact with receiver data using the new `nflog.Store` struct which acts like a mutable persistent key/value store. Every time the notifier is called, the context contains a `*nflog.Store` with any data set by previous runs of the notifier. Since the data is sourced from the nflog, everything works as expected in HA mode and through restarts.

I've chosen to support single key/value pairs (as opposed to `key -> bytes` or just a raw `bytes` field to reduce complexity for the notifier. Nflog entries are persisted through restarts and upgrades, so it's important to avoid breaking backwards compatibility. Short of making every integration specify its own proto, we don't have a good way to avoid compatibility breaks. I think keeping the scope of custom data more limited will help discourage storing complicated structures in the nflog.

The nflog store is automatically cleared after an nflog message containing only resolved alerts. This allows "new" notifications to start from a fresh store. If a notifier doesn't mutate the store, the old values are persisted into the next nflog entry. 

Using the store is optional and easy:
```go
func (n *MyNotifier) Notify(ctx context.Context, alerts ...*types.Alert) (bool, error) {
    // store is always set
    store := notify.NFLogStore(ctx)
    
    persisted_data, ok := store.GetStr("something_useful")
    if !ok {
        persisted_data = n.generateNewPersistedData()
        store.SetStr("something_useful", persisted_data)
    }
   
    // use persisted data when sending the notification
}
```

In HRT, we use this system for storing the first notification's Slack`thread_ts` in our slack notifier. This allows future notifications to update the existing message or respond in the thread. I haven't included any uses of the store in this PR since HRT's slack notifier is a whole different can of worms.

### DIFF
--- a/nflog/nflog.go
+++ b/nflog/nflog.go
@@ -138,7 +138,7 @@ func (s *Store) GetStr(key string) (string, bool) {
 	return strVal.StrVal, true
 }
 
-// SetInt associates an integer value with the provided key, overwriting any existing value
+// SetInt associates an integer value with the provided key, overwriting any existing value.
 func (s *Store) SetInt(key string, v int64) {
 	s.data[key] = &pb.ReceiverDataValue{
 		Value: &pb.ReceiverDataValue_IntVal{
@@ -147,7 +147,7 @@ func (s *Store) SetInt(key string, v int64) {
 	}
 }
 
-// SetFloat associates a float value with the provided key, overwriting any existing value
+// SetFloat associates a float value with the provided key, overwriting any existing value.
 func (s *Store) SetFloat(key string, v float64) {
 	s.data[key] = &pb.ReceiverDataValue{
 		Value: &pb.ReceiverDataValue_DoubleVal{
@@ -156,7 +156,7 @@ func (s *Store) SetFloat(key string, v float64) {
 	}
 }
 
-// SetStr associates a string value with the provided key, overwriting any existing value
+// SetStr associates a string value with the provided key, overwriting any existing value.
 func (s *Store) SetStr(key, v string) {
 	s.data[key] = &pb.ReceiverDataValue{
 		Value: &pb.ReceiverDataValue_StrVal{
@@ -165,7 +165,7 @@ func (s *Store) SetStr(key, v string) {
 	}
 }
 
-// Delete deletes any value associated with the key
+// Delete deletes any value associated with the key.
 func (s *Store) Delete(key string) {
 	delete(s.data, key)
 }


notify/notify.go
```diff
@@ -788,6 +798,13 @@ func (n *DedupStage) Exec(ctx context.Context, _ *slog.Logger, alerts ...*types.
 		return ctx, nil, fmt.Errorf("unexpected entry result size %d", len(entries))
 	}
 
+	isFirstNotification := entry == nil || (len(entry.FiringAlerts) == 0 && len(entry.ResolvedAlerts) > 0 && len(firing) > 0)
```
siavashs: should we break this up and move it to the `switch` above?

```go
	var entry *nflogpb.Entry
    var isFirstNotification bool
	switch len(entries) {
	case 0:
        isFirstNotification = true
	case 1:
		entry = entries[0]
        if len(entry.FiringAlerts) == 0 && len(entry.ResolvedAlerts) > 0 && len(firing) > 0 {
            isFirstNotification = true
        }
	default:
		return ctx, nil, fmt.Errorf("unexpected entry result size %d", len(entries))
	}
```

siavashs: Also should `len(entry.ResolvedAlerts) > 0` be removed? It seems redundant.

siavashs: Also we could use a better name for the boolean, for example:
- `useEmptyStore`

Spaceman1701: > should we break this up and move it to the switch above?

Sure, that seems more clear to me as well

> Also should len(entry.ResolvedAlerts) > 0 be removed? It seems redundant.

I think you're right. One of my test cases fails without this condition, but I think it's a synthetic problem. 

> Also we could use a better name for the boolean, for example:

I disagree here - I think `isFirstNotification` is an accurate name for the state this condition captures. It's meant to be `true` if and only if this is the first notification for a new alert group. In the HRT fork, we have a slightly more fleshed out system for classifying the notification reason - this logic is similar to that, actually. We'll try to upstream that change as well, it's pretty small but we haven't gotten to it yet 😅 .

Spaceman1701:
> Added minor suggestions, but otherwise LGTM 👍 I also ran some code analysis on the tests and the recommendation was to add the following test cases:
> 
> ```
> // Test: Transition from resolved-only to firing alerts should clear receiver data
> // Test: Entry with firing alerts should preserve receiver data  
> // Test: Verify NflogStore contains expected data based on entry state
> ```

I _think_ I do have tests for each of these:
* `Transition from resolved-only to firing alerts should clear receiver data` -> `TestNFLogStore_NoLeakBetweenNotificationSequences` 
* `Entry with firing alerts should preserve receiver data` -> `TestReceiverData_PreserveWhenNotifierDoesNotUpdate` and `TestDedupStageExtractsReceiverData_*`
* `Verify NflogStore contains expected data based on entry state` -> `TestDedupStageExtractsReceiverData_*`

But I'm very happy to add some more test cases if something's missing here

nflog/nflog.go
```diff
@@ -76,6 +76,87 @@ func QGroupKey(gk string) QueryParam {
 	}
 }
 
+type Store struct {
```
ultrotter, line 83: Should we add a docstring for what Store is and its purpose? We could also mention that notifiers are not supposed to pass the store through concurrent goroutines (even if they don't do it now) to avoid issues? Anyway generally a little info about the lifetime of the Store (where it gets created, passed on, who is supposed to use it and how) might help

nflog/nflog.go
```diff
@@ -76,6 +76,87 @@ func QGroupKey(gk string) QueryParam {
...
+	data map[string]*pb.ReceiverDataValue
+}
+
+func NewStore(entry *pb.Entry) *Store {
+	var receiverData map[string]*pb.ReceiverDataValue
+	if entry != nil {
+		receiverData = entry.ReceiverData
```
ultrotter: Should we copy the map here? Or are we ok that e.g. SetInt would set a new int in the entry ReceiverData map?

nflog/nflog.go
```diff
@@ -76,6 +76,87 @@ func QGroupKey(gk string) QueryParam {
...
+
+func (s *Store) SetFloat(key string, i float64) {
+	s.data[key] = &pb.ReceiverDataValue{
+		Value: &pb.ReceiverDataValue_DoubleVal{
+			DoubleVal: i,
+		},
+	}
+}
+
+func (s *Store) SetStr(key, i string) {
```
ultrotter: Should we use s here? Or better just use "v" for all the Set* values names, no matter the type?

nflog/nflog.go
Spaceman1701: Ah yeah, I'll do that.

nflog/nflog.go
Spaceman1701: Good catch... I think it's probably fine as is, but it's not obvious that notifiers are changing the old entry's map. I could see this causing a bug sometime in the future. I'll change it to make a copy.

ultrotter (approved):

PR #4903, opened 2026-01-20
chore: setup stale action

TheMeier:
replace probot config with stale-action


TheMeier:
proposal for message content welcome. also not closing anything at the moment: `days-before-close: -1`

PR #4906, opened 2026-01-21
feat(notify): add notification logging for successfully sent notifications

### LINKED ISSUE #2304: [feature request] Log each successfully sent notification with metadata
**tl;dr:** I think it would be awesome if Alertmanager logged a rich/contextual message whenever it sent a notification that could be used for auditing/downstream analytics/etc. The current best practice of sending every alert to a webhook receiver would work but has some drawbacks that I think are enough to warrant adding some additional logging.

## My Wish :genie:

I run a centralized Alertmanager cluster that sends alerts to hundreds of different destinations/teams/people. I would like to have a way of tracking how many alerts are sent to each destination, integration, what the metadata is for each alert so that we can track it over time.

## The Current Best Practice

Reading docs and #560, it seems like the generally agreed/used practice here is to set up a route that matches all alerts right below the root route and send a webhook to something that does what you're looking for.

This would work fairly well for tracking alerts that have good metadata in labels about where they're going, but there are a few things about the approach that I'm not super excited about:

1. You wouldn't have any way of knowing where the actual alert will be delivered (which integration/channel/etc). So counting things by actual destination (in the unfortunate world where many routes lead to similar destinations) isn't possible. It is possible to count notifications delivered by integration in metrics, but correlating that with inconsistent alert labels would be difficult.
2. Adding a route that matches every alert for auditing means that if you somehow have an alert that doesn't have a matching route below it, it won't fall back to the root route. That's actually what I use the root route for now, catching bad user-defined alerts.
3. (this is probably not a real issue) It adds a non-zero amount of overhead/work for alertmanager and the cluster to track and execute, sending webhooks. They're computers to it probably isn't a big deal, but definitely extra work being done.

I'll definitely acknowledge here that we are not using alertmanager in the way that it should be. If our routing tree was a lot cleaner, and the labels on all of our alerts were more consistent, the webhook approach would be less problematic. But the path of least resistance for me (and hopefully to the benefit of the community) is to see if I can make the change in alertmanager, organizations/people are difficult :)

## Proposal

I took a quick look at the code. My initial thought was:

- Adding to either the `Integration` or `Notifier` interface to expose some rich metadata about the notification destination that could be logged at [notify time](https://github.com/prometheus/alertmanager/blob/1ace0f76b7101cccc149d7298022df36039858ca/notify/notify.go#L655)
- Grabbing the `CommonLabels` or `GroupLabels` from one of the sent alerts
- A very rich structured log could be added that contained everything you want to know about the alert and the destination

I'd be willing to take on the work. But I'll obviously defer to the expertise of the folks here on if this is even a reasonable thing to do, and if so on the approach.

Thank you so much for taking the time to read my request, thank you for all you do maintaining Prometheus and Alertmanager!

willianpaixao:
This PR implements notification logging for Alertmanager.

- Add `--notification-log.file` CLI flag to enable JSON-formatted notification logging
- Log successful notifications with rich metadata (receiver, integration, group key, alert counts, group labels)
- Follow the pattern established by Prometheus query logging

## Motivation

Users need visibility into which notifications were successfully sent for auditing and analytics purposes. The current workaround of using a webhook receiver with `continue: true` has limitations around destination visibility and routing behavior.

### New Files
- `notify/notification_log.go` - Notification logger implementation
- `notify/notification_log_test.go` - Unit tests

### Modified Files
- `notify/notify.go` - Integrate NotificationLogger into the pipeline
- `cmd/alertmanager/main.go` - Add CLI flag and initialization
- `notify/notify_test.go` - Update existing tests for new function signature

### Log Entry Format

Each line in the notification log file is a JSON object:

```json
{
  "timestamp": "2024-01-15T10:30:00.123Z",
  "integration": "slack",
  "integrationIdx": 0,
  "receiver": "team-alerts",
  "groupKey": "{}:{alertname=\"HighMemory\"}",
  "alertsCount": 3,
  "firingCount": 2,
  "resolvedCount": 1,
  "groupLabels": {
    "alertname": "HighMemory"
  }
}
```

### Design Decisions

1. **Metadata only**: Log receiver, integration, group key, counts, and group labels. Avoids sensitive alert data and keeps log files manageable.
2. **Success only**: Only log successful notifications. Failures are already visible in standard logs and metrics.
3. **JSON lines format**: Easy to parse, grep, and integrate with log aggregation systems.
4. **No built-in rotation**: Users can use logrotate or similar tools, following Prometheus conventions.

## Usage

```bash
alertmanager --notification-log.file=/var/log/alertmanager/notifications.log
```

When the flag is not set or empty, notification logging is disabled (default behavior).

## Test Plan

- [x] Unit tests for JSON serialization
- [x] Unit tests for file writing
- [x] Unit tests for concurrent write safety
- [x] Unit tests for noop logger
- [x] Existing notify package tests pass
- [x] Build succeeds
- [ ] Manual integration test: start Alertmanager with flag, send alerts, verify JSON entries in log file

## Future Considerations (Out of Scope)

- Log rotation (users can use logrotate)
- Logging failures (could add with `status: "failure"` in future PR)
- Configurable fields to include/exclude
- Metrics for notification logging

Fixes #2304

### DIFF
--- a/notify/notification_log.go
+++ b/notify/notification_log.go
@@ -1,0 +1,147 @@
+// Copyright 2015 Prometheus Team
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package notify
+
+import (
+	"encoding/json"
+	"errors"
+	"os"
+	"sync"
+	"time"
+)
+
+// ErrNotificationLogClosed is returned when attempting to write to a closed notification log.
+var ErrNotificationLogClosed = errors.New("notification log is closed")
+
+// NotificationLogEntry represents a single notification log entry written to the
+// notification log file. Each entry captures metadata about a successfully sent
+// notification, including the integration type, receiver name, alert counts, and
+// group labels. The entry is serialized as a JSON object on a single line.
+//
+// Example JSON output:
+//
+//	{
+//	  "timestamp": "2024-01-15T10:30:00.123Z",
+//	  "integration": "slack",
+//	  "integrationIdx": 0,
+//	  "receiver": "team-alerts",
+//	  "groupKey": "{}:{alertname=\"HighMemory\"}",
+//	  "alertsCount": 3,
+//	  "firingCount": 2,
+//	  "resolvedCount": 1,
+//	  "groupLabels": {"alertname": "HighMemory"}
+//	}
+type NotificationLogEntry struct {
+	// Timestamp is the time when the notification was successfully sent.
+	Timestamp time.Time `json:"timestamp"`
+	// Integration is the type of notifier used (e.g., "slack", "webhook", "email").
+	Integration string `json:"integration"`
+	// IntegrationIdx is the index of the integration within the receiver configuration.
+	IntegrationIdx int `json:"integrationIdx"`
+	// Receiver is the name of the receiver that processed the notification.
+	Receiver string `json:"receiver"`
+	// GroupKey is the unique identifier for the alert group.
+	GroupKey string `json:"groupKey"`
+	// AlertsCount is the total number of alerts in the notification.
+	AlertsCount int `json:"alertsCount"`
+	// FiringCount is the number of firing alerts in the notification.
+	FiringCount int `json:"firingCount"`
+	// ResolvedCount is the number of resolved alerts in the notification.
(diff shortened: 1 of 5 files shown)


notify/notification_log.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	return &FileNotificationLogger{
+		file: f,
+	}, nil
+}
+
+// Log writes a notification entry to the log file as a JSON line.
+// The data is buffered by the OS and synced to disk when Close is called.
+// Returns ErrNotificationLogClosed if the logger has been closed.
+func (l *FileNotificationLogger) Log(entry *NotificationLogEntry) error {
```
ultrotter, line 100: I am wondering if we should add new locks for this, or a custom notification format...
Would it make sense to use slog for example, and standardize to the standard library?

ultrotter: Also implementation wise, would sending the entries to a channel, and having a goroutine do the marshalling and logging be better, and hold locks for less? I am a bit worried about json.Marshal(...) while holding a lock

ultrotter:
I am closing this PR because of the option to use eventrecorder, plus no response to previous feedback. Thanks and feel free to contribute again!

PR #4917, opened 2026-01-29
config: add global mattermost_api_url and mattermost_api_url_file

madest92:
Default Mattermost receiver webhook_url/webhook_url_file from global settings


TheMeier:
Since https://github.com/prometheus/alertmanager/pull/4812 was merged could you please try to follow it's pattern where possible?

config/config.go
```diff
@@ -703,6 +707,10 @@ func (c *Config) UnmarshalYAML(unmarshal func(any) error) error {
 		}
 		for _, mattermost := range rcv.MattermostConfigs {
 			mattermost.HTTPConfig = cmp.Or(mattermost.HTTPConfig, c.Global.HTTPConfig)
+			if mattermost.WebhookURL == nil && len(mattermost.WebhookURLFile) == 0 {
```
TheMeier, line 686: IMHO this If is not needed and the `cmp.Or` could also be used for the remaining 2 ones. Because further up it is checked only one of them is set. `cmp.Or` would only assign a non-empty value to one or the other but not both.

madest92:
> Since #4812 was merged could you please try to follow it's pattern where possible?

I've reviewed the changes. This doesn't apply to mattermost_api_url/mattermost_api_url_file.

Conflict fixed.

config/config.go
TheMeier: Actually I am wrong. This would allow to set a global WebhookURL and a local WebhookURLFile ending up in both being set in the receiver.

madest92: No possible. The test fails.

```
global:
  mattermost_api_url_file: '/global_mm_file'
route:
  receiver: 'mm-notifications'
receivers:
  - name: 'mm-notifications'
    mattermost_configs:
      - channel: 'alerts3'
        text: 'test'
        webhook_url: 'http://mysecret.example.com/'
```

Then both variables are used(`webhook_url` and `webhook_url_file`). But one of them must be used.

madest92:
@TheMeier 
I'd actually like to do a small refactoring of the MM configuration. Sending messages via attachment, like in Slack.
Move the attachments up a level and add the `define` templates.

Now you need to write like this:
```
mattermost_configs:
  - webhook_url: '.....'
    username: '{{ template "__alertmanager" . }}'
    text: ''
    attachments:
      - color: '{{ if eq .Status "firing" }}danger{{ else }}good{{ end }}'
        title: '{{ .CommonAnnotations.summary }}'
        title_link: '{{ template "__alertmanagerURL" . }}'
        text: '{{ range .Alerts }}{{ .Annotations.description }}{{ end }}'
```

Instead of
```
mattermost_configs:
  - text: 'if you want update default'
    title: 'if you want update default'
    ....
```

What do you think about this?
Yes, it will break the current behavior. But it will be more familiar to everyone.

I can submit changes as part of this merge request, or create a new one once the current one is accepted.
What's best?

TheMeier:
I personally think this should be two separate changes. Adding something is a completly different beast than breaking changes.

madest92:
Okay, then I'm waiting for the current changes to be accepted.

madest92:
@TheMeier hi

Do you have any comments on Mr?

madest92:
@TheMeier hi. Who can give a second approval and merge the changes?

TheMeier:
I think this is now resolved via https://github.com/prometheus/alertmanager/pull/4998

guoard:
Hi @madest92,

I just realized that you had opened a PR for the same feature before mine was merged. I'm sorry I didn't check for similar open PRs beforehand, I should have looked more carefully before submitting mine.

Thank you for your work on this, and apologies again for the overlap.

PR #4922, opened 2026-02-01
Implement status pages in the mantine-ui

sysadmind:
Add the Status>Configuration and Status>Runtime & Build Information pages. These are nearly identical to the pages in Prometheus.

I copied the InfoPageCard and InfoPageStack files directly from Prometheus. Now that we will have a second project using some of these components, we may want to start thinking about a shared library.

image
image

### DIFF
--- a/ui/mantine-ui/src/components/InfoPageStack.tsx
+++ b/ui/mantine-ui/src/components/InfoPageStack.tsx
@@ -1,5 +1,5 @@
-import { Stack } from "@mantine/core";
-import { FC, ReactNode } from "react";
+import { FC, ReactNode } from 'react';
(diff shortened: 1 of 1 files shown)


TheMeier:
Gave it a quick try, looks good to me.

juliusv:
Yeah, the code duplication is definitely annoying and introduces drift risk. I wish sharing code between different repos in JS was as simple as in Go. But in JS you basically have to publish an npm package. So not sure when that becomes worth it...

Other than that, :+1:

TheMeier:
Maybe https://github.com/prometheus/prometheus/blob/main/scripts/sync_repo_files.sh could be used?

Issue #3279, opened 2023-03-06
Silence Regex Match behaving unintuitively 


sinkingpoint:
**What did you do?**

Created a silence with a regex matcher with a double quote in it

**What did you expect to see?**

The silence shouldn't match anything

**What did you see instead? Under which circumstances?**

The silence matches everything

**Environment**

* System information:

	Linux 5.15.49-linuxkit x86_64

* Alertmanager version:

```
/alertmanager $ alertmanager --version
alertmanager, version 0.25.0 (branch: HEAD, revision: 258fab7cdd551f2cf251ed0348f0ad7289aee789)
build user:       root@abe866dd5717
build date:       20221222-14:51:36
go version:       go1.19.4
platform:         linux/amd64
```

* Alertmanager configuration file:

The default from prom/alertmanager:v0.25.0
```
global:
  resolve_timeout: 5m
  http_config:
    follow_redirects: true
    enable_http2: true
  smtp_hello: localhost
  smtp_require_tls: true
  pagerduty_url: https://events.pagerduty.com/v2/enqueue
  opsgenie_api_url: https://api.opsgenie.com/
  wechat_api_url: https://qyapi.weixin.qq.com/cgi-bin/
  victorops_api_url: https://alert.victorops.com/integrations/generic/20131114/alert/
  telegram_api_url: https://api.telegram.org
  webex_api_url: https://webexapis.com/v1/messages
route:
  receiver: web.hook
  group_by:
  - alertname
  continue: false
  group_wait: 30s
  group_interval: 5m
  repeat_interval: 1h
inhibit_rules:
- source_match:
    severity: critical
  target_match:
    severity: warning
  equal:
  - alertname
  - dev
  - instance
receivers:
- name: web.hook
  webhook_configs:
  - send_resolved: true
    http_config:
      follow_redirects: true
      enable_http2: true
    url: http://127.0.0.1:5001/
    max_alerts: 0
templates: []
```

Bit of a long form description for replication:

1. Start an alertmanager: `docker run -d -p 9093:9093 prom/alertmanager:v0.25.0`
2. Send an alert with a single label, instance=foo: `curl -XPOST localhost:9093/api/v2/alerts -H 'Content-Type: application/json' -d '[{"labels":{"instance":"foo"}}]'`
[...]

PR #4942, opened 2026-02-02
fix(ui): handle special chars in silence regex-matchers

TheMeier:
use JSON.encode/decode for silence matchers to properly escape special characters. This changes the display on the silences overview page to also include the quotes around the regex matchers, but ensures that special characters are handled correctly.

IMHO this change on the silence overview makes the UI more consistent. On the detail page nothing changes. On the edit page for existing regex matchers this now correctly includes escaping for special characters.

Fixes: #3279


SoloJacobs:
image
Seems like this has the nice side effect of fixing that button.

Issue #4945, opened 2026-02-04
Update Protobuf generation process


siavashs:
The gogoproto generator is deprecated and requires more dependencies to work with proto files.
Therefore we must use Buf (also used by Prometheus) instead of gogoproto.

References:
- https://github.com/bufbuild/buf
- https://github.com/gogo/protobuf

PR #4955, opened 2026-02-04
Add buf for protobuf generation

SoloJacobs:
HRT uses this internally, and have positive experiences with this project. Thus, this change facilitates them upstreaming their changes, and makes are project easier to work with.

`buf` also has some nice additional tooling, which we could make use of.

Finally, we want to get rid off `gogoproto`, and having tooling will be helpful.

Fixes: #4945


SoloJacobs:
For reference: https://github.com/prometheus/prometheus/pull/16239/commits

PR #4958, opened 2026-02-05
refactor: make dispatcher locking more granular

ultrotter:
We first add a unique index to each route, and test that it's calculated correctly.

Then we pre-allocate the route to aggrgroup map, and actually use a slice instead.

Finally we can remove the lock, changing a few more things so that they are atomic.

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -97,14 +97,15 @@ type Dispatcher struct {
 
 	timeout func(time.Duration) time.Duration
 
-	loadingFinished  sync.WaitGroup
+	loaded   chan struct{}
+	finished sync.WaitGroup
+	ctx      context.Context
+	cancel   func()
+
 	routeGroupsSlice []routeAggrGroups
 	aggrGroupsNum    atomic.Int32
 
 	maintenanceInterval time.Duration
-	finished            sync.WaitGroup
-	ctx                 context.Context
-	cancel              func()
 
 	logger *slog.Logger
 
@@ -155,7 +156,7 @@ func NewDispatcher(
 		propagator:          otel.GetTextMapPropagator(),
 	}
 	disp.state.Store(DispatcherStateUnknown)
-	disp.loadingFinished.Add(1)
+	disp.loaded = make(chan struct{})
 	disp.ctx, disp.cancel = context.WithCancel(context.Background())
 	return disp
 }
@@ -186,7 +187,7 @@ func (d *Dispatcher) Run(dispatchStartTime time.Time) {
 	for _, alert := range initalAlerts {
 		d.routeAlert(d.ctx, alert)
 	}
-	d.loadingFinished.Done()
+	close(d.loaded)
 
 	d.run(it)
 }
@@ -283,17 +284,11 @@ func (d *Dispatcher) doMaintenance() {
 }
 
 func (d *Dispatcher) WaitForLoading() {
-	d.loadingFinished.Wait()
+	<-d.loaded
 }
 
 func (d *Dispatcher) LoadingDone() <-chan struct{} {
-	doneChan := make(chan struct{})
-	go func() {
-		d.WaitForLoading()
-		close(doneChan)
-	}()
-
-	return doneChan
+	return d.loaded
 }
 
 // AlertGroup represents how alerts exist within an aggrGroup.


dispatch/dispatch.go
```diff
@@ -96,20 +97,19 @@ type Dispatcher struct {
...
-	loadingFinished    sync.WaitGroup
-	aggrGroupsPerRoute map[*Route]map[model.Fingerprint]*aggrGroup
-	aggrGroupsNum      int
+	loadingFinished  sync.WaitGroup
+	routeGroupsSlice []routeAggrGroups
+	aggrGroupsNum    atomic.Int32
 
 	maintenanceInterval time.Duration
-	done                chan struct{}
+	finished            sync.WaitGroup
```
Spaceman1701: why change this to a wait group? I think the `done` channel is a little more idiomatic.

dispatch/dispatch.go
```diff
@@ -309,39 +323,26 @@ func (d *Dispatcher) Groups(ctx context.Context, routeFilter func(*Route) bool,
 		return nil, nil, ctx.Err()
 	case <-d.LoadingDone():
 	}
-	d.WaitForLoading()
```
Spaceman1701, line 312: I think we still need this

ultrotter: I don't think we need it: case <-d.LoadingDone(): waits on the group then closes the channel, and we are waiting for the channel to be closed. once it's closed it's strictly after WaitForLoading passes... since the channel would not be closed otherwise.

dispatch/dispatch.go
ultrotter: Right, I thought I needed it to deal with the lock in Stop() but it seems I don't. I'll add a commit to fix this and also make LoadingDone() more efficient

ultrotter:
> It'd be helpful to include some benchmarks.

I will work one some, I have some work in progress ones that show no penalty, but only a very ad-hoc one that shows the gain... Anyway they are only meaningful on the whole patch series, not this change alone

dispatch/dispatch.go
ultrotter: Actually I tried to change it back to a channel and it was problematic: before Stop() took the dispatcher lock, and all was safe, but now it has to be idempotent, and *also* not listen on the done channel if it'll never be closed. While it's arguably possible to achieve with the state atomic pointer and switch cases, it becomes a bit convoluted and relies on assumptions on how we use the dispatcher outside of the module itself. As such I would prefer a waitgroup, and not having to worry, especially as it's not used that often.

ultrotter: I managed to change the loaded signal to a channel, but for now I left finished as a waitgroup for the reasons above

dispatch/dispatch.go
SoloJacobs: I agree with Guido here 👍

SoloJacobs (approved):
LGTM

dispatch/dispatch.go
Spaceman1701: Ah, I see, yeah you're right

dispatch/dispatch.go
```diff
@@ -428,45 +417,41 @@ func (d *Dispatcher) groupAlert(ctx context.Context, alert *types.Alert, route *
...
 
-	d.mtx.Lock()
-	defer d.mtx.Unlock()
-
-	routeGroups, ok := d.aggrGroupsPerRoute[route]
-	if !ok {
-		routeGroups = map[model.Fingerprint]*aggrGroup{}
-		d.aggrGroupsPerRoute[route] = routeGroups
-	}
+	d.routeGroupsSlice[route.Idx].mtx.Lock()
```
Spaceman1701, line 420: small style nitpick: any chance you could reduce the number of times we reference the group via `d.routeGroupsSlice[route.Idx]`? I think it'd be a little more clear if this code assigned the slice to a variable internally:

```go
groups := d.routeGroupsSlice[route.Idx]
groups.mtx.Lock()
defer groups.mtx.Unlock()

...
```

The way it's written now makes it a little unclear that it's only legal to access the groups for `route.Idx`.

dispatch/dispatch.go
```diff
@@ -428,45 +417,41 @@ func (d *Dispatcher) groupAlert(ctx context.Context, alert *types.Alert, route *
...
 		)
 		return
 	}
 
 	ag = newAggrGroup(d.ctx, groupLabels, route, d.timeout, d.marker.(types.AlertMarker), d.logger)
-	routeGroups[fp] = ag
-	d.aggrGroupsNum++
-	d.metrics.aggrGroups.Inc()
+	d.routeGroupsSlice[route.Idx].groups[fp] = ag
+	d.aggrGroupsNum.Add(1)
```
Spaceman1701, line 449: it's probably a good idea to add a comment somewhere explaining why it's ok for `aggrGroupsNum` to be an atomic rather than sharing a lock with group creation. I know we talked about this offline (and the conclusion was that the aggrGroup limit becomes the configured value + the number of dispatcher threads), but it's not documented anywhere (that I see) for the world.

PR #4960, opened 2026-02-05
chore: migrate to google.golang.org/protobuf

siavashs:
- migrate to google.golang.org/protobuf
- drop github.com/gogo/protobuf
- drop github.com/matttproud/golang_protobuf_extensions


Spaceman1701:
This is great! 

Did you verify that the generated binary proto messages are the same?

siavashs:
> This is great! 
> 
> 
> 
> Did you verify that the generated binary proto messages are the same? 

I will setup a cluster with nodes using old and new proto setup tomorrow to test replication.
And also save to disk with old version and load with new version.

siavashs:
I tested both data migration and replication across versions and it worked fine for silences.
I assume it would be fine for nflog as well since we have the same types for time.

siavashs:
> Yes! I think this works correctly. I'm wondering, if there are any benchmarks, we could do? I don't want another performance regression. But I don't have a cluster to test this properly.

For general performance changes of old vs. new protobuf implementation in golang check https://go.dev/blog/protobuf-opaque which claims the performance is better or compatible depending on the field types and might require less allocations.
> 

Interestingly we had `(gogoproto.stdtime) = true` (which was used on `google.protobuf.Timestamp` fields) and it causes gogoproto to suffer from a marshal performance issues: https://github.com/gogo/protobuf/issues/656

Cloudflare has a very noisy alert traffic and this is the number of messages we have across the cluster:
image
I think with this number of messages, small changes in performance of message decoding/encoding will not be significant.

So considering the above I think the performance should be equal if not better.
If you think we need a benchmark I can try to write one.

siavashs:
I ran an analysis to check if we need to implement a benchmark and the answer seems to be no:

> 1. The messages are simple. `Part` is just `{string, []byte}` and `FullState` is repeated `Part`. `MemberlistMessage` is `{string, enum, string, []byte}`. For such simple schemas, the performance difference between gogo and the official implementation is negligible — both are dominated by byte copying, not reflection or field decoding overhead.
> 2. The official protobuf v1.36.11 with open.v1 is already well-optimized. The generated code uses `unsafe` internally for lazy descriptor init, and the marshal/unmarshal paths for simple `bytes/string` fields have very little overhead. The gogo advantage (direct struct encoding) mattered more for deeply nested messages with many fields.
> 3. The real bottleneck is network I/O, not serialization. The `handleOverSizedMessages` goroutine sends to each peer over TCP, and `OversizedMessage` triggers for `payloads > MaxGossipPacketSize/2`. Serialization time is dwarfed by network latency.

SoloJacobs:
Yes, I think you are right, this change does not need a benchmark. I just like to extend coverage, when I'm touching some area. But that is just a personal preference.

PR #4965, opened 2026-02-05
add silence annotations

Spaceman1701:
Silence annotations are... annotations... for silences! 

Just like alert annotations, silence annotations are an optional `LabelSet` which provides structured key->value data on silences. Annotations are added to the underlying cluster model for silences so they're synced properly between nodes in HA mode. 

This is a simple change that adds a lot of value: with structured fields, it's possible to add both machine and human readble metadata for silences. In HRT, we use annotations to indicate the purpose of silences, associate them with machine processes, and encode conventions. The format is flexible so it's possible to encode any number of things.

### DIFF
--- a/cli/silence_add.go
+++ b/cli/silence_add.go
@@ -151,16 +151,19 @@ func (c *silenceAddCmd) add(ctx context.Context, _ *kingpin.ParseContext) error 
 		return errors.New("comment required by config")
 	}
 
-	annotations := make(models.LabelSet, len(c.annotations))
-	for _, a := range c.annotations {
-		matcher, err := compat.Matcher(a, "cli")
-		if err != nil {
-			return err
+	var annotations models.LabelSet
+	if len(c.annotations) > 0 {
+		annotations = make(models.LabelSet, len(c.annotations))
+		for _, a := range c.annotations {
+			matcher, err := compat.Matcher(a, "cli")
+			if err != nil {
+				return err
+			}
+			if matcher.Type != labels.MatchEqual {
+				return errors.New("annotations must be specified as key=value pairs")
+			}
+			annotations[matcher.Name] = matcher.Value
 		}
-		if matcher.Type != labels.MatchEqual {
-			return errors.New("annotations must be specified as key=value pairs")
-		}
-		annotations[matcher.Name] = matcher.Value
 	}
 
 	start := strfmt.DateTime(startsAt)
--- a/silence/silence_test.go
+++ b/silence/silence_test.go
@@ -2837,3 +2837,128 @@ func TestStateDecodingError(t *testing.T) {
 func gosched() {
 	time.Sleep(1 * time.Millisecond)
 }
+
+func TestSilenceAnnotations(t *testing.T) {
+	s, err := New(Options{
+		Metrics:   prometheus.NewRegistry(),
+		Retention: time.Hour,
+	})
+	require.NoError(t, err)
+
+	clock := quartz.NewMock(t)
+	s.clock = clock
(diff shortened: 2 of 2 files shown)


SoloJacobs (changes requested):
I think a bit more test coverage would go here, most importantly a round trip test:
 `post silence (using amtool) -> get silence -> compare annotations.`

`cli/silence_add.go` and `api/v2/compat.go` could also be tested.

I also assume that we expect `amtool` to not have an error, if you post to an older `alertmanager` version?

Spaceman1701:
> I think a bit more test coverage would go here, most importantly a round trip test: `post silence (using amtool) -> get silence -> compare annotations.`
> 
> `cli/silence_add.go` and `api/v2/compat.go` could also be tested.


Sure, I'll add a test case for the round trip, that's a good idea.

> I also assume that we expect `amtool` to not have an error, if you post to an older `alertmanager` version?

Ah yeah, let me also make sure that works - I think if you set annotations and try to post to an old alertmanager it should be expected to break. If you don't, I think it's reasonable to make it work. 

We don't really have a version skew policy, but I guess at some point we'll need to think about guidelines for which versions of `amtool` ought to be compatible with which versions of `alertmanager`...

TheMeier:
`amtool` already [checks if the versions differ](https://github.com/prometheus/alertmanager/blob/a408ed82a0aa6ec80e96ceff88d9532e02271e58/cli/root.go#L140) 

For a users perspective the optimal behaviour would be `amtool` printing a warning if annotations are passed but the server version does not handle them but work otherwise.

Spaceman1701:
> `amtool` already [checks if the versions differ](https://github.com/prometheus/alertmanager/blob/a408ed82a0aa6ec80e96ceff88d9532e02271e58/cli/root.go#L140)
> 
> For a users perspective the optimal behaviour would be `amtool` printing a warning if annotations are passed but the server version does not handle them but work otherwise.

Agreed this would be the nicest behavior. However, it's a bit of work to implement because I don't know what semantic version of alertmanager will start supporting silence annotations... 

I could think of something to make this work, but given that this is a bit of a corner case (the user is using a new version of `amtool` on an old version of `alertmanager` _and_ is trying to set annotations), I don't think we should worry too much.

cli/silence_add.go
```diff
@@ -85,6 +86,7 @@ func configureSilenceAddCmd(cc *kingpin.CmdClause) {
 	addCmd.Flag("end", "Set when the silence should end (overwrites duration). RFC3339 format 2006-01-02T15:04:05-07:00").StringVar(&c.end)
 	addCmd.Flag("comment", "A comment to help describe the silence").Short('c').StringVar(&c.comment)
 	addCmd.Arg("matcher-groups", "Query filter").StringsVar(&c.matchers)
+	addCmd.Flag("annotation", "Set an annotation to be included with the silence").StringsVar(&c.annotations)
```
siavashs:
```suggestion
	addCmd.Flag("annotation", "Set an annotation to be included with the silence (may be repeated)").StringsVar(&c.annotations)
```

SoloJacobs:
I quite like the new test! Thanks for adding it. Regarding the `amtool` - `alertmanager` compatibility: Well at least there is *a* warning. So, LGTM :+1:

ghost:
Hey @Spaceman1701, I’ve tested the silence annotations feature in Alertmanager and it looks like it’s available via the API but not exposed in the UI. Are there any plans to support this in the UI as well?

TheMeier:
Hey deleted user :)

There is already a UI change implemented and merged, it will become available with the next release.

PR #4966, opened 2026-02-05
refactor: convert route groups to a sync map

ultrotter:
This allows us to avoid taking the lock in most cases.

The exception is that we still need it when adding an aggrgroup, as multiple goroutines might be otherwise trying to add theirs, for the same fp.
 
This applies on top of https://github.com/prometheus/alertmanager/pull/4958

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -762,7 +762,9 @@ func (ag *aggrGroup) flush(notify func(...*types.Alert) bool) {
 		// and we don't want to send another one. However, we need to make sure
 		// that each resolved alert has not fired again during the flush as then
 		// we would delete an active alert thinking it was resolved.
-		if err := ag.alerts.DeleteIfNotModified(resolvedSlice); err != nil {
+		// Since we are passing DestroyIfEmpty=true the group will be marked as
+		// destroyed if there are no more alerts after the deletion.
+		if err := ag.alerts.DeleteIfNotModified(resolvedSlice, true); err != nil {
 			ag.logger.Error("error on delete alerts", "err", err)
 		} else {
 			// Delete markers for resolved alerts that are not in the store.
--- a/store/store.go
+++ b/store/store.go
@@ -44,7 +44,6 @@ type Alerts struct {
 	gcCallback    func([]types.Alert)
 	limits        map[string]*limit.Bucket[model.Fingerprint]
 	perAlertLimit int
-	dontdestroy   bool
 	destroyed     bool
 }
 
@@ -66,16 +65,6 @@ func (a *Alerts) WithPerAlertLimit(lim int) *Alerts {
 
 	a.limits = make(map[string]*limit.Bucket[model.Fingerprint])
 	a.perAlertLimit = lim
-
-	return a
-}
-
-// WithDontDestroy sets the dont destroy flag for the Alerts struct.
-func (a *Alerts) WithDontDestroy() *Alerts {
-	a.Lock()
-	defer a.Unlock()
-
-	a.dontdestroy = true
 
 	return a
 }
@@ -130,9 +119,7 @@ func (a *Alerts) GC() []types.Alert {
 		}
 	}
 
-	if len(a.alerts) == 0 && !a.dontdestroy {
-		a.destroyed = true
-	}
+	// GC doesn't destroy the alerts store if it's empty.
 
 	a.Unlock()
 	a.gcCallback(resolved)
@@ -182,7 +169,7 @@ func (a *Alerts) Set(alert *types.Alert) error {
 
 // DeleteIfNotModified deletes the slice of Alerts from the store if not
 // modified.
-func (a *Alerts) DeleteIfNotModified(alerts types.AlertSlice) error {
+func (a *Alerts) DeleteIfNotModified(alerts types.AlertSlice, destroyIfEmpty bool) error {
 	a.Lock()
 	defer a.Unlock()
 	for _, alert := range alerts {
@@ -192,8 +179,8 @@ func (a *Alerts) DeleteIfNotModified(alerts types.AlertSlice) error {
 		}
 	}
 
-	// If the store is now empty, mark it as destroyed (unless dontdestroy is set)
-	if len(a.alerts) == 0 && !a.dontdestroy {
+	// If the store is now empty, mark it as destroyed
+	if len(a.alerts) == 0 && destroyIfEmpty {
 		a.destroyed = true
 	}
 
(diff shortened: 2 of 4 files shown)


dispatch/dispatch.go
```diff
@@ -147,26 +152,33 @@ func NewDispatcher(
...
+	defer d.finished.Done()
 
-	d.mtx.Lock()
 	d.logger.Debug("preparing to start", "startTime", dispatchStartTime)
 	d.startTimer = time.NewTimer(time.Until(dispatchStartTime))
-	d.state = DispatcherStateWaitingToStart
 	d.logger.Debug("setting state", "state", "waiting_to_start")
-	d.aggrGroupsPerRoute = map[*Route]map[model.Fingerprint]*aggrGroup{}
-	d.aggrGroupsNum = 0
+	d.routeGroupsSlice = make([]routeAggrGroups, d.route.Idx+1)
```
siavashs, line 174: I guess route index is zero here always, so should we just use `1` for clarity?

```suggestion
	d.routeGroupsSlice = make([]routeAggrGroups, 1)
```

dispatch/dispatch.go
```diff
@@ -147,26 +152,33 @@ func NewDispatcher(
...
+	d.route.Walk(func(r *Route) {
+		d.routeGroupsSlice[r.Idx] = routeAggrGroups{
+			route: r,
+		}
+	})
+
+	d.aggrGroupsNum.Store(0)
 	d.metrics.aggrGroups.Set(0)
```
siavashs, line 182: Should we use a `GaugeFunc()` instead and just read the atomic `aggrGroupsNum` on scrape?

dispatch/dispatch.go
```diff
@@ -428,45 +415,57 @@ func (d *Dispatcher) groupAlert(ctx context.Context, alert *types.Alert, route *
...
 	}
 
-	ag = newAggrGroup(d.ctx, groupLabels, route, d.timeout, d.marker.(types.AlertMarker), d.logger)
-	routeGroups[fp] = ag
-	d.aggrGroupsNum++
-	d.metrics.aggrGroups.Inc()
+	ag := newAggrGroup(d.ctx, groupLabels, route, d.timeout, d.marker.(types.AlertMarker), d.logger)
+	d.routeGroupsSlice[route.Idx].groups.Store(fp, ag)
+	d.aggrGroupsNum.Add(1)
+	d.metrics.aggrGroups.Set(float64(d.aggrGroupsNum.Load()))
```
siavashs: See my previous comment about using `GaugeFunc()`

dispatch/dispatch.go
ultrotter: No no, the top of the tree is always the maximum index (see tests and route tree implementation) so this creates a slice of 20, if there are 20 routes.

dispatch/dispatch.go
ultrotter: I thought about it, but metrics are created once in main, while dispatcher is recreated on every config reload... It would need a gaugefunc that accesses the pointer, verifies if it's safe, and then an extra public method on the dispatcher to export the value.. Not sure if it's worth it or we actually gain anything to do that, so for now I hadn't changed it that way

dispatch/dispatch.go
siavashs: Cool, I thought since it is the root it has an index of zero.

dispatch/dispatch.go
siavashs: I see, let's keep it as is for now. We can always revisit.

dispatch/dispatch.go
```diff
@@ -268,17 +265,16 @@ func (d *Dispatcher) routeAlert(ctx context.Context, alert *types.Alert) {
 
 func (d *Dispatcher) doMaintenance() {
 	for i := range d.routeGroupsSlice {
-		d.routeGroupsSlice[i].mtx.Lock()
-		for _, ag := range d.routeGroupsSlice[i].groups {
-			if ag.empty() {
-				ag.stop()
-				d.marker.DeleteByGroupKey(ag.routeID, ag.GroupKey())
-				delete(d.routeGroupsSlice[i].groups, ag.fingerprint())
+		d.routeGroupsSlice[i].groups.Range(func(_, ag any) bool {
```
SoloJacobs: Would be nice to only cast the pointer once here.

store/store.go
```diff
@@ -41,6 +44,8 @@ type Alerts struct {
 	gcCallback    func([]types.Alert)
 	limits        map[string]*limit.Bucket[model.Fingerprint]
 	perAlertLimit int
+	dontdestroy   bool
```
SoloJacobs: Hmm, this is a bit hard to reason through. Is there a way to move the destroy property to the AggrGroup itself?

ultrotter:
> Good enough for me. However, there is a commit missing from #4958 .

I know, but it rebases cleanly, so I was hoping github will happily rebase it once I merge that one :)

store/store.go
ultrotter: That's hard because the mutex that protects the data and stores the alerts is here in the store, and the aggrgroup doesn't have an outer mutex itself anymore, so we couldn't do it race-free.

In theory we could never destroy during `GC`, as only the global alertstore runs `GC`, and always destroy during `DeleteIfNotModified` since that's how we clean them up in the dispatcher, but that's even worse since it makes the behavior implicit on how we use this module today.

Ah, what we could do is:
- remove the dontdestroy flag
- pass a destroyIfEmpty bool to DeleteIfNotModified

and then GC will never destroy (ok) and DeleteIfNotModified will do it explicitly.... Would you prefer this option?

dispatch/dispatch.go
ultrotter: I think I can do that, yes

store/store.go
SoloJacobs: Yes, I think that proposal is cleaner. :+1: 

My feeling is that the requirements, which users of `store` api have are slowly diverging. I think it might become worth duplicating that package, and then removing parts, which are not quintessential. The complexity is quite high enough, yet.

Spaceman1701: nit: `dontdestroy` to `doNotDestroy`

ultrotter:
> Good enough for me. However, there is a commit missing from #4958 .

Yes, I was waiting to finalize before rebasing! It's done now!

dispatch/dispatch.go
ultrotter: Done!

store/store.go
ultrotter: Done!

ultrotter: I removed this field as agreed with @SoloJacobs

PR #4970, opened 2026-02-06
bug?: main: change the disp and inhibitor pointers to atomic

ultrotter:
These are used by APIs so in theory could race, though they are atomic on most architectures. Just to be sure, convert to the proper type.

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -419,7 +419,7 @@ func run() int {
 	tracingManager := tracing.NewManager(logger.With("component", "tracing"))
 
 	var (
-		inhibitor *inhibit.Inhibitor
+		inhibitor atomic.Pointer[inhibit.Inhibitor]
 		tmpl      *template.Template
 	)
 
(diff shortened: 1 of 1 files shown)


SoloJacobs:
Are we really expecting these pointers to be updated by multiple threads? If yes, where are those spawned?

siavashs (approved):

ultrotter:
> Are we really expecting these pointers to be updated by multiple threads? If yes, where are those spawned?

No, those pointers are updated in a single place, but they are read in multiple threads (particularly we pass function closures that read them during the API calls). This still needs protection as far as I know.

TheMeier (approved):

cmd/alertmanager/main.go
```diff
@@ -362,13 +363,13 @@ func run() int {
 	}
 	defer alerts.Close()
 
-	var disp *dispatch.Dispatcher
+	var disp atomic.Pointer[dispatch.Dispatcher]
 	defer func() {
-		disp.Stop()
+		disp.Load().Stop()
 	}()
```
siavashs, line 369: Do we still need to wrap this? I assume `Load()` will dynamically load the correct pointer and we don't need to use a `func()`.

ultrotter: Alas, we still need it: https://go.dev/play/p/jUW0nhS-dem

PR #4971, opened 2026-02-06
Improve DedupStage to add the reason for notifying

Spaceman1701:
The `DedupStage.needsUpdate` method decides if the notification pipeline should send a notification by reading the most recent entry from the nflog for the alert group and comparing it to the current state. It returns a boolean, but the internal logic actually identifies a few distinct conditions for notifications.

This is a small change which updates that function to return a value from a new `NotifyReason` enum. The value is added to the context and wired through the template data.

This is useful for two reasons:
1. It allows user templates to include some text about _why_ we're sending a notification. 
2. It makes the behavior of `needsUpdate` a little more clear to readers

This change requires a tiny update to the `needsUpdate` logic where two logical cases where returned via one path.

### DIFF
--- a/notify/notify.go
+++ b/notify/notify.go
@@ -764,8 +764,7 @@ func (n *DedupStage) needsUpdate(entry *nflogpb.Entry, firing, resolved map[uint
 	// If we haven't notified about the alert group before, notify right away
 	// unless we only have resolved alerts.
 	if entry == nil {
-		newAlerts := len(firing) > 0
-		if newAlerts {
+		if len(firing) > 0 {
 			return ReasonFirstNotification
 		}
 		return ReasonDoNotNotify
@@ -773,8 +772,8 @@ func (n *DedupStage) needsUpdate(entry *nflogpb.Entry, firing, resolved map[uint
 
 	// new alerts in the group
 	if !entry.IsFiringSubset(firing) {
-		// If the previous entry has no firing alerts, it was a reason and we
-		// should treat thisas the first notification for the group.
+		// If the previous entry has no firing alerts, it was a resolution and we
+		// should treat this as the first notification for the group.
 		if len(entry.FiringAlerts) == 0 {
 			return ReasonFirstNotification
 		}
@@ -861,6 +860,7 @@ func (n *DedupStage) Exec(ctx context.Context, _ *slog.Logger, alerts ...*types.
 	}
 
 	updateReason := n.needsUpdate(entry, firingSet, resolvedSet, repeatInterval)
+	ctx = WithNotificationReason(ctx, updateReason)
 
 	if updateReason == ReasonFirstNotification {
 		ctx = WithNflogStore(ctx, nflog.NewStore(nil))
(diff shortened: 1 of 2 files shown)


notify/notify.go
```diff
@@ -715,15 +725,60 @@ func hashAlert(a *types.Alert) uint64 {
...
+			return ReasonFirstNotification
+		}
+		return ReasonDoNotNotify
 	}
 
+	// new alerts in the group
 	if !entry.IsFiringSubset(firing) {
-		return true
+		// If the previous entry has no firing alerts, it was a reason and we
+		// should treat thisas the first notification for the group.
```
TheMeier:
```suggestion
		// should treat this as the first notification for the group.
```

notify/notify.go
```diff
@@ -193,6 +194,10 @@ func WithRouteID(ctx context.Context, routeID string) context.Context {
 	return context.WithValue(ctx, keyRouteID, routeID)
 }
 
+func WithNotificationReason(ctx context.Context, reason NotifyReason) context.Context {
```
TheMeier, line 197: where is this actually used to add the reason to and existing context?

notify/notify.go
TheMeier:
```suggestion
		// should treat this as the first notification for the group.
```

notify/notify.go
```diff
@@ -715,15 +725,60 @@ func hashAlert(a *types.Alert) uint64 {
...
+	}
+}
+
+func (n *DedupStage) needsUpdate(entry *nflogpb.Entry, firing, resolved map[uint64]struct{}, repeat time.Duration) NotifyReason {
 	// If we haven't notified about the alert group before, notify right away
 	// unless we only have resolved alerts.
 	if entry == nil {
-		return len(firing) > 0
+		newAlerts := len(firing) > 0
+		if newAlerts {
```
TheMeier: Looks to me like `newAlerts` is used exactly once so why not:

```suggestion
		if len(firing) > 0 {
```

Or `newAlerts := len(firing)` and then use `newAlerts` in comparison conditions on line 768 and 787

notify/notify.go
```diff
@@ -715,15 +725,60 @@ func hashAlert(a *types.Alert) uint64 {
...
+			return ReasonFirstNotification
+		}
+		return ReasonDoNotNotify
 	}
 
+	// new alerts in the group
 	if !entry.IsFiringSubset(firing) {
-		return true
+		// If the previous entry has no firing alerts, it was a reason and we
```
TheMeier: The wording is a bit strange. Reads to me as:  **it** refering to the `entry` is of type `NotifyReason`

notify/notify.go
Spaceman1701: wow yep, this is unset. I'll fix this + add to the `TestDedupStage` that this gets set...

notify/notify.go
Spaceman1701: Ah, this is a typo - it's meant to say `this is a resolution`

TheMeier (approved):

PR #4972, opened 2026-02-07
Experiment DO NOT MERGE move webhook config type into webhook package

TheMeier:
This is an experiment to evaluate a route how notifier specific code could be better spearated.

### DIFF
--- a/notify/webhook/config_test.go
+++ b/notify/webhook/config_test.go
@@ -1,0 +1,108 @@
+// Copyright The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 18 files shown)


notify/webhook/config_test.go
```diff
@@ -0,0 +1,108 @@
+// Copyright The Prometheus Authors
```
SoloJacobs, line 1: I think just creating this file with Unmarshal logic would be a good minimal change. The other changes can be done in a follow-up. That would unblock #4989 @siavashs @TheMeier Thoughts?

TheMeier: Well if you want to use receiver functions of the notifier-config types I would advise to keep everything in one place. Maybe we can set a side a 5-10 timebox on Friday to talk about this experiment.

TheMeier: See https://github.com/prometheus/alertmanager/pull/5004

PR #4617, opened 2025-10-21
Remove abandoned react-app

SuperQ:
The experimental react-app has been abandoned. Remove the code to avoid continued maintence.

PR #4974, opened 2026-02-08
Add feature to build and enable UI V2

sysadmind:
Adds the Makefile and UI embedding components to serve the new UI based on Mantine. This can be enabled when built with the tag `builtinassets`. This is a close match to what Prometheus does to build and embed the UI.

The UI is still very early, so the utility is very limited, but this will allow testing the features more easily as they become available.

### DIFF
--- a/featurecontrol/featurecontrol.go
+++ b/featurecontrol/featurecontrol.go
@@ -1,4 +1,4 @@
-// Copyright Prometheus Team
+// Copyright The Prometheus Authors
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
 // You may obtain a copy of the License at
--- a/ui/ui.go
+++ b/ui/ui.go
@@ -37,9 +37,6 @@ var Assets = func() http.FileSystem {
 	case "alertmanager":
 		// When running Alertmanager (without built-in assets) from the repo root.
 		assetsPrefix = "./ui"
-	case "web":
-		// When running web tests.
-		assetsPrefix = "./ui"
 	case "ui":
 		// When generating statically compiled-in assets.
 		assetsPrefix = "./"
--- a/ui/web.go
+++ b/ui/web.go
@@ -140,11 +140,6 @@ func RegisterUIV2(router *route.Router, logger *slog.Logger) {
 			fmt.Fprintf(w, "Error reading React index.html: %v", err)
 			return
 		}
-		// replacedIdx := bytes.ReplaceAll(idx, []byte("CONSOLES_LINK_PLACEHOLDER"), []byte(h.consolesPath()))
-		// replacedIdx = bytes.ReplaceAll(replacedIdx, []byte("TITLE_PLACEHOLDER"), []byte(h.options.PageTitle))
-		// replacedIdx = bytes.ReplaceAll(replacedIdx, []byte("AGENT_MODE_PLACEHOLDER"), []byte(strconv.FormatBool(h.options.IsAgent)))
-		// replacedIdx = bytes.ReplaceAll(replacedIdx, []byte("READY_PLACEHOLDER"), []byte(strconv.FormatBool(h.isReady())))
-		// replacedIdx = bytes.ReplaceAll(replacedIdx, []byte("LOOKBACKDELTA_PLACEHOLDER"), []byte(model.Duration(h.options.LookbackDelta).String()))
 		w.Write(idx)
 	}
 


siavashs:
CI is failing even after a re-run.
It is not clear why the build is failing for elm, it does not seem that there are any changes to that:
```
cd /__w/alertmanager/alertmanager/ui/app && make script.js
make[1]: Entering directory '/__w/alertmanager/alertmanager/ui/app'
>> building elm-env docker image
/bin/sh: 3: docker: not found
make[1]: *** [Makefile:17: elm-env] Error 127
make[1]: Leaving directory '/__w/alertmanager/alertmanager/ui/app'
make: *** [Makefile:67: ui/app/script.js] Error 2
```

sysadmind:
The problem is that this is now happening in the build process where it was not before and it's already inside docker so it doesn't have access to run docker. I think the logical option is to remove the elm build from the regular build process. The other option would be to enable the docker in docker build.

featurecontrol/featurecontrol.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2023 Prometheus Team
+// Copyright Prometheus Team
```
siavashs:
```suggestion
// Copyright The Prometheus Authors
```

scripts/compress_assets.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
```
siavashs, line 1: We removed this file in #4617 do we need to add it back now again?

ui/web.go
```diff
@@ -23,45 +24,27 @@ import (
...
 
-	r.Get("/lib/*path", func(w http.ResponseWriter, req *http.Request) {
-		disableCaching(w)
+// Register registers handlers to serve files for the web interface.
+func Register(r *route.Router, reloadCh chan<- chan error, logger *slog.Logger, enableUIV2 bool) {
+	if enableUIV2 {
+		RegisterUIV2(r, logger)
+	} else {
+		RegisterUIV1(r, logger)
+	}
```
siavashs, line 45: Should we use `Elm` and `React` for naming these to be more clear?

ui/web.go
```diff
@@ -101,6 +84,86 @@ func Register(r *route.Router, reloadCh chan<- chan error, logger *slog.Logger)
...
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			fmt.Fprintf(w, "Error reading React index.html: %v", err)
+			return
+		}
+		// replacedIdx := bytes.ReplaceAll(idx, []byte("CONSOLES_LINK_PLACEHOLDER"), []byte(h.consolesPath()))
+		// replacedIdx = bytes.ReplaceAll(replacedIdx, []byte("TITLE_PLACEHOLDER"), []byte(h.options.PageTitle))
+		// replacedIdx = bytes.ReplaceAll(replacedIdx, []byte("AGENT_MODE_PLACEHOLDER"), []byte(strconv.FormatBool(h.options.IsAgent)))
+		// replacedIdx = bytes.ReplaceAll(replacedIdx, []byte("READY_PLACEHOLDER"), []byte(strconv.FormatBool(h.isReady())))
+		// replacedIdx = bytes.ReplaceAll(replacedIdx, []byte("LOOKBACKDELTA_PLACEHOLDER"), []byte(model.Duration(h.options.LookbackDelta).String()))
```
siavashs: Leftovers?

ui/ui.go
```diff
@@ -0,0 +1,62 @@
...
+		panic(err)
+	}
+	var assetsPrefix string
+	switch filepath.Base(wd) {
+	case "alertmanager":
+		// When running Alertmanager (without built-in assets) from the repo root.
+		assetsPrefix = "./ui"
+	case "web":
+		// When running web tests.
+		assetsPrefix = "./ui"
```
siavashs: Is this used or dead code?

Makefile
```diff
@@ -27,13 +30,34 @@ STATICCHECK_IGNORE =
 build-all: assets apiv2 build
 
 .PHONY: build
-build: common-build
+build: assets-compress common-build
```
siavashs, line 33: Currently the project ships pre-compiled assets bundle so UI does not have to be built each time.

ui/web.go
```diff
@@ -101,6 +84,86 @@ func Register(r *route.Router, reloadCh chan<- chan error, logger *slog.Logger)
...
+
+func RegisterUIV2(router *route.Router, logger *slog.Logger) {
+	homePage := "/alerts"
+	router.Get("/", func(w http.ResponseWriter, r *http.Request) {
+		http.Redirect(w, r, homePage, http.StatusFound)
+	})
+
+	reactAssetsRoot := "/static/mantine-ui"
+
+	serveReactApp := func(w http.ResponseWriter, _ *http.Request) {
```
siavashs, line 128: Missing cache-control headers?

scripts/compress_assets.sh
sysadmind: It's used in prometheus for the bundling of static assets into a go file. I think we should keep it.

ui/ui.go
sysadmind: Good question. I think we can drop it for now - it was a copy/paste from Prometheus.

ui/web.go
sysadmind: I'm certainly open to renaming them. I don't know that I love the word "react" though because in some future, there may be a UI v3 that is also based on react.

ui/web.go
sysadmind: Yep. I will clean these up. They mat need to be here at some point in the future, but there's no reason to leave the comments sitting around.

Makefile
sysadmind: I think we should re-evaluate that for the mantine based UI. The build output has unique filenames per build. This will cause a lot of churn in git.

ui/web.go
sysadmind: What outcome would you like to see with those?

siavashs: Just noticed they are missing and wondered if we need them.

sysadmind: Personally, if we think they will provide value, I would want to see this happen towards the end of the project. That way we can test what benefits and what tradeoffs they cause.

scripts/compress_assets.sh
SoloJacobs: I don't think we should add `compress_assets.sh`.

* It is very complex, having to both generate .go files and running extra shell scripts as part of the build. It would be better to just directly use Vite for building the gzip files in a `dist` folder. This would allow us to easily add brotli compression later.

* Everytime the ui is opened, `alertmanager` will decompress the files *before sending them over the network*. This can be seen from the `curl` output:

```
curl -s -D - -H "Accept-Encoding: gzip" http://localhost:9093/assets/index-D3YozeYE.js -o /dev/null
HTTP/1.1 200 OK
Accept-Ranges: bytes
Content-Length: 530907
Content-Type: application/javascript
Date: Sat, 07 Mar 2026 19:22:22 GMT
```

So, we are missing out on the main benefit of compression: Reducing the number of TCP round trips.

* The benefits of compression are negligible:

```sh
❯ du -hs  assets/index-D3YozeYE.js.gz  assets/index-pOPVlMO9.css.gz index.html.gz
164K	assets/index-D3YozeYE.js.gz
32K	assets/index-pOPVlMO9.css.gz
4.0K	index.html.gz
❯ du -hs  assets/index-D3YozeYE.js  assets/index-pOPVlMO9.css index.html
520K	assets/index-D3YozeYE.js
216K	assets/index-pOPVlMO9.css
4.0K	index.html
```
Thus, your change will reduce the binary size by 500K .

TLDR; We should keep it simple and not do any compression. If we want to do compression, we should do it properly, and support streaming the files compressed over the network.

ui/ui.go
```diff
@@ -0,0 +1,59 @@
...
+package ui
+
+import (
+	"net/http"
+	"os"
+	"path"
+	"path/filepath"
+	"strings"
+
+	"github.com/shurcooL/httpfs/filter"
```
SoloJacobs, line 25: We should not be using `shurcool`. The files can be easily be routed to the correct location with a few lines of `go`.

ui/web.go
SoloJacobs: The caching headers are need such that users can seamlessly update their alertmanager binaries: If the browser caches the `index.html`, we will have situation that a browser will run an old `index.html` against a new alertmanager server. This will definitely cause failures, as far as I can see, because alertmanager will return 404s for these files:
```html
    <script type="module" crossorigin src="/assets/index-D3YozeYE.js"></script>
    <link rel="stylesheet" crossorigin href="/assets/index-pOPVlMO9.css">
```
We should add the cache-control headers now, since this cannot be fixed retroactively.

The other caches should be fine, since the index files contain the hashes.

ui/web.go
```diff
@@ -101,6 +84,81 @@ func Register(r *route.Router, reloadCh chan<- chan error, logger *slog.Logger)
...
+		indexPath := reactAssetsRoot + "/index.html"
+		f, err := Assets.Open(indexPath)
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			fmt.Fprintf(w, "Error opening React index.html: %v", err)
+			return
+		}
+		defer func() { _ = f.Close() }()
+		idx, err := io.ReadAll(f)
```
SoloJacobs, line 137: Nit: Why not use `io.Copy`

ui/web.go
```diff
@@ -101,6 +84,81 @@ func Register(r *route.Router, reloadCh chan<- chan error, logger *slog.Logger)
...
+
+	for _, p := range reactRouterPaths {
+		router.Get(p, serveReactApp)
+	}
+
+	reactStaticAssetsDir := "/assets"
+	// Static files required by the React app.
+	router.Get(reactStaticAssetsDir+"/*filepath", func(w http.ResponseWriter, r *http.Request) {
+		r.URL.Path = path.Join(reactAssetsRoot+reactStaticAssetsDir, route.Param(r.Context(), "filepath"))
+		fs := server.StaticFileServer(Assets)
```
SoloJacobs, line 157: I don't think we should create a `StaticFileServer` on every request.

featurecontrol/featurecontrol.go
```diff
@@ -81,6 +85,10 @@ func (f *Flags) EnableAutoGOMAXPROCS() bool {
 	return f.enableAutoGOMAXPROCS
 }
 
+func (f *Flags) EnableUIV2() bool {
```
SoloJacobs, line 88: I don't think UI is ready to be shipped with next release, so I would rather not include it in the release. Would it be possible for us to provide a separate build target instead?

ui/mantine-ui/src/components/ThemeSelector.tsx
```diff
@@ -0,0 +1,31 @@
+import { FC } from 'react';
```
SoloJacobs, line 1: Does this possibly belong in a different PR?

ui/web.go
```diff
@@ -101,6 +84,81 @@ func Register(r *route.Router, reloadCh chan<- chan error, logger *slog.Logger)
...
+	}
+
+	// Serve the React app.
+	reactRouterPaths := newUIReactRouterPaths
+
+	for _, p := range reactRouterPaths {
+		router.Get(p, serveReactApp)
+	}
+
+	reactStaticAssetsDir := "/assets"
```
SoloJacobs, line 153: nit: inline

ui/web.go
```diff
@@ -101,6 +84,81 @@ func Register(r *route.Router, reloadCh chan<- chan error, logger *slog.Logger)
...
+			fmt.Fprintf(w, "Error reading React index.html: %v", err)
+			return
+		}
+		w.Write(idx)
+	}
+
+	// Serve the React app.
+	reactRouterPaths := newUIReactRouterPaths
+
+	for _, p := range reactRouterPaths {
```
SoloJacobs, line 149: I'm wondering about this: Aren't we building a SPA? Typically, you would serve everything, not just a fixed set of subpaths, i.e., `localhost:9093/alerts/special-name`  should also serve the react app?

For now, serving 404 is fine I suppose, but we will definitely have to touch this part, once we build deep links. I am wondering whether it is better just have a general `/ui` path.

sysadmind: I think that `/ui` is a fair option if there is general agreement on that direction.

ui/mantine-ui/src/components/ThemeSelector.tsx
sysadmind: Yes I will pull it out.

scripts/compress_assets.sh
sysadmind: The binary file saving are small right now because of tree shaking. There is very little in the UI if you compile it from this PR. As more features and functionality get added, this will grow. I don't have hard data around what that will be, but if there's consensus around embedding the uncompressed files, I'll pull this out.

SoloJacobs:
Hey, just because I saw you might be active here: I did make some progress on what I think the UI should look like.
https://github.com/SoloJacobs/alertmanager/tree/example-ui

This change is not ready, yet: I put the first three commit into review, but the final commit contains the interesting changes:

* Zero-Overhead compression: We directly stream the compressed files from the binary, and don't decompress them unless the client requests it.
* Optimization to the caching, using Vites file hashing.
* Disallow ` go get github.com/prometheus/alertmanager/ui`.

The basic idea is that we have to move to vite sooner or later, so we might as well do it now and find out how it affects users.

ui/web.go
juliusv: Just a comment from the sidelines, since I don't have the capacity to get more deeply involved at the moment:

Since this routing code was taken from the Prometheus server: any UI URL parametrization in the Prometheus server happens only via HTTP parameters (since we encode a lot of different parameters in there on some of the pages, and then we just standardized on that single mechanism), so we only need to explicitly list a couple of fixed paths there. I wonder if for Alertmanager you really want to split this up and store some (prominent) parameters in the path and others (like detailed filters) in HTTP params, or whether it's nicer to just dump them all into HTTP params like in Prometheus and have one unified way of handling them all?

Aside from that: as a user I personally aesthetically prefer UI paths without a `/ui/...` prefix. Maybe the path routing can also be restructured to serve the UI for a set of patterns, or for everything that is *not* `/api/...`, `/-/healthy`, or one of the other special endpoints.

sysadmind:
I think based on the feedback, I want to take a different approach. I have opened a PR for the small UI change - #5106 . After the changes from @SoloJacobs above from the `example-ui` branch, I can rework this build in that spirit, in some smaller increments so that we can align on direction more easily.

PR #4975, opened 2026-02-08
Add Rootly notification integration

kwent:
## Summary

Add native [Rootly](https://rootly.com) receiver for Alertmanager, enabling users to configure `rootly_configs` directly in `alertmanager.yml` instead of using generic webhooks.

The implementation follows the established integration pattern and includes configuration, notifier, wiring, and comprehensive tests.

## Changes

**Configuration** (`config/notifiers.go`, `config/config.go`)
- `RootlyConfig` struct with `url`/`url_file`, `rootly_token`/`rootly_token_file`, `max_alerts`, `timeout`, and `http_config` fields
- `UnmarshalYAML` validation for mutual exclusivity and required fields
- `send_resolved` enabled by default
- `RootlyConfigs` field on `Receiver` struct with `HTTPConfig` inheritance

**Notifier** (`notify/rootly/rootly.go`)
- JSON payload with `version`, `groupKey`, `truncatedAlerts`, and alert data
- Bearer token authentication via `rootly_token` or `rootly_token_file`
- All alerts in a group are sent in full — Rootly processes each alert individually
- Retry on 429/5xx, no retry on other 4xx
- Error response parsing from Rootly API
- Secret redaction via `notify.RedactURL()`

**Wiring** (`config/receiver/receiver.go`, `notify/notify.go`)
- Import and builder loop in `BuildReceiverIntegrations()`
- `"rootly"` registered in notification metrics

**Tests** (`notify/rootly/rootly_test.go`)
- 9 test functions covering retry logic, secret redaction, file-based URL/token loading, alert truncation, notification flow, retry scenarios, error detail parsing, and multi-alert delivery
- Uses `require` (not `assert`), follows Prometheus testing conventions

## Example configuration

```yaml
receivers:
  - name: rootly
    rootly_configs:
      - url: https://rootly.com/webhooks/incoming/alertmanager_webhooks
        rootly_token: <token>
```

## Checklist

- [x] Tests pass (`go test ./notify/rootly/`)
- [x] Linter passes (`make lint`)
- [x] Apache 2.0 license headers on all new files
- [x] DCO sign-off on commits

Signed-off-by: Quentin Rousseau <contact@quent.in>

### DIFF
--- a/notify/rootly/rootly.go
+++ b/notify/rootly/rootly.go
@@ -1,0 +1,241 @@
+// Copyright 2025 Prometheus Team
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rootly
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"log/slog"
+	"net/http"
+	"os"
+	"strings"
+
+	commoncfg "github.com/prometheus/common/config"
+
+	"github.com/prometheus/alertmanager/config"
+	"github.com/prometheus/alertmanager/notify"
+	"github.com/prometheus/alertmanager/template"
+	"github.com/prometheus/alertmanager/types"
+)
+
+const (
+	// MaxPayloadSize is the maximum size of the JSON payload Rootly accepts (256KB).
+	maxPayloadSize = 256 * 1024
+)
+
+// Notifier implements a Notifier for Rootly.
+type Notifier struct {
+	conf    *config.RootlyConfig
+	tmpl    *template.Template
+	logger  *slog.Logger
+	client  *http.Client
+	retrier *notify.Retrier
+}
+
+// New returns a new Rootly notifier.
+func New(conf *config.RootlyConfig, t *template.Template, l *slog.Logger, httpOpts ...commoncfg.HTTPClientOption) (*Notifier, error) {
+	// conf.HTTPConfig is likely to be the global shared HTTPConfig, so we take a
+	// copy to avoid modifying it.
+	httpConfig := *conf.HTTPConfig
+
+	// If a rootly token is provided, we use that one instead of whatever configuration is included in `http_config`.
+	var token string
+	if conf.RootlyToken != "" {
+		token = string(conf.RootlyToken)
+	}
+
+	if conf.RootlyTokenFile != "" {
+		content, err := os.ReadFile(conf.RootlyTokenFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read rootly_token_file: %w", err)
+		}
+		token = strings.TrimSpace(string(content))
+	}
+
+	if token != "" {
+		httpConfig.Authorization = &commoncfg.Authorization{
+			Type:        "Bearer",
+			Credentials: commoncfg.Secret(token),
+		}
+	}
+
+	client, err := notify.NewClientWithTracing(httpConfig, "rootly", httpOpts...)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Notifier{
+		conf:   conf,
+		tmpl:   t,
+		logger: l,
+		client: client,
+		// Always retry on 429 (rate limiting) and 5xx response codes.
+		retrier: &notify.Retrier{
+			RetryCodes:        []int{http.StatusTooManyRequests},
+			CustomDetailsFunc: errDetails,
+		},
+	}, nil
+}
+
+// Message defines the JSON object sent to Rootly endpoints.
+type Message struct {
+	*template.Data
+
+	// The protocol version.
+	Version         string `json:"version"`
+	GroupKey        string `json:"groupKey"`
+	TruncatedAlerts uint64 `json:"truncatedAlerts"`
+}
+
+func truncateAlerts(maxAlerts uint64, alerts []*types.Alert) ([]*types.Alert, uint64) {
+	if maxAlerts != 0 && uint64(len(alerts)) > maxAlerts {
+		return alerts[:maxAlerts], uint64(len(alerts)) - maxAlerts
+	}
+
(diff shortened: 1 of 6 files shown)


siavashs:
Hello @kwent, thank you for submitting this PR.
Have you already tried to use the [Webhook integration](https://prometheus.io/docs/alerting/latest/configuration/#webhook_config) to send notification to Rootly and was there any limitation that required adding a custom integration?

TheMeier:
Since it seems he is CTO and co-founder of rootly he would be in a very good position to actually consider a custom ingestion endpoint at rootly for native webhooks ;) IMHO would that would give much more flexibility, since they can change the behaviour as needed and adapt it independently to match customer demands.

notify/rootly/rootly.go
```diff
@@ -0,0 +1,241 @@
...
+	if buf.Len() <= maxPayloadSize {
+		return buf, nil
+	}
+
+	originalSize := buf.Len()
+
+	// Drop all but the first alert in the message. For most use cases, a single
+	// alert will be created in Rootly for the group, so including more than
+	// one alert in that group is useful but non-essential.
+	msg.Alerts = msg.Alerts[:1]
```
TheMeier: I find that to be and odd design. For one the first alert might have status `reseloved`. But more important you are missing a lot of information that could help identifying causes quickly. 
A constructed example: an alertgroup for node_exporter node being down has multiple alerts for different nodes, on all alerts a label is set that identiefies some common feature (like switch, power-circuit or a physical location). This information will get lost.

As I understand it rootly is heavy on AI, I would assume you would want as much data you can get for inference

siavashs:
> Since it seems he is CTO and co-founder of rootly he would be in a very good position to actually consider a custom ingestion endpoint at rootly for native webhooks ;) IMHO would that would give much more flexibility, since they can change the behaviour as needed and adapt it independently to match customer demands.

In that case they can expose an Alertmanager webhook compatible endpoint on Rootly which would just work with our existing webhook payloads, no?

notify/rootly/rootly.go
kwent: will take a look if we supports this already deeper

kwent:
> > Since it seems he is CTO and co-founder of rootly he would be in a very good position to actually consider a custom ingestion endpoint at rootly for native webhooks ;) IMHO would that would give much more flexibility, since they can change the behaviour as needed and adapt it independently to match customer demands.
> 
> In that case they can expose an Alertmanager webhook compatible endpoint on Rootly which would just work with our existing webhook payloads, no?

Already doing it this way today but some of our customers are always looking more native.

notify/rootly/rootly.go
kwent: Good point! Rootly processes each alert in the payload individually — it iterates over the `alerts` array and creates a separate Rootly alert for each one. Dropping alerts would lose valuable context.

Removed the payload size truncation entirely in cf7ef098. The `encodeMessage` function now simply encodes the JSON without any truncation logic.

siavashs:
> > In that case they can expose an Alertmanager webhook compatible endpoint on Rootly which would just work with our existing webhook payloads, no?
> 
> Already doing it this way today but some of our customers are always looking more native.

This still does not answer my first comment on this PR: what does this native integration provide which cannot be achieved with the webhook integration?
For example is it templating? There are plans to add support for templating to the webhook receiver, so that might satisfy such a requirement without needing a dedicated integration.

kwent:
No difference. I think it's mostly a trust issue from one of customer who raised this issue. Most of our competitors have a native notifier it and we don't, it does not establish good trust and good partnership between us raising questions from some of them.

siavashs:
> No difference. I think it's mostly a trust issue from one of customer who raised this issue. Most of our competitors have a native notifier it and we don't, it does not establish good trust and good partnership between us raising questions from some of them.

Thank you for clarification.

Unfortunately we cannot accept this integration at this time. But we kindly ask you to open a `feature request` issue to propose adding this integration first. Please be specific in the issue on why the integration is needed and what value it brings to the end-user.
The maintainers will make a decision to include this integration in the future if there is enough community demand for the integration to be added.

In the meantime feel free to state on the official Rootly documentation about officially supported integration with Alertmanager using the `Webhook` integration and link back to our official documentation.
You could also list Rootly on the list of integrations that support Alertmanager's Webhook integration on Prometheus website: [link](https://prometheus.io/docs/operating/integrations/#alertmanager-webhook-receiver), [source](https://github.com/prometheus/docs/blob/main/docs/operating/integrations.md#alertmanager-webhook-receiver)
This should help with any potential trust issues that Rootly users or customers might have about integration between Alertmanager with Rootly.

Thank you again for considering to contribute to Alertmanager

kwent:
Thank you for your response.                                                                                                                                                                                  
                                                                                                                                                                                                                
> For example is it templating? There are plans to add support for templating to the webhook receiver, so that might satisfy such a requirement without needing a dedicated integration.                        
                                                                                                                                                                                                                
Yes, templating is exactly what our customers need. We have a customer currently migrating from Opsgenie to Rootly who is blocked because Opsgenie's native Alertmanager integration supports Go-templated fields (description, details) — they use these to inject cluster-specific metadata and construct deep-links back to their observability platform. The generic webhook_configs can't express this since it sends a fixed payload.

Is there an ETA or tracking issue for adding templating support to the webhook receiver? Happy to contribute to that effort as well if it's the preferred path forward. Alternatively, we could add templating support directly to this PR if that would make the case for a native integration stronger.

siavashs:
Thank you for sharing more context.
The templating support in Webhook integration was discussed last week and should be added soon.
I will check with the team if they have already started working on it.

ethanh-canva:
Hi @siavashs, just going to inform from our perspective as the customer.

Our Prometheus alert manager setup is contained within a K8s cluster that contains metadata that links to our custom Grafana scenes application for troubleshooting. In this case we use this metadata on the cluster to construct a link that directs the user form the alert to our Grafana scenes application which contains our alert rules.

The current generic webhook does hold back our development given the fixed payload is unmodifiable from within the cluster. This restricts unique alert receivers to be held in accordance to this fixed payload. If the future development allows for both injection within the cluster and being able to pass said data into the payload it should meet our requirements. Otherwise we may require a Rootly specific integration to ensure the functionality we require.

Side note: would it be possible if we could see the branch for generic webhook upgrades? It could help ensure the functionality we need is met and speed up our development time.

kwent:
@siavashs Could you provide more guidance ? ETA's? Workaround?

akhileshkrishnan:
For anyone who lands here like I did looking for a custom payload template in webhooks for Rootly: it was merged in https://github.com/prometheus/alertmanager/pull/5011

PR #4980, opened 2026-02-08
feat(silence): add dedicated silencer cache

siavashs:
Add a dedicated cache to silencer to drop the dependency on the global marker.
For now this should make get API calls non-blocking for silencer.
This will also enable us to eventually deprecate and remove the global marker.

The new silencer cache is evicted by hooking into provider PostDelete callback.

This change also removes the logic to recreate silencer on reload which was not needed.

### DIFF
--- a/silence/cache.go
+++ b/silence/cache.go
@@ -43,21 +43,21 @@ func (e *cacheEntry) count() int {
 // result is based on.
 type cache struct {
 	entries map[model.Fingerprint]*cacheEntry
-	mu      sync.RWMutex
+	mtx     sync.RWMutex
 }
 
 // delete removes the cacheEntry for the given fingerprint.
 func (c *cache) delete(fp model.Fingerprint) {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
 	delete(c.entries, fp)
 }
 
 // get returns the cacheEntry for the given fingerprint.
 // The returned entry is not a copy, so it should not be modified.
 func (c *cache) get(fp model.Fingerprint) *cacheEntry {
-	c.mu.RLock()
-	defer c.mu.RUnlock()
+	c.mtx.RLock()
+	defer c.mtx.RUnlock()
 	if e, found := c.entries[fp]; found {
 		return e
 	}
@@ -66,7 +66,7 @@ func (c *cache) get(fp model.Fingerprint) *cacheEntry {
 
 // set sets the cacheEntry for the given fingerprint.
 func (c *cache) set(fp model.Fingerprint, entry *cacheEntry) {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
 	c.entries[fp] = entry
 }


silence/cache.go
```diff
@@ -0,0 +1,69 @@
...
+
+package silence
+
+import (
+	"sync"
+
+	"github.com/prometheus/common/model"
+)
+
+type cacheEntry struct {
```
ultrotter, line 24: maybe let's add some comment explaining the fields?

silence/cache.go
```diff
@@ -0,0 +1,69 @@
...
+		pendingIDs: pendingIDs,
+		version:    version,
+	}
+}
+
+func (e *cacheEntry) count() int {
+	return len(e.activeIDs) + len(e.pendingIDs)
+}
+
+type cache struct {
```
ultrotter, line 44: Could we compare between map with mutex and sync.Map? For many reads it could perform better...

silence/cache.go
```diff
@@ -0,0 +1,69 @@
...
+	entry := cacheEntry{}
+	if e, found := c.entries[fp]; found {
+		entry.version = e.version
+		entry.activeIDs = make([]string, len(e.activeIDs))
+		copy(entry.activeIDs, e.activeIDs)
+		entry.pendingIDs = make([]string, len(e.pendingIDs))
+		copy(entry.pendingIDs, e.pendingIDs)
+	}
+	return entry
+}
```
ultrotter, line 65: Do we need to copy here? Or could we assume the reader won't modify the data, given it's private do this module?

silence/cache.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"sync"
+
+	"github.com/prometheus/common/model"
+)
+
+type cacheEntry struct {
+	activeIDs  []string
+	pendingIDs []string
```
ultrotter: We seem to always concatenate these? Would it make sense to only keep the concatenated version in the cache? We always reevaluate their current state so maybe this distinction is not useful anymore...

  The only place we care about active vs pending separation is:
  - At the end (line 285) when updating the marker with only activeIDs
  - But that comes from the re-evaluation loop (lines 262-273), not from the cache

silence/silence.go
ultrotter, line 264: In this loop we could even create just an activeIDs, for the marker, and an allIDs for the cache.

silence/cache.go
siavashs: So currently there is no modification. If we don't copy we pass the underlying `array` along with the new entry returned which can allow the caller to potentially modify an entry in future changes and mutate the data.

ultrotter: I would rather try to find a lint approach to avoid us modifying it in the future, but also rely on the fact that we won't, rather than doing extra data copies while holding a lock when we could just do return c.entries[fp] and be done with it, and tell the caller to treat it as read only. Especially since it's a private struct with private fields in our package, so we just need to make sure our package itself doesn't modify it. or copies it before passing it out of the package itself (which it wouldn't at the moment especially if we change the approach to single slice).

silence/cache.go
siavashs: Fixed, we now keep one slice for all ids.

silence/silence.go
siavashs: Good idea, I implemented it as you mentioned.

silence/cache.go
siavashs: I will try to create a benchmark and compare them.

silence/cache.go
siavashs: Fine with me, I updated the code to return a pointer to an existing or new entry and added documentation that the returned entry must not be modified.

silence/cache.go
siavashs: As we discussed on Slack, `sync.Map` would not be a good option here.
We can always revisit.

silence/cache.go
```diff
@@ -0,0 +1,72 @@
...
+// count returns the number of silence IDs in the cacheEntry.
+func (e *cacheEntry) count() int {
+	return len(e.silenceIDs)
+}
+
+// cache stores the IDs of silences that match an alert and the version of the silences state the
+// result is based on.
+type cache struct {
+	entries map[model.Fingerprint]*cacheEntry
+	mu      sync.RWMutex
```
Spaceman1701: tiny nitpick: we call this `mtx` everywhere else in alertmanager. I think we should be consistent where we can

silence/silence.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2016 Prometheus Team
+// Copyright The Prometheus Authors
```
Spaceman1701, line 1: is this meant to have changed?

cmd/alertmanager/main.go
```diff
@@ -352,7 +354,7 @@ func run() int {
 		marker,
 		*alertGCInterval,
 		*perAlertNameLimit,
-		nil,
+		silencer,
```
Spaceman1701, line 357: I _think_ it's fine, but one thing to note is that the alert deleted hook is called while we have a mutex on the entire alert provider: https://github.com/prometheus/alertmanager/blob/a3a156d6f2f218fe2985925f63ddf5e8224eb20a/provider/mem/mem.go#L172-L183

We should be careful that this doesn't deadlock

silence/silence.go
siavashs: Yes, that is the official new copyright header for Prometheus.

cmd/alertmanager/main.go
siavashs: I think @grobinson-grafana advised against using the callbacks but we do not have a better option yet.
We could spin up a go routine in silencer and return the callback quickly so we do not block the provider.

siavashs: Looking at the code again, same thing is happening with marker, so we are just adding an extra call for a map delete which should be fast.
Silencer itself does not hold locks long on its cache so I think this is going to be fine for now.
One thing we can improve is to batch deletes to both marker and the silence cache.
I can open a separate PR for that as a follow up.

grobinson-grafana: Yeah just be careful with callbacks, we fixed I think a total of 3-4 race condition and deadlock bugs related to callbacks either calling itself (attempting to re-aquire its own mutex) or callbacks from A calling B and callback from B calling A (hence deadlock).

siavashs: I have a draft change which removes locks from callbacks since the caller doesn't really need to hold its internal lock while calling the callback.
This should improve the performance and avoid any potential deadlocks in future.

Maybe when we get the new `EventInterface`, we could potentially hook into that to eventually delete alerts marked as resolved in different components.

PR #4985, opened 2026-02-09
fix: omit empty text field in Mattermost notifications

slashpai:
Add omitempty tag to Text field to prevent sending empty text values to Mattermost, which causes 400 errors. Per the Mattermost API spec, text is only required if attachments are not set.

Related to prometheus-operator/prometheus-operator#8363


TheMeier:
LGTM. Although with the default template, the text field is never empty. So one could argue (which I am not) it is working as designed and users must ensure that custom templates in `mattermost_config.text` must not produce empty strings.

PR #4986, opened 2026-02-09
chore: add pr templates

siavashs:


TheMeier:
LGTM, but the benchmark requirement is kind of a high hurdle

siavashs:
> LGTM, but the benchmark requirement is kind of a high hurdle

We can skip it for changes that are not in the critical pass.
For example notifier integrations don't need a benchmark.
Should I reword it to indicate when it is needed?

TheMeier:
Maybe something like "additionally new benchmark and benchmark results _might_ be required". Or list "critcal" components (nflog, dispatcher, gossip...?)

It is really hard to communicate the intent without freightning of potentiall contributors :/

But I'd be also OK with it as is.

siavashs:
> Maybe something like "additionally new benchmark and benchmark results _might_ be required". Or list "critcal" components (nflog, dispatcher, gossip...?)

Added a new section for this which asks if this change affects performance and the it requires benchmarks under it.

PR #4995, opened 2026-02-10
chore: cleanup proto generation

Spaceman1701:
Now that we've moved to buf for proto code gen, we can cleanup the generation logic a little bit.

This change:
1. Merges all the proto generations into one `buf.yaml` using Buf's module concept
2. Removes the post-run of `goimports` on generated code. I don't think we care too much about the ordering of imports in generated code.

I also thought about using a Buf remote plugin for code generation so that we don't need to do the `go install` dance in `genproto.sh`, but I decided to hold off on that for now. 

The diffs for each generated file come from `goimports` re-ordering the imports.

### DIFF
--- a/scripts/genproto.sh
+++ b/scripts/genproto.sh
@@ -13,17 +13,10 @@ fi
 pushd "internal/tools"
 INSTALL_PKGS="github.com/bufbuild/buf/cmd/buf golang.org/x/tools/cmd/goimports google.golang.org/protobuf/cmd/protoc-gen-go"
 for pkg in ${INSTALL_PKGS}; do
-    go install "$pkg"
+  go install "$pkg"
 done
 popd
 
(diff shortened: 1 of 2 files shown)


scripts/genproto.sh
```diff
@@ -13,17 +13,10 @@ fi
 pushd "internal/tools"
 INSTALL_PKGS="github.com/bufbuild/buf/cmd/buf golang.org/x/tools/cmd/goimports google.golang.org/protobuf/cmd/protoc-gen-go"
```
ultrotter, line 14: Do we need all the packages, including goimports? Also @siavashs and @SoloJacobs

SoloJacobs: Ah, good attention to detail. Created a fix here: https://github.com/prometheus/alertmanager/pull/5008

PR #4998, opened 2026-02-11
Support global Mattermost webhook url

guoard:
This PR adds global configuration support for Mattermost webhook url in Alertmanager.

Until now, `mattermost_webhook_url` and `mattermost_webhook_url_file` could only be set per receiver, which could lead to redundant configuration when multiple receivers used the same webhook.


madest92:
Lol. I waited 2 weeks with the same merge request(

PR #5000, opened 2026-02-11
Fix receiver index in config_test.go error logs

guoard:


config/config_test.go
```diff
@@ -1039,7 +1039,7 @@ func TestVictorOpsDefaultAPIKey(t *testing.T) {
 		t.Fatalf("Invalid victorops key: %s\nExpected: %s", conf.Receivers[0].VictorOpsConfigs[0].APIKey, defaultKey)
 	}
 	if overrideKey != conf.Receivers[1].VictorOpsConfigs[0].APIKey {
-		t.Errorf("Invalid victorops key: %s\nExpected: %s", conf.Receivers[0].VictorOpsConfigs[0].APIKey, string(overrideKey))
+		t.Errorf("Invalid victorops key: %s\nExpected: %s", conf.Receivers[1].VictorOpsConfigs[0].APIKey, string(overrideKey))
```
siavashs, line 1042: Since we are attempting to fix this, let's use `require.Equal()` instead of `if` statements for all tests.

guoard: This pull request addresses a bug fix. The suggestion you mentioned is a refactor and would be better handled in a separate pull request.

siavashs: Fair enough!

PR #5009, opened 2026-02-12
mattermost: flatten attachments into top-level config, keep BC

madest92:
This PR refactors the Mattermost receiver configuration by flattening the attachment structure(like **slack_configs**).

Previously, message formatting required configuring fields inside `attachments[0]`.
Now, commonly used fields such as `title, title_link, color, and text` are promoted to the top level of `mattermost_configs`.

Additionally, more default templates are provided to reduce the need for repetitive manual configuration.

**Example with default template:**
```
receivers:
- name: 'mm'
  mattermost_configs:
  - send_resolved: false
```
image

## Breaking change

This change modifies the configuration schema of `mattermost_configs`.

Old configuration:
```
mattermost_configs:
  - text: ''
    attachments:
      - color: '{{ if eq .Status "firing" }}danger{{ else }}good{{ end }}'
        title: '{{ .CommonAnnotations.summary }}'
        title_link: '{{ template "__alertmanagerURL" . }}'
        text: '{{ range .Alerts }}{{ .Annotations.description }}{{ end }}'
```
New configuration:
```
mattermost_configs:
  - title: '{{ .CommonAnnotations.summary }}'
    title_link: '{{ template "__alertmanagerURL" . }}'
    color: '{{ if eq .Status "firing" }}danger{{ else }}good{{ end }}'
    text: '{{ range .Alerts }}{{ .Annotations.description }}{{ end }}'
```
Users must update existing configurations accordingly.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1071,10 +1071,6 @@ webhook_url_file: <filepath>
 # Defaults to the username set during webhook creation; if no username was set during creation, webhook is used.
 [ username: <string> | default = '' ]
 
-# Markdown-formatted message to display in the post.
-# To trigger notifications, use @<username>, @channel, and @here like you would in other Mattermost messages.
-text: <tmpl_string> | default = '{{ template "mattermost.default.text" . }}'
-
 # Overrides the profile picture the message posts with.
 [ icon_url: <string> | default = '' ]
 
--- a/notify/mattermost/mattermost.go
+++ b/notify/mattermost/mattermost.go
@@ -153,6 +153,7 @@ func (n *Notifier) Notify(ctx context.Context, alert ...*types.Alert) (bool, err
 }
 
 func (n *Notifier) createRequest(tmpl func(string) string) *request {
+	text := tmpl(n.conf.Text)
 	req := &request{
 		Channel:   tmpl(n.conf.Channel),
 		Username:  tmpl(n.conf.Username),
@@ -208,15 +209,15 @@ func (n *Notifier) createRequest(tmpl func(string) string) *request {
 			}
 
 			req.Attachments[idxAtt] = att
-			req.Text = tmpl(n.conf.Text)
+			req.Text = text
 		}
 	} else {
 		req.Attachments = make([]attachment, 1)
 		att := attachment{
+			Text:       text,
 			Fallback:   tmpl(n.conf.Fallback),
 			Color:      tmpl(n.conf.Color),
 			Pretext:    tmpl(n.conf.Pretext),
-			Text:       tmpl(n.conf.Text),
 			AuthorName: tmpl(n.conf.AuthorName),
 			AuthorLink: tmpl(n.conf.AuthorLink),
 			AuthorIcon: tmpl(n.conf.AuthorIcon),


madest92:
@TheMeier hi
We discussed this [here](https://github.com/prometheus/alertmanager/pull/4917#issuecomment-3834571344)

TheMeier:
Why is asset/assets_vfsdata.go here?

TheMeier:
> Users must update existing configurations accordingly.

That does not sound good to me

madest92:
> Why is asset/assets_vfsdata.go here?

I updated the templates. Without this, alertmanager would not see new templates.

madest92:
> That does not sound good to me

I know. Me too. But the current configuration isn't convenient via attachments. I still use **slack_configs**.
```
global:
  slack_api_url: '<mattermost_webhook>'

receivers:
  - name: 'critical'
    slack_configs:
    - send_resolved: true
      title: '{{...}}'
      text: '{{...}}'
``` 

Mattermost was introduced not long ago, so I don't think it's a big problem.

siavashs:
> > That does not sound good to me
> 
> I know. Me too. But the current configuration isn't convenient via attachments. I still use **slack_configs**.
> ...
> Mattermost was introduced not long ago, so I don't think it's a big problem.

That is not a good excuse, we need to make changes backwards compatible.
The best approach is to deprecate the old fields and introduce the new ones.
Eventually we can remove the deprecated fields in a future release where we announce the breaking change.

madest92:
> That is not a good excuse, we need to make changes backwards compatible.

Agreed. I'll make it backwards compatible.

madest92:
@siavashs 

I added backward compatibility

1.  Old behavior when user uses attachment

image

2. New behavior without using attachment. Default templates are already in use
image

madest92:
@TheMeier @siavashs hello.

Is it okey for us?

madest92:
@TheMeier @siavashs hi guys.

Can you please look at the merge again?

Anything else needed?

notify/mattermost/mattermost.go
```diff
@@ -209,7 +208,40 @@ func (n *Notifier) createRequest(tmpl func(string) string) *request {
 			}
 
 			req.Attachments[idxAtt] = att
+			req.Text = tmpl(n.conf.Text)
```
siavashs: We only need to render this template once and not repeatedly inside a loop.

docs/configuration.md
```diff
@@ -1083,6 +1083,22 @@ text: <tmpl_string> | default = '{{ template "mattermost.default.text" . }}'
 
 # Message attachments used for richer formatting options.
 # It is for compatibility with Slack.
+[ fallback: <tmpl_string> | default = '{{ template "mattermost.default.fallback" . }}' ]
+[ color: <tmpl_string> | default = '{{ template "mattermost.default.color" . }}' ]
+[ title: <tmpl_string> | default = '{{ template "mattermost.default.title" . }}' ]
+[ title_link: <tmpl_string> | default = '{{ template "mattermost.default.titlelink" . }}' ]
+[ text: <tmpl_string> | default = '{{ template "mattermost.default.text" . }}' ]
```
siavashs: The `text` field is documented twice, once from existing docs and once for the new functionality.
We should keep one.

siavashs (approved):

PR #5013, opened 2026-02-16
fix: improve gc in provider and store

siavashs:
Break up GC into stages to release the locks earlier.
This should avoid blocking while alerts are deleted from internal caches.
The callbacks are now called without holding locks in the caller.
Add `PostGC` callback to mem provider.
Modify the marker `Delete` method to accept multiple arguments.
Store does not copy deleted alerts to avoid unnecessary allocations.

#### Pull Request Checklist

    - Fixes #<issue number>
    
- Is this a bugfix?
- Is this a new feature?
- Is this a breaking change?

```release-notes
NONE
```

### DIFF
--- a/provider/mem/mem_test.go
+++ b/provider/mem/mem_test.go
@@ -542,7 +542,7 @@ func (l *limitCountCallback) PostStore(_ *types.Alert, existing bool) {
 	}
 }
 
-func (l *limitCountCallback) PostDelete(alert *types.Alert) {
+func (l *limitCountCallback) PostDelete(_ *types.Alert) {
 	l.alerts.Add(-1)
 }
 


provider/mem/mem_test.go
```diff
@@ -540,8 +540,10 @@ func (l *limitCountCallback) PostStore(_ *types.Alert, existing bool) {
 	}
 }
 
-func (l *limitCountCallback) PostDelete(_ *types.Alert) {
-	l.alerts.Add(-1)
+func (l *limitCountCallback) PostDelete(alerts ...*types.Alert) {
+	for range alerts {
+		l.alerts.Add(-1)
```
ultrotter: If this is all we do why not l.alerts.Add(-len(alerts))

store/store.go
```diff
@@ -92,13 +93,29 @@ func (a *Alerts) Run(ctx context.Context, interval time.Duration) {
...
+// gcAlerts deletes resolved alerts and returns a copy of them.
+func (a *Alerts) gcAlerts() (deleted []*types.Alert) {
 	a.Lock()
-	var resolved []types.Alert
+	defer a.Unlock()
 	for fp, alert := range a.alerts {
 		if alert.Resolved() {
 			delete(a.alerts, fp)
-			resolved = append(resolved, types.Alert{
+			deleted = append(deleted, &types.Alert{
```
ultrotter: Is there a reason why we can't return just pointers to the existing alerts... We have removed them from the store, what are we worried about that we have to copy them? We just need to make sure the callbacks don't modify them... We would save a lot of allocations while holding the lock...) If we can't potentially let's check if it isn't better to just do what we did before: at least there would be fewer allocations, just the slice backing array growth, instead of that, plus one per alert...

siavashs: I think it is safe to do so if we trust that no components are going to mutate the alerts.

siavashs: Implemented in the new fix commits pushed.

provider/mem/mem.go
```diff
@@ -170,17 +170,38 @@ func (a *Alerts) gcLoop(ctx context.Context, interval time.Duration) {
...
+	// held in memory in aggregation groups redundantly.
+	deleted := a.gcAlerts()
+
+	// If there are no deleted alerts, there is nothing to do.
+	if len(deleted) == 0 {
+		return
+	}
+
+	// Delete markers for deleted alerts.
+	var ff model.Fingerprints
```
ultrotter: can you ff := make(...) with size zero and capacity len(deleted)?

provider/mem/mem.go
```diff
@@ -71,8 +71,8 @@ type AlertStoreCallback interface {
 	// PostStore is called after alert has been put into store.
 	PostStore(alert *types.Alert, existing bool)
 
-	// PostDelete is called after alert has been removed from the store due to alert garbage collection.
-	PostDelete(alert *types.Alert)
+	// PostDelete is called after alerts have been removed from the store due to alert garbage collection.
+	PostDelete(alerts ...*types.Alert)
```
ultrotter: nit: if we see this as part of the API we could leave it alone, and add a new one which takes the alerts slice, and internally use the new one... But also no super strong opinion about how strict we want to be

siavashs: I think Cortex and Mimir use it.
But fail enough, I will add a new one which even uses FFs which is what we really need.

provider/mem/mem_test.go
```diff
@@ -537,13 +538,18 @@ func (l *limitCountCallback) PreStore(_ *types.Alert, existing bool) error {
 func (l *limitCountCallback) PostStore(_ *types.Alert, existing bool) {
 	if !existing {
 		l.alerts.Add(1)
+		l.gcCount.Add(1)
 	}
 }
 
-func (l *limitCountCallback) PostDelete(_ *types.Alert) {
+func (l *limitCountCallback) PostDelete(alert *types.Alert) {
 	l.alerts.Add(-1)
```
ultrotter, line 546: Uhm don't you have to remove this line, otherwise you will do Add(-1) for each one as PostDelete, then again subtract them all in PostGC!? Do we have no test catching this? Scary!!

siavashs: GC has its own counter since it uses fingerprints and we need to verify that independently.

siavashs: To clarify, here are the independent atomic counters:
```
	alerts  atomic.Int32
	gcCount atomic.Int32
```

Issue #4963, opened 2026-02-05
Add receiver labels


Spaceman1701:
### Proposal

It would be useful to have structured identification for receivers. Today, it's common for users to use structured names for receivers that encode something about their ownership or behavior. It'd be nice if this was more explicit and easier to query. This would be especially useful for #4962

This proposal is essentially to change receiver config so it contains a `labels` key (and add any useful API features to query this):
```yaml
receivers:
  - name: my-useful-receiver
    labels:
      owner: my-team
      kind: heartbeat
```

I can think of two motivating use cases right now:
1. Conventionally defining the owner of a receiver
2. Conventionally defining the category of the receiver (suggested by @SuperQ )

Spaceman1701:
@cxdy is going to take a look at this as an alternative backend implementation for #5015

Spaceman1701:
Two other details I'm thinking about:

1. I think the best thing to do is add a new query parameter for `receiver_matchers` that takes matchers which apply to receiver labels. We can add this to any API which should be able to filter by receiver (@cxdy I'd say as a start, `/alerts` and `/alerts/groups` should have this parameter, but I know your interest is mostly in `/alerts`, so I think it's totally ok to start with that. A follow up PR can add it in other places).
2. We need to decide what to do with the `name` portion of the receiver. For alerts, the `alertname` is a label, so it might make sense to put the value for `name` into `receivername`. It's not quite ideal to have magic that changes `name` -> `receivername`, so the other alternative is to put `name` into a `name` label. I'm not sure which is best.

cxdy:
Also torn on 2, but I think for simplicity keeping it `name` makes the most sense. 
```json
        "receivers": [
          {
            "name": "slack-alert-testing"
          }
        ],
```
It's already called `name` in `/alerts/groups`, no need to rename it in my opinion.

Spaceman1701:
> Also torn on 2, but I think for simplicity keeping it `name` makes the most sense.
> 
>         "receivers": [
>           {
>             "name": "slack-alert-testing"
>           }
>         ],
> 
> It's already called `name` in `/alerts/groups`, no need to rename it in my opinion.

That seems reasonable to me. I think we should add `name` as a label on the receiver (while not removing `name` from the top-level receiver object) so that it will be possible to match receivers by name.

So we'll end up with something like
```json
        "receivers": [
          {
            "name": "slack-alert-testing",
            "labels": {
                "name": "slack-alert-testing",
[...]

PR #5015, opened 2026-02-16
[FEATURE] hidden receivers in UI/API

cxdy:
## Description

This pull request does the following:
- Adds the ability to "hide" alerts from specific receivers in the UI and API responses
- Adds a button to the UI to show alerts from "hidden" receivers
- Adds a URL parameter `?includeHidden=true` to show "hidden" receivers in both UI and API responses

This is set in the `receivers:` block, like this: 
```yaml
receivers:
- name: 'webhook-receiver'
  hidden: true
  webhook_configs:
  - url: 'http://webhook-receiver:8080/alerts'
```

If unset, it will not be hidden.

### Why?

This may be a niche usecase, but in our environment, we have a webhook receiver (internal tool) that is our top-level route, thus it receives every alert that reaches our Alertmanager cluster. This means that all alerts are essentially duplicated in the UI response when hitting the `/alerts` endpoint. 

```console
root@prometheus1:~# curl -s http://127.0.0.1:9093//api/v2/alerts | jq '[.[] | {receiver: .receivers[].name}] | group_by(.receiver) | map({receiver: .[0].receiver, count: length}) | sort_by(-.count) | .[0:1]'
[
  {
    "receiver": "webhook-receiver",
    "count": 8344
  }
]
```
This makes the UI load a bit slower than we'd like, and other folks use the Alertmanager API for various automations and have been complaining of poor performance due to this (amongst other things). 

### Checklist
- Is this a new feature?

### DIFF
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -68,13 +68,16 @@ type API struct {
 	groupMutedFunc groupMutedFunc
 	uptime         time.Time
 
-	// mtx protects alertmanagerConfig, setAlertStatus and route.
+	// mtx protects alertmanagerConfig, setAlertStatus, route, and hiddenReceivers.
 	mtx sync.RWMutex
 	// resolveTimeout represents the default resolve timeout that an alert is
 	// assigned if no end time is specified.
 	alertmanagerConfig *config.Config
 	route              *dispatch.Route
 	setAlertStatus     setAlertStatusFn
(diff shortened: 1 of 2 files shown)


cxdy:
Closing as I'll be working on https://github.com/prometheus/alertmanager/issues/4963 instead which should bring us closer to what I'm looking for.

PR #5020, opened 2026-02-19
[PERF] Reduce memory allocations through pre-sizing collections and batch allocation

Mwea:
#### Pull Request Checklist                                                                                                                                                                                                            
                                                                                                                                                                                                                                         
      - N/A - General performance improvements
      - N/A
  - Is this a bugfix?
      - N/A
  - Is this a new feature?
      - N/A
  - Is this a breaking change?

#  [PERF] Reduce memory allocations in alert processing hot paths through pre-sizing collections and batch allocation.

---
  Summary

  This PR reduces memory allocations in several hot paths by pre-sizing maps/slices and batching allocations. All changes are internal optimizations with no API changes.

---
  
  Benchmarks

  Batch allocate alert copies in flush:
  10 alerts:  Before: 691 ns/op, 12 allocs  →  After: 435 ns/op, 3 allocs  (37% faster, 75% fewer allocs)
  50 alerts:  Before: 3003 ns/op, 52 allocs →  After: 1714 ns/op, 3 allocs (43% faster, 94% fewer allocs)
  100 alerts: Before: 5976 ns/op, 102 allocs → After: 3383 ns/op, 3 allocs (43% faster, 97% fewer allocs)

  DedupStage alert partitioning:
  50 alerts:  Before: 15.8 µs/op, 126 allocs → After: 13.1 µs/op, 110 allocs (17% faster, 13% fewer allocs)
  100 alerts: Before: 29.0 µs/op, 232 allocs → After: 25.7 µs/op, 210 allocs (11% faster, 9% fewer allocs)

  fingerprintEquals pre-sizing:
  3 labels:  Before: 357 ns/op → After: 314 ns/op (12% faster)
  10 labels: Before: 1519 ns/op → After: 1365 ns/op (10% faster)

  getGroupLabels pre-sizing:
  group_by_all (10 labels): Before: 714 ns/op → After: 585 ns/op (18% faster)

  Template.Data caching:
  50 alerts: 4416 B/op vs 8832 B/op (-50% memory), 51 vs 102 allocs (-50%)

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -776,18 +776,17 @@ func (ag *aggrGroup) flush(notify func(...*types.Alert) bool) {
 		alerts        = ag.alerts.List()
 		alertsSlice   = make(types.AlertSlice, 0, len(alerts))
 		resolvedSlice = make(types.AlertSlice, 0, len(alerts))
-		alertCopies   = make([]types.Alert, len(alerts))
 		now           = time.Now()
 	)
-	for i, alert := range alerts {
-		alertCopies[i] = *alert
+	for _, alert := range alerts {
+		a := *alert
 		// Ensure that alerts don't resolve as time move forwards.
-		if alertCopies[i].ResolvedAt(now) {
-			resolvedSlice = append(resolvedSlice, &alertCopies[i])
+		if a.ResolvedAt(now) {
+			resolvedSlice = append(resolvedSlice, &a)
 		} else {
-			alertCopies[i].EndsAt = time.Time{}
+			a.EndsAt = time.Time{}
 		}
-		alertsSlice = append(alertsSlice, &alertCopies[i])
+		alertsSlice = append(alertsSlice, &a)
 	}
 	sort.Stable(alertsSlice)
 
@@ -812,11 +811,6 @@ func (ag *aggrGroup) flush(notify func(...*types.Alert) bool) {
 			}
 		}
 	}
-	// Enforce the contract that after notification the alerts pointed by alertsSlice
-	// cannot be retained as a pointer (to avoid leaking the whole slice in memory).
-	// A notifier that needs to keep a hold of an alert needs to copy it, or just
-	// keep enough information to recognize it again, in case.
-	clear(alertCopies)
 }
 
 type nilLimits struct{}


Mwea:
I believe the added benchmarks for every single commit are probably overkill, but I've left them voluntarily
  available for the first round of review. Happy to squash them into fewer commits or remove any that seem       
  useless once agreed on the overall approach.                                                           

  Let me know if you'd prefer:                                                                                   
  - A single squashed commit with all the pre-sizing optimizations                                               
  - Removing benchmarks that measuring obvious improvements

Mwea:
> Thank you! There a some formatting issues failing in the CI checks.

Fixed it, relica of the refactor 👍🏼

siavashs:
> > Thank you! There a some formatting issues failing in the CI checks.
> 
> Fixed it, relica of the refactor 👍🏼

There are more, please rebase your branch (we just merged an updated `golangci-lint` 2.10) and run `golangci-lint` locally to make sure all issues are fixed.

Mwea:
> > > Thank you! There a some formatting issues failing in the CI checks.
> > 
> > 
> > Fixed it, relica of the refactor 👍🏼
> 
> There are more, please rebase your branch (we just merged an updated `golangci-lint` 2.10) and run `golangci-lint` locally to make sure all issues are fixed.

My bad ! Should be all good now: 
```
 (git)-[perf/pre-size-maps]-   $   make build && make test
>> building binaries
/home/titouan.chary/go/bin/promu build --prefix /home/titouan.chary/Projects/alertmanager 
 >   alertmanager
 >   amtool
>> running all tests
go test -race  ./...
?       github.com/prometheus/alertmanager/api  [no test files]
?       github.com/prometheus/alertmanager/api/metrics  [no test files]
ok      github.com/prometheus/alertmanager/api/v2       (cached)
?       github.com/prometheus/alertmanager/api/v2/client        [no test files]
?       github.com/prometheus/alertmanager/api/v2/client/alert  [no test files]
?       github.com/prometheus/alertmanager/api/v2/client/alertgroup     [no test files]
?       github.com/prometheus/alertmanager/api/v2/client/general        [no test files]
?       github.com/prometheus/alertmanager/api/v2/client/receiver       [no test files]
?       github.com/prometheus/alertmanager/api/v2/client/silence        [no test files]
?       github.com/prometheus/alertmanager/api/v2/models        [no test files]
?       github.com/prometheus/alertmanager/api/v2/restapi       [no test files]
?       github.com/prometheus/alertmanager/api/v2/restapi/operations    [no test files]
?       github.com/prometheus/alertmanager/api/v2/restapi/operations/alert      [no test files]
?       github.com/prometheus/alertmanager/api/v2/restapi/operations/alertgroup [no test files]
?       github.com/prometheus/alertmanager/api/v2/restapi/operations/general    [no test files]
?       github.com/prometheus/alertmanager/api/v2/restapi/operations/receiver   [no test files]
?       github.com/prometheus/alertmanager/api/v2/restapi/operations/silence    [no test files]
?       github.com/prometheus/alertmanager/asset        [no test files]
ok      github.com/prometheus/alertmanager/cli  (cached)
ok      github.com/prometheus/alertmanager/cli/config   (cached)
?       github.com/prometheus/alertmanager/cli/format   [no test files]
ok      github.com/prometheus/alertmanager/cluster      (cached)
?       github.com/prometheus/alertmanager/cluster/clusterpb    [no test files]
ok      github.com/prometheus/alertmanager/cmd/alertmanager     (cached)
?       github.com/prometheus/alertmanager/cmd/amtool   [no test files]
ok      github.com/prometheus/alertmanager/config       (cached)
ok      github.com/prometheus/alertmanager/config/receiver      (cached)
ok      github.com/prometheus/alertmanager/dispatch     (cached)
?       github.com/prometheus/alertmanager/examples/webhook     [no test files]
ok      github.com/prometheus/alertmanager/featurecontrol       (cached)
ok      github.com/prometheus/alertmanager/inhibit      (cached)
ok      github.com/prometheus/alertmanager/limit        (cached)
ok      github.com/prometheus/alertmanager/matcher/compat       (cached)
ok      github.com/prometheus/alertmanager/matcher/compliance   (cached)
ok      github.com/prometheus/alertmanager/matcher/parse        (cached)
ok      github.com/prometheus/alertmanager/nflog        (cached)
ok      github.com/prometheus/alertmanager/nflog/nflogpb        (cached)
ok      github.com/prometheus/alertmanager/notify       (cached)
ok      github.com/prometheus/alertmanager/notify/discord       (cached)
ok      github.com/prometheus/alertmanager/notify/email (cached)
ok      github.com/prometheus/alertmanager/notify/incidentio    (cached)
ok      github.com/prometheus/alertmanager/notify/jira  (cached)
ok      github.com/prometheus/alertmanager/notify/mattermost    (cached)
ok      github.com/prometheus/alertmanager/notify/msteams       (cached)
ok      github.com/prometheus/alertmanager/notify/msteamsv2     (cached)
ok      github.com/prometheus/alertmanager/notify/opsgenie      (cached)
ok      github.com/prometheus/alertmanager/notify/pagerduty     (cached)
ok      github.com/prometheus/alertmanager/notify/pushover      (cached)
ok      github.com/prometheus/alertmanager/notify/rocketchat    (cached)
ok      github.com/prometheus/alertmanager/notify/slack (cached)
ok      github.com/prometheus/alertmanager/notify/sns   1.174s
ok      github.com/prometheus/alertmanager/notify/telegram      (cached)
?       github.com/prometheus/alertmanager/notify/test  [no test files]
ok      github.com/prometheus/alertmanager/notify/victorops     (cached)
ok      github.com/prometheus/alertmanager/notify/webex (cached)
ok      github.com/prometheus/alertmanager/notify/webhook       (cached)
ok      github.com/prometheus/alertmanager/notify/wechat        (cached)
ok      github.com/prometheus/alertmanager/pkg/labels   (cached)
?       github.com/prometheus/alertmanager/pkg/modtimevfs       [no test files]
?       github.com/prometheus/alertmanager/provider     [no test files]
ok      github.com/prometheus/alertmanager/provider/mem (cached)
ok      github.com/prometheus/alertmanager/silence      (cached)
?       github.com/prometheus/alertmanager/silence/silencepb    [no test files]
ok      github.com/prometheus/alertmanager/store        (cached)
ok      github.com/prometheus/alertmanager/template     (cached)
?       github.com/prometheus/alertmanager/test/cli     [no test files]
ok      github.com/prometheus/alertmanager/test/cli/acceptance  3.270s
?       github.com/prometheus/alertmanager/test/testutils       [no test files]
?       github.com/prometheus/alertmanager/test/with_api_v2     [no test files]
ok      github.com/prometheus/alertmanager/test/with_api_v2/acceptance  27.500s
ok      github.com/prometheus/alertmanager/timeinterval (cached)
ok      github.com/prometheus/alertmanager/tracing      (cached)
ok      github.com/prometheus/alertmanager/types        (cached)
ok      github.com/prometheus/alertmanager/ui   (cached)
  (git)-[perf/pre-size-maps]-   $ make lint                                    
>> running golangci-lint
/home/titouan.chary/go/bin/golangci-lint run  ./...
0 issues.
```

siavashs (approved):
LGTM

api/v2/compat.go
```diff
@@ -195,9 +195,9 @@ func OpenAPIAlertsToAlerts(ctx context.Context, apiAlerts open_api_models.Postab
 	_, span := tracer.Start(ctx, "OpenAPIAlertsToAlerts")
 	defer span.End()
 
-	alerts := []*types.Alert{}
-	for _, apiAlert := range apiAlerts {
-		alert := types.Alert{
+	alerts := make([]*types.Alert, len(apiAlerts))
```
Spaceman1701: nit: it's a little more idiomatic to allocate with pre-set capacity but 0 length. 
```suggestion
	alerts := make([]*types.Alert, 0, len(apiAlerts))
```
(you'd also need to change from assigning to `alerts[i]` to `append(alerts, &alert)` again).

This _should_ result in the same number of allocations, but it makes it more clear to the reader than we're allocating a new slice and copying data into it.

dispatch/dispatch.go
```diff
@@ -766,28 +770,32 @@ func (ag *aggrGroup) destroyed() bool {
...
 }
 
+// prepareAlertsForFlush copies alerts and partitions them into all/resolved slices.
+func prepareAlertsForFlush(alerts []*types.Alert, now time.Time) (types.AlertSlice, types.AlertSlice) {
+	alertsSlice := make(types.AlertSlice, 0, len(alerts))
+	resolvedSlice := make(types.AlertSlice, 0, len(alerts))
+	alertCopies := make([]types.Alert, len(alerts))
+
+	for i, alert := range alerts {
+		alertCopies[i] = *alert
```
Spaceman1701: Is the idea here to batch the allocations of the copied alerts? I guess otherwise line 786 will result in one allocation per alert... 

I'm a touch worried that this will leak memory - if anybody keeps a reference to any alert that's part of this slice we cannot deallocate the entire slice.

Have you observed a performance issue with the existing logic that justifies the additional complexity?

siavashs: I might be mistaken but it seems the old code did the same here https://github.com/prometheus/alertmanager/pull/5020/changes#diff-2ec4e2e390946f66324eb00bb9e735280a4d018e65de41e64bebac8bebebd23bL782

api/v2/compat.go
Mwea: Nice catch! updated accordingly

dispatch/dispatch.go
Mwea: Thanks for the review and for raising this concern!                                                                                                                                                             
   
I've updated the commit to inline the change directly in `flush()` rather than extracting a separate function, hopefully this makes the diff easier to follow. The initial extraction was mainly to allow  dedicated unit tests for the alert preparation logic, but it made the actual optimization harder to see.

Regarding the memory concernm I think you're right that if something held a reference to one alert, it would retain the whole array. Looking at the `flush()` flow though, `alertsSlice` is passed to `notify()`, processed, and then released when `flush()` returns. However, if there is a code path where individual alerts are retained longer, that would have been a concern with the original code as well, isn't it ? Does the introduction of `alertCopies` changes that risk ?

Happy to revert if you feel the complexity isn't worth it. About your initial question on the performances, we've been seeing quite a bit of memory pressure on the alertmanager from many small issues, so any low-hanging fruit that reduces allocations feels worth picking up 😄

ultrotter: Well in the original copy the type of AlertSlice was pointers to alert. So holding onto an individual alert would keep that one alert in memory, not the whole slice. Only keeping a reference to the whole slice would keep the whole slice. In the new version the alerts are all part of one slice, so having a pointer to one alert unfortunately holds the whole slice...

If we see it's a problem to allocate and GC all these alerts, would we consider sync.Pool? Though not sure if the sync aspect would slow us down then. Unfortunately it might be hard to guarantee that no code will ever hold onto an alert/store it in a map "for a while" and such ....

ultrotter: @Spaceman1701 what would you think if we zeroed out alertCopies after use? This way any pointers to it would clearly be useless, and any user that mistakenly holds onto them would see the bug and realize they need copy+keep instead... So this would make it safe through documenting the contract (don't hold on to these alert pointers) + making any reference kept useless by clearing the slice.

Spaceman1701: > [@Spaceman1701](https://github.com/Spaceman1701) what would you think if we zeroed out alertCopies after use, or if we put it as part of a struct that we keep in a pool, and reset it before putting it in the pool? This way any pointers to it would break, and we wouldn't be able to hold on to them, so a user needing to keep a cache would need to do copy+cache... ?

I think my preferred solution would be to encapsulate all this logic into a struct that owns the memory, but I suppose in practice it's a minor difference. I also do feel a little more comfortable with the in-lined function since it makes things harder to misuse.

If I'm the only one with this concern, I'm also happy to be outvoted - It's not a strong concern.

ultrotter: Let's go for slice clear for now... It's a little unclear to me how "a struct that owns the memory" would help if the notifiers will still every get access to *types.Alert without a copy each time they do, which would defeat the purpose and duplicate the copies? But maybe we can do that as a future improvement and/or we can chat about it

Mwea:
Revert the batch allocation optimization (b6f01d7f) and the subsequent clear(alertCopies) addition (ab5b200b).                               

The batch allocation aimed to reduce per-alert heap allocations by sharing a backing array. However, this introduced a memory retention concern: holding one alert reference would retain the entire array.
                                                                      
The clear() was added to "enforce the contract" that notifiers should   not retain alert pointers. However, this approach breaks both tests   and any real code that legitimately accesses alerts during or after   the notify callback - it zeroes the alert data, causing fingerprint   calculations to return invalid values.

The complexity and test failures outweigh the allocation savings.   Reverting to per-alert allocations which are simpler and have clear   ownership semantics.

Spaceman1701:
Thanks for the contribution!

PR #5021, opened 2026-02-20
[PERF] Inhibitor: Add inverted index for O(k) rule lookup instead of O(N) linear scan.

Mwea:
#### Pull Request Checklist                                                                                    
                                                                                                                 
      - N/A                                                                                                      
      - N/A
  - Is this a bugfix?
      - N/A
  - Is this a new feature?
      - [X] I have added tests that test the new feature's functionality
  - Is this a breaking change?
  - [x] I will follow [best practices for contributing to this
  project](https://docs.github.com/en/get-started/exploring-projects-on-github/contributing-to-open-source)

#  [PERF] Inhibitor: Add inverted index for O(k) rule lookup instead of O(N) linear scan.

  This PR aAdd an inverted index for inhibit rule target matcher lookup to achieve O(k) rule selection instead of O(N) linear scan, where k = number of labels on the alert and N = number of inhibit rules.               
                                                                                                                                                                                                                  
  - Add benchmarks to measure scaling behavior with different rule distributions
  - Refactor `Mutes()` to extract core checking logic into `checkInhibit()`
  - Implement `ruleIndex` with configurable thresholds for index construction

  ## Motivation

  When alertmanager has many inhibit rules (e.g., hundreds or thousands), the current implementation checks every rule for every alert, resulting in O(N) complexity. In environments with rules targeting different label values (e.g., per-cluster or per-service rules), most of this work is wasted.
  
  ## Benchmark Results                                                                                                                                                                                            
 ### Summary
  | Scenario | Before | After | Improvement | Complexity (before) | Complexity (after) |
  |----------|--------|-------|-------------|---------------------|-------------------|
  | different_targets | 28.2µs | 1.6µs | 18x faster | O(N) | O(k) |
  | no_match | 11.5µs | 0.8µs | 15x faster | O(N) | O(k) |
  | same_target | 143µs | 136µs | ~same | O(N) | O(N) fallback |

  *N = number of rules, k = number of labels on alert*

### Details 

  | Benchmark | Baseline (ns/op) | HEAD (ns/op) | Delta |                                                                                                                                                         
  |-----------|------------------|--------------|-------|                          
  | **BenchmarkMutesScaling/different_targets/rules=10** | 1708 | 1524 | **-10.8%** |
  | **BenchmarkMutesScaling/different_targets/rules=100** | 4023 | 1523 | **-62.1%** |
  | **BenchmarkMutesScaling/different_targets/rules=1000** | 28231 | 1565 | **-94.5%** |
  | BenchmarkMutesScaling/same_target/rules=10 | 2439 | 2396 | -1.8% |
  | BenchmarkMutesScaling/same_target/rules=100 | 14970 | 14380 | -3.9% |
  | BenchmarkMutesScaling/same_target/rules=1000 | 143224 | 135880 | -5.1% |
  | BenchmarkMutesScaling/no_match/rules=10 | 760 | 744 | -2.1% |
  | **BenchmarkMutesScaling/no_match/rules=100** | 1756 | 739 | **-57.9%** |
  | **BenchmarkMutesScaling/no_match/rules=1000** | 11492 | 778 | **-93.2%** |



  When the index is effective (O(1) lookup):
  - Rules have different equality target matchers (e.g., cluster=X)
  - Alert labels allow direct lookup into the index

  When the index falls back to O(N) scan:
  - All rules share the same target matcher (high overlap)
  - Rules use regex matchers only

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **Performance**
  * Faster and more scalable inhibition checks via a new indexing approach that reduces candidate scanning and short-circuits when matches are found, improving throughput for large rule sets.

* **Tests**
  * Added extensive benchmarks covering scaling, overlap ratios, and index thresholds.
  * Added unit tests for indexed candidate selection, deduplication, and mixed-index/linear scenarios.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


inhibit/inhibit_bench_test.go
```diff
@@ -228,3 +228,296 @@ func mustNewMatcher(b *testing.B, op labels.MatchType, name, value string) *labe
...
+				"cluster":   model.LabelValue(lastCluster),
+			},
+		},
+	}
+	require.NoError(b, s.Put(context.Background(), &alert))
+
+	ih := NewInhibitor(s, rules, m, promslog.NewNopLogger())
+	defer ih.Stop()
+	go ih.Run()
+	<-time.After(time.Second)
```
ultrotter: Should we do   ih.WaitForLoading() here instead of effectively a sleep?

inhibit/rule_index.go
```diff
@@ -0,0 +1,204 @@
...
+)
+
+// matcherKey is used as a map key to avoid string concatenation allocations.
+type matcherKey struct {
+	name  string
+	value string
+}
+
+// RuleIndexOptions configures the rule index behavior.
+type RuleIndexOptions struct {
```
ultrotter: For this and the one below DefaultRuleIndexOptions, should we consider private as well? Or do we have a reason to export?

inhibit/rule_index_test.go
```diff
@@ -0,0 +1,397 @@
...
+		return false
+	})
+
+	require.False(t, result)
+	require.Len(t, visited, 1)
+	require.Equal(t, "rule-staging", visited[0].Name)
+}
+
+func TestForEachCandidate_EarlyTermination(t *testing.T) {
+	rules := []*InhibitRule{
```
ultrotter, line 226: Should this test have more than 2 rules to also test the early termination via index? Or is it not necessary?

ultrotter (approved):
Approved, with some minor comments/nits

Mwea:
> Approved, with some minor comments/nits

Thanks ! I will tackle this tomorrow. 
I had some thoughts on proposing a similar solution for the Silencer , but I don't really know it would bring that much benefits as for this one. Any thoughts ? @ultrotter

ultrotter:
> > Approved, with some minor comments/nits
> 
> Thanks ! I will tackle this tomorrow. I had some thoughts on proposing a similar solution for the Silencer , but I don't really know it would bring that much benefits as for this one. Any thoughts ? @ultrotter

We already have some caches in the silencer to improve performace, plus you have the situation that the rules would not be static based on the config, right? This would mean extra locking to make it work which may or may not create an issue? Let's focus on the current patches, then we'll see what we can do if we see that you have scalability issues there! :)

inhibit/inhibit_bench_test.go
Mwea: Done, replaced all occurrences with `WaitForLoading()`.

inhibit/rule_index.go
Mwea: Good point, made both unexported since they're only used internally.

inhibit/rule_index_test.go
Mwea: Added more rules (4 total) to ensure we're testing index-based iteration rather than linear scan at the threshold boundary.

TheMeier (approved):
This possibly causes one change in behaviour, which I consider ok.
Previously Mutes walked ih.rules in config order, so first-match selection was deterministic. Now it is _possible_ that one sees a different inhibition in Alert.InhibitedBy. (Even though it is a slice only one seems to get added)

> while InhibitedBy may contain only a subset of the inhibiting alerts
> // – in practice exactly one ID. (This somewhat confusing semantics might change
> // in the future.)

Mwea:
> This possibly causes one change in behaviour, which I consider ok. Previously Mutes walked ih.rules in config order, so first-match selection was deterministic. Now it is _possible_ that one sees a different inhibition in Alert.InhibitedBy. (Even though it is a slice only one seems to get added)
> 
> > while InhibitedBy may contain only a subset of the inhibiting alerts
> > // – in practice exactly one ID. (This somewhat confusing semantics might change
> > // in the future.)

@TheMeier Would like me to tackle it by documenting the change or do you feel it is good enough to get it merged ?

TheMeier:
@Mwea I personally do not care. But maybe someone else does, who knows?

inhibit/rule_index.go
```diff
@@ -0,0 +1,204 @@
...
+	for labelName, labelValue := range lset {
+		valueMap, ok := idx.exactIndex[string(labelName)]
+		if !ok {
+			continue
+		}
+
+		rules, ok := valueMap[string(labelValue)]
+		if !ok {
+			continue
+		}
```
siavashs, line 188: I ran an analysis and it seems if someone uses a marcher with an empty value this logic changes behaviour: https://app.devin.ai/review/prometheus/alertmanager/pull/5021

I don't know how common is such a configuration though.

Mwea: Good catch! You're right , this was indeed a correctness regression.                                                                               
                                                            
Fixed by excluding empty-value matchers from indexing, they now fall back to linearRules where they're always checked. Added a regression test as well

SoloJacobs:
Hi @Mwea ,

this PR was discussed in the Alertmanager Working Group today. Sadly, I we have decided not to take this PR at this point in time. The reason are as follows:

* The inhibitor is typically no longer a bottle neck for deployments. At least in the workloads we have, slow performance is usually due to the dispatcher. Thus, performance improvements should target that area. If you have contrary data, I would be very interested, if you could share it.
* We just found a regression in the inhibitor caused by the existing performance improvements, so we are a bit more wary of improvements of this kind now.
* We generally think that most of the performance issues are more structural in nature. Specifically, the `provider` implementation, and the architecture it forces downstream, should be reconsidered. Caching and indexing can only do so much.

If we you want to update this PR and providing the specific data, then I'm open to taking it to maintainers again. But I don't want you to invest more time into rebasing, if the actual work required, is conceptual.

Thank you, and have a nice weekend!

PR #4983, opened 2026-02-09
Fixing broken Yaml+ duplicate heading mentioned in Issue#4982

kakabisht:
This PR aims to fix Issue about broken YAML + duplication heading mentioned in https://github.com/prometheus/alertmanager/issues/4982

PR #5034, opened 2026-02-22
fix (docs): Remove duplicate header

kakabisht:
Follow up on https://github.com/prometheus/alertmanager/pull/4983

### DIFF
--- a/docs/high_availability.md
+++ b/docs/high_availability.md
@@ -4,8 +4,6 @@ sort_rank: 4
 nav_icon: network
 ---
 
-# High Availability
-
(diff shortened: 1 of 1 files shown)


SoloJacobs:
Thanks!

kakabisht:
Many thanks @SoloJacobs 
Also apologies for delay

kakabisht:
Hi @SoloJacobs ,

This issue still exists, even after the merge.

SoloJacobs:
It will only go away once we do the next release. However, this problem is not a regression that was introduced in the current alertmanager release, so we won't do a patch release just for this issue.

PR #5035, opened 2026-02-22
chore (internal): Use go tools in `internal/tools`

SoloJacobs:
See commit messages for detailed explanation. Stacks on top of https://github.com/prometheus/alertmanager/pull/5033

#### Pull Request Checklist

```release-notes
NONE
```


internal/tools/go.mod
```diff
@@ -2,11 +2,10 @@ module github.com/prometheus/prometheus/internal/tools
 
 go 1.25.0
```
siavashs, line 3: CI is using 1.26 above but this is not bumped.

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/prometheus/alertmanager
 
-go 1.24.0
+go 1.25.0
```
siavashs, line 3: Same here.

internal/tools/go.mod
siavashs: I see you have opened another PR to bump go version.

SoloJacobs:
@siavashs Should we add a second reviewer here?
Regarding the `go.mod`, I was not sure whether one should be able to use `go 1.24` here, but I don't have any objections against bumping it.

PR #5041, opened 2026-02-23
chore: drop multi error

siavashs:
Drop `MultiError` in favor of `errors.Join()` from Go 1.20+

#### Pull Request Checklist

```release-notes
[CHANGE] Breaking: remove types.MultiError
```

### DIFF
--- a/config/receiver/receiver.go
+++ b/config/receiver/receiver.go
@@ -14,6 +14,7 @@
 package receiver
 
 import (
+	"errors"
 	"log/slog"
 
 	commoncfg "github.com/prometheus/common/config"
@@ -40,7 +41,6 @@ import (
 	"github.com/prometheus/alertmanager/notify/webhook"
 	"github.com/prometheus/alertmanager/notify/wechat"
 	"github.com/prometheus/alertmanager/template"
-	"github.com/prometheus/alertmanager/types"
 )
 
 // BuildReceiverIntegrations builds a list of integration notifiers off of a
@@ -51,12 +51,12 @@ func BuildReceiverIntegrations(nc config.Receiver, tmpl *template.Template, logg
 	}
 
 	var (
-		errs         types.MultiError
+		errs         error
 		integrations []notify.Integration
 		add          = func(name string, i int, rs notify.ResolvedSender, f func(l *slog.Logger) (notify.Notifier, error)) {
 			n, err := f(logger.With("integration", name))
 			if err != nil {
-				errs.Add(err)
+				errs = errors.Join(errs, err)
 				return
 			}
 			integrations = append(integrations, notify.NewIntegration(n, rs, name, i, nc.Name))
@@ -118,8 +118,5 @@ func BuildReceiverIntegrations(nc config.Receiver, tmpl *template.Template, logg
 		add("mattermost", i, c, func(l *slog.Logger) (notify.Notifier, error) { return mattermost.New(c, tmpl, l, httpOpts...) })
 	}
 
-	if errs.Len() > 0 {
-		return nil, &errs
-	}
-	return integrations, nil
+	return integrations, errs
 }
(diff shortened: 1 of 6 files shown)


notify/notify.go
```diff
@@ -536,28 +536,25 @@ func (ms MultiStage) Exec(ctx context.Context, l *slog.Logger, alerts ...*types.
...
+		wg   sync.WaitGroup
+		errs error
 	)
 	wg.Add(len(fs))
 
 	for _, s := range fs {
 		go func(s Stage) {
 			if _, _, err := s.Exec(ctx, l, alerts...); err != nil {
-				me.Add(err)
+				errs = errors.Join(errs, err)
```
SoloJacobs, line 552: I don't think line is thread safe.

siavashs: Good catch! I fixed this one and double checked if there is any other place we might need a lock.

config/receiver/receiver.go
```diff
@@ -118,5 +118,8 @@ func BuildReceiverIntegrations(nc config.Receiver, tmpl *template.Template, logg
 		add("mattermost", i, c, func(l *slog.Logger) (notify.Notifier, error) { return mattermost.New(c, tmpl, l, httpOpts...) })
 	}
 
-	return integrations, errs
+	if errs != nil {
```
SoloJacobs: I think the previous version was better. `integrations` only contains those without errors:
```go
if err != nil {
	errs = errors.Join(errs, err)
	return
}
integrations = append(integrations, notify.NewIntegration(n, rs, name, i, nc.Name))
```

types/types.go
```diff
@@ -301,49 +300,6 @@ func (m *MemMarker) Silenced(alert model.Fingerprint) (activeIDs []string, silen
 		s.State == AlertStateSuppressed && len(s.SilencedBy) > 0
 }
 
-// MultiError contains multiple errors and implements the error interface. Its
-// zero value is ready to use. All its methods are goroutine safe.
-type MultiError struct {
```
SoloJacobs, line 307: Should we also deprecate these? (Yes, that is annoying).

siavashs: If someone is importing `MultiError` from Alertmanager they might be doing something very wrong IMHO.
See my comment about using these packages externally here https://github.com/prometheus/alertmanager/pull/5043#issuecomment-3945905701

I actually found a way to search this and cortex seems to be using it but mimir not, they use `github.com/grafana/dskit/multierror` for example.

https://github.com/search?q=%22types.MultiError%22+%22github.com%2Fprometheus%2Falertmanager%2Ftypes%22&type=code

config/receiver/receiver.go
siavashs: My initial implementation was changing the behaviour of this function, that's why I changed it back.
Either way is fine with me, it is up to the caller to decide what to do.
In this case the caller ignores the integrations if there are any errors, so returning them is not that useful anyway: https://github.com/prometheus/alertmanager/blob/5db677cdc6b6eb8a9cb6991430ec998b1ed92d69/cmd/alertmanager/main.go#L455-L457

SoloJacobs: Ah, I didn't realize that. That's fair enough, I'm also ok with both versions. In my opinion, should still with the version, which returns partial integrations, since we have four eyes on it now. But ok to ignore

types/types.go
SoloJacobs: Yes, but it can be nice to not get punished for being wrong :wink: Anyway, I think we deprecated `pkg` and it is only a convention. We also handled `types/types.go` differently in the other PR, even though it is not under `pkg`. Arguably, this migration is harder then going from types to alerts.

To be fair, in the past, there are some cases, where we simply changed the function signatures in incompatible ways without any announcement. So, maybe we should just remove the functions without a grace period? Why introduce deprecations in the first place, if we don't won't to communicate changes in a consistent manor?

Feels like something we should raise in # promeheus-alertmanger channel.

SoloJacobs: Ah, I just saw your message on slack lol.

config/receiver/receiver.go
siavashs: I updated it to return the integrations.

types/types.go
siavashs: From the discussion on Slack: since Alertmanager has not reached v1.0 we can break the Go API as long as we specify the breaking changes in the CHANGELOG.

PR #5047, opened 2026-02-25
feat(dispatch): add alert markers per group

### LINKED ISSUE #4953: Migrated from Global Marker to per Aggregation Group markers
Markers should have their own dedicated marker instead of a shared global one.
We should ensure that this change does not affect behaviour or API responses.

Also see #4727

siavashs:
This change adds alert markers to the aggregation groups in dispatcher.
Alert markers replace the global marker and are used to track the state of alerts in each aggregation group.

This change touches many components of the alertmanager.
Per Group alert markers are passed to the notifiers and then inhibitor and silencer using context.

The API has no breaking changes:
- /alerts uses a temporary marker to track the state of alerts
- /alerts/groups returns the status from group markers

The metrics are also updated to use group markers.
The `alertmanager_alerts` metric is moved to dispatcher.
The `alertmanager_marked_alerts` metric is removed.
By default it behaves the same as before, by aggregating all alerts in the groups.
Enabling `group-key-in-metrics` flag will cause the metrics to be grouped by `group_key`.

#### Pull Request Checklist

    - Fixes #4953
    
- Is this a new feature?
- Is this a breaking change?

```release-notes
[FEATURE] Introduce per aggregation group AlertMarkers and drop Global Alert Marker
[CHANGE] Add `group-key-in-metrics` feature flag
[CHANGE] Remove `alertmanager_marked_alerts`
[CHANGE] Remove the following from `types` package: `MemMarker`, `AlertState*`, `AlertStatus`
[CHANGE] Move `AlertMarker`, `GroupMarker` to `marker` package
```

## Summary by CodeRabbit

* **New Features**
  * Feature flag to include group-key labeling in dispatcher metrics.
  * New dispatcher metrics and per-state alert counts for improved observability.
  * New runtime markers for per-alert and per-group state tracking, with context propagation.

* **Bug Fixes**
  * Alert status now computed at query/time-of-use for more consistent active/suppressed/unprocessed results.
  * Improved silence and inhibition handling so mute/inhibit state is recorded and observed reliably.

* **Refactor**
  * Configuration simplified: alert-status callback is no longer required.

[Review Change Stack](https://app.coderabbit.ai/change-stack/prometheus/alertmanager/pull/5047?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


MichaHoffmann (approved):

dispatch/dispatch.go
```diff
@@ -408,6 +377,10 @@ func (d *Dispatcher) Groups(ctx context.Context, routeFilter func(*Route) bool,
 				continue
 			}
 			alertGroup.Alerts = filteredAlerts
+			alertGroup.AlertStatuses = make(map[model.Fingerprint]alert.AlertStatus, len(filteredAlerts))
+			for _, a := range filteredAlerts {
+				alertGroup.AlertStatuses[a.Fingerprint()] = ag.marker.Status(a.Fingerprint())
+			}
```
siavashs, line 400: See this analysis https://app.devin.ai/review/prometheus/alertmanager/pull/5047
Basically `Groups()` now uses the per-group alert marker to report status to `/alerts/groups` api call.
But `/alerts` uses a `tempMarker` to replicate a global marker for backwards compatibility.
This results in the 2 API endpoint to not return exactly the same results:

- `/alerts` will keep "predicting" the future (like it did before), to say if an alert will be muted if dispatched now (we want to remove statuses from here in API v3 probably)
- `/alerts/groups` shows the current true status without using "prediction", this is different than before I think

We can use a `tempMarker` for `/alerts/groups` to keep predicting the future.

TheMeier:
Reviewed the new marker package standalone. LGTM

marker/util.go
```diff
@@ -0,0 +1,56 @@
...
+func WithAlertMarker(ctx context.Context, m AlertMarker) context.Context {
+	return context.WithValue(ctx, alertMarkerKey{}, m)
+}
+
+// GetAlertMarker extracts the AlertMarker from the context if present,
+// otherwise returns a no-op marker.
+func GetAlertMarker(ctx context.Context) AlertMarker {
+	m, ok := ctx.Value(alertMarkerKey{}).(AlertMarker)
+	if !ok {
+		return noop
```
ultrotter: While there are cases where a noop marker is ok, there are some situations where we really want an alertmarker to exist, and it would be an error to act on noop. Maybe we should have a function that returns AlertMarker, ok and in case !ok for those functions log an error/increase a metric/do something about it... Alternatively we can have a MustGetAlertMarker that panics if not, or some other option, depending on the situation and code path

marker/group.go
```diff
@@ -0,0 +1,66 @@
...
+	groups map[string]*groupStatus
+
+	mtx sync.RWMutex
+}
+
+// Muted implements GroupMarker.
+func (m *groupMarker) Muted(routeID, groupKey string) ([]string, bool) {
+	m.mtx.RLock()
+	defer m.mtx.RUnlock()
+	status, ok := m.groups[routeID+groupKey]
```
ultrotter: I know this happened also before but do we have any risks of collision here in routeID + groupKey? And should we consider an invalid character to separate them or another way?

marker/alert.go
```diff
@@ -0,0 +1,93 @@
...
+	s, found := m.status[fp]
+	if !found {
+		s = &alertStatus{}
+		m.status[fp] = s
+	}
+	s.SilencedBy = silencedBy
+}
+
+// SetInhibited implements AlertMarker.
+func (m *alertMarker) SetInhibited(fp model.Fingerprint, inhibitedBy []string) {
```
ultrotter, line 52: Just out of curiosity the previous SetIhibited used
func (m *MemMarker) SetInhibited(alert model.Fingerprint, ids ...string) {
And we changed it to []string ... Is there a reason we changed?

types/types.go
```diff
@@ -30,270 +25,3 @@ type AlertSlice = alert.AlertSlice
...
-	status.mutedBy = timeIntervalNames
-}
-
-func (m *MemMarker) DeleteByGroupKey(routeID, groupKey string) {
-	m.mtx.Lock()
-	defer m.mtx.Unlock()
-	delete(m.groups, routeID+groupKey)
-}
-
-func (m *MemMarker) registerMetrics(r prometheus.Registerer) {
```
ultrotter, line 161: We should note in the release note that we removed this metric, as it would be too expensive to calculate it across all markers... ? Unless each time we change it a marker we were to add/remove the relevant number from the metric... But yeah, if not worth it/useful definitely we should say it's gone

marker/marker.go
```diff
@@ -0,0 +1,61 @@
...
+	Status(fingerprint model.Fingerprint) alert.AlertStatus
+
+	// Delete removes markers for the given fingerprints.
+	Delete(fingerprints ...model.Fingerprint)
+}
+
+// GroupMarker helps to mark groups as active or muted.
+// All methods are goroutine-safe.
+//
+// TODO(grobinson): routeID is used in Muted and SetMuted because groupKey
```
ultrotter, line 45: Maybe no point carrying over TODOs for @grobinson-grafana ?

marker/alert.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+// NewAlertMarker returns a new AlertMarker backed by an in-memory map.
+func NewAlertMarker() AlertMarker {
+	return &alertMarker{
+		status: map[model.Fingerprint]*alertStatus{},
+	}
+}
+
+// AlertMarker is an in-memory implementation of AlertMarker.
```
ultrotter: nit: alertMarker is....

dispatch/metric.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+	}
+	return active, suppressed, unprocessed
+}
+
+// emit sends three gauge metrics (one per alert state) with the given counts and optional extra label values.
+func (c *alertStateCollector) emit(ch chan<- prometheus.Metric, active, suppressed, unprocessed int, extraLabels ...string) {
+	ch <- prometheus.MustNewConstMetric(c.desc, prometheus.GaugeValue, float64(active), append([]string{string(alert.AlertStateActive)}, extraLabels...)...)
+	ch <- prometheus.MustNewConstMetric(c.desc, prometheus.GaugeValue, float64(suppressed), append([]string{string(alert.AlertStateSuppressed)}, extraLabels...)...)
+	ch <- prometheus.MustNewConstMetric(c.desc, prometheus.GaugeValue, float64(unprocessed), append([]string{string(alert.AlertStateUnprocessed)}, extraLabels...)...)
```
ultrotter: Do we care about all these allocations in each scrape? Also you seem to call this with either 0 labels or one label, is it worth having the variading extraLabels at all?

If it's worth it you could allocate once and swap the first label.... 
If not you could consider one version which takes just the one label and one which takes 0, or a fixed array of 2 labels, or some other solutions?

marker/group.go
ultrotter: Ok, looking at the details it seems this is very theoretical, so maybe not worth addressing... Though I still don't love concatenated strings as map values it is unlikely to be a practical issue

notify/mute.go
ultrotter, line 78: Are this TODO and the one above still relevant?

marker/group.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// SetMuted implements GroupMarker.
+func (m *groupMarker) SetMuted(routeID, groupKey string, timeIntervalNames []string) {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+	status, ok := m.groups[routeID+groupKey]
+	if !ok {
+		status = &groupStatus{}
+		m.groups[routeID+groupKey] = status
```
ultrotter: Here we calculate routeID+groupKey twice... I think we could instead do:

type groupMarkerKey struct {
	routeID  string
	groupKey string
}

then in e.g. SetMuted:

key := groupMapKey{routeID: routeID, groupKey: groupKey}
m.mtx.Lock()
defer m.mtx.Unlock()
status, ok := m.groups[key]

--

This also removed any allocation since groupMapKey can be on the stack and groupMapKey cointains pointers to existing strings without the need to copy for concatenation

marker/util.go
siavashs: We can add the extra method for forcing a marker and failing, but so far I don't think we have a use-case for it.

marker/alert.go
siavashs: Just for consistency in the interface API for both `SetInhibited` and `SetSilenced`.

marker/marker.go
siavashs: I think this is still a valid TODO item, we did not change the group marker just moved it to the new package.

dispatch/metric.go
siavashs: Good point, I will improve the logic.

notify/mute.go
siavashs: I think he wanted to add/change the metrics here.
I am not sure if we need it.
I can remove these but my changes don't touch these.

types/types.go
siavashs: I put it in the PR description:
```
[CHANGE] Remove `alertmanager_marked_alerts`
```
Do we need to also update any MD files? cc @SoloJacobs

siavashs:
@coderabbitai review

siavashs:
@coderabbitai review

siavashs:
@coderabbitai review

dispatch/dispatch.go
Spaceman1701: I think we should continue to predict the future with this change. I suspect that there are clients that depend on this behavior from `/alerts/groups`

Perhaps in a future change we can add a feature flag to control the compatibility behavior.

Though there's also an argument that predicting the future is useful, maybe it should be an API query param or something

dispatch/metric.go
```diff
@@ -0,0 +1,177 @@
...
+	var active, suppressed, unprocessed int
+	for _, state := range seen {
+		switch state {
+		case alert.AlertStateActive:
+			active++
+		case alert.AlertStateSuppressed:
+			suppressed++
+		default:
+			unprocessed++
+		}
```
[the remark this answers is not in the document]
Spaceman1701: I believe the robot is right

marker/util.go
```diff
@@ -0,0 +1,56 @@
...
+	"context"
+
+	"github.com/prometheus/common/model"
+
+	"github.com/prometheus/alertmanager/alert"
+)
+
+// alertMarkerKey is an unexported type used as context key for
+// AlertMarker to avoid collisions with keys defined in other packages.
+type alertMarkerKey struct{}
```
Spaceman1701: This does work, but I think it's more idiomatic to type def an `int` so that it's possible to define more constant keys if needed. Similar to the notify package.

marker/util.go
Spaceman1701: It might be better to just defer this decision to the caller with the comma-ok pattern (similar to the notify package). This way callers can decide if the missing marker is a problem. That also makes it clear that a missing marker is possible

Spaceman1701 (commented):
Sorry for the long delay before a review. Overall this looks good to me, just a few minor comments. 

Nothing major enough to block this change if someone else approves before I get a chance to re-review!

marker/util.go
siavashs: Good point, adopted that.

dispatch/dispatch.go
siavashs: Push a new commit which use a temp marker to perform predictions.

marker/util.go
siavashs: Pushed a new commit which drops the noop marker in favor of comma-ok.

siavashs:
@coderabbitai review

siavashs:
@coderabbitai review

dispatch/metric.go
```diff
@@ -0,0 +1,192 @@
...
+func newNoopDispatcherMetrics() *DispatcherMetrics {
+	return &DispatcherMetrics{
+		aggrGroups:            prometheus.NewGauge(prometheus.GaugeOpts{}),
+		processingDuration:    prometheus.NewSummary(prometheus.SummaryOpts{}),
+		aggrGroupLimitReached: prometheus.NewCounter(prometheus.CounterOpts{}),
+	}
+}
+
+// NewDispatcherMetrics returns a new registered DispatchMetrics.
+func NewDispatcherMetrics(registerLimitMetrics bool, r prometheus.Registerer, ff featurecontrol.Flagger) *DispatcherMetrics {
```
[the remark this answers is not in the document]
siavashs: This is used by other projects when they import Alertmanager.

siavashs:
@coderabbitai review

dispatch/metric.go
ultrotter, line 47: If it is a deprecated parameter can you rename it to _, and document this in the docstring please?

// NewDispatcherMetrics returns a new registered DispatchMetrics.
// The first parameter is deprecated, but kept for compatibility with external callers 
func NewDispatcherMetrics(_ bool, r prometheus.Registerer, ff featurecontrol.Flagger) *DispatcherMetrics {

dispatch/metric.go
```diff
@@ -0,0 +1,192 @@
...
+				ag := el.(*aggrGroup)
+				active, suppressed, unprocessed := ag.countAlertsByState()
+				labelValues[1] = ag.GroupKey()
+				labelValues[0] = string(alert.AlertStateActive)
+				c.emit(ch, float64(active), labelValues...)
+				labelValues[0] = string(alert.AlertStateSuppressed)
+				c.emit(ch, float64(suppressed), labelValues...)
+				labelValues[0] = string(alert.AlertStateUnprocessed)
+				c.emit(ch, float64(unprocessed), labelValues...)
+				return true
```
ultrotter, line 129: Here do we want to exclude groups that have 0 in each state? or not?

api/v2/api.go
```diff
@@ -470,22 +476,36 @@ func (api *API) getAlertGroupsHandler(params alertgroup_ops.GetAlertGroupsParams
...
+		// Short-circuit on label matchers before the expensive
+		// silencer/inhibitor pipeline.
+		if !alertMatchesFilterLabels(&a.Alert, matchers) {
+			return false
+		}
+
 		// Set alert's current status based on its label set.
-		api.setAlertStatus(ctx, a.Labels)
+		// The inhibitor and silencer write to m via the context.
+		ctx = marker.WithContext(ctx, m)
```
ultrotter, line 502: Do you want to create this context outside the function itself? This way we would avoid creating a new context each time the returned func is called

marker/group.go
```diff
@@ -0,0 +1,79 @@
...
+func (m *groupMarker) SetMuted(routeID, groupKey string, timeIntervalNames []string) {
+	key := newGroupMarkerKey(routeID, groupKey)
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+	status, ok := m.groups[key]
+	if !ok {
+		status = &groupStatus{}
+		m.groups[key] = status
+	}
+	status.mutedBy = timeIntervalNames
```
ultrotter, line 72: If timeIntervalNames is nil or empty, shoudl we actually delete from the map, here?

silence/silence.go
```diff
@@ -293,11 +300,8 @@ func (s *Silencer) Mutes(ctx context.Context, lset model.LabelSet) bool {
 		"active", len(activeIDs),
 		"pending", len(allIDs)-len(activeIDs),
 	)
-	// TODO: remove this sort once the marker is removed.
-	sort.Strings(activeIDs)
```
ultrotter, line 297: Are we sure we shouldn't still sort for the API to be consistent and the results comparable? not necessarily here, but somewhere...

marker/group.go
ultrotter, line 72: In other places you clone the slice, but not here... Any particular reason?

siavashs: I am not sure, but this is just moving existing logic and is not new code.

siavashs: This is just moving existing logic and is not new code.
The PR is not focused on Group Markers used by time intervals so I would leave any changes related to those for later.

api/v2/api.go
siavashs: No, we need to keep it as is since each call to the returned function is a new operation that we want to trace as a child span.
This specific call returns a derived context which is pointing tot he parent.

dispatch/metric.go
siavashs: I guess not, because these are gauge metrics and can drop to zero, so I assume having zero values in between changes is desired compared to missing metrics.
What do you think?

silence/silence.go
siavashs: I am not sure, but does it matter if the IDs are in a specific order or do we guarantee this?

api/v2/api.go
```diff
@@ -433,7 +431,15 @@ func (api *API) getAlertGroupsHandler(params alertgroup_ops.GetAlertGroupsParams
...
-	af := api.alertFilter(matchers, *params.Silenced, *params.Inhibited, *params.Active)
+	// Use a temporary marker to "predict" each alert's current status at
+	// query time (silenced, inhibited, active). The alertFilter calls
+	// setAlertStatus which runs the inhibitor/silencer pipeline and writes
+	// the result into tempMarker. We then use tempMarker for the response
+	// to preserve backward-compatible behavior that clients depend on.
+	// TODO: consider a feature flag or API query param to let callers opt
+	// into using the group's actual marker status instead of the prediction.
+	tempMarker := marker.NewAlertMarker()
+	af := api.alertFilter(ctx, matchers, *params.Silenced, *params.Inhibited, *params.Active, tempMarker)
```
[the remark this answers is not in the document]
Spaceman1701: I think the robot is right, this should return the "predicted" (or calculated) status as it is, not a per-request status

siavashs:
@coderabbitai review

api/v2/api.go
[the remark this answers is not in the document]
siavashs: @coderabbitai does the last commit resolve this issue?

Spaceman1701 (approved):

Issue #2205, opened 2020-03-10
Implement proper retries for integrations


simonpasquier:
Looking at the various integrations, some of them provide hints on how you should retry once you're rate-limited:

- [ ] [OpsGenie](https://docs.opsgenie.com/docs/api-rate-limiting#section-handling-throttles-and-request-delays)
- [ ] [Slack](https://api.slack.com/docs/rate-limits#headers)
- [ ] [PagerDuty](https://v2.developer.pagerduty.com/docs/rate-limiting)
- [ ] [PushOver](https://pushover.net/api#limits)
- [ ] [Discord](https://discord.com/developers/docs/topics/rate-limits#exceeding-a-rate-limit)

Currently the retry behavior for all integrations is a simple back-off mechanism using the [github.com/cenkalti/backoff/v4](https://pkg.go.dev/github.com/cenkalti/backoff/v4) library.

Other integrations ([VictorOps](https://help.victorops.com/knowledge-base/api/), WeChat) don't provide much information about retry strategies so the current implementation is probably good enough for them. The same is true for the webhook integration.

This is a follow up of #2121, #2128 and #2119. In particular this [comment](https://github.com/prometheus/alertmanager/issues/2121#issuecomment-560340181) is relevant.

RajatVaryani:
@simonpasquier I'm interested working on the issue.

vears91:
Hi, is this issue still relevant? I'd like to work on it if it still is

tommysitehost:
Hey I think this is still relevant as more and more backends adopt 429 HTTP code.

vvxxvvxx:
Hi @simonpasquier , is this `unexpected status code 429: {\"retry_after\":1,\"ok\":false,\"error\":\"rate_limited\"}"` issue fixed already?

aleem-99:
Facing same error for slack notifications 
alertmanager version v0.24.0
```notify retry canceled due to unrecoverable error after 1 attempts: channel \"\": unexpected status code 429: {\"retry_after\":1,\"ok\":false,\"error\":\"rate_limited\"}"```

atkinsonm:
👋🏻 from Slack (for public posting purposes, representing just myself). When Slack's API returns code 429, my understanding is it adheres to the [RFC 6585](https://datatracker.ietf.org/doc/html/rfc6585) standard. 429 from Slack's API should be retried.

atkinsonm:
This issue or at least the Slack portion may have been completed by #5389

PR #5048, opened 2026-02-26
slack: retry on 429 and honor Retry-After header

### LINKED ISSUE #2111: Slack Integration does not respect rate limiting
**What did you do?**
Starting seeing the following errors in AlertManager

```
level=error ts=2019-11-21T09:02:58.967Z caller=notify.go:372 component=dispatcher msg="Error on notify" err="cancelling notify retry for \"slack\" due to unrecoverable error: unexpected status code 429: {\"retry_after\":1,\"ok\":false,\"error\":\"rate_limited\"}" context_err=null
level=error ts=2019-11-21T09:02:58.967Z caller=dispatch.go:266 component=dispatcher msg="Notify for alerts failed" num_alerts=2 err="cancelling notify retry for \"slack\" due to unrecoverable error: unexpected status code 429: {\"retry_after\":1,\"ok\":false,\"error\":\"rate_limited\"}"
```

I would expect the dispatcher to respect the `retry_after` and try later rather than failing

Increase volume of alerts triggered

**Environment**

* System information:

	Docker container

* Alertmanager version:

	v0.19.0

* Prometheus version:

	N/A

N/A - contains confidential information

N/A

* Logs:
```
level=error ts=2019-11-21T09:02:58.967Z caller=notify.go:372 component=dispatcher msg="Error on notify" err="cancelling notify retry for \"slack\" due to unrecoverable error: unexpected status code 429: {\"retry_after\":1,\"ok\":false,\"error\":\"rate_limited\"}" context_err=null
level=error ts=2019-11-21T09:02:58.967Z caller=dispatch.go:266 component=dispatcher msg="Notify for alerts failed" num_alerts=2 err="cancelling notify retry for \"slack\" due to unrecoverable error: unexpected status code 429: {\"retry_after\":1,\"ok\":false,\"error\":\"rate_limited\"}"
```

biwwy0:
Make HTTP 429 (Too Many Requests) retryable for the Slack notifier by adding http.StatusTooManyRequests to the retrier's RetryCodes, matching the pattern already used by PagerDuty, OpsGenie, Jira, and incident.io.

Additionally, parse the Retry-After response header on 429 responses and sleep for the indicated duration before returning. This addresses the concern raised when a similar change was reverted in #2128: simply retrying 429 without honoring Retry-After causes rapid-fire retries that risk getting apps permanently disabled by Slack. The sleep respects context cancellation.

Fixes #2111
Ref #2205

#### Pull Request Checklist

    - Fixes #<issue number>
    
- Is this a bugfix?
- Is this a new feature?
- Is this a breaking change?

```release-notes
[ENHANCEMENT] slack: retry on 429 honoring Retry-After header
```

### DIFF
--- a/notify/slack/slack.go
+++ b/notify/slack/slack.go
@@ -22,7 +22,9 @@ import (
 	"log/slog"
 	"net/http"
 	"os"
+	"strconv"
 	"strings"
+	"time"
 
 	commoncfg "github.com/prometheus/common/config"
 
@@ -59,7 +61,7 @@ func New(c *config.SlackConfig, t *template.Template, l *slog.Logger, httpOpts .
 		tmpl:         t,
 		logger:       l,
 		client:       client,
-		retrier:      &notify.Retrier{},
+		retrier:      &notify.Retrier{RetryCodes: []int{http.StatusTooManyRequests}},
 		postJSONFunc: notify.PostJSON,
 	}, nil
 }
@@ -256,6 +258,15 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 	// classify them as retriable or not.
 	retry, err := n.retrier.Check(resp.StatusCode, resp.Body)
 	if err != nil {
+		if resp.StatusCode == http.StatusTooManyRequests {
+			if d := parseRetryAfter(resp.Header.Get("Retry-After")); d > 0 {
+				n.logger.Warn("Rate limited by Slack, waiting before retry", "retry_after_secs", d.Seconds())
+				select {
+				case <-time.After(d):
+				case <-ctx.Done():
+				}
+			}
+		}
 		err = fmt.Errorf("channel %q: %w", req.Channel, err)
 		return retry, notify.NewErrorWithReason(notify.GetFailureReasonFromStatusCode(resp.StatusCode), err)
 	}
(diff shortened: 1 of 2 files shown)


TheMeier:
Hi @biwwy0 thank you for your contribution,

I do think this is good however I think we should take the time and fix this for all notifiers. There is [multiple notifiers currently](https://github.com/search?q=repo%3Aprometheus%2Falertmanager%20http.StatusTooManyRequests&type=code) that just set `http.StatusTooManyRequests` as retryable and cause an immediate retry disregarding the `Retry-After` header. 
On top of that one also needs to think about what happens if `Retry-After` contains a "long" time, If the time spent for the previous request(s) + the retry delay is greater than `group_interval` the notification will get cancelled. (A new one is triggered at that point)

TheMeier:
After checking again there is one issue here. After waiting the `Retry-After` the incremental backup from the notify stage gets added on top. I will try to investigate further how we could do proper 429-handling in the retry-stage globally.

biwwy0:
@TheMeier Please let me know if you want me to change something here? I looked into moving `Retry-After` support into shared retrier, but it seems that every integration has very different return codes/headers and it won't do any good. So decided to think just for one we use mostly.

TheMeier:
I can't follow the reasoning. I think it should be possible to unify the handling of 429 accross all notifiers, regardless of special handling of other return codes in some notifieres. On top the reason for https://github.com/prometheus/alertmanager/pull/2128 also applies for all notifieres that currently just set `http.StatusTooManyRequests` as retrieable, so one could consider it a bug (or at least bad-behaviour).

themavik (commented):
RetryCodes includes 429 and Notify sleeps on integer Retry-After before returning retry=true, which should back off under rate limits. nit: parseRetryAfter only accepts integer seconds; HTTP-date Retry-After parses as 0 and you retry immediately—this could break if Slack sends that form.

SoloJacobs (approved):

Issue #4948, opened 2026-02-04
Event Recorder Interface


siavashs:
Based on the HRT event recorder interfaces.
A event recording interface, protobuf events, and log points wired through all of Alertmanager.

@Spaceman1701 will lead this, @siavashs will contribute or take over if necessary.

siavashs:
Implemented in #5072

PR #5072, opened 2026-03-04
eventrecorder: add structured event recorder

jackrosenthal:
This commit adds a structured event recorder to Alertmanager that
records events.  Events are defined in a protobuf schema and fanned
out to one or more configured outputs (JSONL file, webhook).  The
event recorder provides an audit trail for observability and
analytics, answering questions like "why did this alert fire?", "was
it silenced?", "which receiver got notified?", and "when did this
inhibition rule suppress it?".

The event recorder is gated behind --enable-feature=event-recorder.
When not enabled, a zero-value Recorder is used and all methods are
no-ops.

At this time, there is no filtering of events or event query API.
Downstream software is intended to consume the event recorder via the
webhook or file output and filter and store events in an appropriate
database as desired.  Alertmanager is kept away from this logic.

Ten event types are defined, covering the full alert lifecycle:

| Event | Emitted When |
|---|---|
| `AlertmanagerStartupEvent` | Alertmanager process starts |
| `AlertmanagerShutdownEvent` | Alertmanager process receives SIGTERM |
| `AlertCreatedEvent` | A new alert is inserted via the API |
| `AlertGroupedEvent` | An alert is added to an aggregation group |
| `AlertResolvedEvent` | An alert's end time passes during a flush |
| `NotificationEvent` | A notification is successfully delivered |
| `SilenceCreatedEvent` | A new silence is created |
| `SilenceUpdatedEvent` | An existing silence is updated in place |
| `SilenceMutedAlertEvent` | A silence matches and mutes an alert |
| `InhibitionMutedAlertEvent` | An inhibition rule suppresses an alert |

Each event carries a full protobuf-defined payload with alert details,
labels, annotations, fingerprints, group metadata, silence/inhibition
rule details, and receiver information as applicable.

The event recorder is configured via a new top-level `event_recorder`
section in `alertmanager.yml`, supporting multiple simultaneous
outputs:

```yaml
event_recorder:
  outputs:
    - type: file
      path: /var/log/alertmanager/events.jsonl
    - type: webhook
      url: https://example.com/events
      http_config:   # optional, same as receiver http_config
        tls_config:
          cert_file: /path/to/cert
      timeout: 10s       # optional, default 10s
      workers: 4         # optional, default 4
      max_retries: 3     # optional, default 3
      retry_backoff: 500ms  # optional, default 500ms
```

The configuration is hot-reloadable: sending SIGHUP swaps outputs
atomically (old outputs are closed, new ones created).

Writes JSONL to a file.  Uses fsnotify to watch the parent directory
for file renames and removes (e.g. logrotate), triggering an immediate
reopen.  The constructor blocks until the watcher is ready, avoiding
races in tests.

POSTs each event as a JSON body to a configured URL.  A bounded worker
pool (configurable via `workers`) processes deliveries concurrently so
that a slow webhook does not block the event queue.  Failed deliveries
are retried with exponential backoff (configurable via `max_retries`
and `retry_backoff`), capped at 30 seconds.  When the internal queue
(capacity 1024) is full, events are dropped and counted via the
`alertmanager_event_recorder_webhook_drops_total` Prometheus counter.

`RecordEvent` never blocks the caller.  Events are serialized with
`protojson.Marshal` and placed on a bounded in-memory channel (capacity
8,192, capping memory at roughly 4 MB).  A background goroutine drains
the channel and writes to all configured outputs.  If the channel is
full, the event is dropped and a metric is incremented.  The event
recorder must never slow down alert processing.

Concurrency is handled entirely via message passing — no mutexes.  The
`writeLoop` goroutine exclusively owns output state and config.
Config updates arrive via a `cfgUpdate` channel with an ack signal.
`cluster.Peer` uses `atomic.Pointer`.  `Close` uses `sync.Once`.

`RecordEvent` accepts a `context.Context`.  Recording is opt-in:
callers must pass `eventrecorder.WithEventRecording(ctx)` to enable
recording.  The API update path omits this to avoid flooding the event
queue when checking silences and inhibitions for status display.

Event construction and protobuf conversion helpers live in the
`eventrecorder` package: `NewAlertCreatedEvent`,
`NewSilenceMutedAlertEvent`, `NewSilenceCreatedEvent`,
`NewSilenceUpdatedEvent`, `NewInhibitionMutedAlertEvent`,
`SilenceAsProto`, `InhibitRuleAsProto`, `SilenceMatcherAsProto`,
`MatcherAsProto`, `MatchersAsProto`, `LabelSetAsProto`.  Call sites
in `inhibit`, `silence`, and `provider/mem` are one-liners.

`extractEventType` uses proto reflection (`WhichOneof`) to derive the
event type name from the oneof field, avoiding a manual type switch.

Five Prometheus metrics are exposed:

- `alertmanager_events_recorded_total` (labels: `event_type`, `output`, `result`)
- `alertmanager_event_recorder_bytes_written_total` (labels: `event_type`, `output`)
- `alertmanager_events_dropped_total` (labels: `event_type`)
- `alertmanager_event_serialize_errors_total` (labels: `event_type`)
- `alertmanager_event_recorder_queue_length` (gauge)
- `alertmanager_event_recorder_webhook_drops_total` (labels: `output`)

The `output` label uses a stable identifier (e.g. `file:/path` or
`webhook:https://host/path`) rather than a positional index, so it
remains meaningful across config reloads.  Webhook URLs are sanitized
to strip userinfo and query parameters before use in identifiers,
metrics labels, and log messages.

Both `SilenceCreated` and `SilenceUpdated` events are recorded at the
same level in the `Set` method, rather than split between `setSilence`
(for create) and `Set` (for update).  `setSilence` returns
`(changed, added bool, err error)` so the caller can decide what to
record and whether to broadcast.

Several new context keys were added to `notify/notify.go` to thread
dispatch-time metadata through the notification pipeline:

- `WithGroupMatchers` / `GroupMatchers`: route-level matchers
- `WithRouteLabels` / `RouteLabels`: route-level labels
- `WithFlushId` / `FlushId`: per-flush monotonic counter
- `WithAggrGroupId` / `AggrGroupId`: aggregation group UUID
- `WithMutedAlerts` / `MutedAlerts`: muted alert fingerprints
- `WithNotificationReason` / `NotificationReason`: trigger reason

These are populated in `dispatch/dispatch.go` where the aggregation
group context is set up, and consumed in `notify/event.go` to build
rich `NotificationEvent` and `AlertGroupedEvent` payloads.

Unit tests cover the eventrecorder package:

- `eventrecorder/eventrecorder_test.go`: RecordEvent delivery,
  NopRecorder, zero-value Recorder, WithEventRecording,
  ApplyConfig no-op, extractEventType, LabelSetAsProto,
  MatcherAsProto, MatchersAsProto, eventRecorderConfigEqual.
- `eventrecorder/file_test.go`: SendEvent, reopen-after-rename,
  reopen-after-remove, Close, invalid path.
- `eventrecorder/webhook_test.go`: webhook delivery via
  httptest.Server, concurrent workers, retry on failure, drop after
  max retries, and graceful close draining.

Additionally, tested with a live Alertmanager instance configured with
JSONL file output, a webhook endpoint, inhibition rules, and silences.
All 10 event types were verified:

1. Started Alertmanager with `--enable-feature=event-recorder` and
   `event_recorder` configured.  Verified
   `alertmanagerStartupEvent`.
2. Fired test alerts.  Verified `alertCreated` and `alertGrouped`.
3. Fired `CriticalAlert` (severity=critical) and `WarningAlert`
   (severity=warning) with an inhibition rule suppressing warnings
   when a matching critical alert exists.  Verified
   `inhibitionMutedAlert` with correct rule details and source
   fingerprint.
4. Created a silence matching `SilencedAlert`, then fired
   `SilencedAlert`.  Verified `silenceCreated` and
   `silenceMutedAlert` with silence ID and matchers.
5. Updated an active silence via the v2 API, preserving matchers and
   extending the end time (satisfying `canUpdate`).  Verified
   `silenceUpdated` with the updated silence details and same ID.
6. Waited for notification delivery.  Verified `notification` events
   with receiver name, integration type, and alert details.
7. Fired a short-lived alert (5-second TTL), waited for flush.
   Verified `alertResolved`.
8. Sent SIGTERM.  Verified `alertmanagerShutdownEvent`.
9. Tested hot-reload: started with file-only config, sent SIGHUP with
   file+webhook config, verified new events appeared on the webhook
   and old file continued receiving events.
10. Verified feature flag gating: started without
    `--enable-feature=event-recorder`, confirmed no event file was
    created and no events were recorded.

## Summary by CodeRabbit

* **New Features**
  * Opt-in event recording for startup/shutdown, alerts, silences, notifications, grouping and inhibition using a structured protobuf event envelope; integrated across alert lifecycle, grouping, delivery and inhibition.
  * New file and webhook outputs with JSONL delivery, reopen/rotation handling, worker pools, retries, non-blocking enqueueing, and delivery/drop metrics.
* **Configuration**
  * New event_recorder config (outputs, timeouts, workers, retries) and experimental "event-recorder" feature flag.


siavashs:
Thank you @jackrosenthal 
I'm going to take a look but for now check this agent's report about potential bugs https://app.devin.ai/review/prometheus/alertmanager/pull/5072

jackrosenthal:
Addressed findings from [Devin review](https://app.devin.ai/review/prometheus/alertmanager/pull/5072):

- Fixed nil pointer dereference in `renderMatcher` when `labels.NewMatcher` fails on invalid regex (`silence/silence.go`)
- Drain HTTP response body before closing in webhook output to enable connection reuse (`eventlog/webhook.go`)
- `ApplyConfig` now skips when config is unchanged, avoiding unnecessary output churn and a startup race (`eventlog/eventlog.go`)
- Per-output metrics: `eventsRecorded` and `eventLogBytesWritten` now carry an `output` label; serialization errors and dropped events split into separate counters (`eventlog/eventlog.go`)
- Added `MutesNoRecord` to `Silencer` and `Inhibitor`; `api.Update` closure uses it to avoid flooding the event queue (`silence/silence.go`, `inhibit/inhibit.go`, `cmd/alertmanager/main.go`)

Acking from Devin (no action needed):

- `HTTPConfig.SetDirectory` on range copy is fine — `HTTPConfig` is a pointer
- Defer ordering for shutdown event and `Close()` is correct (LIFO)
- `setSilence` correctly avoids duplicate `SilenceCreated` events on updates

eventlog/eventlog.go
```diff
@@ -0,0 +1,413 @@
...
+		Annotations: LabelSetAsProto(alert.Annotations),
+		StartsAt:    timestamppb.New(alert.StartsAt),
+		EndsAt:      timestamppb.New(alert.EndsAt),
+		Resolved:    alert.Resolved(),
+	}
+}
+
+func MatchersAsProto(matchers labels.Matchers) []*eventlogpb.Matcher {
+	result := make([]*eventlogpb.Matcher, len(matchers))
+	for i, m := range matchers {
```
ultrotter: Should we have a SingleMatcherAsProto func called here. and use it also instead of convertMatcherForEvent() in inhibit/inhibit.go:428-445 so we avoid duplicating all of this?

silence/silence.go
```diff
@@ -845,7 +957,17 @@ func (s *Silences) Set(ctx context.Context, sil *pb.Silence) error {
 		if err := s.checkSizeLimits(msil); err != nil {
 			return err
 		}
-		return s.setSilence(msil, now)
+		if err := s.setSilence(msil, now); err != nil {
+			return err
+		}
```
ultrotter, line 873: Seeing this flow here makes me think: here we are relying on the fact that setSilence will not emit the event itself, relying on the fact that if canUpdate the merge will succeed. Is it always guaranteed? Should we at least log/track if we call setSilence expecting no new silence and one happens against all odds? Or should we consider it out of scope? Also the pattern of calling record event in setsilence but only on one path, and out of it on the other path, is a bit unbalanced. I am wondering if we can always record inside setSilence, or always out

silence/silence.go
```diff
@@ -273,6 +287,18 @@ func (s *Silencer) Mutes(ctx context.Context, lset model.LabelSet) bool {
 			case SilenceStateActive:
 				activeIDs = append(activeIDs, sil.Id)
 				allIDs = append(allIDs, sil.Id)
+
+				recorder.RecordEvent(&eventlogpb.EventData{
```
ultrotter: Would it make sense to emit just one event with all the silences that mute the alert, rather? WDYT?

eventlog/webhook.go
```diff
@@ -0,0 +1,80 @@
...
+	client.Timeout = timeout
+
+	return &WebhookOutput{
+		client: client,
+		url:    cfg.URL.String(),
+	}, nil
+}
+
+// WriteEvent POSTs the pre-serialized JSON bytes to the webhook URL.
+func (wo *WebhookOutput) WriteEvent(data []byte) error {
```
ultrotter: Should we have a goroutine per-POST or a max number of workers, to avoid a slow or temporarily down webhook from stalling us, and/or a small number of retries before dropping? and should we have a metric for retried, dropped, and length of the requests?

eventlog/eventlog.go
```diff
@@ -0,0 +1,413 @@
...
+	c.mu.RLock()
+	outputs := c.outputs
+	c.mu.RUnlock()
+
+	if len(outputs) == 0 {
+		return
+	}
+
+	for i, out := range outputs {
+		idx := strconv.Itoa(i)
```
ultrotter: This index could change on conf reload, right? Should we use file:path or webhook:url for a more stable long term meaning of the output label?

eventlog/eventlog.go
```diff
@@ -0,0 +1,413 @@
...
+
+// Package eventlog provides an audit log of significant Alertmanager
+// events.  Events are serialized as JSON and fanned out to one or more
+// configured outputs (JSONL file, webhook, etc.).
+//
+// RecordEvent never blocks the caller: events are serialized and
+// placed on a bounded in-memory queue.  A background goroutine
+// drains the queue and writes to outputs.  If the queue is full,
+// events are dropped and a metric is incremented.
+package eventlog
```
ultrotter: Should we add any test for the eventlog package?

eventlog/eventlog.go
```diff
@@ -0,0 +1,413 @@
...
+// are closed and new ones are created from the given config.
+func (r Recorder) ApplyConfig(cfg config.EventLogConfig) {
+	if r.core == nil || r.core.logger == nil {
+		return
+	}
+
+	// Skip if the config hasn't changed to avoid closing and
+	// recreating identical outputs.
+	r.core.mu.Lock()
+	if reflect.DeepEqual(cfg, r.core.currentCfg) {
```
ultrotter: Should we use a different comparison here, only on what we care about and avoiding doing a full deepequal on a struct which might have unexported fields and such and is quite complex? 
We could for example keep a hash of the yaml that we used to build the config and only rebuild if the new yaml has a different hash...

eventlog/eventlog.go
```diff
@@ -0,0 +1,413 @@
...
+		r.core.mu.Unlock()
+		return
+	}
+
+	newOutputs := buildOutputs(cfg.Outputs, r.core.logger)
```
ultrotter: Is there any way to build these without holding the lock, then swapping them? Or for example we could have an atomic pointer to the config instead of a lock...

cmd/alertmanager/main.go
```diff
@@ -503,8 +542,8 @@ func run() int {
 		configuredInhibitionRules.Set(float64(len(conf.InhibitRules)))
 
 		api.Update(conf, func(ctx context.Context, labels model.LabelSet) {
-			inhibitor.Load().Mutes(ctx, labels)
-			silencer.Mutes(ctx, labels)
+			inhibitor.Load().MutesNoRecord(ctx, labels)
+			silencer.MutesNoRecord(ctx, labels)
```
ultrotter: I am wondering if instead of a separate MutesNoRecord call in both modules we could pass a context to disable the event recorder for the call?

eventlog/file.go
```diff
@@ -0,0 +1,78 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package eventlog
+
+import (
+	"log/slog"
+	"time"
+
+	"github.com/pengsrc/go-shared/reopen"
```
ultrotter: This is a library that hasn't been updated in years. 8 years, not 2. :) At this point rather than the dependency it might be better to implement the minimum of the needful. Also should we consider using maybe https://github.com/fsnotify/fsnotify to detect if we need to reopen instead of blindly reopening every 5 min?

silence/silence.go
jackrosenthal: I considered this, but I think per-silence events are more useful for downstream consumers.  Each event is self-contained and can be filtered/routed independently, and the schema stays simple (single silence per event rather than a repeated field).  If an alert is muted by 3 silences, 3 separate events is a clearer audit trail than one batched event where the consumer has to iterate.

jackrosenthal:
Addressed review feedback:

- Extracted `eventlog.MatcherAsProto` and removed duplicate `convertMatcherForEvent` from `inhibit/inhibit.go`.
- Moved `SilenceCreated` recording out of `setSilence` into `Set`, so both create and update events are recorded at the same level.  `setSilence` now returns `(added bool, err error)`.
- `WebhookOutput` has a bounded worker pool (`workers`, default 4) with exponential-backoff retries (`max_retries`, `retry_backoff`).  New config fields on `EventLogOutput`.
- Added `Name()` to the `Output` interface (`file:<path>`, `webhook:<url>`).  Metrics use these instead of positional indices.
- Replaced `reflect.DeepEqual` in `ApplyConfig` with `eventLogConfigEqual` comparing fields explicitly.
- `ApplyConfig` builds outputs outside the lock, then swaps under the lock.
- `RecordEvent` takes a `context.Context`.  `eventlog.WithRecordingDisabled(ctx)` suppresses recording.  Removed `MutesNoRecord` from `Silencer` and `Inhibitor`.
- Replaced `pengsrc/go-shared/reopen` with inline file-reopen logic using `fsnotify` for immediate rotation detection.  Removed the stale dependency.
- Added 16 unit tests (`eventlog_test.go`, `webhook_test.go`) covering delivery, retries, context suppression, config comparison, and conversion helpers.

inhibit/inhibit.go
```diff
@@ -209,6 +213,19 @@ func (ih *Inhibitor) Mutes(ctx context.Context, lset model.LabelSet) bool {
...
+					InhibitionMutedAlert: &eventlogpb.InhibitionMutedAlertEvent{
+						InhibitRules: []*eventlogpb.InhibitRule{convertInhibitRuleForEvent(r)},
+						MutedAlert: &eventlogpb.MutedAlert{
+							Fingerprint: uint64(fp),
+							Labels:      eventlog.LabelSetAsProto(lset),
+						},
+						InhibitingFingerprints: []uint64{uint64(inhibitedByFP)},
+					},
+				},
+			})
```
siavashs: Can we add some helper functions to make constructing events less verbose in the emitter side?
For example:

```suggestion
			ih.recorder.RecordEvent(ctx, eventlog.NewInhibitionMutedAlertEvent(r, fp, lset, inhibitedBy))
```

inhibit/inhibit.go
```diff
@@ -399,3 +416,26 @@ func (r *InhibitRule) hasEqual(lset model.LabelSet, excludeTwoSidedMatch bool, n
...
+	for label := range r.Equal {
+		equalLabels = append(equalLabels, string(label))
+	}
+
+	return &eventlogpb.InhibitRule{
+		SourceMatchers: sourceMatchers,
+		TargetMatchers: targetMatchers,
+		EqualLabels:    equalLabels,
+	}
+}
```
siavashs: following my previous comment this should live in the `eventlog` package and be used internally there.

siavashs:
I think the package name `eventlog` is too specific and I would prefer naming the package something like `record` which is more generic and follows the "convention" we have in Alertmanager.
For example `silence.NewSilenser()`, `record.NewRecorder()` or `record.NewEventRecorder()` and then you could call `recorder.Record()` or `recorder.RecordEvent()`, etc.
What do you think?

eventlog/file.go
```diff
@@ -0,0 +1,129 @@
...
+	for {
+		select {
+		case event, ok := <-watcher.Events:
+			if !ok {
+				return
+			}
+			if event.Op&(fsnotify.Rename|fsnotify.Remove) != 0 {
+				fo.reopen()
+				// Re-add the watch since the original inode is gone.
+				_ = watcher.Add(fo.path)
```
ultrotter: Here in case of error we may miss future notifications. Would it make sense to watch the directory instead, only once, and filter the filename we care about from the notification we get from the changes there?

eventlog/webhook.go
```diff
@@ -0,0 +1,172 @@
...
+		retryBackoff = time.Duration(cfg.RetryBackoff)
+	}
+
+	wo := &WebhookOutput{
+		client:       client,
+		url:          cfg.URL.String(),
+		maxRetries:   maxRetries,
+		retryBackoff: retryBackoff,
+		logger:       logger,
+		work:         make(chan []byte, workers),
```
ultrotter: This chan has 4 spaces... So when below you do 
	wo.work <- data
It'll block after 4 workers are busy and 4 events are enqueued. Should we have a larger channel here?
And maybe below drop events and log/increase a metric if the big channel is completely full?

eventrecorder/file.go
```diff
@@ -0,0 +1,129 @@
...
+	"os"
+	"sync"
+
+	"github.com/fsnotify/fsnotify"
+)
+
+// FileOutput writes pre-serialized JSON event bytes to a JSONL file.
+// The file is reopened when fsnotify detects a rename or remove (e.g.
+// logrotate).
+type FileOutput struct {
```
ultrotter, line 29: Can we have some tests for fileoutput and the relevant reopener/notification that all works as we expect?

eventlog/eventlog.go
```diff
@@ -0,0 +1,487 @@
...
+
+	if unchanged {
+		return
+	}
+
+	// Build new outputs without holding the lock (may involve I/O).
+	newOutputs := buildOutputs(cfg.Outputs, r.core.logger)
+
+	// Swap under the lock.
+	r.core.mu.Lock()
```
ultrotter: Here we should check that oldOutputs is unchanged since we unlocked and re-locked. I know it's unlikely, but... If it was an atomic pointer we would use:
func (*Pointer[T]) [CompareAndSwap]

But otherwise we could use a version number stored inside, or a hash, or re-compare eventLogConfigEqual but that seems less ideal...

eventlog/webhook.go
```diff
@@ -0,0 +1,172 @@
...
+
+func (wo *WebhookOutput) postWithRetry(data []byte) {
+	for attempt := range wo.maxRetries {
+		err := wo.post(data)
+		if err == nil {
+			return
+		}
+		wo.logger.Warn("Event log webhook POST failed", "url", wo.url, "attempt", attempt+1, "err", err)
+		if attempt < wo.maxRetries-1 {
+			backoff := time.Duration(float64(wo.retryBackoff) * math.Pow(2, float64(attempt)))
```
ultrotter: Should we have a max backoff here? the largest one with the default values is 256 seconds... ? 
And should we have a metric for dropped values?
also do we need to bring in math.Pow here? (though it shouldn't matter too much, fair enough).

eventlog/eventlog.go
```diff
@@ -0,0 +1,487 @@
...
+
+// writeLoop drains the event queue and writes to outputs.  It runs
+// until the done channel is closed, then drains any remaining events.
+func (c *recorderCore) writeLoop() {
+	defer c.wg.Done()
+	for {
+		select {
+		case req := <-c.events:
+			c.writeToOutputs(req)
+		case <-c.done:
```
ultrotter: Here in this case while we are draining c.events the workers may have already found an empty channel and quit. i.e.
done is closed.
all workers quit because their channel is empty
and then we get scheduled and start adding events.... which nobody handles! :(

We should first tell writeLoop to drain-and-return and only after it's done tell the workers to drain-and-return.

eventlog/eventlog_test.go
```diff
@@ -0,0 +1,252 @@
...
+
+	// Record one event to the initial output.
+	rec.RecordEvent(context.Background(), startupEvent())
+	require.Eventually(t, func() bool {
+		return out1.eventCount() == 1
+	}, time.Second, 10*time.Millisecond)
+
+	// Apply a new config.  Since we can't easily create real outputs
+	// in a test, we manually swap outputs.  But we can verify that an
+	// unchanged config is a no-op.
```
ultrotter: Uhm it does seem that we should also verify that the config changes and new events go to the new output correctly...

eventlog/eventlog.go
```diff
@@ -0,0 +1,487 @@
...
+			r.core.logger.Error("Failed to close old event log output", "err", err)
+		}
+	}
+
+	r.core.logger.Info("Event log configuration reloaded", "outputs", len(newOutputs))
+}
+
+// Close signals the background goroutine to drain remaining events
+// and stop, then closes all outputs.
+func (r Recorder) Close() error {
```
ultrotter: Shall we test that Close drained all events properly?

eventlog/eventlog.go
```diff
@@ -0,0 +1,487 @@
...
+		return
+	}
+
+	data = append(data, '\n')
+
+	select {
+	case r.core.events <- writeRequest{data: data, eventType: eventType}:
+	default:
+		// Queue full; drop event to avoid blocking alertmanager.
+		r.core.metrics.eventsDropped.WithLabelValues(eventType).Inc()
```
ultrotter: Should we add a test for this unhappy path, and potentially other unhappy paths?

silence/silence.go
```diff
@@ -802,18 +823,91 @@ func (s *Silences) toMeshSilence(sil *pb.Silence) *pb.MeshSilence {
...
+	return &eventlogpb.Silence{
+		Id:          sil.Id,
+		Matchers:    matchers,
+		MatcherSets: matcherSets,
+		StartsAt:    sil.StartsAt,
+		EndsAt:      sil.EndsAt,
+		UpdatedAt:   sil.UpdatedAt,
+		CreatedBy:   sil.CreatedBy,
+		Comment:     sil.Comment,
+	}
```
siavashs: This does not access anything from `Silences` instance `s` and should not be part of it.
Since both sides are the conversion are proto types, can we use generated code to handle this conversion?
This should be a helper function in the `eventlog` (whatever we name it) package.

eventlog/eventlog.go
```diff
@@ -0,0 +1,487 @@
...
+func WithRecordingDisabled(ctx context.Context) context.Context {
+	return context.WithValue(ctx, recordingDisabledContextKey{}, true)
+}
+
+// RecordingDisabled reports whether event recording has been suppressed
+// in the given context via WithRecordingDisabled.
+func RecordingDisabled(ctx context.Context) bool {
+	v, _ := ctx.Value(recordingDisabledContextKey{}).(bool)
+	return v
+}
```
siavashs: This reads very odd, we should have it disabled by default and then have an option to enable it.
For example `WithEventRecording()`

eventlog/eventlog.go
```diff
@@ -0,0 +1,487 @@
...
+
+// writeRequest is a single event queued for background writing.
+type writeRequest struct {
+	data      []byte
+	eventType string
+}
+
+// recorderCore holds the mutable state shared by all copies of a
+// Recorder value.  Access is protected by mu.
+type recorderCore struct {
```
siavashs: Maybe `sharedRecorder` would be a better name?

eventlog/eventlog.go
```diff
@@ -0,0 +1,487 @@
...
+// Output is a single event log destination.  Implementations receive
+// pre-serialized JSON bytes and are responsible for delivering them.
+type Output interface {
+	// Name returns a stable identifier for this output, suitable for
+	// use as a Prometheus label value (e.g. "file:/var/log/events.jsonl"
+	// or "webhook:https://example.com/hook").
+	Name() string
+	WriteEvent(data []byte) error
+	io.Closer
+}
```
siavashs: Maybe call it `Destination` and `SendEvent()`?

eventlog/eventlog.go
```diff
@@ -0,0 +1,487 @@
...
+
+// SetClusterPeer sets the cluster peer for HA position tracking.
+func (r Recorder) SetClusterPeer(peer *cluster.Peer) {
+	if r.core == nil {
+		return
+	}
+	r.core.mu.Lock()
+	r.core.peer = peer
+	r.core.mu.Unlock()
```
siavashs: Let's use `defer`.

eventlog/eventlog.go
```diff
@@ -0,0 +1,487 @@
...
+func (r Recorder) ApplyConfig(cfg config.EventLogConfig) {
+	if r.core == nil || r.core.logger == nil {
+		return
+	}
+
+	// Skip if the config hasn't changed to avoid closing and
+	// recreating identical outputs.
+	r.core.mu.RLock()
+	unchanged := eventLogConfigEqual(cfg, r.core.currentCfg)
+	r.core.mu.RUnlock()
```
siavashs: Same here, we should use `defer`.

eventlog/eventlog.go
```diff
@@ -0,0 +1,487 @@
...
+	case *eventlogpb.EventData_SilenceUpdated:
+		return "silence_updated"
+	case *eventlogpb.EventData_SilenceMutedAlert:
+		return "silence_muted_alert"
+	case *eventlogpb.EventData_InhibitionMutedAlert:
+		return "inhibition_muted_alert"
+	default:
+		return "unknown"
+	}
+}
```
siavashs: For functions like this we could potentially do one of these alternatives below.
### Reflection
```go
    msg := event.ProtoReflect()
    od := msg.Descriptor().Oneofs().ByName("event_type") // use actual oneof name
    if od == nil {
        return "unknown"
    }
    fd := msg.WhichOneof(od)
    if fd == nil {
        return "unknown"
    }
    return string(fd.Name()) // returns the proto field name, e.g. "alert_created"
```

### Proto enum
```proto
enum EventType {
    EVENT_TYPE_UNKNOWN = 0;
    EVENT_TYPE_STARTUP = 1;
    EVENT_TYPE_SHUTDOWN = 2;
    EVENT_TYPE_ALERT_CREATED = 3;
    // ...
}
 
message EventData {
    EventType type = 1;
    oneof event_type {
        AlertmanagerStartupEvent alertmanager_startup_event = 10;
        // ...
    }
}
```
Then we can use `.String()` from the generated enum.

eventrecorder/webhook.go
```diff
@@ -0,0 +1,172 @@
...
+	defaultWebhookTimeout      = 10 * time.Second
+	defaultWebhookWorkers      = 4
+	defaultWebhookMaxRetries   = 3
+	defaultWebhookRetryBackoff = 500 * time.Millisecond
+)
+
+// WebhookOutput POSTs each event as a JSON body to a configured URL.
+// Events are processed by a bounded worker pool so that a slow or
+// temporarily unavailable webhook does not block the event log queue.
+type WebhookOutput struct {
```
SoloJacobs, line 46: Should we really call this `Webhook`? This name is already taken by the notifier.

cmd/alertmanager/main.go
```diff
@@ -270,6 +272,35 @@ func run() int {
...
+	eventRecorder.RecordEvent(context.Background(), &eventlogpb.EventData{
+		EventType: &eventlogpb.EventData_AlertmanagerStartupEvent{
+			AlertmanagerStartupEvent: &eventlogpb.AlertmanagerStartupEvent{
+				Version:      version.Version,
+				BuildContext: version.BuildContext(),
+			},
+		},
+	})
+	defer func() {
+		eventRecorder.RecordEvent(context.Background(), &eventlogpb.EventData{
```
SoloJacobs: Can we put this explicitly at the end of the function?

SoloJacobs (dismissed):
General remark: It will be easier to review and merge this, if we did the file.go and webbok.go implementation in a separate PR, but here we go.

a) Some of these events are extremely detailed. E.g., `InhibitionMutedAlertEvent` . We could construct all the relevant information in Alertmanager just from the Silences, the configuration and the alerts. If we precompute results, then this isn't really event sourcing? We already have tracing for adding observability to alertmanager.

b) I don't think we should need any locks in an event-recorder? Can we just use message passing. Below the AIs attempt (Sorry, I let the AI do this, I otherwise would have to delay this review even more):

1. Locking for the `cluster.peer` is weird, we should simply emit an event with the Position, whenever `peer.Position()` is called. We could also keep track of that position inside the `EventRecorder`, but definitely not from

```
func (r Recorder) UpdateClusterPosition(pos int) {
	if r.ch == nil { return }
	p := uint32(pos)
	select {
	case r.ch <- logRequest{clusterPos: &p}:
	default:
		// Queue full. For state updates, we might want to retry or block briefly,
		// but dropping is standard for this non-blocking design.
	}
}
```

2. The EventRecorder should be able to handle config reloads just by message passing:

```
func (c *recorderCore) writeLoop(reqs <-chan logRequest) {
	var currentClusterPos uint32 = 0
	var outputs []Output // <--- Local state, completely lock-free!

	// Ensure we clean up whatever is active when the process finally exits
	defer func() {
		for _, out := range outputs {
			out.Close()
		}
	}()

	for req := range reqs {
		// --- CONTROL MESSAGE: CONFIG RELOAD ---
		if req.newOutputs != nil {
			c.logger.Info("Consumer applying new event log configuration")
			
			// 1. Gracefully close the old outputs
			for _, out := range outputs {
				if err := out.Close(); err != nil {
					c.logger.Error("Failed to close old output", "err", err)
				}
			}
			
			// 2. Swap to the new outputs
			outputs = req.newOutputs
			continue
		}

		// --- CONTROL MESSAGE: CLUSTER POSITION ---
		if req.clusterPos != nil {
			currentClusterPos = *req.clusterPos
			continue
		}

		// --- STANDARD DATA MESSAGE: EVENT ---
		if req.event != nil {
			// (Wrap the event, stamp the cluster position, marshal to JSON...)
			data := marshalEvent(req.event, currentClusterPos)
			
			// Write to the currently active outputs
			for _, out := range outputs {
				out.WriteEvent(data)
			}
		}
	}
}
```

siavashs:
> a) Some of these events are extremely detailed. E.g., `InhibitionMutedAlertEvent` . We could construct all the relevant information in Alertmanager just from the Silences, the configuration and the alerts. If we precompute results, then this isn't really event sourcing? We already have tracing for adding observability to alertmanager.

So event log always records all events, while tracing is for sampling. Technically one could sample traces at 100% but tracing is less efficient than event recording. Event recording is more about analytics than for observability.

SoloJacobs:
> So event log always records all events, while tracing is for sampling. Technically one could sample traces at 100% but tracing is less efficient than event recording. Event recording is more about analytics than for observability.

But we could very easily address the sampling issue by using our existing logger, correct? It wasn't clear from my first review, but the actual code changes I requested were for item b). For item a), what I really need is a coherent strategy for our different logging mechanism. When to pick which emitter (logs, metrics, traces, event log)? How to deal with duplicate events? This feels like the part, which is not explicit here.

As far as I can see, we can't really use the EventRecorder for event sourcing. That is fine, but we really need to be clear about what capabilities this feature gives users. Currently, the PR just describes what events are added, but not why.

siavashs:
> But we could very easily address the sampling issue by using our existing logger, correct? It wasn't clear from my first review, but the actual code changes I requested were for item b). For item a), what I really need is a coherent strategy for our different logging mechanism. When to pick which emitter (logs, metrics, traces, event log)? How to deal with duplicate events? This feels like the part, which is not explicit here.

So some background: `eventlog` is not a very good name (see my comment above about the package naming), the history is that HRT uses this implementation internally to log all the events happening in Alertmanager, this is different from application logs and it is intentionally kept separate.
Now since the meet up we discussed about open sourcing this implementation and making it generic enough for others to use. For example we at CF want to build our analytics system on top of the events recorder and we do not want to use log files but rather messaging or kafka, etc.

I agree the description of PR can be better, or be linked to the issue #4948

eventrecorder/webhook.go
Spaceman1701: While I do think it's a little confusing, I also think it's ok since it's in another package. IMO, It's fairly conventional to have this kind of name duplication in Go since the package name always prefixes the symbol name.

silence/silence.go
Spaceman1701: I think it is guaranteed - the checks before now check the `id` matches so even if `canUpdate` is wrong we'll still end up overwriting an existing silence when we call `setSilence`.

Spaceman1701 (commented):
Thanks for opening this PR! I agree with the general requests around cleanup (e.g. moving event construction into the `eventlog` package), but overall the code and behavior looks good (and familiar 😅 ). 

I also wanted to mention that we discussed this change during the last Alertmanager working group meeting. I think the biggest takeaway is that we should put this change behind an experimental feature flag now - the `eventlog` protos will be the first "public" protos in Alertmanager, so it'd be nice to have some time to iterate on them before we have to support them as part of the stable API.

Spaceman1701:
> > So event log always records all events, while tracing is for sampling. Technically one could sample traces at 100% but tracing is less efficient than event recording. Event recording is more about analytics than for observability.
> 
> But we could very easily address the sampling issue by using our existing logger, correct? It wasn't clear from my first review, but the actual code changes I requested were for item b). For item a), what I really need is a coherent strategy for our different logging mechanism. When to pick which emitter (logs, metrics, traces, event log)? How to deal with duplicate events? This feels like the part, which is not explicit here.
> 

To echo @siavashs a little, the way we use the event recorder in HRT is as a separate structured event reporting system. It's read by machines to generate historical information that drives both analytics and user workflows. 

> As far as I can see, we can't really use the EventRecorder for event sourcing. That is fine, but we really need to be clear about what capabilities this feature gives users. Currently, the PR just describes what events are added, but not why.

The _goal_ is to get enough events that we can do event sourcing, but I agree there are some things missing. Over time we can improve this system to export more events and provide more detail. 

This is also why the events sometimes include "extra" data like the peer position - while a perfect event souring system wouldn't need this, we practically:
1. cannot collect every alertmanager state change right now
2. will sometimes drop events (so duplicating information is useful to avoid data loss)
3. don't want to do full event sourcing to answer every question - sometimes including a little extra data in each event makes certain questions answerable from just the single message

(2) is the most important one, I think: In order to do "true" event sourcing to replicate the state of Alertmanager you would need infrastructure which can guarantee that you never drop an event. This is not really practical in reality, so (in my opinion) it's helpful to "duplicate" some information in each event message.

I don't think having extra events breaks the model, really. If you were implementing an event sourcing system you would just ignore the events that don't contain useful data to you. Perhaps in the future we could even support some filtering directly from the Alertmanager configuration.

inhibit/inhibit.go
jackrosenthal: Done.

inhibit/inhibit.go
jackrosenthal: Done.  Moved to `eventrecorder.InhibitRuleAsProto`.

eventlog/file.go
jackrosenthal: Done.  Changed to watch the parent directory with fsnotify, filtering by the target file path.

eventlog/webhook.go
jackrosenthal: Done.  Increased the queue to 1024.  `SendEvent` is now non-blocking and drops the event when the queue is full, incrementing `alertmanager_event_recorder_webhook_drops_total`.

eventrecorder/file.go
jackrosenthal: Done.  Added five tests: `TestFileOutput_SendEvent`, `TestFileOutput_ReopenAfterRename`, `TestFileOutput_ReopenAfterRemove`, `TestFileOutput_Close`, and `TestFileOutput_InvalidPath`.

eventlog/eventlog.go
jackrosenthal: This is no longer applicable.  The concurrency refactor replaced all locks with message passing, so `writeLoop` exclusively owns the outputs and there is no unlock/relock sequence.

eventlog/webhook.go
jackrosenthal: Done.  Added a max backoff cap of 30 seconds and simplified the calculation to `min(retryBackoff<<attempt, maxBackoff)`, removing the `math` import.  Drop metric added as noted above.

eventlog/eventlog.go
jackrosenthal: Fixed by the concurrency refactor.  `writeLoop` handles its own drain in a `defer` when `<-done` fires, so there is no race between workers quitting and events being enqueued.

eventlog/eventlog_test.go
jackrosenthal: Verified manually with a live instance.  Started with file-only config, sent SIGHUP with file+webhook config, and confirmed new events appeared on the webhook while the file continued receiving events.

eventlog/eventlog.go
jackrosenthal: `TestWebhookOutput_CloseFlushesQueue` and `TestWebhookOutput_MultipleWorkers` both verify that `Close` drains all queued events.

eventlog/eventlog.go
jackrosenthal: Added `TestFileOutput_InvalidPath` and `TestWebhookOutput_DropsAfterMaxRetries`.

silence/silence.go
jackrosenthal: Done.  Moved to `eventrecorder.SilenceAsProto` and `eventrecorder.SilenceMatcherAsProto`.

eventlog/eventlog.go
jackrosenthal: Done.  Renamed to `WithEventRecording`.  Recording is disabled by default; callers opt in.

eventlog/eventlog.go
jackrosenthal: Done.

eventlog/eventlog.go
jackrosenthal: Done.

eventlog/eventlog.go
jackrosenthal: No longer applicable.  The concurrency refactor replaced locks with message passing, so there are no locks to defer.

eventlog/eventlog.go
jackrosenthal: No longer applicable.  The concurrency refactor replaced locks with message passing, so there are no locks to defer.

eventlog/eventlog.go
jackrosenthal: Done.  Went with the reflection approach using `WhichOneof` on the proto descriptor.

eventrecorder/webhook.go
jackrosenthal: I agree with @Spaceman1701 that this is fine since the type is in a separate package (`eventrecorder.WebhookOutput` vs `webhook.Notifier`).

jackrosenthal:
Addressed all review feedback since last round:

> I think the package name `eventlog` is too specific
> ...
> `eventlog` is not a very good name

Renamed the entire feature to `eventrecorder`.  Package, config types (`EventRecorderConfig`/`EventRecorderOutput`), YAML key (`event_recorder`), metric names, comments, and error messages all updated.

> b) I don't think we should need any locks in an event-recorder? Can we just use message passing.

Done.  Adopted the message passing approach you suggested.  `writeLoop` now exclusively owns outputs and config.  Config updates arrive via a `cfgUpdate` channel, and `cluster.Peer` uses `atomic.Pointer`.  No mutexes remain.

> I think the biggest takeaway is that we should put this change behind an experimental feature flag now

Done.  Gated behind `--enable-feature=event-recorder`.  When the flag is not set, a zero-value `Recorder` is used and all methods are no-ops.

jackrosenthal:
Addressed feedback from @coderabbitai:

| Comment | Addressed? | Remarks |
|---|---|---|
| `deleteSilenceHandler` missing `WithEventRecording` | Yes | Added context wrapper. |
| Webhook URL should use `SecretURL` type | Yes | Changed to `SecretURL` to prevent credential leakage during serialization. |
| Nil-config guard in `ApplyConfig` | Yes | Added nil check. |
| `HTTPConfig` missing from config comparison | Yes | Added `reflect.DeepEqual` comparison for `HTTPConfig`. |
| Non-deterministic label ordering in proto conversion | Yes | Sorted label pairs and equal-label lists before serialization. |
| `NewFileOutput` swallows watcher startup errors | Yes | Errors are now propagated to the caller. |
| `Close` doesn't wait for `watchLoop` goroutine | Yes | Added `WaitGroup` and post-close guard. |
| Full webhook URL in `Name()`/metrics/logs leaks credentials | Yes | Sanitized URL strips userinfo and query parameters. |
| `SendEvent` returns nil on queue-full drop | No | Intentional fire-and-forget design.  Drops are tracked via the `alertmanager_event_recorder_webhook_drops_total` counter. |
| `Close` can hang behind retry backlog | Yes | Retry backoff now respects shutdown.  `Close` enforces a 30-second drain deadline. |
| Nil recorder guard in `provider/mem` | No | `Recorder` is a value type.  The zero value is nil-safe (`RecordEvent` checks `r.core == nil`). |
| `setSilence` discards `changed` flag from merge | Yes | `setSilence` now returns `(changed, added, err)`.  Broadcasts and event recording are gated on `changed`. |

jackrosenthal:
Addressed feedback from @coderabbitai:

| Comment | Addressed? | Remarks |
|---|---|---|
| Run `gofumpt` on modified files (config.go, eventrecorder.go, silence.go, api.go) | Yes | Ran `gofumpt` on all modified `.go` files. |
| Avoid eager `NotificationEvent` construction when disabled | No | Not a hot path; runs once per successful notification delivery.  The protobuf allocation is negligible. |
| Cover `ApplyConfig` and `SetClusterPeer` in no-op recorder tests | Yes | Added calls in `TestNopRecorderDoesNotPanic` and `TestZeroRecorderDoesNotPanic`. |
| Use `require.JSONEq` for JSON payload comparison in webhook_test.go | Yes | Replaced `require.Equal` with `require.JSONEq`. |
| `SendEvent` returns nil on queue-full drop | No | Intentional fire-and-forget design.  Drops are tracked via the `alertmanager_event_recorder_webhook_drops_total` counter. |
| Protobuf package versioning (`eventrecorder.v1`) | No | Would be a repo-wide change affecting all proto modules (`clusterpb`, `nflogpb`, `silencepb`), not scoped to this PR. |

jackrosenthal:
Addressed feedback from @coderabbitai:

| Comment | Addressed? | Remarks |
|---|---|---|
| Don't tear down live outputs before replacement set is known-good | Yes | `buildOutputs()` is now called first; old outputs are only closed after the new set is fully initialized.  If any new output fails, the existing outputs are kept. |
| Protobuf package name inconsistent with project pattern | Yes | Renamed `package eventrecorder` to `package eventrecorderpb` to match `nflogpb`, `silencepb`, and `clusterpb`. |
| Use the stored `ag.GroupKey()` value instead of rebuilding it from `LabelSet` | No | False positive.  `ExtractGroupKey` returns a `Key` (string alias) built from the real group key string stored in context by `WithGroupKey(ctx, ag.GroupKey())`.  `groupKey.String()` and `groupKey.Hash()` work correctly. |
| `WithRouteLabels` and `WithMutedAlerts` are defined but never populated | Yes | Ported `WithMutedAlerts` from our fork into `MuteStage.Exec`.  Removed `WithRouteLabels`/`RouteLabels` and the `route_labels` proto field, as there is no upstream equivalent of `RouteOpts.Labels`. |

siavashs:
Sorry for the delay in reviews, I was away and kinda busy but I'm doing another review now.
Thank you for addressing the comments, I think this looks much better now.

silence/silence.go
```diff
@@ -274,6 +277,10 @@ func (s *Silencer) Mutes(ctx context.Context, lset model.LabelSet) bool {
 			case SilenceStateActive:
 				activeIDs = append(activeIDs, sil.Id)
 				allIDs = append(allIDs, sil.Id)
+
+				s.recorder.RecordEvent(ctx, eventrecorder.NewSilenceMutedAlertEvent(
+					eventrecorder.SilenceAsProto(sil), fp, lset,
+				))
```
siavashs, line 283: From https://app.devin.ai/review/prometheus/alertmanager/pull/5072
`silence/silence.go:R278-284`

**SilenceMutedAlert events emitted on every Mutes() call, not just state transitions**

In `silence/silence.go:281-292`, a `SilenceMutedAlertEvent` is recorded every time `Silencer.Mutes()` is called and an active silence matches. Since `Mutes()` is called for every alert during every flush cycle of every aggregation group, this can generate a very high volume of events. For example, 1000 silenced alerts with 10 matching silences across 30-second flush cycles would produce ~10,000 events per cycle. The event queue is only 8192 (`eventlog/eventlog.go:44`), so events would be dropped frequently under this load. Similarly, `InhibitionMutedAlertEvent` in `inhibit/inhibit.go:217-228` and `AlertGroupedEvent` in `dispatch/dispatch.go:775` are emitted on every evaluation, not just on state transitions. This design could benefit from deduplication or rate limiting for high-cardinality scenarios.

---
I think this would be an issue for example at Cloudflare since we have a high number of silences.
Should we switch from recording every event to state transition?

siavashs:
This could be a followup but I guess we might want to allow disabling/enabling recoding of specific event types.

siavashs:
I generated some benchmarks to measure the impact of event recorder: https://github.com/siavashs/alertmanager/commit/1ae2e81ce36c39eea49ef6fa13fdcc1d08af8657

It seems inhibitions are 3.3x slower and Silences are 4.4x slower.
When recording is **enabled**, the dominant cost is `protojson.Marshal` + `extractEventType` (proto reflection) + channel send, adding ~8 µs per event in inhibit and ~8 µs per matching silence.

## Inhibit (`Inhibitor.Mutes`)

| Scenario | Recorder | ns/op | B/op | allocs/op |
|---|---|---:|---:|---:|
| 10 rules, 1 alert | **nop_recorder** | 3,199 | 14,940 | 56 |
| 10 rules, 1 alert | **active/disabled** | 3,385 | 14,943 | 56 |
| 10 rules, 1 alert | **active/enabled** | **11,419** | **19,271** | **137** |
| 100 rules, 1000 alerts | **nop_recorder** | 3,178 | 14,928 | 56 |
| 100 rules, 1000 alerts | **active/disabled** | 2,914 | 14,928 | 56 |
| 100 rules, 1000 alerts | **active/enabled** | **10,307** | **19,210** | **137** |

**Inhibit overhead**: ~3.3× slower when recording enabled, +4.3 KB and +81 allocs per call.
NopRecorder and active/disabled are equivalent (context check is ~free).

## Silence (`Silencer.Mutes`)
| Scenario | Recorder | ns/op | B/op | allocs/op |
|---|---|---:|---:|---:|
| 1000 total, 10 matching | **nop_recorder** | 25,218 | 83,765 | 346 |
| 1000 total, 10 matching | **active/disabled** | 23,950 | 83,756 | 346 |
| 1000 total, 10 matching | **active/enabled** | **105,238** | **128,677** | **1,208** |
| 1000 total, 100 matching | **nop_recorder** | 236,421 | 822,465 | 3,138 |
| 1000 total, 100 matching | **active/disabled** | 221,402 | 822,401 | 3,138 |
| 1000 total, 100 matching | **active/enabled** | **1,027,350** | **1,271,330** | **11,756** |

**Silence overhead**: Scales linearly with matching silences since `RecordEvent` is called per match.
With 10 matches: ~4.4× overhead, +45 KB, +862 allocs.
With 100 matches: ~4.6× overhead, +449 KB, +8,618 allocs.
NopRecorder and active/disabled remain identical.

siavashs:
I have a patch which massively improves the performance https://github.com/siavashs/alertmanager/commit/1d10e77b6afd2333bfbd297afdf272fc66572a37

## Inhibit — Before vs After

| Scenario | Recorder | Before (ns/op) | After (ns/op) | **Speedup** | Before allocs | After allocs |
|---|---|---:|---:|---:|---:|---:|
| 10 rules, 1 alert | nop | 3,044 | 3,169 | ~same | 56 | 56 |
| 10 rules, 1 alert | active/disabled | 3,433 | 3,494 | ~same | 56 | 56 |
| 10 rules, 1 alert | **active/enabled** | **10,719** | **5,350** | **2.0×** | 137 | 89 |
| 100 rules, 1000 alerts | **active/enabled** | **9,570** | **4,204** | **2.3×** | 137 | 85 |

## Silence — Before vs After

| Scenario | Recorder | Before (ns/op) | After (ns/op) | **Speedup** | Before allocs | After allocs |
|---|---|---:|---:|---:|---:|---:|
| 1K total, 10 matching | nop | 24,707 | 25,620 | ~same | 346 | 346 |
| 1K total, 10 matching | **active/enabled** | **105,238** | **29,939** | **3.5×** | 1,208 | 540 |
| 1K total, 100 matching | **active/enabled** | **1,027,350** | **294,584** | **3.5×** | 11,756 | 5,020 |
| 10K total, 10 matching | **active/enabled** | **99,664** | **28,686** | **3.5×** | 1,206 | 538 |

## Summary

- **Baseline (nop/disabled) unchanged** — zero overhead when recording is off.
- **Inhibit recording enabled**: **2–2.3× faster**, allocs dropped from 137 → 85–89 per call.
- **Silence recording enabled**: **3.5× faster**, allocs cut by ~55%. The 100-matching-silence case went from **1.03 ms → 295 µs**.
- The overhead ratio vs nop dropped from **~4.5× → ~1.3×** for silence with 10 matches.

The two changes:
1. **Type switch** eliminated proto reflection in `extractEventType` (zero alloc, ~ns)
2. **Deferred marshal** moved `protojson.Marshal` off the hot path — this was the dominant cost (~7–8 µs per event), now absorbed by the background write loop

siavashs:
@jackrosenthal feel free to pull in my benchmarks and the performance fix to this PR.
If you prefer we can also merge it as is and I open a followup PR.

siavashs (approved):
LGTM, the performance overhead can be fixed with the patches mentioned in previous comments.
I leave it to the author to decide to include them now or we apply them later.
Thank you @jackrosenthal

jackrosenthal:
Cherry-picked your performance patch (1d10e77), @siavashs. Thank you! Let us know if you'd prefer it squashed into the main commit instead.

siavashs:
> Cherry-picked your performance patch ([1d10e77](https://github.com/prometheus/alertmanager/commit/1d10e77b6afd2333bfbd297afdf272fc66572a37)), @siavashs. Thank you! Let us know if you'd prefer it squashed into the main commit instead.

We will squash it during the merge process.

PR #5092, opened 2026-03-14
chore: Remove Docker as a build dependency

SoloJacobs:
The advantages and disadvantages of this PR are described in the commit message. The main motivation is the we use `prom crossbuild`, which does not allow using `Docker` during the build step. Because we can't execute the build steps during the cross build, we are forced to check our build artefacts into the source code.

#### Pull Request Checklist

  (for developers nothing changed Edit: the CONTRIBUTING.md needed an update afterall)

```release-notes
NONE
```


SoloJacobs:
What I did not do yet: Add caching for `node_modules` and ensure dependabot actually finds our `packages.json`.

ui/app/Makefile
siavashs: We should clean up the `clean` target too.

siavashs (changes requested):
We need to clean up the `Makefile` (see comment below) and [`ui/app/ CONTRIBUTING.md`](https://github.com/prometheus/alertmanager/blob/main/ui/app/CONTRIBUTING.md#dev-environment-setup) too

ui/app/Makefile
SoloJacobs: Good catch :+1: I did actually look at the `CONTRIBUTING.md`, but only at shell commands :see_no_evil:

ultrotter:
Have we tried podman?
And could we leave the option of containerized build in case people don't want to pull random npms?

SoloJacobs:
@ultrotter It depends on what you mean.

Before `podman` did not work correctly, see this experiment by Christoph: https://gist.github.com/TheMeier/1ea1ff5a9166d4980309b4621ad8f751
I believe this was due to way we used bind mounts, but I'm not sure.

You always had to pull the `npm` dependencies, this was simply done during the docker build step. This now arguably less random with `package-lock.json`. I don't see anyway of avoiding this, unless we want to host build images, which have our specific dependencies installed.

This PR specifically removes `docker`, so we are compatible with `promu`. You can run your build fully containerized with `podman` like so:

```sh
promu crossbuild --podman -p linux/amd64
```

What you can't do is use the created artifacts locally (except for `alertmanager` and `amtool`). This does not matter right now, since `script.js` is checked into the repo. But might matter in the future. Currently, `promu` does the following steps:

- docker create the container
- docker cp your local files into the container (/app/)
- docker start (runs the build, specifically `make build`)
- docker cp only .build/ back out (everything else, including the `node_modules` is lost)
- docker rm

There used to be bind-mounts in `promu`, which would enable the creation of build `artefacts` locally. We could consider adding them back: https://github.com/prometheus/promu/pull/95

ultrotter:
@SoloJacobs What I meant is having an easy target that can build the tools "in a container" with a Dockerfile that pulls the build deps... It could be easy for the target to do 
"container up"
"build"
"run tests"
"copy artifacts"
"container down"

This is pretty much what you described and might not need (write) bind mounts at all...

SoloJacobs:
To move things forward a bit, I would like to add this extra target after the ui changes have landed. I did add the build target you requested here: https://github.com/prometheus/alertmanager/pull/5102

ultrotter (approved):

PR #5093, opened 2026-03-14
feat(template): add dict, map and append functions

TheMeier:
These functions are useful for constructing complex data structures in templates. For example to compose  K/V arrays to use with the existing `toJSON` function. The dict function allows for creating maps from a list of key-value pairs, the list function creates slices from a list of values, and the append function allows for adding elements to a slice.

The names dict and list where chosen instead of map and slice because there is very widspread use of the sprig library in templates, (eg helm) and these names are used in that library.

#### Pull Request Checklist

    - Fixes #<issue number>
    
- Is this a bugfix?
- Is this a new feature?
- Is this a breaking change?

```release-notes

```

## Summary by CodeRabbit

* **New Features**
  * Added three template helpers to build lists, append items, and construct key/value maps directly in templates.

* **Documentation**
  * Documented the new helpers with usage examples and expected error conditions.

* **Tests**
  * Added tests covering list/dict behaviors, nil handling, invalid-argument and non-string-key error cases, list mutation, and composed template scenarios.

### DIFF
--- a/docs/notifications.md
+++ b/docs/notifications.md
@@ -85,9 +85,12 @@ templating.
 
 | Name             | Arguments                  | Description |
 | ---------------- | -------------------------- | ----------- |
+| append           | slice []any, args ...any   | Returns a new slice with the provided arguments appended to the provided slice. |
 | date             | string, time.Time          | Returns the text representation of the time in the specified format. For documentation on formats refer to [pkg.go.dev/time](https://pkg.go.dev/time#pkg-constants). |
+| dict             | values ...any              | Returns a map of string to any, constructed from the variadic list of key-value pairs. The number of arguments must be even, and the keys must be strings. |
 | humanizeDuration | number or string           | Returns a human-readable string representing the duration, and the error if it happened. |
 | join             | sep string, s []string     | [strings.Join](http://golang.org/pkg/strings/#Join), concatenates the elements of s to create a single string. The separator string sep is placed between elements in the resulting string. (note: argument order inverted for easier pipelining in templates.) |
+| list             | ...any                     | Returns the passed arguments as a slice of interfaces. |
 | match            | pattern, string            | [Regexp.MatchString](https://golang.org/pkg/regexp/#MatchString). Match a string using Regexp. |
 | reReplaceAll     | pattern, replacement, text | [Regexp.ReplaceAllString](http://golang.org/pkg/regexp/#Regexp.ReplaceAllString) Regexp substitution, unanchored. |
 | safeHtml         | text string                | [html/template.HTML](https://golang.org/pkg/html/template/#HTML), Marks string as HTML not requiring auto-escaping. |
--- a/template/template.go
+++ b/template/template.go
@@ -17,6 +17,7 @@ import (
 	"bytes"
 	"embed"
 	"encoding/json"
+	"fmt"
 	tmplhtml "html/template"
 	"io"
 	"net/url"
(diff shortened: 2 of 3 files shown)


template/template.go
```diff
@@ -223,6 +224,28 @@ var DefaultFuncs = FuncMap{
 		}
 		return string(bytes), nil
 	},
+	"list": func(args ...any) ([]any, error) {
+		return args, nil
```
siavashs, line 231: We should return an empty slice if args is empty.
Example usage could be `list | toJson`.

```suggestion
		if args == nil {
			return []any{}, nil
		}
		return args, nil
```

docs/notifications.md
```diff
@@ -85,14 +85,17 @@ templating.
...
 | match            | pattern, string            | [Regexp.MatchString](https://golang.org/pkg/regexp/#MatchString). Match a string using Regexp. |
 | reReplaceAll     | pattern, replacement, text | [Regexp.ReplaceAllString](http://golang.org/pkg/regexp/#Regexp.ReplaceAllString) Regexp substitution, unanchored. |
 | safeHtml         | text string                | [html/template.HTML](https://golang.org/pkg/html/template/#HTML), Marks string as HTML not requiring auto-escaping. |
 | safeUrl          | text string                | [html/template.URL](https://golang.org/pkg/html/template/#URL), Marks string as URL not requiring auto-escaping. |
 | since            | time.Time                  | [time.Since](https://pkg.go.dev/time#Since), returns the duration of how much time passed from the provided time till the current system time. |
+| list             | ...any                     | Returns the passed arguments as a slice of interfaces. |
```
siavashs: This is out of alphabetical order.

siavashs (changes requested):

docs/notifications.md
TheMeier: right, missed since I named it `slice` first ...

fixed

template/template.go
TheMeier: changed as suggested

SoloJacobs: Silly question, but is this actually correct? I tried adding the following test without the nil check, and it still passes:

```go
   title: "Template using empty list piped to toJson",                                                                                                                                                                                     
   in:    `{{ list | toJson }}`,                                                                                                                                                                                                           
   exp:   `[]`,                                                                                                                                                                                                                            
```

It is true, that `json.Marshal` produces `null` if passed a `nil` pointer, but the template engine does not  function pass `nil`.

template/template.go
```diff
@@ -223,6 +224,31 @@ var DefaultFuncs = FuncMap{
 		}
 		return string(bytes), nil
 	},
+	"list": func(args ...any) ([]any, error) {
```
SoloJacobs, line 227: Should we simplify the function signature here to ` func(args ...any) []any`?

template/template.go
TheMeier: There is a tiny difference one comperes to nil the other doesen't
see https://go.dev/play/p/q2Ink2VYsNf

[sprig](https://github.com/Masterminds/sprig/blob/master/list.go#L14) doesn't care. But with the extra logic it is more correct.

template/template.go
TheMeier: yes. updated

template/template_test.go
```diff
@@ -313,7 +313,7 @@ func TestTemplateExpansion(t *testing.T) {
 		},
 		{
 			title: "Template with invalid syntax",
-			in:    `{{ `,
+			in:    `{{ foo }} `,
 			fail:  true,
```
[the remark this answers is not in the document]
TheMeier: removed change

template/template.go
SoloJacobs: I think we are all aware how variadic functions behave, but that was not the point:

The template engine does not call `list`  (the go function) with no arguments, when list (the template function) is called with no arguments. This can be verified by removing the check and running the above test. So, the check doesn't change runtime behavior. We can keep it to make the intent explicit and guard against potential edge cases.

PR #5097, opened 2026-03-15
ui: Move useful parts to `httpserver`

SoloJacobs:
Stacks on top of https://github.com/prometheus/alertmanager/pull/5092

We want to allow users to still be able to import parts of the `ui`, just not the ones depending on `js`. Ideas for naming welcome. I also considered changing the name of `Register` to make it a bit more obvious that this function now does something different, but decided to only change the signature in the end, since we don't plan to expose it.

#### Pull Request Checklist

```release-notes
CHANGE: `ui.Register` no longer exposes all endpoints, one must use `weboperations.Register` as well.
```

## Summary by CodeRabbit

* **Refactor**
  * Operational HTTP endpoints (metrics, reload, health/readiness and debug/pprof) are now registered via a dedicated HTTP handler instead of the UI registration.

* **Tests**
  * Added/updated tests to verify debug/pprof routing with and without a route prefix and to reflect the new handler registration.

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -51,6 +51,7 @@ import (
 	"github.com/prometheus/alertmanager/config/receiver"
 	"github.com/prometheus/alertmanager/dispatch"
 	"github.com/prometheus/alertmanager/featurecontrol"
+	"github.com/prometheus/alertmanager/httpserver"
 	"github.com/prometheus/alertmanager/inhibit"
 	"github.com/prometheus/alertmanager/matcher/compat"
 	"github.com/prometheus/alertmanager/nflog"
@@ -582,7 +583,8 @@ func run() int {
 
 	webReload := make(chan chan error)
 
-	ui.Register(router, webReload, logger)
+	ui.Register(router)
+	httpserver.Register(router, webReload)
 
 	mux := api.Register(router, *routePrefix)
 
--- a/httpserver/httpserver.go
+++ b/httpserver/httpserver.go
@@ -1,0 +1,67 @@
+// Copyright The Prometheus Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package httpserver
+
+import (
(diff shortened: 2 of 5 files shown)


cmd/alertmanager/main.go
```diff
@@ -582,7 +583,8 @@ func run() int {
 
 	webReload := make(chan chan error)
 
-	ui.Register(router, webReload, logger)
+	ui.Register(router)
+	weboperations.Register(router, webReload)
```
ultrotter: I would call it httpserver, or apiserver, or something like that? Or webops if you want to be shorter...

But apiserver should make it clear that it's the server exposing our api (via http, implementaton detail)

SoloJacobs: This is the list of endpoints we have in this package:

GET /metrics
POST /-/reload
GET /-/healthy
GET /-/ready
GET /debug/*subpath

The `api.Register` function actually lives in the `api` package, so I find `apiserver` quite confusing. `httpserver` seems ok to. I think we could then consider moving other functionality in `cmd/alertmanager/main.go` into that package. I think the routing logic from `ui` should maybe also live in this package, since then users have an easier time migrating. So, hopefully `httpserver` captures these different parts well enough.

ultrotter:
If you are thinking of moving things from main there, maybe we should call
it the "alertmanager" package... ? :)

Cheers,

Guido

SoloJacobs:
I specifically meant code such as this 
```go
// Make routePrefix default to externalURL path if empty string.
if *routePrefix == "" {
	*routePrefix = amURL.Path
}
*routePrefix = "/" + strings.Trim(*routePrefix, "/")
logger.Debug("route prefix", "routePrefix", *routePrefix)

router := route.New().WithInstrumentation(instrumentHandler)
if *routePrefix != "/" {
	router.Get("/", func(w http.ResponseWriter, r *http.Request) {
		http.Redirect(w, r, *routePrefix, http.StatusFound)
	})
	router = router.WithPrefix(*routePrefix)
}
```
But splitting up the remainder of `cmd/alertmanager` is also a worthwhile endeavor.  :smile: Currently, I don't have a good grasp what our package layout should look like. All I want is to avoid causing pain to those who depend on alertmanager as a library.

httpserver/httpserver.go
```diff
@@ -0,0 +1,67 @@
...
+// Register registers handlers to serve files for automations.
+func Register(r *route.Router, reloadCh chan<- chan error) {
+	r.Get("/metrics", promhttp.Handler().ServeHTTP)
+
+	r.Post("/-/reload", func(w http.ResponseWriter, req *http.Request) {
+		errc := make(chan error)
+		defer close(errc)
+
+		reloadCh <- errc
+		if err := <-errc; err != nil {
```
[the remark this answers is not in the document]
SoloJacobs: There is an open issue for that already, and it is not related to moving this functionality.

Issue #5012, opened 2026-02-13
notify.DeepCopyWithTemplate edge cases


TheMeier:
### What did you do?

Input like `"{}:{alertname=\"test1\"}"` trips up notify.DeepCopyWithTemplate 

### What did you expect to see?

passing `"{}:{alertname=\"test1\"}"` into notify.DeepCopyWithTemplate to just return the string

### What did you see instead? Under which circumstances?

passing `"{}:{alertname=\"test1\"}"` into notify.DeepCopyWithTemplate returns `interface{} nil`

### System information

Irrelevant

### Alertmanager version

```text
current main branch
```

### Alertmanager configuration file

```yaml
Irrelevant
```

### Prometheus version

```text
Irrelevant
```

### Prometheus configuration file

```yaml
Irrelevant
```

### Logs

```text
Irrelevant
```

TheMeier:
It turns out the reason is https://github.com/prometheus/alertmanager/blob/main/template/template.go#L461

passing such a string into yaml.Unmarshal() returns no error and `map[]` this is unexpected. This might be a bug in the yaml library, need to check if the string is spec compliant. Here is a small test with different yaml libraries: https://go.dev/play/p/u_61B-yFQAF
only github.com/goccy/go-yaml reports an error. 

Quick test in python also reports an error:
```
Error: expected '<document start>', but found '<scalar>'
  in "<unicode string>", line 1, column 3:
    {}/{foo="lbar"}:{alertname=" ... 
      ^
```

TheMeier:
Some more tests:
```
            	Error:      	Not equal: 
            	            	expected: string("{env=\"testing\"}:{alertname=\"TestingAlert\", service=\"api\"}")
            	            	actual  : map[string]interface {}(map[string]interface {}{"env=\"testing\"":interface {}(nil)})
```

```
            	Error:      	Not equal: 
            	            	expected: string("{}/{env=\"testing\"}:{alertname=\"TestingAlert\", service=\"api\"}")
            	            	actual  : map[string]interface {}(map[string]interface {}{})
```

TheMeier:
Maybe this is a vaiable pattern? https://github.com/TheMeier/alertmanager/commit/33227826694c797d6d11454702433b1c4b8e2308

PR #5098, opened 2026-03-18
template: preserve quoted numeric strings in DeepCopyWithTemplate

holger-waschke:
This fixes an issue where a quoted numeric string produced by DeepCopyWithTemplate could end up being serialized as an integer instead of a string.

  One concrete case is Jira custom fields that expect a payload like:

  {"id":"15129"}

  Before this change, the rendered field could become:

  {"customfield_11209":{"id":15129}}

  In that form, id was sent as an integer, which Jira rejected.

  With this change, quoted numeric values are preserved as strings, so the generated payload keeps the expected type.

  A regression test was added to cover this case.


TheMeier:
Thank you @holger-waschke. LGTM
The  addtional checks from proposed from rabbitai sound sensible to me, what do you think?

TheMeier:
Loosely related to https://github.com/prometheus/alertmanager/issues/5012

holger-waschke:
> Thank you @holger-waschke. LGTM The addtional checks from proposed from rabbitai sound sensible to me, what do you think?

Looks reasonable to me, I added the unit test to the template_test.go

TheMeier (approved):

themavik (commented):
test review body 12345 unique

themavik (commented):
Using a string type assert after yaml.Unmarshal fixes quoted numeric template output regressing to float recursion in DeepCopyWithTemplate. nit: TrimSpace-only early return means quoted strings with only whitespace still recurse—odd but probably harmless.

holger-waschke:
Updated branch with main, resolving merge conflicts from https://github.com/prometheus/alertmanager/pull/5304. Ready to be merged.

template/template.go
```diff
@@ -652,10 +652,16 @@ func DeepCopyWithTemplate(value any, tmplTextFunc TemplateFunc) (any, error) {
...
+			if err != nil {
 				// ignore error, thus the string is not an interface
 				return parsed, ok
 			}
+			if inlineString, isString := inlineType.(string); isString {
+				return inlineString, ok
+			}
+			if strings.TrimSpace(parsed) == "" {
+				return parsed, ok
+			}
```
siavashs, line 664: I think a better minimal fix would only decode an explicitly rendered JSON string, it must leave any other YAML string results as raw text.

For example we don't want something like:
```yaml
issue # 1234
```
to render to:
```yaml
issue
```


```suggestion
			if inlineString, isString := inlineType.(string); isString {
                // Decode an explicit JSON string, such as output from toJson.
                // Preserve other strings because YAML can remove comments,
                // whitespace, and line breaks from plain scalar values.
    			if json.Valid([]byte(parsed)) {
    				return inlineString, ok
				}
    			return parsed, ok
			}
```

siavashs (changes requested):
I think this PR fixes a narrow issue, but we need more test cases here since this affects 3 integration: Jira, PagerDuty and Webhook:
- a Jira test which uses `toJson` on a numeric string and expects and output like `{"id":"1234"}` in the body
- regression test for ordinary string like `issue # 1234`
- test for existing numeric values being rendered as integers
- test for a JSON object with numeric strings stays as is

SoloJacobs:
@holger-waschke There is a ci failure now.

holger-waschke:
> @holger-waschke There is a ci failure now.

Whoopsie, all green now.

siavashs (approved):

Issue #4727, opened 2025-11-11
George's knowledge share for Alertmanager maintainers


grobinson-grafana:
# Background

This issue contains all the Alertmanager naunces that I've learned over the years and changes I have been thinking about to try and fix some of the problems I have found (sadly I have found many). It is about time I wrote it up to help future maintainers make informed decisions for the project and understand where there be dragons 🐉.

# Gotchas

The following is a list of "gotchas" that I have found while contributing to Alertmanager:

## GroupKey is supposed to be unique but it is not

The GroupKey is a key derived from the matchers in the route (including any parent routes) and any labels in the `group_by` field of the matching route.

There are a number of cases where GroupKey is not unique such that it is possible for two or more aggregation groups to share the same GroupKey at the same time.

This creates a number of problems:

1. If the receiver is also the same, two or more aggregation groups can share the same notification log.
2. If we want to use GroupKey in the "GroupMarker" then two aggregation groups with conflicted GroupKeys will overwrite each other's markers.

A simple solution to this problem is to add names for routes, and each name must be unique at it's level in the route tree.

You can find the full issue [here](https://github.com/prometheus/alertmanager/issues/3817) and [here](https://github.com/prometheus/alertmanager/issues/3808).

## High availability requires dispatcher timers to be in sync in work properly

Alertmanager organizes replicas into numbered positions (starting from 0), and uses a timeout called the peer timeout to coordinate failover. On flush, each replica waits `position * peer timeout` seconds. When its timer elapses it checks if a lower numbered replica has sent a notification (by checking its local notification log for gossiped entries), and if no notification has been sent, tries to send the notification itself.

The really important point here is that this behavior requires all replicas to flush at the same time, otherwise the failover mechanism breaks completely. In the worst case, flushes can be skewed such that each replica finishes their independent wait stage at exactly the same time, duplicating every notification exactly replica times.

You must avoid any changes that can contribute to timer skew.

## Immediate flushes cause timer skew

There is an unfortunate piece of code (in versions up to 0.29.0 at the time of writing this) that causes immediate flushes to occur. This, ironically, leads to timer skew and duplicate notifications.

Two cases where this happens is when a highly available Alertmanager deployment is restarted at staggered intervals and when the configuration is reloaded at staggered intervals.

I have a tentative fix for staggered restarts [here](https://github.com/prometheus/alertmanager/pull/3419). Unfortunately, this does not prevent the issue when reloading the configuration at staggered intervals.

[...]

PR #5152, opened 2026-04-03
feat(api): add receiver labels and receiver_matchers filter

### LINKED ISSUE #4963: Add receiver labels

cxdy:
Add support for arbitrary key-value labels on receivers, enabling structured identification for querying and filtering. The receiver name is auto-injected as the "name" label for uniform matcher access.

Expose labels in all API receiver responses (GET /receivers, GET /alerts, GET /alerts/groups) and add a receiver_matchers query parameter to all three endpoints for filtering by receiver labels using PromQL-style matchers.

#### Pull Request Checklist

    - Fixes #4963 
- Is this a new feature?

- Adds `labels` to `receivers`
- Returns `receiver.labels` in `/api/v2/alerts`, `/api/v2/alerts/groups`, and `/api/v2/receivers`

#### Good config + response
```yaml
receivers:
- name: this-is-my-receiver-name
  labels:
    owning-team: team-foo-bar
    kind: webhook
    name: this-is-my-receiver-name
  webhook_configs:
  - send_resolved: true
    url: http://127.0.0.1:9014/alerts
```
```bash
root@prometheus:~# curl -s 'localhost:9093/api/v2/receivers' | jq '.[] | {name: .name, labels: .labels}'
{
  "name": "this-is-my-receiver-name",
  "labels": {
    "kind": "webhook",
    "name": "this-is-my-receiver-name",
    "owning-team": "team-foo-bar"
  }
}
{
  "name": "slack-alert-dev",
  "labels": {
    "name": "slack-alert-dev"
  }
}
```
#### Bad config + error
```yaml
receivers:
- name: this-is-my-receiver-name
  labels:
    owning-team: team-foo-bar
    kind: webhook
    name: this-is-not-my-receiver-name
  webhook_configs:
  - send_resolved: true
    url: http://127.0.0.1:9014/alerts
```
```console
time=2026-04-03T12:59:36.559Z level=ERROR source=coordinator.go:116 msg="Loading configuration file failed" component=configuration file=/path/to/alertmanager.yml err="receiver label \"name\" must match receiver name \"this-is-my-receiver\", got \"this-is-not-my-receiver-name\""
```

I'm sort of torn on the error handling though, I currently have it failing hard (seems like that's the general convention in Alertmanager for this sort of thing) if `receiver[i].name != receiver[i].labels.name`, but that seems a little dramatic for what this is. Open to feedback on that one, happy to change it if you agree with me or leave it if you don't! 

## Summary by CodeRabbit

* **New Features**
  * Added receiver_matchers query param to filter receivers, alerts, and alert groups by receiver labels; alerts/alert-groups now respect receiver-label gating.

* **API Changes**
  * Receiver objects include a labels field (auto-injected name label).
  * GET /receivers documents 400/500 responses for invalid matcher input.
  * Client libs and URL builders support multi-valued receiver_matchers.

* **Config**
  * Receiver configs auto-populate/validate the name label.

* **Tests**
  * Added coverage for matcher logic, label propagation, and config label handling.

* **UI**
  * Frontend model encodes/decodes receiver.labels.


Spaceman1701:
This looks awesome, thanks for the contribution!

I am away for a little while, but I will do a code review as soon as I'm able. Perhaps the other maintainers will get to to first 🙂

siavashs (approved):
LGTM, Thank you for your contribution.

siavashs (changes requested):
There is a frontend test error:
```elm
Compiling (80)-- TYPE MISMATCH ------------------------------- src/Views/AlertList/Updates.elm

The 2nd argument to `AlertGroup` is not what I expect:

47|                                                 AlertGroup (Dict.fromList labels) { name = "unknown", labels = Dict.empty } alerts_
                                                                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
This argument is a record of type:

    { labels : Dict.Dict k v, name : String }

But `AlertGroup` needs the 2nd argument to be:

    Data.Receiver.Receiver
```
See the CI job for details.

cxdy:
Ah thanks @siavashs! I got it passing, just rebased and letting it run again.

api/v2/openapi.yaml
```diff
@@ -35,13 +35,26 @@ paths:
...
         '200':
           description: Get receivers response
           schema:
             type: array
             items:
               $ref: '#/definitions/receiver'
+        '400':
+          $ref: '#/responses/BadRequest'
+        '500':
+          $ref: '#/responses/InternalServerError'
```
SoloJacobs: Why is there a 500 error code here? Does this implemenation make it so the receiver endpoint no longer works correctly in all circumstances?

api/v2/openapi.yaml
```diff
@@ -35,13 +35,26 @@ paths:
         - receiver
       operationId: getReceivers
       description: Get list of all receivers (name of notification integrations)
+      parameters:
```
SoloJacobs, line 38: It seems like `receiver_matchers` should not be duplicated? Can't we use a `$ref` here?

config/config.go
```diff
@@ -926,6 +926,9 @@ type Receiver struct {
 	// A unique identifier for this receiver.
 	Name string `yaml:"name" json:"name"`
 
+	// Labels attached to this receiver for querying and filtering.
+	Labels map[string]string `yaml:"labels,omitempty" json:"labels,omitempty"`
```
SoloJacobs, line 942: This needs to be documented in `docs/configuration.md`.

SoloJacobs (changes requested):
Two issues we should discuss:

1) The name auto-injection makes every user opt into this feature silently on upgrade. Since receiver names are already unique identifiers, re-expressing them as a label feels like a category error. I'd like to discuss whether this is the right default before it ships.

2) Returning labels on every `GettableAlert` duplicates static receiver configuration across every alert in the response. There are also an open proposal to remove the receiver field from /alerts altogether https://github.com/prometheus/alertmanager/issues/4727 . Expanding it now moves in the opposite direction. The right place for receiver metadata is GET /receivers. Callers who need the mapping can join on receiver.name client-side.

Suggested direction: keep labels and `receiver_matchers` on GET /receivers only, and drop them from the alert/alert-group responses.

Spaceman1701 (commented):
Sorry for the long delay! Besides the couple in-line comments that @SoloJacobs left, the code looks good to me!

In the meantime, I'm discussing with the other maintainers a little bit about the label injection and the API response types

Spaceman1701:
Hi! Sorry again for the delay! After some discussion, 

- We've agreed that we want to keep the injection of `name` into the receiver labels. The reasoning behind this is that queries that use matchers will often want to match of the receiver name. Without injecting the `name` into the receiver labels, clients would need to query across both the "special" name query parameter and the labels parameter. 
- There's sufficient concern about the increased size of `/alerts` and `/alerts/groups` that we want to avoid returning the entire receiver object for every alert or group. Given that receivers are frequently repeated in these responses, we'd expect a lot of repetition in the labelsets. Our suggestion is to add a `receiverReference` type to the OpenAPI spec which is exactly the same as the `receiver` type was before this PR. `/alerts` and `/alerts/groups` can include the `recieverReference` while `/receivers` can return the full object with the new `labels`.

What do you think of this plan?

api/v2/openapi.yaml
SoloJacobs: I removed this error code now.

SoloJacobs:
@Spaceman1701 I added a fix-up commit here, to address the findings I had. I didn't have time to play around with a gain **after** I made my changes.

Spaceman1701 (approved):
LGTM

TheMeier (approved):
LGTM.

Things i noticed: no UI features added to interact with the new labels, that can be added later. The tests could possibly be simplyfied using testify. But config_test.go is a mixed bag of using testify in some but not all places atm. We might want to clean that up some day.

PR #5172, opened 2026-04-12
docs: clarify weekday range behavior and week start day

### LINKED ISSUE #3213: Document first day of week for time intervals

vincent067:
Clarify that Alertmanager follows Go's time.Weekday convention where Sunday = 0 and Saturday = 6. This differs from ISO 8601 where Monday is the first day of the week.

Also explain that weekday ranges do not wrap around the week boundary, so `saturday:sunday` will not match any day and should be written as `['saturday', 'sunday']` instead.

Fixes #3213

## Summary by CodeRabbit

* **Documentation**
  * Updated `weekday_range` specification documentation with clarification on weekday numbering system (Sunday=0, Saturday=6), explanation of non-wrapping range behavior across week boundaries, and practical guidance for configuring day combinations correctly.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -397,7 +397,13 @@ times:
 `weekday_range`: A list of days of the week, where the week begins on Sunday and ends on Saturday.
 Days should be specified by name (e.g. 'Sunday'). For convenience, ranges are also accepted
 of the form `<start_day>:<end_day>` and are inclusive on both ends. For example:
-`['monday:wednesday','saturday', 'sunday']`
+`['monday:wednesday','saturday', 'sunday']`.
+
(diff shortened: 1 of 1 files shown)


TheMeier (approved):

SoloJacobs (approved):

SoloJacobs:
Hi @vincent067 ,

could you please sign-off on your commit? See the [contributing guidelines.](https://github.com/prometheus/prometheus/blob/main/CONTRIBUTING.md)

Thank you

SoloJacobs (changes requested):
DCO is mandatory here. I approved this by accident, when leaving the original comment.

SoloJacobs:
Hi @vincent067 ,

we haven't heard back from you here in a while, so I will close this PR. Feel free to open a new PR.

Kind regards

PR #5174, opened 2026-04-15
inhibit: fix gcCallback removing live index entry for regex source matchers

### LINKED ISSUE #5162: inhibitions unable to handle regex
### What did you do?

We recently upgraded alertmanager from version 0.28.0 to 0.31.0.
Since upgrading, our inhibition rules for alerts have failed, and we're seeing firing alerts matching the target_matcher, despite source_matchers also firing.

_No response_

Alerts were not inhibited. This was confirmed by curling alertmanager directly when the alert fired. I could see that both the target and source alerts were firing, but the target had no inhibitors.

This was resolved by separating our source_matchers:
previously 
```
    source_matchers:
      - alertname =~ "(alert_one|alert_two|alert_three)"
```
into three separate rules, each with one alert as the source matcher. It is often the case that multiple of these alerts are firing at once.

We also removed some regex from our target_matchers:
```
  - target_matchers:
      - alertname =~ "(alert_four|alert_five|alert_six_.*)"
```
We removed the final `alert_six._*` from the target_matchers.

### System information

Linux 5.15.0-1084-aws x86_64

### Alertmanager version

```text
alertmanager-0.31.0.linux-amd64
```

```yaml

```

### Prometheus version

```text

```

```yaml

```

### Logs

```text

```

alliasgher:
## Root cause

Since the `sindex` performance optimisation landed (v0.28.1 #4119), `gcCallback` unconditionally removes the `sindex` entry for an equal-label fingerprint whenever **any** source alert with that fingerprint is GC'd:

```go
// before
func (r *InhibitRule) gcCallback(alerts []*types.Alert) {
    for _, a := range alerts {
        fp := r.fingerprintEquals(a.Labels)
        r.sindex.Delete(fp)   // always removes, even if a different alert is indexed
    }
}
```

When a **regex** `source_matchers` (e.g. `alertname =~ "(alert_one|alert_two)"`) matches multiple source alerts that share the same equal-label fingerprint, the sindex holds only one of them. If that alert is **not** the indexed one, the delete removes the wrong entry. Worse, even if it is the indexed one, no replacement is promoted — so the remaining active source alert is orphaned in the scache and inhibition silently stops working.

This matches the exact symptom in #5162: splitting the regex into separate per-alertname rules (one alert per rule, no collision) worked around the problem.

## Fix

In `gcCallback`, only remove the index entry when the GC'd alert **is** the indexed one, then scan the source cache for another still-active alert with the same equal-label fingerprint to promote as its replacement:

```go
func (r *InhibitRule) gcCallback(alerts []*types.Alert) {
    for _, a := range alerts {
        eq := r.fingerprintEquals(a.Labels)
        indexed, ok := r.sindex.Get(eq)
        if !ok || indexed != a.Fingerprint() {
            continue
        }
        r.sindex.Delete(eq)
        for _, candidate := range r.scache.List() {
            if r.fingerprintEquals(candidate.Labels) == eq {
                r.sindex.Set(eq, candidate.Fingerprint())
                break
            }
        }
    }
}
```

The scan is O(n) in the source-alert cache size, but GC events are infrequent and the scache only holds alerts that matched the source matchers.

## Test

`TestInhibitRule_gcCallback_preserves_regex_source_match` — fires two regex-matching source alerts with the same `env` equal label, GC's the first, and asserts the target is still inhibited by the second.

Fixes #5162

### DIFF
--- a/inhibit/inhibit.go
+++ b/inhibit/inhibit.go
@@ -390,8 +390,27 @@ func (r *InhibitRule) findEqualSourceAlert(lset model.LabelSet, now time.Time) (
 
 func (r *InhibitRule) gcCallback(alerts []*types.Alert) {
 	for _, a := range alerts {
-		fp := r.fingerprintEquals(a.Labels)
-		r.sindex.Delete(fp)
+		eq := r.fingerprintEquals(a.Labels)
+
+		// Only act on the index entry if the GC'd alert is the one currently indexed.
+		// When multiple source alerts share the same equal-label fingerprint (e.g. via
+		// a regex source_matchers like alertname =~ "(a|b|c)"), the index holds only one
+		// of them. If a different alert is indexed, removing the entry would break
+		// inhibition for that other alert.
+		indexed, ok := r.sindex.Get(eq)
+		if !ok || indexed != a.Fingerprint() {
+			continue
+		}
+
+		// The GC'd alert was the indexed one. Remove it and promote another still-active
+		// source alert with the same equal labels, if any.
+		r.sindex.Delete(eq)
+		for _, candidate := range r.scache.List() {
+			if r.fingerprintEquals(candidate.Labels) == eq {
+				r.sindex.Set(eq, candidate.Fingerprint())
+				break
+			}
+		}
 	}
 }
 
--- a/inhibit/inhibit_test.go
+++ b/inhibit/inhibit_test.go
@@ -620,3 +620,59 @@ func BenchmarkFingerprintEquals(b *testing.B) {
 		})
 	}
 }
+
+// TestInhibitRule_gcCallback_preserves_regex_source_match is a regression test for
+// the case where multiple source alerts match via a regex matcher and share the same
+// equal-label fingerprint. When one of them is GC'd, inhibition for the remaining
+// active source alert must continue to work.
+func TestInhibitRule_gcCallback_preserves_regex_source_match(t *testing.T) {
+	// Build an InhibitRule with a regex source matcher (alertname =~ "src.*") and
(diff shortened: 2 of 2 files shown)


inhibit/inhibit.go
```diff
@@ -390,8 +390,27 @@ func (r *InhibitRule) findEqualSourceAlert(lset model.LabelSet, now time.Time) (
...
+		// inhibition for that other alert.
+		indexed, ok := r.sindex.Get(eq)
+		if !ok || indexed != a.Fingerprint() {
+			continue
+		}
+
+		// The GC'd alert was the indexed one. Remove it and promote another still-active
+		// source alert with the same equal labels, if any.
+		r.sindex.Delete(eq)
+		for _, candidate := range r.scache.List() {
```
TheMeier, line 408: can this potentially become a new performance issue with very large list of matching alerts?

siavashs: I think so, see my comment on https://github.com/prometheus/alertmanager/pull/5449#discussion_r3879887308

SoloJacobs:
Hi @alliasgher ,

We haven't heard back from you on this PR, and it is now outdated: https://github.com/prometheus/alertmanager/pull/5542 has been landed, and addresses the regressions fixed by this PR. 

Just a personal POV:  I think this PR should have been merged much sooner, I simply didn't have time yet to review all of our PRs. It is just difficult to land changes to the inhibitor, without impacting the performance. So, thank you for providing this PR. It also informed the fixes we eventually landed.

Fix will also be part of a patch release: https://github.com/prometheus/alertmanager/pull/5556

Kind regards

Issue #5173, opened 2026-04-14
build v0.32.0 failed on arm64 native


vndroid:
### What did you do?

buildx failed on Arm64, It seems like cause by #5113 .

```dockerfile
FROM golang:1.25-alpine3.23 AS builder
ENV VERSION=0.32.0

RUN set -eux \
    && apk add --no-cache git make curl vim bash npm \
    && addgroup -g 1001 gorelease \
    && adduser -D -u 1001 -G gorelease gorelease

USER 1001:1001

WORKDIR /go/src/github.com/prometheus/alertmanager

RUN set -eux \
    && git clone -b v${VERSION} --single-branch --depth=1 https://github.com/prometheus/alertmanager.git . \
    && make build
```

### What did you see instead? Under which circumstances?

```
784 silly packumentCache corgi:https://registry.npmjs.org/vite-plugin-elm cache-miss
785 verbose stack Error: command failed
785 verbose stack     at promiseSpawn (/usr/local/lib/node_modules/npm/node_modules/@npmcli/promise-spawn/lib/index.js:22:22)
785 verbose stack     at spawnWithShell (/usr/local/lib/node_modules/npm/node_modules/@npmcli/promise-spawn/lib/index.js:124:10)
785 verbose stack     at promiseSpawn (/usr/local/lib/node_modules/npm/node_modules/@npmcli/promise-spawn/lib/index.js:12:12)
785 verbose stack     at runScriptPkg (/usr/local/lib/node_modules/npm/node_modules/@npmcli/run-script/lib/run-script-pkg.js:79:13)
785 verbose stack     at runScript (/usr/local/lib/node_modules/npm/node_modules/@npmcli/run-script/lib/run-script.js:9:12)
785 verbose stack     at /usr/local/lib/node_modules/npm/node_modules/@npmcli/arborist/lib/arborist/rebuild.js:327:17
785 verbose stack     at run (/usr/local/lib/node_modules/npm/node_modules/promise-call-limit/dist/commonjs/index.js:67:22)
785 verbose stack     at /usr/local/lib/node_modules/npm/node_modules/promise-call-limit/dist/commonjs/index.js:84:9
785 verbose stack     at new Promise (<anonymous>)
785 verbose stack     at callLimit (/usr/local/lib/node_modules/npm/node_modules/promise-call-limit/dist/commonjs/index.js:35:69)
786 verbose pkgid elm@0.19.1
787 error code 1
[...]

PR #5191, opened 2026-04-25
fix(ui): Provide prebuilt UI assets in release

### LINKED ISSUE #5163: Include prebuilt UI assets (or tarball) in release artifacts
### Proposal

Since 0.32.0 (as per #5113), compiled UI assets are no longer included in the repository. This means downstream builds now need to build the UI before building the Go binary.

As the downstream maintainer for Fedora & EPEL, this introduces a significant complication in the packaging process. Our previous spec file worked without modification, but now fails with:

    ui/web.go:31:12: pattern app/dist: no matching files found

This change requires adding a full frontend (Elm + tooling) build pipeline into the RPM build, which is considerably more complex and less reproducible in our environment. In addition, such toolchains are often problematic in distribution build systems due to policy and network restrictions.

For comparison, Prometheus ships prebuilt UI assets as part of its release process (`prometheus-web-ui-$version.tar.gz`). This allows downstream packagers to:
- Avoid rebuilding frontend assets
- Keep builds simpler and more reproducible
- Only declare bundled JS dependencies instead of bootstrapping a full toolchain

Would it be possible to publish a tarball with prebuilt UI assets as part of Alertmanager releases (or otherwise include them in release artifacts)?

This would greatly simplify downstream packaging while preserving the current development workflow.

SoloJacobs:
In https://github.com/prometheus/alertmanager/pull/5113 I removed the build artifacts from our source repo. This complicated the packaging process for downstream maintainers significantly.

We will still not check the artifacts into our repo directly, since this is a burden to the review process and opens us up to attacks such as the https://en.wikipedia.org/wiki/XZ_Utils_backdoor .

Instead we provide two new mechanisms to obtain the artifacts:

* alertmanager-web-ui-(file <VERSION).tar.gz will provide the `dist` folder needed for calling `make build`.
* We will upload artifacts via actions/upload-artifact. These can then be easily retrieved for 90 days using `gh`.

Finally, this commit also ensures that `ui/app/dist` is always available during build steps. The reason is that caching `npm` dependencies is not sufficient: `Elm` will download its files again anyway, since it stores them in `~/.elm`. This causes unnecessary network calls.

Technical approach:

* By moving the validation `.build_stamp` to `dist/`, we don't have to upload that file seperately. `embed` will exclude that file.
* Storing artifacts in `.tarballs` is our method for including files in the release. This approach was directly taken from `prometheus/prometheus`.

#### Pull Request Checklist
    - Relates-to: #5173 Fixes: #5163
- Is this a bugfix?
    (Only a manual test on my end).

```release-notes
FIX: Provide prebuilt UI assets in release via alertmanager-web-ui-(file <VERSION).tar.gz
```

## Summary by CodeRabbit

* **Chores**
  * CI now builds and uploads a reusable frontend distribution artifact used by test, build, and release jobs instead of relying on npm cache.
  * Release workflow downloads and reuses the frontend artifact prior to packaging.
  * Added a Makefile target to produce versioned frontend tarballs for distribution.
  * Frontend build recipes and clean steps updated to track builds reliably and avoid stale artifacts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,15 +22,29 @@ jobs:
       - run: make assets
       - run: make apiv2
       - run: git diff --exit-code
+      - run: make assets-tarball
+      - uses: prometheus/promci-artifacts/save@f9a587dbc0b2c78a0c54f8ad1cde71ea29a4b76f # v0.1.0
+        with:
+          directory: .tarballs
+      - uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v4.6.2
+        with:
+          name: ui-dist
+          path: ui/app/dist
+          include-hidden-files: true
 
   build:
     name: Build Alertmanager for common architectures
+    needs: [test_frontend]
     runs-on: ubuntu-latest
     strategy:
       matrix:
         thread: [0, 1, 2]
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v4.0.0
+      - uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16 # v4.1.8
+        with:
+          name: ui-dist
+          path: ui/app/dist
       - uses: prometheus/promci/build@42c3c84c865e5c1ab78543b929f7341eb2ef6123 # v0.6.1
         with:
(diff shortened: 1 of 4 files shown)


SoloJacobs:
Note: I plan to port this into 0.32 branch .

Spaceman1701 (approved):
Lgtm

PR #5222, opened 2026-05-01
cmd/alertmanager: add --config.auto-reload-interval flag

### LINKED ISSUE #5197: Feature request: native config auto-reload support, similar to Prometheus `auto-reload-config`
### Proposal

Alertmanager supports runtime reload via `SIGHUP` or `POST /-/reload`, but it does not appear to support automatic reload when the config file is updated on disk.

This is particularly inconvenient in Kubernetes environments where Alertmanager configuration is managed via ConfigMap or Secret. Even when a mounted file is updated automatically, Alertmanager does not pick up the change on its own, so users still need an external reloader sidecar just to detect the file change and call `/-/reload`.

It would be very helpful if Alertmanager had a native auto-reload mechanism similar to Prometheus `--enable-feature=auto-reload-config`.

Reference: [https://prometheus.io/docs/prometheus/latest/feature_flags/#auto-reload-config](https://prometheus.io/docs/prometheus/latest/feature_flags/#auto-reload-config)

### Current behavior

1. Start Alertmanager with a config file:

```bash
alertmanager --config.file=/etc/alertmanager/alertmanager.yml --web.listen-address=:9093
```

2. Update `/etc/alertmanager/alertmanager.yml` on disk.
3. Alertmanager does not reload the updated config automatically.
4. Manually trigger a reload:

```bash
curl -X POST http://localhost:9093/-/reload
```

5. Only after step 4 is the updated config applied.

### Expected behavior

Alertmanager optionally supports automatic reload when `--config.file` changes on disk, similar to Prometheus `--enable-feature=auto-reload-config`.

This would be especially useful for Kubernetes deployments where configuration is managed via ConfigMap or Secret and updated through mounted files.

### Environment

* Alertmanager version: v0.32.0

mihir-dixit2k27:
## Description

Adds a `--config.auto-reload-interval` flag. When set to a non-zero duration, a background goroutine polls the SHA256 checksum of the config file at the given interval and writes to the existing `webReload` channel on a detected change — the same path taken by `SIGHUP` and `POST /-/reload` — so no new reload logic is introduced.

**Why polling instead of fsnotify?**
Kubernetes `kubelet` uses `AtomicWriter` to update ConfigMap and Secret volume mounts via symlink swaps, which causes fsnotify to stop receiving events after the first update since the original inode is replaced. SHA256 polling works correctly regardless of how the underlying file is updated, which is also why Prometheus uses polling for its equivalent feature.

## Behaviour

- Default is `0s` (disabled). Existing deployments are completely unaffected.
- On a detected change, the reload follows the same code path as `SIGHUP`/`POST /-/reload`: invalid configs are rejected and logged without disrupting the running instance.
- If a reload fails, `lastChecksum` is not updated, so the watcher retries on every subsequent tick until the configuration is valid again.
- The watcher goroutine exits cleanly on `SIGTERM`.

## Changes

| File | Description |
|---|---|
| `cmd/alertmanager/main.go` | New flag, `configFileChecksum()` helper, `runConfigWatcher()` goroutine, watcher start block in `run()` |
| `cmd/alertmanager/main_test.go` | 8 unit tests covering all edge cases |
| `docs/configuration.md` | New auto-reload section |
| `docs/management_api.md` | Cross-reference added under the reload section |

fixes #5197

### DIFF
--- a/cmd/alertmanager/main.go
+++ b/cmd/alertmanager/main.go
@@ -15,12 +15,16 @@ package main
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"fmt"
+	"log/slog"
 	"os"
 	"os/signal"
 	"runtime"
 	"strings"
 	"syscall"
+	"time"
 
 	"github.com/KimMachineGun/automemlimit/memlimit"
 	"github.com/alecthomas/kingpin/v2"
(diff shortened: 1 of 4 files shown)


TheMeier:
Had a quick look, looks ok so far. But why is this better then using prometheus-config-reloader, which is re-usable and in fact used by many projects.

mihir-dixit2k27:
@themeier
The main motivation is avoiding an extra sidecar just for this `prometheus-config-reloader` needs its own image, RBAC, resource requests etc. which is a lot of overhead for simply watching one file.

Also Prometheus also added `--enable-feature=auto-reload-config` natively even though` prometheus-config-reloader `existed, so there's precedent for this in the ecosystem.

Happy to add a docs note pointing to `prometheus-config-reloader` for more advanced use cases if that helps!

Spaceman1701 (commented):
Thanks for the contribution! I've looked briefly at the code and it looks ok to me, but unfortunately I'm a little concerned about this concept.

In cluster mode, Alertmanager nodes need to agree on their config or they will not reach consensus. Decoupling the replacement of the config file and the reload helps ensure that users can coordinate properly - it's not often possible to atomically update a file across many hosts, but it is easy enough to concurrently issue several reload requests. 

I worry that this flag would end up being a footgun for clustered Alertmanager. If enabled, coordinating config reloading becomes very difficult.

mihir-dixit2k27:
@Spaceman1701 That's a really good point, I hadn't thought about it from the cluster coordination angle.
Would it make sense to either add a warning when cluster mode is detected, or explicitly document that this flag is intended for single-node deployments? Happy to go whichever direction you think is right . or if you feel it's a fundamental blocker I'm open to that too.

ultrotter:
We could definitely go with a warning, or even refusing the flag for cluster nodes. What do you think @Spaceman1701 ?

jshah-dev:
Hello,
Are there any plans to include this in an upcoming release? 
This looks like a valuable enhancement that I feel would be very useful and I am looking for this feature in alertmanager for my project
It would be great to see it become part of the official project, as it addresses a genuine use case & would help avoid maintaining custom forks
Looking forward to seeing this move forward if the maintainers feel it's a good fit.

jshah-dev:
Hello ,
Just following up on this PR, as I’m still very interested in seeing this feature move forward.
It addresses a genuine use case for my project, and having it upstream would help avoid maintaining a custom fork or workaround.

SoloJacobs:
Hi @mihir-dixit2k27 , hi @jshah-dev ,

We discussed this feature in the Alertmanager Working Group today, and I'm afraid we cannot take this feature. These are the reasons:

* The reload is a very operationally sensitive operation. Specifically, it is the source of many incidents. It produces significant load on the system.
* The reload needs to be coordinated between different instances of Alertmanager.

Thus, while this feature makes sense in theory, current Alertmanager deployments can just not run this option safely. Of course, once those operational hurdles are no longer an issue, we can reconsider the feature. But that is a multi-year project at this state.

Maybe you have better luck at https://github.com/grafana/prometheus-alertmanager 

I will leave the issue open, so the community has a place to comment on this feature. Maybe it will be re-discussed in the future.

Kind regards

Issue #1996, opened 2019-08-08
Support for sending notification to kafka


jyotiverma03:
**What did you do?**
Trying to send notifications to Kafka from alertmanager

**What did you expect to see?**
Configuration for sending notifications to Kafka along with email and slack

* Alertmanager version:

	 `v0.18.0`

* Prometheus version:

	`v2.11.1` 

Is there any way we can push notifications to Kafka? If yes then how can we do it?

simonpasquier:
We're not adding any more receiver to AlertManager (see [doc](https://prometheus.io/docs/alerting/configuration/#receiver)) so it has to be an external service receiving notifications from AlertManager and forwarding them to Kafka. Unfortunately https://prometheus.io/docs/operating/integrations/#alertmanager-webhook-receiver doesn't list any Kafka integration.

I'm closing it for now. If you have further questions, please use our [user mailing list](https://groups.google.com/forum/#!forum/prometheus-users), which you can also search.

Xtigyro:
@simonpasquier @jyotiverma03 Can we re-open this please?

gotjosh:
Absolutely, let's revisit this.

simonpasquier:
Agree that it sounds useful in general. But details would need to be fleshed out :)

hyingreborn:
Kafka, as a component for processing messages among multiple services, has become the de facto standard. Moreover, when I use Alertmanager, in order to avoid the concurrency problem caused by a large number of messages (when thousands of targets are offline at the same time), I need to forward them to Kafka for subsequent processing every time.

It's hard to imagine that this problem has been going on for six years. ...

siavashs:
I will implement this.

PR #5246, opened 2026-05-19
feat(eventrecorder): add Kafka output

siavashs:
Add a third event recorder destination alongside file and webhook that
produces serialized events to a Kafka topic via **franz-go**.

The event recorder configuration is grouped by output type (receiver-style),
one list per destination kind:
```yaml
  event_recorder:
    file_outputs:
      - path: /var/log/alertmanager/events.jsonl
    webhook_outputs:
      - url: https://example.com/hook
    kafka_outputs:
      - brokers: ["kafka-1:9093", "kafka-2:9093"]
        topic: alertmanager.events
        format: json          # or "protobuf"
        acks: leader          # "none" | "leader" | "all"
        compression: snappy   # "none" | "gzip" | "snappy" | "lz4" | "zstd"
        buffer_size: 1024
        tls_config: { ... }
```

Implementation notes:

- `KafkaOutput` buffers events in a bounded local channel and forwards them to franz-go's async producer via `TryProduce`, which never blocks. When either the local channel or franz-go's internal buffer is full (e.g. a slow or unreachable broker), events are dropped (counted via `alertmanager_event_output_drops_total`) so the upstream pipeline and shutdown are never stalled.
- Broker unreachability at startup is logged at warn level and does not prevent Alertmanager from starting; the connectivity check runs in the background and franz-go retries connections.
- Records use the producing instance's hostname as the message key, keeping per-instance ordering on the same partition.
- The `Destination` interface is unified: each destination receives the structured event and owns its serialization (file/webhook encode JSON via protojson, kafka encodes JSON or binary protobuf per config). This removes the footgun of a protobuf-configured output silently shipping JSON.
- The transport-layer building blocks (config validation, franz-go client/option construction, background ping, error classification, broker-list helpers) live in a new shared `kafka` package so a future Kafka receiver (#1996) can reuse them.
- TLS is supported via prometheus/common's TLSConfig (mTLS or server-only). SASL is intentionally out of scope for this change and can be added later via franz-go's kgo.SASL options.
- Idempotent writes are disabled unless acks=all is explicitly set, to keep the default leader-ack path compatible with franz-go.

Metric changes:

- Rename `alertmanager_event_webhook_drops_total` -> `alertmanager_event_output_drops_total{output}`, shared by webhook and kafka outputs. This is a breaking metric rename; dashboards and alerts referencing the old name need to be updated.
- Add `alertmanager_event_kafka_produce_errors_total{output,error_type}` populated from franz-go's produce callback.

Testing:

- Event recorder unit tests use github.com/twmb/franz-go/pkg/kfake to spin an in-process broker for JSON, protobuf, message-key, drop-on-full, close-flush, initial-ping-failure, name-stability, and config validation cases.
- The shared `kafka` package has its own unit tests for option construction, background ping, error classification, and broker-list helpers.

Dependencies added:
  - github.com/twmb/franz-go
  - github.com/twmb/franz-go/pkg/kfake (test)
  - github.com/twmb/franz-go/plugin/kslog

#### Pull Request Checklist

    - Fixes #<issue number>
    
- Is this a bugfix?
- Is this a new feature?
- Is this a breaking change?

```release-notes
[FEATURE] Event recorder: add Kafka output (event_recorder.kafka_outputs).
[CHANGE] Event recorder: group outputs by type (file_outputs/webhook_outputs/kafka_outputs) instead of a single outputs list with a type field.
[CHANGE] Event recorder: rename alertmanager_event_webhook_drops_total to alertmanager_event_output_drops_total{output}.
```


eventrecorder/kafka.go
```diff
@@ -0,0 +1,358 @@
...
+	close(ko.done)
+	ko.wg.Wait()
+
+	ctx, cancel := context.WithTimeout(context.Background(), ko.flushBudget)
+	defer cancel()
+	if err := ko.client.Flush(ctx); err != nil {
+		ko.logger.Warn("Kafka event recorder flush did not complete within budget; remaining records will be dropped",
+			"output", ko.name, "err", err)
+	}
+	ko.client.Close()
```
TheMeier, line 362: This could hang. Call ko.client.LeaveGroupContext() before

TheMeier:
> A docker-compose example under examples/kafka/ provides a single-node Apache Kafka (KRaft) broker for manual end-to-end verification with a matching alertmanager.yml.

This seems to be missing

siavashs:
> > A docker-compose example under examples/kafka/ provides a single-node Apache Kafka (KRaft) broker for manual end-to-end verification with a matching alertmanager.yml.
> 
> This seems to be missing

Ah, that was for my local tests but it leaked into the commit message.
Do you want me to push that?

eventrecorder/kafka.go
siavashs: Our client is a `producer` and not a `consumer`, therefore it is not joining any groups and calling `LeaveGroupContext()` will be a no-op.
I can push a code doc change to clarify that if you think it is necessary.

eventrecorder/webhook.go
```diff
@@ -15,18 +15,58 @@ package eventrecorder
...
 	commoncfg "github.com/prometheus/common/config"
 )
 
+// validateWebhook validates the webhook-output fields of an Output.
+// Called from Output.UnmarshalYAML when Type == OutputWebhook.
+func (o *Output) validateWebhook() error {
+	if o.URL == nil {
+		return errors.New("event_recorder webhook output requires a url")
+	}
+	return nil
```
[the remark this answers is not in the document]
siavashs: This is a pre-existing bug, we only moved the code, I can fix it in a separate PR or here.

docs/configuration.md
```diff
@@ -2045,3 +2045,120 @@ room_id: <tmpl_string>
...
+delivery, and mute/inhibit suppressions) and fans them out to one or more
+destinations.  Each event is encoded as a structured payload that includes
+a timestamp, the producing instance's hostname, the cluster position (when
+HA clustering is enabled), and the event-specific data.
+
+The recorder is gated behind the `event-recorder` feature flag — pass
+`--enable-feature=event-recorder` on the command line to activate it.
+When the flag is not set, the recorder silently discards all events.
+
+Event recording is configured under the top-level `event_recorder` key.
```
TheMeier, line 2074: Is this enough? Or should there be a section like:

```
# Optional event recorder configuration
[ event_recorder: <event_recorder_config> ]
```
in the top level section?

TheMeier:
Maybe add a note about topic creation to the docs (required if  `auto.create.topics.enable`  is not set?)

TheMeier:
> > > A docker-compose example under examples/kafka/ provides a single-node Apache Kafka (KRaft) broker for manual end-to-end verification with a matching alertmanager.yml.
> > 
> > 
> > This seems to be missing
> 
> Ah, that was for my local tests but it leaked into the commit message. Do you want me to push that?

your call, either add it or remove the text claiming it's there ;)

docs/configuration.md
```diff
@@ -2049,3 +2057,131 @@ room_id: <tmpl_string>
...
+
+**Topic creation:** Alertmanager does not create the topic itself.  The
+target topic must either already exist on the cluster or the brokers
+must be configured with `auto.create.topics.enable=true` (the Kafka
+default, but explicitly disabled in many production clusters).  When
+the topic is missing on a cluster that disables auto-creation, produce
+attempts fail with `UNKNOWN_TOPIC_OR_PARTITION` and are counted in the
+`alertmanager_event_kafka_produce_errors_total` metric.  Create the
+topic up-front with a tool such as `kafka-topics.sh`, choosing the
+partition count and replication factor appropriate for your event
```
Spaceman1701: tiny nitpick, but I think this last sentence is probably a little too much detail for these docs. Those who use this will hopefully know/have their own mechanism for creating topics

eventrecorder/config.go
```diff
@@ -39,48 +28,89 @@ type Config struct {
...
+// per-type field groups are listed below.  Validation and equality
+// comparison for the type-specific fields live in the corresponding
+// output file (file.go, webhook.go, kafka.go).
 type Output struct {
-	Type       string                      `yaml:"type" json:"type"`
-	Path       string                      `yaml:"path,omitempty" json:"path,omitempty"`
-	URL        *amcommoncfg.SecretURL      `yaml:"url,omitempty" json:"url,omitempty"`
+	// Type selects the destination kind.  Must be one of OutputFile,
+	// OutputWebhook, or OutputKafka.
+	Type string `yaml:"type" json:"type"`
```
Spaceman1701: Maybe it's too late since we already have the `Type` field, but I'd rather we structure this config similar to the way we structure receivers with one struct per output type + `WebhookOutputs`, `FileOutputs`, and `KafkaOutputs` on the root config type.

eventrecorder/kafka.go
```diff
@@ -0,0 +1,493 @@
...
+)
+
+// Kafka compression codecs.
+const (
+	KafkaCompressionNone   = "none"
+	KafkaCompressionGzip   = "gzip"
+	KafkaCompressionSnappy = "snappy"
+	KafkaCompressionLZ4    = "lz4"
+	KafkaCompressionZstd   = "zstd"
+)
```
Spaceman1701: I think these should all be typedef-d into named types, e.g.
```go
type KafkaFormat string
type KafkaAcksMode string
type KafkaCompression string
```

eventrecorder/kafka.go
```diff
@@ -0,0 +1,493 @@
...
+	bb := append([]string(nil), b...)
+	sort.Strings(aa)
+	sort.Strings(bb)
+	for i := range aa {
+		if aa[i] != bb[i] {
+			return false
+		}
+	}
+	return true
+}
```
Spaceman1701, line 128:
```suggestion
	aa := append([]string(nil), a...)
	bb := append([]string(nil), b...)
	sort.Strings(aa)
	sort.Strings(bb)
	return slices.Equal(aa, bb)
```

eventrecorder/recorder.go
```diff
@@ -0,0 +1,410 @@
...
+	Name() string
+	SendEvent(data []byte) error
+	io.Closer
+}
+
+// ProtoDestination is an optional extension of Destination for outputs
+// that want to consume events as protobuf messages directly, skipping
+// the JSON serialization step.  Outputs that implement this interface
+// and return true from WantsProto receive SendProto calls instead of
+// SendEvent; SendEvent must remain implemented for fallback callers.
```
Spaceman1701: this seems like a dangerous interface. For the new kafka destination it would be extremely wrong to fallback to a json message for any reason...

Do we have any implementations of `Destination` that don't also implement `ProtoDestination`? Could these be the same interface?

kafka/errors.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Error categories returned by ClassifyError.  These string values
+// are intended for use as bounded-cardinality Prometheus label values.
+const (
+	ErrorCategoryNone    = "none"
+	ErrorCategoryTimeout = "timeout"
+	ErrorCategoryNetwork = "network"
+	ErrorCategoryBroker  = "broker"
+	ErrorCategoryUnknown = "unknown"
+)
```
Spaceman1701, line 34: I think these should have a type alias, e.g. `type ErrorCategory string`

kafka/kafka.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+// Compression codecs supported on the producer wire.
+const (
+	CompressionNone   = "none"
+	CompressionGzip   = "gzip"
+	CompressionSnappy = "snappy"
+	CompressionLZ4    = "lz4"
+	CompressionZstd   = "zstd"
+)
```
Spaceman1701, line 67: I think each one of these should have a type alias so they're all properly associated - e.g.
```go
type Format string
type Acks string
type Compression string
```

kafka/broker.go
```diff
@@ -0,0 +1,49 @@
...
+	bb := append([]string(nil), b...)
+	sort.Strings(aa)
+	sort.Strings(bb)
+	for i := range aa {
+		if aa[i] != bb[i] {
+			return false
+		}
+	}
+	return true
+}
```
Spaceman1701, line 45:
```suggestion
func BrokerListsEqual(a, b []string) bool {
	if len(a) != len(b) {
		return false
	}
	aa := append([]string(nil), a...)
	bb := append([]string(nil), b...)
	sort.Strings(aa)
	sort.Strings(bb)
	return slices.Equal(aa, bb)
}
```

kafka/broker.go
```diff
@@ -0,0 +1,49 @@
+// Copyright The Prometheus Authors
```
Spaceman1701, line 1: tiny nitpick, but I don't think `name.go` is a good name for this file... it seems to be broker utilities

eventrecorder/config.go
siavashs: This will break the schema but it is not a documented feature and we can break it.
Should I address it in this PR or a follow up?

eventrecorder/recorder.go
siavashs: I removed this one. `Destination.SendEvent(event *eventrecorderpb.Event)` now takes the structured event. Each destination owns its serialization (file/webhook → protojson, kafka → JSON or protobuf per its config).

eventrecorder/config.go
siavashs: Updated the config and docs.

Spaceman1701 (approved):

PR #5259, opened 2026-05-29
chore: extract main into an embeddable app package

### LINKED ISSUE #406: Extract main method into package
If we extract the `main` logic, we can actually call it like a standard method. We can then actually make our functional test start Alertmanagers within the test.

This way we don't rely on an existing Alertmanager binary, which also has to be rebuilt before running functional tests. It will also reduce the overhead of our functional to make them less flaky.

siavashs:
## Summary

Resolves the long-standing TODO from [#406](https://github.com/prometheus/alertmanager/issues/406) by extracting the Alertmanager process logic out of [`cmd/alertmanager/main.go`](https://github.com/prometheus/alertmanager/blob/main/cmd/alertmanager/main.go) into a new `app` package, and giving it a lifecycle API (`New` / `Start` / `Addr` / `Reload` / `Stop`) so tests and other binaries can embed Alertmanager in-process instead of building and shelling out to the compiled binary.

[`cmd/alertmanager/main.go`](https://github.com/prometheus/alertmanager/blob/main/cmd/alertmanager/main.go) shrinks from **722 → 207 lines** and now owns only: kingpin flag parsing, logger construction, `versioncollector` registration, feature-flag / GOMEMLIMIT side effects, and translating OS signals into context cancellation (SIGINT/SIGTERM) plus reload events (SIGHUP) consumed by `app.Run`.

### A reusable `app` package

The package is mainly intended for internal use, it can be used by other projects embedding Alertmanager as well.

### A lifecycle API for embedders

```go
New(opts) (*App, error)
(*App).Start() error
(*App).Addr() string             // first listener
(*App).Addrs() []string          // all listeners
(*App).Reload(ctx) error
(*App).Stop(ctx) error
```

`Run` is a thin wrapper (`New` + `Start` + `serveLoop` + `Stop`) with a deferred `Stop` so cleanup also runs on panic. Listeners are bound at `New` time so `Addr()`/`Addrs()` report the real bound ports (including kernel-assigned `:0`) before serving starts.

### Per-instance state, so multiple instances can coexist

The Prometheus collectors that used to be package-level `promauto` variables in [`cmd/alertmanager/main.go`](https://github.com/prometheus/alertmanager/blob/main/cmd/alertmanager/main.go) are now built per `Run()` against `opts.Registerer`, which is threaded through every collaborator (`versioncollector` excepted, which stays process-global in `main.go`). This unblocks running multiple Alertmanager instances in the same process without duplicate-registration panics.

### Deterministic, observable shutdown

`setup` registers teardown steps on a LIFO cleanup stack that `Stop` drains in reverse, mirroring Go's `defer` semantics so shutdown ordering follows construction order automatically. Each step is named; `Stop` runs them all, logs any failure by name, and returns the errors joined with the HTTP-shutdown error. The HTTP shutdown honors a single timeout derived from the context passed to `Stop`.

### Isolated config-reload logic

The config-scoped subgraph lives in a dedicated `reloader` type. `reloader.reload` performs the stop-old → build-new → wait-for-loading → atomic-swap sequence, and `reloader.stop` tears down the live inhibitor and dispatcher at shutdown. The long-lived singletons (nflog, silences, alerts, cluster peer, API, event recorder, tracing) are constructed once and updated in place on reload (`apih.Update`, `eventRec`/`tracing` `ApplyConfig`) rather than rebuilt.

## Behavioural notes

- `prometheus.DefaultRegisterer` is no longer referenced inside `app.Run`; the binary still passes it in via `Options.Registerer`, so on-disk behaviour is identical.
- `srv.Shutdown` now actually runs on `Run` exit (previously the deferred `srv.Close` lived inside the listen goroutine and never ran in practice because `os.Exit` killed the process first). Behaviour for the binary is unchanged; embedded callers now get clean HTTP teardown.
- Systemd socket activation and `vsock://` listen addresses work under both the binary and embedders; the external URL is derived from the actual bound address.
- `tracingManager.Stop` is part of the cleanup stack and always runs, not just on `ctx.Done()`.
- `Start`/`Stop` are concurrency-safe and `/-/reload` works in embedded mode (no deadlock without `Run`/`serveLoop`).
- `--cluster.listen-address` default moved from a const in `cmd/alertmanager` to the exported `app.DefaultClusterAddr`.

## Known follow-ups (out of scope)

- [`matcher/compat.InitFromFlags`](https://github.com/prometheus/alertmanager/blob/main/matcher/compat/parse.go) still mutates package-level state; multi-instance tests with **different** feature flags will collide. Tracked separately.
- Migrating the [v2 acceptance harness](https://github.com/prometheus/alertmanager/tree/main/test/with_api_v2) to drive `app.Run` directly instead of building and spawning the binary. Now mechanically possible thanks to `Addr()` / `Stop()` on `*App`; left for a follow-up PR to keep this one reviewable.

## Verification

Tests live in [`app`](https://github.com/prometheus/alertmanager/tree/main/app) (lifecycle, listen, options, url, reloader, cluster). Highlights:

- **`TestApp_StartStop`** — boot, probe `/-/healthy`, stop, stop again (idempotency).
- **`TestApp_TwoSequentialInstances` / `TestApp_TwoConcurrentInstances`** — multiple instances in one process; guards the metrics-per-Registerer behaviour.
- **`TestApp_ClusteredStartStop`** — gossip clustering enabled; exercises the peer-dependent paths and `clusterWait`.
- **`TestApp_ConcurrentStartStop`** — races `Start` and `Stop` (run under `-race`).
- **`TestApp_EmbeddedReloadDoesNotDeadlock`** — `/-/reload` completes in embedded mode.
- **`TestApp_New_SetupFailureDoesNotDeadlock`** — setup-failure rollback doesn't block.
- **`TestApp_Run_ContextCancel`** — end-to-end `Run` wrapper with ctx cancellation.
- **`TestApp_serveLoop` / `TestApp_Stop_AggregatesCleanupErrors`** — serve-loop exits and aggregated shutdown errors.
- **`TestReloader_*`** — component swap on reload, error path leaves prior state intact, nil-safe stop.
- **`TestListenAll_*` / `TestParseVsockPort` / `TestOptions_Validate` / `TestClusterWait`**.

`app` statement coverage is ~84%; the full suite passes under `-race`.

Closes #406

#### Pull Request Checklist

    - Fixes #406
- Is this a bugfix?
- Is this a new feature?
- Is this a breaking change?

```release-notes
NONE
```


app/app.go
```diff
@@ -0,0 +1,516 @@
...
+		return fmt.Errorf("error creating memory provider: %w", err)
+	}
+	a.onStop(alerts.Close)
+
+	var disp atomic.Pointer[dispatch.Dispatcher]
+	a.onStop(func() {
+		if d := disp.Load(); d != nil {
+			d.Stop()
+		}
+	})
```
ultrotter, line 374: Here in setup() we may also need to do an onStop(...) for the inhibitor or we leak it at Stop()

a.onStop(func() {
    if i := inhibitor.Load(); i != nil { i.Stop() }
})

app/options.go
```diff
@@ -0,0 +1,104 @@
...
+	Flagger    featurecontrol.Flagger
+
+	// Reload triggers a configuration reload each time it receives a
+	// value. The binary translates SIGHUP into sends on this channel;
+	// callers can also drive reloads programmatically. A nil channel
+	// disables external reloads (the /-/reload HTTP endpoint still works).
+	Reload <-chan struct{}
+}
+
+func (o *Options) validate() error {
```
ultrotter, line 148: We should also check that WebConfig.WebConfigFile is non-nil or we risk a panic not in the main binary but from embedders...

internal/app/lifecycle.go
```diff
@@ -0,0 +1,284 @@
...
+			}
+			return fmt.Errorf("alertmanager: HTTP listener failed: %w", err)
+		}
+	}
+}
+
+// listenAll binds TCP listeners for every address in
+// flags.WebListenAddresses. Embedders that need systemd socket
+// activation or non-TCP listeners (vsock, etc.) should drive
+// Alertmanager via cmd/alertmanager instead.
```
ultrotter: Doesn't cmd/alertmanager then use listenAll()? Can we verify/have we verified that that web systemd sockets and vsock addresses keep working after this patch is applied? I think we may need to keep cmd/alertmanager on web.ListenAndServe to support them, no?

internal/app/app.go
```diff
@@ -0,0 +1,516 @@
...
+		c := peer.AddState("sil", silences, reg)
+		silences.SetBroadcast(c.Broadcast)
+	}
+
+	// Start providers before the router potentially sends updates.
+	wg.Go(func() {
+		silences.Maintenance(opts.MaintenanceInterval, filepath.Join(opts.DataDir, "silences"), stopc, nil)
+	})
+
+	a.onStop(func() {
```
ultrotter: Should we move the close(stopc) on stop above, before starting either maintenance? If we have an error in between the onstop is never registered?

app/lifecycle.go
```diff
@@ -0,0 +1,284 @@
...
+		return errors.New("alertmanager/app: App.Reload called before successful New")
+	}
+	return a.coord.Reload()
+}
+
+// Stop gracefully shuts down the App, draining cleanups in reverse
+// registration order so that teardown ordering matches the original
+// defer chain in Run. Safe to call multiple times; safe to call before
+// Start (it will then merely roll back what setup registered).
+func (a *App) Stop(ctx context.Context) error {
```
ultrotter, line 178: Stop returns error but in reality returns only nil, and the internal a.stoppedOnce.Do logs error but doesn't propagate them. It makes sense to either propagate the errors or to make Stop return nothing so nobody expects to see an error from it?

internal/app/lifecycle.go
```diff
@@ -0,0 +1,284 @@
...
+	// shutdown order mirrors the original `defer` chain in Run.
+	cleanups []func()
+
+	startedOnce sync.Once
+	startErr    error
+	// started records whether the serve goroutine in Start was actually
+	// launched. Stop uses this to decide whether draining a.srvc is
+	// meaningful — if Start never ran, nothing will ever close srvc and
+	// the drain would deadlock (e.g., during setup-failure rollback).
+	started bool
```
ultrotter: Should this be an atomic bool. and swapped? At the moment nothing prevents Stop and Start to run in parallel, tho unlikely...

internal/app/app.go
```diff
@@ -0,0 +1,516 @@
...
+		Logger:          logger.With("component", "api"),
+		Registry:        reg,
+		RequestDuration: m.requestDuration,
+		GroupFunc:       groupFn,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create API: %w", err)
+	}
+
+	amURL, err := extURL(logger, os.Hostname, (*opts.WebConfig.WebListenAddresses)[0], opts.ExternalURL)
```
ultrotter: Here you might get a :0 port... Should we calculate extURL after ListenAll?

if len(a.listeners) > 0 {
    amURL, err = extURL(logger, os.Hostname, a.listeners[0].Addr().String(), opts.ExternalURL)
}

internal/app/app.go
```diff
@@ -0,0 +1,516 @@
...
+	)
+	a.coord = configCoordinator
+
+	configCoordinator.Subscribe(func(conf *config.Config) error {
+		// Reload event recorder outputs first so events emitted during
+		// the rest of this callback (e.g., by stopping the old
+		// dispatcher) go to the new outputs.
+		eventRec.ApplyConfig(conf.EventRecorder)
+
+		tmpl, err = template.FromGlobs(conf.Templates)
```
ultrotter: Should we reuse err here?

internal/app/app.go
```diff
@@ -0,0 +1,516 @@
...
+// implicit panic-safety of the original defer-based implementation.
+func Run(ctx context.Context, opts Options) error {
+	a, err := New(opts)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		stopCtx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		defer cancel()
+		_ = a.Stop(stopCtx)
```
ultrotter: Here you pass a 30 seconds timeout but in Stop you put a 5 seconds timeout on top? Should we choose just one? Also should the timeout constants be explicit constants?

internal/app/app.go
```diff
@@ -0,0 +1,516 @@
...
+		// dispatcher before it finishes populating the aggrGroups.
+		go newDisp.Run(startTime.Add(opts.DispatchStartDelay))
+		newDisp.WaitForLoading()
+		disp.Store(newDisp)
+
+		if err := tracingManager.ApplyConfig(conf.TracingConfig); err != nil {
+			return fmt.Errorf("failed to apply tracing config: %w", err)
+		}
+
+		go tracingManager.Run()
```
ultrotter: Is a new tracingmanager started on each config reload, and not stopped then?

internal/app/app.go
```diff
@@ -0,0 +1,516 @@
...
+	})
+	if err != nil {
+		return fmt.Errorf("failed to create API: %w", err)
+	}
+
+	amURL, err := extURL(logger, os.Hostname, (*opts.WebConfig.WebListenAddresses)[0], opts.ExternalURL)
+	if err != nil {
+		return fmt.Errorf("failed to determine external URL: %w", err)
+	}
+	logger.Debug("external url", "externalUrl", amURL.String())
```
ultrotter: external_url here?

internal/app/app.go
```diff
@@ -0,0 +1,516 @@
...
+		return fmt.Errorf("failed to apply initial configuration: %w", err)
+	}
+
+	// Default routePrefix to externalURL path if empty.
+	routePrefix := opts.RoutePrefix
+	if routePrefix == "" {
+		routePrefix = amURL.Path
+	}
+	routePrefix = "/" + strings.Trim(routePrefix, "/")
+	logger.Debug("route prefix", "routePrefix", routePrefix)
```
ultrotter: route_prefix here?

internal/app/lifecycle.go
```diff
@@ -0,0 +1,284 @@
...
+// caller is responsible for calling Stop, ideally via a deferred call so
+// teardown also runs on panic. An App is single-use: calling Start more
+// than once is an error.
+type App struct {
+	opts   Options
+	logger *slog.Logger
+
+	// Lifecycle dependencies retained for use by Start, Reload, and Stop.
+	coord     *config.Coordinator
+	tracing   *tracing.Manager
```
ultrotter: nit: Let's call this tracingMgr to avoid confusing it with the tracing import?

app/lifecycle.go
```diff
@@ -0,0 +1,284 @@
...
+func (a *App) Addr() string {
+	if len(a.listeners) == 0 {
+		return ""
+	}
+	return a.listeners[0].Addr().String()
+}
+
+// Addrs returns all bound listener addresses in the order given by
+// Options.WebConfig.WebListenAddresses.
+func (a *App) Addrs() []string {
```
ultrotter, line 128: Should we add tests for Addrs() ?

internal/app/app.go
siavashs: These are moved from main, which string are you suggestion to replace with `external_url`?

internal/app/app.go
siavashs: This is a pre-existing issue it seems but I will fix it because otherwise we will leak one go routine per reload.

internal/app/app.go
siavashs: No, it has to be local.

internal/app/lifecycle.go
siavashs: Oops, I will fix it.

app/lifecycle.go
siavashs: I think it was missed in one of the refactors.

siavashs:
@coderabbitai review

internal/app/lifecycle.go
```diff
@@ -0,0 +1,328 @@
...
+		a.routerQuit = make(chan struct{})
+		a.routerDone = make(chan struct{})
+
+		// reloadRouter consumes /-/reload requests and opts.Reload sends so
+		// they trigger reloads regardless of whether the caller is using
+		// Run (which also runs serveLoop) or the lifecycle API directly
+		// (which doesn't). Without this goroutine the /-/reload HTTP
+		// handler would block forever in embedded mode because its send
+		// on an unbuffered channel has no receiver.
+		go a.reloadRouter()
```
ultrotter: unless we have a mutex protecting start and stop, here there could be a race between starting this goroutines and then setting a.started.Store(true), vs a stop that may see started as false and claim it stopped thus leaking the goroutines.

app/options.go
```diff
@@ -0,0 +1,112 @@
...
+type Options struct {
+	// Storage and lifecycle.
+	ConfigFile                  string
+	DataDir                     string
+	Retention                   time.Duration
+	MaintenanceInterval         time.Duration
+	MaxSilences                 int
+	MaxSilenceSizeBytes         int
+	SilenceLogging              bool
+	AlertGCInterval             time.Duration
```
ultrotter, line 61: Options{} with 0 as AlertGCInterval may be an issue... Maybe let's have a NewOptions() with sensible defaults? or validate more fields?

app/options.go
```diff
@@ -0,0 +1,112 @@
...
+//
+// Logger, Registerer and Flagger are required dependencies; the remaining
+// fields default to their zero value (which generally matches the kingpin
+// flag default).
+type Options struct {
+	// Storage and lifecycle.
+	ConfigFile                  string
+	DataDir                     string
+	Retention                   time.Duration
+	MaintenanceInterval         time.Duration
```
ultrotter, line 57: Should also be non-zero.... Let's verify about other fields that may misbehave with zero values!

app/cluster_test.go
```diff
@@ -0,0 +1,67 @@
...
+		prometheus.NewRegistry(),
+		"127.0.0.1:0", // bind
+		"",            // advertise
+		nil,           // known peers
+		true,          // wait if empty
+		cluster.DefaultPushPullInterval,
+		cluster.DefaultGossipInterval,
+		cluster.DefaultTCPTimeout,
+		cluster.DefaultProbeTimeout,
+		cluster.DefaultProbeTimeout,
```
ultrotter, line 43: passing DefaultProbeTimeout twice? I think you mean the other constant...

internal/app/options.go
```diff
@@ -0,0 +1,112 @@
...
+	if o.Logger == nil {
+		return errors.New("alertmanager/app: Options.Logger is required")
+	}
+	if o.Registerer == nil {
+		return errors.New("alertmanager/app: Options.Registerer is required")
+	}
+	if o.Flagger == nil {
+		return errors.New("alertmanager/app: Options.Flagger is required")
+	}
+	if o.WebConfig == nil || o.WebConfig.WebListenAddresses == nil || len(*o.WebConfig.WebListenAddresses) == 0 {
```
ultrotter: This may be optional if using systemsd? 
So maybe we can do
if o.WebConfig == nil {
...error
}
usingSystemd := o.WebConfig.WebSystemdSocket != nil && *o.WebConfig.WebSystemdSocket
if !usingSystemd && o.WebConfig.WebListenAddresses == nil || len(*o.WebConfig.WebListenAddresses) == 0 {
ERROR
}

internal/app/lifecycle.go
```diff
@@ -0,0 +1,351 @@
...
+			a.logger.Warn("graceful HTTP shutdown failed", "err", err)
+			stopErr = err
+		}
+	}
+	// HTTP is fully drained; no new /-/reload requests can arrive.
+	// Terminate the reload router and wait for it to exit before
+	// running cleanups (Coordinator is among them).
+	if started {
+		close(a.routerQuit)
+		<-a.routerDone
```
ultrotter: Should we also select on ctx.Done() here to avoid being stuck on Stop here?

app/lifecycle.go
```diff
@@ -0,0 +1,351 @@
...
+	for i, l := range a.listeners {
+		out[i] = l.Addr().String()
+	}
+	return out
+}
+
+// Reload triggers a configuration reload (the programmatic equivalent of
+// SIGHUP). Safe to call concurrently with the running App. The reload is
+// synchronous and not cancellable, so it takes no context.
+func (a *App) Reload() error {
```
ultrotter, line 148: Is there a way here to avoid reloading if a Stop() has started, or making sure a reload won't interfere with a concurrent stop, e.g. stop stops goroutines but a concurrent reload starts new ones?

internal/app/lifecycle.go
```diff
@@ -0,0 +1,351 @@
...
+func (a *App) onStop(name string, fn func() error) {
+	a.cleanups = append(a.cleanups, cleanup{name: name, stop: fn})
+}
+
+// trackClose couples a resource with its teardown: it registers c.Close
+// as a named Stop step right where the resource is acquired, so the two
+// can't drift apart (the gap that previously leaked goroutines/handles).
+// Use it for resources whose Close takes no arguments; everything else
+// uses onStop directly.
+func (a *App) trackClose(name string, c interface{ Close() error }) {
```
ultrotter: Do we need this wrapper? Could we just call onStop?

internal/app/lifecycle_test.go
```diff
@@ -0,0 +1,407 @@
...
+func TestApp_Run_ContextCancel(t *testing.T) {
+	// Exercises the Run wrapper end-to-end: cancel ctx and assert it
+	// returns without error and cleanup has run.
+	ctx, cancel := context.WithCancel(t.Context())
+	done := make(chan error, 1)
+	go func() { done <- Run(ctx, testOptions(t)) }()
+
+	// Give Run a moment to bind. We can't peek inside it for Addr, but
+	// cancelling is unconditionally safe.
+	time.Sleep(200 * time.Millisecond)
```
ultrotter: Sleeps make tests slower and flaky at the same time... Could we use something like require.Eventually?

internal/app/lifecycle_test.go
```diff
@@ -0,0 +1,407 @@
...
+		opts:       Options{Reload: reloadCh},
+		routerQuit: make(chan struct{}),
+		routerDone: make(chan struct{}),
+		webReload:  make(chan chan error),
+	}
+
+	go a.reloadRouter()
+	// Give the router a moment; if the closed channel were mishandled it
+	// would busy-loop and dereference the nil coordinator.
+	time.Sleep(20 * time.Millisecond)
```
ultrotter: Also a bit worried about this sleep... "Give a moment" is very very fishy :/ What exactly are we waiting for? Can we encode it in an actual request? Can we make the code not need these "wait a moment" by being not race prone!?

internal/app/lifecycle_test.go
ultrotter: Also what exactly is "unsafe" before?

Spaceman1701 (commented):
So I think this looks good, but it is a very big change, so I'm not confident that I would've spotted every problem. I'm happy to lean on the tests here.

One thing: I'm not sure that `internal` is the correct package for this. If this really makes alertmanager embeddable, maybe we should put it in some public package so that other modules can depend on this code.

siavashs:
> So I think this looks good, but it is a very big change, so I'm not confident that I would've spotted every problem. I'm happy to lean on the tests here.

I don't know if splitting it would be easy because of how sequential the logic is.
Any suggestion to make it easier for review? 


> One thing: I'm not sure that `internal` is the correct package for this. If this really makes alertmanager embeddable, maybe we should put it in some public package so that other modules can depend on this code.

Fair enough, I'll move it into `app` at root level.

app/lifecycle.go
siavashs: Fixed it by locking.

internal/app/lifecycle.go
siavashs: Dropped it.

internal/app/lifecycle_test.go
siavashs: Dropped the sleep.

internal/app/lifecycle_test.go
siavashs: Also dropped.

app/lifecycle.go
```diff
@@ -0,0 +1,292 @@
...
+	// in normal operation (the router exits as soon as routerQuit is
+	// closed), but a stuck coordinator.Reload could block it, so we also
+	// honour ctx: a caller that passed a deadline gets it back instead of
+	// hanging. Abandoning the goroutine is acceptable — the caller chose
+	// the deadline — and we surface it in the returned error.
+	if started {
+		close(a.routerQuit)
+		select {
+		case <-a.routerDone:
+		case <-ctx.Done():
```
ultrotter: This one may never trigger if ctx is context.Background() as for example in some defers...

app/lifecycle.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+// serveLoop blocks until ctx is cancelled or an HTTP listener fails. It
+// is used by Run only; reload routing is handled by reloadRouter, which
+// is started directly from Start so it is also active for embedders that
+// drive the App lifecycle without using Run.
+func (a *App) serveLoop(ctx context.Context) error {
+	for {
+		select {
+		case <-ctx.Done():
```
ultrotter, line 289: Here we should non-blocking drain serveErrc just in case there are errors while we are also done

ultrotter (approved):

PR #5264, opened 2026-05-30
docs: clarify YAML quoting vs matcher token quoting

### LINKED ISSUE #5099: AlertManager matcher quoting documentation is confusing
### What did you do?

The documentation says a matcher contains three tokens: a label name, an operator, and a regular expression or label value.  It then goes on to say that the label name and the regular expression can be a literal or a double-quoted string.

Matchers are defined in a YAML list (technically a sequence).  For alertmanager.yaml purposes, each item of the list is a blob of text that needs to conform to the above format with three tokens.  The initial quoting is all YAML, which will be on the entire blob, not on elements of the blob.  YAML is not going to parse the label or the regular expression with quoting rules, it will only apply quoting rules to the entire blob.  For example:

```
matchers: 
  - env !~ "prod"   # this is NOT seen by YAML as double quoted, it is plain-style  https://yaml.org/spec/1.2.2/#plain-style
  - "env" =~ qa   # this is incorrectly YAML quoted and likely throws an error.
  - 'env !~ "dev"'   # this is a valid YAML single-quoted https://yaml.org/spec/1.2.2/#single-quoted-style   
  - 'env =~ qa`      # correct single-quoted style without pointless double quotes, and allows literal double quotes and brackets in the regex
  - env !~ preprod  # correct plain-style, will not protect from YAML special characters, but in this case there are none
  - "env !~ prod"    # correct double-quoted style, allows single quotes in the regex and can use \<newline> escape to split long lines
  - "env !~ \"uat\""   # valid, but now we are just hurting ourselves, and for what reason?
``` 

I don't know what use the double-quoting of the label name, regular expression, or label value could have.  Once YAML has dealt with quoting using its well-defined rules, there should be no remaining quoting necessary.   I get that alertmanager may support the stripping of double quoting for backward compatibility and the unshakable desire for people to write thing =~ "regex", but the documentation should help clarify.

It will be highly frustrating for people to double quote their regex in an attempt to deal with YAML lexing and parsing, and not realize that they are just providing plain-style YAML.  The way the documentation is written now, it looks like YAML quoting rules are being applied on the label name and the regular expression.  This is not the case if alertmanager is passing the config through a standard YAML parser and consuming the result.

Please update the documentation to indicate that the entire matcher element should be in plain-style, single quoted, or double quoted.  Indicate that although the regex token can be double quoted for backward compatibility, it is not required, and this double-quoting on its own does not protect YAML special characters.  The safest is to single-quote or double-quote the entire matcher element.

If literal is acceptable as indicated by the documentation, double-quoting of the regex token (and the initial label) should be deprecated, or at best kept for backward compatibility.

_No response_

Double quoted regex token is not protected from YAML special characters unless the entire matcher element is quoted.

### System information

_No response_

### Alertmanager version

```text

```

```yaml

```

### Prometheus version

```text

```

```yaml

```

### Logs

```text

```

Cloud-Architect-Emma:
Fixes #5099

Adds a note under the "UTF-8 matchers" section clarifying the distinction between YAML quoting and matcher token quoting.

Problem: The documentation describes double-quoting of individual matcher tokens (label name, label value) but does not explain that this quoting is handled by Alertmanager's own parser, not by YAML. Users writing env =~ "prod" without surrounding YAML quotes may believe the inner double quotes protect against YAML special characters, which they do not.

Change: Added a blockquote note explaining the three YAML quoting styles (plain, single-quoted, double-quoted) with examples, and explicitly clarifying that inner double quotes on matcher tokens provide no YAML-level protection.

## Summary by CodeRabbit

* **Documentation**
  * Clarified YAML quoting behavior for matcher entries in configuration docs, explaining when to use plain, single-quoted, or double-quoted scalars and warning that quoting must be applied at the YAML level to protect against YAML special characters and ensure matchers are interpreted as intended.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -636,18 +636,19 @@ Double-quoted strings can contain all UTF-8 characters. Unlike unquoted literals
 
 > **Note: YAML quoting vs. matcher token quoting**
 >
-> Each entry in a `matchers:` list is a single YAML value. YAML quoting applies
-> to the *entire* matcher string — it does not parse or protect individual tokens
+> Each entry in a `matchers:` list is a single YAML string that Alertmanager
+> parses after the YAML file has been processed. YAML quoting applies to the
+> *entire* matcher string — it does not parse or protect individual tokens
 > within it. The double-quoting described above is handled by Alertmanager's own
 > parser after YAML has already processed the input.
 >
-> - **Plain style** (no surrounding quotes): works when the matcher contains no
+> - **[Plain style](https://yaml.org/spec/1.2.2/#plain-style)** (no surrounding quotes): works when the matcher contains no
 >   YAML special characters (`{`, `}`, `[`, `]`, `,`, `#`, `|`, `>`, `:`).
 >   Example: `env !~ preprod`
-> - **Single-quoted style** (recommended): protects against all YAML special
+> - **[Single-quoted style](https://yaml.org/spec/1.2.2/#single-quoted-style)**: protects against all YAML special
 >   characters and allows literal double quotes inside the matcher.
 >   Example: `'env =~ "prod|staging"'`
-> - **Double-quoted style**: supports YAML escape sequences; literal double quotes
+> - **[Double-quoted style](https://yaml.org/spec/1.2.2/#double-quoted-style)**: supports YAML escape sequences; literal double quotes
 >   inside the value must be escaped as `\"`.
 >   Example: `"env !~ \"uat\""`
 >
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -633,6 +633,29 @@ Unquoted literals can contain all UTF-8 characters other than the reserved chara
 
 Double-quoted strings can contain all UTF-8 characters. Unlike unquoted literals, there are no reserved characters. However, literal double quotes and backslashes must be escaped with a single backslash. For example, to match the regular expression `\d+` the backslash must be escaped `"\\d+"`. This is because double-quoted strings follow the same rules as Go's [string literals](https://go.dev/ref/spec#String_literals). Double-quoted strings also support UTF-8 code points. For example, `"foo!"`, `"bar,baz"`, `"\"baz qux\""` and `"\xf0\x9f\x99\x82"`.
 
+
+> **Note: YAML quoting vs. matcher token quoting**
+>
+> Each entry in a `matchers:` list is a single YAML value. YAML quoting applies
+> to the *entire* matcher string — it does not parse or protect individual tokens
+> within it. The double-quoting described above is handled by Alertmanager's own
+> parser after YAML has already processed the input.
+>
+> - **Plain style** (no surrounding quotes): works when the matcher contains no
+>   YAML special characters (`{`, `}`, `[`, `]`, `,`, `#`, `|`, `>`, `:`).
+>   Example: `env !~ preprod`
+> - **Single-quoted style** (recommended): protects against all YAML special
+>   characters and allows literal double quotes inside the matcher.
+>   Example: `'env =~ "prod|staging"'`
+> - **Double-quoted style**: supports YAML escape sequences; literal double quotes
+>   inside the value must be escaped as `\"`.
+>   Example: `"env !~ \"uat\""`
+>
+> Writing `env =~ "prod"` without surrounding YAML quotes is valid plain-style
+> YAML. The inner double quotes are stripped by Alertmanager's parser and provide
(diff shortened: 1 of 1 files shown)


Cloud-Architect-Emma:
Hi maintainers, could someone please approve the workflow run so the build and test checks can execute? This is a 

documentation-only change adding a note to clarify YAML quoting vs matcher token quoting as reported in #5099. DCO is passing. 

Thanks!

Cloud-Architect-Emma:
@@Spaceman1701,  @jheiselman  PR is ready for review!

jheiselman:
@Cloud-Architect-Emma I think the documentation looks good! Thank you.

Cloud-Architect-Emma:
@jheiselman, could you approve CI so it can run !

Thanks

jheiselman:
@Cloud-Architect-Emma I am not part of the this project, just a user, so I have no authority here.

docs/configuration.md
```diff
@@ -633,6 +633,29 @@ Unquoted literals can contain all UTF-8 characters other than the reserved chara
...
+>
+> Each entry in a `matchers:` list is a single YAML value. YAML quoting applies
+> to the *entire* matcher string — it does not parse or protect individual tokens
+> within it. The double-quoting described above is handled by Alertmanager's own
+> parser after YAML has already processed the input.
+>
+> - **Plain style** (no surrounding quotes): works when the matcher contains no
+>   YAML special characters (`{`, `}`, `[`, `]`, `,`, `#`, `|`, `>`, `:`).
+>   Example: `env !~ preprod`
+> - **Single-quoted style** (recommended): protects against all YAML special
```
Spaceman1701: I don't think we need to recommend any specific style. At the end of the day, the config file is yaml and there are many ways to write strings. Users can choose to do any of them.

docs/configuration.md
```diff
@@ -633,6 +633,29 @@ Unquoted literals can contain all UTF-8 characters other than the reserved chara
 
 Double-quoted strings can contain all UTF-8 characters. Unlike unquoted literals, there are no reserved characters. However, literal double quotes and backslashes must be escaped with a single backslash. For example, to match the regular expression `\d+` the backslash must be escaped `"\\d+"`. This is because double-quoted strings follow the same rules as Go's [string literals](https://go.dev/ref/spec#String_literals). Double-quoted strings also support UTF-8 code points. For example, `"foo!"`, `"bar,baz"`, `"\"baz qux\""` and `"\xf0\x9f\x99\x82"`.
 
+
+> **Note: YAML quoting vs. matcher token quoting**
+>
+> Each entry in a `matchers:` list is a single YAML value. YAML quoting applies
```
Spaceman1701: It's probably worth making it clear that each entry in the matchers list is a _string_ that is parsed after the yaml file is parsed.

docs/configuration.md
```diff
@@ -633,6 +633,29 @@ Unquoted literals can contain all UTF-8 characters other than the reserved chara
...
+> to the *entire* matcher string — it does not parse or protect individual tokens
+> within it. The double-quoting described above is handled by Alertmanager's own
+> parser after YAML has already processed the input.
+>
+> - **Plain style** (no surrounding quotes): works when the matcher contains no
```
Spaceman1701: It's probably worth it to just link each of these options to the proper yaml docs:
* https://yaml.org/spec/1.2.2/#plain-style
* https://yaml.org/spec/1.2.2/#single-quoted-style
* https://yaml.org/spec/1.2.2/#double-quoted-style

Spaceman1701 (commented):
Thanks for the contribution! 

I think it makes sense to clarify the documentation here.

Cloud-Architect-Emma:
Hi @Spaceman1701, thanks for the feedback! I've addressed all three points:

- Removed the "recommended" label from single-quoted style, no preference is expressed now
- Clarified that each entry is a YAML string that Alertmanager parses after YAML processing
- Added links to the official YAML spec for each quoting style (plain, single-quoted, double-quoted)

Happy to make further adjustments if needed!

docs/configuration.md
Cloud-Architect-Emma: all fixed

docs/configuration.md
Cloud-Architect-Emma: fixed

docs/configuration.md
Cloud-Architect-Emma: fixed

Spaceman1701 (approved):

SoloJacobs (approved):

PR #5266, opened 2026-06-02
cluster: use hostname for TLS validation instead of resolved IP

toller892:
## Problem

When `--cluster.peer` specifies a hostname (e.g. `myhost.example.com:9094`), the `resolvePeers` function resolves it to an IP address. The TLS transport then validates the peer's certificate against this IP, requiring IP SANs. This fails for certificates that only have DNS SANs — which is the common case.

```
tls: failed to verify certificate: x509: cannot validate certificate for 49.12.57.136 because it doesn't contain any IP SANs
```

Users are forced to configure `tls_client_config.server_name` as a workaround, even though the hostname was already provided to `--cluster.peer`.

## Fix

Track the mapping from resolved IP:port back to the original hostname during `resolvePeers`. When dialing via TLS, set `ServerName` to the original hostname so certificate validation checks against DNS SANs.

The mapping is maintained across peer refresh cycles so reconnections also use the correct hostname.

**Changes:**
- `cluster/cluster.go`: `resolvePeers` now returns a `map[string]string` mapping resolved IP:port to original hostname. The `Peer` struct stores a reference to the TLS transport to update the mapping on refresh.
- `cluster/tls_transport.go`: Added `peerHostnames` field and `SetPeerHostnames` method. `DialTimeout` and `WriteTo` use the hostname for TLS `ServerName` when available.
- `cluster/connection_pool.go`: `borrowConnection` accepts an optional hostname parameter for TLS `ServerName`.
- `cluster/cluster_test.go`: Added `TestResolvePeersHostnameMapping` test.

Fixes #5112

## Summary by CodeRabbit

* **New Features**
  * Enhanced TLS peer hostname mapping and validation for secure cluster communication
  * Improved peer resolution with hostname tracking and updates

* **Tests**
  * Added test coverage for peer hostname mapping validation

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -62,6 +62,7 @@ type Peer struct {
 
 	resolvedPeers       []string
 	resolvePeersTimeout time.Duration
+	tlsTransport        *TLSTransport
 
 	mtx    sync.RWMutex
 	states map[string]State
@@ -173,7 +174,7 @@ func Create(
 
 	ctx, cancel := context.WithTimeout(context.Background(), resolveTimeout)
 	defer cancel()
-	resolvedPeers, err := resolvePeers(ctx, knownPeers, advertiseAddr, &net.Resolver{}, waitIfEmpty)
+	resolvedPeers, peerHostnames, err := resolvePeers(ctx, knownPeers, advertiseAddr, &net.Resolver{}, waitIfEmpty)
 	if err != nil {
 		return nil, fmt.Errorf("resolve peers: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


SoloJacobs:
Hi @toller892 ,

without a sign-off on the commits, we can't review this PR.

Note: From the PR description, I don't think this PR has the right shape for fixing the linked issue. Generally, the problem for Alertmanager is that it implements the ordinary PKI model sometimes:  applied per peer, one cert per host, named for the host. Other times, it uses a cluster/mesh style validation. We should probably first pick one which style of setup is the preferred one, and then add the necessary configuration parameters. The current configurations don't clearly allow to guess what users intended to configure.

Kind regards

SoloJacobs:
Hi @toller892 ,

We haven't heard back here in a while, I will close this PR. Feel free to open a new one.

Kind regards

PR #5286, opened 2026-06-04
chore: move context functions into their own file

Spaceman1701:
This is part 4 of breaking up notify.go into a few separate files. I'm generally going to aim to

1. move each stage implementation into its own file, or at least a file of related stages
2. keep common types and utilities in notify.go
    
I'm motivated to do this because making changes in notify.go is becoming difficult because of its length and the number of repeated symbols.

I'm going to do this one piece at a time so the reviews are easier. This PR takes the the context functions and moves them into `context.go`. Nothing about the code or behavior changes.

See #5284 for the previous change.

```release-notes
None
```

## Summary by CodeRabbit

* **Refactor**
  * Consolidated notification context handling into a dedicated module and removed duplicate helpers to simplify internal notification logic. No changes to user-facing behavior or functionality.


ultrotter (approved):
LGTM but I would fix the coderabbit nits while we're at it, they seem sensible! :)

siavashs (approved):

PR #5311, opened 2026-06-18
feat(eventrecorder): add stdout output type

### LINKED ISSUE #5306: Allow event_recorder to send send eventes to stdout
### Proposal

When running alertmanager in a container I would like the ability to capture event records as structured logs along with the other logs to avoid having to manage a separate file with log rotate, setup kafka, or a webhook endpoint to foward these events

mihir-dixit2k27:
**Title:** `eventrecorder: add stdout output type`

Adds a new stdout_outputs destination to the event_recorder configuration. When running Alertmanager in a container, users can now capture structured alert events through the container runtime log driver (Docker, Kubernetes, etc.) without managing a separate file, Kafka cluster, or webhook endpoint. Implementation follows the existing per-type-list pattern used by file_outputs, webhook_outputs, and kafka_outputs. The StdoutOutput implements the Destination interface and serializes events as newline-delimited JSON via protojson, matching the format used by FileOutput. Closes #5306

#### Pull Request Checklist

- Fixes #5306
- Is this a new feature?
- Is this a breaking change?

```release-notes
[FEATURE] event_recorder: Add stdout_outputs destination type to stream alert events as newline-delimited JSON to stdout, enabling container deployments to capture events via the runtime log driver without a separate file, Kafka, or webhook setup.
```

---

#### Description

Adds a `stdout_outputs` destination to the `event_recorder` configuration block.

When running Alertmanager in a container, alert events can now be captured through the container runtime's log driver (Docker, Kubernetes, Loki, etc.) alongside regular logs, without needing a separate file path, Kafka cluster, or webhook endpoint.

**Config example:**
```yaml
event_recorder:
  outputs:
    stdout_outputs:
      - {}
```

**Implementation notes:**
- Follows the existing per-type-list pattern used by `file_outputs`, `webhook_outputs`, and `kafka_outputs`.
- `StdoutOutput` serializes events as newline-delimited JSON via `protojson`, matching the format produced by `FileOutput`.
- `Close()` is a no-op — the process owns `os.Stdout`.
- No new dependencies introduced.


ultrotter (approved):

eventrecorder/stdout_test.go
```diff
@@ -0,0 +1,162 @@
...
+	var got string
+	done := make(chan struct{})
+
+	go func() {
+		defer close(done)
+		got = captureStdout(t, func() {
+			rec.RecordEvent(recordCtx(), startupEvent)
+			// Give the write loop time to drain the event before we close
+			// the pipe and read.
+			time.Sleep(100 * time.Millisecond)
```
ultrotter: Could you use require.Eventually or detect the condition somehow better than using time.Sleep?

eventrecorder/recorder.go
ultrotter: Let's just delete the full list from here.

ultrotter: We should probably also remove this, since we can see it's falling out of date :) Or update if you prefer

TheMeier (approved):

TheMeier:
Thanks @mihir-dixit2k27

When testing I noticed the mixed output format when NOT using `--log.format=json`. Not an issue, but maybe a note is in order about this, or even a gate that disallows using it with default log-format.

mihir-dixit2k27:
> Thanks @mihir-dixit2k27
> 
> When testing I noticed the mixed output format when NOT using `--log.format=json`. Not an issue, but maybe a note is in order about this, or even a gate that disallows using it with default log-format.

Done, added the docs note about `--log.format=json `@TheMeier. Lmk if you will  like any changes to the wording.

Spaceman1701 (commented):
I don't necessarily have a problem with the implementation as is, but how would you feel about changing it so that it's a logger output instead of directly writing to stdout? This would ensure that the logger config would apply to the event recorder.

mihir-dixit2k27:
@Spaceman1701 my thinking is that routing through the logger wraps the event in a log envelope, whereas file_outputs and webhook_outputs emit raw protojson ,wanted stdout to stay consistent with those. Happy to change it if you think it's worth the trade-off.

Spaceman1701:
> @Spaceman1701 my thinking is that routing through the logger wraps the event in a log envelope, whereas file_outputs and webhook_outputs emit raw protojson ,wanted stdout to stay consistent with those. Happy to change it if you think it's worth the trade-off.

I don't feel too strongly, I just wanted to raise this to make sure it's been considered. It's also not like having the stdout event output is mutually exclusive with adding a logger based output later.

mihir-dixit2k27:
> I don't feel too strongly, I just wanted to raise this to make sure it's been considered. It's also not like having the stdout event output is mutually exclusive with adding a logger based output later.

makes sense, thanks for raising it!

PR #5317, opened 2026-06-19
feat(api): add status v3 proto

siavashs:
- add proto file
- run `buf generate`

#### Pull Request Checklist

    - Fixes #<issue number>
    
- Is this a bugfix?
- Is this a new feature?
- Is this a breaking change?

```release-notes
NONE
```


TheMeier (approved):

proto/api/status/v1/status.proto
```diff
@@ -0,0 +1,93 @@
...
+message GetHealthResponse {
+  // ready is true once the API has received at least one configuration
+  // update; before that, requests against business-data services may not
+  // reflect the operator's intended configuration.
+  bool ready = 1;
+  // healthy is true while the process is able to serve requests. Always
+  // true at the application level; intended for completeness and parity
+  // with common probe shapes.
+  bool healthy = 2;
+}
```
ultrotter: Is there a case where we would return not healthy?
And are we worried about the fact that the default struct health response{} would have it as false, and there is no way to distinguish "omitted" from "false"? Maybe we can not have this field, or set it as optional, or use  a nullifiable value just in case?

ultrotter: Also should we consider using "connectrpc.com/grpchealth" implementing official grpc.health.v1.Health which then kubernetes and others can use out of the box to verify health? As is we have our own in-house version and additionally it may report something different from the existing healthy and ready endpoints... Might be confusing?

proto/public/alertmanager/v3/status.proto
```diff
@@ -0,0 +1,93 @@
...
+  // balancers. It does not acquire any internal locks.
+  rpc GetHealth(GetHealthRequest) returns (GetHealthResponse) {}
+}
+
+// AlertmanagerStatus is the top-level status payload.
+message AlertmanagerStatus {
+  VersionInfo version_info = 1;
+  AlertmanagerConfig config = 2;
+  ClusterStatus cluster = 3;
+  google.protobuf.Timestamp uptime = 4;
```
ultrotter: Should we call this start_time? uptime implies a relative and increasing number even if apiv2 called it uptime?

proto/public/alertmanager/v3/status.proto
```diff
@@ -0,0 +1,93 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+syntax = "proto3";
+
+package alertmanager.v3;
```
ultrotter: Should we use apiv3 here and in the directory name, to match what buf also has in its config?

proto/api/status/v3/status.proto
```diff
@@ -0,0 +1,93 @@
...
+  // GetHealth is a lightweight liveness/readiness probe intended for load
+  // balancers. It does not acquire any internal locks.
+  rpc GetHealth(GetHealthRequest) returns (GetHealthResponse) {}
+}
+
+// AlertmanagerStatus is the top-level status payload.
+message AlertmanagerStatus {
+  VersionInfo version_info = 1;
+  AlertmanagerConfig config = 2;
+  ClusterStatus cluster = 3;
```
ultrotter, line 32: Should this be optional, in case we have no cluster, or will we always send the message?

proto/public/alertmanager/v3/status.proto
```diff
@@ -0,0 +1,93 @@
...
+
+import "google/protobuf/timestamp.proto";
+
+option go_package = "github.com/prometheus/alertmanager/api/alertmanagerv3";
```
ultrotter: Should we go api/v3 ? to match api/v2 ? And it could be api/v3/connect for the generated connect modules...

proto/api/status/v3/status.proto
TheMeier: I thought in that case "STATE_DISABLED" is used, which is more explicit then omitting stuff

proto/api/status/v1/status.proto
TheMeier: is "waiting for gossip to settle" a state that is live but not healthy?

proto/api/status/v3/status.proto
siavashs: That is correct, we always have the field and if there is no cluster then we return a disabled value for its status.

proto/api/status/v1/status.proto
siavashs: Currently healthy is meaningless in v2 as it returns a fake value, so we can either drop it in v3 or implement what @TheMeier suggested.

ultrotter: If it's meaningless maybe we can update it to behave it like v3? And let's consider the standard grpc.health.v1.Health response, for v3, so it can work directly with other systems using it?

proto/public/alertmanager/v3/status.proto
siavashs: I think correct name will be `status.v3` here, the generated connect package will be `statusv3connect`.

proto/public/alertmanager/v3/status.proto
siavashs: We don't need to follow what API v2 does, it should not affect our future code structure.
And we can't have `api/v3/connect` because connect generated a complex name, see my reply to you comment above.

proto/api/status/v1/status.proto
siavashs: I dropped it.

ultrotter (approved):

PR #5328, opened 2026-06-25
[RFC] Implement route labels

ultrotter:
This change adds templatable lables to a Route, which are then inherited and overridable in children routes.
These are shared with notifiers, and also available via the /groups call, and thus can be displayed consistently between the notifications and the UI. They are calculated on the fly at notification time, and calculated but then cached on Groups() calls. When the alerts list is changed, in the dispatcher, the cache is invalidated.

#### Pull Request Checklist

- Is this a new feature?
    - [X] I have added tests that test the new feature's functionality
- Is this a breaking change?
    - [X] My changes do not break the existing cluster messages
    - [X] My changes do not break the existing api
- [X] I have added/updated the required documentation
- [X] I have signed-off my commits
- [X] I will follow [best practices for contributing to this project](https://docs.github.com/en/get-started/exploring-projects-on-github/contributing-to-open-source)

```release-notes
[FEATURE] Add optional templatable lables to alert routes
```

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -714,6 +714,7 @@ func (ag *aggrGroup) renderRouteLabels(alerts ...*alert.Alert) model.LabelSet {
 
 	renderedRouteLabels := make(model.LabelSet, len(ag.opts.Labels))
 
+	// The notification reason is not available to route labels and is passed as ReasonUnknown.
 	data := ag.tmpl.Data(ag.opts.Receiver, ag.labels, ag.opts.Labels, notify.ReasonUnknown.String(), alerts...)
 
 	logger := ag.logger.With("data", data)
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -628,13 +628,26 @@ type aggrGroup struct {
 	running  atomic.Bool
 	flushIdx uint64
 
-	// mtx guards routeLabels and routeLabelsDirty.
-	mtx sync.RWMutex
-	// routeLabels holds the route labels rendered against the group's current
-	// alerts. They only change when alerts are inserted or deleted, so they are
-	// rendered lazily: routeLabelsDirty marks when a re-render is needed.
-	routeLabels      model.LabelSet
-	routeLabelsDirty bool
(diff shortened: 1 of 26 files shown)


dispatch/dispatch.go
```diff
@@ -824,6 +945,11 @@ func (ag *aggrGroup) flush(notify func(...*alert.Alert) bool) {
 		if err := ag.alerts.DeleteIfNotModified(resolvedSlice, true); err != nil {
 			ag.logger.Error("error on delete alerts", "err", err)
 		} else {
+			if len(resolvedSlice) > 0 {
+				// The alert set changed; the deletion is already visible via
+				// ag.alerts.
+				ag.invalidateRouteLabels()
+			}
```
siavashs, line 952: This file is massive, shall we move all the aggregation group code out of `dispatch.go`?

template/template.go
```diff
@@ -361,8 +481,22 @@ type Data struct {
...
+	// MarkRouteLabelsResolved. Unexported so it is not reachable from templates
+	// or serialized into JSON webhooks.
+	routeLabelsResolved bool
+}
+
+// MarkRouteLabelsResolved marks d.RouteLabels as already-rendered, so the
+// routeLabels template function returns the values verbatim rather than
+// executing them a second time.
+func MarkRouteLabelsResolved(d *Data) {
+	d.routeLabelsResolved = true
```
siavashs: This is a confusing name, while it is not exposed it can imply these are somehow related to resolved alerts!

siavashs (approved):
Left a few comments, otherwise LGTM

dispatch/dispatch.go
ultrotter: Yes.

template/template.go
ultrotter: ACK will change!!

dispatch/dispatch.go
ultrotter: I can do it in a separate PR though, after this one, so that in this one it's still visible what we are changing...

Spaceman1701 (approved):
lgtm

Issue #924, opened 2017-07-24
Alert limitation discussion


stuartnelson3:
We currently show all possible alerts firing. There is a limit at which we can reasonably load and display all these alerts, but we're not sure what that value is.

From a bandwidth perspective:
From @beorn7 
>And one other aspect: If I'm on-call on the road, I sometimes need to work with low-bandwidth (e.g. shared crappy WiFi) and/or very expensive connection to the internet (e.g. tethering my phone). Even if thousands of alerts load fine with a broadband connection, and the browser on my 16GiB laptop is fine handling that, it might be very inconvenient in a situation like the above, where the first thing I do might be to look at the AM landing page (perhaps even on the browser of my smallish mobile phone).

Although the links from an alert notification include filters to reduce the number of results shown, a mass outage could still affect many instances, which might greatly slow the ability to load.

I personally think if you're on call you should probably stay somewhere that has access to fast, reliable internet, so I'm not proposing any solution to this at the moment, just documenting.

beorn7:
I think hitting an AM link with an filter already active is probably fine. If my own area throws that many alerts, I have enough trouble anyway and might really need bigger tubes.

However, if I just type http://alertmanager in my browser to get on overview of things (or simply use the convenient UI for creating a filter), I don't want to kill my network connection because thousands of alerts not relevant to my work will have to be loaded into my browser.

Kellel:
Have we thought about pagination?

andrewipmtl:
Are there any plans to include some sort of pagination and/or limiting to the results set returned by the alerts/groups API?
We want to make our own front-end solution to display alerts, however, or rules may generate tens of thousands of alerts, and through the API, this may take exceptionally long to load, and may even timeout client-side.

roidelapluie:
Given that the alertmanager is made for company-wide scale, it might make sense to have some kind of pagination or limit in the API. When things are bad you want to be able to reach the front page of AM.

andrewipmtl:
> Given that the alertmanager is made for company-wide scale, it might make sense to have some kind of pagination or limit in the API. When things are bad you want to be able to reach the front page of AM.

Indeed, we've already had instances where the front-page would take minutes to load, and even cause the browsers to detect the tabs/windows as unresponsive.

alvinlin123:
My team at AWS started to work on this; we've gotten many user feedbacks that requires pagination for alert listing APIs. We will submit a API design proposal in coming weeks.

PR #5357, opened 2026-07-02
perf(ui): lazily render the alert list with Html.Lazy and Html.Keyed

wbh1:
The /alerts view rebuilt and re-diffed the entire alert-group tree on every model update, so interactions such as typing in the filter bar were sluggish when many alerts (10K+) were loaded.

This wraps the alert-group rendering in lazy4 so it is skipped while its inputs are unchanged, and render the group and alert lists with Html.Keyed keyed by group identity and alert fingerprint so list diffs are matched by identity rather than position.

It's similar to what is done for listing silences [here](https://github.com/prometheus/alertmanager/blob/1dee97b337fbd7af7d2bcd0f154b1a4da2b62e41/ui/app/src/Views/SilenceList/Views.elm).

Now, even though loading a 30MB+ response of all alerts is still slow on first load, typing in the filter box is at least responsive.

#### Pull Request Checklist

    - Related to https://github.com/prometheus/alertmanager/issues/924 #<issue number>
    - No
- Is this a bugfix?
    - No
- Is this a new feature?
    - No
        - I'm not sure how to benchmark the frontend?
- Is this a breaking change?

```release-notes
[PERF] Improve responsiveness of UI when loading thousands of alerts
```

### DIFF
--- a/ui/app/src/Views/AlertList/Views.elm
+++ b/ui/app/src/Views/AlertList/Views.elm
@@ -7,6 +7,8 @@ import Dict
 import Html exposing (..)
 import Html.Attributes exposing (..)
 import Html.Events exposing (..)
+import Html.Keyed
+import Html.Lazy exposing (lazy4)
 import Set exposing (Set)
 import Types exposing (Msg(..))
 import Utils.Filter exposing (Filter)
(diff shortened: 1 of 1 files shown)


TheMeier:
@wbh1 thank you for contributing.

I am wondering if this might just resurface the issue as discussed in https://github.com/prometheus/alertmanager/issues/3817 that there is no real unique and stable groupKey and constructing one here from receiver-name and labels might be a bad choice.

wbh1:
> I am wondering if this might just resurface the issue as discussed in #3817 that there is no real unique and stable groupKey and constructing one here from receiver-name and labels might be a bad choice.

Thanks @TheMeier! Good point.

I dropped the group-level `Html.Keyed` rendering and gone back to plain positional rendering there.

The main tweak that affects performance (the `lazy4` wrapper that skips re-rendering the whole alert-group tree while its inputs are unchanged) is unchanged since that's what actually addresses the filter-bar lag I'm trying to fix here.

I also kept keyed rendering on the per-group alert list since fingerprint is unique within a group but can drop that if you'd prefer.

TheMeier (approved):

SoloJacobs (approved):

PR #5377, opened 2026-07-07
feat(api): add experimental Connect (ConnectRPC) API alongside v2

siavashs:
Introduce an experimental ConnectRPC API surface alongside API v2. Each
service is independently versioned, beginning with status.v3.StatusService.

Serve Connect and gRPC-Web under the route-prefix-aware /api path while
serving native gRPC, health, and reflection at the server root for standard
gRPC clients. Enable HTTP/2 over TLS and plaintext h2c on the main listener.

Implement GetStatus with cluster and configuration state, mark it as
side-effect-free for Connect HTTP GET requests, and prevent response caching.
Keep API v2 mounted at /api/v2 and remove the obsolete API v1 deprecation
responder.

Add unit and Ginkgo end-to-end coverage for Connect GET, gRPC-Web, native
gRPC, health, reflection, protocol-prefix isolation, and v2 coexistence.

```release-notes
[FEATURE] api: Add an experimental ConnectRPC API served alongside `/api/v2/` under the version-neutral `/api/` prefix, exposing the Connect, gRPC, and gRPC-Web protocols plus the gRPC Health Checking Protocol and server reflection. The first service is `status.v3.StatusService`.
```


ultrotter:
I don't think the all-or-nothing approach will work. 

A user may need time to convert all its workflows/integrations/automations from v2 to v3, and we do need both APIs to work concurrently, for a while.

What we need is a way to disable **v3-ONLY** behavior, until the user is ready to stop v2. Basically we have to allow v3 requests, but refuse any request that would not be compartible with v2 (such as using more than one set of matchers in silences). This request filter will be disabled once the user is ready to pass the flag that enables "full" v3/and contextually disables v2 then.

siavashs:
We will keep v2 enabled while connect API is also enabled, removing the flag. v2 will be deprecated when we release Alertmanager v1 and removed in a later release.

The edge cases like the example you provided will be handled per service implementation in future PRs.

test/e2e/harness_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+import (
+	"context"
+	"net/http"
+	"os"
+	"path/filepath"
+	"time"
+
+	"connectrpc.com/connect"
+	. "github.com/onsi/ginkgo/v2"
```
SoloJacobs, line 24: I don't think we should introduce ginkgo in this PR: We should align on one test framework and use it consistently.

api/v1_deprecation_router.go
```diff
@@ -1,65 +0,0 @@
-// Copyright 2023 Prometheus Team
```
SoloJacobs, line 1: Can we move removal of the v1 to a different PR? And didn't we promise to only delete this in version 1.0.0 somewhere?

api/connect/connect.go
```diff
@@ -0,0 +1,92 @@
...
+	defer api.mtx.Unlock()
+	api.alertmanagerConfig = cfg
+}
+
+// Handler returns an http.Handler serving every ConnectRPC service exposed
+// by the Connect API: the versioned application services, the gRPC Health
+// Checking Protocol (grpc.health.v1.Health), and gRPC server reflection
+// (v1 and v1alpha, for tools such as grpcurl). Procedures are
+// fully-qualified, so the returned handler is mounted at a single prefix.
+func (api *API) Handler(opts ...connect.HandlerOption) http.Handler {
```
SoloJacobs, line 148: With --web.route-prefix=/, port 9093

```
$ ./grpcurl -plaintext localhost:9093 list
Failed to dial target host "localhost:9093": context deadline exceeded

$ ./grpcurl -plaintext localhost:9093 status.v3.StatusService/GetStatus
Failed to dial target host "localhost:9093": context deadline exceeded
```
Is that intended?

api/connect/connect.go
```diff
@@ -0,0 +1,92 @@
...
+func (api *API) Update(cfg *config.Config) {
+	api.mtx.Lock()
+	defer api.mtx.Unlock()
+	api.alertmanagerConfig = cfg
+}
+
+// Handler returns an http.Handler serving every ConnectRPC service exposed
+// by the Connect API: the versioned application services, the gRPC Health
+// Checking Protocol (grpc.health.v1.Health), and gRPC server reflection
+// (v1 and v1alpha, for tools such as grpcurl). Procedures are
```
SoloJacobs, line 146: >Procedures are fully qualified

I felt the global namespace was nice. so `package status.v3` should be `package prometheus.alertmanager.v3`? Just something I'm noting.

api/connect/status.go
```diff
@@ -0,0 +1,94 @@
...
+	statusv3 "github.com/prometheus/alertmanager/api/status/v3"
+	"github.com/prometheus/alertmanager/api/status/v3/statusv3connect"
+)
+
+// Ensure API satisfies the generated StatusService handler interface.
+var _ statusv3connect.StatusServiceHandler = (*API)(nil)
+
+// GetStatus returns the Alertmanager instance and cluster status.
+func (api *API) GetStatus(_ context.Context, _ *connect.Request[statusv3.GetStatusRequest]) (*connect.Response[statusv3.GetStatusResponse], error) {
+	api.mtx.RLock()
```
SoloJacobs: I would like to see some form of test for if GetStatus command takes a long time: We don't want this API to allow starving normal operations.

api/connect/connect.go
```diff
@@ -0,0 +1,92 @@
...
+// fully-qualified, so the returned handler is mounted at a single prefix.
+func (api *API) Handler(opts ...connect.HandlerOption) http.Handler {
+	// serviceNames lists the fully-qualified service names advertised via
+	// health checking and reflection.
+	serviceNames := []string{
+		statusv3connect.StatusServiceName,
+	}
+
+	mux := http.NewServeMux()
```
SoloJacobs, line 176: Does this work correctly with proxies ala --web.route-prefix ?

test/e2e/harness_test.go
siavashs: Usage of ginkgo for e2e tests was discussed on the maintainers chat.

api/connect/connect.go
siavashs: It is fixed now, GRPC was only enabled when TLS was configured but we now support insecure mode too.

api/connect/connect.go
siavashs: We could do that, it makes the service names longer though I think.

api/v1_deprecation_router.go
siavashs: We can remove it in a separate change.

api/connect/status.go
siavashs: good catch, we can use an atomic pointer to avoid using a read lock.

api/connect/connect.go
siavashs: yes, it work correctly with connect HTTP, GRPC has no prefix as expected.

api/api_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"net/http/httptest"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/require"
+)
+
+func TestConcurrencyLimitHandler(t *testing.T) {
```
SoloJacobs, line 27: I feel this test would work better, if written using `synctest`. A prerequisite would be using replacing the HTTP server with `httptest`

api/api_test.go
```diff
@@ -0,0 +1,78 @@
...
+			Name: "test_requests_in_flight",
+		}),
+		concurrencyLimitExceeded: prometheus.NewCounter(prometheus.CounterOpts{
+			Name: "test_concurrency_limit_exceeded_total",
+		}),
+		inFlightSem: make(chan struct{}, 1),
+	}
+	srv := httptest.NewServer(api.concurrencyLimitHandler(dst))
+	t.Cleanup(srv.Close)
+	client := &http.Client{Timeout: time.Second}
```
SoloJacobs: 1 second timout could probably fail in the github ci

api/api_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	req, err := http.NewRequestWithContext(t.Context(), http.MethodPost, srv.URL, nil)
+	require.NoError(t, err)
+	resp, err = client.Do(req)
+	require.NoError(t, err)
+	_ = resp.Body.Close()
+	require.Equal(t, http.StatusNoContent, resp.StatusCode)
+
+	unblock()
+	require.NoError(t, <-firstDone)
```
SoloJacobs: Should we add require here, which ensures the semaphore is handed back?

app/app.go
```diff
@@ -527,9 +527,16 @@ func (a *App) setup() error {
 
 	mux := apih.Register(router, routePrefix)
 
+	protocols := new(http.Protocols)
+	protocols.SetHTTP1(true)
+	protocols.SetHTTP2(true)
+	protocols.SetUnencryptedHTTP2(true)
```
SoloJacobs, line 521: Can you explain a bit how this works? What changes for users? Do they have to change their proxies?

api/api.go
```diff
@@ -167,16 +171,24 @@ func New(opts Options) (*API, error) {
 // true for the concurrency limit, with the exception that it is only applied to
 // GET requests.
 func (api *API) Register(r *route.Router, routePrefix string) *http.ServeMux {
-	// TODO(gotjosh) API V1 was removed as of version 0.27, when we reach 1.0.0 we should removed these deprecation warnings.
-	api.deprecationRouter.Register(r.WithPrefix("/api/v1"))
-
 	mux := http.NewServeMux()
-	mux.Handle("/", api.limitHandler(r))
+	connectHandler := api.connect.Handler()
+	grpcHandler := api.instrumentHandler("", connectHandler)
```
SoloJacobs: AI comment: handler label and the OTel span name are taken from r.URL.Path for the entire URL space, for any HTTP/2 request carrying Content-Type: application/grpc .
This seems like it would cause very high cardinality here.

api/api.go
```diff
@@ -128,6 +131,7 @@ func New(opts Options) (*API, error) {
 	if err != nil {
 		return nil, err
 	}
+	connect := apiconnect.NewAPI(opts.Peer, l.With("api", "connect"))
```
SoloJacobs: If it's experimental, shouldn't it be behind a feature flag?

SoloJacobs (changes requested):

api/api.go
siavashs: In another discussion with @ultrotter we decided not to add a feature flag.
See https://github.com/prometheus/alertmanager/pull/5377#issuecomment-4913296828

api/connect/connect.go
siavashs: Naming was approved in #5317

test/e2e/harness_test.go
siavashs: To clarify further, Ginkgo is scoped to the new Connect API for now.
After we remove API v1/v2 we can onboard any existing e2e tests to Ginkgo and add more behaviour e2e tests.

app/app.go
```diff
@@ -527,9 +527,16 @@ func (a *App) setup() error {
...
 	a.server = &http.Server{
 		// Instrument all handlers with tracing.
-		Handler: tracing.Middleware(mux),
+		Handler:           tracing.Middleware(mux),
+		Protocols:         protocols,
+		ReadHeaderTimeout: 10 * time.Second,
+		IdleTimeout:       90 * time.Second,
```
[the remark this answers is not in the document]
siavashs: I am declining to add a global `ReadTimeout` in favour of RPC-level controls.

api/api.go
siavashs: Added bounded service-prefix metric/trace labels for Connect and gRPC.

api/api_test.go
siavashs: Updated the test.

api/api_test.go
siavashs: Good idea, reimplemented using `synctest`.

api/api_test.go
siavashs: Rewrite the test.

app/app.go
siavashs: This makes the existing listener accept HTTP/1.1, HTTP/2 over TLS, and cleartext HTTP/2 using h2c. It does not force existing clients to use HTTP/2 or change the behaviour of API v2.

Existing users do not need to change their proxy configuration unless they want to expose the new native gRPC endpoint.

api/v1_deprecation_router.go
siavashs: Restored it.

api/api.go
```diff
@@ -190,17 +222,60 @@ func (api *API) Register(r *route.Router, routePrefix string) *http.ServeMux {
...
+		api.instrumentConnectHandler(
+			apiPrefix+"/api",
+			servicePrefixes,
+			http.StripPrefix(apiPrefix+"/api", http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if isGRPCRequest(r) {
+					http.NotFound(w, r)
+					return
+				}
+				connectHandler.ServeHTTP(w, r)
+			})),
```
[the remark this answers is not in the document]
siavashs: This is intentional. Connect HTTP GET requests are already bounded by the Connect unary admission interceptor. The resolved --web.get-concurrency value is used as the capacity for a separate Connect unary pool, so GET and POST invocations receive the same protocol-level admission behavior and return ResourceExhausted when full.

We deliberately keep this pool separate from the regular HTTP GET semaphore so slow Connect clients cannot consume API v2 capacity. Wrapping the mount with api.limitHandler would both double-limit Connect GETs and apply http.TimeoutHandler to streaming health/reflection requests, which is unsafe for streams.

The regular HTTP and Connect limits are independent but use the same configured capacity.

api/connect/connect.go
```diff
@@ -73,21 +143,46 @@ func (api *API) Update(cfg *config.Config) {
...
+	return mux
+}
+
+// ServicePrefixes returns the URL path prefixes ("/<fully-qualified-service>/")
+// for every service registered by Handler. Callers use it to bound the
+// cardinality of metric and trace labels: any request path that does not
+// match one of these prefixes yields a 404 and should not be recorded
+// verbatim.
+func (api *API) ServicePrefixes() []string {
+	_, prefixes := api.buildHandler()
```
SoloJacobs: We should just hard code a list here, and pin prefixes with a test, I think

api/connect/connect.go
```diff
@@ -19,40 +19,110 @@
...
-		logger = promslog.NewNopLogger()
+// is disabled.
+func NewAPI(opts Options) *API {
+	unaryConcurrency := opts.UnaryConcurrency
+	if unaryConcurrency < 1 {
+		unaryConcurrency = max(runtime.GOMAXPROCS(0), 8)
+	}
+	streamConcurrency := opts.StreamConcurrency
+	if streamConcurrency < 1 {
+		streamConcurrency = unaryConcurrency
```
SoloJacobs: Sharing the UnaryConcurrency pool isn't a good fallback.

api/connect/connect.go
```diff
@@ -19,40 +19,110 @@
...
+		admission: &admissionInterceptor{
+			unary:        make(chan struct{}, unaryConcurrency),
+			streams:      make(chan struct{}, streamConcurrency),
+			unaryTimeout: opts.UnaryTimeout,
+		},
+	}
+}
+
+// admissionInterceptor gives unary RPCs and streams independent capacity so
+// slow Connect clients cannot consume the API v2 GET request allowance.
```
SoloJacobs, line 81: How does this affect v2?

SoloJacobs (approved):

SoloJacobs:
LGTM, thanks!

api/connect/connect.go
siavashs: No v2 effects:
- v2 continues using the existing HTTP GET semaphore.
- `--web.get-concurrency` still determines its capacity.
- v2 still returns HTTP 503 when that pool is full.
- `--web.timeout` continues through http.TimeoutHandler.

api/connect/connect.go
siavashs: Yes, I think that is better.

api/connect/connect.go
siavashs: True, I will fix it.

siavashs:
We will switch the version to `v3alpha` as discussed, following this convention https://buf.build/docs/lint/rules/#package_version_suffix

SoloJacobs (approved):

PR #5381, opened 2026-07-13
feat(limit): drop resolved alerts without an admitted firing counterpart

siavashs:
Rework the per-alertname alert limit so resolved notifications no longer consume limit slots, and change how they are handled.

- limit: add Bucket.Remove to free a tracked slot by value/fingerprint.
- store: apply the per-alert limit only to firing alerts. A resolved notification is forwarded only if its firing counterpart was previously admitted (its fingerprint is still tracked); removing it frees the slot the firing alert was holding. Resolves with no admitted firing are dropped as noise, since nothing downstream ever received a firing for them from us.
- provider/mem: add a `state` label (firing/resolved) to alertmanager_alerts_limited_total so dropped firing alerts and dropped resolved notifications can be distinguished.
- docs: document the new resolved handling and metric label in alertmanager.md and configuration.md.
- tests: add Bucket.Remove coverage (presence return, slot freeing, double remove, heap eviction-order integrity, concurrency) and store coverage for the drop-if-unseen resolved semantics.

#### Pull Request Checklist

    - Fixes #<issue number>
    
- Is this a bugfix?
- Is this a new feature?
- Is this a breaking change?

```release-notes
[CHANGE] limit: `alertmanager_alerts_limited_total` now has a `state` label (`firing`/`resolved`) to distinguish dropped firing alerts from dropped resolved notifications. Dashboards/alerts using this metric must be updated.
[CHANGE] limit: With `--alerts.per-alertname-limit` set, a resolved notification is only forwarded if its firing counterpart was previously admitted (which frees its slot); resolved notifications with no admitted firing alert are now dropped.
```


provider/mem/mem.go
```diff
@@ -340,6 +341,11 @@ func (a *Alerts) Put(ctx context.Context, alerts ...*types.Alert) error {
 				if a.flagger.EnableAlertNamesInMetrics() {
 					labels = append(labels, alert.Name())
 				}
+				state := "firing"
+				if alert.Resolved() {
```
ultrotter, line 345: nit: this state may theoretically change between when we checked in a.alerts.Set and now... Maybe we need to put the state of the ErrLimited in the error itself?

store/store.go
```diff
@@ -158,15 +158,28 @@ func (a *Alerts) Set(alert *types.Alert) error {
 	fp := alert.Fingerprint()
 	name := alert.Name()
 
-	// Apply per alert limits if necessary
+	// Apply per alert limits if necessary.
 	if a.perAlertLimit > 0 {
```
ultrotter, line 162: On the "firing" side it might be that if a firing alert is rate-limited, it still shows as resolved in a.alerts[fp] until GC, if it resolved before being subsequently rate-limited. Do we care?

limit/bucket.go
ultrotter, line 133: Do we need this check? Would there be a point for a capacity 0 bucket? Should we fail at NewBucket instead in that case?

provider/mem/mem.go
siavashs: The status should not change, for firing alerts Prometheus sets `EndsAt` to few minutes in future by default, so this will work fine unless someone has a very weirdly short-lived alerts configured on prometheus.

For resolved alerts they are already resolved when they hit Alertmanager.

> Maybe we need to put the state of the ErrLimited in the error itself?

Do you mean a special error for the case of resolved vs. firing?

store/store.go
siavashs: If a firing alert is dropped Alertmanager does not track it and it will not be in the bucket or the store.

limit/bucket.go
siavashs: The bucket limiter was added as a generic implementation, zero here means drop all, we can change it in future to allow all.

docs/alertmanager.md
```diff
@@ -78,11 +78,19 @@ It's important not to load balance traffic between Prometheus and its Alertmanag
 Alertmanager supports configuration to limit the number of active alerts per alertname.
 This can be configured using the [--alerts.per-alertname-limit] flag.
 
-When the limit is reached any new alerts are dropped, heartbeats from already know alerts are processed.
-The known alert (fingerprint) automatically expire to make room for new alerts.
+When the limit is reached any new firing alerts are dropped, while heartbeats from
+already known alerts are still processed. Known alerts (fingerprints) automatically
+expire to make room for new alerts.
+
+Resolved notifications are handled specially. A resolved notification is only
```
Spaceman1701, line 85: We should be careful with the phrasing here: a _notification_ already names the thing that is sent from Alertmanager to an integration. Unless I'm misunderstanding what objects are being used to track alert state in the limiter, I think we should call this a _resolved alert_ since an alert is the thing sent from Prometheus to Alertmanager.

provider/mem/mem.go
Spaceman1701: > for firing alerts Prometheus sets EndsAt to few minutes in future by default

This is only true if the thing sending alerts to Alertmanager is well behaved, I agree with @ultrotter that this can technically race (even if it's unlikely). Even with a well behaved Prometheus, extreme load can cause Alertmanager to receive Alerts milliseconds before they expire. 

That all being said, if this is only used for the metrics labels, it's probably not too big of a problem.

store/store.go
Spaceman1701: I might be misunderstanding, but I think the concern is that a firing alert which is resolved here will still persist in the `a.alerts` map until the `Alerts` GC runs and deletes it. So the alert will be removed from the bucket before actually being removed from Alertmanager's memory. I actually think this is _ok_, but it does mean that some pathological cases can still cause a bunch of load on Alertmanager even with the limiter.

siavashs: I will double check but I don't think we are indirectly changing the GC behaviour here compared to when limits are not applied.

provider/mem/mem.go
siavashs: The limit error now carries the store’s alert-state decision.

store/store.go
siavashs: I checked this again but I don't think it is an issue, if the alert is in the store the resolved instance of it will not be limited and GC will perform its job.
This feature is gated so it should not affect the default behaviour.
If we discovered a bug later we can fix it.

PR #5393, opened 2026-07-17
feat(eventrecorder): add outputs name

siavashs:
Require every output to have a unique and valid UTF-8 name.
Avoid exposing webhook URLs, Kafka brokers, topics, and raw transport errors.

```release-notes
[CHANGE] eventrecorder: All outputs now require a configured `name`, used instead of URLs, brokers, topics, or other destination configuration in metric label values and logs.
```

### DIFF
--- a/eventrecorder/config.go
+++ b/eventrecorder/config.go
@@ -13,6 +13,12 @@
 
 package eventrecorder
 
+import (
+	"fmt"
+
+	"github.com/prometheus/common/model"
+)
+
 // Config configures the event recorder feature.
 //
 // Outputs are grouped by type, one list per destination kind, mirroring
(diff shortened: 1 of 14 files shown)


siavashs:
we actually need to fix this for other outputs too, for example the secret url from webhook leaks into metrics and logs.

eventrecorder/file.go
```diff
@@ -111,7 +125,7 @@ func (fo *FileOutput) reopen() {
 	}
 	f, err := openAppend(fo.path)
 	if err != nil {
-		fo.logger.Error("Failed to reopen event recorder file", "path", fo.path, "err", err)
+		fo.logger.Error("Failed to reopen event recorder file", "output", fo.name)
```
TheMeier: I would like to have the file path also in this error log.

eventrecorder/file.go
```diff
@@ -163,7 +177,8 @@ func (fo *FileOutput) watchLoop(ready chan<- error) {
 			if !ok {
 				return
 			}
-			fo.logger.Error("fsnotify error on event recorder directory", "err", err)
+			_ = err
+			fo.logger.Error("fsnotify error on event recorder directory", "output", fo.name)
```
TheMeier: filepath (or its dirname) would also be helpful in this error log

TheMeier (approved):

eventrecorder/file.go
siavashs: I guess for files it is fine to keep the path, since they are not technically a secret.

eventrecorder/config.go
```diff
@@ -13,6 +13,12 @@
 
 package eventrecorder
 
+import (
+	"fmt"
+)
+
+const maxOutputNameLength = 128
```
Spaceman1701: is this limit really necessary? Do we enforce limits on the lengths of other names in the config?

eventrecorder/config.go
```diff
@@ -26,6 +32,75 @@ type Config struct {
...
+		return "", fmt.Errorf("event_recorder %s output requires a name", kind)
+	}
+	if len(name) > maxOutputNameLength {
+		return "", fmt.Errorf("event_recorder %s output name must not exceed %d characters", kind, maxOutputNameLength)
+	}
+	for _, r := range name {
+		if (r < 'a' || r > 'z') && (r < 'A' || r > 'Z') && (r < '0' || r > '9') && r != '-' && r != '_' && r != '.' {
+			return "", fmt.Errorf("event_recorder %s output name must contain only letters, digits, hyphens, underscores, and periods", kind)
+		}
+	}
```
Spaceman1701: I guess I have the same question here - is there a reason we need to constrain the valid names here?

eventrecorder/recorder.go
```diff
@@ -171,33 +171,42 @@ func NewRecorderFromConfig(cfg Config, instance string, logger *slog.Logger, r p
...
+		logger.Error("Invalid event recorder output configuration")
+		return nil
+	}
 	var outputs []Destination
 	for _, fc := range cfg.FileOutputs {
-		fo, err := NewFileOutput(fc.Path, logger)
+		fo, err := NewFileOutput(fc, logger)
 		if err != nil {
-			logger.Error("Failed to create file event recorder output", "path", fc.Path, "err", err)
+			logger.Error("Failed to create file event recorder output", "output", safeOutputIdentifier("file", fc.Name), "path", fc.Path, "err", err)
```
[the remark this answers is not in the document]
Spaceman1701: I disagree with this, I think having the structured output contain the file path is good.

eventrecorder/config.go
siavashs: If i remember correctly the limit is the safe one for label values.
This is configurable on prometheus for example, but 128 characters should be enough to generate unique names.

eventrecorder/config.go
siavashs: I think validation was added for label values.

eventrecorder/config.go
Spaceman1701: Ah, I see, because the name is exported on a metric. This seems fine to me then.

eventrecorder/config.go
Spaceman1701: I think we still have the Prometheus label parser in the source tree - could we use that for validation instead? I'm worried that this will drift from whatever the Prometheus implementation is.

eventrecorder/config.go
siavashs: Actually Prometheus does not have a universal 128-character safe limit for label values.
Label values may contain arbitrary valid UTF-8, and Prometheus’s `label_value_length_limit` is a per-scrape setting whose default is 0—unlimited.
So my initial comment was me confusing our internal Prometheus config with upstream defaults.
But we should probably keep the safe limit here or make it configurable across all Alertmanager metric label values maybe.

eventrecorder/config.go
siavashs: So the label parser only validates label names not label values. the only validation for label values is to check if they are valid UTF-8 strings.
So we can choose to either drop this custom validation or keep it.

eventrecorder/config.go
siavashs: Dropped the limit.

eventrecorder/config.go
siavashs: We now only check UTF-8 label value validity.

Spaceman1701 (approved):

Issue #4989, opened 2026-02-10
Alertmanager exposes errors one by one


SoloJacobs:
### What did you do?

The current implementation of UnmarshalYAML for certain notifiers (specifically discussed regarding Webhooks) uses an early return pattern to propagate errors. This creates a "whack-a-mole" user experience where only the first encountered error is reported, rather than validating the entire configuration block. This specific issue is only about `receivers:` section of the configuration, but other issues may exist.

### What did you expect to see?

Expected to see all errors, which are independent of one another.

### What did you see instead? Under which circumstances?

A single error (which depends on the notifier.

### Alertmanager version

```text

```

### Alertmanager configuration file

```yaml
See subtasks for configuration examples.
```

### Prometheus version

```text

```

### Prometheus configuration file

```yaml

```

### Logs

```text

```

SoloJacobs:
Note: Issue #4992 needs to be done first, to prepare other changes.

PR #5395, opened 2026-07-17
refactor(config): split validation into per-notifier Validate methods

### LINKED ISSUE #4992: Split validation logic into a separate function for each notifier
This task prepares other fixes in #4989 .

Refactor the validation logic currently scattered across various UnmarshalYAML methods into dedicated, per-notifier validation functions.

Dedicated Validation: Create standalone validation functions for WebhookConfig, EmailConfig, etc., to serve as the single source of truth.

ayushi-work:
Extract validation logic from `UnmarshalYAML` methods into standalone `Validate() error` methods for every notifier config type. Adds a `Validator` interface in `config/common`.

This is a prerequisite for #4989 — once validation lives in separate methods, config loading can call `Validate()` on each notifier independently and collect all errors with `errors.Join` instead of failing on the first error.

- 11 files, +123/−20 lines
- All 22 notifier config types now have `Validate() error`
- Data mutation (defaults, normalization) stays in `UnmarshalYAML`
- Pure validation checks move to `Validate()`

Closes #4992

#### Pull Request Checklist

### DIFF
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -200,7 +200,10 @@ func (c *EmailConfig) UnmarshalYAML(unmarshal func(any) error) error {
 	if err := unmarshal((*plain)(c)); err != nil {
 		return err
 	}
-	// Header names are case-insensitive, check for collisions.
+	// Header names are case insensitive. The normalization loop below
+	// detects duplicates and builds a canonical header map in one pass.
+	// Both the detection and the normalization stay here rather than in
+	// Validate to avoid iterating over the headers a second time.
 	normalizedHeaders := map[string]string{}
 	for h, v := range c.Headers {
 		normalized := textproto.CanonicalMIMEHeaderKey(h)
--- a/config/notifiers.go
+++ b/config/notifiers.go
@@ -107,7 +107,10 @@ func (c *WebexConfig) UnmarshalYAML(unmarshal func(any) error) error {
 	if err := unmarshal((*plain)(c)); err != nil {
 		return err
 	}
+	return c.Validate()
+}
 
+func (c *WebexConfig) Validate() error {
 	if c.RoomID == "" {
 		return errors.New("missing room_id on webex_config")
 	}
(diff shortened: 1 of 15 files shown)


TheMeier:
@ayushi-work that PR is maybe a bit pre mature, the proposal I made in the issue was just an idea. I am not sure it is a good one. And if this is the way to go the interface would probably be less sepcific eg `type NotifierConfg interface`. And further the validation implementations would then go into notify/<name>/config.go or so

ayushi-work:
@TheMeier Hi! I kept the interface small because Validate() is really the only behavior I need for #4989, and it felt cleaner to keep validation separate from other concerns. That said, I'm not attached to the interface itself, if we'd rather see a different interface, I'm happy to change it. For the file layout, the external notifier Validate() methods are already in notify//config.go. The ones in config/notifiers.go remained there because those notifier types have always lived in the config package, and moving them felt like a separate cleanup from what I was trying to do here. This is mostly just the mechanical extraction to make the follow-up work in #4989 easier.

TheMeier:
oh sorry I overlooked that this was already done in notify/$NAME/config.go  where available. i am currently moving ther configs out of notifiy/config.go one by one. so actually it looks quite ok to me

ayushi-work:
> oh sorry I overlooked that this was already done in notify/$NAME/config.go where available. i am currently moving ther configs out of notifiy/config.go one by one. so actually it looks quite ok to me

alright great! let me know if anything needs to be changed!

ayushi-work:
Hi! just bumping this up!

config/notifiers.go
Spaceman1701, line 174: I know it's obvious right now, but would you mind adding a comment that explains why this validation is not in `Validate`? For a future someone who isn't familiar with the history of this code, it might be rather surprising.

Spaceman1701 (approved):
I quite like this change - decoupling unmarshaling from validation is a good idea.

config/notifiers.go
ayushi-work: Done, added a comment explaining why the duplicate detection stays in UnmarshalYAML.

Spaceman1701 (approved):

Spaceman1701:
This looks good to me - Once rebased I can merge this. Thanks for the contribution!

ayushi-work:
> This looks good to me - Once rebased I can merge this. Thanks for the contribution!

hi, I have rebased onto the latest main, so glad to make this contribution. thanks!

SoloJacobs (approved):

SoloJacobs:
Thanks, awesome :+1:

Issue #3641, opened 2023-12-18
Requests to the Pushover API will fail if the the contents of the user key file or token file end with a newline character


cyrozap:
**What did you do?**
I wrote my Pushover user key and app token to files using vim (which adds a newline to the end of each file by default) and set `user_key_file` and `token_file` in `alertmanager.yml` to point to those files.

**What did you expect to see?**
Alertmanager should have successfully sent alerts to Pushover, with a notification appearing on my phone.

**What did you see instead? Under which circumstances?**
When Alertmanager tried sending a notification to Pushover, Pushover returned an HTTP 400 error from its API and said that my application token was invalid.

**Workaround**
Either write the secrets to each file without any newlines using `echo -n secret > secret_file.txt` or after writing the files with newlines strip the newline character from each file using `truncate --size=-1 secret_file.txt`.

**Environment**

* System information:

```
Linux 6.5.6-4k-00001-g6e5115d7f640 ppc64le
```

* Alertmanager version:

```
alertmanager, version 0.26.0 (branch: HEAD, revision: d7b4f0c7322e7151d6e3b1e31cbc15361e295d8d)
  build user:       root@9623f4d745ac
  build date:       20230824-11:10:34
  go version:       go1.20.7
  platform:         linux/ppc64le
  tags:             netgo
```

* Prometheus version:

```
prometheus, version 2.48.1 (branch: HEAD, revision: 63894216648f0d6be310c9d16fb48293c45c9310)
  build user:       root@bd93f1a7309d
  build date:       20231208-23:33:20
  go version:       go1.21.5
  platform:         linux/ppc64le
  tags:             netgo,builtinassets,stringlabels
```

* Alertmanager configuration file:
```
route:
  group_by: ['alertname']
  group_wait: 30s
  group_interval: 5m
  repeat_interval: 1h
  receiver: 'pushover'
receivers:
  - name: 'pushover'
    pushover_configs:
      - user_key_file: 'pushover_user_key.txt'
        token_file: 'pushover_token.txt'
        priority: '0'
inhibit_rules:
  - source_match:
      severity: 'critical'
    target_match:
      severity: 'warning'
    equal: ['alertname', 'dev', 'instance']
```

* Logs:
```
[...]

PR #5398, opened 2026-07-21
notify/pushover: trim whitespace from token and user key files

dpacgdm:
## Summary
- Trim trailing whitespace/newlines when reading `token_file` and `user_key_file` (same pattern as pagerduty/slack/etc.).
- Editors like vim append a newline by default; Pushover then rejects the credentials with HTTP 400.

Related to https://github.com/prometheus/alertmanager/issues/3641

## Test plan
- [x] `go test ./notify/pushover/ -count=1`
- [x] Existing file-backed secret tests now write secrets with a trailing `\n` and still succeed

### DIFF
--- a/notify/pushover/pushover.go
+++ b/notify/pushover/pushover.go
@@ -95,7 +95,8 @@ func (n *Notifier) Notify(ctx context.Context, as ...*types.Alert) (bool, error)
 		if err != nil {
 			return false, fmt.Errorf("read token_file: %w", err)
 		}
-		token = string(content)
+		// Trim trailing newlines from editors like vim that append them by default.
+		token = strings.TrimSpace(string(content))
 	}
 	if n.conf.UserKey != "" {
 		userKey = string(n.conf.UserKey)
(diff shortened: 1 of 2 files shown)


BradKollmyer:
Confirming from production on 0.33.1, with one detail I don't see recorded here
or in #3641: **this fails completely silently, and it kills paging fleet-wide.**

Three-site Prometheus/Alertmanager fleet, zero pages delivered for six days —
1242 failed notifications on one site alone. Targets up, rules evaluating,
alerts firing and visible in the UI; they just never left the building. The only
trace is a notify error in the journal, which nobody reads because the thing
that tells you to look is the thing that's broken.

The trigger is moving credentials from inline `token`/`user_key` into
`token_file`/`user_key_file` — inline, YAML strips the newline; via the file,
nothing does. So it specifically punishes the more secure option.

Small clarification on the API: Pushover accepts a trailing **space** and
rejects only the **newline** (`{"token":"invalid"}`, or `{"user":"invalid"}` for
the user key). `TrimSpace` covers both, so the fix here is unchanged — but it's
why the file looks fine under inspection, since you see 30 plausible characters.
The tell is `wc -c` reporting 31.

Also worth noting for review: `prometheus/common`'s `FileSecret.Fetch` already
`TrimSpace`s file-backed secrets, and this same file `TrimSpace`s the message
body ~20 lines below. The credentials are the exception.

#3641 has been open since December 2023 and this is 8 lines with tests — would
be great to see it land.

TheMeier:
I think it makes sense to add a separate new test instead of modifing the existing one because files without an extra newline are not properly covered anymore. Also Commit signature is missing. Please use the [existing PR template](https://github.com/prometheus/alertmanager/blob/main/.github/pull_request_template.md)

Issue #4346, opened 2025-04-09
Feature request: support read configuration from http endpoint


chenlujjj:
In a large team, maintaining the config file for alert manger can be challenging, so many receivers and route rules.

We have a idea, to build a http service exposing APIs for different team members to read and update their config, the service also expose the whole configuration content in an endpoint, which can be read by alert manger

siavashs:
This is a bad idea, what if the https endpoint is not available on Alertmanager startup?
Alertmanager is a critical system and should not have such dependencies.

But if you still want to load the configuration from an HTTP endpoint, you could use an external provisioner to load the config and make it available to Alertmanager as a file.

I'm closing this issue since I don't think such a feature will be implemented, and there are already external workarounds to achieve the same result.

chenlujjj:
@siavashs Thanks for reply.

Yes, we've used a sidecar to download config from a http server and serve to Alertmanager.

SoloJacobs:
Re-opening this, just so users can report there use cases and we continue gathering feedback from the community. This doesn't mean this feature was approved, in fact we have actively declined the feature here: https://github.com/prometheus/alertmanager/pull/5405

PR #5405, opened 2026-07-27
config: Adding support to read configuration from http endpoint

jshah-dev:
## Description

This PR adds support for loading the Alertmanager configuration from an HTTP endpoint as an alternative to the existing local file-based configuration.

A new command-line option is introduced:

```text
--config.http-url=<url>
```

`--config.http-url` and `--config.file` are mutually exclusive configuration sources.

When `--config.http-url` is configured, Alertmanager retrieves the configuration using an HTTP `GET` request and passes the retrieved configuration through the existing configuration parsing, validation, and application flow.

The HTTP configuration is fetched:

- During Alertmanager startup.
- When the existing `POST /-/reload` endpoint is invoked.

### Motivation

Alertmanager currently expects its configuration to be available on the local filesystem.

In environments where configuration is centrally managed or exposed through a configuration service, this requires an additional mechanism such as a sidecar, shared volume, or synchronization process to materialize the remote configuration as a local file.
Supporting an HTTP configuration source allows Alertmanager to consume configuration directly from a remote configuration service while remaining independent of the underlying storage mechanism.

The intent of this change is therefore not to introduce storage-specific configuration support, but to provide a small, generic HTTP-based configuration source.

### Behavior

Existing file-based configuration continues to work unchanged:
```bash
alertmanager --config.file=/etc/alertmanager/alertmanager.yml
```
HTTP-based configuration can instead be configured using:

```bash
alertmanager --config.http-url=https://config.example.com/alertmanager.yml
```
When using the HTTP source, Alertmanager performs:

```text
HTTP GET
    |
    v
Configuration bytes
    |
    v
Existing parsing and validation
    |
    v
Existing configuration application
```

The same configuration source is used when `POST /-/reload` is invoked. For an HTTP configuration source, this results in a fresh HTTP request so that the latest configuration exposed by the endpoint is loaded.

If fetching or validating the new configuration fails during reload, the currently active configuration remains in use.

### Scope

This change intentionally keeps HTTP configuration loading minimal.

It does **not** introduce:

- Periodic HTTP polling.
- HTTP caching.
- S3/Git/storage-specific integrations.
- Configuration version management.

These capabilities can be considered independently if needed.

### Backward Compatibility

The existing `--config.file` workflow and configuration reload behavior remain unchanged.

This feature is opt-in through `--config.http-url` and does not alter the behavior of existing Alertmanager deployments.

## Testing

Tests have been added for the new HTTP configuration source, including:

- Successful configuration loading over HTTP.
- HTTP error responses.
- Invalid remote configuration.
- Configuration loading during startup.
- Reloading updated HTTP configuration through `POST /-/reload`.
- Preservation of the existing configuration when an HTTP reload fails.
- Mutual exclusion between `--config.file` and `--config.http-url`.
- Existing file-based configuration behavior.

HTTP tests use local test HTTP servers and do not depend on external services.

#### Pull Request Checklist

    - Fixes #<issue number>
    
- Is this a bugfix?
    - [] I have added tests that can reproduce the bug which pass with this bugfix applied
- Is this a new feature?
- Is this a breaking change?

```release-note
[FEATURE] Add support for loading Alertmanager configuration from an HTTP endpoint using the new `--config.http-url` option. Remote configuration is loaded at startup and refreshed when the existing `/-/reload` endpoint is invoked.
```


app/http_config_test.go
```diff
@@ -0,0 +1,195 @@
+// Copyright 2024 Prometheus Team
```
TheMeier: make: *** [Makefile.common:146: common-check_license] Error 1

please use the license header without date. eg https://github.com/prometheus/alertmanager/blob/main/config/config.go#L1-L12

jshah-dev: Thank you @TheMeier for suggestions, will do the changes

app/app.go
```diff
@@ -457,7 +471,13 @@ func (a *App) setup() error {
 	})
 
 	configLogger := logger.With("component", "configuration")
+	if opts.ConfigHTTPURL != "" {
+		loader = config.NewHTTPLoader(opts.ConfigHTTPURL)
```
Spaceman1701: does this need to be reconstructed? It seems like we could use the `loader` that's assigned on either `253` or `258`.

app/config_loader_test.go
```diff
@@ -0,0 +1,195 @@
+// Copyright The Prometheus Authors
```
Spaceman1701, line 1: small nitpick: this seems to be tests for all sorts of config loading, not just `http_config` - I think it would make more sense for this file to be called `config_loader_test.go` or something along those lines.

cmd/alertmanager/main.go
```diff
@@ -97,6 +98,14 @@ func run() int {
 	kingpin.CommandLine.GetFlag("help").Short('h')
 	kingpin.Parse()
 
+	// Validate exactly one configuration source is provided.
+	if *configFile == "" && *configHTTPURL == "" {
+		kingpin.Fatalf("Need to configure one of the following --config.file or --config.http-url")
```
Spaceman1701: I believe this is a behavior regression - there may be production users of Alertmanager who depend on the default value of the `--config.file` flag. We cannot change this behavior without breaking those users.

config/coordinator.go
```diff
@@ -43,9 +46,20 @@ type Coordinator struct {
 // NewCoordinator returns a new coordinator with the given configuration file
 // path. It does not yet load the configuration from file. This is done in
 // `Reload()`.
-func NewCoordinator(configFilePath string, r prometheus.Registerer, l *slog.Logger) *Coordinator {
+func NewCoordinator(loader ConfigLoader, configFilePath string, r prometheus.Registerer, l *slog.Logger) *Coordinator {
+	// Determine the source string for logging
+	source := configFilePath
```
Spaceman1701: this is a bit awkward - perhaps it could be modeled differently. What if `ConfigLoader` required a `Source` method that exposed this value polymorphically? If we do it that way, the `Coordinator` doesn't need to know what the concrete implementation of `loader` actually is.

As it's written now, this is a potential cause of breakage if future changes add new `ConfigLoader` implementations.

config/loader.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unexpected HTTP status %d", resp.StatusCode)
+	}
+
+	// Limit response body size to prevent memory issues
+	// 10MB should be more than enough for any reasonable configuration
+	const maxConfigSize = 10 * 1024 * 1024 // 10MB
```
Spaceman1701: is this a necessary limit? The file loader doesn't appear to have one. I'd lean towards simplicity and leave the limit out.

Spaceman1701 (commented):
Thanks for the contribution! I think this is an interesting idea.

I've left some comments for a few things which I'd want to see fixed.

jshah-dev:
> Thanks for the contribution! I think this is an interesting idea.
> 
> I've left some comments for a few things which I'd want to see fixed.

@Spaceman1701 Thank you for your valuable feedback & comments. I will review it & try doing necessary fixes

jshah-dev:
@Spaceman1701 Thanks for the feedback. I’ve addressed the suggestions in the latest changes. 
Please take another look when you get a chance.

jshah-dev:
Hi @Spaceman1701 @TheMeier , following up on this PR. I’ve incorporated the recently suggested improvements. The latest changes are now pushed and ready for another look. 
It would be great if you could spare some of your valuable time to review changes and share your feedback so we can move the PR forward.

SoloJacobs:
@jshah-dev We have recently updated the golanglint-ci version. You need to install a newer version locally as well.

siavashs:
We rejected this feature request #4346 in the past:
- this introduced an unreliable startup behaviour
- there are existing solutions for fetching the configuration and feeding it to Alertmanager on disk

I just wanted to raise the concerns and past context, I leave the decision to accept or reject this feature to other maintainers who are already engaged in the PR.

SoloJacobs:
Hi @jshah-dev ,

We discussed this feature in the Alertmanager Working Group today, and I'm afraid we cannot take this feature. These are the reasons:

* It seems relatively trivial to setup this feature without it being built into Alertmanager.
* It does not match existing functionality in Prometheus.
* It does not align well with loading certificate and secret files.
* The additional reasons aforementioned by siavashs .

Alertmanager does not take every feature, sadly the total of features we can maintain here is rather small. Maybe you have better luck at https://github.com/grafana/prometheus-alertmanager 

I will re-open the issue, so the community has a place to comment on this feature. Maybe it will be re-discussed in the future.

Kind regards

PR #5406, opened 2026-07-27
api: add active/expired/pending filter params to GET /silences

### LINKED ISSUE #5404: Api silences add parameter like alerts
### Proposal

Add parameters to end point api getSilences
      parameters:
        - in: query
          name: active
          type: boolean
          description: Include active silences in results. If false, excludes active silences.
          default: true
        - in: query
          name: expired
          type: boolean
          description: Include expired silences in results. If false, excludes expired silences.
          default: true
        - in: query
          name: pending
          type: boolean
          description: Include pending silences in results. If false, excludes pending silences.
          default: true

mihir-dixit2k27:
Fixes #5404

`GET /api/v2/silences` returns all silences with no way to filter by state. This means callers who only care about active silences have to fetch everything and filter client-side, which is the same problem that was solved for alerts with the existing `active`, `silenced`, `inhibited`, and `unprocessed` params on `GET /alerts`.

This PR adds three optional boolean query params that are  `active`, `expired`, `pending`  to `GET /api/v2/silences`. Omitting a param includes that state (default behaviour is unchanged). Passing `false` excludes it.

The filter is applied server-side after the existing label-matcher filter, before sorting.

[FEATURE] API: Add active, expired, and pending boolean filter parameters to GET /api/v2/silences to allow filtering silences by state server-side.

### DIFF
--- a/api/v2/api.go
+++ b/api/v2/api.go
@@ -702,6 +702,25 @@ func (api *API) getSilencesHandler(params silence_ops.GetSilencesParams) middlew
 		sils = append(sils, &silence)
 	}
 
+	// Filter by silence state. params.Active/Expired/Pending are always non-nil
+	// because NewGetSilencesParams() initialises them to true; the filter is a
+	// no-op unless the caller explicitly passes false.
+	var filtered open_api_models.GettableSilences
+	for _, sil := range sils {
+		state := string(*sil.Status.State)
+		if !*params.Active && state == "active" {
+			continue
+		}
+		if !*params.Expired && state == "expired" {
+			continue
+		}
+		if !*params.Pending && state == "pending" {
+			continue
+		}
+		filtered = append(filtered, sil)
+	}
+	sils = filtered
+
 	SortSilences(sils)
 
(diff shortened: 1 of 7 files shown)


api/v2/api.go
```diff
@@ -702,6 +702,23 @@ func (api *API) getSilencesHandler(params silence_ops.GetSilencesParams) middlew
 		sils = append(sils, &silence)
 	}
 
+	// Filter by silence state. Defaults come from NewGetSilencesParams(), but
+	// nil-check defensively in case the handler is invoked without them.
+	filtered := open_api_models.GettableSilences{}
+	for _, sil := range sils {
+		state := string(*sil.Status.State)
+		if params.Active != nil && !*params.Active && state == "active" {
```
TheMeier: Maybe better use these constants https://github.com/prometheus/alertmanager/blob/main/silence/state.go#L19 instead of bare strings?

api/v2/api.go
TheMeier: can the filters already be applied here?

TheMeier (approved):

Spaceman1701 (approved):

PR #5409, opened 2026-07-28
feat(eventrecorder): replace event output schema

siavashs:
Replace the original event recorder schema with events/v2. Alert labels,
alert annotations, group labels, silence annotations, and muted-alert labels
are now encoded as maps instead of nested ordered label pairs.

This is a breaking change for every event recorder output. JSON consumers must
handle the new map-based fields, and protobuf consumers must regenerate their
bindings from proto/eventrecorder/events/v2/events.proto.

Remove the legacy eventrecorderpb schema and generated bindings. Register the
new schema as the event recorder Buf module and use it directly for all file,
webhook, Kafka, and stdout outputs without version selection or conversion.

Event producers now construct opaque eventrecorder.Event values through
snapshotting constructors instead of depending on protobuf types. Destinations
receive the typed Event and serialize it as JSON or protobuf. Event metadata is
attached without mutating the constructed payload.

Also include silence annotations in recorded events.

BREAKING CHANGE: Event recorder outputs now use the events/v2 schema and
map-based label and annotation fields. The legacy eventrecorderpb wire format
is no longer supported.

```release-notes
[CHANGE] eventrecorder: The output data format now uses the new `events/v2` schema. This is a breaking change: alert labels, alert annotations, group labels, silence annotations, and muted-alert labels are JSON maps, and protobuf consumers must use the new schema.
```


MichaHoffmann (approved):

eventrecorder/kafka.go
```diff
@@ -224,7 +223,7 @@ func (ko *KafkaOutput) Name() string { return ko.name }
 // SendEvent serializes the event in the configured format (JSON or
 // protobuf) and queues it for asynchronous delivery.  It returns the
 // serialized size (for the bytes-written metric).
-func (ko *KafkaOutput) SendEvent(event *eventrecorderpb.Event) (int, error) {
+func (ko *KafkaOutput) SendEvent(event proto.Message) (int, error) {
```
Spaceman1701: I wonder if there could be a common interface that's more specific than `proto.Message` to user here. As it is, it makes the API a fair bit less type safe.

Spaceman1701 (commented):
I think this change is good in principle, but I don't love that we have to maintain a mapping between v1 and v2 events... I'm generally a little nervous about introducing v2 in general.

How would you feel about changing the v1/v2 selection to be at event construction time? Each `New*Event` function could pick a v1 or v2 version dynamically. This would at least make it harder to forget about the v1 events when making a change.

siavashs:
> I think this change is good in principle, but I don't love that we have to maintain a mapping between v1 and v2 events... I'm generally a little nervous about introducing v2 in general.

My assumption was that eventrecorder v1 will be deprecated and removed until we reach release v1, then there will be no translation until we have v3, etc.

> How would you feel about changing the v1/v2 selection to be at event construction time? Each `New*Event` function could pick a v1 or v2 version dynamically. This would at least make it harder to forget about the v1 events when making a change.

That is fine with me.

eventrecorder/kafka.go
siavashs: Outputs shouldn't really care about the message contents, that what the `proto.Message` does here.
Considering you other comment with this one, if we want to make the outputs use a mode specific type, then we have to delay the v1/v2 translation to be requested by the outputs before encoding.

Spaceman1701:
> > How would you feel about changing the v1/v2 selection to be at event construction time? Each `New*Event` function could pick a v1 or v2 version dynamically. This would at least make it harder to forget about the v1 events when making a change.
> 
> That is fine with me.

Alright, that'd be my preference then - I just want to make sure it's easy and safe to add new events. And I'm fine with the idea of deprecating v1 pretty quickly - I doubt it's widely used at this point (though I'm not sure if we're considering it experimental).

eventrecorder/recorder.go
```diff
@@ -296,39 +293,33 @@ func (c *sharedRecorder) marshalAndSend(req writeRequest, outputs []Destination)
...
-		Instance:  r.core.instance,
-		Data:      event,
-	}
-
+	clusterPosition := uint64(0)
 	if peer := r.core.peer.Load(); peer != nil {
-		wrappedEvent.ClusterPosition = uint32(peer.Position())
+		clusterPosition = uint64(peer.Position())
 	}
+	event = event.withMetadata(timestamppb.Now(), r.core.instance, clusterPosition)
```
Spaceman1701: I think this is a set backwards in terms of modeling the event - I'm not sure that we should be using the same type to represent an event both with and without metadata. I think it's better to differentiate between the `EventData` and the `Event`

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## main / (unreleased)
 
+* [CHANGE] eventrecorder: The output data format now uses the new `events/v2` schema. This is a breaking change: alert labels, alert annotations, group labels, silence annotations, and muted-alert labels are JSON maps, and protobuf consumers must use the new schema.
```
Spaceman1701, line 3: this is behind a feature flag right now, right?

Spaceman1701 (commented):
Generally this LGTM, but I do think we should fix the modeling of the event types if we can.

CHANGELOG.md
siavashs: yes, `event-recorder` is the feature flag.

eventrecorder/recorder.go
siavashs: Good suggestion, I will implement this.

PR #5415, opened 2026-07-31
featurecontrol: log auto-gomemlimit activation at INFO

### LINKED ISSUE #5412: featurecontrol: feature-flag activation is logged at WARN, so a clean start-up emits warnings
### What did you do?

Started Alertmanager v0.33.1 with three feature flags that are required for correct operation in our
deployment:

```
--enable-feature=utf8-strict-mode,auto-gomemlimit,receiver-name-in-metrics
```

A clean start-up log. Enabling a flag we asked for is an acknowledgement, not a warning.

### What did you see instead?

Every one of the three activations is logged at `WARN`, so a healthy start-up emits three warning
lines and nothing else notable:

```json
{"time":"2026-07-30T11:27:37.086790131Z","level":"WARN","source":"featurecontrol.go:160","msg":"UTF-8 strict mode enabled"}
{"time":"2026-07-30T11:27:37.086818731Z","level":"WARN","source":"featurecontrol.go:163","msg":"Automatically set GOMEMLIMIT to match the Linux container or system memory limit."}
{"time":"2026-07-30T11:27:37.086873731Z","level":"WARN","source":"featurecontrol.go:151","msg":"Experimental receiver name in metrics enabled"}
```

In `featurecontrol/featurecontrol.go` at v0.33.1 every branch uses `logger.Warn(...)`
unconditionally — `alert-names-in-metrics`, `receiver-name-in-metrics`, `group-key-in-metrics`,
`classic-mode`, `utf8-strict-mode`, `auto-gomemlimit` and `event-recorder`. There is no level
override and no way to reach the same functionality without the line.

### Why this matters beyond tidiness

`level=WARN` is what monitoring and log pipelines are configured to escalate on. A component that
emits warnings on every clean start teaches its operators that its warnings are noise — and the one
warning that matters here is the delivery-failure retry line, which is also the line that leaks the
Telegram bot token (see #5254). Anything that trains a reader to skip Alertmanager's warnings makes
that worse.

Two of the three flags above are not even experimental in effect: `auto-gomemlimit` is how the Go
heap learns about a container memory limit, and running without it in a limited container is the
worse choice. Announcing the safer configuration as a warning inverts the signal.

### Suggested change

Log the activation of a stable, deliberately requested feature at `INFO`. Keep `WARN` for the flags
whose message actually carries a caution — the unbounded-cardinality ones
(`alert-names-in-metrics`, `group-key-in-metrics`) have a real operational cost and deserve it.

### Environment

* Alertmanager version: `0.33.1` (`revision=2c8da51e03f3dbbed24f9711ca2d76aab4eef9c5`, `go1.26.4`,
  `linux/amd64`), image `quay.io/prometheus/alertmanager@sha256:9e082985f56f4c8c9f724e18f2288c6708f472e56a5286b8863d080434ea065d`
* Configuration: `--log.format=json --log.level=info`, single replica, `--cluster.listen-address=`

byBrodowski:
Narrowing #5412 to the one part that isn't a matter of taste.

@TheMeier's point in the issue is fair: whether a warning-heavy start-up log is a problem at all
depends on how you read logs, and I'm not going to argue anyone out of their preference. But the
agreement in that reply — that `auto-gomemlimit` should be promoted — doesn't depend on taste, so
here it is as a one-line change instead of a longer discussion.

`auto-gomemlimit` is how the Go heap learns about a container memory limit. In a memory-limited
container, running *without* it is the worse of the two configurations. Logging its activation at
`WARN` therefore announces the safer choice as a caution, which is backwards regardless of how
loudly you think a log should speak.

#### What this changes

One `switch` branch in `featurecontrol/featurecontrol.go`:

```go
case FeatureAutoGOMEMLIMIT:
	opts = append(opts, enableAutoGOMEMLIMIT())
-	logger.Warn("Automatically set GOMEMLIMIT to match the Linux container or system memory limit.")
+	logger.Info("Automatically set GOMEMLIMIT to match the Linux container or system memory limit.")
```

The message text is unchanged. Each feature has its own branch with its own log call, so this
reaches nothing else — in particular the unbounded-cardinality metrics flags
(`alert-names-in-metrics`, `group-key-in-metrics`) keep `WARN`, which I think is right: their
message carries a real operational cost, so a caution is the correct level for them. The
`Experimental …` messages also keep `WARN`. I've deliberately left the wider question in #5412
alone; if a maintainer wants the same treatment for `utf8-strict-mode`, that's a separate call and
I'm happy to follow up.

#### Verification

`go build`, `go test` and `go vet` on `./featurecontrol/` pass with the change applied (Go 1.25,
matching `go.mod`). `featurecontrol_test.go` asserts the resulting flag state and makes no
assertion about log level, so it neither needed nor received a change — the behaviour under test is
identical.

#### Pull Request Checklist

- Fixes #5412
- Is this a new Receiver integration? No.
- Is this a bugfix? No — a log-level change, no behaviour change to test.
- Is this a new feature? No.
- Does this change affect performance? No.
- Is this a breaking change? No. No cluster message or API surface is touched.

```release-notes
[CHANGE] Log activation of the `auto-gomemlimit` feature flag at INFO instead of WARN. #5412
```

### DIFF
--- a/featurecontrol/featurecontrol.go
+++ b/featurecontrol/featurecontrol.go
@@ -160,7 +160,7 @@ func NewFlags(logger *slog.Logger, features string) (Flagger, error) {
 			logger.Warn("UTF-8 strict mode enabled")
 		case FeatureAutoGOMEMLIMIT:
 			opts = append(opts, enableAutoGOMEMLIMIT())
-			logger.Warn("Automatically set GOMEMLIMIT to match the Linux container or system memory limit.")
+			logger.Info("Automatically set GOMEMLIMIT to match the Linux container or system memory limit.")
 		case FeatureEventRecorder:
 			opts = append(opts, enableEventRecorder())
 			logger.Warn("Experimental event recorder enabled")


TheMeier:
Thanks @byBrodowski. By promtion I meant move it from a feature-flag to a normal config flag

SoloJacobs:
@TheMeier I don't think we should promote this flag, since there is a good chance this feature will it into the runtime at some point.

SoloJacobs (changes requested):
Had a look, at it seems that we won't get this feature anytime soon:
https://github.com/golang/go/issues/78912
https://github.com/golang/go/issues/75164
So, promoting it might make sense. I'm not sure what the correct command-flags would be here.

@byBrodowski I don't think we will allow for some feature flags to log differently than others, consistency is important.

TheMeier:
how abuot `--auto-gomemlimit.enable` there is already `--auto-gomemlimit.ratio` to configure it's target

TheMeier:
IMHO this is adressed by the change in https://github.com/prometheus/alertmanager/issues/5463. Starting with next release the feature flag will emit an error and users should switch to use `--auto-gomemlimit` command line flag.

PR #5448, opened 2026-08-13
config: keep local SMTP auth secret overrides

sueun-dev:
A receiver-level `auth_secret` should override the global SMTP auth secret defaults. The loader currently merges `auth_secret` and `auth_secret_file` independently, though, so a receiver with a local inline secret also inherits `global.smtp_auth_secret_file`.

That leaves both fields set. `getAuthSecret` checks the file first, so the receiver's inline secret is not used. If the global path is unavailable, the notification fails while trying to read it; if it is available, CRAM-MD5 uses the global file contents instead of the receiver override.

This restores paired inheritance: copy the global secret fields only when neither receiver field is set, matching the existing password/password-file handling. The global/local SMTP fixture now covers a global secret file with both local inline-secret and local secret-file overrides.

Tested with:

- `go test ./config/ -run '^TestGlobalAndLocalSMTPPassword$' -count=1`
- `go test ./config/... ./notify/email/... -count=1`
- `go test -race ./config/... ./notify/email/... -count=1`
- `go vet ./config/... ./notify/email/...`
- `go build ./config/... ./notify/email/... ./cmd/amtool`

#### Pull Request Checklist

    - None
    - [ ] I have already tried to use the Webhook Receiver Integration and third-party integrations before adding this new Receiver Integration
- Is this a bugfix?
- Is this a new feature?
- Is this a breaking change?
    - Not needed: the current docs already say `auth_secret` and `auth_secret_file` are mutually exclusive.

```release-notes
[BUGFIX] Email: Honor a receiver's inline `auth_secret` when a global `smtp_auth_secret_file` is configured.
```

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -408,8 +408,10 @@ func (c *Config) UnmarshalYAML(unmarshal func(any) error) error {
 				ec.AuthPassword = c.Global.SMTPAuthPassword
 				ec.AuthPasswordFile = c.Global.SMTPAuthPasswordFile
 			}
-			ec.AuthSecret = cmp.Or(ec.AuthSecret, c.Global.SMTPAuthSecret)
-			ec.AuthSecretFile = cmp.Or(ec.AuthSecretFile, c.Global.SMTPAuthSecretFile)
+			if ec.AuthSecret == "" && ec.AuthSecretFile == "" {
+				ec.AuthSecret = c.Global.SMTPAuthSecret
+				ec.AuthSecretFile = c.Global.SMTPAuthSecretFile
+			}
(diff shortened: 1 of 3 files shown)


TheMeier (approved):

SoloJacobs (approved):
@TheMeier Hmm, looks like we broke this one: https://github.com/prometheus/alertmanager/pull/4812
Perhaps there are more similar regressions?

TheMeier:
> @TheMeier Hmm, looks like we broke this one: #4812 Perhaps there are more similar regressions?

Yes very likely. :( I really think we should make a formal abstraction for merging configs. Something like this https://github.com/prometheus/alertmanager/commit/7454fd545cf1b61ed0ebeaf91d90992601a25c80

PR #4607, opened 2025-10-13
fix: improve inhibition performance

siavashs:
This change adds a new index per inhibition rule which:
1. extracts the subset of source alert labelset which are in equals
2. calculates the fingerprint of the above
3. maps the calculated fingerprint to the source alert fingerprint
4. performs the same calculation for target alerts
5. uses the above index to find the equal source alerts quickly

This significantly improves the inhibition performance, since there is no need to loop over all source alerts and the equal labels.

The equals index items are garbage collected by callback from `scache`.

Fixes #4606 

<details><summary>Benchmarks</summary>
<p>

```go
goos: darwin
goarch: arm64
pkg: github.com/prometheus/alertmanager/inhibit
cpu: Apple M3 Pro
                                                      │ benchmark-inhibit-base.txt │     benchmark-inhibit-fixed.txt     │
                                                      │           sec/op           │    sec/op     vs base               │
Mutes/1_inhibition_rule,_1_inhibiting_alert-12                         518.5n ± 1%   485.3n ± 11%   -6.40% (p=0.024 n=7)
Mutes/10_inhibition_rules,_1_inhibiting_alert-12                       520.7n ± 1%   484.7n ±  2%   -6.91% (p=0.001 n=7)
Mutes/100_inhibition_rules,_1_inhibiting_alert-12                      537.4n ± 0%   513.7n ±  9%        ~ (p=0.165 n=7)
Mutes/1000_inhibition_rules,_1_inhibiting_alert-12                     672.6n ± 5%   604.0n ±  3%  -10.20% (p=0.001 n=7)
Mutes/10000_inhibition_rules,_1_inhibiting_alert-12                    742.7n ± 4%   702.6n ±  6%   -5.40% (p=0.001 n=7)
Mutes/1_inhibition_rule,_10_inhibiting_alerts-12                       655.3n ± 1%   547.4n ±  7%  -16.47% (p=0.001 n=7)
Mutes/1_inhibition_rule,_100_inhibiting_alerts-12                     1292.0n ± 2%   558.3n ±  5%  -56.79% (p=0.001 n=7)
Mutes/1_inhibition_rule,_1000_inhibiting_alerts-12                    8585.0n ± 1%   586.4n ±  2%  -93.17% (p=0.001 n=7)
Mutes/1_inhibition_rule,_10000_inhibiting_alerts-12                  68664.0n ± 1%   577.4n ±  2%  -99.16% (p=0.001 n=7)
[...]

PR #5449, opened 2026-08-14
inhibit: preserve source-only matches in equal-label index

### LINKED ISSUE #5162: inhibitions unable to handle regex

sueun-dev:
#### Pull Request Checklist

    - Fixes #5162.
    - Related to #5174.
    - [ ] I have already tried to use the Webhook Receiver Integration and 3rd party integrations before adding this new Receiver Integration
- Is this a bugfix?
- Is this a new feature?
        - Same benchmark on the previous PR commit with only the benchmark case applied, compared with this branch: raw `go test -benchmem` geomean over 5 runs was 172555 ns/op -> 6492 ns/op, 83048 B/op -> 1048 B/op, 32 -> 27 allocs/op.
- Is this a breaking change?
- [ ] I have added/updated the required documentation (not needed; this restores the documented self-inhibition behavior)

```release-notes
[BUGFIX] Inhibition: Keep equal-label source alerts indexed correctly when multiple source alerts share the same equal labels.
```


SoloJacobs:
@siavashs I believe this is a regression introduced in https://github.com/prometheus/alertmanager/pull/4607 . Could you have a look?

inhibit/inhibit.go
```diff
@@ -397,6 +397,24 @@ func (r *InhibitRule) findEqualSourceAlert(lset model.LabelSet, now time.Time) (
...
+			continue
+		}
+		if excludeTwoSidedMatch && r.TargetMatchers.Matches(alert.Labels) {
+			continue
+		}
+		return alert, true
+	}
+
+	return nil, false
+}
```
siavashs: This scans the whole cache whenever the indexed source is two-sided which is O(number of source alerts), allocates a copy, locks the shared cache mutex, etc.
This is basically skipping the optimisations #4607 introduced.
Also the current benchmark matrix does not cover this case.

inhibit/inhibit.go
```diff
@@ -412,7 +430,11 @@ func (r *InhibitRule) hasEqual(lset model.LabelSet, excludeTwoSidedMatch bool, n
 	equal, found := r.findEqualSourceAlert(lset, now)
 	if found {
 		if excludeTwoSidedMatch && r.TargetMatchers.Matches(equal.Labels) {
-			return model.Fingerprint(0), false
+			equal, found = r.findEqualSourceAlertFromCache(lset, excludeTwoSidedMatch, now)
```
siavashs: The new method is only called when a source alert is found in cache initially but is disqualified.
The index can be absent while same-equal active sources remain.
`gcCallback` deletes the whole equal-label key when any alert in that bucket is collected.
So we need a regression test where GC removes a non-indexed same-equal alert while a source-only alert remains active. Ideally the index should remove a specific source fingerprint rather than deleting the bucket.

siavashs (changes requested):
I think the intention of the pull request is correct but the implementation is not sufficient to resolve the issue.

A better structure would either:
- index all source fingerprints per equal-label key
- or separately retain an “any source” and “source-only source” candidate.

I think the best would be a hybrid model maybe, something like this:

```
equal-label key
    ├── all source members
    ├── best any-source candidate
    └── best source-only candidate
```

This would not be very efficient for memory but it would be correct when we consider GC.

cc @Spaceman1701

sueun-dev:
Thanks, yes, that was the missing case. I also rechecked #5162/#5174; this now covers the same GC bucket case without scanning the whole source cache.

I reworked the index to keep all source alerts for each equal-label key and separately track the best any-source and source-only candidates. `gcCallback` now removes a collected alert only if that exact alert is still the indexed member, then rebuilds the bucket only when the removed alert was selected.

I added the GC regression you described, a same-fingerprint refresh regression for the GC callback path, and a same-equal/source-only benchmark case.

Checks:
- `go test ./inhibit -run 'TestInhibitRuleGCCallbackDoesNotRemoveRefreshedSameFingerprintSourceAlert|TestInhibitRuleHasEqualKeepsSourceOnlyAlertAfterGCSameEqual' -count=1`
- `go test ./inhibit -run 'TestInhibitRuleHasEqual|TestInhibitRuleHasEqualKeepsSourceOnlyAlertAfterGCSameEqual|TestInhibitRuleGCCallbackDoesNotRemoveRefreshedSameFingerprintSourceAlert' -count=20`
- `go test ./inhibit -count=1`
- `go test ./inhibit -race -count=10`
- `go test ./inhibit ./store ./provider/mem -count=1`
- `go test ./store ./provider/mem -race -count=3`
- `go vet ./inhibit ./store ./provider/mem`
- `go test ./inhibit -run '^$' -bench 'BenchmarkMutes/1_inhibition_rule,_10000_same-equal_alerts,_source-only_candidate$' -benchmem -benchtime=25x -count=5`

For that benchmark, the previous PR commit with only the benchmark case applied was 172555 ns/op, 83048 B/op, 32 allocs/op by raw geomean over 5 runs. This branch was 6492 ns/op, 1048 B/op, 27 allocs/op.

SoloJacobs:
Hmm, from what I can tell the new PR is actually a little bit slower. Test [here](https://github.com/SoloJacobs/alertmanager/tree/test-inhibit) 
```
❯ ./run.sh
=== inhibit
flushes: 238231
Mutes calls: 297539
alerts muted: 197578 (66.40%)
flushes fully muted: 174903 (73.42%)
--- PASS: TestReplay (4.82s)
ok  	github.com/prometheus/alertmanager/inhibit	4.831s
=== patchedinhibit
flushes: 238231
Mutes calls: 297539
alerts muted: 201516 (67.73%)
flushes fully muted: 178227 (74.81%)
--- PASS: TestReplay (4.81s)
ok  	github.com/prometheus/alertmanager/patchedinhibit	4.825s
=== v29inhibit
flushes: 238231
Mutes calls: 297539
alerts muted: 201516 (67.73%)
flushes fully muted: 178227 (74.81%)
--- PASS: TestReplay (10.03s)
ok  	github.com/prometheus/alertmanager/v29inhibit	10.042s
=== 5449updatedinhibit
flushes: 238231
Mutes calls: 297539
alerts muted: 201516 (67.73%)
flushes fully muted: 178227 (74.81%)
--- PASS: TestReplay (5.56s)
ok  	github.com/prometheus/alertmanager/5449updatedinhibit	5.572s
```
My test is not super solid: It uses `synctest` to fastforward time, and there is a bunch of overhead from the provider.
`patchedinhibit` included both the previous 5449 and 5174

Spaceman1701:
Hi! Thanks for putting this together, and for documenting the issue.

Taking a look at the PR, I think it's inevitable that we'll have a small performance regression in order to get correct behavior, but I think that's ok. I do think that there might be a little bit of a less complex indexing scheme that we can use, however. 

How would you feel about starting with just the reproducing test case as a PR? If you add `t.Skip`, we can have the test in the repo, and un-skip it when we accept a fix.

Spaceman1701:
I took a pass at this based on HRT's internal implementation of the inhibitor, and I think #5542 is a little bit simpler than this approach right now.

SoloJacobs (approved):
Hi @sueun-dev ,

Looks like the tests you wrote are revealing another bug in the implementation. Thanks again for this contribution. I'll land them now, so everyone can benefit.

Kind regards

Issue #5247, opened 2026-05-20
Improve Alertmanager Semantics for Partially Muted Alert Groups


Spaceman1701:
### Proposal

This is going to be a long issue 😉

## How Things Work Today

Before we talk about what I think we should change, let me give a brief overview of how notifications work in Alertmanager.

Alertmanager notifications are driven by the Notification Pipeline that’s defined in `notify.go`. Every `group_interval` (more or less), an entire _Alert Group_ of alerts is flushed into the notification pipeline. Alerts are captured at the point of flush, so they cannot change during the execution of the pipeline. This essentially gives the pipeline a point-in-time view of a single group of alerts.

The pipeline completes several steps. At a high level, there are basically three operations: 

1. Apply any muting behavior (silences, inhibits, time intervals)
2. Compare the current state of the Alert Group to the previous state
3. Send the notification and record the result

My focus is on the first two. 

For muting stages, the alerts in the Alert Group are processed and Alertmanager decides which ones are currently muted. Alerts which are muted are removed from the Alert Group for the remainder of the pipeline. For example, if my alertgroup is made of `A`, `B`, and `C`, if `B` is silenced, the pipeline will continue with just `A` and `C`. Since muting happens before comparison to the previous group state, we compare only active alerts. 

Comparison is fairly simple, at least conceptually. Alertmanager writes every successful notification into the Notification Log. Each notification is represented by a log entry: 

https://github.com/prometheus/alertmanager/blob/9aae092dab0eb454a53e10b151295739db7334a0/nflog/nflogpb/nflog.proto#L19-L40

As you can see, the log entry essentially models the state of alerts in the group. When the `DedupStage` performs comparison, it reads the latest nflog entry and compares the active and resolved alerts to the current sets of active and resolved alerts. If anything changed, the the `DedupStage` returns a result indicating that the pipeline should continue. If nothing has changed, the `DedupStage` short circuits the pipeline.

## The Problem

The interaction of these two mechanisms creates a problem: muted alerts are never considered when deciding if the state of the Alert Group has changed. In fact, when an alert is muted, it essentially disappears from the nflog completely. This is how Alertmanager has worked for a very long time, but I believe this is fundamentally a bug.

[...]

PR #5470, opened 2026-08-25
test: document/illustrate notification behavior for muted alerts

suprjinx:
This is a tests-only PR that documents and illustrates the current behavior of muted alerts in the notify pipeline. It's expected that these tests will continue to pass when feature-flagged changes are introduced.

- MuteStage already records muted alert hashes in the context, but the time interval stages do not.
- MultiStage short-circuits once the mute stage has emptied the group, so the dedup stage never observes "all alerts are muted".
- A muted alert is dropped from the notification log entry the next time the group is notified about, even though it is still firing.
- An alert that was notified about and is then muted never produces the matching resolved notification (#226), unit tested and end to end.
- A group with a muted alert is reported as entirely resolved while that alert is still firing, and the alert then opens a new sequence once it is unmuted (#5247).

Assisted-by: Claude Code

#### Pull Request Checklist

    - Relates #226 
    - Relates #5247

- Is this a bugfix?
- Is this a new feature?
- Is this a breaking change?


Spaceman1701 (approved):
Thank you for the contribution, these look very good!

test/with_api_v2/acceptance/silence_test.go
```diff
@@ -181,3 +181,66 @@ receivers:
...
+	return &b
+}
+
+func stringPtr(s string) *string {
+	return &s
+}
+
+func dateTimePtr(t strfmt.DateTime) *strfmt.DateTime {
+	return &t
+}
```
Spaceman1701: It looks like these are unused. (this is the source of the linter failures in CI).

suprjinx: will remove!

PR #5471, opened 2026-08-25
ui: make silence creator and comment optional

### LINKED ISSUE #2998: Silence creator and comment  is required in the UI but not in the API
When creating a silence in the UI and leaving the creator and comment empty, you're met with:
Screenshot 2022-07-11 at 14 34 03

However, if you try to execute a silence creation through the API (as long as you have the keys set the value would be empty).

```bash
## Create Silence
curl -X "POST" "http://localhost:9093/api/v2/silences" \
     -H 'Content-Type: application/json; charset=utf-8' \
     -d $'{
  "matchers": [
    {
      "value": "test",
      "isRegex": false,
      "isEqual": true,
      "name": "test"
    }
  ],
  "comment": "",
  "endsAt": "2022-07-11T15:37:27.659Z",
  "startsAt": "2022-07-11T13:37:27.659Z",
  "createdBy": ""
}'
````

This succeeds - and you get:

```json
{
      "id":"73373e3f-a928-450a-ba75-e9254297b483",
      "status":{
         "state":"active"
      },
      "updatedAt":"2022-07-11T13:42:13.444Z",
      "comment":"",
      "createdBy":"",
      "endsAt":"2022-07-11T15:37:27.659Z",
      "matchers":[
         {
            "isEqual":true,
            "isRegex":false,
            "name":"test",
            "value":"test"
         }
      ],
      "startsAt":"2022-07-11T13:42:13.444Z"
   }
   ```

This doesn't seem right and it's kind of inconsistent with what the UI allows - should we validate at an API level that `""` are not allowed?

### PR DESCRIPTION
**Pull request by neoLsH**
#### Pull Request Checklist

    - Fixes #2998
    - [ ] I have already tried to use the Webhook Receiver Integration and 3rd party integrations before adding this new Receiver Integration
- Is this a bugfix?
- Is this a new feature?
- Is this a breaking change?

The API already accepts silences without a `createdBy` or a `comment` (see `TestValidateSilence`, where a valid silence carries neither), but the web UI rejected the silence form unless both fields were filled in. This made it impossible to create a silence from the UI that the API itself happily accepts.

This change drops the client-side non-empty validation for the Creator and Comment fields so the form matches the API, labels both fields as "(Optional)", and removes the now-unused `stringNotEmpty` helper. Matchers, start/end time and duration are still validated as before.

Verified locally with `make test` in `ui/app` (elm-format, elm-review and the elm-test suite all pass, including new `toSilence`/`validateForm` tests) and a full UI build.

```release-notes
[ENHANCEMENT] ui: The silence form no longer requires a creator and a comment, matching the API. #5471
```


w0rm (approved):

SoloJacobs (commented):
Hmm, the change here clearly does not full-fill #2998 as it was currently intended: The intent being to make Creator and Comment mandatory in the API. However, due to that being a breaking change, we will never change api v2 in that way.

I also don't think asserting that a field should be non-empty in the API is all that helpful: What users really want is that they can trace who silenced what and why.

Currently, the UI is doing a bit of work in that direction: It's a polite reminder that this information is important and useful. So, I think for some organizations this change might be a regression: Somebody might be too lazy to put the information in the UI, but they are probably are not malicious enough to craft an API response, just so they don't have to set the field.

Overall, I'm really undecided whether we should take this patch.

neoLsH:
Thanks a lot for taking the time to review, and for the thoughtful perspective — this is really helpful for me as a first-time contributor here.

You're right that I hadn't considered the full picture. When I picked up #2998, I focused narrowly on the UI/API consistency aspect, and since making the fields mandatory in api v2 isn't possible (as you pointed out, it would be a breaking change), relaxing the UI looked to me like the remaining way to resolve the inconsistency. I didn't think deeply enough about the role the current validation plays beyond validation — as a nudge that encourages people to leave traceability information ("who silenced what, and why"), which clearly has real value for organizations running Alertmanager in teams. I can see how removing it entirely could be a regression there, and I appreciate you pointing that out.

I'd love to discuss what the best resolution is, and I'm happy to go in whatever direction the maintainers prefer. A few possibilities I can think of, in case any is useful:

1. **Soft hint instead of a hard block** — keep creator/comment optional, but show a non-blocking note in the form (e.g. *"Consider adding a creator and a comment so others understand why this alert is silenced"*). This keeps the nudge while not rejecting a payload the API accepts. I'd be glad to implement this.
2. **Leave the UI as it is** and instead close #2998 as "won't fix", since the API direction is off the table and the UI strictness is arguably a feature, not a bug.
3. Something else entirely — e.g. pre-filling `createdBy` from context where available — I'm open to ideas.

Thanks again for the guidance. Just let me know which way you'd like to go, and I'll update (or close) the PR accordingly.

w0rm:
It’s kind of strange that the ui check is used as a guardrail for otherwise a valid API request. I think it makes sense to align them one way or another. 

Also the non empty check is not doing that much of a job, the user can enter gibberish there to bypass validation.

SoloJacobs (approved):

SoloJacobs:
Hi @neoLsH ,

I appreciate the offer to help, but your PR already has the right shape, we just need to make a decision. I will ask a third maintainer, then we will see :-)

Kind regards

SoloJacobs:
This was discussed in the WG meeting. Consensus is:
* We can take this PR.
* But the new API/UI will definitely revert back to being more strict about this behaviour.

PR #5474, opened 2026-08-26
nflog: add muted_alerts to the notification log entry

suprjinx:
Entry currently records the alerts that were firing and resolved at the time of the last notification, but not the ones that were muted. The mute stages drop muted alerts before the dedup stage runs, so they are absent from both lists and indistinguishable from alerts that stopped firing.

Add a muted_alerts field alongside them, an IsMutedSubset helper mirroring IsFiringSubset and IsResolvedSubset, and a mutedAlerts parameter on Log(). Entries that predate this field decode with MutedAlerts unset.

Nothing populates the field yet: SetNotifiesStage passes nil, and the dedup stage does not read it. This is the schema and write path only, so that the behavior change can be reviewed on its own.

Generated code regenerated with scripts/genproto.sh.

Assisted-by: Claude Code

#### Pull Request Checklist

    - Fixes #<issue number>
    
- Is this a bugfix?
- Is this a new feature?
- Is this a breaking change?

```release-notes

```

### DIFF
--- a/nflog/nflogpb/set.go
+++ b/nflog/nflogpb/set.go
@@ -35,6 +35,17 @@ func (m *Entry) IsResolvedSubset(subset map[uint64]struct{}) bool {
 	return isSubset(set, subset)
 }
 
+// IsMutedSubset returns whether the given subset is a subset of the alerts
+// that were muted at the time of the last notification.
+func (m *Entry) IsMutedSubset(subset map[uint64]struct{}) bool {
+	set := map[uint64]struct{}{}
+	for i := range m.MutedAlerts {
+		set[m.MutedAlerts[i]] = struct{}{}
+	}
+
+	return isSubset(set, subset)
+}
+
 func isSubset(set, subset map[uint64]struct{}) bool {
 	for k := range subset {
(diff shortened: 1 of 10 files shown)


nflog/nflogpb/set.go
```diff
@@ -35,6 +35,17 @@ func (m *Entry) IsResolvedSubset(subset map[uint64]struct{}) bool {
 	return isSubset(set, subset)
 }
 
+// IsMutedSubset returns whether the given subset is a subset of the alerts
+// that were muted at the time of the last notification.
+func (m *Entry) IsMutedSubset(subset map[uint64]struct{}) bool {
```
Spaceman1701: I'm not sure we'll end up needing this function, I think we should leave it out until it's clear that it'll be used.

Spaceman1701 (commented):
Thanks for the contribution, generally looks good!

nflog/nflogpb/set.go
suprjinx: removed

Spaceman1701 (approved):

Issue #5486, opened 2026-08-29
UI: build and serve the Mantine application at /ui/


siavashs:
Parent: #5451

## Summary

Build, embed, and release the Mantine application alongside the Elm UI, serving it permanently at `<route-prefix>/ui/` with clean client-side routes.

The existing Elm application remains at the route-prefix root and remains the default UI.

## Build and packaging

- Add first-class Mantine build, test, and lint targets to the root Makefile without duplicating expensive work.
- Configure Vite for hashed assets, third-party license attribution, and gzip/Brotli output compatible with the existing Go asset server.
- Preserve a small uncompressed `index.html` as a no-cache runtime template while precompressing immutable JS, CSS, fonts, images, and license assets.
- Embed both `ui/app/dist` and `ui/mantine-ui/dist` in Alertmanager binaries.
- Update CI artifacts and release builds to carry both distributions.
- Preserve the current Elm prebuilt-assets tarball extraction contract; add a separate Mantine prebuilt-assets tarball rather than silently changing its layout.

## Runtime paths and routing

Register route-prefix-aware handlers for:

- `/ui` → `/ui/`.
- `/ui/` → the Mantine index.
- `/ui/assets/*` → immutable static assets.
- `/ui/*` → the Mantine index for valid clean client routes and direct refreshes.

The SPA fallback must be bounded to the UI mount and must not replace missing assets or shadow `/`, `/favicon.ico`, `/assets/*`, `/api/*`, `/-/*`, `/metrics`, or `/debug/*`.

Inject an escaped `<base href="<route-prefix>/ui/">` and non-inline bootstrap metadata for UI and API base paths into the no-cache index response. This must support nested routes such as `/ui/silences/:id/edit`, arbitrary route prefixes, and CSP hardening without inline executable configuration.

Add a route-prefix-safe “New UI” link to Elm and a “Classic UI” link to Mantine.

## Acceptance criteria

- [ ] One Alertmanager binary serves Elm at `<route-prefix>/` and Mantine at `<route-prefix>/ui/`.
- [ ] Root and non-root route prefixes are covered.
- [ ] Direct loads and refreshes work for `/ui/status`, `/ui/silences/:id`, and `/ui/silences/:id/edit`.
- [ ] Missing `/ui/assets/*` files return 404 and are not replaced by the SPA shell.
- [ ] Dynamic HTML is no-cache and safely injects the runtime base paths.
[...]

PR #5495, opened 2026-08-29
build(ui): add Mantine build pipeline

siavashs:
## Summary

- add first-class Mantine build, test, lint, and clean targets
- include the Mantine build, tests, and lint checks in the corresponding root Make targets
- rebuild Mantine output unconditionally instead of relying on timestamp invalidation
- configure Vite to emit relative hashed assets
- generate gzip and Brotli variants for immutable assets
- generate bundled third-party license notices
- remove the broken reference to the nonexistent `/src/favicon.svg`
- update dedicated UI CI to build, test, and lint through the root Make targets
- build the Mantine distribution in common CI without publishing it
- defer archive layout, Go embedding, and runtime serving to a follow-up change

The Mantine UI is intentionally not included in release artifacts or release binaries before Alertmanager v1.0. The follow-up runtime integration will preserve the existing archive name and extraction path while packaging both UIs together.

Part of #5486

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,7 @@ GOLANGCI_LINT_VERSION := v2.13.1
 include Makefile.common
 
 FRONTEND_DIR             = $(BIN_DIR)/ui/app
+MANTINE_UI_DIR            = $(BIN_DIR)/ui/mantine-ui
 TEMPLATE_DIR             = $(BIN_DIR)/template
 DOCKER_IMAGE_NAME       ?= alertmanager
 
@@ -29,16 +30,16 @@ STATICCHECK_IGNORE =
 build-all: assets apiv2 build
 
 .PHONY: build
-build: ui-elm common-build
+build: ui-elm ui-mantine common-build
 
 .PHONY: test
-test: ui-elm common-test
+test: ui-elm ui-mantine-test common-test
 
 .PHONY: lint
-lint: ui-elm common-lint
+lint: ui-elm ui-mantine-lint common-lint
 
 .PHONY: assets
-assets: $(FRONTEND_DIR)/src/Data ui-elm template/email.tmpl
+assets: $(FRONTEND_DIR)/src/Data ui-elm ui-mantine template/email.tmpl
 
 .PHONY: assets-tarball
 assets-tarball: ui-elm
@@ -48,6 +49,18 @@ assets-tarball: ui-elm
 .PHONY: ui-elm
 ui-elm:
 	cd $(FRONTEND_DIR) && $(MAKE) build
+
+.PHONY: ui-mantine
+ui-mantine:
+	cd $(MANTINE_UI_DIR) && $(MAKE) build
+
+.PHONY: ui-mantine-test
+ui-mantine-test:
+	cd $(MANTINE_UI_DIR) && $(MAKE) test
+
+.PHONY: ui-mantine-lint
+ui-mantine-lint:
+	cd $(MANTINE_UI_DIR) && $(MAKE) lint
 
 $(FRONTEND_DIR)/src/Data: api/v2/openapi.yaml
 	cd $(FRONTEND_DIR) && $(MAKE) src/Data
@@ -71,3 +84,4 @@ clean:
 	- @rm -rf template/email.tmpl \
                   api/v2/models api/v2/restapi api/v2/client
 	- @cd $(FRONTEND_DIR) && $(MAKE) clean
+	- @cd $(MANTINE_UI_DIR) && $(MAKE) clean
--- a/ui/mantine-ui/Makefile
+++ b/ui/mantine-ui/Makefile
@@ -1,0 +1,20 @@
+node_modules: package.json package-lock.json
+	@echo ">> installing Mantine UI dependencies"
+	npm ci
+
+.PHONY: build
+build: node_modules
+	@echo ">> building Mantine UI"
+	npm run build
+
+.PHONY: test
+test: node_modules
+	npm run vitest
+
+.PHONY: lint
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -29,26 +30,43 @@ STATICCHECK_IGNORE =
...
-assets: $(FRONTEND_DIR)/src/Data ui-elm template/email.tmpl
+assets: $(FRONTEND_DIR)/src/Data ui-elm ui-mantine template/email.tmpl
 
 .PHONY: assets-tarball
 assets-tarball: ui-elm
 	mkdir -p .tarballs
 	tar czf ".tarballs/alertmanager-web-ui-$(file <VERSION).tar.gz" -C ui/app dist
 
+.PHONY: mantine-assets-tarball
+mantine-assets-tarball: ui-mantine
```
SoloJacobs: I think we need a way to include mantime into the existing tarball. Anything else is tedious for users. Obviously, only need once we actually include mantime into the build.

ui/mantine-ui/package.json
```diff
@@ -15,7 +15,7 @@
     "check:write": "biome check --write .",
     "vitest": "vitest run",
     "vitest:watch": "vitest",
-    "test": "npm run typecheck && npm run check && npm run vitest && npm run build"
```
SoloJacobs, line 18: This seems like a bug.

ui/mantine-ui/Makefile
```diff
@@ -0,0 +1,30 @@
...
+UI_CONFIG_FILES = Makefile biome.json index.html package.json package-lock.json postcss.config.cjs tsconfig.json vite.config.mjs vitest.setup.mjs
+
+node_modules: package.json package-lock.json
+	@echo ">> installing Mantine UI dependencies"
+	npm ci
+
+.PHONY: build
+build: dist/.build_stamp
+
+dist/.build_stamp: $(UI_DIRS) $(UI_FILES) $(UI_CONFIG_FILES)
```
SoloJacobs: Just FYI: Working with this target isn't fun. The time-based invalidation is wrong in a alot of cases.

Makefile
siavashs: I already checked and changing the tarball contents will break the packaging for Fedora and others.
So it will be a breaking change for linux distros unfortunately.

ui/mantine-ui/package.json
siavashs: tsconfig.json already sets `"noEmit": true`, so both `tsc --noEmit` and the `tsc` inside `npm run build` perform the same type-checking pass. Previously,`npm test` ran `TypeScript` twice.

ui/mantine-ui/Makefile
siavashs: I think the old UI does the same, what is the better alternative?

ui/mantine-ui/package.json
SoloJacobs: Aha :+1:

ui/mantine-ui/Makefile
SoloJacobs: Yes, I know. It was the best I could come with `make`. The reason it here, is that we desperately need caching `elm`. That's because `elm` has it's packaging system on top of the stuff pulled by `npm`. So skipping the elm build was necessary. But with a normal project, we might just be able to use `tsc` built-in capabilities. More correct results, for possibly marginally more build time.

Another option would to use something like `earthly` or `task-go`. But that's certainly out of scope here.

Makefile
SoloJacobs: That seems very surprising to me?
They will all use something like:
```
tar -xf %{S:20} -C ui/app/
```
This means our tar can just unpack
dist/elm and dist/mantime and `go:embed` directives will be `app/dist/elm` and `app/dist/mantime`.

Even if this approach is not correct: I don't want to release two tarballs. That will be a breaking change anyway.

ui/mantine-ui/Makefile
siavashs: I see, I will drop this stamp then for the new UI.

Makefile
siavashs: So the `ui/app` is used by the old UI and the `dist` in its tarball has files in the root, so if we don't want to change tarball extraction path we would have to make mantine put its dist files in the old UI path, etc.

The current change builds the tarball but does not release it yet, my intention was to swap them out at some point. But I will check if we can merge them inside `ui/app` for distribution, etc.

SoloJacobs: Yes, for this pr specifically, I don't think this needs to be solved now. My worry is more nuanced then that: The tar ball is only useful, if we ever want to publish two release artifacts on the release page. No matter which path is correct, (whether we swap the tars at some point, or whether we announce a breaking change), this target is not needed.

In sum, I think we should just delete this make target. We can worry about how to integrate it into go later, that's not the subject of this PR.

Currently, I think we can have both:

* Ship old and new UI in the same binary.
* Don't cause a breaking change for Fedora, etc.

I am still open to being wrong about that. You can delegate this to me, since it's my idea (if you like). My feeling is just that tarball addressing neither strategy coherently.

SoloJacobs (approved):

ui/mantine-ui/Makefile
```diff
@@ -0,0 +1,20 @@
...
+	npm ci
+
+.PHONY: build
+build: node_modules
+	@echo ">> building Mantine UI"
+	npm run build
+
+.PHONY: test
+test: node_modules
+	npm run vitest
```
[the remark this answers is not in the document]
siavashs: Not valid, we already run other checks in other make targets and ci.

PR #5499, opened 2026-08-31
telegram: exclude link targets from message length

alexalok:
## Summary

- Count decoded visible HTML text toward Telegram's 4096-character limit.
- Exclude markup and attribute values such as embedded `href` URLs.
- Preserve fallback for genuinely over-limit visible messages.
- Add regression coverage for long hidden link targets.

## Testing

- `go test ./notify/... -count=1`
- `golangci-lint run ./notify/telegram/...`

#### Pull Request Checklist

- Open issues discussed with maintainers: none.
- Is this a bugfix?

```release-notes
[BUGFIX] Telegram: Count only visible HTML text toward the message limit, excluding embedded link URLs.
```

### DIFF
--- a/notify/telegram/telegram.go
+++ b/notify/telegram/telegram.go
@@ -162,6 +162,8 @@ func htmlTextRuneCount(message string) int {
 				return utf8.RuneCountInString(message)
 			}
 			return count
+		case html.CommentToken, html.DoctypeToken:
+			return utf8.RuneCountInString(message)
 		case html.TextToken:
 			count += utf8.RuneCount(tokenizer.Text())
 		}
--- a/notify/telegram/telegram.go
+++ b/notify/telegram/telegram.go
@@ -158,6 +158,9 @@ func htmlTextRuneCount(message string) int {
 	for {
 		switch tokenizer.Next() {
 		case html.ErrorToken:
+			if len(tokenizer.Raw()) > 0 {
+				return utf8.RuneCountInString(message)
+			}
 			return count
 		case html.TextToken:
 			count += utf8.RuneCount(tokenizer.Text())
--- a/notify/telegram/telegram.go
+++ b/notify/telegram/telegram.go
@@ -22,8 +22,10 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"unicode/utf8"
 
 	commoncfg "github.com/prometheus/common/config"
+	"golang.org/x/net/html"
 	"gopkg.in/telebot.v3"
 
 	"github.com/prometheus/alertmanager/notify"
@@ -31,7 +33,8 @@ import (
 	"github.com/prometheus/alertmanager/types"
 )
 
-// Telegram supports 4096 chars max - from https://limits.tginfo.me/en.
+// Telegram supports up to 4096 characters after entity parsing.
+// See https://core.telegram.org/bots/api#sendmessage.
 const maxMessageLenRunes = 4096
 
 // Notifier implements a Notifier for telegram notifications.
@@ -92,7 +95,7 @@ func (n *Notifier) Notify(ctx context.Context, alert ...*types.Alert) (bool, err
 		if err != nil {
 			return false, err
 		}
-		if len([]rune(messageText)) > maxMessageLenRunes {
+		if htmlTextRuneCount(messageText) > maxMessageLenRunes {
 			messageText = `Alertmanager notification could not be sent: message length exceeds Telegram limits.
 			Please check the template used for producing the message content.`
 		}
@@ -145,6 +148,21 @@ func wrapWithFailureReason(err error) error {
 		return notify.NewErrorWithReason(notify.GetFailureReasonFromStatusCode(apiErr.Code), err)
 	}
 	return err
+}
+
+// htmlTextRuneCount excludes HTML markup and attribute values such as link targets.
+func htmlTextRuneCount(message string) int {
+	tokenizer := html.NewTokenizer(strings.NewReader(message))
+	count := 0
+
+	for {
+		switch tokenizer.Next() {
+		case html.ErrorToken:
+			return count
+		case html.TextToken:
+			count += utf8.RuneCount(tokenizer.Text())
+		}
+	}
 }
 
 func createTelegramClient(apiURL, parseMode string, httpClient *http.Client) (*telebot.Bot, error) {
--- a/notify/telegram/telegram.go
+++ b/notify/telegram/telegram.go
@@ -169,9 +169,73 @@ func htmlTextRuneCount(message string) int {
 		case html.CommentToken, html.DoctypeToken:
 			return utf8.RuneCountInString(message)
 		case html.TextToken:
-			count += utf8.RuneCount(tokenizer.Text())
+			count += telegramHTMLTextRuneCount(tokenizer.Raw())
 		}
 	}
+}
+
+func telegramHTMLTextRuneCount(text []byte) int {
+	count := 0
+	for len(text) > 0 {
+		if text[0] == '&' {
+			if entityLen := telegramHTMLEntityLen(text); entityLen > 0 {
+				count++
+				text = text[entityLen:]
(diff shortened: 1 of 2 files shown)


notify/telegram/telegram.go
```diff
@@ -147,6 +150,21 @@ func wrapWithFailureReason(err error) error {
...
 
+// htmlTextRuneCount excludes HTML markup and attribute values such as link targets.
+func htmlTextRuneCount(message string) int {
+	tokenizer := html.NewTokenizer(strings.NewReader(message))
+	count := 0
+
+	for {
+		switch tokenizer.Next() {
+		case html.ErrorToken:
+			return count
```
SoloJacobs, line 168: This seems like the wrong with to return here?

notify/telegram/telegram.go
```diff
@@ -147,6 +150,21 @@ func wrapWithFailureReason(err error) error {
 	return err
 }
 
+// htmlTextRuneCount excludes HTML markup and attribute values such as link targets.
```
SoloJacobs, line 157: Can you find the corresponding source code? I don't feel comfortable just trusting the documentation.

notify/telegram/telegram_test.go
```diff
@@ -133,6 +134,16 @@ func TestTelegramNotify(t *testing.T) {
 			expText: `Alertmanager notification could not be sent: message length exceeds Telegram limits.
 			Please check the template used for producing the message content.`,
 		},
+		{
```
SoloJacobs, line 151: Let's add a test case for broken html too?

SoloJacobs (changes requested):

notify/telegram/telegram.go
alexalok: `x/net/html` uses `ErrorToken` for normal EOF, so returning the accumulated count remains correct when `Raw()` is empty. An incomplete token can also return `ErrorToken` with unconsumed raw bytes, which made this path return a partial count. Commit fc7887d4 detects that case and conservatively returns the raw rune count instead.

notify/telegram/telegram.go
alexalok: Verified against source. The Bot API calls `parseTextEntities` in [Client.cpp](https://github.com/tdlib/telegram-bot-api/blob/e3e9dd8e5b3d7ab8537cd5a10dc31d5ffa8f82d1/telegram-bot-api/Client.cpp#L11955-L11981). TDLib HTML parsing stores `href` separately as entity data and removes markup from resulting text in [MessageEntity.cpp](https://github.com/tdlib/td/blob/d1085f9cebc5a62379991ae1652673954f229c1f/td/telegram/MessageEntity.cpp#L3437-L3570). The 4096 limit is then checked against parsed text in [MessageContent.cpp](https://github.com/tdlib/td/blob/d1085f9cebc5a62379991ae1652673954f229c1f/td/telegram/MessageContent.cpp#L4809-L4811). Thus the embedded link target is excluded from that limit. There is also a separate 32768-byte raw-input guard before parsing; this regression case remains below it.

notify/telegram/telegram_test.go
alexalok: Added in fc7887d4. The regression renders a malformed anchor at the beginning through `safeHtml`, followed by more than 4096 raw runes. Before the fix, tokenization returned a partial count and the message was sent unchanged; it now takes the conservative fallback path.

notify/telegram/telegram.go
```diff
@@ -147,6 +150,24 @@ func wrapWithFailureReason(err error) error {
...
+func htmlTextRuneCount(message string) int {
+	tokenizer := html.NewTokenizer(strings.NewReader(message))
+	count := 0
+
+	for {
+		switch tokenizer.Next() {
+		case html.ErrorToken:
+			if len(tokenizer.Raw()) > 0 {
+				return utf8.RuneCountInString(message)
```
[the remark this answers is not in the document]
alexalok: Confirmed. `x/net/html` emits unterminated comments as `CommentToken` and unterminated doctypes as `DoctypeToken`; after those tokens consume the payload, the following `ErrorToken` has empty `Raw()`, so the previous guard missed it. Commit 996ef741 treats both token types conservatively by counting the whole raw message and adds regressions for both cases.

SoloJacobs:
Thanks, I think this approach is sound, even described by [Telegram](https://core.telegram.org/api/entities#computing-entity-length) .
 
I'm just worried about the maintenance overhead of this: The API contract here is extremely weak. And the previous implementation is much easier to maintain/understand. I will talk to the other maintainers to find out what they think.

alexalok:
Sure, no problem. Happy to help!

notify/telegram/telegram.go
SoloJacobs: Let's add the input guard? No reason to send messages, which we know will be rejected.

SoloJacobs (changes requested):
Hi @alexalok ,

We talked about this in the maintainer group, and we are open to this approach. I would be happy to see it land. I have computed the following examples by querying the telegram api:
```
1 '&lt;'
1 '&gt;'
1 '&amp;'
1 '&quot;'
1 '&#39;'
1 '&#x27;'
1 '&#128293;'
1 '&#x1F525;'
6 '&nbsp;'
7 '&mdash;'
6 '&apos;'
6 '&copy;'
8 '&hellip;'
6 '&nbspa'
6 '&nbsp '
5 '&ampa'
2 '&amp '
4 '&#0;'
10 '&#1114112;'
5 'a & b'
```
Can you add these as test cases? The count is the length as per telegram.

alexalok:
Hi @SoloJacobs,

Glad to hear the good news! Sure, I will land the requested changes shortly.

notify/telegram/telegram.go
alexalok: Added in `39971770`. HTML-mode notifications now apply Telegram’s 32,768-byte pre-parse input guard before visible-text counting. Tests cover exactly 32,768 bytes being accepted and 32,769 bytes selecting the existing fallback.

alexalok:
Added the requested entity-count cases in `716d1224`. This implements strict Telegram-compatible counting as requested: raw text tokens use Telegram’s entity rules instead of Go’s broader HTML5 decoder. `lt`, `gt`, `amp`, and `quot`, plus valid decimal and hexadecimal numeric entities, count as one rune; unsupported named entities and invalid numeric entities remain literal. All 20 supplied cases are covered. Follow-up `12e44d82` documents why raw text is required and both helper contracts.

If preferred, this can be relaxed back to Go’s built-in HTML decoding. Tradeoff: false negatives in limit checks—Go may decode an entity that Telegram leaves literal, causing us to undercount text Telegram receives.

SoloJacobs (changes requested):
Hey,

looks good :+1: Just one last request on my end: it looks to me like you ported some of the functionality from tdlib? That is the only explanation I have for code like this at least:
```
	if err != nil || code == 0 || code >= utf8.MaxRune || end >= 10 {
```
Could you confirm my theory? And possibly add a reference to this code? Then other people can understand some of the oddities in that function.

Kind regards

alexalok:
Yes, your theory is correct. `telegramHTMLEntityLen` intentionally mirrors TDLib’s `decode_html_entity` behavior so strict Telegram-compatible counting matches what Telegram actually decodes. This avoids Go’s broader HTML5 parser decoding entities that Telegram leaves literal and consequently undercounting visible text.

The `code == 0`, `code >= utf8.MaxRune`, and `end >= 10` checks map directly to TDLib’s corresponding validation. I've added a stable reference to [that TDLib implementation](https://github.com/tdlib/td/blob/d1085f9cebc5a62379991ae1652673954f229c1f/td/telegram/MessageEntity.cpp#L3240-L3286) next to the helper in `02bbd394`.

SoloJacobs (approved):

SoloJacobs:
Thanks !

PR #5500, opened 2026-08-31
fix(events): include alert details in all notify events

siavashs:
Notification events did not include alert details like fingerprint in all cases. This made it defficult to track an alert end to end when querying events produced by event recorder.

This change:
- adds firing, muted, resolved alerts to notify context
- emits the alerts in all notify events' details
-  uses fingerprint instead of nflog's xxhash

This is follow up for #5409


proto/eventrecorder/events/v2/events.proto
```diff
@@ -74,13 +68,13 @@ message AlertCreatedEvent {
 
 // AlertResolvedEvent is emitted when a resolved alert leaves its group.
 message AlertResolvedEvent {
-  GroupedAlert alert = 1;
+  Alert alert = 1;
```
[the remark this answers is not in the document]
siavashs: It is OK, we have not release v2 proto yet.

proto/eventrecorder/events/v2/events.proto
```diff
@@ -51,12 +51,6 @@ message Alert {
   bool resolved = 7;
 }
 
-// GroupedAlert is a reference to an alert within an aggregation group.
-message GroupedAlert {
-  uint64 hash = 1;
-  optional Alert details = 2;
-}
-
```
Spaceman1701, line 59: I'm not sure this is the right thing to do - I think it's useful to have the alert hash recorded here, regardless of whether or not it's needed to join together the alerts in the notification message.

proto/eventrecorder/events/v2/events.proto
```diff
@@ -102,10 +96,10 @@ message Integration {
 
 // NotificationEvent is emitted after a notification is delivered.
 message NotificationEvent {
-  repeated GroupedAlert alerts = 1;
-  repeated GroupedAlert firing_alerts = 2;
-  repeated GroupedAlert resolved_alerts = 3;
-  repeated GroupedAlert muted_alerts = 4;
+  repeated Alert alerts = 1;
+  repeated Alert firing_alerts = 2;
+  repeated Alert resolved_alerts = 3;
```
Spaceman1701, line 101: this causes a lot of duplication in the message, right? What if we went the other direction and ensured `alerts` contains _all_ alerts, and then just provided the hash for `firing/resolved/muted`? Or we could even make firing/resolved/muted a property of the alert... 

I guess I think there's some modeling defect here.

proto/eventrecorder/events/v2/events.proto
siavashs: The hash is internal to Alertmanager, we never expose it in APIs or otherwise so it is not useful for joining an alert notification to an actual received or grouped alert.

proto/eventrecorder/events/v2/events.proto
siavashs: Good catch, I think this tries to follow what is happening inside the notification pipeline but can result in duplicate data.
I think your suggestion is good, but we need to decide which hash we want to keep here, the nflog hash or the fingerprint.

proto/eventrecorder/events/v2/events.proto
siavashs: I restored `GroupedAlert` and switched it to use `fingerprint` instead of `hash`.

proto/eventrecorder/events/v2/events.proto
siavashs: Restored as suggested, `firing`, `resolved` and `muted` variants only include `fingerprint`s now.

Spaceman1701 (approved):

PR #5502, opened 2026-08-31
feat(ui): serve the Mantine application

siavashs:
## Summary

- build the Elm and Mantine distributions as sibling trees under `ui/app/dist/{elm,mantine}`
- embed both distributions in the Alertmanager binary and include both in the existing UI archive
- serve the Mantine application at `/ui/` with its assets isolated under `/ui/assets/`
- use hash-based client routing so root and route-prefixed deployments need no SPA fallback or runtime path injection
- preserve the existing explicit Go compression contract and add only the `.txt` type required by the Mantine license artifact
- leave WOFF/WOFF2 raw while requiring both gzip and Brotli for compressible formats
- verify the real embedded outputs rather than synthetic asset fixtures
- reject invalid asset paths before joining them into either embedded filesystem

The local workflow is:

```bash
make ui-elm ui-mantine
go run ./cmd/alertmanager --config.file=doc/examples/simple.yml
```

Clean BrowserRouter paths and runtime public-prefix handling will follow separately.

Part of #5486

### DIFF
--- a/ui/web.go
+++ b/ui/web.go
@@ -42,6 +42,7 @@ var fileTypes = map[string]struct {
 	".js":    {"text/javascript; charset=utf-8", true},
 	".svg":   {"image/svg+xml", true},
 	".ttf":   {"font/ttf", true},
+	".txt":   {"text/plain; charset=utf-8", true},
 	".woff":  {"font/woff", false},
 	".woff2": {"font/woff2", false},
 }
--- a/ui/web.go
+++ b/ui/web.go
@@ -31,6 +31,24 @@ import (
 //go:embed app/dist
 var asset embed.FS
 
+const (
+	elmRoot     = "app/dist/elm"
+	mantineRoot = "app/dist/mantine"
+)
+
+var (
+	elmFS     = mustSub(elmRoot)
+	mantineFS = mustSub(mantineRoot)
+)
+
+func mustSub(root string) fs.FS {
+	sub, err := fs.Sub(asset, root)
+	if err != nil {
+		panic(err)
+	}
+	return sub
+}
+
 var fileTypes = map[string]struct {
 	contentType  string // https://www.iana.org/assignments/media-types/
 	varyEncoding bool   // Must match build configuration in vite.config.mjs.
@@ -117,11 +135,7 @@ func decompressToReader(f fs.File) (*bytes.Reader, error) {
 
 // Register registers handlers to serve files for the web interface.
 func Register(r *route.Router) {
-	appFS, err := fs.Sub(asset, "app/dist")
-	if err != nil {
-		panic(err) // During build step, we did not embed a directory named `app/dist`.
-	}
-	serve := func(w http.ResponseWriter, req *http.Request, filePath string, immutable bool) {
+	serve := func(w http.ResponseWriter, req *http.Request, appFS fs.FS, filePath string, immutable bool) {
 		ext := strings.ToLower(path.Ext(filePath))
 		fileType, ok := fileTypes[ext]
 		if !ok {
@@ -178,16 +192,37 @@ func Register(r *route.Router) {
 		http.NotFound(w, req)
 	}
 
+	serveAssets := func(w http.ResponseWriter, req *http.Request, appFS fs.FS) {
+		filePath := strings.TrimPrefix(route.Param(req.Context(), "path"), "/")
+		if filePath == "" || !fs.ValidPath(filePath) {
+			http.NotFound(w, req)
+			return
+		}
+		serve(w, req, appFS, path.Join("assets", filePath), true)
+	}
+
 	r.Get("/", func(w http.ResponseWriter, req *http.Request) {
-		serve(w, req, "index.html", false)
+		serve(w, req, elmFS, "index.html", false)
 	})
 
 	r.Get("/favicon.ico", func(w http.ResponseWriter, req *http.Request) {
-		serve(w, req, "favicon.ico", false)
+		serve(w, req, elmFS, "favicon.ico", false)
 	})
 
 	r.Get("/assets/*path", func(w http.ResponseWriter, req *http.Request) {
-		serve(w, req, path.Join("assets", route.Param(req.Context(), "path")), true)
+		serveAssets(w, req, elmFS)
+	})
+
+	r.Get("/ui", func(w http.ResponseWriter, req *http.Request) {
+		http.Redirect(w, req, req.URL.Path+"/", http.StatusFound)
+	})
+
+	r.Get("/ui/", func(w http.ResponseWriter, req *http.Request) {
+		serve(w, req, mantineFS, "index.html", false)
+	})
+
+	r.Get("/ui/assets/*path", func(w http.ResponseWriter, req *http.Request) {
+		serveAssets(w, req, mantineFS)
 	})
 }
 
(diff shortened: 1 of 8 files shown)


ui/web.go
```diff
@@ -35,13 +35,23 @@ var fileTypes = map[string]struct {
...
 	".eot":   {"application/vnd.ms-fontobject", true},
+	".gif":   {"image/gif", false},
 	".html":  {"text/html; charset=utf-8", true},
 	".ico":   {"image/vnd.microsoft.icon", true},
+	".jpeg":  {"image/jpeg", false},
+	".jpg":   {"image/jpeg", false},
 	".js":    {"text/javascript; charset=utf-8", true},
+	".json":  {"application/json; charset=utf-8", true},
+	".mjs":   {"text/javascript; charset=utf-8", true},
+	".otf":   {"font/otf", true},
```
SoloJacobs: Do we really want to serve `otf` and `webp` files? I would consider those not suitable for a modern webapp.

It also seems like the only way with vite to serve `.mjs` is by accident?

We could consider removing the `jpg` entry. We just have to use consistent file endings.

ui/web.go
```diff
@@ -35,13 +35,23 @@ var fileTypes = map[string]struct {
...
 	".css":   {"text/css; charset=utf-8", true},
 	".eot":   {"application/vnd.ms-fontobject", true},
+	".gif":   {"image/gif", false},
 	".html":  {"text/html; charset=utf-8", true},
 	".ico":   {"image/vnd.microsoft.icon", true},
+	".jpeg":  {"image/jpeg", false},
+	".jpg":   {"image/jpeg", false},
 	".js":    {"text/javascript; charset=utf-8", true},
+	".json":  {"application/json; charset=utf-8", true},
+	".mjs":   {"text/javascript; charset=utf-8", true},
```
SoloJacobs: The charset on JSON seems of to me? I don't think that's in an rfc.

ui/web_test.go
```diff
@@ -90,6 +91,37 @@ func TestIndexAssetsAreServedPrefix(t *testing.T) {
...
+func TestMantineBuildOutput(t *testing.T) {
+	mantineFS := os.DirFS("mantine-ui/dist")
+	if _, err := fs.Stat(mantineFS, "."); os.IsNotExist(err) {
+		t.Skip("Mantine UI assets are not built")
+	} else {
+		require.NoError(t, err)
+	}
+
+	count := make(map[string]int)
+	err := fs.WalkDir(mantineFS, ".", func(filePath string, d fs.DirEntry, err error) error {
```
SoloJacobs: This looks like a reimplementation of `walkEmbeddedFiles`?

ui/web.go
```diff
@@ -124,13 +126,29 @@ func decompressToReader(f fs.File) (*bytes.Reader, error) {
...
+
+// Options configures optional web interface assets.
+type Options struct {
+	// MantineAssetsPath is the filesystem path to the built Mantine UI.
+	MantineAssetsPath string
+}
+
 // Register registers handlers to serve files for the web interface.
-func Register(r *route.Router) {
+func Register(r *route.Router, opts ...Options) {
```
SoloJacobs: Is `Options struct` a plan for a new feature? I would rather not introduce this in this commit.

ui/web.go
```diff
@@ -188,15 +206,50 @@ func Register(r *route.Router) {
...
 	})
 
 	r.Get("/favicon.ico", func(w http.ResponseWriter, req *http.Request) {
-		serve(w, req, "favicon.ico", false)
+		serve(appFS, w, req, "favicon.ico", false)
 	})
 
 	r.Get("/assets/*path", func(w http.ResponseWriter, req *http.Request) {
-		serve(w, req, path.Join("assets", route.Param(req.Context(), "path")), true)
+		filePath := strings.TrimPrefix(route.Param(req.Context(), "path"), "/")
```
SoloJacobs: I think the other file layout I suggested would simplify this code a lot.

ui/web_test.go
```diff
@@ -27,6 +30,31 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func writeCompressedAsset(t *testing.T, filePath string, data []byte) {
```
SoloJacobs: This defeats the purpose of the test, we should be using the mantime UI directly.

ui/web_test.go
```diff
@@ -27,6 +30,31 @@ import (
...
+	require.NoError(t, writer.Close())
+	require.NoError(t, os.WriteFile(filePath+".gz", compressed.Bytes(), 0o644))
+	require.NoError(t, os.WriteFile(filePath+".br", []byte("brotli"), 0o644))
+}
+
+func mantineAssetsPath(t *testing.T) string {
+	t.Helper()
+	dir := t.TempDir()
+	require.NoError(t, os.Mkdir(filepath.Join(dir, "assets"), 0o755))
+	writeCompressedAsset(t, filepath.Join(dir, "index.html"), []byte(`<!DOCTYPE html><html><head></head><body><script type="module" src="./assets/index.js"></script></body></html>`))
```
SoloJacobs: Again, let's not test made files.

ui/web.go
```diff
@@ -124,13 +126,29 @@ func decompressToReader(f fs.File) (*bytes.Reader, error) {
 	return bytes.NewReader(data), nil
 }
 
+func fileExists(assetFS fs.FS, filePath string) bool {
```
SoloJacobs: I don't think this check buys anthing.

SoloJacobs: We should probably just ship the UI in the next release? I don't really see why we shouldn't.

ui/web.go
```diff
@@ -188,15 +206,50 @@ func Register(r *route.Router) {
...
+	if err != nil {
+		return
+	}
+	mantineFS := os.DirFS(mantineAssetsPath)
+	if !fileExists(mantineFS, "index.html.gz") || !fileExists(mantineFS, "index.html.br") {
+		return
+	}
+
+	r.Get("/ui", func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Location", "ui/")
```
SoloJacobs: Wouldn't this be solved better with a redirect?

SoloJacobs (changes requested):
I had a look at the first two commits, but not the third. I think we should move this Pr. This way we can already land some things sooner. And they are separate issues anyway.

ui/web.go
siavashs: RFC 8259 requires interoperable JSON to use UTF-8, but `application/json` does not define a charset parameter.
https://datatracker.ietf.org/doc/html/rfc8259#section-8.1

I will drop the charset for this one.

ui/web.go
siavashs: I think the agent was too greedy here, I will remove anything we currently do not serve.

ui/web_test.go
siavashs: True, I will reuse that.

ui/web.go
siavashs: This was added only to package mantine optionally, if we add it by default per your other comment then we can drop it.

ui/web.go
siavashs: I was just being conservative here, but it is fine with me if we ship it.
It is broken in its current form but it is OK.

SoloJacobs (approved):
LGTM, thanks 👍

PR #5503, opened 2026-08-31
notify: Muted alerts wired in

suprjinx:
Adds feature flag `EnableMutedAlertsInNflog` and keeps muted alerts in the nflog Entry for all muting stages. 

#### Pull Request Checklist

  - Relates https://github.com/prometheus/alertmanager/issues/226
  - Relates https://github.com/prometheus/alertmanager/issues/5247
    
- Is this a bugfix?
- Is this a new feature?
- Is this a breaking change?
- [ x] I have signed-off my commits
- [ x] I will follow [best practices for contributing to this project](https://docs.github.com/en/get-started/exploring-projects-on-github/contributing-to-open-source)

Adds a configuration switch for "muted-alerts-in-nflog". When set, muted alerts are retained in the MutedAlerts list of the log entry.

```release-notes

```

Assisted-by: Claude Code


notify/set_notifies_stage.go
```diff
@@ -29,16 +31,40 @@ import (
...
+	if !ok || len(muted) == 0 {
+		return nil
+	}
+
+	hashes := make([]uint64, 0, len(muted))
+	for hash := range muted {
+		hashes = append(hashes, hash)
+	}
+	slices.Sort(hashes)
+
```
Spaceman1701, line 64: is there already a function that does this for resolved/firing alerts? If so, we should try to share the implementation.

notify/notify.go
```diff
@@ -243,15 +253,38 @@ func (rs RoutingStage) Exec(ctx context.Context, l *slog.Logger, alerts ...*aler
...
 // Exec implements the Stage interface.
 func (ms MultiStage) Exec(ctx context.Context, l *slog.Logger, alerts ...*alert.Alert) (context.Context, []*alert.Alert, error) {
+	return execStages(ctx, l, ms, false, alerts...)
+}
+
+// A MutedMultiStage executes a series of stages sequentially, and keeps going
+// when a mute stage has removed every alert from the pipeline. A group whose
+// alerts are all muted is not the same as a group with nothing in it, and the
+// stages that record the group's state have to run either way.
+type MutedMultiStage []Stage
```
Spaceman1701, line 269: I'm not against this approach, but at face value it seems like it might be a bit overkill. Both `MultiStage` and `MutedMultiStage` end up calling the same `execStages` function, so why not move the feature flag check into there? Do you have something more planned for `MutedMultiStage`?

Spaceman1701 (commented):
Thanks! This is looking very good, just a few small comments.

notify/notify.go
suprjinx: You're right, it's not doing much -- but we don't have an easy way to inject the feature flag into MultiStage -- it would need a new attribute to carry the setting down to the `execStages` call site, so it would no longer be a simple type alias for `[]Stage`.

notify/set_notifies_stage.go
suprjinx: The similar cases in the codebase have slight differences (eg, `partitionAlertsByState` in dedup_stage.go doesn't sort). But we can simplify to one-liner by using `slices.Sorted(maps.Keys(muted))`

notify/notify.go
Spaceman1701: Ah, right, you'd need to at least have the `Flagger`. Sure, I think that's a good enough reason.

Spaceman1701 (approved):

Issue #5478, opened 2026-08-29
API v3alpha: harden the ConnectRPC foundation


siavashs:
Parent: #5450

## Summary

Harden the ConnectRPC foundation delivered in #5377 before expanding it with write-heavy and long-lived service methods.

The existing foundation already supports Connect, gRPC-Web, native gRPC, health, reflection, route prefixes, independent unary/stream admission, and bounded instrumentation paths. This issue covers the remaining production-readiness and extensibility gaps.

### Admission observability

Add bounded metrics for:

- Unary RPCs in flight.
- Unary admission rejections.
- Unary durations and deadline expirations.
- Active and rejected streams.
- Stream lifetimes.
- Queue-overflow closures once watch queues exist.

Pass a Prometheus registerer and explicit policy values through the API options. Connect metrics must use fixed service/procedure and outcome labels rather than request-controlled paths.

### Request and stream controls

- Configure explicit incoming and outgoing protobuf message limits; Connect defaults both directions to unlimited.
- Add request-body/read controls that apply before unary protobuf decoding. The current unary interceptor deadline begins after Connect receives the request message.
- Define distinct per-message, idle, lifetime, cancellation, and queue policies for streams.
- Do not wrap streams with `http.TimeoutHandler` or apply a global unary-style timeout to long-lived streams.
- Preserve independent capacity for API v2 GETs, Connect unary RPCs, and Connect streams.

### Service registration

Replace the separately maintained handler registrations, health/reflection service names, and instrumentation prefixes with one explicit canonical service descriptor. Keep the instrumentation allowlist fixed, bounded, and pinned by tests.

### TLS verification

Extend the in-process e2e harness to exercise exporter-toolkit TLS configuration and verify:

- Connect over HTTPS.
- Connect HTTP GET over HTTPS.
- gRPC-Web over HTTPS.
- Native gRPC with TLS and HTTP/2 ALPN.
- Health and reflection.
- Root and non-root route prefixes.

Retain the existing plaintext h2c matrix and document the operational exposure model for h2c.

## Acceptance criteria

- [ ] Admission and stream lifecycle metrics are registered with bounded labels.
- [ ] Oversized incoming messages fail predictably with `ResourceExhausted`.
- [ ] Slow unary uploads are bounded before handler execution.
- [ ] Stream cancellation, idle, and lifetime controls release all capacity.
- [ ] API v2 behavior and capacity remain independent and unchanged.
[...]

PR #5506, opened 2026-09-01
api: centralize Connect service metadata

siavashs:
Keep handler registration, health and reflection names, and bounded procedure instrumentation derived from one catalog, and verify every advertised procedure resolves through the registered API mux.

Part of #5478

### DIFF
--- a/api/api_test.go
+++ b/api/api_test.go
@@ -16,12 +16,18 @@ package api
 import (
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"sync"
 	"testing"
 	"testing/synctest"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/common/promslog"
+	"github.com/prometheus/common/route"
 	"github.com/stretchr/testify/require"
+
+	apiconnect "github.com/prometheus/alertmanager/api/connect"
+	apiv2 "github.com/prometheus/alertmanager/api/v2"
 )
 
 func TestConcurrencyLimitHandler(t *testing.T) {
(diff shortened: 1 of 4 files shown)


api/connect/connect.go
```diff
@@ -43,12 +43,28 @@ type Options struct {
...
+type procedureDescriptor struct {
+	path       string
+	service    string
+	procedure  string
+	streamType connect.StreamType
+}
+
+type serviceDescriptor struct {
+	name       string
+	advertised bool
```
SoloJacobs: Let's remove `advertised` and just rename the variable `applicationServices` to `advertised`. it seems cleaner.

api/connect/connect.go
```diff
@@ -43,12 +43,28 @@ type Options struct {
 	UnaryTimeout      time.Duration
 }
 
+type procedureDescriptor struct {
+	path       string
+	service    string
+	procedure  string
+	streamType connect.StreamType
```
SoloJacobs: Ok to ignore: But we should only add fields in the commit, in which they are acutally needed.

api/api_test.go
```diff
@@ -98,24 +98,24 @@ func TestInstrumentConnectHandlerBoundsCardinality(t *testing.T) {
...
 				http.NotFound(w, r)
 			})
 			h := api.instrumentConnectHandler(
 				mountPrefix,
-				[]string{servicePrefix},
+				[]string{procedure},
 				http.StripPrefix(mountPrefix, inner),
 			)
 
 			for _, procedure := range []string{
```
SoloJacobs: `procedure` is now shadowed, maybe call it `path` instead.

SoloJacobs (approved):
LGTM, some very minor nits. Maybe we should just postpone them to final PR in the stack.

PR #5507, opened 2026-09-01
api: bound Connect admission and message resources

siavashs:
Extend the procedure catalog with service, procedure, and stream metadata, move RPC admission ahead of decoding, and expose configurable resource limits with bounded lifecycle metrics. Preserve successful and specifically coded handler results when request contexts expire.

Part of #5478

```release-notes
[ENHANCEMENT] API: Add configurable ConnectRPC admission, message-size, request-body, and unary timeout controls.
```

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -67,15 +67,33 @@ type Options struct {
 	GroupMutedFunc func(routeID, groupKey string) ([]string, bool)
 	// Peer from the gossip cluster. If nil, no clustering will be used.
 	Peer cluster.ClusterPeer
-	// Timeout for HTTP requests and Connect unary RPCs. The zero value (and
-	// negative values) result in no timeout.
+	// Timeout for HTTP requests and Connect unary RPCs. The zero value and
+	// negative values result in no timeout.
 	Timeout time.Duration
-	// Concurrency limit for GET requests and, independently, Connect unary
-	// RPCs and streams. The zero value (and negative values) result in a
-	// limit of GOMAXPROCS or 8, whichever is larger. Status code 503 is served
-	// for GET requests that would exceed the concurrency limit; Connect calls
-	// receive ResourceExhausted.
+	// Concurrency is the limit for GET requests. The zero value and negative
+	// values result in a limit of GOMAXPROCS or 8, whichever is larger. Status
+	// code 503 is served for requests that would exceed the limit.
 	Concurrency int
+	// ConnectUnaryConcurrency is the independent limit for Connect unary RPCs.
+	// Non-positive values inherit Concurrency. Calls that exceed the limit
+	// receive ResourceExhausted.
+	ConnectUnaryConcurrency int
+	// ConnectStreamConcurrency is the independent limit for Connect streams.
+	// Non-positive values inherit Concurrency. Calls that exceed the limit
+	// receive ResourceExhausted.
+	ConnectStreamConcurrency int
+	// ConnectUnaryTimeout is the timeout for Connect unary RPCs. Zero inherits
+	// Timeout. A negative value disables the Connect unary timeout.
+	ConnectUnaryTimeout time.Duration
+	// ConnectReadMaxBytes limits each incoming Connect protobuf message.
+	// Non-positive values do not set a limit.
+	ConnectReadMaxBytes int
+	// ConnectSendMaxBytes limits each outgoing Connect protobuf message.
+	// Non-positive values do not set a limit.
+	ConnectSendMaxBytes int
+	// ConnectMaxRequestBodyBytes limits the wire size of a Connect unary request
+	// body before decoding. Non-positive values do not set a limit.
+	ConnectMaxRequestBodyBytes int64
 	// Logger is used for logging, if nil, no logging will happen.
 	Logger *slog.Logger
 	// Registry is used to register Prometheus metrics. If nil, no metrics
@@ -87,6 +105,55 @@ type Options struct {
 	// according to the current active configuration. Alerts returned are
 	// filtered by the arguments provided to the function.
 	GroupFunc func(context.Context, func(*dispatch.Route) bool, func(*types.Alert, time.Time) bool) (dispatch.AlertGroups, map[model.Fingerprint][]string, error)
+}
+
+type effectiveOptions struct {
+	logger      *slog.Logger
+	timeout     time.Duration
+	concurrency int
+	connect     apiconnect.Options
+}
+
+func (o Options) resolve() effectiveOptions {
+	logger := o.Logger
+	if logger == nil {
+		logger = promslog.NewNopLogger()
+	}
+	timeout := max(o.Timeout, 0)
+	concurrency := o.Concurrency
+	if concurrency < 1 {
+		concurrency = max(runtime.GOMAXPROCS(0), 8)
+	}
+	unaryConcurrency := o.ConnectUnaryConcurrency
+	if unaryConcurrency < 1 {
+		unaryConcurrency = concurrency
+	}
+	streamConcurrency := o.ConnectStreamConcurrency
+	if streamConcurrency < 1 {
+		streamConcurrency = concurrency
+	}
+	unaryTimeout := o.ConnectUnaryTimeout
+	switch {
+	case unaryTimeout == 0:
+		unaryTimeout = timeout
+	case unaryTimeout < 0:
+		unaryTimeout = 0
+	}
+	return effectiveOptions{
+		logger:      logger,
+		timeout:     timeout,
+		concurrency: concurrency,
+		connect: apiconnect.Options{
+			Peer:                o.Peer,
+			Registerer:          o.Registry,
+			UnaryConcurrency:    unaryConcurrency,
+			StreamConcurrency:   streamConcurrency,
+			UnaryTimeout:        unaryTimeout,
+			ReadMaxBytes:        max(o.ConnectReadMaxBytes, 0),
+			SendMaxBytes:        max(o.ConnectSendMaxBytes, 0),
+			MaxRequestBodyBytes: max(o.ConnectMaxRequestBodyBytes, 0),
+		},
+	}
 }
 
 func (o Options) validate() error {
@@ -111,14 +178,7 @@ func New(opts Options) (*API, error) {
 	if err := opts.validate(); err != nil {
 		return nil, fmt.Errorf("invalid API options: %w", err)
 	}
-	l := opts.Logger
-	if l == nil {
-		l = promslog.NewNopLogger()
-	}
-	concurrency := opts.Concurrency
-	if concurrency < 1 {
-		concurrency = max(runtime.GOMAXPROCS(0), 8)
-	}
+	effective := opts.resolve()
 
 	// The Connect API is always mounted alongside API v2.
 	v2, err := apiv2.NewAPI(
@@ -127,18 +187,13 @@ func New(opts Options) (*API, error) {
 		opts.GroupMutedFunc,
 		opts.Silences,
 		opts.Peer,
-		l.With("version", "v2"),
+		effective.logger.With("version", "v2"),
 		opts.Registry,
 	)
 	if err != nil {
 		return nil, err
 	}
-	connect := apiconnect.NewAPI(apiconnect.Options{
-		Peer:              opts.Peer,
-		UnaryConcurrency:  concurrency,
-		StreamConcurrency: concurrency,
-		UnaryTimeout:      opts.Timeout,
-	})
+	connect := apiconnect.NewAPI(effective.connect)
 
 	requestsInFlight := prometheus.NewGauge(prometheus.GaugeOpts{
 		Name:        "alertmanager_http_requests_in_flight",
@@ -160,14 +215,14 @@ func New(opts Options) (*API, error) {
 	}
 
 	return &API{
-		deprecationRouter:        NewV1DeprecationRouter(l.With("version", "v1")),
+		deprecationRouter:        NewV1DeprecationRouter(effective.logger.With("version", "v1")),
 		v2:                       v2,
 		connect:                  connect,
 		requestDuration:          opts.RequestDuration,
 		requestsInFlight:         requestsInFlight,
 		concurrencyLimitExceeded: concurrencyLimitExceeded,
-		timeout:                  opts.Timeout,
-		inFlightSem:              make(chan struct{}, concurrency),
+		timeout:                  effective.timeout,
+		inFlightSem:              make(chan struct{}, effective.concurrency),
 	}, nil
 }
 
@@ -184,8 +239,9 @@ func (api *API) Register(r *route.Router, routePrefix string) *http.ServeMux {
 	mux := http.NewServeMux()
 	connectHandler := api.connect.Handler()
 	// ConnectRPC procedure paths are the only bounded label values on the
-	// Connect/gRPC surface; any other path yields a 404 and must not be
-	// recorded verbatim, or a client could inflate metric/trace cardinality.
+	// Connect/gRPC surface. The Connect handler rejects every other path. Do not
+	// record unmatched paths verbatim because a client could inflate metric and
+	// trace cardinality.
 	procedures := api.connect.Procedures()
 	// Native gRPC is served at the server root, so match against the raw
 	// request path (no mount prefix).
--- a/api/connect/connect.go
+++ b/api/connect/connect.go
@@ -29,6 +29,7 @@ import (
 	"connectrpc.com/connect"
 	"connectrpc.com/grpchealth"
 	"connectrpc.com/grpcreflect"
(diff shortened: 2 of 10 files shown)


api/api.go
```diff
@@ -133,12 +139,31 @@ func New(opts Options) (*API, error) {
 	if err != nil {
 		return nil, err
 	}
-	connect := apiconnect.NewAPI(apiconnect.Options{
-		Peer:              opts.Peer,
-		UnaryConcurrency:  concurrency,
-		StreamConcurrency: concurrency,
-		UnaryTimeout:      opts.Timeout,
+	unaryConcurrency := opts.ConnectUnaryConcurrency
+	if unaryConcurrency < 1 {
```
SoloJacobs, line 143: This default block is a bit misplaced, concurrency is defined elsewhere. Also: api/connect/connect.go has some very similiar code. I think we should move the Options resolution all in one place, ideally with a [Parse, don't validate ](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/) approach.

api/connect/connect.go
```diff
@@ -53,6 +62,62 @@ type serviceDescriptor struct {
...
+		for _, collector := range []prometheus.Collector{
+			metrics.unaryInFlight,
+			metrics.unaryRejected,
+			metrics.unaryDuration,
+			metrics.unaryDeadlines,
+			metrics.streamsActive,
+			metrics.streamsRejected,
+		} {
+			if err := reg.Register(collector); err != nil {
+				return nil, fmt.Errorf("register Connect API metrics: %w", err)
```
SoloJacobs, line 114: I would call panic here.

api/connect/connect.go
```diff
@@ -147,39 +227,141 @@ func (api *API) serviceDescriptors() []serviceDescriptor {
...
+	return procedureDescriptor{service: "unknown", procedure: "unknown", streamType: streamType}
+}
+
+func (i *admissionInterceptor) enter(desc procedureDescriptor) (func(), error) {
+	labels := prometheus.Labels{"service": desc.service, "procedure": desc.procedure}
+	if desc.streamType == connect.StreamTypeUnary {
 		select {
 		case i.unary <- struct{}{}:
-			defer func() { <-i.unary }()
+			if i.metrics != nil {
```
SoloJacobs, line 259: I would rather drop the `i.metrics != nil` guards. Pass `newRPCMetrics(nil)` in the tests.

api/connect/connect.go
```diff
@@ -147,39 +227,141 @@ func (api *API) serviceDescriptors() []serviceDescriptor {
...
+		}
+		return nil, connect.NewError(connect.CodeResourceExhausted, errors.New("maximum concurrent streams reached"))
+	}
+}
+
+func (i *admissionInterceptor) startTime(ctx context.Context) time.Time {
+	if started, ok := ctx.Value(requestStartContextKey{}).(time.Time); ok {
+		return started
+	}
+	return time.Now()
```
SoloJacobs, line 298: I rather panic, than return `time.Now()`.

(depends on the suggestion I made below)

api/connect/connect.go
```diff
@@ -147,39 +227,141 @@ func (api *API) serviceDescriptors() []serviceDescriptor {
...
+}
+
+func normalizeContextError(ctx context.Context, err error) error {
+	if err == nil {
+		return nil
+	}
+	if connect.CodeOf(err) != connect.CodeUnknown {
+		return err
+	}
+	cause := context.Cause(ctx)
```
SoloJacobs, line 322: Can we remove this early reutnr?

api/connect/connect.go
```diff
@@ -147,39 +227,141 @@ func (api *API) serviceDescriptors() []serviceDescriptor {
...
+	return func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
+		desc := i.descriptor("", connect.StreamTypeUnary)
+		if req != nil {
+			desc = i.descriptor(req.Spec().Procedure, connect.StreamTypeUnary)
+		}
+		started := i.startTime(ctx)
+		if controller, ok := ctx.Value(responseControllerContextKey{}).(*http.ResponseController); ok {
+			_ = controller.SetReadDeadline(time.Time{})
+		}
+		if _, admitted := ctx.Value(admittedContextKey{}).(struct{}); !admitted {
```
SoloJacobs, line 346: `admitted` is false according to claude, if:

>  client posts to a URL under one of the two reflection prefixes but with the wrong method name on the end. The outer control handler looks that URL up in its table, doesn't find it, and steps aside — so it takes no concurrency slot and stamps nothing on the request. The reflection service accepts the request anyway, because unlike the other two services it never checks the URL, only the HTTP method and content type. The interceptor then runs, finds no stamp, and does the admission itself. That's the admitted=false branch.

Can we test this?

(I also suggested to remove this path completely further down).

api/connect/connect.go
```diff
@@ -147,39 +227,141 @@ func (api *API) serviceDescriptors() []serviceDescriptor {
...
+	metrics      *rpcMetrics
 }
 
-func (i *admissionInterceptor) WrapUnary(next connect.UnaryFunc) connect.UnaryFunc {
-	return func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
+func (i *admissionInterceptor) descriptor(path string, streamType connect.StreamType) procedureDescriptor {
+	if procedure, ok := i.procedures[path]; ok {
+		return procedure
+	}
+	return procedureDescriptor{service: "unknown", procedure: "unknown", streamType: streamType}
```
SoloJacobs, line 251: I would like to avoid this unknown procedure, a panic seem preferable. Can we implement taht?

api/connect/connect.go
```diff
@@ -147,39 +227,141 @@ func (api *API) serviceDescriptors() []serviceDescriptor {
...
+		return connect.NewError(connect.CodeCanceled, context.Canceled)
+	default:
+		return err
+	}
+}
+
+func (i *admissionInterceptor) WrapUnary(next connect.UnaryFunc) connect.UnaryFunc {
+	return func(ctx context.Context, req connect.AnyRequest) (connect.AnyResponse, error) {
+		desc := i.descriptor("", connect.StreamTypeUnary)
+		if req != nil {
```
SoloJacobs, line 339: I don't think `connect` will pass a `req` like this.

api/connect/connect.go
```diff
@@ -233,9 +420,71 @@ func (api *API) Procedures() []string {
 	return procedures
 }
 
+func (api *API) controlHandler(next http.Handler, errorWriter *connect.ErrorWriter) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		desc, ok := api.procedures[r.URL.Path]
+		if !ok {
+			next.ServeHTTP(w, r)
```
SoloJacobs, line 427: Maybe we can just  return `connect.CodeUnimplemented` here, and then drop `admittedContextKey{}` altogether?

api/connect/connect.go
```diff
@@ -233,9 +420,71 @@ func (api *API) Procedures() []string {
...
+			_ = r.Body.Close()
+			_ = errorWriter.Write(w, r, err)
+			return
+		}
+		defer release()
+
+		controller := http.NewResponseController(w)
+		started := time.Now()
+		observed := &atomic.Bool{}
+		ctx := context.WithValue(r.Context(), admittedContextKey{}, struct{}{})
```
SoloJacobs, line 441: We should add some helpers here, similiar to what the notifiers do.

api/connect/connect.go
```diff
@@ -233,9 +420,71 @@ func (api *API) Procedures() []string {
...
+		ctx = context.WithValue(ctx, observationContextKey{}, observed)
+		ctx = context.WithValue(ctx, requestStartContextKey{}, started)
+		ctx = context.WithValue(ctx, responseControllerContextKey{}, controller)
+		if desc.streamType == connect.StreamTypeUnary && api.admission.unaryTimeout > 0 {
```
SoloJacobs, line 445: `desc.streamType == connect.StreamTypeUnary` should probably only be checked once for readbility, and the if block below extended.

api/connect/connect.go
```diff
@@ -233,9 +420,71 @@ func (api *API) Procedures() []string {
...
+		}
+		defer release()
+
+		controller := http.NewResponseController(w)
+		started := time.Now()
+		observed := &atomic.Bool{}
+		ctx := context.WithValue(r.Context(), admittedContextKey{}, struct{}{})
+		ctx = context.WithValue(ctx, observationContextKey{}, observed)
+		ctx = context.WithValue(ctx, requestStartContextKey{}, started)
+		ctx = context.WithValue(ctx, responseControllerContextKey{}, controller)
```
SoloJacobs, line 444: Should we really pass all `ctxt` values unconditionally, it seems most are only used by `StreamTypeUnary`.

cmd/alertmanager/main.go
```diff
@@ -60,11 +60,17 @@ func run() int {
...
-		routePrefix    = kingpin.Flag("web.route-prefix", "Prefix for the internal routes of web endpoints. Defaults to path of --web.external-url.").String()
-		getConcurrency = kingpin.Flag("web.get-concurrency", "Maximum number of GET requests processed concurrently. If negative or zero, the limit is GOMAXPROC or 8, whichever is larger.").Default("0").Int()
-		httpTimeout    = kingpin.Flag("web.timeout", "Timeout for HTTP requests. If negative or zero, no timeout is set.").Default("0").Duration()
+		webConfig                  = webflag.AddFlags(kingpin.CommandLine, ":9093")
+		externalURL                = kingpin.Flag("web.external-url", "The URL under which Alertmanager is externally reachable (for example, if Alertmanager is served via a reverse proxy). Used for generating relative and absolute links back to Alertmanager itself. If the URL has a path portion, it will be used to prefix all HTTP endpoints served by Alertmanager. If omitted, relevant URL components will be derived automatically.").String()
```
SoloJacobs, line 64: Should we mark them as experimental somehow? Technically speaking the feature is still subject to change.

api/connect/connect.go
```diff
@@ -53,6 +62,62 @@ type serviceDescriptor struct {
...
+		unaryRejected: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "alertmanager_api_connect_unary_admission_rejections_total",
+			Help: "Total number of Connect unary RPCs rejected by admission control.",
+		}, labels),
+		unaryDuration: prometheus.NewHistogramVec(prometheus.HistogramOpts{
+			Name:    "alertmanager_api_connect_unary_request_duration_seconds",
+			Help:    "Duration of Connect unary RPCs.",
+			Buckets: prometheus.DefBuckets,
+		}, outcomeLabels),
+		unaryDeadlines: prometheus.NewCounterVec(prometheus.CounterOpts{
```
SoloJacobs, line 91: Isn't this redundant? I think `unaryDuration` already has the deadline_exceeded reason.

api/connect/status_test.go
```diff
@@ -228,27 +230,147 @@ var _ = Describe("StatusService", func() {
 		Entry("gRPC-Web", http.MethodPost, []connect.ClientOption{connect.WithGRPCWeb()}),
 		Entry("gRPC", http.MethodPost, []connect.ClientOption{connect.WithGRPC()}),
 	)
+
+	It("rejects oversized incoming messages before handler execution", func() {
```
SoloJacobs, line 234: We should add some tests for stream admission too.

api/connect/connect.go
```diff
@@ -53,6 +62,62 @@ type serviceDescriptor struct {
...
+	unaryDeadlines  *prometheus.CounterVec
+	streamsActive   *prometheus.GaugeVec
+	streamsRejected *prometheus.CounterVec
+}
+
+func newRPCMetrics(reg prometheus.Registerer) (*rpcMetrics, error) {
+	labels := []string{"service", "procedure"}
+	outcomeLabels := []string{"service", "procedure", "outcome"}
+	metrics := &rpcMetrics{
+		unaryInFlight: prometheus.NewGaugeVec(prometheus.GaugeOpts{
```
SoloJacobs, line 78: Can we make the naming a bit more consistent with existing metrics?
Here is an AI suggestion:
```
alertmanager_api_connect_unary_requests_in_flight stays as is
alertmanager_api_connect_stream_requests_in_flight replaces alertmanager_api_connect_streams_active
alertmanager_api_connect_unary_concurrency_limit_exceeded_total replaces alertmanager_api_connect_unary_admission_rejections_total
alertmanager_api_connect_stream_concurrency_limit_exceeded_total replaces alertmanager_api_connect_stream_admission_rejections_total
alertmanager_api_connect_unary_request_duration_seconds stays as is
alertmanager_api_connect_unary_deadline_exceeded_total gets deleted
```
I'm not so attached to this, though.

api/connect/connect.go
```diff
@@ -61,33 +126,43 @@ type API struct {
...
+		readMaxBytes:    opts.ReadMaxBytes,
+		sendMaxBytes:    opts.SendMaxBytes,
+		maxRequestBytes: opts.MaxRequestBodyBytes,
 	}
 	api.services = api.serviceDescriptors()
 	for _, service := range api.services {
 		for _, procedure := range service.procedures {
 			api.procedures[procedure.path] = procedure
 		}
 	}
```
SoloJacobs, line 157: Should intialize the rejection counters to zero here?

SoloJacobs (changes requested):
Pretty nice change, lots of comment by me.

api/connect/connect.go
siavashs: I don't think so, because an expired context can replace a successful handler result with `CodeDeadlineExceeded`.
This fixes an issue discovered earlier by coderabbit.

api/connect/connect.go
siavashs: We use `MustRegister()` now.

api/connect/connect.go
siavashs: We now reject with `CodeUnimplemented` error.

api/connect/connect.go
siavashs: I removed this fallback path instead of preserving it.
`controlHandler` now rejects every unregistered procedure with `CodeUnimplemented`.

api/connect/connect.go
siavashs: I replaced the optional start-time lookup with a required `unaryRequestState`.
`unaryRequestStateFromContext` now panics when the state is missing.

api/connect/connect.go
siavashs: We use `req.Spec()` and `conn.Spec()` directly now.

api/connect/connect.go
siavashs: `controlHandler` now takes an early branch for streams.
It creates `unaryRequestState`, the response controller, and the timeout context only for unary procedures.

api/connect/connect.go
siavashs: I kept this metric because it has different semantics from the duration outcome.
The **duration** histogram records the final normalized handler result.
The result can remain either `ok` or a specific Connect `code` after the context expires.

The **deadline** counter records expiration of the configured server timeout independently.
It now uses a dedicated timeout cause, so parent or client deadlines do not increment it.

api/connect/status_test.go
siavashs: Done!

api/connect/connect.go
siavashs: Applied the suggested naming for the stream in-flight gauge and both concurrency-limit counters. I kept `unary_deadline_exceeded_total` because it measures configured server-timeout expiration independently from the final duration outcome.

api/connect/connect.go
SoloJacobs: I looked at this a bit more, and I think the race doesn't go away, unless we replace the signature with:
```
normalizeContextError(err error) error 
```
And then add this block:
```
				err := error(errors.New("rpc ended before handler execution"))
				switch cause := context.Cause(ctx); {
				case errors.Is(cause, unaryTimeoutError{}):
					err = connect.NewError(connect.CodeDeadlineExceeded, context.DeadlineExceeded)
				case errors.Is(cause, context.Canceled):
					err = connect.NewError(connect.CodeCanceled, context.Canceled)
				}
```
to the call site where it is needed. Checking `ctx` after the function, which was cancelled returned is inherently racy.

SoloJacobs (changes requested):
LGTM, one more thing

api/connect/connect.go
siavashs: Fixed. `normalizeContextError` now depends only on the error returned by the handler, so a concurrent context cancellation cannot reclassify a generic handler error.
Context-based classification now happens only in the outer fallback (where handler execution was not observed).

SoloJacobs (approved):

PR #5538, opened 2026-09-03
feat(ui): migrate mantine to StatusService v3alpha

siavashs:
- replace the Mantine UI's hand-written API v2 status model
- use generated Protobuf-ES bindings for status.v3alpha.StatusService
- use Connect-Query with a route-prefix-aware Connect-Web transport
- render the new start time and cluster state fields
- add coverage for generation, transport routing, response validation, and page rendering

image

image


ui/mantine-ui/buf.gen.yaml
```diff
@@ -0,0 +1,8 @@
+version: v2
+inputs:
+  - directory: ../../proto/api
+plugins:
+  - local: protoc-gen-es
+    out: src/gen
+    opt:
+      - target=ts
```
SoloJacobs, line 8: Should we add `clean: true`?

ui/mantine-ui/package.json
```diff
@@ -18,6 +19,10 @@
     "test": "npm run check && npm run vitest && npm run build"
   },
   "dependencies": {
+    "@bufbuild/protobuf": "^2.14.0",
```
SoloJacobs, line 22: Can we reuse the buf from `go tools`?

ui/mantine-ui/src/data/connect.ts
```diff
@@ -0,0 +1,16 @@
...
+  new URL('../api', baseURI).toString();
+
+export const createAPITransport = (
+  baseURI: string | URL = document.baseURI,
+  fetch?: typeof globalThis.fetch
+) =>
+  createConnectTransport({
+    baseUrl: getAPIBaseURL(baseURI),
+    useHttpGet: true,
+    ...(fetch === undefined ? {} : { fetch }),
```
SoloJacobs, line 13: Just pass `fetch` here, I think?

ui/mantine-ui/src/data/status.ts
```diff
@@ -1,30 +1,57 @@
...
+    case ClusterStatus_State.DISABLED:
+      return 'disabled';
+    case ClusterStatus_State.SETTLING:
+      return 'settling';
+    case ClusterStatus_State.READY:
+      return 'ready';
+    case ClusterStatus_State.UNSPECIFIED:
+      return 'unspecified';
+    default:
+      return 'unknown';
```
SoloJacobs, line 43: I think the better pattern here would be to have an error. The ui is bundled with the srever.

ui/mantine-ui/src/data/status.ts
```diff
@@ -1,30 +1,57 @@
...
+  type GetStatusResponse,
+  StatusService,
+} from '@/gen/status/v3alpha/status_pb';
+
+type Status = AlertmanagerStatus & {
+  versionInfo: NonNullable<AlertmanagerStatus['versionInfo']>;
+  config: NonNullable<AlertmanagerStatus['config']>;
+  cluster: NonNullable<AlertmanagerStatus['cluster']>;
+  startTime: NonNullable<AlertmanagerStatus['startTime']>;
 };
```
SoloJacobs, line 16: We could add a 
```
export const REQUIRED_STATUS_FIELDS = ['versionInfo', 'config', 'cluster', 'startTime'] as const;
```
and reuse in test.

ui/mantine-ui/src/data/status.ts
```diff
@@ -1,30 +1,57 @@
...
+
+export const useStatus = () =>
+  useSuspenseQuery(
+    StatusService.method.getStatus,
+    {},
+    {
+      select: statusFromResponse,
+      retry: false,
+      refetchOnWindowFocus: false,
+      gcTime: 0,
```
SoloJacobs, line 55: I think we can just remove the `gcTime: 0` here?

SoloJacobs (approved):

ui/mantine-ui/src/data/status.ts
siavashs: We could, but that means if a status is required and is removed by accident from the shared list then we also lose the test case for it.

ui/mantine-ui/src/data/status.ts
siavashs: The status endpoint also returns `Cache-Control: no-store`. It seems keeping `gcTime: 0` matches that freshness policy.

ui/mantine-ui/src/data/connect.ts
```diff
@@ -0,0 +1,16 @@
+import { createConnectTransport } from '@connectrpc/connect-web';
+
+export const getAPIBaseURL = (baseURI: string | URL = document.baseURI) =>
+  new URL('../api', baseURI).toString();
```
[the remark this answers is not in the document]
siavashs: Your assessment is invalid.

ui/mantine-ui/Makefile
```diff
@@ -1,4 +1,15 @@
 SRC_FILES = $(shell find src test-utils -type f)
+PROTO_FILES = $(shell find ../../proto/api -type f -name '*.proto')
```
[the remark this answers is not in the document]
siavashs: We track proto directories now instead of using a manifest.

PR #5541, opened 2026-09-03
Add support for http debugging

jaymzh:
This depends on https://github.com/prometheus/common/pull/985 for the
infrastructure, but adds an --http.debug flag that uses the
prometheus/common debug roundtripper to provide debug logs with
credentials censored.

**NOTE**: Will fail tests until  https://github.com/prometheus/common/pull/985 is merged

### DIFF
--- a/cli/root.go
+++ b/cli/root.go
@@ -42,6 +42,7 @@ var (
 	output          string
 	timeout         time.Duration
 	httpConfigFile  string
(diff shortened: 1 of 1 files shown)


SoloJacobs (commented):
Hi @jaymzh ,

thanks for the PR. Please open an issue for this feature instead explaining, the use case.

Kind regards

PR #5542, opened 2026-09-04
update inhibitor index to store all matched alerts

Spaceman1701:
This fix is based on HRT's inhibitor implementation. It updates the index so that each equal-labels key holds a `FingerprintSet` instead of a `Fingerprint` so that the index stores _every_ matching source alert instead of _one_ matching source alert.

This allows the inhibitor to search through all the alerts in the index for the set of equal-labels to select the best candidate. Worst case, this does degrade inhibitor performance to `O(n)`, but in practice it should be similar to today (since the first element of the set is most likely a viable candidate). 

```release-notes
[BUGFIX] fix various issues with the inhibitor improperly skipping inhibits when the indexed source alert is resolved but other alerts are present.
```

### DIFF
--- a/inhibit/index.go
+++ b/inhibit/index.go
@@ -19,41 +19,63 @@ import (
 	"github.com/prometheus/common/model"
 )
 
-// index contains map of fingerprints to fingerprints.
+// index contains map of fingerprints to sets of fingerprints.
 // The keys are fingerprints of the equal labels of source alerts.
-// The values are fingerprints of the source alerts.
+// The values are the fingerprints of all source alerts sharing those equal labels.
 // For more info see comments on inhibitor and InhibitRule.
 type index struct {
 	mtx   sync.RWMutex
-	items map[model.Fingerprint]model.Fingerprint
+	items map[model.Fingerprint]model.FingerprintSet
 }
 
 func newIndex() *index {
 	return &index{
-		items: make(map[model.Fingerprint]model.Fingerprint),
+		items: make(map[model.Fingerprint]model.FingerprintSet),
 	}
 }
 
-func (c *index) Get(key model.Fingerprint) (model.Fingerprint, bool) {
+// Get returns a copy of the fingerprints indexed under key.
+func (c *index) Get(key model.Fingerprint) ([]model.Fingerprint, bool) {
 	c.mtx.RLock()
 	defer c.mtx.RUnlock()
 
-	fp, ok := c.items[key]
-	return fp, ok
+	set, ok := c.items[key]
+	if !ok {
+		return nil, false
+	}
+	fps := make([]model.Fingerprint, 0, len(set))
+	for fp := range set {
+		fps = append(fps, fp)
+	}
+	return fps, true
 }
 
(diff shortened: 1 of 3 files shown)


SoloJacobs:
```
❯ ./run.sh
=== inhibit
flushes: 238231
Mutes calls: 297539
alerts muted: 197630 (66.42%)
flushes fully muted: 174951 (73.44%)
--- PASS: TestReplay (4.70s)
ok  	github.com/prometheus/alertmanager/inhibit	4.710s
=== patchedinhibit
flushes: 238231
Mutes calls: 297539
alerts muted: 201516 (67.73%)
flushes fully muted: 178227 (74.81%)
--- PASS: TestReplay (4.98s)
ok  	github.com/prometheus/alertmanager/patchedinhibit	4.988s
=== v29inhibit
flushes: 238231
Mutes calls: 297539
alerts muted: 201516 (67.73%)
flushes fully muted: 178227 (74.81%)
--- PASS: TestReplay (10.05s)
ok  	github.com/prometheus/alertmanager/v29inhibit	10.062s
=== 5449updatedinhibit
flushes: 238231
Mutes calls: 297539
alerts muted: 201516 (67.73%)
flushes fully muted: 178227 (74.81%)
--- PASS: TestReplay (5.66s)
ok  	github.com/prometheus/alertmanager/5449updatedinhibit	5.665s
=== 5542updatedinhibit
flushes: 238231
Mutes calls: 297539
alerts muted: 201516 (67.73%)
flushes fully muted: 178227 (74.81%)
--- PASS: TestReplay (4.71s)
ok  	github.com/prometheus/alertmanager/5542updatedinhibit	4.718s
```

inhibit/index.go
```diff
@@ -19,41 +19,63 @@ import (
...
-	return fp, ok
+	set, ok := c.items[key]
+	if !ok {
+		return nil, false
+	}
+	fps := make([]model.Fingerprint, 0, len(set))
+	for fp := range set {
+		fps = append(fps, fp)
+	}
+	return fps, true
```
siavashs: We could probably replace this cloning loop with:
```suggestion
	return slices.Collect(maps.Keys(set)), true
```

inhibit/index.go
```diff
@@ -19,41 +19,63 @@ import (
...
+	}
+	set[value] = struct{}{}
 }
 
-func (c *index) Delete(key model.Fingerprint) {
+// Delete removes value from the set of fingerprints indexed under key,
+// removing the key entirely once the set is empty.
+func (c *index) Delete(key, value model.Fingerprint) {
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
```
siavashs, line 67: I was wondering if the lock can move per set now instead of being global.

siavashs (approved):
LGTM, added few comments.

inhibit/index.go
Spaceman1701: Eh, probably, but it'd be a bit more complex - this function can still delete the entire set, so if we went to a per-set lock, we'd sometimes need to upgrade to a global lock, which is a little annoying.

PR #5551, opened 2026-09-10
docs: clarify that time intervals are not inherited by child routes

### LINKED ISSUE #3211: Clarify lack of route time interval inheritance in docs
The [`<route>`](https://prometheus.io/docs/alerting/latest/configuration/#route:~:text=optional%20configuration%20parameters%20are%20inherited%20from%20its%20parent%20node%20if%20not%20set) docs state _optional configuration parameters are inherited from its parent node if not set_. `active_time_intervals` and `mute_time_intervals` are currently [copied](https://github.com/prometheus/alertmanager/blob/ecb66f76b0058957e5045114ae01f8b98994451e/dispatch/route.go#L63) as part of `RouteOps`, however they are then [overwritten](https://github.com/prometheus/alertmanager/blob/ecb66f76b0058957e5045114ae01f8b98994451e/dispatch/route.go#L120-L121) without the `nil` check that guards other parameters, removing the inheritance. Is this intentional?

The effect is a parent will not implicitly mute its children, e.g. given the following, `a` and `b` will still receive every alert all the time:

```yaml
...
route:
  receiver: default
  routes:
  - routes:
    - receiver: a
      continue: true
    - receiver: b
    mute_time_intervals:
    - everyday
time_intervals:
- name: everyday
  time_intervals:
  - weekdays:
    - sunday:saturday
```

This tripped us up, as we wanted to notify two receivers with the same mute condition, and the only way to do this is to move the mute down into each child, e.g.

```yaml
route:
  receiver: default
  routes:
  - routes:
    - receiver: a
      mute_time_intervals:
      - everyday
      continue: true
    - receiver: b
      mute_time_intervals:
      - everyday
```

If this is intended behaviour, I think it's worth making more explicit in the docs. I'm happy to submit a PR on confirmation.

xenon898:
## What

The `<route>` docs say "optional configuration parameters are inherited from its parent node if not set", but `mute_time_intervals` and `active_time_intervals` are unconditionally set from the current route's config in [`newRoute`](https://github.com/prometheus/alertmanager/blob/main/dispatch/route.go#L137-L138) — they lack the `nil`-check that `group_by`, `group_wait`, `group_interval`, and `repeat_interval` use, so they are never inherited.

This trips users up when they set `mute_time_intervals` on a parent route expecting it to mute the whole subtree (see #3211). This PR documents the actual behavior in `docs/configuration.md`:

- the `<route>` "inherited from its parent" sentence now notes the exception
- the `mute_time_intervals` field comment explains it is not inherited, and how to mute a subtree
- the `active_time_intervals` field comment notes it is not inherited

Docs-only, no functional change.

Closes #3211

Follow-up to #3538 (attempted to change the behavior; closed unmerged).


SoloJacobs (changes requested):
https://github.com/prometheus/alertmanager/pull/5440 Already meantions this: Please don't close the original issue.

xenon898:
Apologies — I did not spot #5440 before opening this. It predates this PR and is more complete (it also lists which parameters *are* inherited), so closing in its favour. Thanks @SoloJacobs; #3211 stays open for #5440 to resolve.

SoloJacobs:
@xenon898 Looks like the other contributors won't cleanup there changes. Do you want to pick them up?

xenon898:
Thanks @SoloJacobs — happy to pick this up.

Before I do, want to confirm direction rather than burn a third attempt on the wrong one. Looking at `newRoute` (dispatch/route.go): every other route option follows the same pattern — start from the parent's resolved value, only overwrite if the child explicitly sets it (`receiver`, `group_by`, `group_wait`, `group_interval`, `repeat_interval`; `labels` additionally merges rather than replacing). `mute_time_intervals`/`active_time_intervals` are the only fields that don't follow this — they're unconditionally set from the child's own YAML, so an omitted field becomes `[]` instead of inheriting the parent's value.

Two ways to close this out:
1. **Docs-only**: accurately document the current behavior (fixing #5440's inaccurate "merging" framing, and dropping any Fixes/Closes trailer since the inconsistency itself remains).
2. **Behavior fix**: make these two fields follow the same inherit-unless-set pattern as the others, matching @benridley's original "this is not intentional" from #3538 — then document the corrected behavior.

Which do you want? Happy to do either.

SoloJacobs:
We will do a docs only change, and keep the issue open. The project is not ready for breaking changes inside the config. Another note: the new receiver labels are also not consistent with the other types of inheritance (as you pointed out). That's just something to keep in mind to not overclaim.

xenon898:
Opened #5565 with the docs-only change — named exactly which fields inherit (receiver, group_by, group_wait, group_interval, repeat_interval), described labels as merged rather than inherited, and mute_time_intervals/active_time_intervals as not inherited at all. No Closes/Fixes trailer, #3211 stays open. Thanks for the direction.

PR #5552, opened 2026-09-10
docs: note that weekday ranges cannot wrap around the week

### LINKED ISSUE #3213: Document first day of week for time intervals

xenon898:
## What

A `weekday_range` such as `saturday:sunday` is rejected by the parser with `start day cannot be before end day` ([timeinterval.go](https://github.com/prometheus/alertmanager/blob/main/timeinterval/timeinterval.go)), because the start day index must be less than or equal to the end day index (Sunday = 0 … Saturday = 6). The docs don't mention this, so users try to express a weekend as `saturday:sunday` and hit a config error (see #3213).

This adds a sentence to the `weekday_range` description in `docs/configuration.md` explaining that ranges cannot wrap around the end of the week, and that individual days should be listed instead.

Docs-only.

Closes #3213


xenon898:
Duplicate of #5172, which is already approved and awaiting merge. Apologies for the noise — closing in its favour.

PR #5556, opened 2026-09-11
Prepare 0.34.1

Spaceman1701:
Prepare 0.34.1 with the inhibitor fixes from #5542, #5449, #5559 and dependency updates from #5468


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## main / (unreleased)
 
+## 0.34.1 / 2026-09-11
+
+* [BUGFIX] fix various issues with the inhibitor improperly skipping inhibits when the indexed source alert is resolved but other alerts are present.
```
SoloJacobs, line 5: - We uppercase the start of these.
- link the PR number at the end.
- Optionally have the `inhibit` component at the start.

Could we also make it a bit more user-facing?
```
[BUGFIX] inhibit: Fix alerts not being muted when a source alert resolves while other matching source alerts are still firing. #5542
```

SoloJacobs (changes requested):

SoloJacobs (approved):

PR #5561, opened 2026-09-14
Muted alerts logic

suprjinx:
This PR builds on the work of PR #5503 to implement the changed notification logic for muted alerts.

The general approach follows the recommendation of @Spaceman1701, revised a bit in the `#promethus-alertmanager` slack channel. Alerts are now in a binary state of "firing" or "resolved", while maintaining a separate set of "muted" alerts. We store the same alert up to twice - it could be firing or resolved, and additionally muted in either case.

Relates-to: #5247 
Relates-to: #226 

#### Pull Request Checklist

    - Fixes #<issue number>
    
- Is this a bugfix?
- Is this a new feature?
- Is this a breaking change?

Assisted-by: Claude Code


notify/dedup_stage.go
```diff
@@ -95,6 +102,140 @@ func (n *DedupStage) needsUpdate(entry *nflogpb.Entry, firing, resolved map[uint
...
+		return ReasonNewResolvedAlerts
+	}
+
+	// The group is still firing but none of it can be shown, and there is
+	// nothing else to say, so the sequence closes as muted, not as resolved.
+	if len(visibleFiring) == 0 {
+		if len(notifiedFiring) > 0 {
+			return ReasonAllAlertsMuted
+		}
+		return ReasonDoNotNotify
```
Spaceman1701, line 236: I'm not sure this is right - if all alerts in a group are muted, we'll fall into this branch even if the repeat interval has elapsed. 

Since nflog entries eventually expire, we'll end up losing the state of the group. I think this needs to be handled somewhat like `send_resolved` where the `RetryStage` has some special logic that handles an alert group which is fully muted for integrations that don't know how to handle mutes themselves: https://github.com/prometheus/alertmanager/blob/b4095436688a51272959ce4ed8e41103d71b2cfb/notify/retry_stage.go#L95-L113

It is a bit more complex, however, since we need to handle the case of "is now fully muted, but wasn't before" and "continues to be fully muted" separately. Still, I _think_ the `NotificationSequence` exposes enough information to do that.

notify/event.go
```diff
@@ -93,7 +110,7 @@ func newNotificationEvent(ctx context.Context, alerts []*alert.Alert, integratio
 	firingHashes, _ := FiringAlerts(ctx)
 	resolvedHashes, _ := ResolvedAlerts(ctx)
 	details := alertDetailsByHash(alerts)
-	muted, _ := mutedAlertDetails(ctx)
+	muted := mutedAlertDetails(ctx)
```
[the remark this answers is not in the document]
Spaceman1701: I'm not sure whether the bot is correct or not...

notify/muted_alerts_test.go
```diff
@@ -353,3 +371,318 @@ func TestDedup_MutedAlertBreaksNotificationSequence(t *testing.T) {
...
+	require.Equal(t, ReasonFirstNotification, reason)
+	require.Equal(t, []*alert.Alert{a}, notified)
+	require.Equal(t, 1, p.writes)
+	firstWrite := p.entry.Timestamp.AsTime()
+
+	// Nothing changed and the repeat interval has not elapsed.
+	notified, reason, _ = p.flush(base.Add(10*time.Minute), a, b)
+	require.Equal(t, ReasonDoNotNotify, reason)
+	require.Empty(t, notified)
+	require.Equal(t, 1, p.writes, "a flush that notifies nobody should not write to the notification log")
```
Spaceman1701, line 414: This is related to my comment above, but we do sometimes want to write the nflog when there's nothing to notify about - I think specifically when repeat_interval has elapsed.

notify/muted_alerts_test.go
```diff
@@ -353,3 +371,318 @@ func TestDedup_MutedAlertBreaksNotificationSequence(t *testing.T) {
...
+		muted:  []uint64{1},
+		want:   ReasonDoNotNotify,
+	}, {
+		name:     "nothing to notify about when a group only ever resolved",
+		resolved: []uint64{1},
+		want:     ReasonDoNotNotify,
+	}, {
+		// The receiver was shown the group, so it is told the group is over,
+		// even though the resolved alert is now muted.
+		name:     "muted alert resolves the group the receiver was told about",
```
Spaceman1701: I just want to call out that if we were to enable this, it would be a behavior change. It might be acceptable in the sense that it only can generate resolved notifications for groups that sent a notification before being muted, but some for some users it may be unwelcome.

Spaceman1701 (commented):
This looks really good! I think there's a small problem with the way continuously muted groups are handled, but otherwise this seems right to me.

notify/context.go
```diff
@@ -45,7 +45,7 @@ const (
 	keyFlushID
 	keyGroupMatchers
 	keyRouteLabels
-	keyMutedAlertDetails
+	keyNotificationSequence
```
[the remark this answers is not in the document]
suprjinx: not in use

notify/notify_test.go
```diff
@@ -948,7 +959,7 @@ func TestMutedGroupIsRecordedInNflog(t *testing.T) {
 			alert := &types.Alert{Alert: model.Alert{Labels: model.LabelSet{"alertname": "muted"}}}
 			stage := test.newStage([]Stage{
 				muteAllStage{},
-				NewDedupStage(&integration, tnflog, recv),
+				NewDedupStage(&integration, tnflog, recv, test.ff),
```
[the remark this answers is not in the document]
suprjinx: all calls are correct

notify/event.go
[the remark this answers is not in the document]
suprjinx: > I'm not sure whether the bot is correct or not...

I think the code is correct, bot incorrect -- we are building details for what's visible, and we don't want to include muted. I will rename the vars so it's clearer.

notify/muted_alerts_test.go
suprjinx: Should we toggle this with the feature flag?

notify/dedup_stage.go
suprjinx: Good catch. I added a new `ReasonStillMuted` for the continuing case. ReasonAllAlertsMuted no longer returns early — it writes the entry that closes the sequence — and on later flushes, ReasonStillMuted rewrites that entry on each repeat interval so it stays ahead of the expiry. Does that capture the nuance?

notify/muted_alerts_test.go
suprjinx: Now updating the nflog entry in this case, via ReasonStillMuted handling.

notify/muted_alerts_test.go
Spaceman1701: We discussed this a bit in the Alertmanager working group meeting, and the conclusion there is that we cannot introduce this behavior without an option to toggle it off.

I don't think a feature flag is fine grained enough. I'd prefer to handle this is _per-integration_, like `send_resolved`. Essentially, we need some flag per-integration that indicates that a receiver _always_ wants resolved notifications, regardless of whether the group is muted. 

I actually think we could generalize this a little bit, and give the user a per-notification-reason filter:
```yaml
receiver: my_receiver
slack_configs:
  - reason_filters:
       all_alerts_resolved: true
       alerts_added_to_group: false
       all_alerts_resolved_muted: true
```
but this is both a little awkward for users and a little too much to add to the scope of this change. 

So for lack of a better idea, maybe we could just add a new `always_send_resolved` flag on the integration config? When this is `true`, notifiers would receive resolved messages for groups that are muted. When `false`, they would have identical behavior as today.

For implementation, I think this can be done in the `RetryStage`, similar to the handling of `send_resolved` today: based on the value of the flag, the actual call to `Notify` can be short-circuited.

notify/dedup_stage.go
Spaceman1701: Yeah, I think this works. One small question: When we get to the branch on line 233, are we guaranteed that the group has _any_ alerts in it (muted or otherwise)?

notify/retry_stage.go
```diff
@@ -92,22 +92,24 @@ func (r RetryStage) Exec(ctx context.Context, l *slog.Logger, alerts ...*alert.A
...
 		}
-		if len(firing) == 0 {
-			return ctx, alerts, nil
-		}
 		for _, a := range alerts {
 			if a.Status() != model.AlertResolved {
 				sent = append(sent, a)
 			}
 		}
+		if len(sent) == 0 {
```
Spaceman1701, line 110: if we're in the `ReasonStillMuted` case, do muted alerts end up in the `sent` list? Based on my reading, they do, but I might be missing something. If that's the case, I think notifiers will be shown a list of muted alerts, which is probably wrong (unless all notifiers are updated to handle this case).

I think what we want to do here is handle notifier mute awareness here, and use this short-circuit case to provide "default" behavior for notifiers that don't handle muted alerts (or configured not to).

notify/set_notifies_stage.go
```diff
@@ -74,6 +71,16 @@ func (n SetNotifiesStage) Exec(ctx context.Context, l *slog.Logger, alerts ...*a
 	)
 	defer span.End()
 
+	// With the feature enabled this stage is reached on flushes that deliver
+	// nothing, and most of them are still worth recording: a muted group has
+	// gone quiet, and the entry has to say so. A flush the dedup stage found
+	// nothing to say about is the exception, since rewriting the entry would
+	// refresh its timestamp and defer the repeat interval forever.
```
Spaceman1701: this has actually always been the case - when a receiver has `send_resolved: false`, we'll reach this function for a group that delivered nothing. I think we should try to write the pipeline so that we only reach this stage if we intend to write an nflog entry.

notify/dedup_stage.go
suprjinx: at line 233, we must have at least one firing alert and all alerts are muted.

notify/muted_alerts_test.go
suprjinx: @Spaceman1701 makes sense -- perhaps the new flag should mention muting, eg `send_resolved_when_muted: true|false` or `muted_group: ignore|resolve`?

notify/retry_stage.go
suprjinx: I believe the incoming `alerts` param are only the visible alerts (muted have been subtracted by muting stages). So those are the only candidates for the `sent` list.

notify/set_notifies_stage.go
suprjinx: good idea, removing this logic and placing the decision in `notify.execStages`

PR #5565, opened 2026-09-15
docs: clarify which route options are inherited by child routes

### LINKED ISSUE #3211: Clarify lack of route time interval inheritance in docs

xenon898:
## What

The `<route>` docs claimed all optional parameters are inherited from the parent if not set, but that's only true for `receiver`, `group_by`, `group_wait`, `group_interval`, and `repeat_interval`. `labels` are merged (parent + child) rather than inherited-or-replaced, and `mute_time_intervals`/`active_time_intervals` are not inherited at all — [`newRoute`](https://github.com/prometheus/alertmanager/blob/main/dispatch/route.go) sets them unconditionally from the child's own config, so an omitted field is simply empty regardless of the parent.

This documents the actual behavior, phrased to name exactly which fields inherit rather than claiming a blanket rule.

## Scope

Per @SoloJacobs on #5551: **docs-only** for now — a behavior change here would be a breaking config change the project isn't ready to make. This PR does **not** close #3211, since the underlying inconsistency itself remains unresolved; only the documentation is being corrected.

Supersedes #5440 and #5551 (both stalled/closed).

#### Pull Request Checklist

- Related open issue: #3211 (discussed with @SoloJacobs in #5551 — intentionally not using `Fixes`, since this PR only corrects the docs; the underlying behavior inconsistency in #3211 is still open)
- Is this a bugfix? It corrects incorrect documentation; there's no code behavior to add a test against
- Is this a new feature? No

Documentation-only correction, no behavior change.

```release-notes
[BUGFIX] docs: clarify which route options are inherited by child routes
```

---


SoloJacobs (approved):

SoloJacobs:
Hi,

looks good to me :+1: Thanks for the contribution. By the way: You can make the Commit messages a bit shorter, they are not so useful, if nobody reads them. So only the most important details should be included there.

kind regards

SoloJacobs:
Do you want to pick up https://github.com/prometheus/alertmanager/pull/5172 as well? We shouldn't use that code, since was not signed-off on.

xenon898:
Yes, happy to pick up #5172 as well. The content there looks already reviewed and approved (by you and @TheMeier), so I'll reuse it as-is and just fix the sign-off, rather than rewrite it, since the wording review is already settled. Will open a fresh PR once that's done.Also, noted on the commit messages. Will keep them shorter going forward.

PR #5566, opened 2026-09-16
docs: clarify weekday_range does not wrap around the week

### LINKED ISSUE #3213: Document first day of week for time intervals

xenon898:
Reuses the content from #5172 (already approved by @TheMeier and @SoloJacobs), which was blocked only on a missing DCO sign-off. This PR carries the same change with a signed-off commit.

Clarifies that Alertmanager follows Go's `time.Weekday` convention (Sunday = 0, Saturday = 6), differing from ISO 8601, and that `weekday_range` values can't wrap around the week boundary (`saturday:sunday` silently matches nothing; use `['saturday', 'sunday']` instead).

Fixes #3213

#### Pull Request Checklist

- Fixes #3213
- Is this a bugfix? Docs-only, no code behavior to test
- Is this a new feature? No

```release-notes
[BUGFIX] docs: clarify weekday_range does not wrap around the week
```

---


SoloJacobs (changes requested):
Welp, I should have said: You cannot use the code from PR in any shape or form. The sign-off is necessary, so we have a form of legal attestation for the code in question. Please only have a look at the original issue opened by the user: https://github.com/prometheus/alertmanager/issues/3213

I think me might have to find another person, which has not read the PR text ... 😢

xenon898:
Understood, apologies. I misread "not signed off" as a missing formality rather than an authorship/attestation issue. Closing this PR.I've already read the original PR's text, so I won't attempt a reworded version myself. I can't honestly guarantee it wouldn't be influenced by it. Thanks for catching this before it went further.

PR #5571, opened 2026-09-17
docs: document OAuth2 support for jira_config's http_config

### LINKED ISSUE #5475: Jira integration: Support for API token auth method
### Proposal

Currently the Jira integration only support PAT - personal access tokens.
This is not ideal and not the correct way to do system to system integrations.
Are there any plans (soon) to support this?

xenon898:
The `jira_config` docs only described PAT-based auth (`basic_auth` / `authorization`) for `http_config`, even though it's the same shared `<http_config>` type every notifier uses and already supports OAuth2 (verified against `OAuth2` in `prometheus/common@v0.70.1`: `client_id`, `client_secret_file`, `token_url`, `scopes` are all real fields). That's what #5475 ran into.

Adds a line noting OAuth2 is supported, and a worked example. The example uses independently-chosen placeholder values, not the one posted in the issue thread, since that suggestion was given as a plain comment rather than a signed-off contribution.

Fixes #5475

#### Pull Request Checklist

- Fixes #5475
- Is this a bugfix? Documentation-only, no code behavior to test
- Is this a new feature? No

```release-notes
[BUGFIX] docs: document OAuth2 support for jira_config's http_config
```

---


SoloJacobs (changes requested):
Hey,

I'm afraid this change is doing a bit more than it should:
* Jira doesn't actually support like OAuth2 without clicking in the browser. This means Alertmanager cannot do this either. What we can actually do is specify an idp via http config.
* The example defines a non-existing scope. We should probably remove that example altogether and point users to general http_config.

So, please make the change a bit more minimal. The main goal is just to check what may lead users to conclude one can't use OAuth with an idp. The OAuth solution by jira does not work for automations.

xenon898:
Pushed an update: removed the worked example entirely and corrected the explanation. It now says OAuth2 client_credentials works via an external identity provider through the standard http_config, and explicitly notes Jira's own OAuth2 needs interactive browser consent and can't be used for Alertmanager's automated notifications. Thanks for the correction.

SoloJacobs (approved):

SoloJacobs (approved):

Issue #5479, opened 2026-08-29
API v3alpha: add shared contracts and reusable infrastructure


siavashs:
Parent: #5450

## Summary

Establish the shared `common.v3alpha` protobuf contracts and reusable Connect API infrastructure needed by AlertService, SilenceService, ReceiverService, and NotificationService.

Public handlers and conversion code must use internal Alertmanager types rather than generated API v2 OpenAPI models.

## Shared protobuf contracts

Add `common.v3alpha` messages for:

- `Matcher`.
- OR-of-AND `MatcherSet`.
- Receiver identity and labels where shared between services.
- State-filter primitives.
- Pagination request and response fields.
- Structured validation details.
- Partial-result details.
- Required-feature details.

Keep this package minimal because it becomes a dependency of independently versioned services.

## Pagination

Define one live-view pagination contract for list RPCs:

- Deterministic resource-specific sort keys.
- A unique ID or fingerprint tie-breaker.
- Opaque tokens bound to resource type, sort cursor, and filter digest.
- Default and maximum page sizes.
- `InvalidArgument` for negative sizes and malformed, mismatched, or expired tokens.
- Explicit documentation that concurrent mutation may move records between pages and tokens do not provide snapshot isolation.

## Reusable infrastructure

- Add matcher, filtering, conversion, pagination, and token helpers over internal Alertmanager types.
- Wire `featurecontrol.Flagger` through app, API, and Connect options.
- Support disabled-by-default capability flags that reject incompatible operations with structured `FailedPrecondition` details naming the required feature.
- Expand the Connect options structure for alert, silence, group, receiver, feature, and future notification dependencies.
- Publish one immutable reload snapshot containing configuration, routes, receiver metadata, and the reload-safe alert-status callback.
- Never hold configuration or reload locks while scanning stores or sending stream messages.
- Centralize stable Connect error translation for `InvalidArgument`, `NotFound`, `ResourceExhausted`, `Unavailable`, `FailedPrecondition`, `Aborted`, and `Internal`, while preserving cancellation and deadline codes.

## Acceptance criteria

- [ ] `common.v3alpha` source and generated bindings are Buf-compliant and reproducible.
- [ ] Shared messages have documented ownership and compatibility semantics.
- [ ] Pagination tokens are opaque, filter-bound, deterministic, and covered for invalid inputs.
- [ ] Capability gating returns machine-readable `FailedPrecondition` details.
- [ ] Reload snapshots are immutable and race-safe.
[...]

PR #5578, opened 2026-09-18
feat(api): add common v3alpha

siavashs:
Define `common.v3alpha` contracts for:
- matchers
- OR-of-AND matcher sets
- receiver metadata
- resource states
- pagination
- validation errors
- partial results
- required features

Add shared matcher conversion, filtering, signed page tokens, Connect error translation, and capability gates.

Wire alert, silence, group, notification, and feature dependencies into the Connect API.

Publish configuration, routes, receiver metadata, and the alert-status callback through one immutable reload snapshot after the inhibitor and dispatcher finish loading.

Store enabled features in one map as a cleaner solution.

Part of #5479


api/connect/common_test.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/prometheus/common/model"
+	"github.com/prometheus/common/promslog"
+	"google.golang.org/protobuf/proto"
+
+	commonv3alpha "github.com/prometheus/alertmanager/api/common/v3alpha"
+	"github.com/prometheus/alertmanager/featurecontrol"
+	"github.com/prometheus/alertmanager/pkg/labels"
+)
+
+var _ = Describe("shared matchers", func() {
```
ultrotter, line 33: Do we really need to write tests this way, and import an external library for it? I mean golang has a test framework embedded and it works. What's the advantage?

PR #5579, opened 2026-09-20
feat(config): allow weekday ranges to wrap around the end of the week

### LINKED ISSUE #3213: Document first day of week for time intervals

TheMeier:
#### Pull Request Checklist

    - Fixes #3213
    
- Is this a new feature?

```release-notes
[ENHANCEMENT] CONFIG: allow weekday ranges to wrap around the end of the week
```


SoloJacobs (approved):
I like it! Technically a new feature, so it should be signed off in a meeting, or at least approved in our slack channel. We can also consider removing the first day of the week from the docs, since it is no longer relevant.

ultrotter (approved):

